Issue #1517, opened 2017-12-13
Stabilization branch: Xbox controller does not work in GamepadTest.unity scene


NeerajW:
## Overview
Connected Xbox controller.
Unity remoting the test scene into immersive HMD.

## Expected Behavior
A press makes cube bigger.

## Actual Behavior
Nothing happens.

## Steps to reproduce
_(Links to sample github project preferred)_

keveleigh:
Looks like GitHub didn't automatically close these issues when merged into master. I'll close these manually and look into why.

PR #1521, opened 2017-12-14
XboxControllerHandlerBase fixes

keveleigh:
Overview
---
Simplified the state machine in `XboxControllerHandlerBase` to reduce confusion and cases where states may conflict with each other.
Also resolved a bug where `GamePadHandlerBase` was registering as a global listener twice, thus receiving all events from the `InputManager` twice.

Changes
---
- Fixes: #1517

### DIFF
--- a/Assets/HoloToolkit-Examples/Input/Scripts/XboxControllerHandlerTest.cs
+++ b/Assets/HoloToolkit-Examples/Input/Scripts/XboxControllerHandlerTest.cs
@@ -25,7 +25,7 @@ namespace HoloToolkit.Unity.InputModule.Tests
         private Vector3 newPosition;
         private Vector3 newRotation;
 
-        private void Start()
+        protected virtual void Start()
         {
             initialPosition = transform.position;
         }
--- a/Assets/HoloToolkit-Examples/Input/Scripts/XboxControllerHandlerTest.cs
+++ b/Assets/HoloToolkit-Examples/Input/Scripts/XboxControllerHandlerTest.cs
@@ -25,9 +25,8 @@ namespace HoloToolkit.Unity.InputModule.Tests
         private Vector3 newPosition;
         private Vector3 newRotation;
 
-        protected override void Start()
+        private void Start()
         {
-            base.Start();
             initialPosition = transform.position;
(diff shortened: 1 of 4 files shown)


Assets/HoloToolkit/Input/Scripts/GamePad/XboxController/XboxControllerHandlerBase.cs
```diff
@@ -56,22 +49,21 @@ public virtual void OnXboxInputUpdate(XboxControllerEventData eventData)
...
                 HandleNavigation(eventData);
-
-                if (!HoldStarted && !RaiseOnce && !NavigationStarted)
-                {
-                    HandStartedRoutine = StartCoroutine(HandleHoldStarted(eventData));
-                }
             }
 
-            if (XboxControllerMapping.GetButton_Up(SelectButton, eventData))
+            if (eventData.XboxA_Up)
```
StephenHodgson: this removes the customization of the selection buttons

keveleigh: Ahh yes, this was left over from when I was investigating the double events. Reverting.

Assets/HoloToolkit-Examples/Input/Scripts/XboxControllerHandlerTest.cs
```diff
@@ -25,9 +25,8 @@ public class XboxControllerHandlerTest : XboxControllerHandlerBase
         private Vector3 newPosition;
         private Vector3 newRotation;
 
-        protected override void Start()
+        private void Start()
```
StephenHodgson: We should probably make this `protected virtual` in case someone wants to add onto the functionality of this class

keveleigh: Since it's a test class, that seems unlikely, but I'll change it.

StephenHodgson: Good point

Issue #1505, opened 2017-12-11
BoundaryManager should be disabled on HoloLens


timGerken:
## Overview
This code will fail when called on HoloLens. End result is the floor quad gets instantiated at (0,0,0) which is sub-optimal on a non-opaque device.
https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/f2596ee39d740f2eab1209c869ea1e55d0478a49/Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs#L87

## Expected Behavior
When on HoloLens, boundary manager is disabled

## Actual Behavior
When on HoloLens, boundary manager draws a floor quad at (0,0,0)

## Steps to reproduce
add MixedRealityCameraParent to a new unity project and run on HoloLens.

## Unity Editor Version
2017.2.0p2-MRTP5

## Mixed Reality Toolkit Release Version
branch v2017.2.0.1-Stabilization
SHA f2596ee39d740f2eab1209c869ea1e55d0478a49

timGerken:
Updated issue as this occurs on both RS1 and RS4 based HoloLens

keveleigh:
Looks like GitHub didn't automatically close these issues when merged into master. I'll close these manually and look into why.

PR #1526, opened 2017-12-14
Removing BoundaryManager from HoloLens

keveleigh:
Overview
---
If a HoloLens is detected, now the `BoundaryManager` script is disabled.

Changes
---
- Fixes: #1505.

### DIFF
--- a/Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
+++ b/Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
@@ -84,7 +84,11 @@ namespace HoloToolkit.Unity.Boundary
             }
             else
             {
-                XRDevice.SetTrackingSpaceType(transparentTrackingSpaceType);
+                // Removed for now, until the HoloLens tracking space type story is more clear.
+                //XRDevice.SetTrackingSpaceType(transparentTrackingSpaceType);
+
+                enabled = false;
+                return;
             }
 
             // Render the floor based on if you are in editor or immersive device.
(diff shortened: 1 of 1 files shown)


Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
```diff
@@ -84,7 +84,11 @@ protected override void Awake()
             }
             else
             {
-                XRDevice.SetTrackingSpaceType(transparentTrackingSpaceType);
+                // Removed for now, until the HoloLens tracking space type story is more clear.
+                //XRDevice.SetTrackingSpaceType(transparentTrackingSpaceType);
+
+                enabled = false;
```
StephenHodgson: Should we destroy the component instead of disabling, like we do in other scripts?

keveleigh: My hesitation to destroy was to preserve the Singleton instance, but if devs are properly checking `IsInitialized` and whatnot, that should be okay.

Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
```diff
@@ -84,7 +84,11 @@ protected override void Awake()
...
+                return;
```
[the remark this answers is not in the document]
keveleigh: Bounds isn't nullable, so it won't nullref
image

Addressing Stephen's comment below about destroying instead of disabling will prevent the method from being called as well.

Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
StephenHodgson: That's a good point as well.

PR #1532, opened 2017-12-15
Updated initialized this object in structs

StephenHodgson:
Overview
---
Fixed issue with `RayStep` and `Vector3Smoothed` struct not initializing this() before assigning values in constructor.

Also public fields were not PascalCase.


Assets/HoloToolkit/Input/Scripts/Focus/RayStep.cs
```diff
@@ -9,43 +9,43 @@ namespace HoloToolkit.Unity
     [Serializable]
     public struct RayStep
     {
-        public RayStep(Vector3 origin, Vector3 terminus)
+        public RayStep(Vector3 origin, Vector3 terminus) : this()
```
keveleigh: This looks like a good change, but, to help me understand, what is this fixing in this case? I was under the impression ` : this()` was only needed if not all properties are set inside the constructor. In this case, all four are set.

StephenHodgson: `this` is required _because_ we're setting the values in the constructor.

Idk, I just keep getting a compiler warning about it, and figured we should heed the warnings.

keveleigh: Do you happen to have the warning? I can't seem to repro it.

Definitely agreed on fixing warnings though :)

StephenHodgson: image

Railboy: Better safe than sorry. Though I can't seem to raise the warning either - I've tried 4.6, 3.5, 2.0 and 2.0 subset.

NeerajW: Wow could not repro either :) Perhaps Stephen has some flags turned on in VS?

StephenHodgson:
Here's another struct that needs the same treatment
image

NeerajW:
@keveleigh can we get this merged?

Issue #1548, opened 2017-12-19
2017.3.0f3 Released


StephenHodgson:
[Unity Engine 2017.3.0f3](https://unity3d.com/unity/whats-new/unity-2017.3.0) was released today.

Windows Mixed Reality Specific Changes:
---
- XR: Added a new option in the Windows MR Player Settings called Enable Depth Buffer Sharing. This allows the OS to better stabilize images without the need to manually set the focus plane. For more information about the benefits of image stabilization, see Microsoft's documentation on Hologram Stability.
- XR: Added XR support to Experimental Scriptable Render Pipelines.
- XR: Improved the way XR devices pause/unpause applications, particularly when pausing as a result of application focus being lost. Unity now handles overlapping pause conditions correctly, regardless of the order in which they occur.
- XR: Removed padding on double-wide texture used with Single Pass.
Windows: Added support for proxy setup on UWP.
- XR: Added support for Single-Pass stereo rendering on MacOS.
- XR: Added XRSettings.useOcclusionMesh to allow for disabling the occlusion Mesh.
- XR: Improved rendering performance of Daydream and Gear VR apps.
- XR: Improved VR Terrain Trees: Fixed billboard trees rotating all the time as viewer moves in VR, better transition of 2D to 3D trees, fast "sort-independent" trees with alpha-to-coverage blending and MSAA , billboard Texture aliasing improvements. (733507, 946544)
- XR: Updated Google VR NDK to 1.80.
- XR: Updated Vuforia to version 7.0.36
- XR: XR now supports jittered projection matrices (for use with TAA). (873685)
- XR: Added PlayerSettings Get/SetVirtualRealitySupported, Get/SetVirtualRealitySDKs, and GetAvailableVirtualRealitySDKs for aquiring and setting the XR Settings Virtual Reality Supported toggle and Virtual Reality SDKs list.
- XR: RenderViewportScale is now disabled in Deferred Rendering.
- Universal Windows Platform: Fix building project when its product name ends with a space or contains unicode characters (885265, 942554, 945345)
- Universal Windows Platform: Fixed "AssemblyConverter running on already processed assemblies. There is something wrong with your build." error message when building on top of a built project that had its UnityOverwrite.txt file deleted (940463)
- Universal Windows Platform: Fixed an inconsitent with other platforms error message "Script Error (script name): Update() can not take parameters." when those magic methods are defined in a certain order on .NET scripting backend (950025)
[...]

Issue #1398, opened 2017-11-19
Can not debug Hololens UWP-Applications


brean:
## Overview
My Hololens/MRTK-Applications are crashing and the output is not showing why, I am also not seeing any debug log.
This might be a bug in Unity and/or Visual Studio.

## Expected Behavior
Visual Studio should show the exact position  the Code and the full log of the stacktrace.

## Actual Behavior
The Program is compiled and run as Unity-UWP-Project from Visual Studio. When an error occurs I see the error message "A debugger is attached to <your app>.exe but not configured to debug this unhandled exception".

## Steps to reproduce
create a new Unity-Project copy the Mixed Reality Project into it, use the build Menu to add all MR and UWP settings and add a script that has an runtime error to your scene.
Build the app from File > Build settings as unity project, activate the "Unity C# Projects" and "Development build". After building open the Project with Visual Studio and run it as "Debug", "x86" on your Local Machine

## Unity Editor Version
2017.2.0f3

## UWP-SDK Version 
10.0.16299.0

## Visual Studio Version
15.4.0 (Community 2017)

## Mixed Reality Toolkit Release Version
current master https://github.com/Microsoft/MixedRealityToolkit-Unity/commit/2c59d07b574cbde904b7ad6bd2e249438597440d

samharper94:
For HoloLens apps, you need to run "Release", "x86", on Remote Machine and Run without Debugging (Ctrl+F5)

brean:
@samharper94 I want to debug my code to see why my app crashed, so "Release" build is not an option for me and I can not run it as remote without a remote device. Without the debugger it can be very frustrating/impossible finding some bugs.

Running the app locally on my notebook worked with older Unity Versions and UWP 15xxx, but since the update to the current Unity Version with the latest UWP and the Windows-Fall Creators update it does not work anymore. Because I updated everything in one session I don't know what exactly broke this functionality.

brean:
I just created a small Unity-Project without MRTK, put a "throw Exception"-script in it build it as UWP application with Hololens as target device and run it from Visual Studio and it worked: The debugger halted at the exception. So it has to be something that the toolkit does... But I have absolutely no idea how to find this because its the debuger itself that causes the problems.

brean:
[...]

PR #1549, opened 2017-12-19
Updated Editor to 2017.3.0p2

StephenHodgson:
- Fixes: #1552 Removed previous Editor compatibility.
- Fixes: #1548 Updated the Editor to 2017.3.0p2
- Fixes: #1398 Fixes debugging on HoloLens
- Shader Upgrade NOTE: upgraded instancing buffer 'Props' to new syntax.
- Remove stabilization plane logic
- Add app setup wizard support for enabling the shared depth buffer.
- Remove duplicate SDK verification in build window utilities.

TODO
---

- [x] Write a check to see if shared depth buffer is enabled when the editor starts and UWP build target is specified.

Known Issues
---

[`WWW` and `UnityWebRequests` have false positive errors for UWP apps built against x86](https://issuetracker.unity3d.com/issues/wsa-isnetworkerror-always-return-true-when-running-an-uwp-app-on-86x-architecture).


Assets/HoloToolkit/UX/Shaders/3DTextShader.shader
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+// Upgrade NOTE: upgraded instancing buffer 'Props' to new syntax.
```
timGerken: Do we need to keep these upgrade notes?

StephenHodgson: I thought they might be useful, but i can removed if needed.

timGerken: Maybe add the comment to the PR description.
Every time Unity goes and changes their shader syntax, these notes get littered throughout the affected shaders. Over time, they lose their cachet.

DDReaper:
Are these shader changes, breaking changes?  So what happens if you use an older version of 2017?
P.S.
I was going to ask whether we were updating :S

timGerken:
@DDReaper Good Question. It looks like these changes were introduced in [2017.3.0b4](https://unity3d.com/unity/beta/unity2017.3.0b4)...

StephenHodgson:
Indeed.  That is a good point.

StephenHodgson:
Yup. It's a breaking change.
image

StephenHodgson:
If that's the case we should just drop all the support for prev versions of the editor as well.

StephenHodgson:
Going to reopen this when I finish the work needed for 1552

StephenHodgson:
image

DDReaper:
We really should remove the ProjectVersion file and just keep the version tag on the github page stating supported versions.
Seen this done in other Unity projects to avoid constantly having to update the file.

StephenHodgson:
I agree, but some people made a good point, that in the unitypackage, there's nothing that really states what the current MRTK version is, and when people inherit project's they're sometimes left wondering.

DDReaper:
I created a "Package.Json" file in the root of some of my packages to resolve that, using standard notation.   It was suggested to me by the community
e.g.
https://bitbucket.org/UnityUIExtensions/unity-ui-extensions/src/d3671c3711b8e4078fdb271a91fc14e84c363611/package.json?at=master&fileviewer=file-view-default

StephenHodgson:
@NeerajW was this the option you were talking about during the shiproom meeting?
image

StephenHodgson:
@NeerajW I looked at figuring out how to update the Depth Buffer Sharing in the editor, but it seems like Unity did not provide an easy way to do so when setting the Virtual Reality SDK.

It also seems that this is not enabled by default.

keveleigh:
@DDReaper That Package.Json looks way nicer than the [txt file](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/Dev_Working_Branch/Assets/HoloToolkit/MRTKVersion.txt) I added 😆

I don't think I agree with removing ProjectVersion.txt though, since Unity can use that to update project settings, APIs, shaders, etc between versions, like we saw when moving to 2017.3. I'm not sure I trust it to guess which version the project is currently on before attempting to update the project.

DDReaper:
Other projects I've worked with @keveleigh removed it because Unity automatically does that upgrade regardless.  When it does a library rebuild, it automatically runs the script updater to check no matter what is recorded in the projectversion file.

The only thing that file does is to inform the Project Wizard what it was last built against :D 

Plus, in the asset, we don't use it anyway

keveleigh:
Sweet, I think this is just missing the readme updates where the version is referenced. Otherwise, looks good!

StephenHodgson:
Still waiting to address the Stabilization Plane.

StephenHodgson:
Blocked until stabilization plane is removed and app setup wizard supports enabling the shared depth buffer

StephenHodgson:
Something else to consider is the verifying installed SDK reflection methods have been changed as well too.

StephenHodgson:
@NeerajW , @keveleigh This PR is now ready for review.

Tested out the HoloLens version with the shared depth buffer. Couldn't really tell the difference. 

Maybe others can test this out as well.

StephenHodgson:
After a bit of testing I ran into issues with the camera.
```
Failed to get rendering parameters for main camera - view transforms may get out of sync! Error code: '0x80004002'
 
(Filename: C:\buildslave\unity\build\Runtime/VR/HoloLens/VRDeviceHoloLens.cpp Line: 554)
```

PR #1554, opened 2017-12-19
Tile Generation Update

StephenHodgson:
Overview
---
- Fixes some issues with generating duplicate textures, when we can just reuse the ones we have, and issues where we were multiplying by zero instead of 1.
- Added Windows Phone icon settings back into the generation

### DIFF
--- a/Assets/HoloToolkit/Utilities/Scripts/Editor/SetIconsWindow.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/Editor/SetIconsWindow.cs
@@ -255,8 +255,18 @@ namespace HoloToolkit.Unity
 
             if (iconSize.x == 1240 && iconSize.y == 1240 || iconSize.x == 2480 && iconSize.y == 1200) { return texturePath; }
 
-            string filePath = string.Format("{0}/{1}_{2}x{3}.png", _outputDirectoryName, type.ToString(), iconSize.x, iconSize.y);
+            string filePath = string.Format("{0}/{1}_AppIcon_{2}x{3}.png", _outputDirectoryName, Application.productName, iconSize.x, iconSize.y);
+            if (iconSize.y == 3200)
+            {
+                Debug.LogWarning(type.ToString() + "Scale: " + scale);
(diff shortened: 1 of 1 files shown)


Assets/HoloToolkit/Utilities/Scripts/Editor/SetIconsWindow.cs
```diff
@@ -251,13 +251,19 @@ private static string CloneAndResizeToFile(PlayerSettings.WSAImageType type, Pla
             if (string.IsNullOrEmpty(texturePath)) { return string.Empty; }
 
             var iconSize = GetUWPImageTypeSize(type, scale);
+
             if (iconSize == Vector2.zero) { return string.Empty; }
 
             if (iconSize.x == 1240 && iconSize.y == 1240 || iconSize.x == 2480 && iconSize.y == 1200) { return texturePath; }
 
-            string filePath = string.Format("{0}/{1}_{2}x{3}.png", _outputDirectoryName, type.ToString(), iconSize.x, iconSize.y);
+            string filePath = string.Format("{0}/{1}_AppIcon_{2}x{3}.png", _outputDirectoryName, Application.productName, iconSize.x, iconSize.y);
```
keveleigh: This change will create duplicates with different names if tile generation was run before this PR, right? That's probably fine, since this name change looks more descriptive for the long run.

StephenHodgson: It removes all the old files in the folder before regenerating, so it should be fine.

StephenHodgson: Also, I make them more generic bc I made it so it didn't duplicate existing texture sizes.  Now we generate only what we need, and no more duplicate sizes exist.  It's possible that a single texture is being used for many icons.

keveleigh: Aha I see that now! Nice

PR #1555, opened 2017-12-20
MixedRealityTeleport option to stay in same horizontal plane, fixed Editor

LocalJoost:
- Added an option "Stay on the floor" to actually stay in the same horizontal plane, regardless whether or not the user puts the teleport marker on something higher. 
- Added a cube to step on to the InteractiveMeshCursor to be able to test the above
- Fixed MixedRealityTeleportEditor because it did never serialize only three of the twelve properties it edits.

### DIFF
--- a/Assets/HoloToolkit-Examples/Input/Scenes/InteractiveMeshCursor.unity
+++ b/Assets/HoloToolkit-Examples/Input/Scenes/InteractiveMeshCursor.unity
@@ -88,6 +88,7 @@ LightmapSettings:
     m_PVRFilteringAtrousPositionSigmaDirect: 0.5
     m_PVRFilteringAtrousPositionSigmaIndirect: 2
     m_PVRFilteringAtrousPositionSigmaAO: 1
+    m_ShowResolutionOverlay: 1
   m_LightingDataAsset: {fileID: 0}
   m_UseShadowmask: 0
 --- !u!196 &4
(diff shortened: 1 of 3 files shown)


Assets/HoloToolkit-Examples/Input/Scenes/InteractiveMeshCursor.unity
```diff
@@ -362,6 +363,94 @@ MonoBehaviour:
...
+  m_PrefabParentObject: {fileID: 0}
+  m_PrefabInternal: {fileID: 0}
+  serializedVersion: 5
+  m_Component:
+  - component: {fileID: 1166889117}
+  - component: {fileID: 1166889116}
+  - component: {fileID: 1166889115}
+  - component: {fileID: 1166889114}
+  m_Layer: 0
+  m_Name: CubeToStandOn
```
keveleigh: Is this adding a new cube to test out the new setting for teleporting? I don't think `InteractiveMeshCursor` is the right scene for that. Maybe `MotionControllerTest` or a new `TeleportingTest` scene?

Assets/HoloToolkit/Input/Scripts/Utilities/Managers/MixedRealityTeleport.cs
```diff
@@ -287,11 +290,18 @@ public void DoStrafe(Vector3 strafeAmount)
         /// <param name="worldPosition"></param>
         public void SetWorldPosition(Vector3 worldPosition)
         {
+            var orginaly = transform.position.y;
```
keveleigh: typo: `orginal` -> `original`.
Also, wouldn't hurt to capitalize the `Y` as well, to make it a bit clearer what this variable is and harder to confuse with the word `originally` or something.

Assets/HoloToolkit-Examples/Input/Scenes/InteractiveMeshCursor.unity
LocalJoost: It is adding an extra cube indeed. Do you see that as a real problem? There is a cube in the scene now, it's just a bit too high to easily place the cursor on. If you want it gone, I will delete it again

keveleigh: The only problem I see is discoverability, if this new cube is intended to be the test for this new feature, since it's in a cursor test scene.

LocalJoost:
Well, it was more a show off feature, to show you it actually works

Joost van Schaik

Windows Development MVP

"The world as I envision it in my head is such a more interesting place"

PR #1265, opened 2017-10-30
Accessibility changes to make extending the Toolkit easier to maintain.

markgrossnickle:
These changes will allow people to maintain their extensions to the Toolkit without having to change accessibility each time they grab a new build.  

Whenever we update the toolkit we have to go back and make these accessibility changes so it would be nice to have them made into the core so that we and others like us do not need to continuously change the core code. 

The one class that is more than an accessibility change is GazeManager.cs which I reverted the Ray.Origin back to maintaining a reference. The functionality should be the same for existing users but if there is concern I understand ditching this change.  The benefit of having it saved is that you can override the GazeOrigin for custom gaze interactions.  We do so when allow users to draw with the manipulation gesture.

PR #1557, opened 2017-12-20
Accessibility changes to make extending the Toolkit easier to maintain.

markgrossnickle:
Overview
---
These are the changes requested from:
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1265#

I've updated the code to address the conflicts.  It was easier to just create a new fork and redo it than try to fix the prior. I'll close the prior ticket above.

The main goal of this ticket to keep the Toolkit simple and focused on the most common needs while allowing others to extend where they deem fit.  There are no breaking changes. Besides a null check the only changes involve switching functions or properties to protected instead of private (which I realize can lead to breaking changes down the road if we are not careful).

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
@@ -106,9 +106,9 @@ namespace HoloToolkit.Unity.InputModule
         /// <summary>
         /// Position, scale and rotational goals for cursor
         /// </summary>
-        private Vector3 targetPosition;
-        private Vector3 targetScale;
-        private Quaternion targetRotation;
+        protected Vector3 targetPosition;
+        protected Vector3 targetScale;
+        protected Quaternion targetRotation;
 
         /// <summary>
         /// Indicates if the cursor should be visible
@@ -124,7 +124,7 @@ namespace HoloToolkit.Unity.InputModule
 
         #region MonoBehaviour Functions
 
-        private void Awake()
+        protected virtual void Awake()
         {
             // Use the setter to update visibility of the cursor at startup based on user preferences
             IsVisible = isVisible;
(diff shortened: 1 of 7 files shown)


StephenHodgson:
@NeerajW this isn't based on the Dev branch, and for some reason it's not letting me update that.

NeerajW:
@StephenHodgson thank you for catching. I just went blind on that.

StephenHodgson:
@markgrossnickle Let's address this in MRTK vNEXT.  Please join us in the discussion about how to make the reboot more extensible.

markgrossnickle:
Sounds good.  Where is the reboot hosted or being discussed currently?

StephenHodgson:
Join us over on Slack or hop on the discussion over on #1754

HBJKZLH:
Untiy [Universal Windows Platform] Platform does not support.

The Socket code is written and the Thread code is written and the FileStream code is written, you know what?



I used the network communication section to find that this part of the code could not be transferred to the [Universal Windows Platform].

After I tested Unity 2017.2.0f3 (64-bit) version,Unity 2017.3.1f1 (64-bit) version,Unity 2017.2.1f1 (64-bit) version.



They don't support it.



Do you have any feedback on untiy?









When I wrote socket and IO and Thread related code in the program, untiy packaging times was wrong,
The socket.Connect function is not supported.
The socket.Receive function is not supported, and the 
socket.Send function is not supported.
The filestream.close function is not supported,
New Thread(recv) is not supported,
Does not support Thread. IsBackground,
Does not support Thread. To Start,




But if I just write in the code.

The Socket client = null;

Thread the Thread = null;

FileStream filedata = null;



This way, the Universal Windows Platform is passed.

I think it should be the flaw in the Platform library of untiy [Universal Windows Platform].


Please help me to feedback this question,








------------------ 原始邮件 ------------------
发件人: "Stephen Hodgson"<notifications@github.com>;
发送时间: 2018年3月2日(星期五) 凌晨1:40
收件人: "Microsoft/MixedRealityToolkit-Unity"<MixedRealityToolkit-Unity@noreply.github.com>;
抄送: "Subscribed"<subscribed@noreply.github.com>; 
主题: Re: [Microsoft/MixedRealityToolkit-Unity] Accessibility changes tomake extending the Toolkit easier to maintain. (#1557)




Join us over on Slack or hop on the discussion over on #1754

PR #1560, opened 2017-12-21
Fixed bug in UserNotifications example that didn't correctly check for local user.

kewlniss:
Overview
---
Simple fix to have the code behave as it was intended - only notify when a remote user joined instead of all users (including the local user).

### DIFF
--- a/Assets/HoloToolkit-Examples/Sharing/SharingService/Scripts/UserNotifications.cs
+++ b/Assets/HoloToolkit-Examples/Sharing/SharingService/Scripts/UserNotifications.cs
@@ -35,17 +35,17 @@ namespace HoloToolkit.Sharing.Tests
             usersTracker.UserLeft += NotifyUserLeft;
         }
 
-        private static void NotifyUserJoined(User user)
+        private void NotifyUserJoined(User user)
         {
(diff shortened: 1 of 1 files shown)


Assets/HoloToolkit-Examples/Sharing/SharingService/Scripts/UserNotifications.cs
```diff
@@ -35,17 +35,17 @@ private void Connected(object sender = null, EventArgs e = null)
...
+        private void NotifyUserJoined(User user)
         {
-            if (user.IsValid() && localUser != user)
+            if (user.IsValid() && localUser.GetID() != user.GetID())
             {
                 Debug.LogFormat("[User Notifications] User {0} has joined the room.", user.GetName());
             }
         }
 
-        private static void NotifyUserLeft(User user)
```
StephenHodgson: Why remove static?

kewlniss: Hmm, why should static be there? The actual UserJoined and UserLeft events on SessionUsersTracker is not a static class. This makes it more inlined with the well-behaved RemoteHeadManager example class.

Issue #1573, opened 2017-12-29
UI components not working with multiple scenes


christian4423:
## Overview
I had all of my components nested under game objects. This was getting messy for me and I decided to switch to scenes and use UnityManager.SceneManager

Everything is working as expected, except my Input UI Elements are now are not recognized by my gaze cursor. Following the setup for the keyboard input in the Holokit examples, I was able to get it set up before. 

But now I have this issue. 

Here are some screenshots below

![pt1](https://user-images.githubusercontent.com/6324867/34428171-01413900-ec18-11e7-9a80-5f8daa6d98d9.PNG)

If you look at the canvas in the inspector you will see the Camera I need to reference is not allowed. 

![pt2](https://user-images.githubusercontent.com/6324867/34428175-09cb836e-ec18-11e7-8d34-888cf0867849.PNG)

If you look at Inspector when Input Manager is selected, It says UI Raycast Camera is the camera under EventSystem

Let me know if any more information is needed for this or if there is a better way to handle scene management in unity that will allow use of the camera. 

## Unity Editor Version
2017.2
## Mixed Reality Toolkit Release Version
v2017.2.1.0

genereddick:
There are many issues related to multi-scene projects, but see #1420 on this topic. 

You can either create a script and run the code below on loading the new scene. This will update an existing Canvas to the correct camera:
    
    FocusManager.Instance.UpdateCanvasEventSystems();

Or, try creating a new Canvas (and then modifying it to be the same as the old version). For me, I wasn't able to update an existing Canvas with the correct camera, but a new Canvas was OK.

StephenHodgson:
I really don't think using `FocusManager.Instance.UpdateCanvasEventSystems();` is the best solution and was never really intended to be used at run-time.

You should probably set the UIRaycasetCamera directly on your canvas.

```
// Set the Canvas reference in the inspector
private void Awake()
{
    if(myCanvas.isRootCanvas && sceneCanvases[i].renderMode == RenderMode.WorldSpace)
    {
        myCanvas..worldCamera = FocusManager.UIRaycastCamera;
    }
}
```

christian4423:
Sorry, I will review these comments ASAP and respond with any success. I did not get alerted that people responded to this!

Issue #1534, opened 2017-12-15
RaycastCamera cannot manually be assigned to canvas camera field


Alexees:
## Overview
Dragging the RaycastCamera into the Camera field in WorldSpace mode of a canvas does not work, it doesn't get assigned. This scenario is only possible when adding a prefabbed canvas with WorldSpace configured already, or when adding MixedReality stuff after a canvas has already been setup that way. All other scenarios prompt for the RaycastCamera to be assigned.
Workaround: Select WorldSpace Render Mode again

## Expected Behavior
Assigning a camera should be possible at any time.Or Canvases should be checked for the the right camera being assigned on adding the MR content to a scene or when adding a new prefab to the scene that contains one.

## Actual Behavior
Assignment doesn't work

## Steps to reproduce
Create a canvas and configure it for WorldSpace rendering. Add MR stuff. Try to assign the RaycastCamera to the currently empty camera field.

## Unity Editor Version
MRTP5

## Mixed Reality Toolkit Release Version
current

StephenHodgson:
This was kinda by design.  It was decided that opting out by default was better than overriding values in people's projects.

StephenHodgson:
Although I do have a issue open to make this process easier to do when you add the Input Manager prefab into the scene.

Issue #1420, opened 2017-11-22
Easy Legacy uGUI Canvas Raycast Camera updater


StephenHodgson:
## Overview
Provide an easy way to update the default canvas raycasting camera for projects using uGUI.  New users are not affected, because we're already prompting them when they create new canvases.

Currently the only easy way to do this is by using the scene configuration menu and re-adding the InputManager.

More advanced users can also manually call the update using `FocusManager.Instance.UpdateCanvasEventSystems();`

## Expected Behavior
When devs open legacy projects, we should prompt them once if they wish to update their canvases default raycasting camera if the canvas' render mode is set to world space.

## Actual Behavior
Currently older projects are automatically opted out to using the new UIRaycastCamera as the default raycaster, and devs need to manually assign the camera to each canvas who's render mode is set to world space.

## Unity Editor Version
Any/All

## Mixed Reality Toolkit Release Version
405bb16be62b5d1eb18f8e5f39d9d1fe9a962013

genereddick:
Is the MixedRealityCamera the UIRaycastCamera?

StephenHodgson:
No, See #1442

genereddick:
One thing to note is that for multi-scene projects, there won't necessarily be a camera after the iniitial scene, so you will need to guide them to use the UpdateCanvasEventSystems call: 
    
    FocusManager.Instance.UpdateCanvasEventSystems()

jasonhbartlett:
I think the Keyboard prefab in the MRTK might need an update due to this.  If you just drag and drop it into the scene it won't automatically work (ie - won't be detected by gaze/raycast).  To fix it you just have to click the Canvas component Render Mode drop-down and re-select World Space (cause it was already selected).  This triggers your little warning about "In order for the Input Manager to work properly with uGUI raycasting..."  After that it works.  
Just was scratching my head for a few minutes trying to figure out what the deal was.  Hope to save someone having to remember this in the future.

kmikkel:
Suggesting to add a checkbox to the FocusManager to call UpdateCanvasEventSystems on scene loading.

StephenHodgson:
>Suggesting to add a checkbox to the FocusManager to call UpdateCanvasEventSystems on scene loading.

No, That's not a good idea.  It should be serializing the scene before you ever have to load a scene.

Strepto:
[...]

PR #1575, opened 2018-01-02
Update for UIRaycast Camera and Canvas Referencing

StephenHodgson:
Overview
---
This PR aims to fix the issues with using the uGUI Canvases and the MRTK's input system.

Changes
---
- Fixes: #1573 Fixes issues with using canvases across scenes, and assigning the UIRaycastCamera when a new scene is loaded.
- Fixes: #1420 Gives devs the ability to update the scenes canvases via the Scene Capabilities Wizard.
- Fixes: #1534 Fixes issues with setting the UIRaycastCamera on the Canvas Component itself.


Caltato:
Hi, I was going to start a new issue until I saw this. I'm unable to assign any camera to a Canvas' camera field when its render mode is set to Screen Space and I was wondering if this PR will solve this.

Currently running on the latest release for Unity 2017.2.1f1.

StephenHodgson:
Form my tests, yes

Caltato:
Alright I shall test it out. Thanks!

Caltato:
Hey @StephenHodgson , sorry but it still seems like I can't assign cameras to canvases. Even if I add a new canvas and camera to the scene, I can't drag it into the field. Is there something else that could be causing this?

StephenHodgson:
Idk, I can't reproduce this

Caltato:
Woah okay it suddenly worked O.o I have no idea how. I literally just kept trying the same thing over and over again. Thanks anyway! Sorry for the bother

StephenHodgson:
Can I get some love on this PR?

DDReaper:
OK, I'm struggling to see the use case for this "Canvas helper".
We create a script to keep a reference to the canvas, that is attached to the canvas?
We then also automatically add / remove the Canvas helper from the canvas automatically.  

Seems odd as to why not just reference the canvas itself.

Also, there is almost no use of the Canvas Helper in these scripts unless I'm missing something.

StephenHodgson:
Canvas helper ensures it's reference is serialized in the scene or prefab so we don't have to get the component at runtime and so our references don't break though scene transitions

StephenHodgson:
A few Test Cases:

1. Create a blank scene with nothing but a canvas on it.
2. Create a second scene with only the basic MRTK prefabs (Input System, Camera, and cursor)
3. Start in the second scene with the MRTK content and load into the first scene, and the canvas will not work.

This PR essentially fixes that.

DDReaper:
I tried those replication steps on your updated code, yet the canvas still had no camera assigned when run?
Did I miss something?

StephenHodgson:
Was the canvas helper attached, and was the canvas set to world space?

DDReaper:
Canvas was a world space canvas, but no, the canvas helper wasn't attached. By your replication instructions, I would have thought that was automatic

StephenHodgson:
Nope, it was the bug I was trying to point out.

StephenHodgson:
I agree a more elegant fix would need to be solved in the core editor logic.

The biggest problem here is the way Unity implemented the Graphic UI features, and why they have to do two different kinds of raycasting for different kinds of graphic objects.

paulalom:
Hi,

I had the same problem as Caltatao where I couldn't attach any camera to any canvas (on either world space or screen space settings). I flipped all of the switches and fiddled with all of the dials on all of the related game objects to no effect. I am on fall creators update, I reinstalled the MRToolkit today from master and am using unity version 2017.2.1f1. 

I created a new project, deleted the camera and directional light, dragged the following files/folders into my assets:
MixedRealityToolkit-Unity-master\Assets\HoloToolKit
MixedRealityToolkit-Unity-master\Assets\HoloToolKit.meta
MixedRealityToolkit-Unity-master\Assets\mcs.rsp
MixedRealityToolkit-Unity-master\Assets\mcs.rsp.meta

Then I applied MRTookit scene and project settings through the MRToolkit menu (this is where the issue is, as I can apply cameras to canvases before I do this), added a canvas, then attempted to add the MixedRealityCamera to the canvas camera property by dragging and dropping. This resulted in an empty camera field (ie nothing changed).

I figured out that if I reload the scene at this point, **I can now add cameras to the canvas**. 
Restarting unity removes the camera from the canvas and I need to re-load the scene before I can add the camera back again.

Assets/HoloToolkit/Utilities/Scripts/Editor/SceneSettingsWindow.cs
```diff
@@ -95,7 +97,23 @@ protected override void ApplySettings()
                 }
 
                 PrefabUtility.InstantiatePrefab(AssetDatabase.LoadAssetAtPath<GameObject>(AssetDatabase.GUIDToAssetPath(InputSystemPrefabGUID)));
-                FocusManager.Instance.UpdateCanvasEventSystems();
+                Values[SceneSetting.UpdateCanvases] = true;
```
keveleigh: Is this required for adding the input system prefab, or should we give the choice of one or the other instead of forcing it even if I don't check the Update Canvases box?

StephenHodgson: This is required when using the input system.

StephenHodgson: New setups are auto opt in, while existing setups are opt out.

Issue #1577, opened 2018-01-03
InteractiveElementInstantiate-Example throwing nullpointer


brean:
I just started the new year with a clean checkout of the MRTK.

I don't know how that happened (I guess I just forgot to save or something got tangled up during the merge) but the InteractiveGroupSet-Script attached to the "NoteButtonSet"-GameObject in my example ([InteractiveElementInstance on Dev_Working_Branch](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/Dev_Working_Branch/Assets/HoloToolkit-Examples/UX/Scenes/InteractiveElementInstantiate.unity)) does not have the GameObject for the Grid set in the Inspector anymore so the Script throws a NullPointer if you just press play.

Should I just attach it again and create a pull request for the .unity-file or should I write something that searches the scene and attach it programmatically?

In general what is best practice for that? Should scripts search for components in the scene if they are not set in the inspector?

StephenHodgson:
Best practice is to set the reference in the inspector, and fallback if the field/property is null on Start/Awake

PR #1578, opened 2018-01-04
Set Grid in inspector, get Grid using code if it is not defined

brean:
Overview
---
- Set the Grid in inspector/unity file
- Added code to find the grid by name if it is not set (as described in #1577)
- Removed some unused Scaler for Canvas in inspector/unity file

Changes
---
- Fixes: #1577 .


Assets/HoloToolkit-Examples/UX/Scripts/InteractiveGroup.cs
```diff
@@ -36,7 +40,7 @@ private void Start()
                 UpdateData();
             }
         }
-
+
```
StephenHodgson: We can probably revert this line.

Issue #1584, opened 2018-01-04
Build deployment fails if only one Windows SDK installed on 2017.2.1.0


johnnycrich:
Hi there, I am _brand_ new to MR development and wanted to file what I believe to be a significant issue, with the caveat that I could be entirely missing a key element of the intended behavior (though it seems unlikely).

## Overview
If only one Windows SDK version is installed locally, in my case 10.0.16299.0, the build deployment process fails because BuildDeployWindow expects to find more than one installed version upon open.

## Expected Behavior
I would expect the solution deployment to UWP to only ever need one Win SDK install. As such, the catch for finding no installs ought to be the below, since we'd want to look for an empty array:
`...
if (currentSDKVersionIndex < 0)
...`

This change at BuildDeployWindow/326 is my workaround.

## Actual Behavior
The build process will throw an exception if only one Win SDK is found in **...Windows Kits\10\Lib**, instead of finding none.

## Steps to reproduce
1. Follow steps as per [Getting Started](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/GettingStarted.md).
2. Attempt to Build SLN and APPX after importing the MixedRealityToolkit package and adding any scene to build queue.

## Unity Editor Version
2017.2.1f1

## Mixed Reality Toolkit Release Version
2017.2.1.0

StephenHodgson:
Thanks and welcome to the community!

This is a great opportunity for you to submit your first pull request outlining these changes.

johnnycrich:
Submitted! 👍

Alexees:
Nice one, had that yesterday.

Dmidcap:
I had the same issue (running Unity 2017.2.0f3) and Mixed Reality Toolkit 2017.2.1.0. Your fix removed the error message I was getting, but I still can't Build the SLN from the build window.

Error message prior to fix line 326:
![unable-to-find-sdk](https://user-images.githubusercontent.com/3589780/35125241-8471f542-fc6e-11e7-96ec-7e25cf611c30.png)

After fixing line 326, error message is gone but still unable to click build.
![unable-to-build](https://user-images.githubusercontent.com/3589780/35125240-845b2f42-fc6e-11e7-9419-f23bd001fe82.png)

PR #1585, opened 2018-01-04
Fix for build error when one Win SDK installed

johnnycrich:
Overview
---
Currently, if only one Windows SDK version is installed locally, the build deployment process fails because BuildDeployWindow expects to find more than one installed version upon open. With this fix, the process will throw an exception if only no SDKs are found.

Changes
---
- Fixes: #1584 .


keveleigh:
Should [line 309](https://github.com/johnnycrich/MixedRealityToolkit-Unity/blob/9e6963438b0e69a6c8ce83a882a73e7b5ee0a8a5/Assets/HoloToolkit/BuildAndDeploy/Editor/BuildDeployWindow.cs#L309) be updated to start at `-1`? Otherwise, if I'm reading this correctly, `currentSDKVersionIndex` will stay 0 with no SDKs installed, and this check will no longer pass.

johnnycrich:
@keveleigh Thanks for the catch! Updated.

keveleigh:
Thanks for the fix!

PR #1588, opened 2018-01-05
Added ability to update the keywords at runtime.

StephenHodgson:
Overview
---
Added the ability to update the keywords at run time.

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Editor/SpeechInputSourceEditor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Editor/SpeechInputSourceEditor.cs
@@ -19,7 +19,7 @@ namespace HoloToolkit.Unity.InputModule
             persistentKeywordsProperty = serializedObject.FindProperty("PersistentKeywords");
             recognizerStart = serializedObject.FindProperty("RecognizerStart");
             confidenceLevel = serializedObject.FindProperty("recognitionConfidenceLevel");
-            keywordsAndKeys = serializedObject.FindProperty("Keywords");
+            keywordsAndKeys = serializedObject.FindProperty("keywords");
         }
 
         public override void OnInspectorGUI()
--- a/Assets/HoloToolkit/Input/Scripts/InputSources/SpeechInputSource.cs
+++ b/Assets/HoloToolkit/Input/Scripts/InputSources/SpeechInputSource.cs
@@ -20,7 +20,7 @@ namespace HoloToolkit.Unity.InputModule
     /// Edit -> Project Settings -> Player -> Settings for Windows Store -> Publishing Settings -> Capabilities
     /// or in your Visual Studio Package.appxmanifest capabilities.
     /// </summary>
-    public partial class SpeechInputSource : BaseInputSource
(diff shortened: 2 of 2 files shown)


keveleigh:
What's the benefit here vs adding a second (or third, etc) `SpeechInputSource` and setting the new keywords in it? As I understand it, `KeywordRecognizers` are optimized for spinning up new recognizers and turning off unused ones, instead of adding/removing keywords.

StephenHodgson:
There's no way to add a new component and set its keyword array before the `KeywordRecognizers` gets initialized.

keveleigh:
I don't believe that's true.

I created a test app and set a string on a component that I added at runtime:
image

image


Here were my results when running:
image

StephenHodgson:
Okay, but not every machine does this at the same speed.

Remember that kinda stuff is non deterministic.  It's all based on the speed of your machine!

keveleigh:
@StephenHodgson Is it? Unity is single-threaded, right? So it makes sense to me that the remainder of the `Update` finishes before the added component's `Start` is ever called. I'm not sure where the non-deterministic part is, since this isn't script execution order or anything. It's a known component added at runtime.

I'm looking for official documentation about this either way.

StephenHodgson:
What if we don't call it from an update loop, but instead from a raised event or Action?

You can't tell me without certainty you'll set the string before that game object is initialized and Start called on it.

keveleigh:
I did it from an event, and I'm seeing identical behavior:

image

image

StephenHodgson:
>As I understand it, KeywordRecognizers are optimized for spinning up new recognizers and turning off unused ones, instead of adding/removing keywords.

Also, we're not adding or removing any keywords.  We are doing exactly what you said. We're turning off the `KeywordRecognizer`, disposing it, and then creating a new one with a new set of keywords.

keveleigh:
> We're turning off the KeywordRecognizer, disposing it

is more than adding a new one to the existing one.

StephenHodgson:
What if you want to completely change the keywords all together?

This supports that

keveleigh:
> What if you want to completely change the keywords all together?

`Stop()` the existing recognizer, then create a second one. Then, if you want to completely change back, the system is optimized for that, since you're turning on/off existing recognizers.
The underlying [SpeechRecognizer constraints](https://docs.microsoft.com/en-us/uwp/api/windows.media.speechrecognition.speechrecognizer#Windows_Media_SpeechRecognition_SpeechRecognizer_Constraints) do better when disabling them versus removing them, to save a [`CompileConstraintsAsync` call](https://docs.microsoft.com/en-us/uwp/api/Windows.Media.SpeechRecognition.SpeechRecognizer#Windows_Media_SpeechRecognition_SpeechRecognizer_CompileConstraintsAsync). That's the main reason you can't edit a `KeywordRecognizer`'s keywords after it's created.

This scenario is already supported, but if we want to add something to make it more obvious, I want to make sure it's done the right way.

Additionally, while I haven't yet found an official source either way, these posts describe the `Start` behavior I'm seeing:
https://gamedev.stackexchange.com/a/135905
https://gamedev.stackexchange.com/a/115597

PR #1590, opened 2018-01-05
Removed invalid attributes on properties.

StephenHodgson:
`[HideInInspector]` and `[SerializeField]` are invalid attributes on properties.


NeerajW:
I suspect there are many other places in the code base that do this though :)

StephenHodgson:
Nope, I already checked

Issue #1593, opened 2018-01-05
Duplicate BoundingBoxes.fbx file


StephenHodgson:
## Overview
There's two `BoundingBoxes.fbx` assets. One in UX/Models/BoudindingBoxes and one in UX/Resources/Models/

The latter should be removed to reflect the ongoing folder conventions.

StephenHodgson:
@Railboy

PR #1595, opened 2018-01-05
Removed duplicate bounding box asset 

StephenHodgson:
Changes
---
- Fixes: #1593


StephenHodgson:
@Railboy

Railboy:
This change breaks the Bounding Box prefab. I recommend that we rename the UX mesh asset instead, since it will likely be modified further as we incorporate more bounding box functionality.

StephenHodgson:
How does it break the prefab?  I double checked and all the mesh references are linked to the asset in the UX/Models folder

StephenHodgson:
Also, this duplicated asset is causing issues with duplicated meta files.

Railboy:
We've got several unmerged bounding box prefabs that aren't using the UX meshes yet. But on reflection I guess we have to break them sooner or later.

NeerajW:
@Railboy could we help approve or comment on this?

Issue #1576, opened 2018-01-02
Removing PressType from Generic Input Events


StephenHodgson:
## Overview
The PressType is really only ever used for the Interaction Input Sources, and doesn't really have a place for generic input types.

I propose we move some of the Interaction Inputs into their own classes and interfaces that inherit from the Base Events.  This will help ensure we can have better cross platform inputs, and will reduce the complexity of the Input System.

Railboy:
I like this idea in theory. It couldclear up a lot of the messiness we're encountering with multi-pointers.

Issue #334, opened 2016-11-18
Research a better way to get CursorModifier in Cursor.cs


StephenHodgson:
I love the new input module, but there's gotta be a better way we can get the cursor modifier than GetComponent whenever we get a new focused object.

Probably not an easy way to around this, but it would be nice.

Cursor.cs line [253](https://github.com/Microsoft/HoloToolkit-Unity/blob/ExtensibleInputModule/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs#L253):
```
protected virtual void OnFocusedObjectChanged(GameObject previousObject, GameObject newObject)
{
    TargetedObject = newObject;
    if (newObject != null)
    {
        OnActiveModifier(newObject.GetComponent<CursorModifier>());
    }
}
```

paseb:
The only thing here is that OnActiveModifier is also for setting the modifier to null.  So as-is it's not ideal but still handles the base side of having OnActiveModifier handle both new modifier or no modifier.  Ideally I think we would the modifier to self-register on focus and the ability to override.  Like having a modifier that does not require focus but is enabled through other mechanisms.

Danta1st:
I do not really see a way around this without sacrificing the flexibility provided. 

While the GetComponent is sort of slow, it should not _really_ be a problem unless you are performing hundreds or thousands during the same update cycle.

I would much rather we focused on cleaning the interface by either: 
1. Removing the dead methods: _GetModifiedPosition_, _GetModifiedRotation_, _GetModifiedScale_
2. or, Removing the all-in-on method: _GetModifiedTransform_
where I would prefer the individuality of split methods. That is go with option 2, and refactor accordingly.

StephenHodgson:
@Danta1st great suggestions. I'd say they warrant their own issue.

StephenHodgson:
Fixed in vNEXT

Issue #1602, opened 2018-01-05
Controller Pointers Base Branch


Railboy:
## Overview

This branch adds a basic line pointer example to the MRTK:
https://github.com/Railboy/MixedRealityToolkit-Unity/tree/Controller_Pointers_Base

## Issues

This is a subset of the pointer examples we're working on in #1486. Some changes were necessary to make these classes function in the current dev branch. Those changes have been marked with a comment: 
`// This will be enabled when multi-pointer support is incorporated`
If we open a PR we'll want to strip these out - they're just to facilitate discussion.

Minor issues that are addressed in #1486:
 - Cursor lags behind pointer's target point
 - Single cursor is swapped between multiple pointers
 - IPointerInput.OwnsInput can result in false positives

## Files

- Adds ControllerPointerBase.cs, FocusPointerInput.cs, InteractibleHighlight.cs and LinePointer.cs classes to HoloToolkit/UX/Scripts/Pointers
- Adds ControllerPointerExamples scene to HoloToolkit-Examples/UX/Scenes.
- Adds LinePointerBasic prefab to HoloToolkit-Examples/UX/Prefabs
- Minor changes to several Line classes.

## Questions

Do we feel this is worth merging? How could it be improved?

StephenHodgson:
@Railboy I started working on another branch to show my perspective and a different take on the purposed changes.  I think we should have a comprehensive meeting about and compare notes.

Railboy:
@StephenHodgson Taking a look now. At first blush I don't think your proposed changes conflict with this example in a significant way.

wiwei:
Closing issues older than 180 days. If this is still an issue, please reactivate with recent information.

Issue #1342, opened 2017-11-11
Update Mouse Support to reflect the Shell (Cliffhouse) inputs


genereddick:
## Overview
Using latest Unity & MRTP editions, even when I do not have "Target Occluded Devices" set in the Configure menu, attempting to Run from the Unity Editor opens the app in a connected Windows Headset or simulator, adds a floor plane, doesn't work with the mouse and other issues.

Maybe I'm missing a configuration setting or the expected behavior has changed?

## Expected Behavior
Unless targeting occluded devices, run should not create a floor plane, should preserve the correct placement of the camera (at head level, not floor level), should use the correct camera background, should continue to work with the mouse.

## Actual Behavior
Opens in connected Headset. Sets camera initial position incorrectly (positioned for Windows, not HoloLens). Inserts floor plane into scene. Can't use mouse.

## Unity Editor Version
Unity 2017.2.0p1-MRTP4, but happens in others

## Mixed Reality Toolkit Release Version
HoloTookit-Unity-v1.2017.2.0
also current Master

StephenHodgson:
I don't believe we have proper mouse support that's like in the cliff house yet, but I believe a lot of the preliminary stuff is getting imported with the MRDL merges.

genereddick:
When you click run from within the Unity Editor, how does it determine whether to run in Transparent vs Opaque mode? I had expected, perhaps incorrectly, that setting the checkbox in the Configure Menu for "Target Occluded Devices" would control this -- but it is not clear which settings this effects?

Specifically, on my system I have an HMD. The "Target Occluded Devices" is unchecked and my intention is the deploy to HoloLens, but I want to test within the editor on my PC without deploying. However, when I run within the Unity editor, sometimes it launches the MR Portal, sometimes not, but in every case I get the opaque configuration with the camera and floor misplaced, etc.

StephenHodgson:
I agree, having better support overall for this of switching would be ideal.

Although in my opinion, I think being able to test specific features out with the HMD would effectively be easier to do instead of pushing to the HoloLens each time.  I have a feeling that a lot of apps will be cross platform compatible in this way.

As far as when the mixed reality portal launches, the only deciding factor there is if you have XR enabled or not.

StephenHodgson:
The option for targeting occluded devices is somewhat depricated. 

[...]

PR #1603, opened 2018-01-06
Input System Refactoring

StephenHodgson:
Overview
---
This is purely an exploratory request to refactor the input system. See #1602 

The biggest changes to note:
- Each of the event data classes now have overloaded initialization methods for more refined source data
    - `InputEventData` now contains `PressType` and `Handedness`
- Added concepts for OnPointerUp/Down/Clicked
    - `IInputClickHandler` was renamed to `IPointerHandler`
- InputUp/Down/Pressed is separated out into its own thing, and doesn't raise pointer clicks to uGUI
- Renamed `IFocusable` to `IFocusHandler`
- `IPointingSource` now inherits from `IInputSource`
    - Pointers now can have cursors assigned to them
    - Pointers no longer require a Cursor, but a Cursor does require a Pointer.
- `CursorModifier` now implements `IFocusHandler` and assigns itself when focused to the pointer
- Combined and/or removed a few of the Input Handler Interfaces to make things cleaner
- Simplified raising sources by removing the `sourceId` parameter since it's already embedded in the input source data and added a `Name` property
- `InteractionInputSources` is now a singleton
- The `InputManager` assigns the source id as it's raised to ensure unique ids across all sources.
- Input Sources now inherit from `IEquatable` and are now considered to be unique based on source id.
- Renamed `CustomInputSources` to `SimulatedInputSources`

TODO
---
- [ ] Check that #1662 is fixed.
- [x] Support #1602 Multi Pointer functionality.
- [x] Implement Pointer interface in `GazeManager`, `InteractionInputSources`, and `PointerSources`
- [x] Remove `XboxControllerInputSource` and replace it with a class that is more similar to `InteractionInputSources` that tracks and keeps input sources mapped in the editor up to date
- [ ] Fix: #1621 & #1342 Native support for Mouse with similar interactions as cliff house (spatial mouse)
- [ ] Native support for keyboard
- [ ] #1653 Add grip position and rotation to the supported input info.

Changes
---
- Fixes: #1576 
- Fixes: #334


StephenHodgson:
Currently broken.  We need to register the pointers correctly.
Ended up removing the `SimpleSinglePointerSelector` for Muli Pointer feature.

Railboy:
Overview of features to pull in from #1486

- Rename IFocusable to IFocusTarget and implement FocusTarget base class (see **IFocusable -> IFocusTarget** post)
- Add IFocuser, refactor IPointingSource and PointerResult (See **IFocuser, IPointingSource & PointerResult** post)
- Add EventOrigin to input events (see **Overview** post)

StephenHodgson:
@Railboy, which scenes do I need to focus on for testing?

Railboy:
InputManagerTest is good for making sure we haven't broken pointers in general.

I have a scene called ControllerPointerExamples that we can use to test multi-pointers specifically. That relies on a controller pointer class that I plan on incorporating after you're finished. But you could easily write a simple pointer class for testing.

StephenHodgson:
I noticed there's a difference in the Line Scripts in the Working Dev branch and the MulitPointers (Specifically with parabolas) which should I deffer to?

Railboy:
MultiPointers has the latest Line classes.

StephenHodgson:
- Added Previous and Current Pointer Targets to the PointerResult struct.
- Added IFocusChangedHandler, and made the IFocusHandler inherit from it.   This makes it easier for scripts to know when global focus changed events happen. (Like for the Cursor, who doesn't need to implement everything from the IFoucsHandler)
- Added a Focus Target to the IPointingSource interface.
- Added FocusTarget base class for GameObjects that will handle being focused.

StephenHodgson:
@Railboy I'm still not fully convinced we need the `EventOrigin` string you want to include with that change.

I looked through that other PR to see exactly where it's being used, but I can't seem to find a good example.  Could you point me in the right direction?

Railboy:
@StephenHodgson That's okay, at this point it makes more sense to just trust that our team understands our own needs and that we're capable of determining that the current system doesn't address them. Once the PR is merged we'll create a scene that demonstrates EventOrigin in use and I promise it'll click for you right away.

StephenHodgson:
Okay.  Well I think this is good for now.  I'll expose the `EventOrigin` when we add the specific functionality for it with the included demo scene.  I just wanted to understand exactly what your needs were to better accommodate an appropriate solution.

StephenHodgson:
From feedback at the Weekly Shiproom Meeting:

- [ ] Break `PointerInputSource` from `InputSource`, and make them their own entities.
- [ ] Implement `MultiPointers` into the `InputSource` data and have the `InputSource` keep track of it's internal `pointerId`'s.
- [ ] Give ability to name the pointer in constructor, and make the `PointerName` property have get and set capabilities.

yacuzo:
Can you please extract an interface from the various EventData objects? I should not need an EventSystem (needed to construct an InputEventData object) to do a unit test for, say, a PointerClickHandler. With an interface i can mock the actual class in tests.

Could you also extract an interface for the InputManager itself? Same problems with testability and tight coupling appears here too. Any class I make which does calls to the InputManager is not fesable to test due to the singleton reference.

Ideally, for looser coupling and to make this friendly for dependency injection (Zenject for example), all of the InputManager's logic should be in a plain C# class, together with a MonoBehaviour(Singleton) which only links the C#-class (new it up in Awake) to the unity lifecycle. Those who are fine with singletons can still use the singleton, while those who have other architectures can use the plain C#-class directly. The same concept could be applied to all managers/singletons. If the managers need some other C#-class or some singleton, then the dependency should be injected in the managers constructor by the MonoBehaviour singleton.  

I can elaborate or clarify if this is hard to read. I really would like this entire package to be friendly to unit tests and dependency injection. As it stands I have to extract interfaces and do other dirty changes to use MRTK, and that makes it hard/tedious to stay up to date with changes.

StephenHodgson:
I'm totally on board with DI, but seems difficult to do in Unity because of the object life-cycles and such 
 (you can't guarantee an object will even be in the scene).  Usually I do the same in a lot of instances where if an object doesn't need to be a Monobehaviour its not.

For the most part I've been trying to pattern a lot of the changes off of Unity's own Input Modules and how they're handling different things here and there.

As for the EventSystem requirement, I don't think what you're asking for is possible.  The EventData classes all inherit from Unity's base event data type, and those require the EventSystem to function properly.  We do have unit tests for the input system, and I've updated those as well going through these changes.

Overall though, I do want to try and get away from using the Standalone Input Module, but that might take some time, as we need to make sure that the same functionality can be guaranteed in our input system.

I really do like your suggestion about DI though, and I'll take a pass at seeing what can be done to do this.  That might need it's own PR just for Singletons in general.

A third option is to just get rid of the Input System in the toolkit altogether.  A lot of people have issues with it and MSFT also just wants to get back to basics with simple scripts that just handle everything in one (although, I'd continue to use and modify it in a separate project).

yacuzo:
Thanks for the detailed answer. 

Yea, definitly a separate PR for the singletons. I just figured I'd throw in the suggestion here since it was related to what i was describing.

maxouellet:
About DI, I wrote some kind of DependencyResolver a while back that I've been using in my projects instead of singletons. It's not a dependency injector with a pull model though: it's essentially a static class that wraps a Dictionary<Type, object>. It verifies that anything registered to it is an interface, and allows consumers to retrieve an interface from it. Doing that, you can change all the singletons so that they implement some interface and register themselves with the DependencyResolver on awake, and they anything that depends on the singletons just queries the interface they want from it.

I never gave a serious thought about submitting it to HoloToolkit because it makes things a little harder to understand, which would impact things like the HoloAcademy for example (for the same reasons that @StephenHodgson mentioned around the input system being complex). I do think it's better than using singletons for professional codebases though, for the reasons that @yacuzo pointed out.

StephenHodgson:
I think the DI vs Singleton conversation should move into an issue where we can get a larger discussion going.  I def like where this is going though.

maxouellet:
@StephenHodgson I honestly don't know when/if I'll have the time to look at this. This is a huge changeset. I wish it was divided into smaller, easier to consume changes, it would make it easier to review.

StephenHodgson:
@Railboy, the MultiPointers branch is the extension of this PR that specifically addresses the concerns you had and the work needed for the Teleporting stuff.

If we can get some time to sit down and talk 1:1 I'd like to go over the stuff with you.

StephenHodgson:
@maxouellet yeah, I wish smaller chunks would be better here, but the rippling changes required for it all to work seems unavoidable.

StephenHodgson:
Going to close this and start working on smaller PRs

PR #1608, opened 2018-01-08
Layer extension optimization

StephenHodgson:
Overview
---
There's lots of unnecessary memory allocations when sorting layers.
Changed the `IEnumerable` parameters to a basic array.
Updated Focus Manager assertions to remove memory allocations when getting the type even if no assertion is thrown.
Removed the list of `ComparableRaycastResult` in `EventSystemExtensions`, because we don't need it.


Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs
```diff
@@ -552,8 +552,8 @@ private void RaycastPhysics(PointerData pointer, LayerMask[] prioritizedLayerMas
             RayStep rayStep = default(RayStep);
             RaycastHit physicsHit = default(RaycastHit);
 
-            Debug.Assert(pointer.PointingSource.Rays != null, "No valid rays for " + pointer.GetType());
```
keveleigh: Doesn't this remove context from the assertion statement? Why would we want to remove that?

Assets/HoloToolkit/Common/Scripts/Extensions/LayerExtensions.cs
```diff
@@ -54,15 +53,16 @@ public static int Activation
         /// <param name="layer">Layer to search for</param>
         /// <param name="layerMasks">List of LayerMasks to search</param>
         /// <returns>LayerMaskList index, or -1 for not found</returns>
-        public static int FindLayerListIndex(this int layer, IEnumerable<LayerMask> layerMasks)
+        public static int FindLayerListIndex(this int layer, LayerMask[] layerMasks)
```
keveleigh: Changing the parameters of a method would be a breaking change, right? Is this change worth that?

StephenHodgson: No, it's not a breaking change.

Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs
StephenHodgson: `GetType()` is causing an enormous amount of allocations during this time.

Assets/HoloToolkit/Common/Scripts/Extensions/LayerExtensions.cs
StephenHodgson: Yes, this change is very worth it, considering it removes _all_ allocations for this check.

StephenHodgson:
@NeerajW  It does:
image

Sorry to have dismissed your review.

StephenHodgson:
@Zod- any chance I could get your eyes on this as well?

Assets/HoloToolkit/Common/Scripts/Extensions/EventSystemExtensions.cs
```diff
@@ -25,17 +24,26 @@ public static RaycastResult Raycast(this EventSystem eventSystem, PointerEventDa
...
-        private static RaycastResult PrioritizeRaycastResult(LayerMask[] layerMaskPrio)
+        private static RaycastResult PrioritizeRaycastResult(LayerMask[] priority)
         {
-            ComparableRaycastResults.Clear();
-            foreach (var raycastResult in RaycastResults)
+            ComparableRaycastResult maxResult = default(ComparableRaycastResult);
+
+            for (var i = 0; i < RaycastResults.Count; i++)
             {
-                if (raycastResult.gameObject == null) { continue; }
```
Zod-: I'd keep `var raycastResult = RaycastResults[i];`

StephenHodgson: Good catch

StephenHodgson:
@Railboy had to update the test scene on @Zod-'s request.  Should be good now.

Zod-:
Do we need the clear here? Atleast when running the tests it doesn't seem to care if we do it or not.
```cs
RaycastResults.Clear();
eventSystem.RaycastAll(pointerEventData, RaycastResults);
return PrioritizeRaycastResult(layerMasks);
```

Zod-:
Ok, I've checked the source of `EventSystem.RaycastAll`. The first thing it does is clear the results itself so we don't have to do that.

https://bitbucket.org/Unity-Technologies/ui/src/6a975b5d7ec71f5cf99b615bc31d1d14e6e246db/UnityEngine.UI/EventSystem/EventSystem.cs?at=2017.1&fileviewer=file-view-default#EventSystem.cs-182

Assets/HoloToolkit/Common/Scripts/Extensions/LayerExtensions.cs
keveleigh: It's a public method though, so devs might call it in their apps too.

What/how much do we gain from removing all allocations?

Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs
keveleigh: Do you have a measured performance increase?

StephenHodgson:
Where do we call clear?

StephenHodgson:
@keveleigh, sorry made one last tweak per @Zod-

keveleigh:
@StephenHodgson Sounds good to me :)

StephenHodgson:
Should be good now.

PR #1609, opened 2018-01-08
Renamed core HLTK folders to MRTK

DDReaper:
All .MD files updated to new MixedRealityToolKit name
Spellings fixed

Overview
---
Overhaul of namespaces / classes within the project, e.g.
Rename HoloToolkit folders -> MixedRealityToolkit verbiage.
Rename namespaces like HoloToolkit.Unity to:
MixedRealityToolkit.Input.
MixedRealityToolkit.Common
MixedRealityToolkit.SpatialMapping etc.

Tasks
---
[x] Update documentation from HoloToolKit to MixedRealityToolkit
[x] Rename Core HT -> MRTK folder
[x] Rename Examples HT -> Examples MRTK folder
[x] Rename Unit Tests HT -> Unit Tests MRTK folder


Assets/MixedRealityToolkit-Examples.meta
```diff
@@ -0,0 +1,10 @@
+fileFormatVersion: 2
+guid: 65c66d518b6f7d442b6cf5e5cced7329
```
keveleigh: It might be best to rename the previously existing HoloToolkit, HoloToolkit-UnitTests, and HoloToolkit-Examples folder meta files, instead of having Unity generate new ones. I'm not sure if changing this guid under people might cause unintended issues, but it'd probably be best to (even ever so slightly) lower the amount of change we're introducing here as much as possible.

DDReaper: Ok, I'll try resurrecting the old one's and renaming them.  
As these are root folders however, it's unlikely to cause issue.

PR #1613, opened 2018-01-10
Fix for hidden members in AttachToController

StephenHodgson:
Fix for hidden members in AttachToController

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
@@ -2,9 +2,6 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using UnityEngine;
-#if UNITY_WSA && UNITY_2017_2_OR_NEWER
-using UnityEngine.XR.WSA.Input;
-#endif
 
 namespace HoloToolkit.Unity.InputModule
 {
@@ -13,14 +10,6 @@ namespace HoloToolkit.Unity.InputModule
     /// </summary>
     public class AttachToController : ControllerFinder
     {
-#if UNITY_WSA && UNITY_2017_2_OR_NEWER
-        [Header("AttachToController Elements")]
-        [SerializeField]
-        protected new InteractionSourceHandedness handedness = InteractionSourceHandedness.Left;
-#endif
-        [SerializeField]
-        protected new MotionControllerInfo.ControllerElementEnum element = MotionControllerInfo.ControllerElementEnum.PointingPose;
-
         public bool SetChildrenInactiveWhenDetached = true;
 
         [SerializeField]
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/ControllerFinder.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/ControllerFinder.cs
@@ -25,6 +25,7 @@ namespace HoloToolkit.Unity.InputModule
 
     #if UNITY_WSA && UNITY_2017_2_OR_NEWER
         public InteractionSourceHandedness Handedness { get { return handedness; } set { handedness = value; } }
+        [SerializeField]
         protected InteractionSourceHandedness handedness = InteractionSourceHandedness.Left;
     
     #endif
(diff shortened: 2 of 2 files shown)


radicalad:
Thanks Stephen, This got super messy with abstracting ControllerFInder. I'll clone and test it, but at first glance it looks fine to me.

Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
```diff
@@ -13,14 +10,6 @@ namespace HoloToolkit.Unity.InputModule
...
     public class AttachToController : ControllerFinder
     {
-#if UNITY_WSA && UNITY_2017_2_OR_NEWER
-        [Header("AttachToController Elements")]
-        [SerializeField]
-        protected new InteractionSourceHandedness handedness = InteractionSourceHandedness.Left;
-#endif
-        [SerializeField]
-        protected new MotionControllerInfo.ControllerElementEnum element = MotionControllerInfo.ControllerElementEnum.PointingPose;
-
```
radicalad: Inherited protected variables in the base class aren't exposed in the editor. We want them protected by default in the base class *ControllerFinder* so they're hidden by default when extending. *AttachToController* redefines these variables as *SerializedFields* so they're available via editor. This is why the new keyword is required. These changes remove functionality from *AttachToController*.

StephenHodgson: Ah okay, I see. Then we should name them something different.

StephenHodgson:
I'd like to address the concerns @radicalad brought up.

Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
radicalad: Are you suggesting the fix be new *SerializedField* variables in *AttachToController* with getter/setters that apply the values to the protected variable in *ControllerFinder*? That makes sense to me - avoids the *new* keyword and still lets element / handedness be editable. Why is an approach like this cleaner than overriding the existing variables with *new*? Really I wish *virtual* was an option for variables and we'd be fine.

StephenHodgson: Is the inheriting class supposed to set to be able to update/change these fields?

radicalad: Yep. That's why I was just overriding those two variables in the inheriting class - does nothing more than exposing them in the editor.

Railboy: We could use an MRTKEditor attribute to display the property in the editor. Then we could avoid using 'new' and avoid creating new serialized variables.

```
[EditableProp]
public InteractionSourceHandedness Handedness { get { return handedness; } set { handedness = value; } }
```

StephenHodgson: Well, just adding the `[SerializedField]` attribute to the base class' protected field should be enough to expose them in the editor.  If we're not trying to override or change the values in the inherited class then we don't need to override or call a new field.

StephenHodgson: I'd like to avoid the `MRTKEditor` stuff until we get that sorted out.

radicalad: Right, but that exposes it in the base class as well which I don't want.

StephenHodgson: Then why would we want to expose it in the inherited class?

StephenHodgson: The only thing I'm trying to fix in the PR is the fact that we're hiding inherited members.

radicalad: I tried to keep ControllerFinder as passive as possible - that way we can maximize the usefulness of the script. In the case of Solvers, SolverHandler extends ControllerFinder (for the controller finding ability - go figure) but handles the logic of when to get controllers differently, basically if you set the SolverHandler's enum to "head" it will bypass ControllerFinder etc. AttachToController is much more straightforward it and editor-oriented. So it makes sense to just expose those variables.

I talked with @Railboy about the intent here. The reason we're in this spot to begin with is because I got pushback about duplicating the logic that existed in AttachToController. So there are a few options:

Leave ControllerFinder alone, and then in AttachToController we have a variable that has an associated property that overrides the public property in the base class (currently in ControllerFinder, Handedness is virtual, we'd have to change Element as well) and handle the logic there.

Duplicate logic in AttachToController and they're their own things and don't try to complicate things.

Use the MRTK editor as @Railboy proposed which would expose Properties in the editor.

PR #1614, opened 2018-01-10
MRTK 2017.2.1.1 Patch Release

StephenHodgson:
Overview
---
Addresses the Patch Release for MRTK 2017.2.1.0

Changes
---
- Fixed building for Window Standalone.
- Fixed unused field warning in `BoundaryManager`
- Added unity packages from 2017.2.1.0 release
- Added unity packages for 2017.2.1.1 release

PR #1642, opened 2018-01-16
Merge Master into Dev Working Branch

StephenHodgson:
See #1614

### DIFF
--- a/Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
+++ b/Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
@@ -33,8 +33,8 @@ namespace HoloToolkit.Unity.Boundary
         // This puts the origin (0, 0, 0) on the floor if a floor has been established during setup via MixedRealityPortal.
         private TrackingSpaceType opaqueTrackingSpaceType = TrackingSpaceType.RoomScale;
 
-        //[SerializeField]
         // Removed for now, until the HoloLens tracking space type story is more clear.
+        //[SerializeField]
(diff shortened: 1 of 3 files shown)


Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
```diff
@@ -191,10 +191,6 @@ public void CalculateBoundaryVolume()
                     }
                 }
             }
-            else
```
keveleigh: Looks like this is removing something that was added in the Dev branch.

StephenHodgson: Do we really need a log that this always returns false when it's stated in the comments above?

StephenHodgson: This is kinda strange, I did a straight merge from the master branch, without any edits.

keveleigh: I think this was added by @NeerajW as a way to validate when that API becomes fixed on Unity's end.

StephenHodgson: Idk, having optional debug logs like this is probably useful.  IMO, Production apps wouldn't want this kind of thing spamming the console.

StephenHodgson: Not to mention the performance impact debug statements have in general.

Issue #1644, opened 2018-01-17
[Suggestion] Use the base open standards repository for gLTF rather than embedded classes


SimonDarksideJ:
## Overview
gLFT is a standard and the Kronos group already provide an official GitHub repo source for this functionality.

I would recommend we SubModule this functionality rather than use separate copied code in the repo

https://github.com/KhronosGroup/UnityGLTF

Unless there is some reason for providing alternate code for this feature.

StephenHodgson:
We've talked about this a few times before, but ultimately we decided submodules raise the barrier to entry for users with little to no experience using git.

DDReaper:
True, however, it is a standard used in GitHub and those users are more likely to only use the packaged MRTK asset which should include the version of the gLTF code (only the bits used by MRTK) in the package.

The primary mechanism for uses picking up the toolkit should be the compiled asset and not source.  Source is for contributors and we can provide docs for how to get started, downloading the project and submitting for contributors

StephenHodgson:
Agreed in all counts, but [using submodules directly inside a Unity project is quite complex and has some substantial setup and advanced concepts](https://www.rageagainstthepixel.com/expert-import-mrtk).

DDReaper:
Except you wouldn't have a submodule in Unity :D 
you would either have:
* Source from the asset package (which would contain the required source), each new asset uses the latest source directly.
* Source from the Repo, would be just another folder.  Only difference is that it's someone elses code.  Not directly managed by MRTK.  PR's could be needed to update the commit reference for the submodule to use a newer version.

StephenHodgson:
It would be nice if they actually published releases in the form of asset packages as well.

DDReaper:
Ha ha,  Now that's a thought. I'll raise an issue on their Site.  Would make life much easier :D

StephenHodgson:
I already did, but realized I submitted a duplicate.

DDReaper:
Yup, it's been there a while though
https://github.com/KhronosGroup/UnityGLTF/issues/7

StephenHodgson:
There's this one too:
https://github.com/KhronosGroup/UnityGLTF/issues/80

DDReaper:
Hmm, in Git lands, I think 7 trumps 80 :D

keveleigh:
> Currently we are likely lagging behind improvements from the standard. (Which was updated only days ago)

[...]

PR #1650, opened 2018-01-18
glTF new drop and updates

keveleigh:
Overview
---
This update includes the latest drop from https://github.com/KhronosGroup/UnityGLTF. This drop cleans up quite a bit of excess code, as the schema scripts are now found in the GLTFSerialization DLL.

It removes a script I created (GLTFComponentStreamingAssets) in favor of new support in this drop.

It also replaces our previous (incorrect, as we are not using specular materials) use of the Unity Standard (Specular) shader in favor of the project's GLTFStandard shader.

Changes
---
Related to #1644


Assets/HoloToolkit-Examples/Input/Materials/HoloToolkit_Focusable.mat
```diff
@@ -8,7 +8,8 @@ Material:
   m_PrefabInternal: {fileID: 0}
   m_Name: HoloToolkit_Focusable
   m_Shader: {fileID: 4800000, guid: f1f3948b20c230044bda31b620ddd37f, type: 3}
-  m_ShaderKeywords: _EMISSION
+  m_ShaderKeywords: _EMISSION _SPECULARHIGHLIGHTS_ON _USEAMBIENT_ON _USEDIFFUSE_ON
```
StephenHodgson: Does updating this material required for the gLTF update?

Assets/HoloToolkit/Utilities/Scripts/GLTF/Scripts/GLTFComponent.cs.meta
```diff
@@ -1,11 +1,11 @@
 fileFormatVersion: 2
-guid: 3628db10e32278e499da95887f1bc989
-timeCreated: 1488842155
+guid: 9087a95f22bcee64ebc45f6deec08714
```
StephenHodgson: Will this meta file change break the scene references in other people's projects?

DDReaper:
My only concern is the same as @StephenHodgson regarding the .meta files, should try to retain the original GUID's if we can.  Unless these are all new from the source repo.

A few concerns with the Kronos scripts, but they need to be addressed in the base repo, not here.

Assets/HoloToolkit/Utilities/Scripts/GLTF/Scripts/GLTFComponent.cs.meta
keveleigh: This is an update from the source repo. Looks like they changed it there, so I'm leaving this edit here to match it.

Assets/HoloToolkit-Examples/Input/Materials/HoloToolkit_Focusable.mat
keveleigh: The MotionControllerTest scene's cube is broken without it. I suppose it's not required for the glTF drop itself, but it fixes the main test scene for the glTF models.

PR #1659, opened 2018-01-23
Fix for hidden members in AttachToController

### LINKED ISSUE #1661: Missing Script References in SolverExamples
## Overview
Missing Script references in the SolverExamples scene.

48a9ac506c430325340e248c8421604e54de1804

StephenHodgson:
Overview
---
Fix for hidden members in AttachToController.  Second attempt.

This time I simply made ControllerFinder Abstract so no one could use it directly on a GameObject.  I couldn't think of any reason to really have it unless you needed to inherit anyway. (Mostly a helper).

Added a industry standard editor for SolverHandler to make sure we don't show the fields we didn't need to change.

Also added editors for the Controller Finder and Attach To Controller classes for aesthetics.

image

Changes
---
Fixes #1661

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
@@ -16,16 +16,16 @@ namespace HoloToolkit.Unity.InputModule
         public bool SetChildrenInactiveWhenDetached = true;
 
         [SerializeField]
-        protected Vector3 positionOffset = Vector3.zero;
+        protected Vector3 PositionOffset = Vector3.zero;
 
         [SerializeField]
-        protected Vector3 rotationOffset = Vector3.zero;
+        protected Vector3 RotationOffset = Vector3.zero;
 
         [SerializeField]
-        protected Vector3 scale = Vector3.one;
+        protected Vector3 ScaleOffset = Vector3.one;
 
         [SerializeField]
-        protected bool setScaleOnAttach = false;
+        protected bool SetScaleOnAttach = false;
 
         public bool IsAttached { get; private set; }
 
@@ -38,7 +38,7 @@ namespace HoloToolkit.Unity.InputModule
 
 #if UNITY_WSA && UNITY_2017_2_OR_NEWER
             // Look if the controller has loaded.
-            if (MotionControllerVisualizer.Instance.TryGetControllerModel(handedness, out ControllerInfo))
+            if (MotionControllerVisualizer.Instance.TryGetControllerModel(Handedness, out ControllerInfo))
             {
                 AddControllerTransform(ControllerInfo);
             }
@@ -50,7 +50,7 @@ namespace HoloToolkit.Unity.InputModule
         protected override void AddControllerTransform(MotionControllerInfo newController)
         {
 #if UNITY_WSA && UNITY_2017_2_OR_NEWER
-            if (!IsAttached && newController.Handedness == handedness)
+            if (!IsAttached && newController.Handedness == Handedness)
             {
                 base.AddControllerTransform(newController);
 
@@ -58,12 +58,12 @@ namespace HoloToolkit.Unity.InputModule
 
                 // Parent ourselves under the element and set our offsets
                 transform.parent = ElementTransform;
-                transform.localPosition = positionOffset;
-                transform.localEulerAngles = rotationOffset;
+                transform.localPosition = PositionOffset;
+                transform.localEulerAngles = RotationOffset;
 
-                if (setScaleOnAttach)
+                if (SetScaleOnAttach)
                 {
-                    transform.localScale = scale;
+                    transform.localScale = ScaleOffset;
                 }
 
                 // Announce that we're attached
@@ -77,7 +77,7 @@ namespace HoloToolkit.Unity.InputModule
         protected override void RemoveControllerTransform(MotionControllerInfo oldController)
         {
 #if UNITY_WSA && UNITY_2017_2_OR_NEWER
-            if (IsAttached && oldController.Handedness == handedness)
+            if (IsAttached && oldController.Handedness == Handedness)
             {
                 base.RemoveControllerTransform(oldController);
 
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
@@ -13,14 +13,6 @@ namespace HoloToolkit.Unity.InputModule
     /// </summary>
     public class AttachToController : ControllerFinder
     {
-#if UNITY_WSA && UNITY_2017_2_OR_NEWER
-        [Header("AttachToController Elements")]
-        [SerializeField]
-        protected new InteractionSourceHandedness handedness = InteractionSourceHandedness.Left;
-#endif
-        [SerializeField]
-        protected new MotionControllerInfo.ControllerElementEnum element = MotionControllerInfo.ControllerElementEnum.PointingPose;
-
         public bool SetChildrenInactiveWhenDetached = true;
 
         [SerializeField]
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
@@ -2,9 +2,6 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using UnityEngine;
-#if UNITY_WSA && UNITY_2017_2_OR_NEWER
-using UnityEngine.XR.WSA.Input;
-#endif
 
 namespace HoloToolkit.Unity.InputModule
 {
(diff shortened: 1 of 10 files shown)


StephenHodgson:
>any classes inheriting from ControllerFinder will need a new custom inspector or else the most important fields will be hidden by default.

Yeah, I thought about that for a bit, and figured the same would have also been true if I went and made a custom inspector for the solver handlers instead. That was a lose, lose situation.  I decided to go with controller finder because these fields seem to be optional.

>the logic exposing them will be hidden from developers in an editor class which many will not even realize exists

I disagree. Developers who are familiar with unity will be able to tell right away that it's a custom inspector because there's no script object field at the top of the inspector.

I've also been thinking about how to make the editor scripts more accessible be mirroring the editor folder to match the scripts folder.  @DDReaper, could we address that change in your refactoring? (I think you did, but just wanted to check)

>If MRTKEditors were used that logic would be visible within the class as attributes on the fields.

I don't really see the advantage here because we're also using attributes to control the visibility if the fields in the inspector no matter which editor class we inherited from.

One thing to note, is any class that inherited from controller finder, could have an abstract base editor it needs to inherit from as well.

Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
```diff
@@ -13,27 +13,19 @@ namespace HoloToolkit.Unity.InputModule
...
-        protected new InteractionSourceHandedness handedness = InteractionSourceHandedness.Left;
-#endif
-        [SerializeField]
-        protected new MotionControllerInfo.ControllerElementEnum element = MotionControllerInfo.ControllerElementEnum.PointingPose;
-
         public bool SetChildrenInactiveWhenDetached = true;
 
         [SerializeField]
-        protected Vector3 positionOffset = Vector3.zero;
+        protected Vector3 PositionOffset = Vector3.zero;
```
keveleigh: From what I can tell, we've mostly been using lower camel case for protected names in the Toolkit.

Assets/HoloToolkit/Input/Scripts/Utilities/ControllerFinder.cs
```diff
@@ -16,33 +16,42 @@ namespace HoloToolkit.Unity.InputModule
...
+    public abstract class ControllerFinder : MonoBehaviour
     {
-        #region public members
-        public MotionControllerInfo.ControllerElementEnum Element { get { return element; } }
-        public Transform ElementTransform { get { return elementTransform; } private set { elementTransform = value; } }
-        #endregion
+        public MotionControllerInfo.ControllerElementEnum Element
+        {
+            get { return element; }
+            set { element = value; }
```
keveleigh: The old version did not have a `set`.

Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
StephenHodgson: Just following [MSDN naming conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-fields)

Assets/HoloToolkit/Input/Scripts/Utilities/ControllerFinder.cs
StephenHodgson: Indeed, I added this intentionally.

keveleigh: Why? What does changing the `Element` at runtime do?

StephenHodgson:
> Can we solve this problem by making Element and Handedness public, thus exposing them in the Inspector for anything inheriting this? What do we lose in that case?

That's what I originally did in the first PR, but @radicalad [said he didn't want these fields exposed in the editor](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1613#discussion_r161838863).

> SolverHandler extends ControllerFinder (for the controller finding ability - go figure) but handles the logic of when to get controllers differently, basically if you set the SolverHandler's enum to "head" it will bypass ControllerFinder etc.

Assets/HoloToolkit/Input/Scripts/Utilities/ControllerFinder.cs
StephenHodgson: Potentially you could update the logic being performed in the Solver Handler, or change it's position, without having to destroy and create a new instance of the component/object.

StephenHodgson:
>This is a class we expect others to inherit from, so now they all need custom inspectors?

Not every class will need a custom inspector. For example, all the solvers that inherit from Solver Handler, do not need it.

StephenHodgson:
@Railboy I added a base editor class to inherit from for easier accessibility for future classes.

keveleigh:
> all the solvers that inherit from Solver Handler

I'm not talking about Solver Handler though. I'm specifically talking about classes that inherit directly from `ControllerFinder`.

> Besides, what's so wrong with using editor scripts?

I didn't say anything was wrong with using editor scripts. What's wrong is requiring editor scripts to be written for any class that inherits from a provided abstract class in the Toolkit. Custom editor scripts for MixedRealityCameraParent, etc, are great, but having them be a requirement to see everything possible from an inherited class seems broken to me.

keveleigh:
> basically if you set the SolverHandler's enum to "head" it will bypass ControllerFinder etc.

This sounds like SolverHandler should have a custom editor script to hide these values if the enum is set to "head" in the Inspector, instead of hiding them for everybody.

Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
keveleigh: Consistency seems more important, but we should change them all if we want to follow those naming conventions.

StephenHodgson:
>having them be a requirement to see everything possible from an inherited class seems broken to me.

They're not a requirement, but they do help if you want to have a good separation of concerns for each editor class.

keveleigh:
> They're not a requirement

If I inherit from `ControllerFinder` without a custom editor, the main two settings (`Element` and `Handedness`) won't show up in the inspector, right? That makes the Inspector settings unusable.

StephenHodgson:
> This sounds like SolverHandler should have a custom editor script to hide these values if the enum is set to "head" in the Inspector, instead of hiding them for everybody.

See https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1659#issuecomment-359881956

Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
StephenHodgson: Well as far as I can tell, I'm the single biggest contributor and I follow the MSDN guidelines. 
So, it might be better to follow industry standard and fix this in a separate issue.

keveleigh:
>I decided to go with controller finder because these fields seem to be optional.

What's the purpose of `ControllerFinder` if these fields are optional? 

>the same would have also been true if I went and made a custom inspector for the solver handlers

Do we expect anybody to inherit from `SolverHandler`? `Solver` is the abstract class.

>imo, hiding fields and and using the new keywords is a bigger cause for concern.

I don't think we should hide anything. It should all be public in the base class.

>See https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1659#issuecomment-359881956

I don't see anything in there relating to the second point you quoted.

StephenHodgson:
>I don't think we should hide anything. It should all be public in the base class.

fields should never be public and breaks encapsulation rules.

keveleigh:
>Besides the Properties are public and gives access to modifying these fields if necessary.

But doesn't expose them in the editor for any class that inherits, which is the problem we're solving.

StephenHodgson:
> But doesn't expose them in the editor for any class that inherits, which is the problem we're solving.

Which is why we have an editor class for it.

radicalad:
> This time I simply made ControllerFinder Abstract so no one could use it directly on a GameObject. I couldn't think of any reason to really have it unless you needed to inherit anyway. (Mostly a helper).

I'm late to the party, but based on everything I've read, @StephenHodgson you're correct in that Controller finder isn't necessarily helpful by itself - that was why I had initially hidden the variables for the editor.

> This sounds like SolverHandler should have a custom editor script to hide these values if the enum is set to "head" in the Inspector, instead of hiding them for everybody.

I completely agree that SolverHandler should have the custom editor, it is unlikely it will be extended as it just handles the update loop. The real meat is in extending Sovler. This will minimize the editor overhad for anyone extending AttachToController OR ControllerFinder.

StephenHodgson:
>I completely agree that SolverHandler should have the custom editor

@radicalad Cool, yeah I think we're all in agreement here. Just finishing up the custom editor for the solver handler now.

StephenHodgson:
@radicalad just waiting for a reply on those missing script references in the Solver Example Scene for a final review of this PR

Assets/HoloToolkit/Input/Scripts/Utilities/AttachToController.cs
keveleigh: From a *very* quick, non-scientific search, I found 17 files that use PascalCase, 17 files that use camelCase (not counting the SWIG-generated files, which all use camelCase), and 2 files that use underscores.

StephenHodgson:
Okay, this PR is good for review.

PR #1562, opened 2017-12-21
Sharing DLL Update

markgrossnickle:
Overview
---
Back in August dimatomp fixed the AutoDiscovery bug in the sharing service:
https://github.com/Microsoft/MixedRealityToolkit/pull/106

However, as far as I can tell no one has updated the dlls or the exe for the Unity toolkit users. I'm committing mine here which have been working for HoloMaps and a couple of other apps on the Hololens but have not been tested elsewhere. 

Changes
---

I compiled new dlls and a service exe based on the updated MRTK code.

- Fixes: #493  Auto discovery: https://github.com/Microsoft/MixedRealityToolkit/issues/70

PR #1666, opened 2018-01-24
Updating the native MRTK namespace

keveleigh:
Overview
---
This PR brings over the namespace changes from the native Toolkit.

Changes
---
Related to Microsoft/MixedRealityToolkit/pull/140


Assets/MixedRealityToolkit/Sharing/Scripts/SyncModel/SyncObject.cs
```diff
@@ -124,7 +119,7 @@ private void InitializePrimitives()
             // Scan the type of object this is a look for the SyncDataAttribute
             Type baseType = GetType();
 
-#if USE_WINRT
+#if WINDOWS_UWP
```
StephenHodgson: Should we be using `UNITY_WSA` or `WINDOWS_UWP` throughout the project?

What's the difference?

DDReaper: I was going to create another issue to standardise on our #IF statements throughout the project.  As we head towards a more multi-platform architecture, having single definitions for each state will be critical.

DDReaper: #1667 created for sorting out later

Assets/MixedRealityToolkit/Common/Scripts/CrossPlatform/TypeUtils.cs
```diff
@@ -6,7 +6,7 @@
 using System.Reflection;
 #endif
 
-namespace MixedRealityToolkit.Common
```
DDReaper: Why has this namespace changed, as it is located in the "Common" folder?

Assets/MixedRealityToolkit/Common/Scripts/CrossPlatform/ReflectionExtensions.cs
```diff
@@ -1,10 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-namespace MixedRealityToolkit.Common.Extensions
```
DDReaper: Why has this namespace changed, as it is located in the "Common\Extensions" folder?
Plus it dirties the base MRTK namespace

Assets/MixedRealityToolkit/UX/Scripts/Buttons/Utilities/CompoundButtonToggle.cs
```diff
@@ -112,11 +112,8 @@ private void OnDisable()
 
             // Get the profile field of the target component and set it to the on profile
             // Store all icons in iconLookup via reflection
-#if USE_WINRT
-            FieldInfo fieldInfo = Target.GetType().GetTypeInfo().GetField("Profile");
```
DDReaper: Whereas I didn't like these blocks everywhere for reflection. Is this flattening of the lookup going to cause future issues?

DDReaper:
A little concerned over some of the namespace changes, as it breaks the new standard for naming.
Just need to agree if it's reasonable or not.
Putting classes in the base MRTK namespace should be avoided if they don't make sense there, e.g. reflection classes.

DDReaper:
Thought I had removed any and all HoloToolkit references, so good catch there @keveleigh

DDReaper:
Ok, just noticed after looking at #1562, that this PR doesn't include the recompiled and signed DLL's that the SDK code depends on?

Assets/MixedRealityToolkit/Sharing/Scripts/SyncModel/SyncObject.cs
keveleigh: `UNITY_WSA` runs both in the compiled UWP app and in the editor. From my tests, `WINDOWS_UWP` is equivalent to our use of `UNITY_WSA && !UNITY_EDITOR`, running only in the compiled UWP app.

Assets/MixedRealityToolkit/Common/Scripts/CrossPlatform/ReflectionExtensions.cs
keveleigh: These extensions implement some reflection methods that don't exist in UWP but do exist elsewhere. Since the SWIG generated files use these methods, putting this file into a different namespace would require manually adding a using statement every time they're regenerated. In the interest of cutting down on manual steps and preventing the need to edit generated files, these would have to be somewhere visible to the SWIG files.

These two files are more similar to `SystemException.cs` and `MetroHandleRef.cs` than the other extension scripts, since they're redefining things that exist elsewhere. I believe these two files used to live without any namespace.

keveleigh:
Ahh yup, good catch. I'll drop the updated binaries in here as well.

Assets/MixedRealityToolkit/Common/Scripts/CrossPlatform/TypeUtils.cs
NeerajW: I believe this is also needed by SWIG files.
@keveleigh we should atleast write a comment in these files explaining why they don't have a namesapace.

NeerajW:
>>Thought I had removed any and all HoloToolkit references, so good catch there @keveleigh
Wow I caught them in your merged PR @DDReaper  ;)

Assets/MixedRealityToolkit/UX/Scripts/Buttons/Utilities/CompoundButtonToggle.cs
StephenHodgson: See https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1666#discussion_r163919711

DDReaper:
I would challenge your point then @keveleigh / @NeerajW on the namespace.
How often do we regenerate these files? From what I can see, last time was over a year ago.

So is it not better to have a cleaner API for a minute or two handwork in the rare occasion we are ever likely to regenerated those dell's?

StephenHodgson:
Yeah, would we want to update the SWIG generator to spit out the updated namespaces?

Assets/MixedRealityToolkit/UX/Scripts/Buttons/Utilities/CompoundButtonToggle.cs
keveleigh: `USE_WINRT` was defined in a different file and not included in our builds, so this line was never called. Replacing it with what `USE_WINRT` was previously defined as caused a build error, which leads me to believe this line of code hasn't been called in a while haha.

image

StephenHodgson: Ahh, I thought it was from the reflection extensions

DDReaper:
I agree with @StephenHodgson that we should look to fix the generator to recognise the new namespaces than dirtying the API.
Can we revert those changes back for now and look to implement a separate bit of work to look at improving the automation generation later @NeerajW / @keveleigh ?

Assets/MixedRealityToolkit/Common/Scripts/CrossPlatform/ReflectionExtensions.cs
DDReaper: My point there would be (as per my comment below) then we should look to fix the generator for the future, rather than muddying / dirtying the API with rogue classes in the base namespace.
Fixing the source rather than the symptom.

Assets/MixedRealityToolkit/Common/Scripts/CrossPlatform/TypeUtils.cs
DDReaper: My Point @NeerajW is that (agreeing with @StephenHodgson ) is that we should fix the generator (in the future) rather than dirty the API.

Assets/MixedRealityToolkit/UX/Scripts/Buttons/Utilities/CompoundButtonToggle.cs
DDReaper: Yes, I had to fix those errors, which only shows up on a UWP build.
Fixed by altering it's name accordingly.

keveleigh:
> How often do we regenerate these files? From what I can see, last time was over a year ago.

That's true. Part of that was the long delay in figuring out how to properly sign the binaries. We probably would have had ~2 updates in that time period otherwise, which, true, isn't the most frequent event.

> would we want to update the SWIG generator to spit out the updated namespaces

I've been unable to figure out how to do this, but I also don't think it'd be the best route to take. This `using` would only be necessary in UWP builds, while the SWIG files themselves are usable cross-platform.

keveleigh:
I'll revert the changes, but we should address it soon. I don't think hand-editing autogenerated files is the correct way to go.

DDReaper:
Hmm, did yo unot regenerate the .PDB files as well @keveleigh ?
MixedRealityToolkit\SpatialUnderstanding\Plugins\x86\SpatialUnderstanding.pdb
"Line 56: d:\s\forrestholotoolkit\spatialunderstanding\src\buildoutput\obj\standalonerelease\win32\spatialunderstanding\dllmain.obj"

DDReaper:
Has anyone any ideas what these files are for?
\External\MixedRealityToolkit\CrossPlatform\DocSource\Doxyfile.txt

HoloToolKit references throughout them.

DDReaper:
I also got some new warnings appearing in the editor, but those are unrelated to the namespace change, so we can look at those later.

NeerajW:
@DDReaper and @StephenHodgson, @keveleigh and I agree with you on keeping namespaces to Common. We will find a way to update the Swig dumped output to include the using statement. Perhaps as part of the header dumping code.

keveleigh:
@DDReaper 
re:pdb files
I do have the updated PDB files along with the DLLs in #1670

DDReaper:
OK, looks good now. Merging

Issue #493, opened 2017-01-31
Sharing Stage Discovery Client fails to find systems to connect to


StephenHodgson:
When setting `AutoDiscoverServer == true` the Discovery Client fails to find server.

Repro Steps:

1. Open Sharing test scene.
2. Set `AutoDiscoverServer` to true on the sharing prefab.
3. Start `SharingService` from editor.
4. Observe that the client never connects.

NeerajW:
@alexdrenea any thoughts here?

DrNeuroSurg:
Any news here ?

HodgsonSDAS:
I think part of the issue is embedded into the sharing service binaries that come from the main [HoloToolkit](https://github.com/Microsoft/HoloToolkit) (which is written in C++).

I've been doing a lot of work trying to make sure everything is working again with the test scenes.
Unfortunately I haven't gotten this to work.

yacuzo:
I use the Holotoolkit library directly for sharing (not the scripts in this repo), and i have the same problem, so I made an issue at the Holotoolkit repo: https://github.com/Microsoft/HoloToolkit/issues/70.

StephenHodgson:
https://github.com/Microsoft/HoloToolkit/pull/106

davidkline-ms:
Should be fixed by #1670.

PR #1435, opened 2017-11-25
Platforms2

trzy:
Add QueryTopology_FindLargestPositionsSittable() to Unity and update the Spatial Understanding demo with this query. This is the companion to: Microsoft/HoloToolkit#86

Note: this should replace Platforms #651.

PR #1670, opened 2018-01-26
Signed MixedRealityToolkit binary updates

keveleigh:
Overview
---
This is a new drop from https://github.com/Microsoft/MixedRealityToolkit, adding versioning info (starting at 1.0.0.1 for this drop) to all the DLLs and EXEs, as well as digitally signing the binaries with the official Microsoft certificate for authenticity.

The updated Sharing binaries will come from #1666.

In the future, drops will happen as necessary from the native repo, instead of all at once like this.

Changes
---
- Unblocks: #1435 
- Supercedes: #1562 
- Fixes: #493 
- Probably others...


StephenHodgson:
@DDReaper I think he just wants to verify they work

keveleigh:
> when we already have them planned for the NameSpace Refactor?

I only have the sharing binaries planned to be updated in the namespace refactor branch. The rest can go directly into the dev branch, since there weren't any namespaces to be updated there.

StephenHodgson:
Will not including the ARM version of the Motion Controller prevent us from building for that platform target?

keveleigh:
The motion controller DLL is only used in the editor, while built apps call the platform API directly, so we don't need to ship any UWP versions of them.

PR #1678, opened 2018-01-30
Fixed int bug in column calculation of ObjectCollection

Weasy666:
Column calculation was buggy because of some `int` math.  
As an Example, when calculating columns with a `NodeList.Count` of `4` and `Rows` of `5` (both are `int` values) the result is 0. That resulted in all CollectionNodes being positioned in one place.

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/Collections/ObjectCollection.cs
+++ b/Assets/HoloToolkit/UX/Scripts/Collections/ObjectCollection.cs
@@ -176,7 +176,7 @@ namespace HoloToolkit.Unity.Collections
                     break;
             }
 
-            _columns = Mathf.CeilToInt(NodeList.Count / Rows);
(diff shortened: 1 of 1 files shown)


Weasy666:
Please wait a moment before merging. I maybe found another bug.  
Currently when i have 4 Objects in my collection and the LayoutType set to RowThanColumn shouldn't the 4 objects be positioned in one row/line? I get just a single column and changing the LayoutType to ColumnThanRow doesn't change a thing for me in the editor.

StephenHodgson:
Are you in play mode, or just in the editor?

Weasy666:
Just in the editor and pressing the `Update Collection` button.

StephenHodgson:
Make sure you use the execute in edit mode attribute.

Weasy666:
The Update button is working fine.  
Shouldn't i get a row instead of a column here?
image

Here i'm getting a row, but it's kinda forced with `Rows` set to 1.
image

Issue #1689, opened 2018-02-04
DictationInputManager and focus do not play well together


vbandi:
## Overview
InputManager is architected to only send events to focused objects. This is fine for things like air-tapping, but speech is different. You may start to speak, move your cursor away from the object, and the object will never receive the dictation. It is unrealistic to expect the user to keep his/her head on a small microphone target all through dictation.

Adding a globalEventListener is not a good solution either. In this case, the object will receive all events from other objects, too, so to mitigate this, the developer has to add tons of checks all over the place.

## Steps to reproduce
- Open up the keyboard sample in HoloLens or the editor. 
- Air-tap on the microphone icon
- move the cursor off the keyboard
## Actual Behavior
What you speak doesn't get shown in the input field.

## Expected Behavior
The keyboard (and anything using the DictationInputSource) should receive dictation text until the dictation finishes, even if the gameobject is no longer in focus.

## Suggestions for fixing

:
- Consider whether DictationInputManager is a good candidate to being an InputModule at all or if it should be handled differently
 - If it is to stay an InputModule, provide a mechanism for gameobjects that initiated the dictation to keep receiving dictation events regardless of focus as long as the dictation session is active. E.g. StartDictation should have an optional IDictationHandler parameter, and if not null, dictation events should be passed to this object until dictation is complete.

## Unity Editor Version
2017.3

## Mixed Reality Toolkit Release Version
2017.2.0.1

vbandi:
Seems to have solved my issue for now. Thanks!

PR #1690, opened 2018-02-04
Dictation Focus

StephenHodgson:
Overview
---
Added modal focus for the currently listening game object during dictation.

Changes
---
- Fixes: #1689

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/InputSources/DictationInputManager.cs
+++ b/Assets/HoloToolkit/Input/Scripts/InputSources/DictationInputManager.cs
@@ -180,6 +180,7 @@ namespace HoloToolkit.Unity.InputModule
             if (hasListener)
             {
                 InputManager.Instance.PopModalInputHandler();
+                hasListener = false;
             }
 
             Microphone.End(DeviceName);
--- a/Assets/HoloToolkit/Input/Scripts/InputSources/DictationInputManager.cs
+++ b/Assets/HoloToolkit/Input/Scripts/InputSources/DictationInputManager.cs
@@ -53,6 +53,7 @@ namespace HoloToolkit.Unity.InputModule
 
         private static bool isTransitioning;
         private static bool hasFailed;
+        private static bool hasListener;
 #endif
 
         #region Unity Methods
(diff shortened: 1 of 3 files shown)


Assets/HoloToolkit/Input/Scripts/InputSources/DictationInputManager.cs
```diff
@@ -103,11 +103,12 @@ protected override void OnDestroy()
         /// <summary>
         /// Turns on the dictation recognizer and begins recording audio from the default microphone.
         /// </summary>
+        /// <param name="listener">GameObject listening for the dictation input.</param>
         /// <param name="initialSilenceTimeout">The time length in seconds before dictation recognizer session ends due to lack of audio input in case there was no audio heard in the current session.</param>
         /// <param name="autoSilenceTimeout">The time length in seconds before dictation recognizer session ends due to lack of audio input.</param>
         /// <param name="recordingTime">Length in seconds for the manager to listen.</param>
         /// <returns></returns>
-        public static IEnumerator StartRecording(float initialSilenceTimeout = 5f, float autoSilenceTimeout = 20f, int recordingTime = 10)
+        public static IEnumerator StartRecording(GameObject listener, float initialSilenceTimeout = 5f, float autoSilenceTimeout = 20f, int recordingTime = 10)
```
keveleigh: What if I'm using a global listener and I don't want/need to pass a specific GameObject in? Can we give `listener` a default value (probably `null`), like we've done with the others? That'd also prevent this from being a breaking change.

StephenHodgson: Good question. 

I don't think it's a good idea to make the object you're using to start the dictation a global listener.
But if you did, you'd have a script on the object that listens for the click handler:

```
OnInputClicked(EventData eventData)
{
    StartCoroutine(DictationManager.Instance.StartRecording(gameObject);
{
```

keveleigh: Ahh, I didn't mean the starting object being a global listener, but the one being sent the dictation results. If I want a button to start recording, but I want the results to be sent to a global listener (or two), my button would be forced to pass in some GameObject anyway.

StephenHodgson: I don't think I follow.

StephenHodgson: Okay, I'll make the listener optional.

PR #1708, opened 2018-02-09
Fix build for non windows platforms

GAnatoliy:
Overview
---

Using UnityEngine.Windows.Speech namespace prevents build for other platforms than UWP (e.g. iOS).
Platform dependent compilation is used in other places in code, e.g. HoloToolkit.UI.Keyboard.Keyboard

But it is missed in HoloToolkit.Unity.Buttons.CompoundButtonSpeech that prevents build for ios.

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/Buttons/CompoundButtonSpeech.cs
+++ b/Assets/HoloToolkit/UX/Scripts/Buttons/CompoundButtonSpeech.cs
@@ -3,9 +3,6 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 //
 using UnityEngine;
-#if UNITY_WSA || UNITY_STANDALONE_WIN
-using UnityEngine.Windows.Speech;
-#endif
 using HoloToolkit.Unity.InputModule;
 using HoloToolkit.Unity;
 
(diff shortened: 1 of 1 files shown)


StephenHodgson:
@DDReaper was this because of the namespace refactor?

keveleigh:
@StephenHodgson Looks like this is going into master, so unrelated to the refactor. We should probably take this in Stabilization?

StephenHodgson:
Stabilization branch should build to other platforms just fine. I tested it before making it.

StephenHodgson:
If anything, this class needs to use the ISpeechHandler.

keveleigh:
Which platforms did you try? This PR calls out iOS specifically, but I'd imagine a `Windows` namespace would fail to build on any non-Windows platform.

StephenHodgson:
yeah, the appropriate fix is for the class to use the speech input handler.

StephenHodgson:
Ahh maybe that stuff needs to get removed then.  There's no reason to have it.  The source should handle the confidence level and such.

StephenHodgson:
@keveleigh doesn't build for Android either

StephenHodgson:
Also doesn't build for webGL.  Should we support that as well?

StephenHodgson:
@GAnatoliy even with this change, we have a [dependency on glTF which also has issues compiling for iOS.](https://github.com/KhronosGroup/UnityGLTF/issues/97).

DDReaper:
Any reason this change is just removing references and not holding them behind a precompiler flag?

I agree however, if MRTK is to support projects built for other platforms, it will need a much wider review than the changes here.

keveleigh:
@DDReaper it looks like the references were left over from [a conversion from the MRDL](https://github.com/Microsoft/MRDesignLabs_Unity_Tools/blob/master/HUX/Scripts/Buttons/CompoundButtonSpeech.cs) to the Toolkit's `ISpeechHandler`. Of course, now it isn't clear where (if at all) the keyword from the button's text is actually being registered on a recognizer, so I think this might be broken anyway.

PR #1711, opened 2018-02-11
UX Namespace Changes

StephenHodgson:
Overview
---
- There was a duplicate class named KeyboardManager in two difference namespaces.
    - Updated the UX KeyboardManager name to VirtualKeyboardManager.
- Moved Cursor Scripts from Input feature folder to UX feature folder and updated namespaces.
- Moved Cursor Prefabs from Input feature folder to UX feature folder.
- Moved other UX specific assets from Input feature folder to UX feature folder.
- Cleaned up and removed a few duplicate assets.
- Organized the Input and UX editor folders to mirror their script folder counterparts.


Assets/MixedRealityToolkit/InputModule/Prefabs/HoloLensCamera.prefab
```diff
@@ -52,7 +52,6 @@ GameObject:
   m_Component:
   - component: {fileID: 4000010560871102}
   - component: {fileID: 20000013304850028}
-  - component: {fileID: 92000010221946138}
```
DDReaper: I thought we removed this prefab some time back?  How come it's back?

Assets/MixedRealityToolkit/InputModule/Prefabs/MotionControllers.prefab
```diff
@@ -1,58 +0,0 @@
-%YAML 1.1
```
DDReaper: Did you mean to remove this prefab? Can't see it elsewhere

Assets/MixedRealityToolkit/UX/Scripts/Keyboard/KeyboardInputField.cs
```diff
@@ -6,7 +6,7 @@
 using UnityEngine.EventSystems;
 using UnityEngine.UI;
 
-namespace MixedRealityToolkit.UX.Keyboard
+namespace MixedRealityToolkit.UX.VirtualKeyboard
```
DDReaper: If we've moved these to a new VirtualKeyboard namespace, the folder name should be renamed as well..

Assets/MixedRealityToolkit/InputModule/Scripts/Focus/SimpleSinglePointerSelector.cs
```diff
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using MixedRealityToolkit.InputModule.Cursor;
 using MixedRealityToolkit.InputModule.EventData;
 using MixedRealityToolkit.InputModule.Gaze;
 using MixedRealityToolkit.InputModule.InputHandlers;
 using MixedRealityToolkit.InputModule.InputSources;
+using MixedRealityToolkit.UX.Cursors;
```
NeerajW: This sucks that now to use simple Gaze and Cursor in my app I need both folders Input and UX?

NeerajW: This scenario is needed by pretty much every app. I am supportive of this. Sorry if I should have caught this earlier.

NeerajW: Perhaps a UX folder under Input for things that depend purely on Input like Cursor and Motion Controllers?

Assets/MixedRealityToolkit/InputModule/Prefabs/MotionControllers.prefab
StephenHodgson: I don't believe we were using it any longer. The attach to controller and motion controller visualization scripts made this obsolete. Also it was missing the script reference because it was removed.

Assets/MixedRealityToolkit/InputModule/Scripts/Focus/SimpleSinglePointerSelector.cs
StephenHodgson: I thought we agreed that all visual representations would be UX?

Also, after my input refactor we won't have the requirements to use a cursor on our pointers.

StephenHodgson: No matter how we approach this, most likely the ux will be dependent on the input folder in some way. But we won't have the Input depending on the ux.

Assets/MixedRealityToolkit/InputModule/Prefabs/MotionControllers.prefab
StephenHodgson: Okay, so I went back to look. This got wrapped up into the MixedRealityCameraPartent prefab.  But it's not been deprecated. The Visualizer was the missing script reference on this. I can revert, but not sure if it's really needed if it's already on the camera prefab.

Assets/MixedRealityToolkit/InputModule/Prefabs/HoloLensCamera.prefab
StephenHodgson: Yup, this prefab was never removed.

Assets/MixedRealityToolkit/InputModule/Prefabs/MotionControllers.prefab
StephenHodgson: Okay, I'll be sure to put it back and fix the script reference.

Assets/MixedRealityToolkit/InputModule/Scripts/Focus/SimpleSinglePointerSelector.cs
DDReaper: When you say "I need to use both UX and Input folders in my project" @NeerajW it makes me shudder.

IMHO, the MRTK project is a full fledged asset, meant to be used / shipped as a whole.  It doesn't bloat the project or cause any issue by leaving the main MRTK folder intact.  in some way, that would be wholly resolved by having all the code wrapped in a single DLL (but let's not go there)

With Unity, there isn't any practical reason to remove asset code from a project and if people feel they need to , then we need to address that.  Simply put, just removing "stuff" without understanding why people do it is likely to cause runtime issues later because they removed too much.

DDReaper:
Seems one of my other comments and it seems my comment on moving the InputManager (and other core managers) from the InputModule\Utilities\Managers went astray
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1711/files#diff-384f5a9b4efc7a5f7ff67e4cffbb24d0

IMHO, as these are in the top level InputModule namespace, then those scripts should live at the top level as their namespace dictates.  files should match their namespace locations for ease of reference.

So, either change their namespace back to InputModule.Utilities.Managers or have the scripts back in their original folder. (i'm a bit particular about API architecture, if you can tell :P)

Assets/MixedRealityToolkit/InputModule/Scripts/Focus/SimpleSinglePointerSelector.cs
NeerajW: @StephenHodgson I don't recall a discussion around us agreeing on everything UX but I could be wrong for sure.

@DDReaper what you said is the exact opposite feedback I get. People don't want to import a ton of code into their projects. Hence I did the semi-modularization by reducing cross feature dependencies.

My scenario is fairly simple as I stated above: In my app I need Input from gaze and I need to visualize it. Looks like I will need a lot of code to do that contained in two separate feature folders Input and UX. Previously I could just import Input and be on my way.
These type of massive code pieces are a part of what deter people from using and contributing.

I agree from an asset prespective having things in UX makes sense. I just want to make sure we address the feedback on MRTK around poor modularization. :(

Assets/MixedRealityToolkit/InputModule/Prefabs/MotionControllers.prefab
NeerajW: Exactly, I believe this is a child of the MRCameraParent.

Assets/MixedRealityToolkit/InputModule/Scripts/Focus/SimpleSinglePointerSelector.cs
NeerajW: May be we solve this with Unity packages for scenarios. @keveleigh thoughts? We chatted on this yesterday briefly.

DDReaper: Based on this @NeerajW , then we may need a small re-think on the Multi-VR solution to breed more modularization, if we intend (rather large) chunks to be able to be split off.

StephenHodgson:
Yeah there wasn't an easy way to reply to that.  

I noticed there was still some stragglers in the folder, so we should move those up too.

Assets/MixedRealityToolkit/InputModule/Scripts/Focus/SimpleSinglePointerSelector.cs
StephenHodgson: @DDReaper I think we should def focus on making some feature areas decoupled. For example spatial mapping and understanding shouldn't have dependencies on input for example.

DDReaper:
Any update @StephenHodgson ? I know you were busy last week, would be good to get this cleaned up and closed down

StephenHodgson:
Been in vacation. Might not be able to address this until after Wednesday.

Assets/MixedRealityToolkit/InputModule/Prefabs/MotionControllers.prefab
StephenHodgson: right, if it's on the MRCameraParent, do we need it as it's own prefab?

StephenHodgson:
I'm just going to do this in vNEXT

PR #1717, opened 2018-02-14
Resolve future issues... now

DDReaper:
Resolved two issues that will become a problem in 2018, as they are fixable now it seems sensible to resolve them.

Overview
---
2018 highlighted some obsolete code / functions that were now removed.  The new versions of these functions exist now, so we should fix them ahead of them being a problem

Changes
---
- BuildsNSLTools updated to force conversion of it's build output to a string (as Unity will change it from a string to a Type)
 - Changed the Destroy method used in Tagalong script (as destroyobject is scheduled for removal)

Remaining Issues
---
 - EnumMaskField is being replaced with EnumFlagsField (but not in 2.17.2 build so cannot be fixed now)
 - WSA Images are getting tidied up (removing old types and Phone), we should probably remove these now, but need to validate the effect on current WSA builds

### DIFF
--- a/Assets/HoloToolkit/Utilities/Scripts/Tagalong.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/Tagalong.cs
@@ -291,7 +291,7 @@ namespace HoloToolkit.Unity
                             {
                                 Light clonedLight = Instantiate(DebugPointLight, closestHit.point, Quaternion.identity) as Light;
                                 clonedLight.color = Color.red;
-                                DestroyObject(clonedLight, 1.0f);
+                                Destroy(clonedLight, 1.0f);
                             }
 #if UNITY_EDITOR
                             DebugDrawLine(DebugDrawLines, cameraPosition, targetCoord, Color.red);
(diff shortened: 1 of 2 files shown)


DDReaper:
Confirmed, we can replace the usage of EditorGUILayout.EnumMaskField with EditorGUILayout.EnumFlagsField for 2017.3

Assets/HoloToolkit/Utilities/Scripts/Tagalong.cs
```diff
@@ -291,7 +291,7 @@ private bool AdjustTagalongDistance(Vector3 cameraPosition, out Vector3 newPosit
                             {
                                 Light clonedLight = Instantiate(DebugPointLight, closestHit.point, Quaternion.identity) as Light;
                                 clonedLight.color = Color.red;
-                                DestroyObject(clonedLight, 1.0f);
+                                Destroy(clonedLight, 1.0f);
```
StephenHodgson: Do we want to check if application is editor and use destroy immediate if true?

DDReaper: Possibly but this PR was only to future proof the current functionality rather than change it.
Better we have a separate PR to review "Destroy"s use through out the project?

PR #1745, opened 2018-02-23
Turn off the influence of the physics engine while dragging.

michaellperry:
If you attach both RigidBody and HandDraggable to an object, then it will be affected by both gravity and the user's motion controller while dragging. The result is that it flops on the ground like a struggling flounder. This change turns off the influence that the physics engine has over the object while the user is in control of it.

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/HandDraggable.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/HandDraggable.cs
@@ -64,6 +64,7 @@ namespace HoloToolkit.Unity.InputModule
         private IInputSource currentInputSource;
         private uint currentInputSourceId;
         private Rigidbody hostRigidbody;
+        private bool hostRigidbodyWasKinematic;
 
         private void Start()
         {
@@ -120,6 +121,7 @@ namespace HoloToolkit.Unity.InputModule
             isDragging = true;
             if (hostRigidbody != null)
             {
+                hostRigidbodyWasKinematic = hostRigidbody.isKinematic;
                 hostRigidbody.isKinematic = true;
             }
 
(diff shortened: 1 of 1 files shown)


DDReaper:
I don't personally see any issue here but want a second opinion, @StephenHodgson

michaellperry:
Added the braces.

Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/HandDraggable.cs
```diff
@@ -293,6 +297,10 @@ public void StopDragging()
             isDragging = false;
             currentInputSource = null;
             currentInputSourceId = 0;
+            if (hostRigidbody != null)
+            {
+                hostRigidbody.isKinematic = false;
```
keveleigh: What if `isKinematic` was true to begin with?

DDReaper: That's a fair point, if the hand draggable object was never meant to be affected, then this would change it's behaviour in the scene. Nice catch @keveleigh 

@michaellperry two suggestions:
1: Make this configurable, so the user can opt to have the IsKinematic enabled / disabled with this option (Name it something like "FreezePhysicsWhenGrabbed").  Only do the behaviour if enabled.
2: Alter the flow of the change to:
* Have a private "Altered" bool to denote the setting was changed
* Only set  "Altered" bool to true if isKinematic was altered (e.g. from true to false)
* Only set isKinematic back to true of the "Altered" bool was true (and reset "Altered" back to false)

michaellperry:
I decided to go with option number 2. Option 1 would introduce a new degree of freedom that could accidentally be set incorrectly. Please have a look.

Issue #1679, opened 2018-01-31
DebugLog of CanShareAnchors is wrong


robonich:
https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/0d54a2df552833d1cda692da62bc9df7c09a589d/Assets/HoloToolkit/SharingWithUNET/Scripts/PlayerController.cs#L288

I think the true line is 
```csharp
Debug.LogFormat("local player {0} share anchors ", (CanShareAnchors ? "does" : "does not"));
```

It's a very easy mistake but it confused me when I debugged UNET sharing, so please fix it.

PR #1748, opened 2018-02-23
A few more small bug fixes

keveleigh:
Overview
---
Updates the debug log from a SharingWithUNET component to be correct.

Reverts the addition of `requiredComponent: ` from `HapticsTest` which was preventing the editor from working.
image

Updates the `TryGetGeometry` call in `BoundaryManager` to actually return something in WMR.

Changes
---
- Fixes: #1679


StephenHodgson:
@DDReaper _is_ this broken in stabilization?

DDReaper:
If this is only being fixed now in stabilisation, then it'll be unchanged in dev, so might as well ensure we fix forward for reference

Issue #1760, opened 2018-02-27
Send Slider Value on Update


Gaweph:
## Overview
Allows for value to be sent when invoking **OnUpdateEvent** event of slider control (Slider Gesture Control).
## Expected Behavior
Ability to specify an object method that accepts the updated value of the slider
## Actual Behavior
Slider does not send value when invoking update method.
## Steps to reproduce
Select an object and a method to invoke.  Must specify the value to pass through.
![image](https://user-images.githubusercontent.com/6419944/36723955-9841d088-1ba9-11e8-9e70-f8cee14941bd.png)

## Unity Editor Version
2017.3.0f3 Personal
## Mixed Reality Toolkit Release Version
2017.2.1.0

PR #1761, opened 2018-02-27
Send Slider Value on Update

Gaweph:
Overview
---
Allows for value to be sent when invoking **OnUpdateEvent** method.

Changes
---

Can Specify a method that accepts a float

image

image

Previous behavior works correctly (specifying a method and supplying a value)

image

image

- Fixes: #1760 .

### DIFF
--- a/Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
+++ b/Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using UnityEngine;
-using UnityEngine.Events;
+using HoloToolkit.Unity;
 
 namespace MixedRealityToolkit.Examples.UX.Controls
 {
@@ -24,7 +24,7 @@ namespace MixedRealityToolkit.Examples.UX.Controls
         public GameObject CenteredDot;
 
         [Tooltip("Sends slider event information on Update")]
-        public UnityEvent OnUpdateEvent;
+        public UnityEventFloat OnUpdateEvent;
 
         /// <summary>
         /// The value of the slider
@@ -36,7 +36,7 @@ namespace MixedRealityToolkit.Examples.UX.Controls
                 if (mSliderValue != value)
                 {
                     mSliderValue = value;
-                    OnUpdateEvent.Invoke();
+                    OnUpdateEvent.Invoke(mSliderValue);
                 }
             }
             get
--- a/Assets/MixedRealityToolkit/Utilities/Scripts/UnityEventFloat.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Scripts/UnityEventFloat.cs
@@ -1,0 +1,9 @@
+﻿using UnityEngine.Events;
+
+namespace HoloToolkit.Unity
+{
+    [System.Serializable]
+    public class UnityEventFloat : UnityEvent<float>
+    {
+    }
+}
--- a/Assets/MixedRealityToolkit/Utilities/Scripts/UnityEventFloat.cs.meta
+++ b/Assets/MixedRealityToolkit/Utilities/Scripts/UnityEventFloat.cs.meta
@@ -1,0 +1,13 @@
+fileFormatVersion: 2
+guid: 536378fce6b62754e8e9607c2ace31b2
+timeCreated: 1519726146
+licenseType: Free
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 


Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using UnityEngine;
-using UnityEngine.Events;
+using HoloToolkit.Unity;
```
DDReaper: Please check namespaces, as these were all updated to MixedRealityToolkit

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
```diff
@@ -24,7 +24,7 @@ public class SliderGestureControl : GestureInteractiveControl
         public GameObject CenteredDot;
 
         [Tooltip("Sends slider event information on Update")]
-        public UnityEvent OnUpdateEvent;
+        public UnityEventFloat OnUpdateEvent;
```
StephenHodgson: Why are we even using a unity event here anyway?

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
Gaweph: This refers to the namespace where the **UnityEventFloat** class is located (added as part of this PR).  Should this class use a different namespace (it matches those present alongside it)?

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
Gaweph: I believe events are being used so that one or many callback methods can be specified from within the editor.

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
Gaweph: I did not notice that the namespace had changed in dev.  Should I change the namespace? Or should a minor change like this target the stabilisation branch? (This is my first PR).  Advice welcome.

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
StephenHodgson: yeah that's exactly the problem. It's better to do it in code.

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
StephenHodgson: Best bet is to retarget the PR for the stabilization branch if you want to see these fixes in the next release

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
Gaweph: I do not agree.  If a user wants to use the slider prefab and drags it onto the scene, then the current implementation allows for them to wire up some functionality without writing any code.

This is also true for the current implementation of the buttons, checkbox and other UX prefabs provided by the holotoolkit-examples package

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
Gaweph: OK, thanks.

Assets/MixedRealityToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
StephenHodgson: It's not any different than having a float field and a game object field and referencing it that way.

Believe me, it's a lot easier to deal with.

StephenHodgson: > This is also true for the current implementation of the buttons, checkbox and other UX prefabs provided by the holotoolkit-examples package.

Yeah, those are also doing it wrong.

Gaweph: I think I understand, you could just have a reference to a game object instead and have the code change that value directly on update.

The current implementation of the controls using the event system does however allow for other things, such as triggering multiple methods on multiple objects.

Gaweph: I think this discussion is addressing a larger issue, namely "should the toolkit controls use unity events or some other system".

This PR is addressing a minor issue in that the update event of the slider does not allow the updated value to be sent.  

Would it be pragmatic to add this update to the current system?  And discuss a refactor to the controls event system in another issue?  I believe that there will be arguments for both sides of this discussion.

Gaweph:
Hello,

@StephenHodgson Should I close this PR?  As  at present I am not going to rework the example controls to use something else other than events, I was simply trying to resolve a minor problem with the current events system with this change.

StephenHodgson:
You can leave or open. It's not too radical a change. I had planned on opening a PR against your branch but didn't have time over the weekend. Keep your eyes peeled.

Gaweph:
Ok, I look forward to seeing what you come up with 👍

StephenHodgson:
Strange, I pulled your branch and I'm seeing different things there than here in this PR.

Gaweph:
Closing this PR will reopen against the Stabilization Branch

Issue #1730, opened 2018-02-19
Ambiguous call GetTypeInfo() in dev branch with .Net 4.6 runtime version.


eirikhollis:
## Overview
When importing the latest dev branch with .Net 4.6, the following error appears:

MixedRealityToolkit/Sharing/Scripts/SyncSettings.cs(43,33): error CS0121: 
The call is ambiguous between the following methods or properties: 'MixedRealityToolkit.Common.Extensions.ReflectionExtensions.GetTypeInfo(this System.Type)' and 'System.Reflection.IntrospectionExtensions.GetTypeInfo(this System.Type)'

## Expected Behavior
No errors.

## Actual Behavior
Ambiguous call in [SencSettings.cs#L43](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/Dev_Working_Branch/Assets/MixedRealityToolkit/Sharing/Scripts/SyncSettings.cs#L43).

## Steps to reproduce
Import latest dev branch in 2017.3.1p1.
Switch Scripting Runtime Versions to Experimental .Net 4.6
Error appears after restarting Unity.

## Unity Editor Version
2017.3.1p1

## Mixed Reality Toolkit Release Version
dev_branch 19.02.2018

StephenHodgson:
Chances are this will be removed when we make the MRTK target 4.6 by default.

genereddick:
Same issue for me. My temporary work around removes the error though not sure about any actual impact on functionality:

Changed MixedRealityToolkit.Sharing.SyncSettings LINE 38: 
    var typeInfo = type.GetTypeInfo();
To:
    var typeInfo = ReflectionExtensions.GetTypeInfo(type);

StephenHodgson:
Yup, it's a problem with the namespaces.

DDReaper:
I did resolve this in the namespace update, so unsure as to why it's happening again.

StephenHodgson:
I vote we just move the project up to .NET 4.x and just remove the offending code.
This will enable us to do a lot of awesome new stuff (like use the device portal wrapper)

keveleigh:
I'd be hesitant to move to 4.x until Unity stops labeling it "experimental". Is there a known timeframe for that?

Of course, I also have little personal experience yet with how "experimental" it still is.

StephenHodgson:
Everything in Unity is experimental.

keveleigh:
[...]

PR #1772, opened 2018-02-27
.NET 4.6 Compatibility

StephenHodgson:
Fixed issue with compiling against .NET 4.6 and the `GetTypeInfo` Reflection Extension

Changes
---
- Fixes: #1730

### DIFF
--- a/Assets/HoloToolkit/Common/Scripts/CrossPlatform/Reflection/ReflectionExtensions.cs
+++ b/Assets/HoloToolkit/Common/Scripts/CrossPlatform/Reflection/ReflectionExtensions.cs
@@ -179,11 +179,12 @@ namespace HoloToolkit
 
     public static class ReflectionExtensions
     {
+#if !NET_4_6
         public static Type GetTypeInfo(this Type type)
         {
             return type;
         }
-
+#endif
         public static Type AsType(this Type type)
         {
             return type;


Assets/HoloToolkit/Common/Scripts/CrossPlatform/Reflection/ReflectionExtensions.cs
```diff
@@ -179,11 +179,12 @@ public static object[] GetCustomAttributes(this Type type, Type attributeType, b
 
     public static class ReflectionExtensions
     {
+#if !NET_4_6
         public static Type GetTypeInfo(this Type type)
```
DDReaper: So, what do we use in place of this and does omitting it for standalone cause issue?

Assets/HoloToolkit/Common/Scripts/CrossPlatform/Reflection/ReflectionExtensions.cs
```diff
@@ -179,11 +179,12 @@ public static object[] GetCustomAttributes(this Type type, Type attributeType, b
 
     public static class ReflectionExtensions
     {
+#if !NET_4_6
```
keveleigh: Does this define work? I'm still seeing the error in 2017.2.1f1.

image

StephenHodgson: Restart the editor.

Assets/HoloToolkit/Common/Scripts/CrossPlatform/Reflection/ReflectionExtensions.cs
StephenHodgson: The platform target shouldn't matter.

Assets/HoloToolkit/Common/Scripts/CrossPlatform/Reflection/ReflectionExtensions.cs
StephenHodgson: https://docs.unity3d.com/Manual/PlatformDependentCompilation.html

keveleigh: I know it exists in documentation, but I've never seen it work.

I just restarted the editor, deleted all the Unity generated files and folders, and I still see the error.  Which Unity version are you using?

StephenHodgson: 2017.2, 2017.1, and 2017.3

keveleigh: Strange, I tried 
* 2017.3.1f1
* 2017.2.1p2
* 2017.2.1f1
* 2017.1.3f1
* 5.6.5f1

and 5.6 was the only one that didn't give me the error.

StephenHodgson: What's stranger, is that I never get had the issue in the first place.  I wonder if it has something to do with what .NET workloads are installed on the machine.

StephenHodgson: The ambiguous reference is exactly what we're trying to fix here.

StephenHodgson: Are you sure you pulled in the fix?

keveleigh: Aha! I went back and double checked everything, and I realized you've based this on the Stabilization branch. The issue isn't present there, only in the Dev_Working branch.

StephenHodgson: Yeah the same will need to be addressed in the dev branch.  Although I'd like to see MRTK vNEXT use strictly use .NET 4.6.

keveleigh: This issue isn't present in Stabilization at all, even without this change.
This fix doesn't fix the Dev branch, since it seems like the define doesn't work like it should. It seems the code isn't able to disambiguate between extension methods in different namespaces. I think the issue is that the Toolkit extensions are now in a namespace in the Dev branch, with a corresponding `using` which didn't used to be present.

StephenHodgson: You're right.

Issue #1784, opened 2018-03-02
Scene with spatial understanding components crashes on IL2CPP (HoloLens)


Holo-Krzysztof:
## Overview
I wanted to have a simple scene that visualizes the spatial mapping mesh as the HoloLens scans. The example scene only worked in the editor, so I've rebuilt the scene with the same prefabs and deployed it on a HoloLens which worked on the .NET backend. However, when on IL2CPP it crashes when calling the native function GeneratePlayspace_InitScan.

I looked into the native sources and the underlying DLL seems to use the cdecl calling convention. However, as far as i know the CLR on Windows x86 defaults to stdcall for P/Invoke, so I'm not entirely sure why it works at all if that is the case.

## Expected Behavior
The spatial mesh is rendered as is the case on .NET backend.

## Actual Behavior
The app crashes with the error message "Run-Time Check Failure #0 - The value of ESP was not properly saved across a function call.  This is usually a result of calling a function declared with one calling convention with a function pointer declared with a different calling convention."

## Steps to reproduce
- Clone the repo at https://github.com/Holo-Krzysztof/SpatialMappingCrashRepro
- Build the project with IL2CPP (default for UWP since Unity 2017.2)

## Unity Editor Version
2017.2.1
## Mixed Reality Toolkit Release Version
2017.2.1

StephenHodgson:
Might want to open this on the main Mixed Reality Toolkit

Holo-Krzysztof:
The bug is in the bindings though; I guess there's no error on the native side.
It'd be possible to change the calling convention to __stdcall there, but I'm not sure whether that's desired or if it breaks other platforms. So I think the better solution is to annotate the `DllImport`s with `CallingConvention.Cdecl`.

StephenHodgson:
Ahh, thanks for clearing that up.

Holo-Krzysztof:
I'm going to try to do that and report back if it works.

PR #1787, opened 2018-03-03
Changed SpatialUnderstanding DllImports to cdecl.

Holo-Krzysztof:
Overview
---
Explicitly added calling convention to SpatialUnderstanding imports to prevent application from crashing when built using IL2CPP.

Changes
---

- Fixes: #1784 .


Holo-Krzysztof:
@keveleigh That may very well be the case - I didn't look too much at other managed imports but while searching the native DLL I've found that exports use the EXTERN_C macro from winnt.h which implies __cdecl.
If the managed imports lack a calling convention specifier then native calls are going to crash like in this case.

keveleigh:
@Holo-Krzysztof Interesting, thanks! They all work for .NET, but I'll go through and test them in IL2CPP too.

PR #1797, opened 2018-03-07
Updated ButtonTraditional prefab

Jarodshow:
The ButtonTraditional prefab was using an old 3DTextPrefab, and wasn't using the new standard text shader

## Before
actual-tradbutton

## After 
expected-tradbutton


cre8ivepark:
@Jarodshow Tested the scenes and it looks good. Thanks for the update!

Jarodshow:
For future reference, should small bug fixes be pushed into Stabilization?

StephenHodgson:
@Jarodshow yes

Issue #1801, opened 2018-03-07
Dead Link in Documentation (404 - page not found)


CarpeTempus:
## Overview
Hi, I was just reading the documentation and found a dead link:
Open https://github.com/Microsoft/MixedRealityToolkit-Unity/tree/master/Assets/HoloToolkit/UX in your Browser. Scroll down until you see step 4. with the text
> Follow the same steps for the sample text layout prefabs in HoloToolkit-Examples\Text

and click on the Link "HoloToolkit-Examples\Text".

## Expected Behavior
documentation should open

## Actual Behavior
Error 404 page not found

cre8ivepark:
@CarpeTempus Thanks for opening this issue!

PR #1802, opened 2018-03-07
UX Document Fixes

StephenHodgson:
Changes
---
- Fixes: #1801


keveleigh:
Is this meant to target Stabilization?

StephenHodgson:
@keveleigh yes. That way master doesn't get ahead of it.

keveleigh:
@StephenHodgson We should close Stabilization. I was under the impression that we'd create a stabilization branch off Dev, take bug fixes for some amount of time, merge the stabilization branch into master, delete the stabilization branch, then bring the bug fixes back into Dev to complete the cycle.

The release was merged, and Stabilization's only purpose is to stabilize towards a release. Otherwise, we'll get PRs targeting Stabilization incorrectly, and doc changes in master like this now turn into three PRs (branch -> Stabilization -> master -> dev) instead of "just" two.

Issue #1800, opened 2018-03-07
2017.2.1.2 Patch Release compile errors


Backer-Sultan:
## Overview
I'm getting two compile errors on an empty project with only the toolkit package imported.

## Expected Behavior
The toolkit compiles successfully.

## Actual Behavior
Getting two compile errors:
1. Assets/HoloToolkit/SharingWithUNET/Scripts/PlayerController.cs(341,13): error CS0246: The type of namespace name `UNetSharedHologram` could not be found. Are you missing an assembly reference?

2. Assets/Holotoolkit/SharingWithUNET/Scripts/PlayerController.cs(342,13): error CS0841: A local variable `ush` cannot be used before it is decleared

## Steps to reproduce
1. Start a new project.
2. Import package "HoloToolkit-Unity-2017.2.1.2". 
3. Click "Play".

## Unity Editor Version
2017.2.1.p2

## Mixed Reality Toolkit Release Version
2017.2.1.2

CarpeTempus:
Same here. I was curious and tried it with Unity 2017.3.1p2 as well but unfortunately with the same result..

mrmaxmagna:
got same error

askor:
@mrmaxmagna it's fixed in 2017.2.1.3 Hot Fix

PR #1803, opened 2018-03-07
Moved UNET Sharing code out into examples

StephenHodgson:
Overview
---
Sharing with UNET is really just an example anyway. Too many example assets in the core toolkit.

Changes
---
- Fixes: #1800

See [Draft Release](https://github.com/Microsoft/MixedRealityToolkit-Unity/releases/edit/untagged-797a1aa3074edafad192)


keveleigh:
I disagree that the entirety of SharingWithUNET is an example.

The only script I'd agree could be moved is `SharedAnchorDebugText`. `UNetSharedHologram` (the problem script) looks like a useful Toolkit script, syncing a holograms transform across the network (like several scripts in the Sharing folder do), and should be moved into the main Toolkit.

Move the assets if you want, but how is, for example, UIContainer any different from ManualIpConfigUtility in the Sharing folder? They're both assets designed to facilitate connecting to the service.

PR #1818, opened 2018-03-10
App Bar Build Fixes

StephenHodgson:
Overview
---
Fixed an issue with building for the standalone player.
Misc formatting.

### DIFF
--- a/Assets/MixedRealityToolkit/UX/Scripts/AppBar/AppBar.cs
+++ b/Assets/MixedRealityToolkit/UX/Scripts/AppBar/AppBar.cs
@@ -1,15 +1,14 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/UX/Scripts/AppBar/AppBar.cs
```diff
@@ -383,54 +398,51 @@ private void RefreshTemplates ()
...
                 }
             }
-            List<ButtonTemplate> defaultButtonsList = new List<ButtonTemplate>();
+
+            var defaultButtonsList = new List<ButtonTemplate>();
+
             // Create our default button templates based on user preferences
             if (UseRemove)
             {
-                defaultButtonsList.Add(GetDefaultButtonTemplateFromType(ButtonTypeEnum.Remove, numCustomButtons, UseHide, UseAdjust, UseRemove));
```
keveleigh: Why is `UseRemove` being removed in these lines?

StephenHodgson: Because the parameter wasn't being used in the method, and the method was private.

PR #1821, opened 2018-03-11
Enabled .NET 4.x for project

StephenHodgson:
Overview
---
Added .NET 4.x support and added project dialog box to prompt users to switch.

Removed the compiler definitions for the reflection extensions.


Assets/MixedRealityToolkit/Sharing/Scripts/SyncSettings.cs
```diff
@@ -100,7 +99,7 @@ public void Initialize()
             }
         }
 
-        private static Assembly[] GetAssemblies()
+        private static System.Reflection.Assembly[] GetAssemblies()
```
StephenHodgson: This fixes namespace issues by moving it inside the class instead of declaring it above.

Assets/MixedRealityToolkit/Utilities/Scripts/Editor/EnforceEditorSettings.cs
```diff
@@ -52,6 +53,22 @@ static EnforceEditorSettings()
                 }
             }
 
+            if (PlayerSettings.scriptingRuntimeVersion != ScriptingRuntimeVersion.Latest)
+            {
+                if (EditorUtility.DisplayDialog(
+                        "Change the Scripting Runtime Version to 4.x?",
+                        "The Mixed Reality Toolkit would like to change the Scripting Runtime Version to use .NET 4.x.\n\n" +
+                        "In order for the change to take place the Editor must be restarted.\n\n" +
+                        "WARNING: If you do not make this change, then it's likely you're build will not complete successfully.\n\n" +
```
davidkline-ms: you're -> your.

davidkline-ms: What makes builds "likely" vs "will" fail? I don't want developers to think of this as the kid who cried wolf.

StephenHodgson: Thanks! I'll be sure to clean up the language a bit.

ProjectSettings/ProjectSettings.asset
```diff
@@ -733,7 +732,7 @@ PlayerSettings:
     UNet: 0
     Unity_Ads: 0
   facebookSdkVersion: 7.9.1
-  apiCompatibilityLevel: 2
+  apiCompatibilityLevel: 3
```
keveleigh: What is compatibility level 3? I only have one possible value.
image

Assets/MixedRealityToolkit/Sharing/Scripts/SyncSettings.cs
keveleigh: If this fixes the namespace issue, do we still need to move to 4.6?

keveleigh: Also, if we do move to 4.6, this only *sorta* fixes the namespace issue. With this change, it's still using the Toolkit's old extension method instead of the new official one. Should we not move to the platform's extension method and delete the Toolkit's, since the old Toolkit one only existed to fill the gap of an official one?

ProjectSettings/ProjectSettings.asset
StephenHodgson: because that's the enum value

Assets/MixedRealityToolkit/Sharing/Scripts/SyncSettings.cs
StephenHodgson: That's a good point, I could remove the them

ProjectSettings/ProjectSettings.asset
keveleigh: Right, but enum value 2 (and 0, and 1, and probably any number since I only have one option) was .NET 4.6 for me too. I'm curious which version of Unity seemingly has more than one value here.

StephenHodgson: Unsure, but maybe the editor is doing something to ensure the compatibility level is set to this value because it's using the 4.6 scripting runtime version.

keveleigh: Yeah, must be

SimonDarksideJ:
Any update @StephenHodgson  on the fixes?

StephenHodgson:
I can't get this to build on a different machine. Gotta check this out first.

StephenHodgson:
Starting over on this

Issue #1825, opened 2018-03-12
AudioOccluder does not properly restore lowPass filter cutoff frequency


davidkline-ms:
## Overview
If a GameObject has a LowPass filter attached and configured, the AudioOccluder script will not propertly restore the original frequency.

## Expected Behavior
AudioOccluder (or the AudioEmitter) to restore the originally configured LowPass filter frequency.

## Actual Behavior
If a LowPass filter already exists on a component, the initial frequency is not cached and the cutoff will be set to 22kHz when the occlusion effect is removed.

## Mixed Reality Toolkit Release Version
All versions

davidkline-ms:
Fix for this issue is in the Patch4_Dev branch and will be part of the next Mixed Reality Toolkit release.

PR #1838, opened 2018-03-15
[master] Fix occlusion effect, add lofi and example scene

davidkline-ms:
Overview
---
This change fixes a bug in AudioOccluder.cs where the effect would restore the incorrect low pass filter frequency when the original sound source had a preexisting low pass filter attached,

Also renames AudioEmitter->AudioInfuencerManager to be more descriptive of the usage and purpose. Changes made here were to support the AudioOccluder fix.

IAudioInfluencer.cs was modified to simplify the method signatures.

AudioLoFiEffect was added as both a verification of the AudioOccluder fix and as a useful script for experiences wishing to simulate various sound sources (ex: AM radio).

Added Examples scene to demonstrate LoFi and validate the occluder fix.

All-up: improve the source comments / documentation of publics.

Update readme.md to match script names, dependencies and usage/behaviors.

Changes
---
- Fixes: #1825  .


Assets/HoloToolkit/SpatialSound/Scripts/AudioEmitter.cs
```diff
@@ -1,139 +0,0 @@
...
-using System.Collections.Generic;
-using UnityEngine;
-
-namespace HoloToolkit.Unity
-{
-    /// <summary>
-    /// Class which supports IAudioInfluencers being used with audio sources.
-    /// </summary>
-    [RequireComponent(typeof(AudioSource))]    
-    public class AudioEmitter : MonoBehaviour
```
SimonDarksideJ: Cannot just rename classes when the intention for the script has not changed.  Revert this change and mark the class as [Obsolete("AudioEmitter is being deprecated, please use AudioInfluencerManager instead")].  This causes a braking change which should be avoided for Master. (fine for dev_working)

What is the reason for the file rename?  Does it change the base reason for the script or is it just cosmetic / influential.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
```diff
@@ -9,47 +9,62 @@ namespace HoloToolkit.Unity
     /// <summary>
     /// Class that implements IAudioInfluencer to provide an occlusion effect.
     /// </summary>
-    public class AudioOccluder : MonoBehaviour, IAudioInfluencer
+    /// <remarks>
+    /// For best results, ensure that all sound emitting objects have an attached
+    /// AudioInfluencerManager. Failing to do so may result in incorrect frequency
+    /// and/or volume restoration when the RemoveEffect method is called.
+    /// </remarks>
+    public class AudioOccluder
```
SimonDarksideJ: Do not separate out dependencies on to separate lines as thi makes it harder to read.  The standard used in MRTK is to have all dependencies on the same line as the class.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
```diff
@@ -9,47 +9,62 @@ namespace HoloToolkit.Unity
...
+        /// the effect is applied.
+        /// The VolumePassThrough range is from 0.0 - 1.0 (0-100%), inclusive.
+        /// The default value is 1.0.
+        /// </remarks>
         [Tooltip("Percentage of the audio source volume that will be heard after applying occlusion.")]
         [Range(0.0f, 1.0f)]
         public float VolumePassThrough = 1.0f;
 
-        // Update is not used, but is kept so that this component can be enabled/disabled.
-        private void Update()
```
SimonDarksideJ: have you tested that this component can now be enabled / disabled as per the comment.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,222 @@
...
+            sourceQualityFilterSettings.Add(
+                AudioSourceQuality.FmRadio,
+                new FilterSettings(30, 15000));
+        }
+
+        /// <summary>
+        /// Source quality options that match common telephony and
+        /// radio broadcast options.
+        /// </summary>
+        public enum AudioSourceQuality
```
SimonDarksideJ: Please extract enum in to it's own file.  We should never embed enums unless they are private

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,222 @@
...
+            /// The frequency range used is a bit wider than that of human
+            /// hearing. It closely resembles the range used for audio CDs.
+            /// </remarks>
+            FullRange
+        }
+
+        /// <summary>
+        /// Settings for the filters used to simulate a low fidelity sound source.
+        /// </summary>
+        public struct FilterSettings
```
SimonDarksideJ: Please extract structs in to their own file, should never embed structures.

Assets/HoloToolkit-Examples/SpatialSound/Scenes/AudioLoFiOcclusionTest.unity
```diff
@@ -0,0 +1,686 @@
+%YAML 1.1
```
SimonDarksideJ: This demo scene needs updating to the new standards defined by @cre8ivepark 
Should have:
* Information dialog describing scene
* More information as to what it is demonstrating
* Should be more of a Show normal, LoFi and HiFi so as to better demonstrate what the scripts provide. (use gaze to activate /  deactivate each)

Assets/HoloToolkit/SpatialSound/Scripts/AudioEmitter.cs
StephenHodgson: On top of that, it's harder to see if there were any changes to this class.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
```diff
@@ -64,20 +79,33 @@ private void Update()
...
+            lowPass.cutoffFrequency = neutralFrequency;
             lowPass.enabled = false;
 
-            // Note: Volume attenuation is reset in the emitter.
+            // Note: Volume attenuation is reset in the AudioInfluencerManager, attached to the sound emitting object.
         }
+
+        // Update is not used, but is kept so that this component can be enabled/disabled.
+        private void Update()
+        { }
```
StephenHodgson: kinda odd formatting, it's probably safe to put this on a single line.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
StephenHodgson: @SimonDarksideJ it was moved to the end of the file for some reason.  We can probably revert this change.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
StephenHodgson: Agreed, there's not many anyway so no need to break them up.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
```diff
@@ -9,47 +9,62 @@ namespace HoloToolkit.Unity
...
+            AudioSource audioSource = soundEmittingObject.GetComponent<AudioSource>();
             if (audioSource == null)
             {
                 Debug.LogWarning("The specified emitter does not have an attached AudioSource component.");
                 return;
             }
 
             // Audio occlusion is performed using a low pass filter.                
-            AudioLowPassFilter lowPass = emitter.GetComponent<AudioLowPassFilter>();
+            AudioLowPassFilter lowPass = soundEmittingObject.GetComponent<AudioLowPassFilter>();
```
StephenHodgson: What was the purpose of renaming this? Just for cosmetic reasons?
Could we revert these just so it's a bit easier to focus on the real changes?

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
```diff
@@ -9,47 +9,62 @@ namespace HoloToolkit.Unity
...
-        /// <param name="audioSource">The AudioSource that will be impacted by the effect.</param>
-        public void ApplyEffect(GameObject emitter,
-                                AudioSource audioSource)
+        /// <param name="soundEmittingObject">The GameObject on which the effect is to be applied.</param>
+        public void ApplyEffect(GameObject soundEmittingObject)
         {
             if (!isActiveAndEnabled)
             { return; }
 
+            AudioSource audioSource = soundEmittingObject.GetComponent<AudioSource>();
```
StephenHodgson: Why change this method to exclude an audio source?

What if a dev already has an audio source cached and they wanted to use it instead of having to get the component here?

Let's just check if the provided audio source is null, then then get it if needed.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,222 @@
...
+            /// </summary>
+            public float HighPassCutoff
+            { get; private set; }
+
+            /// <summary>
+            /// FilterSettings constructor.
+            /// </summary>
+            /// <param name="highPassCutoff">High pass filter cutoff frequency.</param>
+            /// <param name="lowPassCutoff">Low pass filter cutoff frequency.</param>
+            public FilterSettings(
```
StephenHodgson: we could probably put all the parameters on a single line.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        private void Update()
+        {
+            FilterSettings fs = sourceQualityFilterSettings[SourceQuality];
+            if (fs != filterSettings)
+            {
+                // If we have an attached AudioInfluencerManager, we need to let it know
+                // about our filter settings change.
+                AudioInfluencerManager influencerManager = gameObject.GetComponent<AudioInfluencerManager>();
```
StephenHodgson: If this is a Manager, then shouldn't we be able to get it's singleton instance instead of getting the component?

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        // The audio filter settings that match the selected source quality.
+        private FilterSettings filterSettings;
+
+        // The filters used to simulate the source quality.
+        private AudioLowPassFilter lowPassFilter;
+        private AudioHighPassFilter highPassFilter;
+
+        // Collection used to look up the filter settings that match the selected
+        // source quality.
```
StephenHodgson: let's use the summary tags here

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,222 @@
...
+    [RequireComponent(typeof(AudioLowPassFilter))]
+    [RequireComponent(typeof(AudioHighPassFilter))]
+    [DisallowMultipleComponent]
+    public class AudioLoFiEffect : MonoBehaviour
+    {
+        /// <summary>
+        /// The quality level of the simulated audio source (ex: AM radio).
+        /// </summary>
+        [Tooltip("The quality level of the simulated audio source.")]
+        public AudioSourceQuality SourceQuality;
```
StephenHodgson: please use encapsulation with a serialized private field so you can set the quality in the editor.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
```diff
@@ -0,0 +1,242 @@
...
+            int count = Physics.RaycastNonAlloc(cameraTransform.position,
+                                                direction,
+                                                hits,
+                                                distance,
+                                                Physics.DefaultRaycastLayers,
+                                                QueryTriggerInteraction.Ignore);
+            
+            for (int i = 0; i < count; i++)
+            {
+                IAudioInfluencer ai = hits[i].collider.gameObject.GetComponentInParent<IAudioInfluencer>();
```
StephenHodgson: could we rename `ai` to `influencer` or something like that?

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
```diff
@@ -0,0 +1,242 @@
...
+                {
+                    // Apply the influencer's effect.
+                    influencer.ApplyEffect(gameObject);
+                }
+
+                // Find and remove the audio influencers that are to be removed from the audio source.
+                List<IAudioInfluencer> influencersToRemove = new List<IAudioInfluencer>();
+                foreach (IAudioInfluencer prev in previousInfluencers)
+                {
+                    MonoBehaviour mbPrev = prev as MonoBehaviour;
```
StephenHodgson: This seems odd. Let's replace `MonoBehaviour` with `GameObject` instead.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
```diff
@@ -0,0 +1,242 @@
...
+                    {
+                        influencersToRemove.Add(prev);
+                    }
+                }
+                RemoveInfluencers(influencersToRemove);
+
+                previousInfluencers = influencers;
+                lastUpdate = now;
+            }
+	    }
```
StephenHodgson: formatting seems off

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
```diff
@@ -0,0 +1,242 @@
...
+        // Potential effects manipulated by an audio influencer and their key
+        // properties.
+        private AudioLowPassFilter lowPassFilter;
+        private float nativeLowPassCutoffFrequency;
+        private AudioHighPassFilter highPassFilter;
+        private float nativeHighPassCutoffFrequency;
+
+        private void Awake() 
+        {           
+            audioSource = gameObject.GetComponent<AudioSource>();
```
StephenHodgson: let's check for null here in case the dev set it in the inspector

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
```diff
@@ -0,0 +1,242 @@
...
+    public class AudioInfluencerManager : MonoBehaviour
+    {
+        /// <summary>
+        /// Frequency below the nominal range of human hearing.
+        /// </summary>
+        /// <remarks>
+        /// This frequency can be used to set a high pass filter to allow all
+        /// audible frequencies through the filter.
+        /// </remarks>
+        public static readonly float NeutralLowFrequency = 10.0f;
```
StephenHodgson: could this be const instead?

StephenHodgson: if it's expected for devs to modify this value let's remove the readonly modifier.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
```diff
@@ -0,0 +1,242 @@
...
+        /// Time, in seconds, between audio influence updates.
+        /// </summary>
+        /// <remarks>
+        /// The UpdateInterval range is betweel 0.0 and 1.0, inclusive.
+        /// The default value is 0.25.
+        /// A value of 0.0f indicates that updates occur every frame.
+        /// </remarks>
+        [Tooltip("Time, in seconds, between audio influence updates.  0 indicates to update every frame.")]
+        [Range(0.0f, 1.0f)]
+        public float UpdateInterval = 0.25f;
```
StephenHodgson: let's use encapsulation with a private serialized backing field.
(same for all public fields below)

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
```diff
@@ -0,0 +1,242 @@
...
+        [Tooltip("Maximum number of objects that will be considered when looking for influencers.")]
+        [Range(1, 25)]
+        [SerializeField]
+        private int MaxObjects = 10;
+
+        // Time of last audio processing update.
+        private DateTime lastUpdate = DateTime.MinValue;
+
+        // The source of the audio.
+        private AudioSource audioSource;
```
StephenHodgson: let's serialize this so devs can set it in the inspector

Assets/HoloToolkit/SpatialSound/Scripts/AudioEmitter.cs
davidkline-ms: Putting the old name back is no problem. The reason for the change is that the old name wasn't very accurate in describing what the script does.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
davidkline-ms: The audio source parameter shouldnt have been there originally. It's not a huge deal to put it back and ignore it.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
davidkline-ms: This is a component level manager. It manages the influencers that are effecting the AudioSource attached to the game object. I can rename it (in vNext) to be a Controller if that seems better.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
davidkline-ms: Are we using summary tags for privates as well as publics? Will the auto documentation tool do the correct filtering?

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
davidkline-ms: I am able to set the quality in the editor. this also allows it to be changed via code.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: actually, I have a better way of doing this that I am coding up now.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: sure. it is a holdover from the previous version of the code

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
davidkline-ms: It got moved to get it out of the way. I tried eliminating it and was no longer able to enable/disable the component.

@SimonDarksideJ, yes, I have tested (and verified via Unity docs) that Update is one of the functions that allows for enabling/disabling components.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
davidkline-ms: I'm going to restore a set of overloaded methods and mark them as deprecated.  They will call the AudioSource-less versions.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: This is why I generally use static readonly instead of const: https://blogs.msdn.microsoft.com/davidklinems/2006/10/27/const-vs-static-readonly/

Devs are not expected to modify either of the neutral frequencies.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: the values can currently be set in the inspector as a public field. marking them private and serialized will prevent scripts from modifying these values at runtime.

the intent is to allow both editor and runtime changes.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: the values can currently be set in the inspector as a public field. marking them private and serialized will prevent scripts from modifying these values at runtime.

the intent is to allow both editor and runtime changes.

davidkline-ms: the AudioSource value is read from the one attached to the component. if we encapsulate the audio source, we prevent the devs from setting values such as 3d audio settings, etc. if we expose this to the inspector, we could be enabling the audio influences being applied to the wrong sound source.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
davidkline-ms: I'll extract the enum and the struct from the class. They are specific to working with this class, so I'll leave them in the source file, but make their names clearly imply that they are for use with the AudioLoFiEffect.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
davidkline-ms: I'll extract the enum and the struct from the class. They are specific to working with this class, so I'll leave them in the source file, but make their names clearly imply that they are for use with the AudioLoFiEffect.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
davidkline-ms: This was changed for clarity.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: this comes from querying the game object. this class, by design, does not expose the AudioSource to the inspector. this was true of the earlier version as well.

Assets/HoloToolkit/SpatialSound/Scripts/AudioEmitter.cs
SimonDarksideJ: Which is fine, but class name changes have to be controlled else they just break projects.  Given you are changing the logic and function of this class, it might be better to mark the old as obsolete (so you don't inadvertently alter existing projects using it) and create your new class to supercede it in a future release.

You would need to do A/B testing to ensure your updates don't break existing implementations.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
SimonDarksideJ: So from your testing, this update call has to stay in or consumers can't enable / disable it on the fly

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
SimonDarksideJ: It's not just about editor access @davidkline-ms , it's more of a stanards thing with Unity.  The recommended standard to use with such fields is to have a Private Field that is serialised (which makes it visible in the editor), then only add the public variant if you intend for the field to be accessed outside of the class.

Assets/HoloToolkit/SpatialSound/Scripts/AudioOccluder.cs
davidkline-ms: Yes. Per the documentation for MonoBehaviour, one or more of the following must be present to be able to enable/disable on the fly.

Note: There is a checkbox for disabling MonoBehaviour on the Unity Editor. It disables functions when unticked. If none of these functions are present in the script, the Editor does not display the checkbox. The functions are:
Start()
Update()
FixedUpdate()
LateUpdate()
OnGUI()
OnDisable()
OnEnable()

Update is convenient, but then again, so is Start().  I am willing to have an empty method for either of those two.

Assets/HoloToolkit/SpatialSound/Scripts/AudioEmitter.cs
davidkline-ms: The rename was 100% about being more accurate about what the class does. The core functionality doesn't change. some logic definitely does (ex: provide a means to fix #1825, and to actually handled disabled influencers).

I am fine using the old name and logging an issue to rename in vNext.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: sure. I can do that.

davidkline-ms: I'll take a look

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: what part seems off? either my editor is set different from yours or I am just not seeing it :)

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
davidkline-ms: @SimonDarksideJ, are you advocating having BOTH a private and a public field? or are you advocating for a private field and a public set of Get/Set methods? or are you recommending a differently named public property?

For something that was originally intended as being public, it seems that a single public field makes sense.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
StephenHodgson: Right, and if we use a public property to access them, then we still have both, as intended.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
StephenHodgson: spaces could be off.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
StephenHodgson: >  are you advocating having BOTH a private and a public field?

No private field with public property. We're trying to ensure encapsulation.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
StephenHodgson: I guess that depends on the tool we use.

Assets/HoloToolkit/SpatialSound/Scripts/AudioInfluencerManager.cs
davidkline-ms: The compiler complains about casting to GameObject via the as keyword (used to avoid InvalidCastException). I will see if there is a better solution than casting to a MonoBehaviour.

Assets/HoloToolkit-Examples/SpatialSound/Scenes/AudioLoFiOcclusionTest.unity
davidkline-ms: in progress

davidkline-ms: Example scene updated and renamed. Removed the occlude as that is not what is being demonstrated.

Scene showcases three of the five sound quality settings.

davidkline-ms:
PR feedback addressed. Please re-review this change.

Thanks!

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
StephenHodgson: In either case I'd like to at least inspect it and see a description via tooltip

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,238 @@
...
+            sourceQualityFilterSettings.Add(
+                AudioLoFiSourceQuality.FmRadio,
+                new AudioLoFiFilterSettings(30, 15000));
+        }
+    }
+
+    /// <summary>
+    /// Settings for the filters used to simulate a low fidelity sound source.
+    /// </summary>
+    public struct AudioLoFiFilterSettings
```
keveleigh: Could probably be marked private, since it isn't used outside this class.

Assets/HoloToolkit/SpatialSound/Scripts/AudioLoFiEffect.cs
davidkline-ms: Summary tags were added throughout.

StephenHodgson: Thanks!

davidkline-ms:
@SimonDarksideJ. @StephenHodgson requested changes addressed. Can you plz revisit? Thx!

davidkline-ms:
@StephenHodgson. yes they will be submitted to dev this week. I just need to fix the namespace and rename AudioEmitter first

PR #1846, opened 2018-03-19
MRTK vNEXT coding guidelines

StephenHodgson:
MRTK vNEXT coding guidelines


CodingGuidelines.md
```diff
@@ -0,0 +1,168 @@
...
+
+private void Update()
+{
+    myRenderer.Foo(Bar);
+}
+ ```
+
+---
+
+Public classes, structs, and enums should all go in their own files.
```
timGerken: I'd love a link to a citation as to why this is a best practice.

CodingGuidelines.md
```diff
@@ -0,0 +1,168 @@
...
+
+This document outlines the recommended coding guidelines for the Mixed Reality Toolkit.  The majority of these suggestions follow the [recommended standards from MSDN](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions).
+
+## Spaces vs Tabs
+Please be sure to use 4 spaces when contributing to this project.
+
+## Naming Conventions
+
+Always use `PascalCase` for public properties, and `camelCase` for private properties and fields.
+> The only exception to this is for data structures that require the fields to be serialized by the `JsonUtility`.
```
timGerken: an example of the exception would be useful

CodingGuidelines.md
```diff
@@ -0,0 +1,168 @@
...
+    }
+}
+```
+
+## Encapsulation
+Always use private fields and public properties if access to the field is needed from outside the class or struct.
+
+If you need to have the ability to edit your field in the inspector, it's best practice to follow the rules for Encapsulation and serialize your backing field.
+
+>The only exception to this is for data structures that require the fields to be serialized by the `JsonUtility`.
```
timGerken: An example of the exception case would be useful here as well.

timGerken:
I'm so glad to see this. It would be great if something similar were in master as well...

SimonDarksideJ:
That's the plan @timGerken once this has been agreed.  We regually hit these standards in PR reviews which wastes a lot of time.  Will be better if contributors are aware of our expectations before submitting.

CodingGuidelines.md
SimonDarksideJ: It's for a couple of reasons, a simple case is so that enum / struct changes don't affect the classes they are used by.
Organisationally, the intention is that PUBLIC enums / structs will be used by many classes, so should be separate.  Only if it is a private enum for a specific class should it be included in the class.

The other reason is because Unity! (the default answer for most cases)

CodingGuidelines.md
SimonDarksideJ: Agreed, I'm going to do a PR on @StephenHodgson branch to add my thoughts (esp as the next section actually breaks this principle :P)

CodingGuidelines.md
SimonDarksideJ: An example is provided in this section Do's / Dont's

CodingGuidelines.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Do:
+
+ ```
+private void Foo(){ }
+public void Bar(){ }
+ ```
+
+## Use Braces
+Always use braces whenever possible.
```
davidkline-ms: I would remove "whenever possible" :)

CodingGuidelines.md
davidkline-ms: @SimonDarksideJ, I don't see an example of an exception.

CodingGuidelines.md
```diff
@@ -0,0 +1,168 @@
...
+     set{ myValue = value }
+ }
+ ```
+
+## Best Practices
+Some of the target platforms of this project require us to take performance into consideration.  What this in mind we should always be careful of allocating memory in frequently called code in tight update loops or algorithms.
+
+---
+
+Use `for` instead of `foreach` when possible.
```
davidkline-ms: When would it not be possible to use a for? all enumerable collections support a Size/Count property, to the best of my recollection.
 If there are situations where a foreach is preferred (or acceptable) we should document an exception.

CodingGuidelines.md
davidkline-ms: I do see times where it is best to keep a struct/enum in the same source file as the class for which it is associated. For example, a class has a custom settings structure that is specific to that class. It feels appropriate to have the struct co-located in the class' source file for easy visual reference.

CodingGuidelines.md
```diff
@@ -0,0 +1,168 @@
...
+### Do:
+
+ ```
+[SerializeField] // You could set the reference in the inspector.
+private Renderer myRenderer;
+
+private void Awake()
+{
+    // If you didn't set it in the inspector, then we cache it on awake.
+    if(myRenderer == null)
```
davidkline-ms: Let's add a rule that places spaces between flow control statements, ex: if, for, while.

I was convinced awhile ago that "if (..)" is easier to read than "if(…)" and VS automatically formats that way :)

CodingGuidelines.md
```diff
@@ -0,0 +1,168 @@
+# Coding Guidelines
+
+This document outlines the recommended coding guidelines for the Mixed Reality Toolkit.  The majority of these suggestions follow the [recommended standards from MSDN](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions).
+
+## Spaces vs Tabs
+Please be sure to use 4 spaces when contributing to this project.
+
+## Naming Conventions
+
+Always use `PascalCase` for public properties, and `camelCase` for private properties and fields.
```
keveleigh: What about [protected fields](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1659#discussion_r163358898)?

StephenHodgson: PascalCase

SimonDarksideJ:
I've addressed the above comments in a PR to @StephenHodgson branch, as I was adding some of my own bits to it.

CodingGuidelines.md
```diff
@@ -61,12 +191,76 @@ private Foo()
...
+public class MyClass
+{
+    private MyStruct myStructreference;
+    private MyEnumType myEnumReference;
+}
+ ```
+---
+
+## Best Practices, including Unity recommendations
+Some of the target platforms of this project require us to take performance into consideration.  What this in mind we should always be careful of allocating memory in frequently called code in tight update loops or algorithms.
```
davidkline-ms: What -> With

CodingGuidelines.md
```diff
@@ -61,12 +191,76 @@ private Foo()
...
+
 ## Encapsulation
+---
+
 Always use private fields and public properties if access to the field is needed from outside the class or struct.
 
 If you need to have the ability to edit your field in the inspector, it's best practice to follow the rules for Encapsulation and serialize your backing field.
 
->The only exception to this is for data structures that require the fields to be serialized by the `JsonUtility`.
+>The only exception to this is for data structures that require the fields to be serialized by the `JsonUtility`, where a data class is required to have all public fields for the serialization to work.
```
davidkline-ms: Can you add a comment to co-locate the private field and the public property. This makes it easier to see, at a glance, what backs the property and that the field is modifiable by script.

Thx!

CodingGuidelines.md
```diff
@@ -141,28 +350,42 @@ private void Update()
 
 ---
 
-Public classes, structs, and enums should all go in their own files.
+Cache references to materials, do not call the ".material" each time.
```
davidkline-ms: Add a mention to using .sharedMaterial as an option for setting materials on objects?

StephenHodgson:
Here's a great link showing current property/field performance differences.

https://jacksondunstan.com/articles/2968
@davidkline-ms

davidkline-ms:
Excellent writeup! I will no longer fret over the perf difference between properties and fields :)

CodingGuidelines.md
davidkline-ms: @StephenHodgson, I think the What -> With fix is close to the last thing we need changed.

PR #1848, opened 2018-03-20
TapToPlace: extends the usage of collider to place objects

joaocpaiva:
Introduces the ability to use new reference points such as top or bottom
of the collider, helpful for common scenarios where user wants to place
the object on the floor or ceiling.

I used the TapToPlaceScene to test all combinations (None, Center, Bottom, Top) to produce the desired behavior.


joaocpaiva:
This is functionality I needed in an app and I believe it can help other developers.

PR #1854, opened 2018-03-21
TapToPlace: extends the usage of collider to place objects

joaocpaiva:
Introduces the ability to use other parts of the collider like top or bottom, useful for common scenarios where user wants to place the object on the floor or ceiling.

Used the TapToPlace scene to test combinations (DontUse, UseCenter, UseBottom and UseTop) to validate the desired behavior.

### DIFF
--- a/Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
+++ b/Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
@@ -23,17 +23,6 @@ namespace MixedRealityToolkit.SpatialMapping
     [RequireComponent(typeof(Interpolator))]
     public class TapToPlace : MonoBehaviour, IInputClickHandler
     {
-        /// <summary>
-        /// The collider reference point used to place the object.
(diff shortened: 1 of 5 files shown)


Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
```diff
@@ -23,6 +23,17 @@ namespace MixedRealityToolkit.SpatialMapping
     [RequireComponent(typeof(Interpolator))]
     public class TapToPlace : MonoBehaviour, IInputClickHandler
     {
+        /// <summary>
+        /// The collider reference point used to place the object.
+        /// </summary>
+        public enum ColliderPlacement
```
davidkline-ms: Per code style guidelines (see #1846), please move public enums outside of the class. I also recommend naming it TapToPlaceColliderPlacement to make it clear which class it is for (unless you believe it has a more generic usage.

davidkline-ms:
@davidkline-ms note to self to target this fix for Patch4

PR #1855, opened 2018-03-21
[master] fixes #1831 - replace update with lateupdate in FixedAngularSize

### LINKED ISSUE #1831: FixedAngularSize
## Overview
When using the FixedAngularSize.cs on the cursor, i observed a flicker when it was going from an intersection of an object that was far to an empty area (whereby placing the cursor at the default distance).
I found that the scaling of the cursor was done in the FixedAngularSize.cs Update(), but its positioning from the GazeManager.Update(). So if the FixedAngularSize update is called first and then the Gazemanager.Update is done afterwards you would get the wrong scale for one frame.

I fixed this by changeing Update to LateUpdate in FixedAngularSize.cs to make sure that the scaling of the cursor is computed after its placement.

## Expected Behavior
No flicker
## Actual Behavior
Flickers when going form far to near.

## Steps to reproduce
Add component to a cursor. SO that the cursor stays always the same size, irrespective of near or far.
Then try to look at a few objects you have setup near and far from the camera. When the cursor leaves the focus of an object i observed a slight jump (imminent resize) of the cursor. The explanation is above.
its a really simple fix....

## Unity Editor Version
2017.3.1.p2
2017.2.1.1

davidkline-ms:
Overview
FixedAngularSize was previously using Update to adjust the element's visible size. This caused a flicker when applied to a cursor and this update came before the update performed by the GazeManager (as reported in #1831).

Changes
---
* Replaced Update with LateUpdate to ensure the scaling occurs after other modifications.
* Encapsulated the SizeRatio field per the pending code style guide and applied a range check in the setter.
* Add a range to SizeRatio (was defined as 0 - 1 in SetSizeRatio())
* Add summary tag comments throughout.

- Fixes: #1831 .

### DIFF
--- a/Assets/HoloToolkit/Utilities/Scripts/FixedAngularSize.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/FixedAngularSize.cs
@@ -12,13 +12,43 @@ namespace HoloToolkit.Unity
     /// </summary>
     public class FixedAngularSize : MonoBehaviour
     {
-        [Tooltip("Off sets the scale ratio so that text does not scale down too much. (Set to zero for linear scaling)")]
-        public float SizeRatio = 0;
+        /// <summary>
+        /// Offsets the scale ratio so that text does not scale down too much. (Set to zero for linear scaling)
+        /// </summary>
+        [Tooltip("Offsets the scale ratio so that text does not scale down too much. (Set to zero for linear scaling)")]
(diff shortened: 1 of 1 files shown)


StephenHodgson:
Should we be fixing this in dev as well?

davidkline-ms:
@StephenHodgson, yes. I plan on fixing this and other updates taken to master in dev this week.

Issue #1221, opened 2017-10-22
Events for TapToPlace component


stevenjt:
Hi, it would be useful to have starting/ending placement events for this component to make it easier to integrate with other components. I have implemented two events OnStartedPlacing and OnStoppedPlacing as a PR.

wiwei:
Closing issues older than 180 days. If this is still an issue, please reactivate with recent information.

PR #1222, opened 2017-10-22
Events added to TapToPlace

stevenjt:
OnStartedPlacing and OnStoppedPlacing events added to TapToPlace

PR for #1221

PR #1865, opened 2018-03-23
TapToPlace: add events for placement of objects

joaocpaiva:
Uses interface to route events for every object that is placed. It fires one event when placement starts and another once placement is completed.

Used TapToPlace test scene to validate the events are triggered as expected.

This change comes as result of functionality I required for my scenario and I am sure it can be useful for other developers (as I have seen mentioned in other threads such as [PR 1222](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1222))

Changes
- Fixes: #1221


Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
```diff
@@ -153,6 +174,12 @@ public virtual void OnInputClicked(InputClickedEventData eventData)
             eventData.Use();
         }
 
+        private void RaisePlacingEvent(ExecuteEvents.EventFunction<ITapToPlaceHandler> handler, GameObject selectedObject)
```
StephenHodgson: This method and it's delegate counterpart should me moved into the input manager class.

joaocpaiva: @StephenHodgson in fact that was my initial approach but then I noticed that all current events implement/need the corresponding IInputSource and I was not sure what to do for the case of the new IPlacementHandler. Any thoughts on that?

StephenHodgson: No need to add a new source.  The input source is just as it says, the root source of the input, which in this case would be from the Tap event (passing in the event data's input source should be enough to satisfy this condition.) Thoughts?

joaocpaiva: Check out the new iteration and tell me what you think.

Assets/MixedRealityToolkit/InputModule/Scripts/InputManager.cs
```diff
@@ -245,8 +247,20 @@ private void Start()
...
-            // Use focused object when OverrideFocusedObject is null.
-            GameObject focusedObject = (OverrideFocusedObject == null) ? FocusManager.Instance.TryGetFocusedObject(eventData) : OverrideFocusedObject;
+            
+            GameObject focusedObject;
+
+            if (OverrideFocusedObject != null)
+            {
+                // Use override focused object if not null.
+                focusedObject = OverrideFocusedObject;
+                eventData.selectedObject = OverrideFocusedObject;
```
StephenHodgson: we don't generally use/set `eventData.selectedObject` unless it's for uGUI related things. generally this is null.

We should revert these lines.

Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
```diff
@@ -55,7 +57,7 @@ public class TapToPlace : MonoBehaviour, IInputClickHandler
 
         private Dictionary<GameObject, int> layerCache = new Dictionary<GameObject, int>();
         private Vector3 PlacementPosOffset;
-
+
```
StephenHodgson: got some extra spaces on this line

Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
```diff
@@ -182,6 +186,22 @@ private void StopPlacing()
...
+            RaisePlacingEvent(InputManager.Instance.RaisePlacingCompleted);
+        }
+
+        private void RaisePlacingEvent(Action<IInputSource, uint, object[]> raisePlacingEvent)
+        {
+            // Set override focused object before trigger the event, otherwise 
+            // the selectedObject defined in the event data might be wrong because,
+            // the object being placed may not have the focus by now (which occurs
+            // for placing completed for example).
```
StephenHodgson: why not just use InputManager.Push/Pop modal input?

Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
```diff
@@ -21,7 +23,7 @@ namespace MixedRealityToolkit.SpatialMapping
     /// </summary>
     [RequireComponent(typeof(Collider))]
     [RequireComponent(typeof(Interpolator))]
-    public class TapToPlace : MonoBehaviour, IInputClickHandler
+    public class TapToPlace : MonoBehaviour, IInputClickHandler, IInputSource
```
StephenHodgson: We probably don't need to make this an input source, but just pass the click input source when we raise the event.

Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
```diff
@@ -149,35 +151,44 @@ public virtual void OnInputClicked(InputClickedEventData eventData)
...
             ToggleSpatialMesh();
             RemoveWorldAnchor();
         }
 
-        private void StopPlacing()
+        private void StopPlacing(IInputSource inputSource)
         {
             var layerCacheTarget = PlaceParentOnTap ? ParentGameObjectToPlace : gameObject;
             layerCacheTarget.ApplyLayerCacheRecursively(layerCache);
+            InputManager.Instance.RaisePlacingCompleted(inputSource, 0);
```
joaocpaiva: @StephenHodgson I think this iteration addresses your concerns? 

My only concern with this version is, the PlacementEventData.selectedObject for OnPlacingCompleted is not the object that is being placed. For example, in the TapToPlace test scene the selectedObject is the 'roommesh' instead of the 'cube' that is being placed. That does not seem the desired behavior of these events?

StephenHodgson: Yes, this is what I was.

I'll have to take a look at the class you're taking about, but the selectedObject is something that should be handled by the focus manager.

If you pop/pash the modal Input handler, you should be able to specify an object that will temp get all input directly.

joaocpaiva: I believe the reason is:

In InputManager, even though the modalInput is 'cube' the the focusedObject is 'roomMesh' and the eventData.selectedObject is always set as the focusedObject.

StephenHodgson: eventData.selectedObject should always be the focusedObject.

joaocpaiva: So does it mean the behavior is by design and we should proceed without any modifications?

Summarizing the issue:
1. the selectedObject when OnPlacingStarted triggers is 'cube' as I would expect.
2. the selectedObject when OnPlacingCompleted triggers is 'roommesh' instead of the 'cube' that is being placed.

The events are routed to the listeners registered as expected. It's just the selectedObject when OnPlacingCompleted triggers that makes me wonder whether it's a bug or just by design. Ideally I would like to know what game object is being placed when those events fire ... but maybe I am overthinking the API.

StephenHodgson: >Ideally I would like to know what game object is being placed when those events fire

You could add a GameObject property to the eventData to get this info.

joaocpaiva: Oh that sounds good to me! I wasn't sure passing the GameObject around would be well received since there was no other event doing it,

Assets/MixedRealityToolkit/SpatialMapping/Scripts/TapToPlace.cs
```diff
@@ -275,89 +274,5 @@ private static Vector3 GetGazePlacementPosition(Vector3 headPosition, Vector3 ga
             }
             return headPosition + gazeDirection * defaultGazeDistance;
         }
-
-        #region IInputSource Implementation
```
joaocpaiva: @StephenHodgson I just noticed I had forgot to remove this unused code so I pushed one last commit.

Assets/MixedRealityToolkit-Examples/SpatialMapping/Scripts/TapToPlaceTest.cs
```diff
@@ -0,0 +1,27 @@
...
+using MixedRealityToolkit.InputModule.EventData;
+using MixedRealityToolkit.InputModule.InputHandlers;
+using UnityEngine;
+
+namespace MixedRealityToolkit.Examples.SpatialMapping
+{
+    /// <summary>
+    /// Class to test TapToPlace interactions, such as placement of object events.
+    /// </summary>
+    public class TapToPlaceTest : MonoBehaviour, IPlacementHandler
```
StephenHodgson: This test can go into the InputTest.cs in the examples folder.

Issue #1868, opened 2018-03-23
ObjectCollection does not remove moved GameObject that is no longer a child from NodeList


MSDNAndi:
## Overview
UpdateCollection() in ObjectCollection does not remove moved GameObject that is no longer a child from NodeList. It only removes GameObjects from NodeList that do not exist anymore, resulting in a wrong number of children and wrong layout
## Expected Behavior
When a GameObject is not a child, it should not be in the NodeList holding the children that should be laid out. 
## Actual Behavior
See above.
## Steps to reproduce
Add a GameObject (e.g. the cube primitive) as a child to a parent GameObject that has ObjectCollection, then call or select UpdateCollection, then move the GameObject elsewhere so it is no longer a child of the GameObject, select or call UpdateCollection on the parent GameObject. NodeList still contains the moved GameObject, resulting in the layout considering too many children.

MSDNAndi:
While I believe I contributed a fix, for the issue described, I still find funky behavior when it comes to hierarchical object collections and their members, where I am not sure about the reason for it yet. I am not sure yet how to describe and repro the observed behavior I am still seeing.

MSDNAndi:
I think I need to rework the fix. I should not have used IsChildOf as it uses deep children.

if (NodeList[i].transform == null || (IgnoreInactiveTransforms && !NodeList[i].transform.gameObject.activeSelf) || NodeList[i].transform.parent==null || ! (NodeList[i].transform.parent.gameObject==this.gameObject) )

Seems to be the better fix. I am not sure if I should open another issue and work against that new issue, if I should submit against this issue, or if I should just edit it again.

davidkline-ms:
Please open a new issue and pull request.

Thanks!

StephenHodgson:
Please keep issues open until they're merged with the master branch.

MSDNAndi:
Sorry, didn't realize. Will they auto-close then?
________________________________

davidkline-ms:
@MSDNAndi, Yes. We will close the issues when the release complete.

PR #1869, opened 2018-03-23
[master] Update ObjectCollection.cs

MSDNAndi:
Fixes issue with GameObject staying in NodeList even though it is not a child anymore.

Overview
---
Remove GameObject from NodeList, when it is not a child of this.transform anymore
!NodeList[i].transform.IsChildOf(this.transform))

Changes
---
- Fixes: #1868


davidkline-ms:
Looks like a clean fix. @cre8ivepark should take a look also.

davidkline-ms:
Retargeted to the patch4_dev branch.

Issue #1860, opened 2018-03-22
[master -> dev] integrate audio occlusion fixes and add the lofi effect


davidkline-ms:
Master recently took some bug fixes to audio occlusion (#1838) and a new LoFi effect. bring these changes into dev with the following modifications:
* namespace updates
* folder location updates
* rename AudioEmitter.cs to AudioInfluencerController.cs to better match the intent of the class

PR #1873, opened 2018-03-23
[master->dev] integrate recent audio occlusion fixes

davidkline-ms:
Overview
---
* Updates AudioInfluencerController (formerly AudioEmitter), IAudioInfluencer and AudioOcclusion scripts to match recent changes in the master (Patch4_Dev) branch.
* Add AudioLoFiEffect script to provide a simple low fidelity effect (also a great test of the occlusion fix!)
* Move scripts into new MixedRealityToolkit.SpatialSound.Effects namespace
* Add -Examples\SpatialSound\Scenes\AudioLoFi and related files
* Update SpatialSound\README.md

Changes
---
- Fixes: #1860.


Assets/MixedRealityToolkit-Examples/SpatialSound/Scripts/LoFiFilterSelection.cs
```diff
@@ -0,0 +1,104 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using MixedRealityToolkit.InputModule.EventData;
+using MixedRealityToolkit.InputModule.InputHandlers;
+using MixedRealityToolkit.SpatialSound.Effects;
+
```
StephenHodgson: missing namespace

Assets/MixedRealityToolkit-Examples/SpatialSound/Scripts/LoFiFilterSelection.cs
```diff
@@ -0,0 +1,104 @@
...
+                emitterMaterial = AmRadio;
+                break;
+
+            case AudioLoFiSourceQuality.FullRange:
+                emitterMaterial = FullRange;
+                break;
+        }
+
+        // Set the material on the emitter.
+        gameObject.GetComponent<Renderer>().sharedMaterial = emitterMaterial;
```
StephenHodgson: good practice to cache the renderer reference in start or awake so we don't have to get the component each time we set the emitter material.

Assets/MixedRealityToolkit/SpatialSound/Scripts/Effects/AudioInfluencerController.cs
```diff
@@ -0,0 +1,297 @@
...
+                {
+                    // Apply the influencer's effect.
+                    influencer.ApplyEffect(gameObject);
+                }
+
+                // Find and remove the audio influencers that are to be removed from the audio source.
+                List<IAudioInfluencer> influencersToRemove = new List<IAudioInfluencer>();
+                foreach (IAudioInfluencer prev in previousInfluencers)
+                {
+                    MonoBehaviour mbPrev = prev as MonoBehaviour;
```
StephenHodgson: still not sure if this is correct.  What about classes that implement the IAudioInfluencer that don't inherit from MonoBehaviour?

davidkline-ms: then prev will be null and the following would handle it appropriately.
                        ((mbPrev != null) && !mbPrev.isActiveAndEnabled))
If the IAudioInfluencer does not inherit from MonoBehaviour, it is likely not able to be disabled anyhow.

Assets/MixedRealityToolkit-Examples/SpatialSound/Scripts/LoFiFilterSelection.cs
davidkline-ms: fair enough :)

Assets/MixedRealityToolkit/SpatialSound/Scripts/Effects/AudioInfluencerController.cs
StephenHodgson: That's what I mean, what if a dev decided to make their own IAudioInfluencer, and didn't want to have their class inherit from Monobehavior?

davidkline-ms: If they make their own IAudioInfluencer that does not inherit from MonoBehavior, they will not be able to attach it to their game objects, which would make the system not function at all for them. It is not designed for use without the influencer being a MonoBehavior.

StephenHodgson: Fair enough

StephenHodgson:
Why did we merge this in without resolving the missing namespace?

davidkline-ms:
The namespace was added:
+namespace MixedRealityToolkit.Examples.SpatialSound 
+    public class LoFiFilterSelection : MonoBehaviour, IInputClickHandler

Issue #1440, opened 2017-11-26
[master] Haptics doesn't work when running in-editor


keveleigh:
## Overview
The UWP haptics APIs are currently exposed in a way that makes them incompatible with running in the Unity editor.

My recommended fix would be by adding to the DLL (we can rename it to make it more accurate) I wrote to expose the motion controller model API, [found in the MixedRealityTookit here](https://github.com/Microsoft/MixedRealityToolkit/blob/master/UnityEditorMotionControllerModelDLL/MotionControllerModel/MotionControllerModel.cpp).

Additional methods could be written, with identical code to lines 18-39, but then calling the haptics APIs like in the [existing Toolkit extensions](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit/Utilities/Scripts/Extensions/InteractionSourceExtensions.cs#L53).

I initially experimented with this addition when writing the initial DLL, but remember being blocked around creating a `TimeSpan` object, without time at the time to fully investigate.

## Expected Behavior
The haptics APIs work when running in the editor.

## Actual Behavior
The haptics APIs only work in a deployed app.

## Steps to reproduce
Run the MotionControllerTest scene and press grasp or menu to try to trigger haptics.

## Unity Editor Version
All

## Mixed Reality Toolkit Release Version
All

StephenHodgson:
Any way we can extend this to also tie in haptics from non spatial controllers as well?

keveleigh:
Yeah, maybe we can use the [UWP GamePad Vibration APIs](https://docs.microsoft.com/en-us/windows/uwp/gaming/gamepad-and-vibration#vibration-and-impulse-triggers-overview) or the [SimpleHapticsControllers from the UWP RawGameController APIs](https://docs.microsoft.com/en-us/uwp/api/windows.gaming.input.rawgamecontroller) to expose haptics there as well.

PR #1881, opened 2018-03-26
[Dev_Working] Adding motion controller haptics to editor; updating and renaming DLL

keveleigh:
Overview
---
This update adds haptics support while running in the Editor. It also renames the DLL to better reflect the newly multi-feature support, instead of just models.

To test, run the MotionControllerTest scene and press either the grip button or the menu button on your motion controller.

The corresponding DLL updates can be found in Microsoft/MixedRealityToolkit#146 and Microsoft/MixedRealityToolkit#153.

This is the dev branch version of #1880.

Changes
---
- Fixes: #1440.


jimmy12day:
The scripts is changed in latest version as a extension. How to access the Starthaptics method now in unity?

keveleigh:
The same code is present in master, included in https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/1880, and nothing has really changed since then.

The [HapticsTest](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit-Examples/Input/Scripts/HapticsTest.cs) script included in the examples should show you how to use `StartHaptics`.

Issue #1874, opened 2018-03-23
[master] update unity version to 2017.2.1p2


davidkline-ms:
Update the toolkit's unity version to match the recommendation on https://docs.microsoft.com/en-us/windows/mixed-reality/install-the-tools

StephenHodgson:
We agreed that we would only move the project to full releases of Unity, and not patches.

Although, I do like the idea that we should keep in line with the dev docs.

davidkline-ms:
Alignment with the dev docs is 100% the goal. Customers that install recommended tools should not have to see the "the versions don't match" dialog. :)

keveleigh:
Related to #1628.

> not patches

We shouldn't arbitrarily move to the latest patch, but Unity has put many MR-specific fixes into patches, often making them the recommended version.

Our current version (2017.2.1f1) was never an officially recommended version for MR development, due to bug fixes that never made it in out of the MRTP builds (motion controllers not being detected when playing from the editor, among others). These bug fixes *did* make p2.

davidkline-ms:
Definitely not advocating to updating for every patch that is released. Only to match the recommended guidance on the Windows Mixed Reality documentation.

PR #1882, opened 2018-03-26
[Patch4] Update projectversion to match the current recommended Unity release

davidkline-ms:
We are recommending 2017.2.1p, yet every user with that build sees the "do you want to update" dialog. This doesn't feel right.

Fixes: #1874


davidkline-ms:
@StephenHodgson, by MRTKVersion file are you meaning the Roadamp.md?

davidkline-ms:
Nevermind…. I was looking in the wrong place.

PR #1892, opened 2018-03-29
Update CustomInputSource.cs

MrDChristop:
Enable Navigation Gesture emulation using the mouse. Until know only the Manipulation Gesture would work.

Overview
---
Enables the navigation Gesture. 
I measured and compared its behavior to the Hololens emulator available in Unity when an xbox controller is connected.

Changes
---
Only adds the navigation gestuure functionality


StephenHodgson:
@MrDChristop this made it into the dev branch already, correct?

PR #1894, opened 2018-03-29
Automate building of MRTK artifacts

jwittner:
Adds scripts and package definitions for automating creation of artifacts for the project.

Creates three NuGet Packages:
* Microsoft.MixedReality.Toolkit
* Microsoft.MixedReality.Toolkit.Sharing
* Microsoft.MixedReality.Toolkit.Examples

Creates two Unity Packages:
* Microsoft.MixedReality.Toolkit
* Microsoft.MixedReality.Toolkit.Examples

Note: Sharing is split out because it has large dependencies on things outside of the repo. Over time we can split out more of the library since dependencies in NuGet are easy and work well. I know this is something other customers have wanted. We could also depend on the NewtonSoft.Json NuGet package instead of embedding the DLL into the project, for example. If you're interested in NuGet for Unity, checkout [NuGetForUnity](https://github.com/GlitchEnzo/NuGetForUnity/releases).

Open, of course, for suggestions and questions.


StephenHodgson:
Should we update the internal namespaces to match? or vice versa?

jwittner:
Perhaps the naming was a bit hopeful, but it's in line with Microsoft standards and other OSS projects.

I bet we can get a glTF NuGet package up that properly takes a dependency on NewtonSoft.Json. =)

jwittner:
Boom - https://www.nuget.org/packages/glTF2Loader/

StephenHodgson:
Maybe we should align MRTK vNEXT namespace to `Microsoft.MixedReality.<FeatureName>`

@SimonDarksideJ

StephenHodgson:
NuGet is just an optional thing. No forcing anyone to use it. We're going to also be releasing unity asset packages as well.

StephenHodgson:
I'm also a bit concerned about having this in master at the moment as well. 

There's going to be a major namespace change and other braking changes when we move from HoloToolkit to MixedRealityToolkit namespaces. @jwittner  Do we want to make the NuGet package HoloToolkit until we get vNEXT complete?

jwittner:
Given that the repo and the project has been changed to the new name I think it makes more sense to reinforce the change and adopt the new standard instead of introducing more work that needs to change in the future. There's no technical necessity that the code folders match and it's the folders, not the package, that's out of compliance with the standard. 

Regarding the splitting out of the other components into their own packages - love the idea, lots of customers want that. It will be a big effort from my tests the last time I tried. Sharing was literally the only component folder that didn't have crossing or circular dependencies with the others.

jwittner:
Regarding the changing namespaces and folders, I think updating these specifications and the build script in light of those is relatively simple work for the value gained in enabling people to easily package up their own forks and test them in the same manner that the official releases are built.

StephenHodgson:
Definitely think we should enforce the HoloToolkit vs MixedRealityToolkit namespaces. This is a distinct change, and we've already put it out there that projects built with the HoloToolkit will not be compatible with the MixedRealityToolkit.

jwittner:
> My only issue with switching to NuGet rather than .UnityAsset files is adoption. Most Unity devs will struggle with NuGet, especially as it's another dependency in their project. (Would be better if Unity natively supported NuGet)

It is another dependency, but access is pretty easy as it's available on the Unity Asset Store and functions similar to how NuGet works in Visual Studio. I've ran into lots of .NET enterprise developers moving into Unity for MR and it's a natural fit for them. As Unity catches up it's .NET support, we're seeing more existing technologies on NuGet work out of the box. NuGet also breaks a dependency on Unity for an OSS cross platform self hostable package distribution solution.

I agree though that native integration would be better. I'll step down from the soap box now. :)

StephenHodgson:
2018 has a package manager that uses NuGet

jwittner:
@StephenHodgson Where did you read that it uses NuGet? Certainly looks like a layer on top, but I didn't see anything official in the docs about NuGet.

The new docs (ex, [here](https://docs.unity3d.com/Packages/com.unity.package-manager-ui@1.8/manual/index.html#unity-package-manager)), seem like they might be build on [DocFX](https://dotnet.github.io/docfx/), which is awesome because it's great.

I'll also note that in the Package Manager UI docs, all the packages are [SemVer](https://semver.org/) compatible 3 component versions. ;)

davidkline-ms:
Sorry for the late response (been out a few days). @StephenHodgson, I like having **Toolkit** in the namespace. Especially with us looking at integrating our documentation with the platform docs.

StephenHodgson:
@davidkline-ms yup, @SimonDarksideJ is going to update the namespace in vNEXT to align with this pattern.

StephenHodgson:
> Where did you read that it uses NuGet? Certainly looks like a layer on top, but I didn't see anything official in the docs about NuGet.

@jwittner maybe I missread. Sorry. It was that there was an issue where NuGetPackage manager and theirs used a similar folder.

SimonDarksideJ:
namespace is currently updated to 
```
Microsoft.MixedReality.xxx 
```
Currently, is there some request to change again?

The asset is called the MixedRealityToolkit.  Not sure what value it adds elongating the namespace by adding more.

davidkline-ms:
@SimonDarksideJ,
It doesn't add much value in the code (folks know they are using the toolkit), but it may add value with respect to the integrated documentation.

SimonDarksideJ:
So what's the final call (while I'm actually editing the code at the moment?

davidkline-ms:
I vote for keeping Toolkit. :)

SimonDarksideJ:
Done and done.  No going back now in the next meeting.
```
Microsoft.MixedReality.Toolkit.xxx
```
It is.

Now I have so far.
```
Microsoft.MixedReality.Toolkit.Internal.Definitions;
Microsoft.MixedReality.Toolkit.Internal.Interfaces;
Microsoft.MixedReality.Toolkit.Internal.Managers
```
(I hope namespaces don't have the same floor as the NTFS :P max 256 chars)

StephenHodgson:
> (I hope namespaces don't have the same floor as the NTFS :P max 256 chars) 

😆

SimonDarksideJ:
Why are there shader changes included as part of this PR @jwittner ?
You should revert this files if they are not part of your changes, to avoid unintentional automatic upgrade changes.

jwittner:
@SimonDarksideJ Bah! Nice catch, must've got caught in my updates - resolving now.

jwittner:
I'm down to move it to vNEXT if it can stand as is. =)

jwittner:
@StephenHodgson Ok, to clarify - we want this change applied to 'MRTK-Version-Next', and we want the 'HoloToolkit' naming as a PR against 'may18_dev'?

StephenHodgson:
@jwittner that is correct.

Issue #1872, opened 2018-03-23
[dev] nullreference exception when attempting to tap


davidkline-ms:
## Overview
When I run the InputTapTest, or any other scene with an object that supports tap, I hit a NullReferenceException on the call to 
                ExecuteEvents.ExecuteHierarchy(inputEventData.selectedObject, pointerInputEventData, ExecuteEvents.pointerDownHandler);
in InputMananger.RaiseSourceDown()

## Expected Behavior
The tap to complete w/o exception.

## Actual Behavior
A null reference is hit when RaiseSourceDown calls ExecuteEvents.ExecuteHierarchy() is called. On inspection, inputEventData.selectedObject is null. This seems odd because the cursor is on the object implementing IInputClickHandler.

## Steps to reproduce
* Set -Examples\Input\Scenes\InputTapTest as the active scene in the app.
* Run the project on HoloLens
* Tap on the blue cube.

## Unity Editor Version
2017.2.1p2

## Mixed Reality Toolkit Release Version
Dev_Working_Branch

PR #1896, opened 2018-03-30
Fixed null ref when pointer data does not have a selected object.

StephenHodgson:
Overview
---
Updated Focus Manager `TryGetFocusedObject` to make sure we set the pointer data's selected object with the focus details.

Also added a null check in the Input Manager's `RaiseInputUp` and `RaiseInputDown` methods before we execute the pointer handlers for the `UnityEngine.Graphic` objects.

Changes
---
- Fixes: #1872

### DIFF
--- a/Assets/MixedRealityToolkit/InputModule/Scripts/Focus/FocusManager.cs
+++ b/Assets/MixedRealityToolkit/InputModule/Scripts/Focus/FocusManager.cs
@@ -340,21 +340,11 @@ namespace MixedRealityToolkit.InputModule.Focus
 
         public GameObject TryGetFocusedObject(BaseEventData eventData)
         {
-            FocusDetails? details = TryGetFocusDetails(eventData);
-
-            if (details == null)
-            {
-                return null;
-            }
-
             IPointingSource pointingSource;
             TryGetPointingSource(eventData, out pointingSource);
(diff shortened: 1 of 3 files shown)


calebcannon:
I hit this issue while testing Unity 2017.4f1. The fix here didn't solve it for me, but I was able to correct it by moving the call to InitializeEventDatas from Awake() to Start() in InputManager.cs. The cause seems to be that InputEventData.selectedObject is not a field, but a property. That property is pulling the selected object from an internal reference to the EventSystem that was in activeat the time of the InputManagers creation, but wasn't setup properly because EventSystem.current was null at the time Awake was called.  E.g., this is another singleton bug related to Awake/Start usage.

StephenHodgson:
@calebcannon thanks for the heads up.  That's a different issue entirely that will be addressed in the Input Refactor.

jerrygg:
qq 20180413163919
The problem still exists

StephenHodgson:
Is the event data null? or the pointing source?

davidkline-ms:
the value of pointingSource shouldn't matter here, as it is the out param. my gut says it's the eventData

StephenHodgson:
Yeah, that should pretty much be impossible.  We're creating all our event datas in the InputManager.

Could be a script execution order issue.

davidkline-ms:
@jerrygg,
What version of Unity are you using? That will greatly help our investigation.
Thanks!

PR #1920, opened 2018-04-05
Mrtk version next - mixed reality manager

SimonDarksideJ:
Overview
---
First pass on the Mixed Reality manager and it's accompanying MixedRealityManagerConfiguration scriptable configuration system. (a bit blank as there aren't any managers yet)

Still ToDo
---

- [x] Update scripts to documentation standards
- [x] Review accompanying documentation
- [ ] Write Unit tests for manager (@StephenHodgson  has offered to help post merge)

Changes
---
- Introduces new MixedRealityManager - (_Core\Managers)
- Introduces definitions for creating manager classes (IManager / BaseManager)
Need to work with Stephen about the Manager Abstraction through interfaces.
- Adds documentation for creating managers 
(initial draft, will be updated following the addition of some :D )
- Updates Script icons (image may be replaced by @cre8ivepark )
- Edited .Meta files in project to remove unwanted Unity stuff (Unity Free version and creation date)
- New System base interfaces created (samples) for Boundary and InputSystem.  These identify the generic way the toolkit references features (_Core\Interfaces)
- Added sample implementations for Boundary and Input systems (InputSystem - Boundary)
- Added an event to allow active profile changing, informing managers when configuration changes.  Note, configuration properties cannot be updated at runtime, only the entire profile at the present.

*Note, this isn't the final manager by a long shot.  Through working with new managers, this will likely evolve in Sprint 1 and beyond.  Especially the Configuration Profile.

Discussion Point
---
I have two example implementations for Manager construction

1. Base Manager registers all events and exposes virtual events that are overridden / consumed by child managers  (BaseManager / MixedRealityInputManager)

2. The child manager registers all events independently (MixedRealityBoundarySystem)

Which is the preferred path?


SimonDarksideJ:
Not sure why the Conflict files are complaining.  Might try a separate PR to update them manually.

Assets/MixedRealityToolkit/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <summary>
+        /// MixedRealityBoundaryManager constructor
+        /// </summary>
+        public MixedRealityBoundaryManager()
+        {
+            //Attach to Manager events
+            MixedRealityManager.Instance.InitializeEvent += InitializeInternal;
+            MixedRealityManager.Instance.UpdateEvent += Update;
+            MixedRealityManager.Instance.ProfileUpdateEvent += ProfileUpdate;
+            MixedRealityManager.Instance.DestroyEvent += Destroy;
```
StephenHodgson: If we're using interfaces for the managers, we don't need to subscribe, we can directly call the methods in each child manager.

Assets/MixedRealityToolkit/_Core/Definitions/BaseManager.cs
```diff
@@ -6,26 +6,31 @@
...
+    public class BaseManager : Interfaces.IMixedRealityManager
     {
 
         /// <summary>
-        /// 
+        /// Controls whether the manager is enabled and active in the scene
         /// </summary>
         [SerializeField]
+        [Tooltip("Is the selected manager enabled")]
         private bool enabled;
```
StephenHodgson: I don't like that we have enables on pure c# class., the summary is misleading bc they won't be in the scene.

We should only consider a manager enabled if it's enabled via the profile, and listed in the MixedRealityManager's active manager list.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        #endregion
+
+        #region Focus Control
+
+        [Header("Focus Options")]
+        /// <summary>
+        /// Enable and configure the Focus component for the Mixed Reality Toolkit
+        /// </summary>
+        public bool EnableFocus;
```
StephenHodgson: focus should be part of the input system.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -0,0 +1,109 @@
...
+        #endregion
+
+        #region Motion Controllers
+
+        /// <summary>
+        /// Enable and configure the controller rendering for the Mixed Reality Toolkit
+        /// </summary>
+        [Header("Motion Controllers")]
+        [Tooltip("Enable the Motion Controllers on Startup")]
+        public bool EnableControllers;
```
StephenHodgson: controllers should also be part of the input system.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -0,0 +1,109 @@
...
+        #endregion
+
+        #endregion
+
+        #region ISerializationCallbackReceiver Interface
+
+        /// <summary>
+        /// Unity function to prepare data for serialization, unused in the Mixed Reality Toolkit
+        /// </summary>
+        public void OnBeforeSerialize() { }
```
StephenHodgson: Actually we do need to use this.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -0,0 +1,251 @@
...
+        #region Mixed Reality Manager Events
+
+        /// <summary>
+        /// Delegate for the Initialize event
+        /// </summary>
+        public delegate void InitializeHandler();
+        /// <summary>
+        /// Initialize event called after all managers have been registered. Informs all managers to initialize (Awake)
+        /// </summary>
+        public event InitializeHandler InitializeEvent;
```
StephenHodgson: Instead of subscribing to events, let's just call the methods directly.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -0,0 +1,251 @@
...
+
+        /// <summary>
+        /// Delegate for the Profile Update event
+        /// </summary>
+        public delegate void ProfileUpdateHandler();
+        /// <summary>
+        /// The ProfileUpdate event fires whenever the MixedRealityManager active profile is replaced.
+        /// *Note Profile options cannot be changed once the project is running.
+        /// </summary>
+        public event ProfileUpdateHandler ProfileUpdateEvent;
```
StephenHodgson: Instead of ProfileUpdate, let's call this "Reset" so the managers can do what they need.  Also gives us a more generic way to reset the managers as needed.

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/Singleton.cs
```diff
@@ -0,0 +1,170 @@
...
+namespace Microsoft.MixedReality.Toolkit.Internal.Utilities
+{
+    /// <summary>
+    /// Singleton behavior class, used for components that should only have one instance.
+    /// <remarks>Singleton classes live on through scene transitions and will mark their 
+    /// parent root GameObject with <see cref="Object.DontDestroyOnLoad"/></remarks>
+    /// </summary>
+    /// <typeparam name="T">The Singleton Type</typeparam>
+    [DisallowMultipleComponent]
+    public abstract class Singleton<T> : MonoBehaviour where T : Singleton<T>
```
StephenHodgson: I'm a little apprehensive to putting this class into vNEXT.  I thought we were going to move away from the Singleton pattern.

StephenHodgson:
@SimonDarksideJ I'm gong to open a PR with some suggested changes.

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/Singleton.cs
StephenHodgson: I removed this class from the project.  Most of its guts now live in MixedRealityManager. It's the only class that will be a singleton tied to a game object reference in the scene.  All MRTK managers will be pure c# classes managed by this instance.

SimonDarksideJ: I may disagree on this.  While I agree and accept that in the MRTK-vNext, the only class using the Singleton will be the manager.
However, that's not to say that other projects will already be consuming the singleton from the MRTK for their own projects, so it's still "useful" (even though we don't like it) for consumers projects.   Thoughts?

Assets/MixedRealityToolkit/Boundary/MixedRealityBoundarySystem.cs
SimonDarksideJ: OK, that's why I put up both choices for us to determine.  The subscribers in children makes more sense as each manager can choose what to subscribe to, however, having it in the base class is cleaner :D 
Right, that's the path then

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
SimonDarksideJ: This was only there as an example as it was separate before.  I expect a LOT of change to happen in this configuration profile as vNext evolves.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
SimonDarksideJ: We may use it later, just not used for now :D

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: OK, THat is certainly another way.  Having the methods defined in the interface and virtuals in the base.  Fair enough.

SimonDarksideJ:
Done some updates post your updates @StephenHodgson , mainly to:

* enforce BaseManager use
* brought back the "Enabled" property to allow components to be "turned off" at runtime
* added code to call the Reset function
* changed profile configuration options to be read-only at runtime (we can't currently update managers with new values yet)


Still think we should leave the Singleton in the project as it's very useful for consumers to use (if they wish) and reduce the bloat in the MRManager from it.  Discuss :D

Assets/MixedRealityToolkit/Boundary/MixedRealityBoundarySystem.cs
SimonDarksideJ: Agreed with your new method

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
SimonDarksideJ: Agreed

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
SimonDarksideJ: Not sure I agree, as they are part of the active SDK, sure the InputSystem should grab these references, but at the time of SDK discovery, the input system wont exist yet (and we won't know which one we are using)

These are set (or will be) by the SDK discovery

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
SimonDarksideJ: Agreed.  But it won't "do" anything yet until we build a custom editor (the rabbit hole I fell in) to configure this

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: Agreed

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/Singleton.cs
StephenHodgson: Is rather see ya move away from it because it's caused us too much pain in the past, and the manager system is supposed to be the way forward. It'll reduce confusion and promote adoption of the system when people try to add to it.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
StephenHodgson: I already updated the usage to serialize and deserialize the managers appropriately.

Assets/MixedRealityToolkit/_Core/Definitions/BaseManager.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// The base manager implements the IMixedRealityManager interface and provides default properties for all managers
+    /// </summary>
+    public class BaseManager : Interfaces.IMixedRealityManager
+    {
+        /// <summary>
+        /// Controls whether the manager is enabled and active in the scene
+        /// </summary>
+        [SerializeField]
+        [Tooltip("Is the selected manager enabled")]
+        private bool enabled;
```
StephenHodgson: I don't like that we have enables on pure c# class., the summary is misleading bc they won't be in the scene.

We should only consider a manager enabled if it's enabled via the profile, and listed in the MixedRealityManager's active manager list.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -0,0 +1,104 @@
...
+        #region Mixed Reality Manager configurable properties
+
+        /// <summary>
+        /// Enable and configure the Input System component for the Mixed Reality Toolkit
+        /// </summary>
+        [Header("Input Settings")]
+        [Tooltip("Enable the Input System on Startup")]
+        [SerializeField]
+        private bool enableInputSystem = true;
+        public bool EnableInputSystem { get { return enableInputSystem; } }
```
StephenHodgson: Why change this to read-only?

Assets/MixedRealityToolkit/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class MixedRealityBoundaryManager : BaseManager, IMixedRealityBoundarySystem
+    {
+        /// <summary>
+        /// MixedRealityBoundaryManager constructor
+        /// </summary>
+        public MixedRealityBoundaryManager()
+        {
+            // TODO define any constructor requirements
+        }
+        /// <summary>
```
StephenHodgson: nit space

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -0,0 +1,323 @@
...
+        //[SerializeField]
+        //[Tooltip("The collection of currently active / detected controllers")]
+        //private Controller[] activeControllers = null;
+
+        ///// <summary>
+        ///// The Active Headset property maintains the Headsets/SDK detected by the Mixed Reality manager on startup
+        ///// </summary>
+        //[SerializeField]
+        //[Tooltip("The currently active / detected Headset or SDK")]
+        //private Headset activeHeadset = default(Headset);
```
StephenHodgson: I just commented these out until we use them

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -0,0 +1,323 @@
...
+        #endregion Mixed Reality Manager Profile configuration
+
+        #region Active SDK components
+
+        /// <summary>
+        /// The Active Controllers property lists all the controllers detected by the Mixed Reality manager on startup
+        /// </summary>
+        //[SerializeField]
+        //[Tooltip("The collection of currently active / detected controllers")]
+        //private Controller[] activeControllers = null;
```
StephenHodgson: I just commented these out until we use them

Assets/MixedRealityToolkit/Boundary/MixedRealityBoundarySystem.cs
SimonDarksideJ: FFS, grrr

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: Fair enough

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: Fair enough

PR #1946, opened 2018-04-12
Trim end of keywords

LuckyXII:
Fix so recognized text "example_keyword" and keyword added with SpeechInputSource "example_keyword " (notice the extra space) are treated equally
Appending an extra space when adding keywords is an easy to make mistake and also easy to miss while debugging, causes events not being raised even though keyword is correctly recognized

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/SpeechInputHandler.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/SpeechInputHandler.cs
@@ -62,7 +62,7 @@ namespace HoloToolkit.Unity.InputModule
             for (int index = 0; index < keywordCount; index++)
             {
                 KeywordAndResponse keywordAndResponse = Keywords[index];
-                string keyword = keywordAndResponse.Keyword.ToLower();
+                string keyword = keywordAndResponse.Keyword.ToLower().TrimEnd();
 
                 if (responses.ContainsKey(keyword))
                 {


davidkline-ms:
Changed the merge base to may18_dev as changes are not taken directly into master.

Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/SpeechInputHandler.cs
```diff
@@ -62,7 +62,7 @@ protected virtual void Start()
             for (int index = 0; index < keywordCount; index++)
             {
                 KeywordAndResponse keywordAndResponse = Keywords[index];
-                string keyword = keywordAndResponse.Keyword.ToLower();
+                string keyword = keywordAndResponse.Keyword.ToLower().TrimEnd();
```
davidkline-ms: What is the behavior if there is whitespace at the beginning of a keyword? Should we trim both ends?

LuckyXII:
Prepend whitespace(s) did not cause any unexpected behavior it was added
without any whitespace to the dictionary.

Maybe toLower and Trim should be called in speechInputSource instead
though, this way anyone wanting to use it but write their own handler
wouldn't need to think about whitespace and cases

Sorry for requesting it to master i just committed it and sent it

davidkline-ms:
@LuckyXII,

No worries about the master request, we catch those :)

I like the idea of putting the fix in speechInputSource.

Thanks!
David

LuckyXII:
No worries glad I could contribute! Keep up the awesome work

Issue #1949, opened 2018-04-12
[Patch4] Address potential NullReferenceException on Select


davidkline-ms:
Hitting this in 2017.2.1.4-rc, need to port #1896 from dev->patch4

PR #1954, opened 2018-04-13
Fix crash (null reference exception) when air-tapping objects

davidkline-ms:
Add protections for null in FocusMananger and InputManager based on dev_working_branch changes from #1896.

This issue was found on 2017.2.1.4 RC and is needed for the release.

Fixes: #1949 .

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs
@@ -353,6 +353,10 @@ namespace HoloToolkit.Unity.InputModule
 
         public bool TryGetPointingSource(BaseEventData eventData, out IPointingSource pointingSource)
         {
+            // pre-initialize pointingSource to null, assuming we will return false
+            pointingSource = null;
+            if (eventData == null) { return false; }
+
             for (int i = 0; i < pointers.Count; i++)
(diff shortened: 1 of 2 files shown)


Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs
```diff
@@ -343,25 +343,20 @@ public void UnregisterPointer(IPointingSource pointingSource)
...
-
-            return details.Value.Object;
+            return pointerInputEventData.selectedObject;
         }
 
         public bool TryGetPointingSource(BaseEventData eventData, out IPointingSource pointingSource)
         {
+            // pre-initialize pointingSource to null, assuming we will return false
+            pointingSource = null;
+            if (eventData == null) { return false; }
```
StephenHodgson: Not sure how/why the event data could be null. 

Seems like it's just treating the symptom not the cause.

SimonDarksideJ: I agree, this is just hiding the problem rather than fixing the issue.
I guess we'll review in vNext

davidkline-ms:
this was me being EXTRA paranoid. until vNext, I would rather have this method return false instead of throwing

PR #1973, opened 2018-04-18
Patch4 dev

67688FEA:
Overview
---
Sometimes the target is too far away to make the cursor too small to see,and sometimes we want the cursor size to remain the same.
So I modified the Cursor. cs, added a bool variable 'IsCameraCursorDistance', and renamed the 'SurfaceCursorDistance' to 'CursorDistance', the 'IsCameraCursorDistance' control the 'CursorDistance' means the distance is from the Cursor to hit or  camera,the former retain the original function, and the latter can keep the Cursor size unchanged.
Then in the method UpdateCursorTransform(),I will calculate the cursor position based on 'IsCameraCursorDistance' and 'CursorDistance'.

Changes
---
- Fixes: #.


67688FEA:
To resubmit

67688FEA:
To resubmit

PR #1974, opened 2018-04-18
Add a function that keeps the same distance from cursor to the camera.

67688FEA:
Overview
---
Sometimes the target is too far away to make the cursor too small to see,and sometimes we want the cursor size to remain the same,that means keeps the same distance from the camera.
So I modified the Cursor. cs, added a bool variable 'IsCameraCursorDistance', and renamed the 'SurfaceCursorDistance' to 'CursorDistance', the 'IsCameraCursorDistance' control the 'CursorDistance' means the distance is from the Cursor to hit or  camera,the former retain the original function, and the latter can keep the Cursor size unchanged.
Then in the method UpdateCursorTransform(),I will calculate the cursor position based on 'IsCameraCursorDistance' and 'CursorDistance'.

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
@@ -35,11 +35,19 @@ namespace HoloToolkit.Unity.InputModule
         [Tooltip("The maximum distance the cursor can be with nothing hit")]
         public float DefaultCursorDistance = 2.0f;
 
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


davidkline-ms:
Changed the base from master -> may18_dev as master does not receive direct checkins.

davidkline-ms:
@67688FEA, Have you tried adding the FixedAngularSize script onto your cursor? that will keep it the same size visually without needing to change the distance.

Thanks!

67688FEA:
@davidkline-ms Thanks for your advice!

Issue #1770, opened 2018-02-27
Add Spatial Mapping layer to "Apply Mixed Reality Project Settings"


davidkline-ms:
## Overview
The Spatial Mapping Manager defaults to using layer 31 for the mesh. Projects generally need to set layer 31 to Spatial Mapping for easy reference (ex: configuring the Focus Manager).

I'd like to propose adding an option to the Apply Mixed Reality Project Settings dialog to set the name of layer 31 to Spatial Mapping.

davidkline-ms:
This was released in the 2017.4.0.0 release candidate

Issue #1443, opened 2017-11-27
Spatial Mapping Physics Layer isn't visible in MRTK project's layer


StephenHodgson:
When users import and use the MRTK, they need to know about the special layers reserved for Spatial Mapping (layer 31) and we need a way to prompt users to let them know we'd like to add it to their project on import.

PR #1991, opened 2018-04-19
add option to set default spatial mapping layer

davidkline-ms:
This change adds a checkbox (selected by default) to the Apply Mixed Reality Project Settings dialog.

When selected, layer 31 is checked to see if it is in use (with a name other than Spatial Mapping) and if not, sets the layer name to "Spatial Mapping". If we are able to set the layer, the Spatial Perception capability is also set (setting the layer implies it will be used).

If we fail to set the layer (the preparation code fails, or layer 31 has a name that implies it is in use, a message is displayed stating that we failed and that the user should check Tags & Layers in the inspector.

- Fixes: #1443.
- Fixes: #1770


StephenHodgson:
Is there a PR for this in Dev?

davidkline-ms:
There will be one for vNext when @SimonDarksideJ's PR is (#1995) approved.

davidkline-ms:
Given your (@StephenHodgson) comment in #1995, I will do a dev branch PR.

Issue #1913, opened 2018-04-04
FPSDisplay.cs calls InitBuffer every Update() when no TextMesh component added


Troy-Ferrell:
## Overview
FPSDisplay.cs has the following if check every update() call:
if (fpsBuffer == null || fpsBuffer.Length != frameRange || textMesh == null)
         
The problem with this code is it only checks the textMesh property, not the uGUIText property. Thus, if the developer only attached to the uGUIText property, not the textMesh, then the if condition is evaluated true resulting in a call to InitBuffer() every update frame. 

This results in the following:
1) wasted performance (GC allocation of new int buffer as well as call to GetComponent<TextMesh>()) every frame 
2) Invalid FPS displayed to the nGuiText property component (since the bufferIndex is reset to 0 every frame)

Furthermore, note this is an issue if no TextMesh or Text component can be found every frame as well

## Expected Behavior
The buffer is created once and the script only tries to get a TextMesh and a Text Component once via GetComponent<>()

Ideally, the InitBuffer() is just called in Start() instead of using this if statement every update method

## Actual Behavior
The if statement is evaluated true every update method, resulting in 0 being displayed as FPS and wasted GC allocation every frame

## Steps to reproduce
Create a Text component to some canvas, attach the FPS display script and hook up the Text component the nGuiText property
Run game
See actual behavior

## Unity Editor Version
2017.3

## Mixed Reality Toolkit Release Version
Latest on master - 2017.2.1.3

davidkline-ms:
This was released in the 2017.4.0.0 release candidate

PR #1993, opened 2018-04-20
fix fpsdisplay perf, add comments and frame range

davidkline-ms:
FpsDisplay was repeatedly fetching components if the user did not use both a TextMesh and a uGui Text component. Fixed by adding Start() and calling InitBuffer there instead of in Update.

Also added a range (1-300) to the number of frames used to calculate FPS and some documentation comments.

Fixes: #1913 .

### DIFF
--- a/Assets/HoloToolkit/Utilities/Scripts/FpsDisplay.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/FpsDisplay.cs
@@ -12,16 +12,26 @@ namespace HoloToolkit.Unity
     [RequireComponent(typeof(TextMesh))]
     public class FpsDisplay : MonoBehaviour
     {
+        /// <summary>
+        /// Reference to TextMesh component where the FPS should be displayed.
+        /// </summary>
         [Tooltip("Reference to TextMesh component where the FPS should be displayed.")]
         [SerializeField]
         private TextMesh textMesh;
 
+        /// <summary>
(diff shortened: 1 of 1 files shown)


Assets/HoloToolkit/Utilities/Scripts/FpsDisplay.cs
```diff
@@ -12,16 +12,26 @@ namespace HoloToolkit.Unity
...
         [Tooltip("Reference to uGUI text component where the FPS should be displayed.")]
         [SerializeField]
         private Text uGUIText;
 
+        /// <summary>
+        /// How many frames should we consider into our average calculation?
+        /// </summary>
         [Tooltip("How many frames should we consider into our average calculation?")]
         [SerializeField]
+        [Range(1, 300)]
```
StephenHodgson: Why 300?

davidkline-ms: it's 5 seconds at 60fps. no reason other than that.  this field feels like it needs a range value. do you have a suggested max number of frames?

StephenHodgson: Nope just curious

StephenHodgson:
@davidkline-ms is there a PR for dev?

davidkline-ms:
@StephenHodgson, thanks for keeping me honest. There will be one today :)

PR #2002, opened 2018-04-23
Updated the Mixed Reality manager to now also support runtime Managers.

SimonDarksideJ:
Overview
---
Updated the Mixed Reality manager to now also support runtime Managers.  Main difference is that design time managers are unique (e.g. InputSystemManager) whereas there maybe many runtime managers.
Also added support for the Disable and Enable MonoBehaviour functions.

See "/External/Documentation/MixedRealityManagerConstruction.md" for more details

Also added some simple Unit Tests (WIP) for the MR Manager, to test DI implementation and use.

Changes
---
- Fixes: Resolved some null reference issues when initialising
- Fixes: Singleton pattern updated in manager to be Unit Test Friendly
- Additions: Assembly refs added for Unit Testing only atm.  For further review.

TODO
---
- [ ] Review new Input System components for manager adoption (design and runtime)
- [ ] Evolve (still much to do )

### DIFF
--- a/Assets/MixedRealityToolkit-Tests/MixedRealityManagerTests.cs
+++ b/Assets/MixedRealityToolkit-Tests/MixedRealityManagerTests.cs
@@ -1,0 +1,403 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Managers;
+using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.TestTools;
+
+namespace Microsoft.MixedReality.Toolkit.Tests
+{
+    public class MixedRealityManagerTests
+    {
+        public void CreateMRManager()
+        {
+            // Create The MR Manager
+            GameObject go = GameObject.Find("Mixed Reality Manager");
+            if (!go)
+            {
+                go = new GameObject("Mixed Reality Manager");
+            }
+            go.AddComponent<MixedRealityManager>();
+        }
+
+        [Test]
+        public void Test01_CreateMixedRealityManager()
+        {
+            // Create The MR Manager
+            CreateMRManager();
+            GameObject go = GameObject.Find("Mixed Reality Manager");
+            Assert.AreEqual("Mixed Reality Manager", go.name);
+        }
+
+
+        [Test]
+        public void Test02_TestNoMixedRealityConfigurationFound()
+        {
+            LogAssert.Expect(LogType.Error, "No Mixed Reality Configuration Profile found, cannot initialize the Mixed Reality Manager");
+            CreateMixedRealityConfiguration(MixedRealityManager.Instance);
+        }
+
+        [Test]
+        public void Test03_CreateMixedRealityManager()
+        {
+            //Test the Manager
+            CreateMixedRealityConfiguration(MixedRealityManager.Instance);
+            Assert.IsNotNull(MixedRealityManager.Instance);
+
+            // Create Test Configuration
+            Assert.IsEmpty(MixedRealityManager.Instance.ActiveProfile.ActiveManagers);
+            Assert.IsEmpty(MixedRealityManager.Instance.MixedRealityComponents);
+        }
+
+        [Test]
+        public void Test04_CreateMixedRealityInputManager()
+        {
+            // Create The MR Manager & Test configuration
+            CreateMixedRealityConfiguration(MixedRealityManager.Instance);
+
+            // Add Input System
+            MixedRealityManager.Instance.AddManager(typeof(Internal.Interfaces.IMixedRealityInputSystem), new InputSystem.MixedRealityInputManager());
+
+            // Tests
+            Assert.IsNotNull(MixedRealityManager.Instance.ActiveProfile);
+            Assert.IsNotEmpty(MixedRealityManager.Instance.ActiveProfile.ActiveManagers);
+            Assert.AreEqual(MixedRealityManager.Instance.ActiveProfile.ActiveManagers.Count, 1);
+            Assert.IsEmpty(MixedRealityManager.Instance.MixedRealityComponents);
+
+        }
+
+        [Test]
+        public void Test05_TestGetMixedRealityInputManager()
+        {
+            // Create The MR Manager & Test configuration
+            CreateMixedRealityConfiguration(MixedRealityManager.Instance);
+
+            // Add Input System
+            MixedRealityManager.Instance.AddManager(typeof(Internal.Interfaces.IMixedRealityInputSystem), new InputSystem.MixedRealityInputManager());
+
+            // Retrieve Input System
+            var inputSystem = MixedRealityManager.Instance.GetManager<Internal.Interfaces.IMixedRealityInputSystem>();
+
+            // Tests
+            Assert.IsNotNull(inputSystem);
+        }
+
+        [Test]
+        public void Test06_TestMixedRealityInputManagerDoesNotExist()
+        {
+            // Create The MR Manager & Test configuration
+            CreateMixedRealityConfiguration(MixedRealityManager.Instance);
+
+            //Check for Input System
+            var inputSystemExists = MixedRealityManager.Instance.ManagerExists<Internal.Interfaces.IMixedRealityInputSystem>();
+
+            // Tests
+            Assert.IsFalse(inputSystemExists);
+        }
+
+
+        [Test]
+        public void Test07_TestMixedRealityInputManagerExists()
+        {
+            // Create The MR Manager & Test configuration
+            CreateMixedRealityConfiguration(MixedRealityManager.Instance);
+
+            // Add Input System
+            MixedRealityManager.Instance.AddManager(typeof(Internal.Interfaces.IMixedRealityInputSystem), new InputSystem.MixedRealityInputManager());
+
+            //Check for Input System
+            var inputSystemExists = MixedRealityManager.Instance.ManagerExists<Internal.Interfaces.IMixedRealityInputSystem>();
+
+            // Tests
+            Assert.IsTrue(inputSystemExists);
+        }
+
+        [Test]
+        public void Test08_CreateMixedRealityComponent()
+        {
+            // Create The MR Manager & Test configuration
+            CreateMixedRealityConfiguration(MixedRealityManager.Instance);
+
+            var component = new TestComponentManager1();
+
+            //Add test component
+            MixedRealityManager.Instance.AddManager(typeof(ITestComponentManager1), component);
+
+            // Tests
+            Assert.IsNotNull(MixedRealityManager.Instance.ActiveProfile);
+            Assert.IsEmpty(MixedRealityManager.Instance.ActiveProfile.ActiveManagers);
+            Assert.AreEqual(MixedRealityManager.Instance.MixedRealityComponents.Count, 1);
+        }
+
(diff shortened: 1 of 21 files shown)


Assets/MixedRealityToolkit-Tests/MixedRealityManagerTests.cs
```diff
@@ -0,0 +1,403 @@
...
+            GameObject go = GameObject.Find("Mixed Reality Manager");
+            if (!go)
+            {
+                go = new GameObject("Mixed Reality Manager");
+            }
+            go.AddComponent<MixedRealityManager>();
+        }
+
+        [Test]
+        public void Test01_CreateMixedRealityManager()
```
StephenHodgson: Should we put `Test<number>_` prefix on these?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -4,6 +4,8 @@
 using Microsoft.MixedReality.Toolkit.Internal.Definitions;
 using Microsoft.MixedReality.Toolkit.Internal.Interfaces;
 using System;
+using System.Linq;
```
StephenHodgson: Tisk tisk. Any way to remove this? Or are we using a Linq specific method?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -31,19 +33,47 @@ public class MixedRealityManager : MonoBehaviour
...
+            // Reset all active managers in the registry
             foreach (var manager in ActiveProfile.ActiveManagers)
             {
                 manager.Value.Reset();
             }
+
+            // Reset all registered runtime components
+            foreach (var manager in MixedRealityComponents)
+            {
+                manager.Item2.Reset();
```
StephenHodgson: What's Item2?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -92,11 +130,22 @@ private void Initialize()
                 //Enable Boundary (example initializer)
                 AddManager(typeof(IMixedRealityBoundarySystem), new InputSystem.MixedRealityBoundaryManager());
             }
+
+            //TODO should this be optional?
+            //Sort the managers based on Priority
+            var orderedManagers = ActiveProfile.ActiveManagers.OrderBy(m => m.Value.Priority);
```
StephenHodgson: I do like this, but let's try and order without Linq

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -117,11 +166,15 @@ public static MixedRealityManager Instance
                 {
                     return instance;
                 }
-
+//Modified to enable Unit Tests
+#if UNITY_EDITOR
+                if (UnityEditor.EditorApplication.isPlaying && !searchForInstance)
```
StephenHodgson: Let's replace with `Application.IsEditor` so we can get rid of the `#Ifdef`'s

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityManager.cs
```diff
@@ -24,6 +34,16 @@ public interface IMixedRealityManager
...
 
+        /// <summary>
+        /// Optional Disable function to pause the manager.
+        /// </summary>
+        void Disable();
+
+        /// <summary>
+        /// Optional Enable function to enable / re-enable the manager.
+        /// </summary>
+        void Enable();
```
StephenHodgson: Let's sort these and make them in order:

- Initialize
- Enable
- Update
- Disable
- Destroy

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -93,9 +93,12 @@ void ISerializationCallbackReceiver.OnAfterDeserialize()
             // *Note This will only take effect once the Mixed Reality Toolkit has a custom editor for the MixedRealityConfigurationProfile
 
             ActiveManagers.Clear();
-            for (int i = 0; i < initialManagers.Length; i++)
+            if (initialManagers != null && initialManagers.Length > 0)
```
StephenHodgson: ` initialManagers?.Length > 0`

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
timGerken: @StephenHodgson , I don't see anything in the coding standards doc regarding not using Linq. Should there be; perhaps with an actual reason  as to why it shouldn't be used?

StephenHodgson: Actually there is, in a few places.  Give me a few min to find the relevant links (get it? 😆 )

timGerken: I opened it and searched for "linq" and came up empty.

StephenHodgson: [Windows Mixed Reality Developer Documentation:](https://docs.microsoft.com/en-us/windows/mixed-reality/performance-recommendations-for-unity#garbage-collections)

>Tips for avoiding allocations. The ideal way to avoid allocations (and thus collections) is to allocate everything you need at the startup of the application and just reuse that data while the app runs. There are plenty of cases where it's not possible to avoid allocations (for example APIs that return allocated objects), so this isn't actually possible in practice, but it's a good goal to shoot for.
>
>There are also some unintuitive reasons why allocations may occur even when you're not directly allocating something, here are some tips to avoid that:
>
> - **Do not use LINQ, as it causes heavy allocations.**
> - Do not use lambdas, as they cause allocations.
> - Beware of boxing! A common case for that is passing structs to a method that takes an interface as a parameter. Instead, make the method take the concrete type (by ref) so that it can be passed without allocation.
> - Prefer structs to classes whenever you can.
> - Default implementations for value equality and GetHashcode uses reflection, which is not only slow but also performs a lot of allocations.
> - Avoid foreach loops on everything except raw arrays and List. Each call potentially allocates an Enumerator. Prefer regular for loops whenever possible. (See https://jacksondunstan.com/articles/3805 for more info)

timGerken: Maybe that should be called out explicitly in the [MRTK coding standards](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/MRTK-Version-Next/CodingGuidelines.md) document as well. Thoughts? What about the next line regarding lambdas?

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
StephenHodgson: Actually `for (int i = 0; i < initialManagers?.Length; i++)`

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
StephenHodgson: Yes, we should probably align our coding standards to point and quote the Mixed Reality Documentation.

StephenHodgson: I'll use lambda's but only when necessary and when we start getting into async code that perf stuff won't really be as big of a deal.

timGerken: Any time I see a comment like the one at the top of this thread, I'd like to see a reason behind the suggestion. It will help future contributors understand better the logic and thinking behind the guidelines.

StephenHodgson: Understandable

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityManager.cs
SimonDarksideJ: Sure

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: I forgot about that one, thanks

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: Lol, without linq, it's horrible for loops :P

Assets/MixedRealityToolkit-Tests/MixedRealityManagerTests.cs
SimonDarksideJ: Yes, because Unity CANNOT sort Tests :D

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
SimonDarksideJ: Fair enough

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
StephenHodgson: fine with me. *brute force*

StephenHodgson:
One last comment, if we're using a custom assembly definition, we should also set one up for the rest of the project. Unity recommendations is either all in or all out on this one.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: Can we keep this one Linq thing in, as there are no other "nice" solutions.  Plus it's is only used once on startup.

StephenHodgson: Fine with me.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: I've removed ALL bar one implementation.  Because there just isn't another clean way to Sort :D

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: Tuple

StephenHodgson: Can we just name them as they are?

StephenHodgson: Or is this name hard coded in the pattern?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
StephenHodgson: That's fine with me. It's only called once on startup anyway.

Issue #2004, opened 2018-04-24
Add events to listen when a BoundingBoxRig is either activated or deactivated


rigma:
## Overview

On a project using MRTK where I am currently involved, I need to know when a bounding box rig is either activated or deactivated by the user.

To do so, I would like to add and use events, which are owned by the bounding box rig, that will be used to track the state of a rig.

## Proposed Behavior

When a bounding box rig is activated, a `BoundingBoxRig.Activated` event is fired.
When a bounding box rig is deactivated, a `BoundingBoxRig.Deactivated` event is fired.

davidkline-ms:
@rigma, Feel free to code up your proposal and submit a pull request. We love contributions!

rigma:
I am doing it right now!

PR #2011, opened 2018-04-25
Adding events for the BoundingBoxRig script

rigma:
Overview
---
This adds two events to the `BoundingBoxRig` script that can be listened to know when the rig is either activated or deactivated by the user. This follows the proposal made in #2004.

Changes
---
- Adding `BoundingBoxRig.Activated` and `BoundingBoxRig.Deactivated` events to the `BoundingBoxRig` script.
- Updating the example documentation to mention those events.


Assets/MixedRealityToolkit/InputModule/Scripts/InputEventData/BoundingBoxRigActivityEventData.cs
```diff
@@ -0,0 +1,28 @@
...
+    public class BoundingBoxRigActivityEventData : BaseInputEventData
+    {
+        /// <summary>
+        /// The bounding box rigged object.
+        /// </summary>
+        public GameObject BoundingBoxRiggedObject { get; private set; }
+
+        public BoundingBoxRigActivityEventData(EventSystem eventSystem) : base(eventSystem) { }
+
+        public void Initialize(IInputSource source, uint sourceId, object tag, GameObject boundingBoxRiggedObject)
```
StephenHodgson: Tag should be an object array and last, as it's optional.

`public void Initialize(IInputSource source, uint sourceId, GameObject boundingBoxRiggedObject, object[] tags = null)

Assets/MixedRealityToolkit/UX/Scripts/BoundingBoxes/BoundingBoxRig.cs
```diff
@@ -170,11 +171,13 @@ private bool ShowRig
 
         public void Activate()
         {
+            InputManager.Instance.RaiseBoundingBoxRigActivated(null, 0, gameObject);
```
StephenHodgson: If we're not going to pass an input source or source id we can probably remove them.

rigma: Okay, I though it was the standard way to call those methods, I'll change that.

StephenHodgson: No worries, it would be if the bounding box itself was in input source, but its not.

Assets/MixedRealityToolkit/InputModule/Scripts/InputEventData/BoundingBoxRigActivityEventData.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <summary>
+        /// The bounding box rigged object.
+        /// </summary>
+        public GameObject BoundingBoxRiggedObject { get; private set; }
+
+        public BoundingBoxRigActivityEventData(EventSystem eventSystem) : base(eventSystem) { }
+
+        public void Initialize(GameObject boundingBoxRiggedObject, object[] tags = null)
+        {
+            BaseInitialize(null, 0, tags);
```
StephenHodgson: We could still see weird stuff here if the input source is null.

Let's inherit from the BaseEventData instead of the BaseInputEventData.

rigma: Sure thing: I'm doing the change right now.

Assets/MixedRealityToolkit/InputModule/Scripts/InputEventData/BoundingBoxRigActivityEventData.cs
```diff
@@ -0,0 +1,29 @@
...
+using MixedRealityToolkit.InputModule.InputSources;
+using UnityEngine.EventSystems;
+using UnityEngine;
+
+namespace MixedRealityToolkit.InputModule.EventData
+{
+    /// <summary>
+    /// Describes activity of a bounding box rig events.
+    /// </summary>
+    public class BoundingBoxRigActivityEventData : BaseEventData
```
StephenHodgson: suggested name: `BoundingBoxEventData`

rigma:
I agree with your names. They are more easier to remember!

PR #2020, opened 2018-04-26
Some cleanup to conform to the coding guidelines.

StephenHodgson:


Assets/MixedRealityToolkit/_Core/Utilities/Scripts/RaycastResultComparer.cs
```diff
@@ -4,22 +4,12 @@
...
-
-        public ComparableRaycastResult(RaycastResult raycastResult, int layerMaskIndex = 0)
-        {
-            RaycastResult = raycastResult;
-            LayerMaskIndex = layerMaskIndex;
-        }
-    }
-
+    /// <summary>
+    /// Compares the Raycast Results from Unity's Graphic & Physics Raycasters.
```
StephenHodgson: use html `&amp;`

Assets/MixedRealityToolkit/InputSystem/Focus/TeleportTarget.cs
```diff
@@ -8,34 +8,19 @@ namespace Microsoft.MixedReality.Toolkit.InputSystem.Focus
 {
     public class TeleportTarget : InteractableHighlight, ITeleportTarget
     {
-        [Header("Hot Spot Options")]
         [SerializeField]
-        private bool overrideOrientation;
+        [Header("Hot Spot Options")]
```
davidkline-ms: Why reorder the Header and SerializeField attributes? The header applies to more than just this one field, correct?

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/ClipPlane.cs
```diff
@@ -32,18 +32,16 @@ private void OnDisable()
             ToggleClippingPlane(false);
         }
 
-#if UNITY_EDITOR
         private void Update()
         {
-            if (Application.isPlaying)
+            if (Application.isPlaying || Application.isEditor)
```
davidkline-ms: should this be && instead of || ? The original code got executed when it was in the editor, this looks like it will occur when the app is running on device as well.

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/HoverLight.cs
```diff
@@ -58,18 +59,16 @@ private void OnDisable()
             UpdateHoverLight();
         }
 
-#if UNITY_EDITOR
         private void Update()
         {
-            if (Application.isPlaying)
+            if (Application.isPlaying || Application.isEditor)
```
davidkline-ms: should this be && instead of || ? The original code got executed when it was in the editor, this looks like it will occur when the app is running on device as well.

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/Int3.cs
```diff
@@ -1,249 +0,0 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
davidkline-ms: Thank you for getting rid of this file.... it always makes me think that it should break us into the debugger (yes, i know i am old.... )

StephenHodgson: Haha, yea `Vector3Int` should be used instead.

Assets/MixedRealityToolkit/InputSystem/Focus/TeleportTarget.cs
StephenHodgson: It doesn't matter the order. I'm a bit strange about having lines grow. haha.  In reality if we want to do it the professional way we should remove the header attribute and do a custom inspector.

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/ClipPlane.cs
StephenHodgson: || is correct, but I needed to add a !

Should be `if (Application.isPlaying || !Application.isEditor)`

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/HoverLight.cs
StephenHodgson: Same applies here as well.

Issue #1923, opened 2018-04-09
AutoConfigureMenu broken without .Net 4.6


jbienzms:
## Overview
AutoConfigureMenu is broken in Dev_Working_Branch unless the player is set to .Net 4.6. 

## Expected Behavior
Mixed Reality Toolkit menu should appear and work in a new Unity project regardless of scripting runtime setting.

## Actual Behavior
Mixed Reality Toolkit menu only appears after changing the runtime to .Net 4.6. Console error messages do not explain the problem. 

This appears to be an issue with ActionExtensions.cs now depending on the null coalescing operator, which in turn depends on .Net 4.0+. This causes the project to fail to compile, which keeps the AutoConfigureMenu from appearing.

Since AutoConfigureMenu is the primary way most users will get their project setup correctly, it should work even without changing the framework version. If this is not possible (ill advised), setting the proper framework version needs to be called out and highlighted as a critical step in the Getting Started guide.

## Unity Editor Version
2017.4.0f1

## Mixed Reality Toolkit Release Version
Dev_Working_Copy

StephenHodgson:
My apologies, this was my fault.

The is supposed to be a dialogue that should pop up if you don't have it enabled. I guess because there are compiler errors that code never runs.

hannesne:
Maybe we could include a .Net 3.5 compatible compiled dll that is guaranteed to load independently from the rest of the MRTK, that can check prerequisites?

StephenHodgson:
I think the only real way to do this is to use [custom assembly definitions](https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html) but they only work in 2017.3+

StephenHodgson:
I think amending the docs is probably the best course of action for the time being.

jbienzms:
Interesting note about custom assembly definitions. I didn't realize those existed, so thank you for linking to the docs. How are you thinking this could help solve the problem? I could see if there was a way to mark the script as compiling against a different runtime that would work, but does this also control build order? If we say that it's targeted for Editor, will that guarantee it to compile even if the rest of the MRTK fails to build?

StephenHodgson:
Yes, it also controls the assembly build and load order.

StephenHodgson:
[...]

PR #2026, opened 2018-04-30
More custom assembly updates.

StephenHodgson:
Seeing as custom assembly code doesn't automatically exclude editor classes, and we need to define our own, I renamed the editor folder to inspectors.

Also fixed some issues with standalone compilation.

Ported over settings enforcement script and updated it to use .net standard 2, but not complain if users target the broader .net 4.x runtime.

See #1923


SimonDarksideJ:
I'll not merge this one until you say you're ready this time @StephenHodgson  :D

Assets/MixedRealityToolkit/_Core/Utilities/Scripts/Editor/Setup/EnforceEditorSettings.cs
```diff
@@ -0,0 +1,118 @@
...
+                    "Enable Visible Meta Files",
+                    "Later"))
+                {
+                    EditorSettings.externalVersionControl = "Visible Meta Files";
+                    Debug.Log("Updated external version control mode: " + EditorSettings.externalVersionControl);
+                    refresh = true;
+                }
+            }
+
+            if (PlayerSettings.scriptingRuntimeVersion != ScriptingRuntimeVersion.Latest)
```
SimonDarksideJ: We have checks on the ScriptingRuntime and ScriptingConpatibility
Should we also not have one for the Scripting Backend, to ensure they use IL2CPP as the new standard for 2018?

StephenHodgson: Done

Issue #1983, opened 2018-04-18
Cursor.prefab is missing the point light off holograms


keveleigh:
## Overview
In Cursor.prefab, the CursorOffHolograms component used to be a point light. In the process of adding the new hover light (to work with the new Toolkit Standard shader), this light was incorrectly removed.

Additionally, in the port to master, the hover light was incorrectly added to CursorOffHolograms instead of CursorOnHolograms.

## Expected Behavior
[All] When I'm not looking at a hologram using Cursor.prefab, I should see a floating point light.
[master] When I'm looking at a hologram using Cursor.prefab and the Toolkit Standard shader, I should see a light around my cursor.

## Actual Behavior
[All] Nothing is displayed when looking away from holograms.
[master] The hover light does not work.

## Steps to reproduce
1. Create a scene with Cursor.prefab and an object using the Toolkit's Standard shader.
2. Look at the object.
3. Look away from the object.

## Mixed Reality Toolkit Release Version
2017.2.1.4 and latest dev

davidkline-ms:
This was released in the 2017.4.0.0 release candidate

PR #2035, opened 2018-05-02
[May18] Cursor light fix (and other minor fixes)

keveleigh:
Overview
---
Some minor updates that were found while updating the Academy courses to the most recent Toolkit release.

Partner PR to #2036.

Changes
---
- Fixes: #1983.

PR #2036, opened 2018-05-02
[Dev] Cursor light fix (and other minor fixes)

keveleigh:
Overview
---
Some minor updates that were found while updating the Academy courses to the most recent Toolkit release.

Partner PR to #2035.

Changes
---
- Fixes: #1983.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Scripts/Billboard.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Scripts/Billboard.cs
@@ -30,6 +30,7 @@ namespace MixedRealityToolkit.Utilities
         /// The axis about which the object will rotate.
         /// </summary>
         [Tooltip("Specifies the axis about which the object will rotate.")]
+        [SerializeField]
         public PivotAxis PivotAxis = PivotAxis.XY;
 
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit/Utilities/Scripts/Billboard.cs
```diff
@@ -30,6 +30,7 @@ public class Billboard : MonoBehaviour
         /// The axis about which the object will rotate.
         /// </summary>
         [Tooltip("Specifies the axis about which the object will rotate.")]
+        [SerializeField]
         public PivotAxis PivotAxis = PivotAxis.XY;
```
StephenHodgson: public fields are already serialized, so this attribute is redundant, unless you plan to also make the field private per coding guidelines.

keveleigh: It's public here, but private in the master branch. Looks like there are probably some other changes to be made.

keveleigh: Changes from the original Dev -> Master PR have now been migrated.

Issue #1935, opened 2018-04-10
Headset Adjustment


Merglasch:
## Overview
I recently came back to this and wanted to use the Headset Adjustment scene. The scene and the screen space/camera canvas modes both don't work.
## Expected Behavior
HeadsetAdjustment should show a border on the camera on startup and load the next level on command.
## Actual Behavior
The scene shows nothing except the cursor (whyever there is even a cursor object in there). There is no border image in the entire project as far as I can see. Trying to use my own image with a screen space or camera space canvas doesn't work, as it is invisible for the hololenscamera. Also I cannot put any camera inside the camera field of the canvas. It just silently fails to do it and the field stays empty.

## Steps to reproduce
just start the HeadsetAdjustment scene and try.

## Unity Editor Version
2017.3.p3

## Mixed Reality Toolkit Release Version
2017.2.1.2 (But this scene doesn't seem to have changed since then?)

Merglasch:
tried again with the latest mrtk hotfix version. Now at least the camera assignment to a canvas seems to work.

davidkline-ms:
it appears that this scene lost it's display canvas sometime late 2017. I'll add a prefab that can be added to the camera (where it was in the earlier version of the scene).

davidkline-ms:
This was released in the 2017.4.0.0 release candidate

PR #2055, opened 2018-05-07
fix headset adjustment scene (#1935), add prefab

davidkline-ms:
At some point in late 2017, the camera object in the HeadsetAdjustment.unity scene lost the canvas that displayed the message and bounding box.

This change creates HeadsetAdjustmentScene.prefab and adds it as a child of the camera in the scene.
Fixes: #1935


davidkline-ms:
closing until this sudden appearance of Travis CI is cleared up

PR #1904, opened 2018-04-02
Async Building and Windows Device Portal Commands

StephenHodgson:
Overview
---
- Added a heavily edited version of the [Unity Async/Await Utilities](https://github.com/svermeulen/Unity3dAsyncAwaitUtil) by @svermeulen.  It has the same license type as the project (MIT)
- Fixed an issue where users were unable to build without either specifying a wsa certificate, or building through the normal Unity Editor Build window because the certificate had not been generated yet.  (Unity Build will generate one for you if one doesn't exist.)
- Adds async building of appx project. Typical Unity Player Building still blocks the editor, but APPX building and installing happens asynchronously.
- Did a bit of cleanup for coding guidelines and renamed a few classes for clarity.
- Added ability to build appx from command line argument.
- Added `EditorAssemblyReloadManager` to control reloading assemblies in the editor so running async tasks don't get cancelled prematurely.
- Added StartAsyncProcess extension methods.
- Added Rest Utililities.
- Updated data structures for Windows Device Portal.
- Fixes #1904 
- Moved the windows device portal classes out of the editor so any application can use it at runtime.
- Fixes #2124 and other conflicts with IL2CPP builds where the toolkit has implemented some apis from mscorlib.

TODO
---
- [x] Update Device Portal with new async Rest calls.
- [x] A bit of cleanup and renaming a few async methods with Async suffix.
- [x] Fix a issue with device connections when prev paired device is unreachable the string name is lost and it's impossible to select it again in the drop down.
- [x] Add all the documentation to the public PME's in the PR.
- [x] Fix an issue with CSRF Token refreshing.
- [x] Fix an issue with incrementing versions.
- [x] Compatibility with 2018.1
- [x] Make quick options HoloLens target make sure to target HoloLens only. #1959
- [x] Fixes #2119 

Breaking Changes
---
- Removed backwards compatibility to Visual Studio 2015.
- Changed command line tool hook for VSO build task.
- Removed unused `XDEGuestLocator.cs`

PR #2059, opened 2018-05-08
Mrtk v next async build tools device portal

StephenHodgson:
Overview
---
Ported Build tools, async/await, device portal features.


davidkline-ms:
How does this change relate to #1904?

StephenHodgson:
It's a port of the PR into vNEXT

SimonDarksideJ:
Missing menu actions for:

* Project Settings - Should be fit for purpose, even if some options not yet applied.
* Scene Settings - Good as a start point as we define what a scene looks like. (commenting out stuff we don't use yet)
* Others ?

Assets/MixedRealityToolkit/_Core/Extensions/ProcessExtensions.cs
```diff
@@ -0,0 +1,112 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#if UNITY_EDITOR || !UNITY_WSA
```
SimonDarksideJ: This isn't really needed unless you also want to prevent use at runtime (which we don't)
Can we avoid #if tags unless we really need them.

Assets/MixedRealityToolkit/_Core/Utilities/AsyncUtility/AwaitYieldInstructions/WaitForBackgroundThread.cs
```diff
@@ -0,0 +1,35 @@
+// MIT License
```
SimonDarksideJ: I thought we agreed to not use other License texts throughout the project.
@davidkline-ms ?

Assets/MixedRealityToolkit/_Core/Utilities/AsyncUtility/AwaitYieldInstructions/WaitForUpdate.cs
```diff
@@ -0,0 +1,35 @@
+// MIT License
```
SimonDarksideJ: And here @davidkline-ms

Assets/MixedRealityToolkit/_Core/Utilities/AsyncUtility/Internal/AsyncCoroutineRunner.cs
```diff
@@ -0,0 +1,62 @@
+// MIT License
```
SimonDarksideJ: And again?

Assets/MixedRealityToolkit/_Core/Utilities/AsyncUtility/Internal/SyncContextUtility.cs
```diff
@@ -0,0 +1,78 @@
+// MIT License
```
SimonDarksideJ: Gremlins?

Assets/MixedRealityToolkit/_Core/Utilities/AsyncUtility/License.md
```diff
@@ -0,0 +1,21 @@
+MIT License
```
SimonDarksideJ: Questionable.

Assets/MixedRealityToolkit/_Core/Utilities/ProcessResult.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Utilities
```
SimonDarksideJ: Should this be in definitions?  It's a Struct after all.

Assets/MixedRealityToolkit/_Core/Utilities/WebRequestRest/Response.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Utilities.WebRequestRest
```
SimonDarksideJ: Same again, should be in definitions.  In a subfolder if need be.

Assets/MixedRealityToolkit/_Core/BuildAndDeploy/BuildDeployWindow.cs
```diff
@@ -0,0 +1,1507 @@
...
+        #region Internal Types
+
+        private enum BuildDeployTab
+        {
+            UnityBuildOptions,
+            AppxBuildOptions,
+            DeployOptions
+        }
+
+        private enum BuildPlatformEnum
```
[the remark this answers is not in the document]
SimonDarksideJ: For the MVP, we aren't targeting any platforms that run on arm. So let's leave it off now and reevaluate when we need it.

StephenHodgson:
So I need clear action about the licensing changes.  What exactly do I need to do?

SimonDarksideJ:
Agreed, @davidkline-ms do we:

1: Use the MS License at the top of the files and include the license.txt from the source (or)
2: Use the source License at the top of each file and include the license.txt from the source 

I was under the impression @StephenHodgson that you had modified a lot of this code and improved it, hence changing it enough beyond the original license.

SimonDarksideJ:
But you should be able to move the other Structs / Classes and defs though while you wait @StephenHodgson  :D

davidkline-ms:
Re: License files

In general, my understanding is we need to include the license files (option 2) whenever we add an external project to MRTK and not remove existing headers.

In this case, the question is more, what should @StephenHodgson do wrt copyright notices in his changed files.  I can follow up with my PM on that if needed.

StephenHodgson:
>In general, my understanding is we need to include the license files (option 2) whenever we add an external project to MRTK and not remove existing headers.

Well we already satisfy this condition.

As for the changed files, I think it's fine to keep them the way they are. I don't see these files changing anymore unless there's bug fixes or underlying changes made by Unity.

Assets/MixedRealityToolkit/_Core/Utilities/WebRequestRest/Response.cs
StephenHodgson: I don't agree with moving these out of the related namespaces

Assets/MixedRealityToolkit/_Core/Utilities/ProcessResult.cs
StephenHodgson: Sure I'll move any structs in the utilities folder into definitions, but structs that are related specifically to a utility feature needs to stay together.

Assets/MixedRealityToolkit/_Core/Utilities/AsyncUtility/AwaitYieldInstructions/WaitForBackgroundThread.cs
StephenHodgson: I think adding the license headers makes it very clear, in case the files get moved in the future.

Assets/MixedRealityToolkit/_Core/BuildAndDeploy/BuildDeployPreferences.cs
```diff
@@ -0,0 +1,100 @@
...
+        private const string EditorPref_UseSSL = "_BuildDeployWindow_UseSSL";
+        private const string EditorPref_ProcessAll = "_BuildDeployWindow_ProcessAll";
+
+        public static string BuildDirectory
+        {
+            get { return EditorPrefsUtility.GetEditorPref(EditorPref_BuildDir, "UWP"); }
+            set { EditorPrefsUtility.SetEditorPref(EditorPref_BuildDir, value); }
+        }
+
+        public static string AbsoluteBuildDirectory
```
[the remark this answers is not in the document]
StephenHodgson: NO, it should be used to set whatever path you've got.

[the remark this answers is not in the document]
StephenHodgson: There's absolute no reason to override this.

StephenHodgson:
> - Project Settings - Should be fit for purpose, even if some options not yet applied.
> - Scene Settings - Good as a start point as we define what a scene looks like. (commenting out stuff we don't use yet)

That's not part of the build tools, so it wasn't added, nor should it be.  A lot of this will be replaced by the Mixed Reality Configuration Profiles we came up with.

Assets/MixedRealityToolkit/_Core/Extensions/ProcessExtensions.cs
StephenHodgson: I don't understand.  We do need them. This should be availible for any platform except Unity. Right now the only implementation is in the Editor.

StephenHodgson: > avoid #if tags unless we really need them.

I agree.

StephenHodgson: Why wasn't some of these concerns brought of in the initial PR?

Issue #2068, opened 2018-05-10
Warning cleanup "The private field is assigned but its value is never used"


davidkline-ms:
Identify and cleanup "The private field is assigned but its value is never used" warnings

davidkline-ms:
There are some in the SpectatorView feature that was introduced in the may18_dev branch.

keveleigh:
Found one more when building Standalone:

![image](https://user-images.githubusercontent.com/3580640/40802062-68e3b432-64c9-11e8-9b30-2fe9d6528ecf.png)

PR #2077, opened 2018-05-11
Addresses #MRTK-Unity:2068: private fields being assigned but not used

timGerken:
Overview
---

- Addresses #2068 - private fields being assigned but not used
- Some minor whitespace cleanup that should have happened in original PR

Changes
---

- wrapped recording related fields in #if Unity_IOS
- removed lastError altogether
- disabled compiler warning via pragma for fields with [SyncVar] or [SerializeField] attributes
- Fixes: #2068

### DIFF
--- a/Assets/HoloToolkit/SpectatorView/Scripts/Recording/ReplayKitRecorder.cs
+++ b/Assets/HoloToolkit/SpectatorView/Scripts/Recording/ReplayKitRecorder.cs
@@ -19,63 +19,64 @@ namespace HoloToolkit.Unity.SpectatorView
         /// <summary>
         /// Controls container gameObject
         /// </summary>
-        [Tooltip("Controls container gameObject")] 
+        [Tooltip("Controls container gameObject")]
         [SerializeField]
         private GameObject controls;
 
+#pragma warning disable 0414
         /// <summary>
         /// Seconds to countdown before recording
         /// </summary>
-        [Tooltip("Seconds to countdown before recording")] 
+        [Tooltip("Seconds to countdown before recording")]
         [SerializeField]
         private int countDownNumber = 3;
+#pragma warning restore 0414
 
-        /// <summary>
-        /// If an error ocurred, this variable will hold the last error message
-        /// </summary>
-        private string lastError = "";
-
+#if UNITY_IOS
         /// <summary>
         /// Is the component preparing for recording (Counting down)
         /// </summary>
         private bool preparingForRecording;
+#endif
 
         /// <summary>
         /// Record button gameObject
         /// </summary>
-        [Tooltip("Record button gameObject")] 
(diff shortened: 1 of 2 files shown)


Assets/HoloToolkit/SpectatorView/Scripts/Recording/ReplayKitRecorder.cs
```diff
@@ -19,63 +19,64 @@ public class ReplayKitRecorder : MonoBehaviour
...
 
+#pragma warning disable 0414
         /// <summary>
         /// Seconds to countdown before recording
         /// </summary>
-        [Tooltip("Seconds to countdown before recording")] 
+        [Tooltip("Seconds to countdown before recording")]
         [SerializeField]
         private int countDownNumber = 3;
+#pragma warning restore 0414
```
StephenHodgson: I think using pragma keywords aren't the best solution.

timGerken: do you have a better one?

timGerken: yes, but the countdown delay is/can be set in the editor. I can wrap it in an #if UNITY_IOS block like the other updates, but I though it was more important for the inspector's view of the app to look as close as possible to each other.

keveleigh: Causing it to disappear in the Inspector for different build types can also cause lost prefab values as the build types are switched back and forth.

StephenHodgson:
@davidkline-ms why was this merged? I was hoping to get more feedback on how to solve this without disabling the warnings.

davidkline-ms:
@StephenHodgson, I merged based on the commentary between tim and kurtis (especially kurtis' comment regarding lost prefab values).

We could, if you believe strongly, open another issue to remove the pragma and investigate a better #if solution.

Assets/HoloToolkit/SpectatorView/Scripts/Recording/ReplayKitRecorder.cs
StephenHodgson: I understand that, but we don't do this in other places where we have the same problem.

keveleigh: Do you have an example of how it was fixed in another place that would solve this situation? 
If we have cases where things are disappearing from the Inspector, we should reevaluate what we did there. If we have cases where this was solved without `#pragma` and without things disappearing from the Inspector, we should probably put a note in the coding guidelines letting people know our expectations.

Issue #2009, opened 2018-04-24
[2017.2.1.4 Patch Release] Hololens. BoundingBoxGizmoExample. Any scene selections cause NullReferenceException


jchristof:
## Overview
Tested in Hololens emulator
[2017.2.1.4 Patch Release] Hololens. BoundingBoxGizmoExample. Clicks on objects or empty scene space cause exceptions
## Expected Behavior
Selecting an object shows a bounding box with app bar
## Actual Behavior
eventData.selectedObject produces NullReferenceException
```
public void OnInputClicked(InputClickedEventData eventData)
        {
            if (Isinteractable(eventData.selectedObject))
            {
                InputClicked(eventData.selectedObject, eventData);
            }
        }
```
## Steps to reproduce
New project. Load BoundingBoxGizmoExample scene. Build for Hololens. Click anywhere in the scene and observe the output window.
## Unity Editor Version
Unity_2017_2.1p2
## Mixed Reality Toolkit Release Version
[2017.2.1.4 Patch Release]

davidkline-ms:
@jchristof, What version of the HoloLens OS are you running?

Thx!

jchristof:
@davidkline-ms Hi David:
Windows version	14393.1358.x86fre.rs1_release.170602-2252

cre8ivepark:
@jchristof Does it only happen in the emulator?

davidkline-ms:
Assigning to @cre8ivepark

davidkline-ms:
@jchristof, is the code snippet you provided something you added to a script?
Thanks!

jchristof:
Have not tried it on device.

jchristof:
No not my script it came from a clean install of the toolkit and example
project package.

davidkline-ms:
found it. thanks! i'll was not seeing an exception on my flighted HoloLens. I will continue to investigate on one with the version you mentioned.

jchristof:
Thanks guys!

davidkline-ms:
I have not been able to repro this using the emulator. I recall fixing this issue for the 2017.2.1.4 release...

@jchristof, if you re-download the 2017.2.1.4 package, or try the may18_dev branch and can still reproduce the issue, please reopen it.

Thanks!
David

davidkline-ms:
I stand corrected.... I just saw the exception.

davidkline-ms:
my exception is occurring in OnSourceLost, but may be related to this.

[...]

PR #2086, opened 2018-05-11
address BoundingBoxRigHandle nullreferenceexception

davidkline-ms:
BoundingBoxRigHandle has a potential to encounter a NullReferenceException in OnSourceLost. This change avoids the exception as well as moving the call to eventData.Use inside the if block.

This MAY be related to what was reported in #2009. I was unable to reproduce the exact issue when I encountered this one.

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBoxGizmoHandle.cs
+++ b/Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBoxGizmoHandle.cs
@@ -370,7 +370,8 @@ namespace HoloToolkit.Unity.UX
         }
         public void OnSourceLost(SourceStateEventData eventData)
         {
-            if (eventData.SourceId == inputDownEventData.SourceId)
+            if ((inputDownEventData != null) &&
+                (eventData.SourceId == inputDownEventData.SourceId))
             {
                 inputDownEventData = null;
                 ResetRigHandles();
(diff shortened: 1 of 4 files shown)


davidkline-ms:
@cre8ivepark is likely the best to answer the question regarding inputDownEventData.

davidkline-ms:
FYI: OnInputUp is setting inputDownEventData to null, so this exception will happen if a source is lost if not in a down state.

Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBoxGizmoHandle.cs
```diff
@@ -368,14 +368,17 @@ public void OnInputUp(InputEventData eventData)
         public void OnSourceDetected(SourceStateEventData eventData)
         {
         }
+
         public void OnSourceLost(SourceStateEventData eventData)
         {
-            if (eventData.SourceId == inputDownEventData.SourceId)
+            if ((inputDownEventData != null) &&
```
StephenHodgson: event data should never be null.  If it is then, something is being used improperly in another location.

davidkline-ms:
To reiterate, the change in this PR is to match the Update method and check to see if the **cached** inputDownEventData is null before attempting to dereference.

Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBoxGizmoHandle.cs
StephenHodgson: I believe it is wrong, but we can address a fix in vNEXT

davidkline-ms:
this PR got messed up due to me trying to keep it in sync with may18_dev.... going to re-branch and re-submit.

Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBoxGizmoHandle.cs
StephenHodgson: This might be related to those null references in the InputManger you were getting.

events should _never_ be null

PR #2092, opened 2018-05-14
MRTK - vNext - Windows Mixed Reality Device Support

SimonDarksideJ:
Overview
---
Adding the initial Windows MR Device and Controller definitions, including some refactoring of core / input to accommodate.

Changes
---
 - Adds new WMRDevice (temp name, will rename for merge)
-  Adds new Windows Mixed Reality Controller input source
 - Adds new InputDatum classes for managing input data from device.
 - Changed Controller struct to an Interface for better management

TODO
---

- [ ] Add events for buttons / controllers / pointers / grip
- [ ] Consolidate input events raised
- [ ] Rationalise InputTypeDatum to meet current requirements (some old data taken is no longer needed)
- [ ] Review 6DoF implementation
- [ ] Test with fake controllers in scene
- [ ] Wire up pointers -> @StephenHodgson 
- [ ] Rename Device name based on popular demand
- [ ] Part the red sea 🌊 

Notes
---
Focusing on the "Making it work" for the moment.  Documentation and rationalization of the new functionality will happen before the final merge.
So please restrict comments to architecture / design for the moment please.


Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/Raw.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputTypes
+{
+    public struct Raw<T> : IMixedRealityInputTypeDatum<T>
+    {
+        public bool IsSupported { get; }
```
davidkline-ms: need doc comments

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/SingleAxis.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputTypes
+{
+    public struct SingleAxis : IMixedRealityInputTypeDatum<float>
```
davidkline-ms: SingleAxisInputType?

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/SingleAxis.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public bool IsSupported { get; }
```
davidkline-ms: need doc comments

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/SixDoF.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+using System;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputTypes
+{
+    public struct SixDoF : IMixedRealityInputTypeDatum<Tuple<Vector3, Quaternion>>
```
davidkline-ms: SixDegreesOfFreedomInputType?  Abbreviations in type name aren't much cared for in .net

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/SixDoF.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            IsSupported = true;
+            IsAvailable = true;
+            Pressed = false;
+            Touched = false;
+            Changed = false;
+            Value = startPosition;
+        }
+
+        public bool IsSupported { get; }
```
davidkline-ms: need doc comments

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/ThreeDoF.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputTypes
+{
+    public struct ThreeDoF : IMixedRealityInputTypeDatum<Quaternion>
```
davidkline-ms: ThreeDegreesOfFreedomInputType?

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/ThreeDoF.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public bool IsSupported { get; }
```
davidkline-ms: need doc comments

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -0,0 +1,98 @@
...
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions;
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.InputTypes;
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+using System.Collections;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Devices.WindowsMixedReality
+{
+    public struct WindowsMixedRealityController : IMixedRealityController<SixDoF>
```
davidkline-ms: need doc comments

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -0,0 +1,98 @@
...
+        public ControllerState ControllerState { get; set; }
+
+        public InteractionDefinition[] Interactions { get; set; }
+
+        public GameObject ControllerModel { get; set; }
+
+        //TODO - Needs re-implementing, as there may be more than left == right
+        #region IEquality Implementation
+
+        public static bool Equals(IMixedRealityInputSource left, IMixedRealityInputSource right)
```
davidkline-ms: re: above todo comment... maybe use names other than left && right

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            return obj.GetHashCode();
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                int hashCode = 0;
+                hashCode = (hashCode * 397) ^ (int)SourceId;
```
davidkline-ms: where did the 397 come from?

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
```diff
@@ -1,48 +1,49 @@
...
         /// </summary>
-        public string Id { get; set; }
+        Handedness Handedness { get; }
 
         /// <summary>
-        /// The designated hand that the controller is managing, as defined by the SDK / Unity.
+        /// Outputs the current position of the controller, as defined by the SDK / Unity.
         /// </summary>
-        public Handedness Handedness { get; set; }
+        T Controller { get; }
```
davidkline-ms: ControllerPosition

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
```diff
@@ -1,48 +1,49 @@
...
 
         /// <summary>
-        /// Outputs the current position of the controller, as defined by the SDK / Unity.
+        /// Outputs the current Pointer position of the controller, as defined by the SDK / Unity.
         /// </summary>
-        public Vector3 Position { get; set; }
+        T Pointer { get; }
```
davidkline-ms: PointerPosition

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
```diff
@@ -1,48 +1,49 @@
...
 
         /// <summary>
-        /// Outputs the current rotation of the controller, as defined by the SDK / Unity.
+        /// Outputs the current Grip position of the controller, as defined by the SDK / Unity.
```
davidkline-ms: GripPosition

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityDevice.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
+{
+    public interface IMixedRealityDevice
```
davidkline-ms: needs doc comments

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputTypeDatum.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem
+{
+    public interface IMixedRealityInputTypeDatum<TReading>
```
davidkline-ms: need doc comments

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -142,13 +138,23 @@ private void Initialize()
                 AddManager(manager.Key, manager.Value);
             }
 
+            #endregion Managers Initialization
```
davidkline-ms: Registration

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -142,13 +138,23 @@ private void Initialize()
...
+            #region Managers Initialization
+
             //Initialize all managers
             foreach (var manager in ActiveProfile.ActiveManagers)
             {
                 manager.Value.Initialize();
             }
 
-            #endregion Managers Initialization
+            #endregion
```
davidkline-ms: add description of the region (like above)

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/Digital.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputTypes
+{
+    public struct Digital : IMixedRealityInputTypeDatum<bool>
```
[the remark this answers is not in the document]
SimonDarksideJ: A possible contender.  
These datums are needed to store the various data types used by Input in a uniformed (multi-platform) manner

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/Digital.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public bool IsSupported { get; }
```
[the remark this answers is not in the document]
SimonDarksideJ: Rofl, they are needed everywhere in these updates.  First Rule of testing, make sure it tests!

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/DualAxis.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputTypes
+{
+    public struct DualAxis : IMixedRealityInputTypeDatum<Vector2>
```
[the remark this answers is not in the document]
SimonDarksideJ: Considering

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/None.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputTypes
+{
+    public struct None : IMixedRealityInputTypeDatum<bool>
```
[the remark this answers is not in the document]
SimonDarksideJ: It's taken from the Enum atm, not sure it's needed yet but that will come from actual testing.

Assets/MixedRealityToolkit/_Core/Definitions/InputTypeDatum/DualAxis.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public bool IsSupported { get; }
```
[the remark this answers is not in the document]
SimonDarksideJ: Like EVERYWHERE :D

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
SimonDarksideJ: ??  It's the position of the Grip, same as Pointer and the base Controller.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityDevice.cs
SimonDarksideJ: yup

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputTypeDatum.cs
SimonDarksideJ: agreed

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: Fair enough

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: OK

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputTypeDatum.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        bool IsSupported { get; }
+        bool IsAvailable { get; }
+        bool Pressed { get; }
+        bool Touched { get; }
```
StephenHodgson: What about inputs that don't support touch?

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputTypeDatum.cs
StephenHodgson: I think this interface is somewhat redundant. Input Events and the capabilities will determine what is availible and when.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityDevice.cs
StephenHodgson: Seems to be the same as other interfaces we've made.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
```diff
@@ -1,48 +1,49 @@
...
         /// </summary>
-        public Quaternion Rotation { get; set; }
+        T Grip { get;}
```
StephenHodgson: Not all controllers will have grip.  Not a fan of the changes to this class. It's not as generic as it used to be.

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -23,5 +27,93 @@ public struct InteractionDefinition
...
+        {
+            switch (AxisType)
+            {
+                case AxisType.Raw:
+                    Changed = value == rawDatum;
+                    rawDatum = value;
+                    break;
+                case AxisType.Digital:
+                    bool newBoolValue;
+                    Assert.IsTrue(bool.TryParse(value.ToString(), out newBoolValue));
```
StephenHodgson: Should probably use the `UnityEngine.Debug.Assert` if this isn't a NUnit test class

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -81,6 +81,32 @@ public bool EnableControllers
...
+            get { return leftControllerModel; }
+            private set { leftControllerModel = value; }
+        }
+
+        [SerializeField]
+        [Tooltip("Default Right Controller Model")]
+        private GameObject rightControllerModel;
+
+        /// <summary>
+        /// Enable and configure the controller rendering for the Mixed Reality Toolkit
```
StephenHodgson: should the tooltip info and summary match?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
```diff
@@ -0,0 +1,489 @@
...
+using UnityEngine;
+using UnityEngine.XR.WSA.Input;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Devices.WindowsMixedReality
+{
+    public class WMRDevice : IMixedRealityDevice
+    {
+        private readonly Dictionary<uint, IMixedRealityInputSource> activeControllers = new Dictionary<uint, IMixedRealityInputSource>();
+        Vector3 controllerPosition, pointerPosition, gripPosition = Vector3.zero;
+        Quaternion controllerRotation, pointerRotation, gripRotation = Quaternion.identity;
```
StephenHodgson: nit, let's declare our fields on separate lines, and use access modifiers.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
```diff
@@ -0,0 +1,489 @@
...
+        public void Initialize()
+        {
+            inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>();
+
+            InitializeSources();
+        }
+
+        private void InitializeSources()
+        {
+            InteractionManager.InteractionSourceDetected += InteractionManager_InteractionSourceDetected;
```
StephenHodgson: I'd like to use the actual `Windows.Spatial.UI` namespace to get controller data.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
StephenHodgson: the number can/should be random. This number is used to salt the hash number.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityDevice.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+using System.Collections.Generic;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
+{
+    public interface IMixedRealityDevice
+    {
+        void Initialize();
```
StephenHodgson: We seem to use these interface methods a lot, should we make a base base interface for basic methods like this we keep reusing?

ie. Initialize, Enable, Disable, Destory, etc?

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
```diff
@@ -12,17 +12,41 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem
     /// </summary>
     public interface IMixedRealityInputSource : IEqualityComparer
     {
+        /// <summary>
+        /// The ID assigned to the Controller
+        /// </summary>
         uint SourceId { get; }
 
+        /// <summary>
+        /// The Name assigned to the Controller
```
StephenHodgson: Not event a controller, but the source itself. Remember an input source is more than just a controller, it could be a mouse, or a keyboard as well.

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
```diff
@@ -12,17 +12,41 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem
...
+        /// </summary>
         string SourceName { get; }
 
+        /// <summary>
+        /// Outputs the current state of the controller, whether it is tracked or not. As defined by the SDK / Unity.
+        /// </summary>
+        ControllerState ControllerState { get; }
```
StephenHodgson: Let's rename to SourceState or similar.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -78,19 +78,14 @@ private void ResetConfiguration()
...
         ///// The Active Headset property maintains the Headsets/SDK detected by the Mixed Reality manager on startup
         ///// </summary>
-        //[SerializeField]
-        //[Tooltip("The currently active / detected Headset or SDK")]
-        //private Headset activeHeadset = default(Headset);
+        [SerializeField]
+        [Tooltip("The currently active / detected Headset or SDK")]
+        private IMixedRealityDevice activeDevice = default(IMixedRealityDevice);
+
+        public IMixedRealityDevice ActiveDevice { get { return activeDevice; } }
```
StephenHodgson: I like this a bit better, but wondering if it'll limit us to how we use the controllers.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
SimonDarksideJ: I'll fix those later, with so much change, I want to make it "work" (tm) first.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
SimonDarksideJ: tbc, when it works :P

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
SimonDarksideJ: Windows.UI.Input.Spatial.SpatialInteractionManager
Although I'm concerned about the "GetForCurrentView" call, as Unity may interfere.  In my experience, such a call can only work in the UWP app.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityDevice.cs
SimonDarksideJ: Will review once we get further.

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
SimonDarksideJ: Ahh, forgot to update the descriptions after moving them. Will fix that as it'll easily be missed

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
SimonDarksideJ: Agreed

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: We can review as we progress, as it's the Interface it implements, we have a bit more freedom. If it was a concrete class we would have issue.

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityConfigurationProfile.asset
```diff
@@ -16,4 +16,8 @@ MonoBehaviour:
     classReference: Microsoft.MixedReality.Toolkit.InputSystem.MixedRealityInputManager,
       Microsoft.MixedReality.Toolkit.InputSystem
   enableControllers: 1
+  leftControllerModel: {fileID: 1774971220009634, guid: 89aca206fbe292649bca7605a95fa801,
```
davidkline-ms: please forgive if this is a dumb question... are we moving away from getting the model from the controller firmware?

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
SimonDarksideJ: Used the same method as used in other parts of the toolkit. Will review and update.

I don't particularly like this approach, but it will do until "I can find something better"

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
SimonDarksideJ: Didn't work in the Unity project.  Although I could type and have the code recognised.  It caused errors in the Unity project.

Maybe there is some magic to have this only recognised in the Player project (and ignored in the main one?)  /cc @StephenHodgson

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
davidkline-ms: Werent we going to have Assert wrappers? Then we could have Unity and runtime asserts.... Please.

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
SimonDarksideJ: fixed the interface descriptions

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
SimonDarksideJ: Renamed to InputSourceState, including enum

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
davidkline-ms: The tooltip should be a subset of the summary tag, up to and including a subset of 100% :)

That said, I am fine with Simon doing a doc comment pass as a separate issue, if that works best.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
```diff
@@ -35,6 +40,11 @@ public void Initialize()
 
         private void InitializeSources()
         {
+            //Tried but it fails to build and causes errors in Unity :S (recognized in the Player project)
```
davidkline-ms: Am i reading correctly that this compiles in VS but not in the Unity Editor?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
```diff
@@ -50,10 +60,25 @@ private void InitializeSources()
             }
         }
 
+        //TODO - kept for reference - clean later.
+        //private void SpatialManager_SourcePressed(Windows.UI.Input.Spatial.SpatialInteractionManager sender, Windows.UI.Input.Spatial.SpatialInteractionSourceEventArgs args)
```
davidkline-ms: Future note: We may need to confirm the API contract / OS version to ensure that functionality is present. I can work with you on that if needed.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
```diff
@@ -246,12 +271,18 @@ private void UpdateInteractionSource(InteractionSourceState interactionSourceSta
...
+            //TODO - Review, as we no longer have a MixedRealityCameraParent, this will cause issue.
             if (CameraCache.Main.transform.parent != null)
             {
                 gripPosition = CameraCache.Main.transform.parent.TransformPoint(gripPosition);
                 gripRotation.eulerAngles = CameraCache.Main.transform.parent.TransformDirection(gripRotation.eulerAngles);
             }
             sourceData.Grip.Update(gripPosition, gripRotation);
+
+            // Update Touchpad
+            if (interactionSourceState.touchpadTouched)
```
davidkline-ms: Note: Touchpad (also, Grip, Thumbstick, Menu) are features that may need an API contract / os version check

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityConfigurationProfile.asset
StephenHodgson: That's correct we will be loading the model from the driver, but we still want to give devs the option to use their own custom model if they so choose.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
StephenHodgson: Most likely from an open bug I have reported about the custom assembly definitions

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
StephenHodgson: I _really_ *really* disagree with wrapping asserts and debug logs.

StephenHodgson: In either case, I was just trying to point out that Simon used the wrong namespace for the asserts.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
StephenHodgson: Good thing we already have a contract checker class.  @davidkline-ms do you know if that contract checking uses reflection or something like that under the hood?

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
SimonDarksideJ: Will review.  ANd I know you are against Wrappers @StephenHodgson but it's much better to control that from a Single class and use our own implementation, rather than spat Unity stuff everywhere as it leaves us open to issues if Unity introduce a breaking change.  I prefer to be in control, not unity.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
SimonDarksideJ: Again, we'll refine once we have an MVP.  The MVP is classes as a single API version with WMR controllers.  Then we can extend to support the next SDK/Controller - OpenVR

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
StephenHodgson: Well unfortunately that impacts others who use the toolkit.  I don't like when dependencies change the way logs and assertions are handled.  For example, we have tools and things we use that are directly impacted by a change like this.

StephenHodgson: We're using Unity, do it the Unity way. Accept the breaking changes when they come. Not _if_ they come.

SimonDarksideJ: On the namespace thing.  It IS using the UnityEngine.Debug.Assert. in the same way you do in the Input manager :D

P.S. Wrappers would make this MORE clear :D

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
```diff
@@ -88,47 +88,99 @@ private IMixedRealityInputSource GetOrAddWindowsMixedRealityController(Interacti
 
             bool controllerTracked = ReadMixedRealityControllerValues(ref interactionSourceState);
 
-            // Define Controller capabilities
-            var controllerCapabilities = new List<InputType>();
+            var DetectedController = new WindowsMixedRealityController()
+            {
+                SourceId = interactionSourceState.source.id,
+                Handedness = interactionSourceState.source.handedness == InteractionSourceHandedness.Left ? Handedness.Left : Handedness.Right,
```
StephenHodgson: How do we handle hands or controllers that done have any handedness, like the clicker?

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
```diff
@@ -19,12 +19,34 @@ public class BaseGenericInputSource : IMixedRealityInputSource
...
+            SourceName = name;
+            Pointers = pointers ?? new[] { GazeProvider.GazePointer };
+            Interactions = new Dictionary<InputType, InteractionDefinition>();
+            foreach (var interaction in interactions)
+            {
+                Interactions.Add(interaction.InputType, interaction);
+            }
+        }
+
         public BaseGenericInputSource(string name, InputType[] capabilities, IMixedRealityPointer[] pointers = null)
```
StephenHodgson: Maybe we should get rid of the capabilities all together?

Assets/MixedRealityToolkit-SDK/Scripts/ControllerVisualizer.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.MixedReality.Toolkit.Internal.Utilities;
+using UnityEngine;
+using UnityEngine.XR.WSA.Input;
+
+namespace Microsoft.MixedReality.Toolkit.SDK
+{
+    public class ControllerVisualizer : MonoBehaviour
+    {
+        GameObject leftController;
+        GameObject rightController;
```
StephenHodgson: Maybe this class should be a single instance on any controller being visualized?

Assets/MixedRealityToolkit/_Core/Definitions/AxisType.cs
```diff
@@ -30,10 +30,18 @@ public enum AxisType
         /// </summary>
         DualAxis,
         /// <summary>
-        /// Dual Axis analogue input.
+        /// Position OR Rotation Axis analogue input. <-- Experiment
         /// </summary>
         ThreeDoF,
```
StephenHodgson: Let's go ahead and remove this one.

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -1,6 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
+using UnityEngine;
+using UnityEngine.Assertions;
```
StephenHodgson: Using wrong namespace for assertions

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -23,5 +30,141 @@ public struct InteractionDefinition
...
         public InputType InputType { get; set; }
+
+        #endregion Interaction Properties
+
+        #region Definition Datums
+
+        /// <summary>
+        /// The data storage for a Raw / None Axis type.
+        /// </summary>
+        private object rawDatum { get; set; }
```
StephenHodgson: Let's drop the datum suffix on these properties and use data instead. Datum is plural for data and we only have a single item.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
SimonDarksideJ: Well, we have Handedness.None for those, or we can re-think.

I'm also considering restructuring the relationship between "device" and "Controllers" as I'm starting to feel we need a separate "detection" logic for controllers.


Something less that the forced approach of VRTK and definitely separate to the "all in one" approach in current HoloToolkit.

But want to formalise these first controllers, first.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
SimonDarksideJ: Already done and updated the InputSources you already have after I pulled your patest changes.

Updated the Capabilities logic to use the InteractionDefinitions dictionay, which is more efficient.

But I seriously need to rethink the "Supports Capability" logic, as it returns true if ONE of the required capabilities is valid, which could cause issue.

Assets/MixedRealityToolkit-SDK/Scripts/ControllerVisualizer.cs
SimonDarksideJ: This is a pure test only (as highlighted in the commit), just a playgound while I test access.

The method here is already wrong and I need it to infer from the InputSystem, not direct from the controller.
Ulitmately though, both direct and input system will be required to cover advanced cases.

Assets/MixedRealityToolkit/_Core/Definitions/AxisType.cs
SimonDarksideJ: Rofl, I was going to remove the two new ones as I found a way around it following our discussion.

So, which method do you prefer??

* A single 3DoF
* Separate Position and Rotation 3DoF

#Fight

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
SimonDarksideJ: Fair enough, got to get our connotations right.

Assets/MixedRealityToolkit/_Core/Definitions/AxisType.cs
StephenHodgson: > Separate Position and Rotation 3DoF

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
StephenHodgson: Should be `UnityEngine.Debug`.

That class you have in there is reserved for testing

StephenHodgson:
One thing I want us to keep in mind, is that the interaction manager from Unity handles more than just the WMR Controllers, but also the clicker, and hands.  We're not taking these into account.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
```diff
@@ -8,6 +8,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.XR.WSA.Input;
```
StephenHodgson: Why?

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
```diff
@@ -33,24 +56,42 @@ public BaseGenericInputSource(string name, InputType[] capabilities, IMixedReali
...
+
+        public Handedness Handedness { get; } = Handedness.None;
 
-        public bool SupportsInputCapability(InputType[] capabilities)
+        public Dictionary<InputType, InteractionDefinition> Interactions { get; }
+
+
+        public virtual void SetupInputSource(InteractionSourceState interactionSourceState) { }
+
+        public virtual void UpdateInputSource(InteractionSourceState interactionSourceState) { }
```
StephenHodgson: Generic input shouldn't care, or know about WSA.XR inputs.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseInputSource.cs
```diff
@@ -2,10 +2,12 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Collections;
+using System.Collections.Generic;
 using Microsoft.MixedReality.Toolkit.Internal.Definitions;
 using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
 using Microsoft.MixedReality.Toolkit.Internal.Managers;
 using UnityEngine;
+using UnityEngine.XR.WSA.Input;
```
StephenHodgson: Same here, we shouldn't be including this in base input sources.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
SimonDarksideJ: Temp only. Need to separate device state to Mrtk state (so) but was to lazy to do it atm 😎

Assets/MixedRealityToolkit/InputSystem/Sources/BaseInputSource.cs
SimonDarksideJ: See above

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
SimonDarksideJ: Same as above, wip

SimonDarksideJ:
> One thing I want us to keep in mind, is that the interaction manager from Unity handles more than just the WMR Controllers, but also the clicker, and hands. We're not taking these into account.

It's getting there, all part of the larger refactoring work. 

There will be a "clickerInputSource", "XboxControllerInputSource", "JoystickInputSource", etc. Each will have their own config profile which will be picked up by whatever device is running. (see Slack brain dump)

Still heavily WIP, but I want to hammer it out the next few days. 

Main area I'll need help with is pointer registration, as it's too tightly coupled atm and the pointer code should be abstracted out of the base input.

Thinking :

1. Device discovery
2. Config registration 
3. Controller initialization
4. Pointer registration and coupling

Each part being done by its own discrete function code to avoid being hard bound.  Pointer code should create pointers (based on the controller make up). And then be attached to the controller. Instead of all being in the I put source atm.

SimonDarksideJ:
Just spotted the issues /pr's being addressed today about grip /pointer offsets. Is there something else we need to take account of?

Assets/MixedRealityToolkit/InputSystem/Sources/TouchscreenInputSource.cs
```diff
@@ -23,7 +23,7 @@ public class TouchscreenInputSource : BaseInputSource
     {
         private const float K_CONTACT_EPSILON = 2.0f / 60.0f;
 
-        public override InputType[] Capabilities => new[] { InputType.Pointer, InputType.PointerPosition };
+        public override Dictionary<InputType, InteractionDefinition> Interactions => new Dictionary<InputType, InteractionDefinition>() { { InputType.Pointer, new InteractionDefinition() { InputType = InputType.Pointer, AxisType = AxisType.DualAxis, Changed = false, Id = InputSystem.GenerateNewSourceId().ToString() } } };
```
StephenHodgson: Why are we setting the interaction definitions id to a generated id for an input source from the manager?

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
```diff
@@ -1,32 +1,78 @@
...
+        /// The list of Pointers attached to the Input Source
+        /// </summary>
         IMixedRealityPointer[] Pointers { get; }
 
         /// <summary>
-        /// Returns the input info that the input source can provide.
+        /// Details the list of available buttons / interactions available from the Input Source.
         /// </summary>
-        InputType[] Capabilities { get; }
+        Dictionary<InputType,InteractionDefinition> Interactions { get; }
```
StephenHodgson: As much as I like this new patter, I feel like it's a bit redundant to have the input type both as a key and in the interaction definition itself.

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
```diff
@@ -1,32 +1,78 @@
...
+        /// </summary>
+        /// <param name="interactionSourceState">SDK data to inform construction</param>
+        void SetupInputSource(UnityEngine.XR.WSA.Input.InteractionSourceState interactionSourceState);
+
+        // TODO - Need to generalize the Input Type?
+        /// <summary>
+        /// Update function to update controller
+        /// </summary>
+        /// <param name="interactionSourceState">SDK data to inform update</param>
+        void UpdateInputSource(UnityEngine.XR.WSA.Input.InteractionSourceState interactionSourceState);
```
StephenHodgson: We shouldn't need to call update on input sources.  I'd rather let devs get the raw data at the time they need it instead of polling on updates.

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
```diff
@@ -1,32 +1,78 @@
...
-        InputType[] Capabilities { get; }
+        Dictionary<InputType,InteractionDefinition> Interactions { get; }
 
+        // TODO - Need to generalize the Input Type?
         /// <summary>
-        /// Returns whether the input source supports the specified input types.
+        /// Setup function to define controller
+        /// </summary>
+        /// <param name="interactionSourceState">SDK data to inform construction</param>
+        void SetupInputSource(UnityEngine.XR.WSA.Input.InteractionSourceState interactionSourceState);
```
StephenHodgson: Also, let's get rid of this XR.WSA reference. Generic Base classes shouldn't reference specific platform types.

StephenHodgson:
> Main area I'll need help with is pointer registration, as it's too tightly coupled atm and the pointer code should be abstracted out of the base input. 

I disagree here.  We will always have a pointer of some sort.

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -9,6 +13,9 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
     /// </summary>
     public struct InteractionDefinition
     {
+
+        #region Interaction Properties
+
         /// <summary>
         /// The ID assigned to the Input
```
StephenHodgson: Not sure that I agree with the ID being a string. Let's follow the pattern of the rest of the ids we keep as uints

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
SimonDarksideJ: There is that thought, but it happens when it's working in both roles. Will see if I can remove the duplication

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
SimonDarksideJ: It means the controller does the lifting of translating raw input in to its data classes. That would be static for each controller and shouldn't really be touched (it's particular to that controller)

The Dev should be more interested with what they then do with that data.

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
SimonDarksideJ: Agreed, this was just being lazy as we need an internal state definition. But sure I my mind how that looks yet

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
SimonDarksideJ: Agreed, I think it was an oversight

SimonDarksideJ:
> I disagree here. We will always have a pointer of some sort for each input source

Not saying that. I just mean that the code to generate a pointer should be separate, so you can attach it to any controller

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -23,5 +30,141 @@ public struct InteractionDefinition
...
+        }
+        #endregion Get Operators
+
+        #region Set Operators
+
+        public void SetValue(object newValue)
+        {
+            if (AxisType == AxisType.Digital)
+            {
+                Changed = newValue == rawData;
```
StephenHodgson: shouldn't this be `!=`?

StephenHodgson:
> I just mean that the code to generate a pointer should be separate, so you can attach it to any controller

Okay, that makes sense.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
StephenHodgson: Fixed the support capabilities in my pr

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
```diff
@@ -1,32 +1,83 @@
...
+        /// <summary>
+        /// The list of Pointers attached to the Input Source
+        /// </summary>
         IMixedRealityPointer[] Pointers { get; }
 
         /// <summary>
-        /// Returns the input info that the input source can provide.
+        /// Details the list of available buttons / interactions available from the Input Source.
+        /// </summary>
+        Dictionary<InputType,InteractionDefinition> Interactions { get; }
```
StephenHodgson: Let's replace input type key with the Input action we added or else we can only define a single input type, when we need to do multiple.

StephenHodgson:
Not so sure about the InputActions. They're a mirror of the InputType.  Thinking that an InputAction could be a struct with an id, and a custom name that users can set.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
```diff
@@ -31,26 +35,45 @@ public BaseGenericInputSource(string name, InputType[] capabilities, IMixedReali
 
         public string SourceName { get; }
 
+        public InputSourceState InputSourceState { get; }
+
+        public Handedness Handedness { get; }
```
StephenHodgson: This shouldn't be in the base generic class, but instead in the controller interface.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseInputSource.cs
```diff
@@ -37,26 +39,45 @@ public virtual string SourceName
...
+
+        public virtual IMixedRealityPointer[] Pointers { get; private set; }
+
+        public virtual Dictionary<InputType, InteractionDefinition> Interactions { get; }
+
+        public virtual InputType[] Capabilities { get; }
+
+        public void SetupInputSource<T>(T state) { }
 
-        public virtual InputType[] Capabilities => new[] { InputType.None };
```
StephenHodgson: Please revert.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseInputSource.cs
```diff
@@ -37,26 +39,45 @@ public virtual string SourceName
             set { name = value; }
         }
 
-        public virtual IMixedRealityPointer[] Pointers => null;
+        public virtual InputSourceState InputSourceState { get; }
+
+        public virtual Handedness Handedness { get; }
```
StephenHodgson: Base input source shouldn't know about handedness

Assets/MixedRealityToolkit/InputSystem/Sources/DictationInputManager.cs
```diff
@@ -67,6 +68,7 @@ public class DictationInputManager : BaseInputSource
 
         public override InputType[] Capabilities => new[] { InputType.Voice };
 
+
```
StephenHodgson: We can probably revert any changes in this class.

Assets/MixedRealityToolkit/InputSystem/Sources/SpeechInputSource.cs
```diff
@@ -53,6 +54,7 @@ private enum RecognizerStartBehavior { AutoStart, ManualStart }
 
         public override InputType[] Capabilities => new[] { InputType.Voice };
 
+
```
StephenHodgson: same here.

Assets/MixedRealityToolkit/_Core/Definitions/ControllerState.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
     /// The Controller State defines whether a controller or headset is currently being tracker or not.
     /// This enables developers to be able to handle non-tracked situations and react accordingly
     /// </summary>
-    public enum ControllerState
+    public enum InputSourceState
```
StephenHodgson: Why rename?

Assets/MixedRealityToolkit/_Core/Definitions/Headset.cs
```diff
@@ -33,6 +33,6 @@ public struct Headset
         /// <summary>
         /// Outputs the current state of the headset, whether it is tracked or not. As defined by the SDK / Unity.
         /// </summary>
-        public ControllerState ControllerState { get; set; }
+        public InputSourceState ControllerState { get; set; }
```
StephenHodgson: But then not rename the property?

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -22,6 +36,146 @@ public struct InteractionDefinition
         /// <summary>
         /// The primary action of the button as defined by the controller SDK.
         /// </summary>
-        public InputType InputType { get; set; }
+        public InputAction InputAction { get; set; }
```
StephenHodgson: Get access only.

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -22,6 +36,146 @@ public struct InteractionDefinition
...
+
+        #endregion Interaction Properties
+
+        #region Definition Data items
+
+        /// <summary>
+        /// The data storage for a Raw / None Axis type.
+        /// </summary>
+        private object rawData { get; set; }
```
StephenHodgson: These could be fields, correct?

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinitionMapping.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
+{
+    public struct InteractionDefinitionMapping
+    {
+        public InputType InputType { get; set; }
+        public AxisType AxisType { get; set; }
+        public InputAction InputAction { get; set; }
```
StephenHodgson: I still feel like `InputType ==InputAction` because of their enum types.  Maybe we can change `InputAction` to a struct with some data that users can set per interaction, like the action id, a custom name, etc...

Assets/MixedRealityToolkit/_Core/Definitions/ControllerState.cs
SimonDarksideJ: I'll be reverting following the discussion last night, yay more work.

Assets/MixedRealityToolkit/_Core/Definitions/Headset.cs
SimonDarksideJ: tbc

Assets/MixedRealityToolkit/InputSystem/Sources/DictationInputManager.cs
SimonDarksideJ: ok

Assets/MixedRealityToolkit/InputSystem/Sources/SpeechInputSource.cs
SimonDarksideJ: ok

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
SimonDarksideJ: I think we resolved this in talk last night:

* InputType is the Physical Input Type (maybe we should rename it)
* Input Action is the Logical Action that is (should) be sent to the InputSystem

The mapping directs a specific input to output the action the dev has configured.  Multiple inputs can point to the same action.

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSource.cs
```diff
@@ -1,32 +1,83 @@
...
+        /// Returns whether the Input Source supports the specified input types.
+        /// </summary>
+        /// <param name="inputInfo">Input types that we want to get information about.</param>
+        bool SupportsCapability(InputType inputInfo);
+
+        /// <summary>
+        /// Register pointers against the controller
+        /// </summary>
+        /// <param name="pointers">Array of pointers to attach</param>
+        void RegisterPointers(IMixedRealityPointer[] pointers);
```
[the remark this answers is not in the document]
SimonDarksideJ: That would be fair, but I leave to you to review since I'm unplugging inputsource from the controllers now.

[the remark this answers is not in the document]
SimonDarksideJ: But the other Pointer code needs to go in to it's own static class for reuse, leaving only 

* RegisterPointers (meant to initialise or replace all current pointers)
* Unregister (although this could be achieved with RegisterPointers(null)

Possibly Register Pointers needs renaming if the array is intended to be modified through it.

Assets/MixedRealityToolkit/InputSystem/Sources/BaseGenericInputSource.cs
```diff
@@ -31,26 +35,45 @@ public BaseGenericInputSource(string name, InputType[] capabilities, IMixedReali
...
+                    return true;
                 }
             }
 
             return false;
         }
 
+        public void RegisterPointers(IMixedRealityPointer[] pointers)
+        {
+            if (pointers == null) { throw new ArgumentNullException(nameof(pointers)); }
```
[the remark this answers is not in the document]
SimonDarksideJ: Following comment below, if you intend Register to allow pointers to clear

But if it is already Null, you shouldn't allow assigning null.  You should only call register if there is something to register?

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinitionMapping.cs
SimonDarksideJ: As discussed above:   I think we resolved this in talk last night:

* InputType is the Physical Input Type (maybe we should rename it)
* Input Action is the Logical Action that is (should) be sent to the InputSystem

The mapping directs a specific input to output the action the dev has configured. Multiple inputs can point to the same action.

SimonDarksideJ:
Closing and rebasing due to the amount of features developed and exported in to separate PR's.
Will submit new PR with just Device changes

PR #2104, opened 2018-05-15
fix #2103 - allow unsafe code

### LINKED ISSUE #2103: [may18] "allow unsafe code" not auto-checked via mcs.rsp
With the addition of Spectator View, projects including the folder in their project must tell Unity (and Visual Studio) that unsafe code is allowed.

Unity should read the mcs.rsp file (included in the MRTK) and set the checkbox in the Visual Studio project.

This appears to not be happening.

The current workarounds are:
- Manually check "Allow Unsafe Code" **each time** the solution is built from Unity
- Do not generate C# projects when building in Unity (let Unity generate the assemblies).

This may be that the MRTK does not have the correct rsp file for the specific version of Unity or it may be a Unity issue.

davidkline-ms:
Add csc.* to fix Unity not enabling Allow Unsafe Code in the generated projects when building for the .NET Backend.

Tested on Unity versions:
- 2017.1.3p4
- 2017.2.1p2
- 2017.4.3f1

Fixes: #2103


davidkline-ms:
verified that with BOTH files present, Unity successfully compiles the dll with IL2CPP.

davidkline-ms:
Updated PR description to match the current contents.

StephenHodgson:
Do we also need to make sure this file gets included into the asset package?

What happens if someone has a file like this already declared?

davidkline-ms:
@StephenHodgson,

I will be adding these files to the asset package for the May release. If the user has one, the package importer should show it as already there (and not override)

StephenHodgson:
> the package importer should show it as already there (and not override)

that's not exactly how that works. It'll only override if the meta file id is the same.

davidkline-ms:
We had been shipping .rsp files in earlier releases.  Is the concern that ours may overwrite an existing rsp?

PR #2129, opened 2018-05-18
MRTK/Standard Shader Improvements

Cameron-Micka:
Overview
---
Integrating the latest work to improve the MRTK/Standard shader. Improvements include:

- Support for non-D3D11 APIs (tested on iOS and Android devices using OpenGL ES and Metal)
- Support for three hover lights (useful with IHMDs with multi-pointers)
- Improved rounded corner algorithm
- Closer match to the Unity/Standard shader
- Improved shader based anti-aliasing support
- Minor optimizations

aa
(Shader based anti-aliasing example.)

### DIFF
--- a/Assets/MixedRealityToolkit/Common/Scripts/HoverLight.cs
+++ b/Assets/MixedRealityToolkit/Common/Scripts/HoverLight.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Collections.Generic;
 using UnityEngine;
 
 namespace MixedRealityToolkit.Common
@@ -18,9 +19,12 @@ namespace MixedRealityToolkit.Common
         [SerializeField]
         private Color color = new Color(0.3f, 0.3f, 0.3f, 1.0f);
 
-        private int hoverPositionID;
-        private int hoverRadiusID;
-        private int hoverColorID;
+        // Two hover lights are supported at this time.
+        private const int hoverLightCount = 2;
+        private static List<HoverLight> activeHoverLights = new List<HoverLight>(hoverLightCount);
+        private static int[] hoverLightIDs = new int[hoverLightCount];
+        private static int[] hoverLightColorIDs = new int[hoverLightCount];
+        private static int lastHoverLightUpdate = -1;
 
         public float Radius
         {
@@ -50,12 +54,13 @@ namespace MixedRealityToolkit.Common
 
         private void OnEnable()
         {
-            Initialize();
+            AddHoverLight(this);
         }
 
         private void OnDisable()
         {
-            UpdateHoverLight();
+            RemoveHoverLight(this);
+            UpdateHoverLights(true);
         }
 
 #if UNITY_EDITOR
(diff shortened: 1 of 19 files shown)


StephenHodgson:
> Support for two hover lights (useful with IHMDs with two pointers)

Kinda curious how to support _n_th nmber of pointers.

Assets/MixedRealityToolkit/Common/Scripts/HoverLight.cs
```diff
@@ -18,9 +19,12 @@ public class HoverLight : MonoBehaviour
         [SerializeField]
         private Color color = new Color(0.3f, 0.3f, 0.3f, 1.0f);
 
-        private int hoverPositionID;
-        private int hoverRadiusID;
-        private int hoverColorID;
+        // Two hover lights are supported at this time.
+        private const int hoverLightCount = 2;
```
StephenHodgson: Is this a hard limit?  What's preventing us from supporting any number?

Assets/MixedRealityToolkit/Common/Shaders/Standard.shader
```diff
@@ -456,46 +491,61 @@ Shader "MixedRealityToolkit/Standard"
...
 #endif
 
                 albedo *= _Color;
 
                 // Hover light.
 #if defined(_HOVER_LIGHT)
-                fixed pointToHover = (1.0 - saturate(length(_HoverPosition - i.worldPosition.xyz) / _HoverRadius)) * _HoverColor.a;
+                fixed pointToHover0 = HoverLight(_HoverLight0, i.worldPosition.xyz, _HoverLightColor0.a);
+#if defined(_MULTI_HOVER_LIGHT)
+                fixed pointToHover1 = HoverLight(_HoverLight1, i.worldPosition.xyz, _HoverLightColor1.a);
```
StephenHodgson: Oh I see, we need to define a fixed point for each in the shader itself.

StephenHodgson: Is there a way to make this more generic to support any number of HoverLights?

Cameron-Micka: There isn't an explicit hard limit, but I wanted to settle on a number that didn't incur much overhead (especially for the HoloLens and "Core" Windows MR PCs). Two seemed like a good number since usually we have zero to two motion controllers/pointers (I'm sure that will change before we know it haha).

If there is an outcry for more hover lights, I'm happy to add a more generic solution in the future!

StephenHodgson: Could we do 3 for now?

Cause chances are you'll have gaze as well as two controller pointers.

Cameron-Micka: 3 lights will be an easy change, I'll make those changes and upload them tomorrow. Since you are  most likely to need 2+ hover lights on an immersive HMD we should have performance to spare. :)

Cameron-Micka:
@StephenHodgson I just pushed a commit that increases the HoverLight support to 3 lights. And, it should be simple to increase the amount if need be in the future (just need to change a constant in the shader and HoverLight.cs).

Alexees:
Hi, I just dropped the HolographicButton prefab from june18_dev into a scene on Android platform and it looks like like this. Is the GL ES support somewhere to be found already
holographicbutton
?

StephenHodgson:
@Alexees please open new issues for questions. Thanks.

Issue #1978, opened 2018-04-18
Caret not showing on MRT Keyboard Input Field


Vangaorth:
## Overview
The 'KeyboardTest' scene inside any Examples package has everything already set up. Otherwise, create a new scene with MixedRealityCameraParent prefab, DefaultCursor, Input Manager prefab, MixedRealityToolkit Keyboard prefab, a main canvas with a child InputField which script has been replaced by a KeyboardInputField component and which 'text' property has been filled with some short single line text.

## Expected Behavior
On Hololens, when I click/tap on the GameObject containing the KeyboardInputField, the MixedRealityToolkit keyboard appears and the caret on the inputfield of the keyboard should be visible and blanking.

## Actual Behavior
The MixedRealityToolkit keyboard appears but the caret of the kayboard input field does not (on Hololens).

## Steps to reproduce
Uuse [this repo](https://github.com/Vangaorth/HolotoolkitCaretIssue)

Otherwise, create a new project, switch to Universal Windows Platform, create a new scene, import an MixedRealityToolkit >= 2017.2.1.0, import the relative 'Example' unity package (found on the releases page), open the KeyboardTest scene, search for the email input field and fill the 'text' property of its 'KeyboardInputField' with some short text (e.g., an email address),  from MixedRealityToolkit menu, select Configure -> Apply MR Project Settings -> Apply, add the scene to the built scenes, build and deploy. On Hololens, tap the GameObject containing the KeyboardInputField.

## Unity Editor Version
2017.2.1p2

## Mixed Reality Toolkit Release Version
2017.2.1.4 Release Candidate 2 but the problem is in every MRT version starting from MixedRealityToolkit for Unity 2017.2.1f1 (included). MixedRealityToolkit for Unity 2017.2.0p1 MRTP4 and previous are not affected by this problem.

davidkline-ms:
Re-enabling the caret causes issue #1940 to repro. I am investigating a fix/workaround to that issue that will accompany the restoration of the caret.

davidkline-ms:
This was released in the 2017.4.0.0 release candidate

Issue #1940, opened 2018-04-11
XAML: Holotoolkit triggers own and system keyboard


IrisSReflekt:
## Overview
I use the Holotoolkit keyboard and gaze at its input field (the one directly on the keyboard) 

## Expected Behavior
nothing should happen except for maybe focus change 

## Actual Behavior
As soon as I gaze at the input field, I am thrown out of my app and into the 2d keyboard app. if i enter my text there and submit, the text is submitted into the input field of the holotoolkit keyboard and i have to click enter on that one once more to finally submit the text. If i gaze again on the input field the external keyboard opens again. This can make the user be stuck in a loop if he keeps looking there to check his input

## Steps to reproduce
build with xaml
open holotoolkit keyboard
gaze at the input field

## Unity Editor Version
2017.2.0f3 - building for XAML with 10.0.16299.0 sdk

## Mixed Reality Toolkit Release Version
2017.2.1.0

StephenHodgson:
Duplicate of #1370

StephenHodgson:
Well not exactly a duplicate. Essentially we need to remove the keyboard prefab from the project for those that are using RS4 in the HoloLens

IrisSReflekt:
I dont think we are using rs4, i just checked os and firmware version numbers:
os: 10.0.14393.2007
firmware: 0005.05.P.1606.291.233.R

Also, if you would remove the keyboard prefab, that would mean we have to use the integrated 2d one? because we actually want to only use the one from holotoolkit to prevent breaking the 3d immersion

cybergoulion:
It's not specificallyon topic but I also exported my project in XAML because I needed to open a registered filetype file in the desktop version using OnFileActivated(FileActivatedEventArgs args). I didn't try this on the hololens version yet, but I cannot see another way of doing it. Is there a right/recommended way to do this in a D3D project?

PR #2132, opened 2018-05-18
Re-enable keyboard caret, move test scene

davidkline-ms:
Overview
---
This change re-enables the caret in the MRTK keyboard. This MAY result in both the MRTK and System keyboards being displayed at the same time (#1940) -- we will continue to investigate this issue separately. Submitting this as not having a caret is a significant usability issue.

The keyboard test scene has been moved from Examples\Input to Examples\UX to avoid confusion, as that is where the scripts reside.

Fixes: #1978


davidkline-ms:
possibly, but not 100% from what I have investigated

Issue #2131, opened 2018-05-18
May18 branch: grip position not reported for in-Editor hands


PJBowron:
## Overview
I believe there was an oversight in my changes to address #1824 - the Left/RightHandInputControl prefabs do not have 'grip position' selected as a supported feature. Additionally the `CustomInputControl` that governs in-Editor input doesn't update grip position/rotation in the `SourcePose`.

I think perhaps that a subsequent change has introduced more use of `TryGetGripPosition` than previously existed? It's these calls that are failing now, however, they always would have, had they been there before my change!?

**Note: on device, behavior works as expected.**

I've got a local fix; I'll post a PR asap.

## Expected Behavior
Running in Editor, attempts to obtain grip position succeed and return the general hand position.

## Actual Behavior
Attempts fail, grip position returns as 0,0,0

## Steps to reproduce
1. setup an MRTK scene
2. add an object with a `HandDraggable` script
3. the object cannot be moved in-Editor, but can on device

## Mixed Reality Toolkit Release Version
May18 branch; SHA: 01876ebc6d6ca23319a5ee4aff784a5bc538e9f4

PJBowron:
@davidkline-ms

keveleigh:
Ah, looks like this happened in your change. `CustomInputSource.cs`, the script used for the hands, was updated in #1870 and #2050 to properly return the grip position from `TryGetGripPosition`, instead of faking it and returning the regular position in the past.

https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2050/files#diff-d55c55a745b71a905534124d85148516L176

PJBowron:
Hopefully the changes in #2133 should do the trick.
I don't really understand why I didn't see this when I made the first change... I must have had other changes that didn't make it into the submission... either way, my bad :(

davidkline-ms:
This was released in the 2017.4.0.0 release candidate

PR #1870, opened 2018-03-23
Adding GripPosition and GripRotation to SupportedInputInfo and adding support to WSA input source and related classes

PJBowron:
Overview
---
WSA input sources do not currently report their capabilities correctly in regards to Pointer position/rotation and grip position/rotation. This PR attempts to fix the situation by adding GripPosition/GripRotation input info types, and clearly differentiating between the Pointer and Grip types. InteractionInputSource and a number of related classes are updated to accommodate use of the new types.

Changes
---
* added PointerPosition (Position-equivalent), PointerRotation (Rotation-equivalent), GripPosition and GripRotation  to SupportedInputInfo enumeration
* deprecated existing Position and Rotation enum ids

- Fixes: #1824 .

PR #2133, opened 2018-05-18
Fix to allow in-Editor hands to properly support and report a 'grip' position

PJBowron:
Overview
---
Fixes an issue introduced #1870 (& #2050 ) whereby the in-Editor hands no longer report their position, if data is requested through `TryGetGripPosition`

Changes
---
- Fixes: #2131  .


PJBowron:
For transparency, this fix hasn't been tested on device: I won't be able to do so until until Tuesday.
It's hard to see that there could be any adverse effect however.

PR #2134, opened 2018-05-19
Small tweaks to input system

StephenHodgson:
Fixed some issues with querying Input Source Capabilities and updated Interaction Definition struct.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Definitions/InputType.cs
+++ b/Assets/MixedRealityToolkit/_Core/Definitions/InputType.cs
@@ -4,8 +4,7 @@
 namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
 {
     /// <summary>
-    /// The InputType defines the types of input exposed by a controller.
-    /// Denoting the available buttons / interactions that a controller supports.
+    /// The InputType defines the types of input exposed by a controller or Input Source.
     /// </summary>
     public enum InputType
     {
--- a/Assets/MixedRealityToolkit/_Core/Definitions/InputType.cs
+++ b/Assets/MixedRealityToolkit/_Core/Definitions/InputType.cs
@@ -4,14 +4,24 @@
 namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
 {
     /// <summary>
-    /// The InputType defines the types of input exposed by a controller or Input Source.
+    /// The InputType defines the types of input exposed by a controller.
+    /// Denoting the available buttons / interactions that a controller supports.
     /// </summary>
     public enum InputType
     {
         None = 0,
         Voice,
+        /// <summary>
+        /// 6 Dof Pointer with position and rotation.
+        /// </summary>
         Pointer,
+        /// <summary>
+        /// 3 Dof Pointer with only position.
+        /// </summary>
         PointerPosition,
+        /// <summary>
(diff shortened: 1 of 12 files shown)


Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -9,19 +12,178 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
...
+        /// </summary>
+        private Quaternion RotationData { get; set; }
+
+        private Tuple<Vector3, Quaternion> TransformData { get; set; }
+
+        #endregion Definition Data items
+
+        #region Get Operators
+
+        public object GetRaw()
```
SimonDarksideJ: As discussed, I personally don't like the multiple GET option as it complicates the API.

We should have the following allowing the model to store and retrieve the data as it sees fit:
* Set Data
* Get Data


Where as this is now:


* Set Data

* Get Bool

* Get Float

* Get....


Which makes the code look very messy.  I'd prefer to keep the original model I laid out and only return to this if we hit issues.

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
```diff
@@ -9,19 +12,178 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
...
 
         /// <summary>
-        /// The primary action of the button as defined by the controller SDK.
+        /// The position data storage for a 3DoF or 6DoF Axis type.
         /// </summary>
-        public InputType InputType { get; set; }
+        private Vector3 PositionData { get; set; }
+
+        /// <summary>
+        /// The rotation data storage for a 3DoF or 6DoF Axis type.
```
SimonDarksideJ: This isn't used for 6DoF unless I've missed something?  We have the Tuple for that.

Assets/MixedRealityToolkit/_Core/Definitions/InputType.cs
```diff
@@ -11,8 +11,17 @@ public enum InputType
     {
         None = 0,
         Voice,
+        /// <summary>
+        /// 6 Dof Pointer with position and rotation.
```
SimonDarksideJ: I like this new separation

SimonDarksideJ:
Best to pull this @StephenHodgson if you are now changing on my PR

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
StephenHodgson: Just wanted to avoid boxing

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
StephenHodgson: I wonder if we should have all 3. Might need to add the tuple property and initializer

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinition.cs
StephenHodgson: Can we do both?

StephenHodgson:
closed in favor of #2135

PR #2135, opened 2018-05-21
Major Refactoring for Input System

StephenHodgson:
Overview
---
Organized and moved quite a bit around.

- Removed a bunch of mapping classes we probably won't use
- Noticed we added pointer registration to the base input source, and I removed it again. That's not where we want to add that functionality but in a higher derived input source class that is specifically for spatial controllers that can have one or more or dynamically generated pointers.
- I'm going to try and make sure all our input sources are pure c#
- Refactored most to be except for InteractionInputSources and SpeechInputSource. SpeechInputSource will need a config profile to setup speech keywords and actions.
- Moved the controller visualization classes and helpers out of the input system too
- Ported a change to the SystemType to support structs


Assets/MixedRealityToolkit/InputSystem/Utilities/Enums/OculusControllerInputType.cs
```diff
@@ -1,46 +0,0 @@
...
-
-using System;
-
-namespace Microsoft.MixedReality.Toolkit.InputSystem.Utilities
-{
-    [Flags]
-    public enum OculusControllerInputType
-    {
-        None                           = 0,
-        ButtonOnePress                 = 1 << 0,
```
SimonDarksideJ: Maybe we should take some of these Enum Values for the Device Input Type (yay thought of a better name)

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/InputType.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
SimonDarksideJ: Given the device nature of the use of this enum, I'll remove it from Input (or create a new one in _Core)

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/InteractionDefinition.cs
```diff
@@ -0,0 +1,195 @@
...
+        private Vector3 positionData;
+
+        /// <summary>
+        /// The rotation data storage for a 3DoF type.
+        /// </summary>
+        private Quaternion rotationData;
+
+        private Tuple<Vector3, Quaternion> transformData;
+
+        #endregion Definition Data items
```
SimonDarksideJ: you were supposed to leave in the single Get, but I'll re-add this after merge

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputAction.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem
```
SimonDarksideJ: Needs to be in "_Core" as it's consumed by BOTH device and Input

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/InteractionDefinition.cs
StephenHodgson: Ah, my bad.

PR #2144, opened 2018-05-21
HostTransform for TwoHandManipulatable

Jarodshow:
## Overview
Updated TwoHandManipulatable to use `HostTransform` for `BoundingBoxRig`.


davidkline-ms:
There will be some porting from master into vNext. Not 100% sure if we will use dev as a stepping stone due to the fact that it will be locked shortly and would require two passes at the namespace / folder structure.

keveleigh:
Looks like there's definitely been some divergence between the branches in this file.

I don't see this change in master though: https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/may18_dev/Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/TwoHandManipulatable.cs#L120

keveleigh:
@Jarodshow What does this change do/fix? Looks like this is a good change, so I'm mostly curious how the bug manifests itself.

PR #2159, opened 2018-05-24
Fix for choosing closest grabbable object to grab

paulmriordan:
Sort method was comparing distance of first object to itself, rather than second object.

Let me know if I need to sign a contribution licence agreement.


keveleigh:
Closed in favor of #2159

PR #2160, opened 2018-05-24
Fix for dragging of rotation handles of BoundBoxRig

paulmriordan:
Current method of rotation is not consistent; the object is rotated different directions
depending on orientation of object.

This commit adds an option (rotateAroundPivot) which allows the user to rotate an object in a more natural way.

To avoid breaking existing usage, this new rotation is enabled via a bool in BoundingBoxRig.

Existing behaviour:
boundingboxgizmoold

New behaviour:
boundingboxgizmonew


davidkline-ms:
I am working on a port of this change that targets the june18_dev branch. I will reference it and close this one when I submit the PR.

keveleigh:
Closed in favor of #2349, due to inactivity here.

Issue #2167, opened 2018-05-25
[master] Create HoloToolkit-Preview folder/package for early or experimental releases


davidkline-ms:
This would effectively isolate features that
- are experimental
- may change (breaking) more rapidly than other features
- may not be ready for wide adoption
- might not pass WACK (due to native dependencies)

Issue #2163, opened 2018-05-24
may18 rc - WACK fails - spectatorview


davidkline-ms:
## Overview
Compiling a project using the 2017.4.0.0 release candidate fails WACK with errors regarding the pinvokes into the spectatorviewplugin.dll -- the dll is NOT present in the build.

## Steps to reproduce
- Import the 2017.4.0.0 release candidate and examples unitypackages
- Open your favorite scene
- Genereate solution
- Build master config
- Generate store package
- Run local WACK

davidkline-ms:
I am going to investigate adding a #define to opt-in to building the SpectatorView components

StephenHodgson:
Is this because we're missing a copy of it in both the normal plugins folder and a wsa folder?

davidkline-ms:
Most likely. I am working on an option to avoid this issue for customers that are not using SpectatorView, yet import the entire unitypackage.

davidkline-ms:
With all required DLLs present, the API check fails when SpectatorView is enabled.

API gzclose in zlib1.dll is not supported for this application type. opencv_core341.dll calls this API.
API gzeof in zlib1.dll is not supported for this application type. opencv_core341.dll calls this API.
API gzgets in zlib1.dll is not supported for this application type. opencv_core341.dll calls this API.
API gzopen in zlib1.dll is not supported for this application type. opencv_core341.dll calls this API.
API gzputs in zlib1.dll is not supported for this application type. opencv_core341.dll calls this API.
API gzrewind in zlib1.dll is not supported for this application type. opencv_core341.dll calls this API.

davidkline-ms:
Removing Bug and adding External tag as this is an issue with OpenCV and not the MRTK code

PR #2168, opened 2018-05-25
create HoloToolkit-Preview, moved SpectatorView there until OpenCV passes WACK

davidkline-ms:
Overview
---
Without the SpectatorViewPlugin and associated OpenCV DLLs, WACK fails the PInvoke calls in the new SpectatorView feature. With these DLLs present, WACK fails due to APIs provided by zlib1.dll not being supported. zlib is being called by OpenCV_core341.dll

The recent change to set ExactSpelling in the SpectatorView DllImport attributes was causing the stated EntryPoint values to seemingly be ignored.

Changes
---
Create a new project folder called HoloToolkit-Preview -- An asset package called HoloToolkit-Unity-Preview-<version> will be released alongside the main package.

Change the SpectatorView namespace to HoloToolkit.Unity.Preview.SpectatorView and move it into a new folder. Once it's dependencies are able to pass WACK, the Preview portion of the namespace will be removed and SpectatorView will be moved into the HoloToolkit project folder.

Addresses: #2163 
Fixes: #2167


Assets/HoloToolkit-Preview/SpectatorView/Scripts/SpatialSync/MarkerDetector.cs
```diff
@@ -6,29 +6,29 @@
...
-namespace HoloToolkit.Unity.SpectatorView
+namespace HoloToolkit.Unity.Preview.SpectatorView
 {
     /// <summary>
     /// Manages the OpenCV wrapper to detect a marker.
     /// </summary>
     public class MarkerDetector
     {
-        [DllImport("SpectatorViewPlugin", EntryPoint="MarkerDetector_Initialize", ExactSpelling=true)]
+        [DllImport("SpectatorViewPlugin", EntryPoint="MarkerDetector_Initialize")]
```
StephenHodgson: Do we want exact spelling? 

Or was this part of the problem we had with building and compiling?

davidkline-ms: I got reports that that was blocking the runtime from finding the entry points. ExactSpelling quiets the compiler warnings, but apparently ignores the EntryPoint param in the attribute.

PR #2169, opened 2018-05-25
[May18] Disable keyboard for the MRTK keyboard's inputfield

keveleigh:
Overview
---
This sets the slider input field in the MRTK keyboard to have a keyboardType that isn't defined, which prevents a keyboard from popping up.

This PR does not address the situation where both the MRTK and the system keyboard pop up on other input fields. This only addresses the situation where the MRTK keyboard triggers the system keyboard itself.

Changes
---
- Fixes: #1940

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/Keyboard.cs
+++ b/Assets/HoloToolkit/UX/Scripts/Keyboard.cs
@@ -251,6 +251,8 @@ namespace HoloToolkit.UI.Keyboard
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Assets/HoloToolkit/UX/Scripts/Keyboard.cs
```diff
@@ -251,6 +251,8 @@ protected override void Awake()
                 }
             }
 
+            InputField.keyboardType = (TouchScreenKeyboardType)(-1);
```
davidkline-ms: Recommend a comment re: why here.

PR #2175, opened 2018-05-27
Updated MixedRealityManager Initialization and Config Profile setup

StephenHodgson:
Now if there's not a MixedRealityManager in the scene and a script reference calls for an instance, it'll create a new manager object in the scene, then ping it.  If more than one config profile exists in the project a object window picker shows up to pick the config profile of your choice. If only the default profile exists, then it sets it automatically.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealitySpeechCommandsProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealitySpeechCommandsProfileInspector.cs
@@ -25,12 +25,7 @@ namespace Microsoft.MixedReality.Toolkit.Inspectors
 
         private void OnEnable()
         {
-            if (MixedRealityManager.Instance.ActiveProfile == null)
-            {
-                Debug.LogError("No Active Profile set on the Mixed Reality Manager.");
-                Selection.activeObject = MixedRealityManager.Instance;
-                return;
-            }
+            if (MixedRealityManager.Instance.ActiveProfile == null) { return; }
 
             speechCommands = serializedObject.FindProperty("speechCommands");
             actionLabels = MixedRealityManager.Instance.ActiveProfile.InputActionsProfile.InputActions.Select(
--- a/Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealitySpeechCommandsProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealitySpeechCommandsProfileInspector.cs
@@ -25,6 +25,13 @@ namespace Microsoft.MixedReality.Toolkit.Inspectors
 
         private void OnEnable()
         {
+            if (MixedRealityManager.Instance.ActiveProfile == null)
+            {
+                Debug.LogError("No Active Profile set on the Mixed Reality Manager.");
+                Selection.activeObject = MixedRealityManager.Instance;
+                return;
+            }
+
             speechCommands = serializedObject.FindProperty("speechCommands");
             actionLabels = MixedRealityManager.Instance.ActiveProfile.InputActionsProfile.InputActions.Select(
                 action => new GUIContent(action.Description)).Prepend(new GUIContent("None")).ToArray();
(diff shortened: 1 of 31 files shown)


Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -495,15 +537,15 @@ public IEnumerable<IMixedRealityManager> GetManagers(Type type)
...
+            if (ActiveProfile == null)
             {
                 Debug.LogError($"Unable to add a new {type.Name} Manager as the Mixed Reality manager has to Active Profile");
                 return null;
             }
 
             if (type == null) { throw new ArgumentNullException(nameof(type)); }
 
-            List<IMixedRealityManager> managers = new List<IMixedRealityManager>();
+            var managers = new List<IMixedRealityManager>();
```
SimonDarksideJ: Why change this to a var?  It's a fixed implementation with a concrete type, a var is just going to add unnecessary overhead

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -484,7 +526,7 @@ public IMixedRealityManager GetManager(Type type, string managerName)
         /// <returns>An array of Managers that meet the search criteria</returns>
         public IEnumerable<IMixedRealityManager> GetManagers(Type type)
         {
-            return GetManagers(type, "");
+            return GetManagers(type, string.Empty);
```
SimonDarksideJ: Following testing, we need to add a null check here

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -294,6 +274,83 @@ private void Awake()
...
         {
+            EnableAllManagers();
+        }
+
+        /// <summary>
+        /// The MonoBehaviour Update event, which is then circulated to all active managers
+        /// </summary>
+        private void Update()
+        {
+            UpdateAllManagers();
```
SimonDarksideJ: I approve of this abstraction,  I've been doing it more and more in refactoring.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -191,13 +159,13 @@ public static MixedRealityManager Instance
                 switch (objects.Length)
                 {
                     case 0:
-                        instance = new GameObject("MixedRealityManager").AddComponent<MixedRealityManager>();
+                        instance = new GameObject(nameof(MixedRealityManager)).AddComponent<MixedRealityManager>();
```
SimonDarksideJ: Not entirely sure why we need to have this, but not opposed to it.  The MR Manager is the one component in the toolkit that can't really be stripped out,. Unless we are now also looking at a consumable component model?

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealitySpeechCommandsProfileInspector.cs
```diff
@@ -25,12 +25,7 @@ public class MixedRealitySpeechCommandsProfileInspector : MixedRealityBaseConfig
 
         private void OnEnable()
         {
-            if (MixedRealityManager.Instance.ActiveProfile == null)
-            {
-                Debug.LogError("No Active Profile set on the Mixed Reality Manager.");
```
SimonDarksideJ: Why remove this error?  The MRManager cannot operate without a config, unless you have now replaced this with the popup you mentioned.
Even with the popup however, it is still a valid error if you try to run the project without one

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
StephenHodgson: syntactic sugar I suppose.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
StephenHodgson: in case we ever rename the class

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealitySpeechCommandsProfileInspector.cs
StephenHodgson: Because I added logic to prevent users from going into play mode if no active profile is set.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
StephenHodgson: I usually go with var if the type is explicit

PR #2184, opened 2018-05-28
MRTK - vNext - Windows Mixed Reality Device Support - rebase

SimonDarksideJ:
Overview
---
Adding the initial Windows MR Device and Controller definitions, including some refactoring of core / input to accommodate.

Changes
---
 - Adds new WMRDevice (temp name, will rename for merge)
-  Adds new Windows Mixed Reality Controller input source
 - Adds new InteractionDefinition structure for managing input data from device and controller mappings.
 - Changed Controller struct to an Interface for better management - Post Alpha clean up required
- Add events for buttons / controllers / pointers / grip
- Consolidate input events raised
- Documentation pass on new classes (and a few standards corrections on others)
- Wired up default pointers 
- Scaffolds - definitions added for other SDK's frameworks and controllers added but are ***NOT IMPLEMENTED*** (don't get excited, a ways off that yet) - *Future PR*
- Parted the red sea 🌊 

Known Issues
---
 None

Notes
---
Ready for Merge


Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/SpeechCommands.cs
```diff
@@ -6,26 +6,44 @@
...
         public SpeechCommands(string keyword, KeyCode keyCode, InputAction action)
         {
             this.keyword = keyword;
             this.keyCode = keyCode;
             this.action = action;
         }
 
+        /// <summary>
+        /// The input string keyword that will be listened for by the SpeechInputSource
+        /// </summary>
```
StephenHodgson: Summary tags in wrong place.

Also let's do a doc pass in a separate PR.

I just wanna see the controller implementation.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// The registered Input Source for this controller
+        /// </summary>
+        IMixedRealityInputSource InputSource { get; }
+
+        /// <summary>
+        /// Mapping definition for this controller, linking the Physical inputs to logical Input System Actions
+        /// </summary>
+        Dictionary<DeviceInputType, InteractionDefinition> Interactions { get; }
```
StephenHodgson: Still not sure this should use the device input type as the key

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/InputAction.cs
```diff
@@ -22,6 +22,7 @@ public InputAction(uint id, string description)
         {
             this.id = id;
             this.description = description;
+            this.axis = AxisType.None;
```
StephenHodgson: Shouldn't this be getting passed into the constrictor?

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
```diff
@@ -11,6 +12,34 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem
     [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Mixed Reality Input Actions Profile", fileName = "MixedRealityInputActionsProfile", order = 1)]
     public class MixedRealityInputActionsProfile : ScriptableObject
     {
+        private readonly string[] inputActionsMRTKDefaults =
```
StephenHodgson: Not sure if I like the MRTK abbreviation in the name.

`defaultInputActions` would be suitable

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealitySpeechCommandsProfile.cs
```diff
@@ -11,6 +11,9 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem
     [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Mixed Reality Speech Commands Profile", fileName = "MixedRealitySpeechCommandsProfile", order = 2)]
     public class MixedRealitySpeechCommandsProfile : ScriptableObject
     {
+        /// <summary>
+        /// Array of configured speech commands to be consumed by the SpeechInputSource
+        /// </summary>
```
StephenHodgson: Why put summary tag on private field?

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinitionMapping.cs
```diff
@@ -0,0 +1,32 @@
...
+    public struct InteractionDefinitionMapping
+    {
+        /// <summary>
+        /// The physical button / control you want to map
+        /// </summary>
+        public DeviceInputType InputType { get; set; }
+
+        /// <summary>
+        /// The Axis of the Action to be used in your project 
+        /// *Note list will be filtered to only those controller inputs that support this Axis
```
StephenHodgson: Replace *Note with `<remarks><remarks/>`

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityControllerMappingProfile.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public Handedness ControllingHand
+        {
+            get { return controllingHand; }
+            private set { controllingHand = value; }
+        }
+
+        [SerializeField]
+        [Tooltip("Controller type to instantiate at runtime.")]
+        [Implements(typeof(IMixedRealityController), TypeGrouping.None)]
```
StephenHodgson: Let's use flat grouping

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -116,6 +116,36 @@ public bool RenderMotionControllers
...
         }
 
+        [SerializeField]
+        [Tooltip("Default Left Controller Model")]
+        private GameObject leftControllerModel;
+
+        /// <summary>
+        /// Enable and configure the controller rendering for the Mixed Reality Toolkit
+        /// </summary>
+        public GameObject LeftControllerModel
```
StephenHodgson: I think the controller models should be moved into the controller configuration profiles.

Assets/MixedRealityToolkit/_Core/Devices/Generic/XBoxController.cs
```diff
@@ -0,0 +1,68 @@
...
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions;
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Devices;
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+using System.Collections.Generic;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Devices.Generic
+{
+    // TODO - Implement
+    public struct XBoxController : IMixedRealityController
```
StephenHodgson: Not sure about the name here.
Might just be GamepadController per windows namespace naming.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
```diff
@@ -0,0 +1,437 @@
...
+using Microsoft.MixedReality.Toolkit.Internal.Managers;
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.XR.WSA.Input;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Devices.WindowsMixedReality
+{
+    // TODO - name not final
+    public class WMRDevice : IMixedRealityDevice
```
StephenHodgson: What about Windows mixed reality device manager?

Assets/MixedRealityToolkit/_Core/Extensions/CollectionsExtensions.cs
```diff
@@ -113,5 +113,32 @@ public static void DisposeElements<TElement>(this IList<TElement> elements)
...
+
+        /// <summary>
+        /// Overload extension to enable saving of an InteractionDefinition inside a Dictionary collection
+        /// *Note can only use generics (in both here and InteractionDefinition)
+        /// </summary>
+        /// <typeparam name="T">Type of input being saves</typeparam>
+        /// <param name="input">The InteractionDefinition dictionary reference (generics, performed on a Dictionary)</param>
+        /// <param name="key">The specific DeviceInputType value to update</param>
+        /// <param name="value">The data value to be updated</param>
+        public static void SetDictionaryValue<T>(this Dictionary<Definitions.Devices.DeviceInputType, Definitions.Devices.InteractionDefinition> input, Definitions.Devices.DeviceInputType key, T value)
```
StephenHodgson: Could this be generic?

Assets/MixedRealityToolkit/_Core/Extensions/InputActionsExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public static class InputActionsExtensions
+    {
+        /// <summary>
+        /// Get an Input Action by name from the current Array
+        /// </summary>
+        /// <param name="input">InputAction Array to query</param>
+        /// <param name="actionName">Input Action name to search for</param>
+        /// <returns>A specific InputAction if found, else null</returns>
+        public static InputAction GetActionByName(this InputAction[] input, string actionName)
```
StephenHodgson: Why do we need this?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -309,124 +329,6 @@ private void OnDestroy()
 
         #endregion MonoBehaviour Implementation
 
-        private void InitializeAllManagers()
```
StephenHodgson: Why move these and make them out of order?

Assets/MixedRealityToolkit/_Core/Devices/Generic/XBoxController.cs
StephenHodgson: But also, we could reuse the controller type for more than just Xbox devices

Assets/MixedRealityToolkit/_Core/Definitions/Handedness.cs
```diff
@@ -10,24 +10,28 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
     /// It is up to the developer to determine whether this affects the use of a controller or not.
     /// "Other" defines potential controllers that will offer a "third" hand, e.g. a full body tracking suit.
     /// </summary>
-    [Flags]
```
StephenHodgson: How does this change affect classes like `ControllerFinder`?

Assets/MixedRealityToolkit/_Core/Definitions/InteractionDefinitionMapping.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Devices;
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
+{
+    // TODO - Need to think about adding filtering to the Controller definition?
+    /// <summary>
+    /// A mapping construct, linking the Physical inputs of a controller to a Logical construct in a runtime project
+    /// </summary>
+    public struct InteractionDefinitionMapping
```
StephenHodgson: Isn't this the same as `InteractionDefinition`? Does this replace it?

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityDevice.cs
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
+{
+    /// <summary>
+    /// Mixed Reality Toolkit device definition, used to instantiate and manage a specific device / SDK
+    /// </summary>
+    public interface IMixedRealityDevice
```
StephenHodgson: Not sure if this is needed. Windows doesn't give us specific devices to choose from, but only controllers, gamepads, and their data.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
SimonDarksideJ: It's the simplest way to provide 1-1 mapping with a physical input without lengthy lookups or other costly mechanisms.  But agree that we should try and iterate on this later, for now, this works for the MVP.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/InputAction.cs
SimonDarksideJ: Axis is Optional, which is why there was an overload for the AxisType.  I left it off orrigially as it broke the Inspector, just added the overload and it's now fine?  Oh well.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
SimonDarksideJ: Renamed

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealitySpeechCommandsProfile.cs
SimonDarksideJ: Lol, my bad, moved

Assets/MixedRealityToolkit/_Core/Extensions/CollectionsExtensions.cs
SimonDarksideJ: Yes, because the values we set are generic and this now uses the generic SET method you created :D 
Else, I would need to create overloaded versions for each value type.

Assets/MixedRealityToolkit/_Core/Extensions/InputActionsExtensions.cs
SimonDarksideJ: It was added orriginally when we had an InputActions array, it is a little redundant now. However, I'd like to keep it for now as we haven't yet fully settled and re can remove this in the Alpha pass

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: I moved them with the other Manager management code and put them in the order in which they are used there:

* Container Add
* Initialise
* ...

Assets/MixedRealityToolkit/_Core/Definitions/Handedness.cs
SimonDarksideJ: We'll need to re-evaluate controller finder and adapt when we get to that implementation.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityDevice.cs
SimonDarksideJ: It's the generic device interface for ALL SDK's not just Windows

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WMRDevice.cs
SimonDarksideJ: renamed to "WindowsMixedRealityDevice" (it's a device not a manager :D)

Assets/MixedRealityToolkit/_Core/Devices/Generic/XBoxController.cs
SimonDarksideJ: Done

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
StephenHodgson: Thinking the events we raise to the managers should be in their own region

SimonDarksideJ:
Note, this branch is mainly for test only now.  Following the recent changes, I'll likely have to re-implement this against the base vNext once #2204 is merged.

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityControllerConfigurationProfile.asset
```diff
@@ -67,6 +67,19 @@ MonoBehaviour:
...
       rotationData: {x: 0, y: 0, z: 0, w: 0}
       changed: 0
+    - id: 0
+      axisType: 7
+      inputType: 3
+      inputAction:
+        id: 0
+        description: 
+        axisConstraint: 0
+      boolData: 0
```
SimonDarksideJ: I'm thinking @StephenHodgson that we shouldn't be serializing these data fields.  Thoughts?

StephenHodgson: That's correct.

SimonDarksideJ: OK, will update with the next batch

StephenHodgson:
Tests aren't passing:
image

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityControllerConfigurationProfile.asset
StephenHodgson: After looking at this again, yes, we do want to make sure we're serializing these fields.

SimonDarksideJ:
Will retest and validate post the crash fixes

Assets/MixedRealityToolkit-Tests/MixedRealityManagerTests.cs
```diff
@@ -283,7 +283,8 @@ private static void InitializeMixedRealityManager()
 
             // Test the Manager
             Assert.IsNotNull(MixedRealityManager.Instance);
-            MixedRealityManager.Instance.ActiveProfile = ScriptableObject.CreateInstance<MixedRealityConfigurationProfile>();
+            var configuration = ScriptableObject.CreateInstance<MixedRealityConfigurationProfile>();
+            MixedRealityManager.Instance.ActiveProfile = configuration;
```
StephenHodgson: why break this up onto different lines?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -380,7 +385,7 @@ public void AddManager(Type type, IMixedRealityManager manager)
             else
             {
                 MixedRealityComponents.Add(new Tuple<Type, IMixedRealityManager>(type, manager));
-                manager.Initialize();
+                if(!isMixedRealityManagerInitializing) manager.Initialize();
```
StephenHodgson: braces

StephenHodgson: also, seeing as this is a singleton, how are we initializing more than once at a time?

Assets/MixedRealityToolkit-Tests/MixedRealityManagerTests.cs
SimonDarksideJ: Helped with debugging to find out what default options were set. Could also probably add test cases for the config values.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
SimonDarksideJ: It's not the singleton that was being initialized twice, is was the Runtime DeviceManager.

In MRManager.Initialise, we add managers and then call InitializeAllManagers.
BUT, when you add a RunTime manager, we also initialize it (as it wasn't expected to add runtime managers during MRManager initialization, now we do)

SO we need to ensure Runtime managers added on MixedRealityManager startup don't initialize themselves.  But it's needed at runtime of course.

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityControllerConfigurationProfile.asset
SimonDarksideJ: Make up your mind :P 

But we need to find a good use case for serialising them.  These are input values direct from the controller, serialising doesn't help unless you want to enable some sort of suspend / resume.  But on resume, any serialized values will be immediately replaced with the next controller values.

SimonDarksideJ:
Success, latest test with the Input System enabled now work and don't crash

ProjectSettings/EditorBuildSettings.asset
```diff
@@ -4,4 +4,8 @@
 EditorBuildSettings:
   m_ObjectHideFlags: 0
   serializedVersion: 2
-  m_Scenes: []
```
StephenHodgson: please revert.

Assets/MixedRealityToolkit/_Core/Extensions/InputActionsExtensions.cs
StephenHodgson: Yeah I'd like to discourage people trying to look for actions by string

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityDevice.cs
```diff
@@ -1,35 +1,14 @@
...
     /// <summary>
     /// Mixed Reality Toolkit device definition, used to instantiate and manage a specific device / SDK
     /// </summary>
-    public interface IMixedRealityDevice
+    public interface IMixedRealityDevice : IMixedRealityManager
     {
-        /// <summary>
-        /// Initialize function to setup the device, called after all managers are initialized by the Mixed Reality Manager
-        /// </summary>
-        void Initialize();
```
StephenHodgson: Did we also remove these from all the scaffolding?

SimonDarksideJ: It wasn't removed, it was simply replaced with the same methods in the Manager interface (which actually had more)

ProjectSettings/EditorBuildSettings.asset
SimonDarksideJ: Reverted

Assets/MixedRealityToolkit-Tests/MixedRealityManagerTests.cs
```diff
@@ -53,8 +53,8 @@ public void Test03_CreateMixedRealityManager()
             InitializeMixedRealityManager();
 
             // Create Test Configuration
-            Assert.IsEmpty(MixedRealityManager.Instance.ActiveProfile.ActiveManagers);
-            Assert.IsEmpty(MixedRealityManager.Instance.MixedRealityComponents);
+            Assert.AreEqual(MixedRealityManager.Instance.ActiveProfile.ActiveManagers.Count, 0);
+            Assert.AreEqual(MixedRealityManager.Instance.MixedRealityComponents.Count, 1);
```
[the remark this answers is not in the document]
SimonDarksideJ: Sorted

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -1,39 +1,251 @@
...
         public Handedness ControllerHandedness { get; }
 
-        public IMixedRealityInputSource InputSource { get; }
+        public IMixedRealityInputSource InputSource { get; private set; }
+
+        public Dictionary<DeviceInputType, InteractionMapping> Interactions { get; private set; }
+
+        public void SetupInputSource<T>(T state)
+        {
+            InteractionSourceState interactionSourceState = CheckIfValidInteractionSourceState(state);
```
davidkline-ms: No behavior change if CheckIfValid fails?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -1,39 +1,251 @@
...
+            SetupFromInteractionSource(interactionSourceState);
+        }
+
+        public void UpdateInputSource<T>(T state)
+        {
+            InteractionSourceState interactionSourceState = CheckIfValidInteractionSourceState(state);
```
davidkline-ms: No behavior change if CheckIfValid fails?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -1,39 +1,251 @@
...
+            {
+                Interactions.Add(DeviceInputType.SpatialGrip, new InteractionMapping(3, AxisType.SixDoF, DeviceInputType.SpatialGrip, inputActions.GetActionByName("Grip")));
+
+                Interactions.Add(DeviceInputType.GripPress, new InteractionMapping(4, AxisType.SingleAxis, DeviceInputType.GripPress, inputActions.GetActionByName("Grab")));
+            }
+
+            // If the controller has a menu button, add it to the controller capabilities
+            if (interactionSourceState.source.supportsMenu)
+            {
+                Interactions.Add(DeviceInputType.Menu, new InteractionMapping(5, AxisType.Digital, DeviceInputType.Menu, inputActions.GetActionByName("Menu")));
```
davidkline-ms: It feels really odd to have a button described as an axis. I can understand the trigger as it could tell you how far it was depressed, but a basic button?

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
```diff
@@ -11,6 +13,34 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem
...
+            "Pointer",
+            "Walk",
+            "Look",
+            "Interact",
+            "Pickup",
+            "Inventory",
+            "ConversationSelect"
+        }; // Examples only, to be refined later.
+
+        private readonly AxisType[] defaultInputActionsAxis =
```
davidkline-ms: Could we possibly rename this from AxisType to ControlType? That would allow us to have basic buttons, axes, analog triggers, etc.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
```diff
@@ -11,6 +13,34 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem
...
     public class MixedRealityInputActionsProfile : ScriptableObject
     {
+        private readonly string[] defaultInputActions =
+{
+            "Select",
+            "Menu",
+            "Grip",
+            "Pointer",
+            "Walk",
+            "Look",
```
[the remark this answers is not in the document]
StephenHodgson: These are just examples.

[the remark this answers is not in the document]
StephenHodgson: I'm wondering if we even want to create default ones for people.  We will have the default that they can copy/start from.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -184,5 +185,36 @@ void ISerializationCallbackReceiver.OnAfterDeserialize()
...
+            {
+                if (ControllersProfile.MixedRealityControllerMappingProfiles[i].Controller == systemType && ControllersProfile.MixedRealityControllerMappingProfiles[i].Handedness == hand)
+                {
+                    return ControllersProfile.MixedRealityControllerMappingProfiles[i];
+                }
+            }
+            return default(MixedRealityControllerMapping);
+        }
+
+        // TODO - Which is better?
```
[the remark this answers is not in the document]
StephenHodgson: We could make it a static and include this to the Type controller parameter so we don't have to pass in the type as well.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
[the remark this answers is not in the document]
davidkline-ms: We should consider at least "Select" as a default as that allows users to click and should be available for all controllers (including hands and simple clickers)

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
```diff
@@ -0,0 +1,482 @@
...
+  m_StaticEditorFlags: 0
+  m_IsActive: 1
+--- !u!4 &1941297680
+Transform:
+  m_ObjectHideFlags: 0
+  m_PrefabParentObject: {fileID: 0}
+  m_PrefabInternal: {fileID: 0}
+  m_GameObject: {fileID: 1941297679}
+  m_LocalRotation: {x: -0, y: -0, z: -0, w: 1}
+  m_LocalPosition: {x: 0, y: 1, z: -10}
```
keveleigh: The controller position is way off into space because the Head object is at (0, 1, -10). Updating this to (0, 0, 0), as well as reducing the cube to a MUCH smaller size haha ((0.1, 0.1, 0.1) worked fine), the controllers show up correctly.

Moving to (0, 0, 0) is a temporary fix though, as we'll need to take into account an object on top of the camera moving it, via teleportation or something similar.

In master, we use [`CameraCache.Main.transform.parent.TransformPoint`](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit/Input/Scripts/InputSources/InteractionInputSource.cs#L737) to take this into account.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
keveleigh: It follows Unity's terminology, where they're all "input axes", but the type can be different things:
image

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
davidkline-ms: Nice find, Kurtis (aka @keveleigh)!

keveleigh: Now with gif:

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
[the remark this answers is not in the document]
StephenHodgson: Of course we will have defaults, but I think they should be pre-serialized in an asset when we release.

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
SimonDarksideJ: Great find @keveleigh 
In the Controller code it is transforming using ``` CameraCache.Main.transform.parent.TransformPoint```

If you look in 
https://github.com/SimonDarksideJ/MixedRealityToolkit-Unity/blob/867acc556d5e4b1b827e2dd18faf8e3e29ef60a3/Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs#L189

You'll see that is transformed already.  Obviously the default setup I created was slightly flawed.
So we'll need to think of a generic way to correct this automatically for projects.

Don't recall ever seeing any such guidance regarding scale for creating custom hands, so we'll likely need some docs on it.  The Cube's I initially created we the standard cube (e.g 1m, which might explain their size :D)

Great find and glad it was something so simple.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
SimonDarksideJ: Axis is correct in my understanding @davidkline-ms , as it describes the type of input data, from Axis 1d through to Axis 6d.  Digital is the only exception as it's an on/off state, pressed or not.

We can only go so far down the generic route when describing these things.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
SimonDarksideJ: Good point, will review

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
SimonDarksideJ: Good point, will review

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -184,5 +185,36 @@ void ISerializationCallbackReceiver.OnAfterDeserialize()
         }
 
         #endregion  ISerializationCallbackReceiver Implementation
+		
+		#region Mixed Reality Controller Mapping helpers
+
+        // TODO - needs validation to ensure duplicates are not added?
```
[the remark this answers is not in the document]
SimonDarksideJ: in a TODO for a later PR, iterating on what exactly we need

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
```diff
@@ -34,5 +64,42 @@ public class MixedRealityInputActionsProfile : ScriptableObject
         /// The action to use for pointing events.
         /// </summary>
         public InputAction PointerAction => pointerAction;
+
+        /// <summary>
+        /// Reset the current InputActions definitions to the Mixed Reality Toolkit defaults
+        /// If existing mappings exist, they will be preserved and pushed to the end of the array
+        /// </summary>
+        /// <returns>Default MRTK Actions plus any custom actions (if already configured)</returns>
```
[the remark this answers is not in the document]
SimonDarksideJ: The defaults are for only if you do not create any.
It's a different task to update the inspector to load defaults.  @StephenHodgson  and I are still testing / toying with the correct pattern for this.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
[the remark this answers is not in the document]
SimonDarksideJ: The defaults need to move to "Per-Controller" anyway, these are just test for now.  As each controller type will have it's own defaults :D

Assets/MixedRealityToolkit/_Core/Definitions/Devices/Headset.cs
```diff
@@ -5,6 +5,7 @@
 
 namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.Devices
 {
+    // TODO - currently not used, consider removing maybe?
```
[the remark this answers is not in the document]
SimonDarksideJ: Fair enough, we aren't using it yes as the device doesn't initialise it.. yet.  But we can review in a later PR as we iterate forward

SimonDarksideJ:
One comment to all, we need to ensure we only fix / change what is critically necessary in this PR.  We need this as a baseline to then iterate forward, starting from what works (now thanks to @keveleigh ) and making subtle but necessary improvements.

We desperately need to move away from such big PR's as this.

Next steps after this PR include:

* Implement controllers for the remainder of the functionality in the old InteractionInputSources class (e.g. Gesture and Navigation)
* Iterating on controller design as it applies to other SDK's
* Implement some of the SDK components that consume the InputSystem Actions

This PR is only the start of that journey.

So the only fix above I see is crucial, is @davidkline-ms point on the CheckIfValidInteractionSourceState(), which I'll address.

Thoughts?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
davidkline-ms: I can live with it, then :)

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
SimonDarksideJ: Double checked.  The function will only ever return a valid InteractionSource state.  If the type of the parameter passed in is not an InteractionSourceState, it will throw an ArgumentOutOfRangeException.

Don't think we need anything more there really.  In reality, the check isn't needed, but I like to walk on the safe side and be sure.

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
keveleigh: @SimonDarksideJ Ahh yep, I realized after I commented that it actually isn't the teleportation issue I as I originally thought (especially since you already took that into account!), but actually an issue with the camera itself starting at a non-(0,0,0) location. I just checked, and we have this same problem in master too.

davidkline-ms:
@SimonDarksideJ, Yep. I'll happily approve after that fix :)

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
StephenHodgson: That's partly why I wanted to use what we already have, but I'm glad it's working now.

StephenHodgson:
One other thing to note, it that the camera position isn't in the center of the world like it should be in your example scene.

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
keveleigh: > I wanted to use what we already have

Have = have in master or somewhere else?

keveleigh:
@StephenHodgson

>the camera position isn't in the center of the world

Yeah, that's what we were talking about above that caused the issue with the controllers. The code (and Unity's API) assumes the camera starts at the origin. We don't handle this case in master either, but assume the camera starts at the origin.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
StephenHodgson: Goes back to that essence of input I talked about.

Technically digital isn't an axis per se but it's still the base and subset of all the things you can do with a signal. Essentially digital is just a clamped sinusoidal with a ceiling and floor value.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDevice.cs
```diff
@@ -0,0 +1,466 @@
...
+            var detectedController = new WindowsMixedRealityController(
+                ControllerState.NotTracked,
+                controllingHand,
+                inputSystem.RequestNewGenericInputSource($"Mixed Reality Controller {controllingHand}"),
+                new Dictionary<DeviceInputType, InteractionMapping>()
+                );
+
+            detectedController.SetupInputSource(interactionSourceState);
+
+            activeControllers.Add(interactionSourceState.source.id, detectedController);
```
StephenHodgson: This is the boxing I'm taking about. Not a big deal.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionsProfile.cs
SimonDarksideJ: See above, the terminology is correct

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDevice.cs
davidkline-ms: I'd like to avoid boxing if at all possible... It can generate garbage VERY quickly.

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
SimonDarksideJ: Just confirmed and replicated locally.  Noticed a few interesting points we'll have to educate devs about, such as visualizer position (seems centered on the Touchpad), camera settings, etc.

It also highlights the need for:
* An SDK component to create a default "Rig"
* A Rig validator, to avoid the issues detected here.
Plus some other stuff, going to create some tasks for "next steps" once this is merged

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDevice.cs
StephenHodgson: I think it's not that big a deal here since we only create a few devices at a time. Generally this won't happen that often.

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
StephenHodgson: I meant the `InteractionInputSources` class. But @SimonDarksideJ has assured me we will do an A/B test to make sure nothing falls between the cracks.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -261,12 +219,12 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialGrip) && controller.Interactions[DeviceInputType.SpatialGrip].Changed)
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions[DeviceInputType.SpatialGrip].Changed)
                 {
                     inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, controller.Interactions[DeviceInputType.SpatialGrip].GetValue<Tuple<Vector3, Quaternion>>());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.TouchpadTouch) && controller.Interactions[DeviceInputType.TouchpadTouch].Changed)
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions[DeviceInputType.TouchpadTouch].Changed)
```
StephenHodgson: doesn't this cause that crash again?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -156,13 +159,14 @@ private IMixedRealityController GetOrAddWindowsMixedRealityController(Interactio
...
+            IMixedRealityInputSource controllerInputSource = inputSystem?.RequestNewGenericInputSource($"Mixed Reality Controller {controllingHand}");
 
             //TODO - Controller Type Detection?
             //Define new Controller
             var detectedController = new WindowsMixedRealityController(
                 ControllerState.NotTracked,
                 controllingHand,
-                inputSystem.RequestNewGenericInputSource($"Mixed Reality Controller {controllingHand}"),
+                controllerInputSource,
                 new Dictionary<DeviceInputType, InteractionMapping>()
```
StephenHodgson: we can just pass null here, correct?

StephenHodgson: or nothing cause null is default

SimonDarksideJ: If the RequestNewGenericInputSource returns null, then it will be null.
That was the pattern we settled on, unless you want to alter it now?

StephenHodgson: in the constructor it will create a new dictionary buy default is all I'm saying. So you don't have to make a new one here.

SimonDarksideJ:
#Falls off chair :D

Assets/MixedRealityToolkit-Tests/Scenes/BasicControllerView/01-Basic.unity
keveleigh: >such as visualizer position (seems centered on the Touchpad)

Assuming we're using the pointer position, it's centered at the tiny sphere here:
ezgif-2-3ac46a4ce7

StephenHodgson: Real question is how'd you get the Samsung device to show up?

PR #2187, opened 2018-05-28
New APIs that enable access to MS Graph

joaocpaiva:
**Overview**
1. Adds APIs that enable access to MS Graph for Read operations. Other C(R)UD operations will come later.
2. Adds default authentication provider that uses MSAL to allow user authentication. Developers have the option to override with their own implementation though.

**Disclaimer**
This is an early review as I am looking for guidance to ensure compliance with MRTK-Version-Next best practices, including:
1. Folder and file structure.
2. Namespaces and code organization.
3. Guideline to incorporate plugins such as MSAL and JSon. Are plugins permitted at all?

**Known issue**
MSAL is currently not working when using IL2CPP, which seems to be a bug in Unity. I will be tracking this issue with Unity team and to start I created a new thread in their [forums](https://forum.unity.com/threads/uwp-assemblies-not-working-when-using-il2cpp-but-work-on-net-scripting-backend.533401/).


StephenHodgson:
> MSAL is currently not working when using IL2CPP

Unfortunately that's what we're currently targeting for vNEXT.

Assets/MixedRealityToolkit-Examples/Graph/GraphConnectorTest.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    /// Class to test GraphConnector using Unity coroutine model.
+    /// </summary>
+    public class GraphConnectorTest : MonoBehaviour
+    {
+        /// <summary>
+        /// Called by Unity on the frame when the script is enabled just before any of the Update methods is called the first time.
+        /// </summary>
+        /// <returns>The Unity coroutine.</returns>
+        private IEnumerator Start()
```
StephenHodgson: async?

Assets/MixedRealityToolkit-Examples/Graph/GraphConnectorTestAsync.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    /// <summary>
+    /// Class to test GraphConnector using async model.
+    /// </summary>
+    public class GraphConnectorTestAsync : MonoBehaviour
+    {
+        /// <summary>
+        /// Called by Unity on the frame when the script is enabled just before any of the Update methods is called the first time.
+        /// </summary>
+        private async void Start()
```
StephenHodgson: Ah there's two! disregard my last comment then 😄

Assets/MixedRealityToolkit-Examples/Graph/GraphProfile.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Examples
+{
+    /// <summary>
+    /// Class that represents a partial Graph profile.
+    /// </summary>
+    public class GraphProfile
```
StephenHodgson: can this be a struct?

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        /// <summary>
+        /// Makes POST request via Graph's REST APIs.
+        /// </summary>
+        /// <typeparam name="T">The returned type for graph response.</typeparam>
+        /// <param name="graphRequest">The graph request to make.</param>
+        /// <returns>The Unity coroutine.</returns>
+        public IEnumerator MakeRequestPost<T>(GraphRequest<T> graphRequest) where T : GraphResponse, new()
+        {
+            using (UnityWebRequest webRequest = UnityWebRequest.Post(graphRequest.Uri, graphRequest.Body))
```
StephenHodgson: Let's use the REST library found in MixedRealityToolkit/_Core/Utilities/WebRequestRest

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// <param name="graphRequest">The graph request to make.</param>
+        /// <param name="webRequest">The web request.</param>
+        /// <returns>The Unity coroutine.</returns>
+        private IEnumerator MakeRequest<T>(GraphRequest<T> graphRequest, UnityWebRequest webRequest) where T : GraphResponse, new()
+        {
+            string authToken = string.Empty;
+
+            try
+            {
+#if UNITY_EDITOR
```
StephenHodgson: `if(Application.isEditor)`

Assets/MixedRealityToolkit/Graph/GraphResponse.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Networking;
+
+namespace Microsoft.MixedReality.Toolkit.Graph
+{
+    /// <summary>
+    /// Class that represents a Graph response.
+    /// </summary>
+    public class GraphResponse
```
StephenHodgson: could this be a struct?

Assets/MixedRealityToolkit/Graph/Plugins/Newtonsoft.Json.dll.meta
```diff
@@ -0,0 +1,103 @@
+fileFormatVersion: 2
+guid: abc5aed9aec861241ac2df4d98e3d04b
```
StephenHodgson: We're not going to be putting this dll into vNEXT. It causes too much trouble with other packages that also include it.

We should be using the `Unity.JsonUtility` where possible.

Assets/MixedRealityToolkit-Examples/Graph/GraphConnectorTestAsync.cs
StephenHodgson: Since we're specifying that vNEXT requires .NET 4.x support we can take out the IEnumerator version of this class.

Assets/MixedRealityToolkit/Graph/GraphResponse.cs
joaocpaiva: This one can't because it has virtual methods.

Assets/MixedRealityToolkit/Graph/Plugins/Newtonsoft.Json.dll.meta
joaocpaiva: Agreed!!

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
joaocpaiva: Oh, I like it.

joaocpaiva:
@StephenHodgson it's confirmed that the issue with MSAL in IL2CPP is a Unity bug. 

I filed a bug in the Unity data base and status can be tracked [here](https://fogbugz.unity3d.com/default.asp?1043171_p9ksn5jtpcmqntsi). I think for now we can assume this will be fixed in Unity and commit the code as is (knowing we have a temporary known issue)?

StephenHodgson:
@joaocpaiva that's not the only IL2CPP issue: [I've got one open for .NET Standard 2 as well](https://forum.unity.com/threads/il2cpp-net-standard-2-and-missing-language-features.533022)

Assets/MixedRealityToolkit/Graph/README.md
```diff
@@ -0,0 +1,59 @@
...
+	* **Graph Access Scopes** is the array that lists all [access permissions](https://developer.microsoft.com/en-us/graph/docs/concepts/permissions_reference) required in your scenario.
+
+7. All set. 
+
+	Look up **GraphConnectorTestAsync.cs** for examples on how to use the GrahConnector.
+
+# Managed bytecode stripping with IL2CPP
+IL2CPP analyzes all assemblies and removes methods that are never directly called. If something is only accessed through reflection, it will be removed unless it is specified in link.xml files. Read more about it [here](https://docs.unity3d.com/Manual/IL2CPP-BytecodeStripping.html) 
+
+# Known issues
```
joaocpaiva: @StephenHodgson I tracked down these two bugs in Unity that are preventing MSAL from working on IL2CPP. Once these bugs are fixed, this library works.

I'd like to wrap-up this PR as soon as possible. Should we proceed with this change with that caveat (this won't work until those bugs are fixed in Unity) or do you have some other suggestion?

Assets/MixedRealityToolkit/Graph/README.md
```diff
@@ -0,0 +1,59 @@
...
+Two Unity bugs prevent Microsoft.Identity.Client.dll from working in UWP:
+
+	1. IL2CPP project failing to P/Invoke into kernel32.dll!GetNativeSystemInfo. 
+
+		Issue Tracker: https://issuetracker.unity3d.com/issues/uwp-il2cpp-project-failing-to-p-slash-invoking-into-etw-logging-functions
```
joaocpaiva: @StephenHodgson I tracked down these two bugs in Unity that are preventing MSAL from working on IL2CPP. Once these bugs are fixed, this library works.

I'd like to wrap-up this PR as soon as possible. Should we proceed with this change with that caveat (this won't work until those bugs are fixed in Unity) or do you have some other suggestion?

StephenHodgson: I'll be sure to pull your branch and do a more in depth review this weekend.

StephenHodgson: Hey @joaocpaiva I saw that 2018.1.6 is out with fixes for this. Going to do a dry run and see if I can get it setup.

StephenHodgson:
@joaocpaiva everything seemed to work well for me. I wanna do a pass on where this code should live and set up a custom assembly definition for it to match the pattern in the rest of vNEXT.  Please keep your eyes open for a PR soon.

Assets/MixedRealityToolkit/Graph/Plugins/WSA/Microsoft.Identity.Client.dll.meta
```diff
@@ -0,0 +1,114 @@
+fileFormatVersion: 2
+guid: cbf1e91417ffe6c46b7e1b1e444ef8a7
```
StephenHodgson: Where does this dll come from?

Is there source code that goes along with this?

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// Beta Graph endpoint.
+        /// </summary>
+        public static readonly string GraphBetaEndpoint = "https://graph.microsoft.com/beta/";
+
+        /// <summary>
+        /// The application id registered to access the Graph.
+        /// </summary>
+        [SerializeField]
+        [Tooltip("The App Id registered in https://apps.dev.microsoft.com to access the Graph.")]
+        private string graphAppId = "83f5c300-1163-48b5-9790-f7a86d02349c";
```
StephenHodgson: should we remove this and add a check if string is null or empty?

Assets/MixedRealityToolkit/Graph/Plugins/WSA/Microsoft.Identity.Client.dll.meta
StephenHodgson: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet

StephenHodgson: I wonder if there is a nice way to reference this dependency via GitHub.

My only concern here is if people already use a library in their projects and they get conflicts.

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
```diff
@@ -0,0 +1,121 @@
...
+using Microsoft.MixedReality.Toolkit.Graph.Internal;
+using System.Threading.Tasks;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Graph
+{
+    /// <summary>
+    /// Unity script that enables access to the Graph.
+    /// </summary>
+    public class GraphConnector : MonoBehaviour
```
StephenHodgson: Does this need to inherit from Mono?

Assets/MixedRealityToolkit/Graph/link.xml
```diff
@@ -0,0 +1,13 @@
...
+	<assembly fullname="System.Runtime.Serialization" preserve="nothing">
+		<type fullname="System.Runtime.Serialization.Configuration.DataContractSerializerSection" preserve="all" />
+		<type fullname="System.Runtime.Serialization.Configuration.SerializationSectionGroup" preserve="all" />
+		<type fullname="System.Runtime.Serialization.Configuration.DeclaredTypeElementCollection" preserve="all" />
+		<type fullname="System.Runtime.Serialization.XmlObjectSerializerWriteContext" preserve="all" />
+		<type fullname="System.Runtime.Serialization.XmlWriterDelegator" preserve="all" />
+	</assembly>
+
+	<assembly fullname="Microsoft.Identity.Client">
+		<type fullname="Microsoft.Identity.Client.Internal.*" preserve="all" />
```
StephenHodgson: Can you explain what this does please?

Assets/MixedRealityToolkit/Graph/Prefabs/GraphConnector.prefab
```diff
@@ -0,0 +1,57 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
```
StephenHodgson: Any prefabs should live in the SDK layer.

Assets/MixedRealityToolkit/Graph/README.md
StephenHodgson: Tried this out but getting null references in your scene, is the auth correct for the example?

image

StephenHodgson: I'm also getting this 
image

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        /// <summary>
+        /// Makes GET request via Graph's REST APIs. 
+        /// </summary>
+        /// <typeparam name="T">The returned type for graph response.</typeparam>
+        /// <param name="endpoint">The Graph endpoint to call.</param>
+        /// <param name="api">The Graph API to call.</param>
+        /// <returns>The async task.</returns>
+        public async Task<T> MakeRequestGetAsync<T>(string endpoint, string api) where T : GraphResponse, new()
```
StephenHodgson: I'd like to see this only return a Response so that devs can use it as they please. I think having graph response types constrains this a bit.  I think it's best to move that stuff to examples in case people want to know how to utilize and get content from the requests.

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            string authToken = string.Empty;
+
+            if (Application.isEditor)
+            {
+                authToken = testAuthToken;
+
+                // Passthrough all HTTPS traffic in the Unity Editor. Should not be used in production code.
+                System.Net.ServicePointManager.ServerCertificateValidationCallback += (o, certificate, chain, errors) => true;
```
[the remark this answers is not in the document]
StephenHodgson: I think this is bc of a known TLS bug in Unity.

Assets/MixedRealityToolkit/Graph/link.xml
StephenHodgson: Nvm it was in the readme that it's to prevent stripping byte code.

Assets/MixedRealityToolkit/Graph/README.md
joaocpaiva: Yes, MSAL does not work in the Unity editor because the SDK requires web browser which is not available in Unity.

There are two options as workaround to test Graph in the editor:
1. Use network monitor like Fiddler, to sneak peak the Bearer token and copy/paste into GraphConnector inspector as the Test Auth Token
2. Provide custom implementation of IGraphAuthentication instead of using MsalGraphAuthentication. Obviously a real implementation is not trivial (requires OAuth, UI for users to enter credentials, etc).

Furthermore, in future releases of MSAL this scenario should be unblocked using the device code flow (which enables us to sign-in from a phone for instance), or another device. But Device code flow is not yet implemented in MSAL.Net (this is coming).

Do you have a particular concern about leaving this as is, knowing these caveats. I guess I can elaborate a bit more in the README.md?

Assets/MixedRealityToolkit/Graph/Plugins/WSA/Microsoft.Identity.Client.dll.meta
joaocpaiva: @StephenHodgson I also hate the fact that Unity does not have a nice way to reference NuGet packages, or am I wrong about that? 

Options I can think of: we make this DLL disabled by default in Unity, so it does not cause any conflicts whatsoever, until users explicitly enable it. Or we don't even include the DLL and we mention in the documentation that developers require MSAL DLL to use MsalGraphAuthentication.cs, provided with the SDK?

StephenHodgson: We could add a auto downloader if they so choose to enable these features

StephenHodgson: It would just grab the latest from the github releases

joaocpaiva: @StephenHodgson how do I hook to the auto downloader?

StephenHodgson: We would have to write something in Unity that checks if they already have the dll in their project and then grab it from a direct link on github.  Or we could use a submodule approach and take it in directly.

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
joaocpaiva: The reason why it's inheriting from Mono is to allow user to set the Graph settings through the Unity editor, using the inspector to make it Unity friendly. I will be fine to change if you feel strong about it?

Currently there are only 3 settings (App Id, Access Scopes, Auth Token) and there might be more to add in the future to support more (uncommon) authentication scenarios (e.g. Authority, RedirectUri, Extra Scopes, etc).

Assets/MixedRealityToolkit/Graph/link.xml
joaocpaiva: I will add an XML comment in place!

Assets/MixedRealityToolkit/Graph/Plugins/WSA/Microsoft.Identity.Client.dll.meta
joaocpaiva: Got it. We really need a system to reference NuGets from Unity. The change you suggest makes totally sense to me but I don't have the cycles to do it. Can we leave the DLL embedded for now and create a separate work item to create a system that allows managing NuGets that everyone can benefit and hook in? Honestly, I hate the fact that Unity does not integrate with NuGet yet.

By the way, I stumbled up on https://github.com/GlitchEnzo/NuGetForUnity. Might be interesting to take a look.

Assets/MixedRealityToolkit/Graph/GraphConnector.cs
StephenHodgson: You can do that with the ScriptableObjects as well. Please revisit my PR against your branch.

StephenHodgson: I don't think this class should be a monobehaviour

Assets/MixedRealityToolkit/_Core/Utilities/MotionController/MotionControllerVisualizer.cs
```diff
@@ -17,6 +17,7 @@
 //using UnityGLTF;
 
 #if !UNITY_EDITOR
+using Microsoft.MixedReality.Toolkit.Internal.Extensions;
```
StephenHodgson: Please revert.

joaocpaiva:
@StephenHodgson I just saw your PR, but after already had submitted a new commit with a refactor to use ScriptableObject .. so it is not based on your proposal :( What do you think? Let me know. But I am happy to change it to match your proposal.

joaocpaiva:
@StephenHodgson I pushed new changes that incorporate your earlier proposal for the Scriptable Objects. Please have a look once you get a chance.

StephenHodgson:
Will do!

davidkline-ms:
Marking blocked until I can sync with @joaocpaiva.

Assets/MixedRealityToolkit-SDK/Features/Input/Prefabs.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: ddec047ef9d7e3740aaa8ecceb9a9636
+guid: 085c5b2bce218574298073b8375efab6
```
StephenHodgson: We should probably revert this change, as it is unrelated to Graph.

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityConfigurationProfile.asset
```diff
@@ -25,3 +25,5 @@ MonoBehaviour:
   enableControllerProfiles: 1
   controllersProfile: {fileID: 11400000, guid: 4319257468937c74ebb69e41792d60c5, type: 2}
   enableBoundarySystem: 1
+  enableGraphAccessProfile: 1
```
StephenHodgson: Do we want to enable this by default?

joaocpaiva: Yeah, this file should be reverted, my mistake.

StephenHodgson: I'm not sure I want to add this to the confirmation profile. In my pr I just assigned a profile from the project window onto my script that needed to know about graph.

I suppose this is something we can discuss later on.

joaocpaiva: Sorry, I meant, Yes - you are right :) I should have reverted this file before pushing last changes, but I will do that next.

davidkline-ms:
Marked as Do Not Merge. I will be getting together with @joaocpaiva soon to discuss.

PR #2188, opened 2018-05-28
Fixed some issues with handedness declarations.

StephenHodgson:
Overview
---
Values were wrong, and added Both enum.


StephenHodgson:
Should work fine in dropdowns

PR #2198, opened 2018-05-29
Moving definitions following review

SimonDarksideJ:
Overview
---
Moving specific definitions to the Utilities following review
Also removing the Controller definition as it was replaced with an interface and updating the Handedness enum.

Changes
---
Moves:
* AxisType
* Handedness
* RotationConstraintType

Handedness enum updated following review
*Note, error will occur in Controller struct which is being removed in #2197


Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Handedness.cs
```diff
@@ -30,6 +30,11 @@ public enum Handedness : byte
         /// </summary>
         Both = Left | Right,
         /// <summary>
+        /// Global catchall, used to map actions to any controller (provided the controller supports it)
+        /// *Note, by default the specific hand actions will override settings mapped as both
+        /// </summary>
+        Any = None | Left | Right | Other,
```
StephenHodgson: Not sure if this is needed.

SimonDarksideJ: I like catch-alls and also good if I want to configure just one mapping for All controllers / SDK types

Issue #2166, opened 2018-05-25
Can I Change length of Motion Controller pointing ray when if lost focus from a object?


TakahiroMiyaura:
## Overview
Can I Change length of Motion Controller pointing ray when if lost focus from a object?
I feel short it in the setting of MRTK2017.4.0.0-RC1. 
When we lost focus from a object,we would be better to "Ray" is pointing far as possible.  We will be easer to understand the state of lost focus .

![1](https://user-images.githubusercontent.com/8484268/40525537-8f735b3c-601b-11e8-8873-c171601c66e4.png)

## Steps to reopen 
Open scene "InputNavigationRotateTest.unity" and execution in HoloToolkit-Unity-2017.4.0.0-rc.unitypackage.

## Unity Editor Version
Unity 2017.4.3f1
## Mixed Reality Toolkit Release Version
MRTK 2017.4.0.0-RC1

gobanjoboy:
It seems I have been able to extend it some, but there seems to be a cap some where. It doesn't really seem to use my 200 value.

![screenshot_1](https://user-images.githubusercontent.com/174855/40552235-b72bead4-6004-11e8-8582-dbf94241bc4f.png)

gobanjoboy:
Okay I figured it out.

Unless I configured something that didn' t need to be set. I found three places that needed to be updated.

![screenshot_1](https://user-images.githubusercontent.com/174855/40553057-242342b6-6007-11e8-9f90-e0c743d7d0ee.png)
![screenshot_2](https://user-images.githubusercontent.com/174855/40553058-2433ec6a-6007-11e8-8e62-6fa8cc5f6af9.png)

keveleigh:
Yep, Default Cursor Distance is (at the moment) where you'd change it. I'm working on a change that'll make this a little more configurable, especially depending on the type of pointer your cursor is attached to (you likely want the default distance further from the controllers than from your gaze, etc). 

I'll also look into the observed cap above. I'm not sure what that might be.

cre8ivepark:
@TakahiroMiyaura @keveleigh I had similar feedback. Thanks for opening the issue!

gobanjoboy:
@keveleigh the Cap I spoke about was because I was not editing all three spots. I was only changing the DefaultCursor. I don't have any cap issues.

gobanjoboy:
Has there been any talk of scaling the cursor bigger as it gets further way?

StephenHodgson:
[...]

PR #2201, opened 2018-05-29
[May18] Updating pointing rays to extend out to the pointing extent off holograms

keveleigh:
Overview
---
This PR updates the pointing ray visualization to extend out the full length of the pointer's pointing extent (or the FocusManager's default, if no override is specified), instead of the previous cursor default.

This is implemented in a way to preserve the default behavior of the Gaze cursor being placed at 2m, which is closer than the default pointing extent.

Changes
---
- Fixes: #2166.

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/BaseControllerPointer.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/BaseControllerPointer.cs
@@ -11,18 +11,8 @@ public abstract class BaseControllerPointer : AttachToController
     protected float CurrentPointerOrientation;
 
     [SerializeField]
-    [Range(0.5f, 50f)]
-    private float extentOverride = 2f;
-
-    [SerializeField]
     [Tooltip("Source transform for raycast origin - leave null to use default transform")]
     protected Transform RaycastOrigin;
-
-    public float? ExtentOverride
-    {
-        get { return extentOverride; }
-        set { extentOverride = value ?? FocusManager.Instance.GlobalPointingExtent; }
-    }
 
     /// <summary>
     /// The Y orientation of the pointer target - used for touchpad rotation and navigation
--- a/Assets/HoloToolkit/Input/Prefabs/LinearControllerPointer.prefab
+++ b/Assets/HoloToolkit/Input/Prefabs/LinearControllerPointer.prefab
@@ -62,7 +62,6 @@ MonoBehaviour:
   ScaleOffset: {x: 1, y: 1, z: 1}
   SetScaleOnAttach: 0
   CurrentPointerOrientation: 0
-  extentOverride: 2
   RaycastOrigin: {fileID: 0}
   LineColorSelected:
     serializedVersion: 2
--- a/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
@@ -25,13 +25,18 @@ namespace HoloToolkit.Unity.InputModule
             get { return pointer; }
             set
             {
+                // This value is used to determine the cursor's default distance.
+                // It is cached here to prevent repeated casting in the update loop.
                 pointerIsInputSourcePointer = value is InputSourcePointer;
                 pointer = value;
             }
         }
-
         private IPointingSource pointer;
 
+        /// <summary>
+        /// Cached value if the pointer is of type InputSourcePointer,
+        /// to prevent repeated casting in the update loop.
+        /// </summary>
         private bool pointerIsInputSourcePointer = false;
 
         /// <summary>
@@ -122,6 +127,10 @@ namespace HoloToolkit.Unity.InputModule
         private Vector3 targetScale;
         private Quaternion targetRotation;
 
+        /// <summary>
+        /// Keeps track of the starting setting for DefaultCursorDistance,
+        /// to revert after a pointer overrides the value.
+        /// </summary>
         private float originalDefaultCursorDistance;
 
         /// <summary>
--- a/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
@@ -20,7 +20,19 @@ namespace HoloToolkit.Unity.InputModule
         /// <summary>
         /// The pointer that this cursor should follow and process input from.
         /// </summary>
-        public IPointingSource Pointer { get; set; }
+        public IPointingSource Pointer
+        {
+            get { return pointer; }
+            set
+            {
+                pointerIsInputSourcePointer = value is InputSourcePointer;
+                pointer = value;
+            }
+        }
+
+        private IPointingSource pointer;
+
+        private bool pointerIsInputSourcePointer = false;
 
         /// <summary>
         /// Minimum distance for cursor if nothing is hit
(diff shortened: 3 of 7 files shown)


Assets/HoloToolkit/Input/Scripts/Utilities/BaseControllerPointer.cs
```diff
@@ -10,20 +10,10 @@ public abstract class BaseControllerPointer : AttachToController
     [Range(0f, 360f)]
     protected float CurrentPointerOrientation;
 
-    [SerializeField]
-    [Range(0.5f, 50f)]
-    private float extentOverride = 2f;
```
StephenHodgson: This most def should stay here.

Assets/HoloToolkit/Input/Scripts/Utilities/BaseControllerPointer.cs
```diff
@@ -10,20 +10,10 @@ public abstract class BaseControllerPointer : AttachToController
...
-
     [SerializeField]
     [Tooltip("Source transform for raycast origin - leave null to use default transform")]
     protected Transform RaycastOrigin;
 
-    public float? ExtentOverride
-    {
-        get { return extentOverride; }
-        set { extentOverride = value ?? FocusManager.Instance.GlobalPointingExtent; }
```
StephenHodgson: This should stay as well.

Assets/HoloToolkit/Input/Scripts/Utilities/BaseControllerPointer.cs
keveleigh: This class was added by me in the PR that added the visualization. The visualizer doesn't need to know about the pointer's extent all the time.

Why should it stay?

keveleigh: This was present in the MultiPointer Refactor branch, but that's because pointers and pointer visualization were one and the same. Here, they're separate, and tracking the extent in multiple places isn't necessary.

StephenHodgson: A bit busy right now, but I'll be sure to explain in detail later.

keveleigh:
> Cursor.cs shoudn't know about cursor distance

What does that mean? Cursor has always known about its default distance, as well as asking the FocusManager where it should place itself.

StephenHodgson:
The cursor should know know where where the pointer end is, and its transform is directly tied to it.  The cursor itself shouldn't be controlling the properties of the pointer it is attached to.

keveleigh:
>The cursor should only know where where the pointer end is, and its transform is directly tied to it.

Yes, that's what's happening on [line 330](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2201/files#diff-9427d8431b452b1871885c9abbe1af1dR330).

I'm not sure I understand. Where is the cursor controlling properties of the pointer it's attached to?

StephenHodgson:
> I'm not sure I understand. Where is the cursor controlling properties of the pointer it's attached to?

The extent is a property of the pointer, not the cursor.
I'm fine with the logical changes, but I think that we should be checking the distance in the pointer, not the cursor.

The cursor is a child of the pointer object and shouldn't be controlling/modifying the pointer's extent properties.  The code you added to 326-340 should most likely live in InputSourcePointer.

StephenHodgson:
Not to mention the fact that removing the Extent override from the pointer is a huge breaking change.

keveleigh:
>Not to mention the fact that removing the Extent override from the pointer is a huge breaking change.

That's not what's happening here. The removal is not the `extentOverride` present in [`InputSourcePointer`](https://github.com/keveleigh/HoloToolkit-Unity/blob/PointingRaysExtra/Assets/HoloToolkit/Input/Scripts/Focus/InputSourcePointer.cs#L41). The removed `extentOverride` was was added in [my PR adding pointer visualizations](https://github.com/Microsoft/MixedRealityToolkit-Unity/blame/may18_dev/Assets/HoloToolkit/Input/Scripts/Utilities/BaseControllerPointer.cs#L15), to the classes that are solely responsible for the visualizations (and not actual pointing/interaction). It was not actually doing anything, except being [immediately overwritten with the default cursor distance](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2201/files/56b9a6ecc6ce59c2ef9d8f5644af3c6f96871cf0#diff-269f5fd5cb2a1585a50c71319e1706b6L312) (again, because this class only cares about visualization, not pointing).

Now, instead of the pointer visualizer having a cached extent that needed to be updated any time the pointer's extent was updated, it [gets told about it at update-time](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2201/files/56b9a6ecc6ce59c2ef9d8f5644af3c6f96871cf0#diff-3c6ba40f6acbe9dfe24ae2c206de77deR132).

In the old refactor branch, the same class was responsible for being a pointer and for visualizing the pointer. In this implementation, to prevent breaking changes, separate classes are responsible for each.

keveleigh:
> The cursor is a child of the pointer object

Perhaps in vNext or in the old refactor branch, but that's not true in this implementation. InputSourcePointer has no idea about any cursor that's attached to it, and Cursor decides where it should place itself.

> and shouldn't be controlling/modifying the pointer's extent properties

Which line in Cursor.cs controls or modifies any of the pointer's properties? It asks for them, but I don't see where it changes anything.

Assets/HoloToolkit/Input/Scripts/Utilities/BaseControllerPointer.cs
SimonDarksideJ: I would ask if there is actually a good reason to remove the property.  It seems a useful override if you wanted to limit the pointer extent.  Unless this is being removed because it's causing a problem.

Likely this will be re-implemented differently in vNext, so I would vote to keep it in for backwards compatibility for now.

keveleigh: The removal is not the `extentOverride` present in [`InputSourcePointer`](https://github.com/keveleigh/HoloToolkit-Unity/blob/PointingRaysExtra/Assets/HoloToolkit/Input/Scripts/Focus/InputSourcePointer.cs#L41). The removed `extentOverride` was added in [my PR adding pointer visualizations](https://github.com/Microsoft/MixedRealityToolkit-Unity/blame/may18_dev/Assets/HoloToolkit/Input/Scripts/Utilities/BaseControllerPointer.cs#L15), to the classes that are solely responsible for the visualizations (and not actual pointing/interaction). It was not actually doing anything, except being [immediately overwritten with the default cursor distance](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2201/files/56b9a6ecc6ce59c2ef9d8f5644af3c6f96871cf0#diff-269f5fd5cb2a1585a50c71319e1706b6L312) (again, because this class only cares about visualization, not pointing). I shouldn't have added it in the first place.

StephenHodgson:
Ah @keveleigh I think you're right. I'm getting confused with the implementation in vNEXT and the input refactor branch.

StephenHodgson:
vNEXT has the cursor transform position being driven by the pointer directly.

keveleigh:
@StephenHodgson Yeah, much of this is a side-effect of how we started with just a gaze cursor and attempted to start from there and build up. From my understanding, it's definitely better implemented in vNext!

StephenHodgson:
@SimonDarksideJ we recently added those public properties in the patch branch, so no breaking changes for people in release, that I'm aware of.

davidkline-ms:
@StephenHodgson, correct. They were added in RC2 of this release.

PR #2222, opened 2018-05-31
Eliminated unnecessary check that resulted in bad behavior at short distances

radicalad:
Overview
---
Removed the flip check which caused a temporary gimble lock when the solver was used in small numbers < .01.

Works as expected now.

Changes
---
- Fixes: 1
SolverRadialView.cs


radicalad:
@StephenHodgson this is just a fix, when we bring Solvers to vNext we certainly can optimize them.

PR #2224, opened 2018-05-31
Fix build warnings

keveleigh:
Overview
---
Mostly, they were "[x] is never assigned to", with some "[x] is assigned to, but never used" thrown in.

From this: 
image
to this:
image

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBoxRig.cs
+++ b/Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBoxRig.cs
@@ -15,7 +15,7 @@ namespace HoloToolkit.Unity.UX
         [Header("Flattening")]
         [SerializeField]
         [Tooltip("Choose this option if Rig is to be applied to a 2D object.")]
-        private BoundingBox.FlattenModeEnum flattenedAxis;
+        private BoundingBox.FlattenModeEnum flattenedAxis = default(BoundingBox.FlattenModeEnum);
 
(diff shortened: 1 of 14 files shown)


Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBoxRig.cs
```diff
@@ -15,7 +15,7 @@ public class BoundingBoxRig : MonoBehaviour
         [Header("Flattening")]
         [SerializeField]
         [Tooltip("Choose this option if Rig is to be applied to a 2D object.")]
-        private BoundingBox.FlattenModeEnum flattenedAxis;
+        private BoundingBox.FlattenModeEnum flattenedAxis = default(BoundingBox.FlattenModeEnum);
```
StephenHodgson: None?

keveleigh: I'm not sure what this means. Are you saying to explicitly set this to a specific default value?

image

StephenHodgson: Hmm, should def have a none option on that enum.

vNEXT I suppose.

keveleigh: `DoNotFlatten` is kinda a `None`, but that name change would make it more obvious.

PR #2233, opened 2018-06-02
Fixed recursive issues with Source/Pointer creation in gaze provider

StephenHodgson:


Assets/MixedRealityToolkit/InputSystem/Gaze/GazeProvider.cs
```diff
@@ -246,14 +256,9 @@ private void OnDestroy()
...
 
-        private void InitializeInputSource()
-        {
-            gazeInputSource = new BaseGenericInputSource("Gaze");
-        }
-
         private IMixedRealityPointer InitializeGazePointer()
         {
-            return gazePointer = new InternalGazePointer(this, "Gaze Pointer", gazeInputSource, raycastLayerMasks, maxGazeCollisionDistance, gazeTransform, stabilizer);
+            return gazePointer = new InternalGazePointer(this, "Gaze Pointer", null, raycastLayerMasks, maxGazeCollisionDistance, gazeTransform, stabilizer);
```
SimonDarksideJ: If input source is not needed in creation, why is it part of the Gaze Constructor?
Does this need a little refactoring?

StephenHodgson: It's tricky, [check line 75](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2233/files#diff-42c2559cca49c73eea6265c6f99893a3R75).  We immediately set it after we create our gaze input source.

Part of the problem was that all input sources default to the gaze pointer if no pointers were passed into the constructor.  Well, if we're the gaze source, and we don't have a gaze pointer yet, we have a chick vs egg problem. So the only way I found around it was to initialize the pointer's source parent after we've created it.

Issue #1744, opened 2018-02-22
Keyboard/UICollection Broken in Unity 2017.3.1p1


markgrossnickle:
## Overview
The top left key in each UICollection is either misplaced or invisible when viewing on the device and building from Unity 2017.3.1p1
 

## Expected Behavior
All keys should be visible and placed how they are in the editor.

## Actual Behavior
When I display the keyboard.prefab the Q key is no where to be seen and the &123 is displaced but in view.   If I change to other keyboards the same thing happens in that the first element of the UICollection is missing or displaced.

## Steps to reproduce
Use Unity 2017.3.1p1 and then instantiate a keyboard.prefab in your scene.   

## Unity Editor Version
2017.3.1p1

Not reproducable in 2017.2

## Mixed Reality Toolkit Release Version
2017.2.1.1 (not certain)

Possible fix:
Add this:

            if (!Application.isEditor) { return; }

To the CollectItems and UpdateLayout methods.

jghospital:
I also suffered this bug, solved eith the proposed fix.

ivan2007:
@markgrossnickle  also for me the proposed fix works.
Thank you

keveleigh:
From some debugging today, it looks like Unity 2017.3.1p1 had an issue with setting a RectTransform's anchorPosition to (0,0), which [line 196](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit/UX/Scripts/UICollection.cs#L196) would do for the first buttons in each collection, when `Vector2 currentOffset = Vector2.zero;`. I checked by adding a small ϵ to either `x` or `y`, and the issue went away.

I can't repro this is 2017.4.3f1, so I'm going to mark this as a Unity Bug which has since been fixed. I can't find a public Unity bug though.

PR #2242, opened 2018-06-04
Keyboard Fix for Missing Keys Bug

markgrossnickle:
Overview

There was a bug with the keyboard (#1744) which causes the first button in each row to not appear or at least be positioned correctly.

Please note, the bug fixed by this Pull Request was deemed to be a Unity bug that has now been reported as fixed so perhaps this pull request can be ignored.

However, I am still leaving this code in my local respository for the slight performance improvement since the layouts never need to be updated at runtime.   Each keyboard layout is saved as a child gameObject so if you switch between layouts this code never needs to be called.  The only reason I can think of is if you try to customize the keyboard at runtime, but in that case you should just call update layout manually.  

Perhaps I am missing something? @keveleigh
 

Changes
Each of the layouts now only updates in the Editor.
 

- Fixes: # 
#1744

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/UICollection.cs
+++ b/Assets/HoloToolkit/UX/Scripts/UICollection.cs
@@ -65,6 +65,8 @@ namespace HoloToolkit.UI.Keyboard
             // Verify this is attached to a GameObject with a rect transform
             rectTransform = GetComponent<RectTransform>();
 
+            if (!Application.isEditor) { return; }
+
             // Collect children items already added (likely added in the Editor)
             CollectItems();
             UpdateLayout();
@@ -121,8 +123,7 @@ namespace HoloToolkit.UI.Keyboard
         }
 
         private void CollectItems()
-        {
-            if (!Application.isEditor) { return; }
+        { 
             Items.Clear();
 
             foreach (Transform childTransform in transform)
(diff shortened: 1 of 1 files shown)


keveleigh:
@markgrossnickle Definitely a good idea for a performance improvement! Especially since the alignment shouldn't change from what was determined in the Editor.
<p>
My only hesitation to add these checks would be that it breaks [`AddItem`](https://github.com/markgrossnickle/MixedRealityToolkit-Unity/blob/c79b63c5451e5733c698d90e57a9635e6c0aed9e/Assets/HoloToolkit/UX/Scripts/UICollection.cs#L95) and [`RemoveItem`](https://github.com/markgrossnickle/MixedRealityToolkit-Unity/blob/c79b63c5451e5733c698d90e57a9635e6c0aed9e/Assets/HoloToolkit/UX/Scripts/UICollection.cs#L108), which both update the layout after an item is added or removed at runtime. What about adding the `if (!Application.isEditor) { return; }` check in `Start()` instead, which is where the problematic calls are happening, instead of in the calls themselves?

markgrossnickle:
@keveleigh  Good catch on the add/remove item.  

I'll change it to add in the Start and the Update.

davidkline-ms:
Thanks @markgrossnickle!

PR #2243, opened 2018-06-04
cleanup .net build changes

davidkline-ms:
Changes
---
- Move IsClass, IsInterface, IsAbstract and GetInterfaces(Type) from NetNativeReflectionExtensions.cs to ReflectionExtensions.cs
- Delete NetNativeReflectionExtensions.cs as it is no longer needed and had duplicate definitions of methods found in ReflectionExtensions.cs
- Wrap contents of ReflectionExtensions.cs in #if WINDOWS_UWP &amp;&amp; !ENABLE_IL2CPP
- Add missing using statements to files using ReflectionExtensions
- Replace implementation of SystemType.GetReference() with a backend agnostic version

Tested by compiling a project in the following configurations
- .NET backend
    - .NET 4.x API compat level
    - .NET Standard 2.0 API compat level
- IL2CPP backend
    - .NET 4.x API compat level
    - .NET Standard 2.0 API compat level


Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SystemType.cs
```diff
@@ -24,12 +27,9 @@ public static string GetReference(Type type)
...
             }
 
-#if WINDOWS_UWP && !ENABLE_IL2CPP
-            return $"{type.FullName}, {type.GetAssembly().GetName().Name}";
-#else
-            return $"{type.FullName}, {type.Assembly.GetName().Name}";
-#endif
-
+            string[] qualifiedNameComponents = type.AssemblyQualifiedName.Split(new char[] { ',' });
+            Debug.Assert(qualifiedNameComponents.Length >= 2);
```
StephenHodgson: What's with this?  This change breaks my serialization for the this.

Assets/MixedRealityToolkit/_Core/Extensions/NetNativeReflectionExtensions.cs
```diff
@@ -1,211 +0,0 @@
...
-// Licensed under the MIT License. See LICENSE in the project root for license information.
-
-#if !ENABLE_IL2CPP && WINDOWS_UWP
-
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Reflection;
-
-namespace Microsoft.MixedReality.Toolkit.Internal.Extensions
```
StephenHodgson: Why remove?

davidkline-ms: Most of this file was a duplicate of ReflectionExtensions.cs.  I moved the methods needed there.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SystemType.cs
davidkline-ms: The reason was to eliminate a platform specific block.

davidkline-ms: How did the serialization break? This should be returning the same string as the other version.

Assets/MixedRealityToolkit/_Core/Extensions/NetNativeReflectionExtensions.cs
StephenHodgson: Nice catch, I didn't realize I had already ported the reflection extensions

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SystemType.cs
StephenHodgson: because each platform is referencing the assembly a bit differently is seems.

StephenHodgson: it might be because there's some hidden characters added/removed in this.

StephenHodgson: I just need to update the profiles with the updated serialization and it should be fine.

Issue #2227, opened 2018-05-31
Proposal: Fixing a bug the causes FaceOrigin collections to face the wrong direction when collections are rotated.


Ecnassianer:
## Overview

Previously collections with FaceOrigin selected as their Orientation Type which were not perfectly vertical (i.e, transform.up == Vector3.up) would align their elements in crazy ways. I fixed this for my team's project and am I proposing pulling similar changes into the MRTK June release. 

Vertical sphere:
![image](https://user-images.githubusercontent.com/231247/40812345-404767b4-64ea-11e8-99f3-66203c330783.png)

Sphere rotated 90 degrees:
![image](https://user-images.githubusercontent.com/231247/40812377-6622b6b4-64ea-11e8-9aec-319a0328f1f3.png)

ಠ_ಠ

Cylinders also have a strange layout when rotated:
![image](https://user-images.githubusercontent.com/231247/40812440-be6e9b9e-64ea-11e8-8ce3-0dc4af49446a.png)

I'm proposing consolidating all the per-shape orientation methods into a single method that behaves uniformly regardless of shape type or orientation. This means the "FaceOrigin" Orientation Type will always face collection elements toward the origin regardless of surface type. This is likely an undesirable Orientation Type for vertically aligned Cylinders which previously relied on FaceOrigin to face each element toward their central axis, rather than toward their origin (I think we can assume no one was using the odd rotated cylinder layout). IMO, this was incorrect behavior for a "FaceOrigin" type, even though a "correct" behavior for FaceOrigin on a Cylinder is probably not very desirable (it creates a weird flattened sphere where the higher or lower you are on the cylinder the more angled down or up you are, even though all elements are vertically aligned). 

To restore the previously desirable behavior of Cylinder collections, I propose adding a new Orientation Type called "FaceCenterAxis". On Cylinders this restores the old behavior. On spheres this creates a new potentially useful arrangement:
![image](https://user-images.githubusercontent.com/231247/40812565-55863b22-64eb-11e8-96b4-80aca23751a8.png)

The same is true of FaceOrigin Plane collections, however the old behavior of FaceOrigin for planes is exactly the same as the FaceForward Orientation Type, so no new Orientation Type is required.

This should be considered a breaking change, since all Cylinder and Plane collections that want the old behavior would have to be switched to use the FaceCenterAxis or FaceForward orientations. (They cannot be updated in place, since FaceOrigin needs to remain the same in the enum so spheres don't break).

[...]

PR #2246, opened 2018-06-05
Fixing a bug the causes FaceOrigin collections to face the wrong direction when collections are rotated (and some bonuses) TAKE 2

Ecnassianer:
I dug myself into a githole in that other branch. I gave up trying to revert my reverts of merges of merges and just put my changes in a new branch. Sorry for making you check it twice. It should have all the changes I made in response to previous review feedback. The only bit we should be waiting for is @paseb to weigh in on the RadialMapping() question. I'll poke him in person tomorrow.

image

---------------------------------------------
This is a pull request for the completion of my proposal:
#2227


paseb:
@StephenHodgson, I couldn't add directly to the comments but here my responses to the above comments.  Radial mapping represents a radial menu so the grid source is not what is being used though there's no really need to make a new vector so assigning it directly is fine.  Since this is a single relative axis rotation gimbal lock isn't an issue.  If we were going to do any multi-axis rotations we should be using matrix transforms.

keveleigh:
Cool, so it looks like addressing the "update source vs new vector creation" concern is all that's needed here for approval.

Assets/HoloToolkit/UX/Scripts/Collections/ObjectCollection.cs
```diff
@@ -62,23 +62,94 @@ public class ObjectCollection : MonoBehaviour
...
+
+        public float CellHeight
+        {
+            get { return cellHeight; }
+            set { cellHeight = value; }
+        }
+
+        /// <summary>
+        /// Margin between objects horizontally.
+        /// </summary>
```
StephenHodgson: Summary tags should be above the public properties not the attributes for the private fields

Assets/HoloToolkit/UX/Scripts/Collections/ObjectCollection.cs
```diff
@@ -395,6 +396,65 @@ public void UpdateCollection()
...
+
+                case OrientTypeEnum.FaceOriginReversed:
+                    node.transform.rotation = Quaternion.LookRotation(this.transform.position - node.transform.position, this.transform.up);
+                    break;
+
+                case OrientTypeEnum.FaceCenterAxis:
+                    centerAxis = Vector3.Project(node.transform.position - this.transform.position, this.transform.up);
+                    pointOnAxisNearestNode = this.transform.position + centerAxis;
+                    node.transform.rotation = Quaternion.LookRotation(node.transform.position - pointOnAxisNearestNode, this.transform.up);
+
```
StephenHodgson: a couple of extra lines before breaks.

Assets/HoloToolkit/UX/Scripts/Collections/ObjectCollection.cs
```diff
@@ -434,6 +494,27 @@ private Vector3 CylindricalMapping(Vector3 source, float radius)
...
+        /// <param name="source">The source <see cref="Vector3"/> to be mapped to cylinder</param>
+        /// <param name="radius">This is a <see cref="float"/> for the radius of the radial</param>
+        /// <param name="row">This is a <see cref="int"/> for the radius of the radial</param>
+        /// <param name="column">This is a <see cref="int"/> for the radius of the radial</param>
+        /// <returns></returns>
+        private Vector3 RadialMapping(Vector3 source, float radius, int row, int column)
+        {
+            Radius = radius >= 0 ? Radius : radius;
+       
+            Vector3 newPos = new Vector3(0f, 0f, (Radius/Rows) * row);
```
StephenHodgson: can we just assign the vector instead of creating a new one?

PR #2265, opened 2018-06-07
Fixed issue with Icon Manager

StephenHodgson:
Icon manager wasn't writing to the meta file correctly.

It now does so, and I added a check to make sure we don't overwrite the set icon, and a flag for when we do want to overwrite it.


SimonDarksideJ:
Can we add the ability to assign multiple icons for files with different names (since we standardised on names).  Which will make it much easier for doing global passes?

StephenHodgson:
Maybe later. This just fixes the fact that it didn't work.

PR #2282, opened 2018-06-11
Remove InteractionMaping Get|SetValue<T> and update relevant tests

davidkline-ms:
Overview
---
The generic Get|SetValue methods , as asserted in the test cases, were slower (and produced garbage) than their specific counterparts.

Changes
---
This change:
* removes the generic GetValue<T> and SetValue<T> methods
* renames the Get<type> methods (ex: GetRaw -> GetRawValue)
* updates calling code
* removes test cases that utilized the generic methods
* renumbered the test cases to eliminate the gaps that were introduced

### DIFF
--- a/Assets/MixedRealityToolkit-Tests/InteractionDefinitionTests.cs
+++ b/Assets/MixedRealityToolkit-Tests/InteractionDefinitionTests.cs
@@ -73,107 +73,12 @@ namespace Microsoft.MixedReality.Toolkit.Tests
             Assert.IsFalse(inputDef.Changed);
         }
 
-        // todo: remove Generic Get|Set tests
-        //[Test]
-        //public void Test03_TestObjectGenericChanged()
-        //{
-        //    var inputDef = new InteractionMapping(1, AxisType.Raw, DeviceInputType.None, new InputAction(1, "None"));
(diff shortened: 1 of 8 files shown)


Assets/MixedRealityToolkit-Tests/InteractionDefinitionTests.cs
```diff
@@ -24,7 +24,7 @@ public void Test01_TestObjectChanged()
             var testValue1 = (object)1f;
             var testValue2 = (object)false;
 
-            var initialValue = inputDef.GetRaw();
+            var initialValue = inputDef.GetRawValue();
```
StephenHodgson: These were here to show Simon the difference in performance. The tests are largely duplicated if you update these methods.

We can just remove altogether.

SimonDarksideJ:
David, by removing the Generics Get and Set from the Interaction mapping, you've managed to break the project.  There is the Dictionary Extension which uses the Generic methods which is used by devices.


Currently the ONLY overload that will now get used is the "Object" Set, which means no data will be stored or returned.  The extension is needed as dictionary's return a copy of the mapping, so this fixes that.

I'm still working on an alternative, however, the individual overloads are the backup method and the generics were the main, which is now gone.

PR #2283, opened 2018-06-11
add #if around extensions namespace as appropriate, add IsOpaque

davidkline-ms:
Overview
---
Found some references to the Microsoft.MixedReality.Toolkit.Internal.Extensions that should be in closed in #if WINDOWS_UWP && !ENABLE_IL2CPP.

Added IsOpaque to the headset definition to allow differentiation between opaque (vr) and see through (ex: HoloLens).

Also tweaked controller state comments and added OrientationOnly state.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Attributes/ExtendsAttribute.cs
+++ b/Assets/MixedRealityToolkit/_Core/Attributes/ExtendsAttribute.cs
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using System;
 using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+#if WINDOWS_UWP && !ENABLE_IL2CPP
 using Microsoft.MixedReality.Toolkit.Internal.Extensions;
+#endif // WINDOWS_UWP && !ENABLE_IL2CPP
+using System;
 
 namespace Microsoft.MixedReality.Toolkit.Internal.Attributes
 {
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerState.cs
```diff
@@ -18,10 +18,14 @@ public enum ControllerState
...
         Tracked,
         /// <summary>
-        /// The controller is currently not tracked and has relative positioning.
+        /// The controller is currently not tracked.
         /// </summary>
         NotTracked,
         /// <summary>
+        /// The controller is currently only returning orientation data.
+        /// </summary>
+        OrientationOnly,
```
StephenHodgson: What about position only?

Assets/MixedRealityToolkit/_Core/Attributes/ExtendsAttribute.cs
```diff
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using System;
 using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+#if WINDOWS_UWP && !ENABLE_IL2CPP
 using Microsoft.MixedReality.Toolkit.Internal.Extensions;
+#endif // WINDOWS_UWP && !ENABLE_IL2CPP
```
StephenHodgson: What are these for?

davidkline-ms: These were added when I noticed that VS was telling me that the namespace was not being used when building for IL2CPP and removing it blocked building with .NET backend.

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerState.cs
davidkline-ms: I can add that. I was also planning on moving Other above Tracked so that the value does not change.

PR #2287, opened 2018-06-12
fix assorted errors in the CodingGuidelines doc

davidkline-ms:
There were some errors in CodingGuidelines.md including:
* Incorrect namespace
* Shifting 0
* Spaces in enum values
* Spelling errors


CodingGuidelines.md
```diff
@@ -240,7 +240,7 @@ public class MyClass
 }
  ```
 
-## Initilise Enums.
+## Initilize Enums.
```
timGerken: This is still spelled wrong...

PR #2293, opened 2018-06-13
Added tests for InteractionMappings dictionary

SimonDarksideJ:
Overview
---

Added tests for how the InteractionMappings are used in the device space.  Highlighting the issues now created by removing the generics functions.  These tests all fail with the following error (simply because SETVALUE cannot distinguish a type like Get can)

SetValue(object) is only valid for AxisType.Raw InteractionMappings

Also added another extension for the InteractionMapping Dictionary as I missed that changed was updated between gets.

Known Issues
---

Tests now fail as Dictionary updates ONLY update the raw data type without the generic SET method.

Changes
---
- Added Dictionary Tests for InteractionMappings
- Added new InteractionMapping Dictionary extension to allow changed to be updated for a mapping
- Updated used of "Changed" in the devices to use the new GetChanged extension

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
@@ -181,7 +181,7 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Devices.WindowsGaming
                     inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, controller.Interactions[DeviceInputType.SpatialGrip].GetTransform());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions..GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
                 {
                     if (controller.Interactions[DeviceInputType.TouchpadTouch].GetBooleanValue())
                     {
--- a/Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
@@ -171,17 +171,17 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Devices.WindowsGaming
 
             if (MixedRealityManager.Instance.ActiveProfile.EnableInputSystem)
             {
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions[DeviceInputType.SpatialPointer].Changed)
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.SpatialPointer))
                 {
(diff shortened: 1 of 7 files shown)


davidkline-ms:
My apologies, Simon. I should have added named Set methods.

davidkline-ms:
We should find a generic implementation that does not box twice per update.

davidkline-ms:
Actually, I am pretty certain the tests will not fail, based on this:

Code:
    class Program
    {
        static void Main(string[] args)
        {
            CustomObject co = new CustomObject();
            CustomStruct cs = new CustomStruct(56);
            bool b = true;
            int i = 17;
            float f = 8.9f;
            double d = 43.82;
            string s = "a basic string";
            object o = new object();

            Console.WriteLine("Boolean");
            SetValue(b);

            Console.WriteLine("Int");
            SetValue(i);

            Console.WriteLine("Float");
            SetValue(f);

            Console.WriteLine("Double");
            SetValue(d);

            Console.WriteLine("String");
            SetValue(s);

            Console.WriteLine("CustomObject");
            SetValue(co);

            Console.WriteLine("CustomStruct");
            SetValue(cs);

            Console.WriteLine("Object");
            SetValue(o);

            Console.WriteLine();
        }

        static void SetValue(object val)
        {
            Console.WriteLine("SetValue(object)");
        }

        static void SetValue(bool val)
        {
            Console.WriteLine("SetValue(bool)");
        }

        static void SetValue(int val)
        {
            Console.WriteLine("SetValue(int)");
        }

        static void SetValue(float val)
        {
            Console.WriteLine("SetValue(float)");
        }

        static void SetValue(double val)
        {
            Console.WriteLine("SetValue(double)");
        }

        static void SetValue(string val)
        {
            Console.WriteLine("SetValue(string)");
        }

        static void SetValue(CustomObject val)
        {
            Console.WriteLine("SetValue(CustomObject)");
        }

        static void SetValue(CustomStruct val)
        {
            Console.WriteLine("SetValue(CustomStruct)");
        }
    }
 
Output:
Boolean
SetValue(bool)
Int
SetValue(int)
Float
SetValue(float)
Double
SetValue(double)
String
SetValue(string)
CustomObject
SetValue(CustomObject)
CustomStruct
SetValue(CustomStruct)
Object
SetValue(object)

SimonDarksideJ:
No worries, it wasn't fault for not adding the test cases to mark the dependency.
Been trying to find a way around that but not found one so far. Besides the convert action doesn't cause boxing, only the tuple did

davidkline-ms:
Converting value types didn't box?!

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
```diff
@@ -171,17 +171,17 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions[DeviceInputType.SpatialGrip].Changed)
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.SpatialGrip))
                 {
                     inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, controller.Interactions[DeviceInputType.SpatialGrip].GetTransform());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions[DeviceInputType.TouchpadTouch].Changed)
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions..GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
```
davidkline-ms: .. -> .

SimonDarksideJ: Lol, whoops, fixed.

davidkline-ms: We all make mistakes :)

PR #2299, opened 2018-06-13
Mrtk version next interaction mapping<T>

SimonDarksideJ:
Overview
---
Review implementation (NOT FOR MERGE YET) of an alternate implementation for InteractionMapping<T>

Have a look through the updated implementation and use in 
* WMRDeviceManager
* WMR Controller
* Collections Extension

In testing, this is on par with the base InteractionMapping tests, so is much faster than the old generics method.

Note
---
Ignore the changes in the other device files, as these are just scaffolds.  
Implementation is test only, once agreed I'll refactor all the stuff in to the correct places :D 

Changes
---
- Several new definitions in the InteractionMapping.cs class file.  New InteractionMapping<T> and interface added.
 - Updated implementation in the WMR Device manager and WMR Controller defs


Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
```diff
@@ -244,4 +243,129 @@ public void SetValue(Tuple<Vector3, Quaternion> newValue)
 
         #endregion Set Operators
     }
+
+    public class Test
+    {
+        System.Collections.Generic.List<IInteractionMapping> list = new System.Collections.Generic.List<IInteractionMapping>();
+        void helloWorld()
+        {
+            var interaction = new InteractionMapping<float>(0, AxisType.SingleAxis, DeviceInputType.None, new InputAction(1, "None"));
```
davidkline-ms: I REALLY like the idea of typing the interaction mapping. much easier to read and consume

Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
```diff
@@ -244,4 +243,129 @@ public void SetValue(Tuple<Vector3, Quaternion> newValue)
...
+        /// Has the value changed since the last reading.
+        /// </summary>
+        bool Changed { get; }
+    }
+
+    public class InteractionMapping<TReadingType> : IInteractionMapping
+    {
+        public InteractionMapping(uint id, AxisType axisType, DeviceInputType inputType, InputAction inputAction)
+        {
+            this.id = id;
```
davidkline-ms: would be great to add a check that TReadingType is compatible with the specified AxisType. if we are certain that this is internal MRTK _Core code only, then an Assert will do.

Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
```diff
@@ -244,4 +243,129 @@ public void SetValue(Tuple<Vector3, Quaternion> newValue)
...
+            private set
+            {
+                changed = value;
+            }
+        }
+        #endregion Interaction Properties
+
+        #region Operators
+
+        public TReadingType GetData() => currentReading;
```
davidkline-ms: Love how much easier to read this is

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -53,7 +53,7 @@ public WindowsMixedRealityController(ControllerState controllerState, Handedness
         public IMixedRealityInputSource InputSource { get; private set; }
 
         /// <inheritdoc/>
-        public Dictionary<DeviceInputType, InteractionMapping> Interactions { get; private set; }
+        public Dictionary<DeviceInputType, IInteractionMapping> Interactions { get; private set; }
```
davidkline-ms: This just struck me... it is very possible that a given controller may have multiple controls (ex: Game Controllers generally have two thumbsticks) for a given DeviceInputType… is this going to bite us?

Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
SimonDarksideJ: I'll have a look, however Checking the type at runtime will incur a boxing cost most likely.
Main point is it shouldn't matter using this implementation. Main reason for the check in the old version was that we had multiple datum storage, now we only have one. 

The type set when the mapping was created enforces the data type stored.  Although I suppose we could have a "Validator" instead of building it in to the mapping.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
SimonDarksideJ: DeviceInputType will GROW,  So I fully expect to add Thumbstick 1 / Thumbstick 2 when we do Xbox controllers.
@StephenHodgson does disagree with me on this, but I cannot think of another way to uniquely identify differert hardware types in a fasterway than this lookup.  Most SDK's (Xbox Gamepad is a good example) use an enum, not an ID, then apply different "GetValue" methods for each control.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
```diff
@@ -31,7 +31,7 @@ public interface IMixedRealityController
         /// <summary>
         /// Mapping definition for this controller, linking the Physical inputs to logical Input System Actions
         /// </summary>
-        Dictionary<DeviceInputType, InteractionMapping> Interactions { get; }
+        Dictionary<DeviceInputType, IInteractionMapping> Interactions { get; }
```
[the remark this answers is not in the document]
SimonDarksideJ: Not really, as it would further complicate the lookup.  The reason the DeviceInputType is the key, is so that when a button is pressed, the code takes the input and looks for the mapping for that input.
With it being an enum, it's an ID lookup in any case and as fast as we can make it.

Physical -> SDK -> InputType -> InteractionMapping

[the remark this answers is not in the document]
StephenHodgson: I still think having the device input type as the key limits us with assigning multiple types here.

SimonDarksideJ:
So, comments please @StephenHodgson @keveleigh @davidkline-ms on whether we agree with this approach.

If so, I'll tidy it up and make the PR ready, then we don't need the "generics" fix putting back in as this replaces it.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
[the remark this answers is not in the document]
SimonDarksideJ: How does it limit us?  A key is unique, so we can't assign multiples.  Plus the interactions are what are configured for that physical input.  Theres not that many to create an issue.  
Although, yes, we do need more input types as we add more physical controllers with more buttons.

It was suggested we also have a filter on the controller for what physical inputs a controller has for easier config

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
StephenHodgson: I agree, I think we should be using something different as the key

Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
StephenHodgson: Ideally we shouldn't be making new ones anyway. They should be done in the controller's config profile.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
SimonDarksideJ: OK, then let's finish up this PR which was to address a different issue. Then create a task to revisit this later, as it is currently working.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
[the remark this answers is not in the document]
StephenHodgson: >A key is unique, so we can't assign multiples. 

That's exactly the problem.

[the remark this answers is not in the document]
StephenHodgson: >we do need more input types as we add more physical controllers with more buttons.

No we don't. A button is a button

SimonDarksideJ:
PR ready

Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
```diff
@@ -84,164 +84,30 @@ private set
...
+        /// Get the value of the Interaction Mapping data
+        /// </summary>
+        /// <param name="newValue"></param>
+        public void GetValue(TReadingType newValue)
         {
-            if (AxisType != AxisType.SixDoF)
-            {
-                Debug.LogError("SetValue(Tuple<Vector3, Quaternion>) is only valid for AxisType.SixDoF InteractionMappings");
-            }
+            Changed = currentReading == null && newValue != null ||
```
davidkline-ms: We could probably make it just currentReading != newValue.
Especially since value types can not be null.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityGestureController.cs
```diff
@@ -84,15 +85,41 @@ public void SetupFromInteractionSource(InteractionSourceState interactionSourceS
...
         /// </summary>
         /// <param name="mappings">Configured mappings from a controller mapping profile</param>
-        private void SetupFromMapping(InteractionMapping[] mappings)
+        private void SetupFromMapping(IInteractionMapping[] mappings)
         {
             for (uint i = 0; i < mappings.Length; i++)
             {
                 // Add interaction for Mapping
-                Interactions.Add(mappings[i].InputType, new InteractionMapping(i, mappings[i].AxisType, mappings[i].InputType, mappings[i].InputAction));
+                switch (mappings[i].AxisType)
```
davidkline-ms: not sure this should be populated yet. Gesture controllers need some more thinking.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDevice.cs
```diff
@@ -170,17 +170,20 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialGrip) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.SpatialGrip))
                 {
-                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, controller.Interactions[DeviceInputType.SpatialGrip].GetTransform());
+                    var data = controller.Interactions[DeviceInputType.SpatialGrip] as InteractionMapping<Tuple<Vector3, Quaternion>>;
+                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, data.GetValue());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
+                if (controller.Interactions.ContainsKey(DeviceInputType.TouchpadTouch) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
```
davidkline-ms: Does OpenVR support touchpad touch?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -220,17 +220,20 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
                 {
-                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, controller.Interactions[DeviceInputType.SpatialGrip].GetTransform());
+                    var data = controller.Interactions[DeviceInputType.SpatialGrip] as InteractionMapping<Tuple<Vector3, Quaternion>>;
+                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, data.GetValue());
                 }
 
                 if (controller.Interactions.ContainsKey(DeviceInputType.TouchpadTouch) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
                 {
-                    if (controller.Interactions[DeviceInputType.TouchpadTouch].GetBooleanValue())
+                    var data = controller.Interactions[DeviceInputType.TouchpadTouch] as InteractionMapping<bool>;
```
davidkline-ms: Need TouchpadPress also

Assets/MixedRealityToolkit/_Core/Devices/Simulator/SimulatedDevice.cs
```diff
@@ -170,17 +170,20 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialGrip) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.SpatialGrip))
                 {
-                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, controller.Interactions[DeviceInputType.SpatialGrip].GetTransform());
+                    var data = controller.Interactions[DeviceInputType.SpatialGrip] as InteractionMapping<Tuple<Vector3, Quaternion>>;
+                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, data.GetValue());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
+                if (controller.Interactions.ContainsKey(DeviceInputType.TouchpadTouch) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
```
davidkline-ms: Also need TouchpadPress

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
```diff
@@ -173,17 +173,20 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialGrip) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.SpatialGrip))
                 {
-                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, controller.Interactions[DeviceInputType.SpatialGrip].GetTransform());
+                    var data = controller.Interactions[DeviceInputType.SpatialGrip] as InteractionMapping<Tuple<Vector3, Quaternion>>;
+                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialGrip].InputAction, data.GetValue());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
+                if (controller.Interactions.ContainsKey(DeviceInputType.TouchpadTouch) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.TouchpadTouch))
```
davidkline-ms: Do we need this? Xbox controller does not have a touchpad.

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
```diff
@@ -193,19 +196,22 @@ private void InteractionSourceUpdated(InteractionSourceState state)
                     }
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.Touchpad))
+                if (controller.Interactions.ContainsKey(DeviceInputType.Touchpad) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.Touchpad))
```
davidkline-ms: Do we need this? Xbox controller does not have a touchpad.

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
```diff
@@ -193,19 +196,22 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
+                    var data = controller.Interactions[DeviceInputType.Touchpad] as InteractionMapping<Vector2>;
+                    inputSystem?.Raise2DoFInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.Touchpad].InputAction, data.GetValue());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.ThumbStick))
+                if (controller.Interactions.ContainsKey(DeviceInputType.ThumbStick) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.ThumbStick))
                 {
-                    inputSystem?.Raise2DoFInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.ThumbStick].InputAction, controller.Interactions[DeviceInputType.ThumbStick].GetVector2Value());
+                    var data = controller.Interactions[DeviceInputType.ThumbStick] as InteractionMapping<Vector2>;
```
davidkline-ms: here is where we need two thumbsticks.

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
```diff
@@ -193,19 +196,22 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
+                    inputSystem?.Raise2DoFInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.ThumbStick].InputAction, data.GetValue());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.Trigger))
+                if (controller.Interactions.ContainsKey(DeviceInputType.Trigger) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.Trigger))
```
davidkline-ms: we will also need two triggers

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
```diff
@@ -193,19 +196,22 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
                 {
-                    inputSystem?.RaiseOnInputPressed(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.Select].InputAction, controller.Interactions[DeviceInputType.Trigger].GetFloatValue());
+                    var data = controller.Interactions[DeviceInputType.Trigger] as InteractionMapping<float>;
+                    inputSystem?.RaiseOnInputPressed(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.Select].InputAction, data.GetValue());
```
davidkline-ms: Xbox controller has 8 buttons that can be used. it also has a d-pad that we should be sure exists in our DeviceInputType.

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
```diff
@@ -193,19 +196,22 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
+                if (controller.Interactions.ContainsKey(DeviceInputType.Touchpad) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.Touchpad))
                 {
-                    inputSystem?.Raise2DoFInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.Touchpad].InputAction, controller.Interactions[DeviceInputType.Touchpad].GetVector2Value());
+                    var data = controller.Interactions[DeviceInputType.Touchpad] as InteractionMapping<Vector2>;
+                    inputSystem?.Raise2DoFInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.Touchpad].InputAction, data.GetValue());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.ThumbStick))
+                if (controller.Interactions.ContainsKey(DeviceInputType.ThumbStick) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.ThumbStick))
```
davidkline-ms: ThumbstickPress needed (x2)

Assets/MixedRealityToolkit/_Core/Devices/OpenXR/OpenXRDevice.cs
```diff
@@ -170,17 +170,20 @@ private void InteractionSourceUpdated(InteractionSourceState state)
             {
                 if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.SpatialPointer))
                 {
-                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialPointer].InputAction, controller.Interactions[DeviceInputType.SpatialPointer].GetTransform());
+                    var data = controller.Interactions[DeviceInputType.SpatialPointer] as InteractionMapping<Tuple<Vector3, Quaternion>>;
+                    inputSystem?.Raise6DofInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.SpatialPointer].InputAction, data.GetValue());
                 }
 
-                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialPointer) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.SpatialGrip))
+                if (controller.Interactions.ContainsKey(DeviceInputType.SpatialGrip) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.SpatialGrip))
```
davidkline-ms: let's not implement the OpenXR controller until the spec has been completed / published

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -242,17 +245,20 @@ private void InteractionSourceUpdated(InteractionSourceState state)
...
                 {
-                    inputSystem?.Raise2DoFInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.Touchpad].InputAction, controller.Interactions[DeviceInputType.Touchpad].GetVector2Value());
+                    var data = controller.Interactions[DeviceInputType.Touchpad] as InteractionMapping<Vector2>;
+                    inputSystem?.Raise2DoFInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.Touchpad].InputAction, data.GetValue());
                 }
 
                 if (controller.Interactions.ContainsKey(DeviceInputType.ThumbStick) && controller.Interactions.GetDictionaryValueChanged(DeviceInputType.ThumbStick))
                 {
-                    inputSystem?.Raise2DoFInputChanged(controller.InputSource, controller.ControllerHandedness, controller.Interactions[DeviceInputType.ThumbStick].InputAction, controller.Interactions[DeviceInputType.ThumbStick].GetVector2Value());
+                    var data = controller.Interactions[DeviceInputType.ThumbStick] as InteractionMapping<Vector2>;
```
davidkline-ms: Need ThumbstickPress

Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
SimonDarksideJ: Tuple<T,T> is NOT a value type!, so we have to have this for this use case

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityGestureController.cs
SimonDarksideJ: Scaffold, but I shall remove the copied code

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDevice.cs
SimonDarksideJ: Scaffold, but I shall remove the copied code

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
SimonDarksideJ: Will check as I thought I had added that.

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
SimonDarksideJ: Scaffold, but I shall remove the copied code

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
SimonDarksideJ: Scaffold, but I shall remove the copied code

Assets/MixedRealityToolkit/_Core/Devices/WindowsGaming/WindowsGamingDevice.cs
SimonDarksideJ: Scaffold, but I shall remove the copied code

Assets/MixedRealityToolkit/_Core/Devices/OpenXR/OpenXRDevice.cs
SimonDarksideJ: Scaffold, but I shall remove the copied code

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
SimonDarksideJ: Duplicate comment.  However, I will double check

Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
davidkline-ms: Changed = currentReading == null && newValue != null || 
           currentReading != null && newValue == null || 
           currentReading != null && newValue != null && 
           !currentReading.Equals(newValue);

It just feels like there is some redundancy here that isn't needed... I'll try a few things out in a simple app and update my comments.

davidkline-ms: so, in the case of using just ==, Tuples with the same contents are considered different (ex: it is reference equals only).

in the case of value types, I need to confirm whether or not using a T to declare the type will report the same warnings as checking an explicitly defined one (ex: bool) for null.

davidkline-ms: I see where using .Equals is required now (instances of T cannot use ==). That said, I still think we can make this check easier to read.

davidkline-ms: I did some testing (please forgive the comment spam and subsequent deletions)… The following feels easier to read and should accomplish the same thing:

Changed = (currentReading != null) ? !current.Equals(newValue) : (newValue != null);

My results showed the following:
b1 (false) != b2 (true) True
b1 (false) != b3 (false) False

t1 (43, 7.15) != t2 (17, 4.3) True
t1 (43, 7.15) != t3 (43, 7.15) False
tn1 (null) != tn2 (null) False
tn1 (null) != t2 (17, 4.3) True
t1 (43, 7.15) != tn2 (null) True

Which matched the output from the version in the PR

StephenHodgson: This is a ton of boxing in this op:
image

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
StephenHodgson: > DeviceInputType will GROW, So I fully expect to add Thumbstick 1 / Thumbstick 2 when we do Xbox controllers.

We should most def not do this.

Assets/MixedRealityToolkit/_Core/Extensions/CollectionsExtensions.cs
```diff
@@ -135,11 +136,11 @@ public static T[] ExportDictionaryValuesAsArray<T>(this Dictionary<uint,T> input
...
         /// <param name="value">The data value to be updated</param>
         public static void SetDictionaryValue<T>(
-            this Dictionary<Definitions.Devices.DeviceInputType, Definitions.Devices.InteractionMapping> input, 
+            this Dictionary<Definitions.Devices.DeviceInputType, Interfaces.IInteractionMapping> input, 
             Definitions.Devices.DeviceInputType key, T value)
         {
-            var entry = input[key];
-            entry.SetValue(value);
+            var entry = input[key] as InteractionMapping<T>;
+            entry.GetValue(value);
```
StephenHodgson: Should we check if the entry is not null?

SimonDarksideJ:
Took a PR from @StephenHodgson with some great enhancements but need to re-work stuff to make it work

Assets/MixedRealityToolkit/_Core/Extensions/CollectionsExtensions.cs
SimonDarksideJ: Quite possibly, worth adding

Assets/MixedRealityToolkit/_Core/Definitions/Devices/InteractionMapping.cs
SimonDarksideJ: So, do you have a working alternative?
Noting we have to handle BOTH value and reference types

StephenHodgson: Using direct methods 😝

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
keveleigh: > We should most def not do this.

How do you see it being implemented?

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityController.cs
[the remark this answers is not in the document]
keveleigh: > No we don't. A button is a button

We currently have explicit `Menu` and `GripPress` (etc) buttons defined. Why are those not just "button"? How do you differentiate between different buttons in the controller mapping profile if they're all just `Button`?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SixDof.cs
```diff
@@ -0,0 +1,102 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities
+{
+    [Serializable]
+    public struct SixDof : IEqualityComparer
```
StephenHodgson: I need to pull a few of these changes out and make them their item PR.

Sorry about this.

SimonDarksideJ:
Looks like we need to pull and rebase this PR again. Downside of moving so fast but yet more lessons learned.

SimonDarksideJ:
Restarting this PR due too many underlying conflicts with other PR's

Issue #2300, opened 2018-06-13
MoveToPosition.TargetObject continues to target its own game object instead of the target


ForrestTrepte:
## Overview

MoveToPosition.TargetObject specifies "the GameObject with the position the requires animation". However, this variable is only used in GetPosition. The code that sets position sets this.transform.position/localPosition.

## Expected Behavior

TargetObject is moved.

## Actual Behavior

This GameObject is moved.

## Steps to reproduce

1. Open the CycleArray scene.
1. Create a new sphere gameobject of scale .25, .25, .25.
1. Modify PositionCube > MoveToPosition > TargetObject to reference the sphere instead of None.
1. Run the scene and click on CyclePosition.

expected: The sphere is moved.
actual: The CyclePosition cube is moved.

## Unity Editor Version
2017.4.2f2
## Mixed Reality Toolkit Release Version
june18_dev 0f9cff76

PR #2301, opened 2018-06-13
fix MoveToPosition.TargetObject

ForrestTrepte:
Overview
---
MoveToPosition was setting its own transform instead of the TargetObject's transform.

Changes
---
- Fixes: #2300.


keveleigh:
Thanks for this fix!

PR #2306, opened 2018-06-15
Properly expose MonoBehaviour methods in Cursor.cs

### LINKED ISSUE #2294: Interactive mesh cursor default distance set to 0
## Overview
As you may know, the HoloLens cursor is following the surface of a box collider by default when it can or it will remain at a default distance otherwise: with the interactive mesh cursor behavior, it's not the case.

When we go anywhere other than a surface of a collider, the interactive mesh cursor is renderer at a distance of 0 with the camera. In other words, we can't see it. And that have a huge bad impact with the UX when using an application using this cursor.

## Expected Behavior
The cursor's distance with the camera is set to the default distance when it's not on a collider surface.

## Actual Behavior
The cursor is rendered on the camera position.

## Steps to reproduce
1. Sets up a Unity project for an HoloLens application
2. Place a new Cube in the scene
3. Use the InteractiveMeshCursor as your cursor

This [project](https://github.com/rigma/InteractiveMeshCursor-Bug) is a reproduction of this issue.

## Unity Editor Version
2018.1.0f2

2017.4.0.0

keveleigh:
InteractiveMeshCursor was unknowingly hiding Cursor's Awake. This wasn't an issue previously, since the only thing in Awake actually didn't really do anything due to a different bug (also fixed here). 

Update Cursor to make its Awake (and others) overridable.

This commit also exposes `isVisible` in the Editor, as the comment in
Awake() indicates that it should be settable before startup.

Also, removes an extra SetVisibility call that already happens in the
public setter.

Fixes #2294.

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
@@ -118,6 +118,9 @@ namespace HoloToolkit.Unity.InputModule
         protected ICursorModifier TargetedCursorModifier;
 
         private uint visibleHandsCount = 0;
+
+        [SerializeField]
+        [Tooltip("Set this to specify if the Cursor should start visible or invisible in the scene.")]
         private bool isVisible = true;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


keveleigh:
@rigma, please take a look and make sure this fixes your issue! Thanks!

Issue #2310, opened 2018-06-16
Change AppBar Icon


brean:
## Overview
I don't know if it is even possible with the current implementation, but I like to change the icon for the Hide, Adjust, ...-Buttons in the AppBar to show Sprites instead of the UIButtonSquareIcon-Mesh. 
You can set the "Overwrite Icon" Flag after the buttons are created (in play mode) of the CompontButtonIcon of the Button instance, but they get destroyed after exiting play-mode.

## Expected Behavior
You should be able to overwrite the Button icons in the Inspector somehow, e.g. by configuring the profile for the AppBar-Prefab

## Actual Behavior
There is no way (at least I found none) to change the Mesh to Sprite

## Steps to reproduce
Try to change the Icon of an AppBar-Button

## Unity Editor Version
2018.1f1

## Mixed Reality Toolkit Release Version
june18_dev

StephenHodgson:
Why not just make your own prefab?

brean:
Oh, I have not thought of that. So I just created my own ButtonHolographicSprite prefab based on ButtonHolographic prefab, I can overwrite the icon in the icon-settings and set a sprite there but then every Button has the same icon. I can create my own ComponentIconButton-Script that uses Sprites but I think there should be a checkbox in the ComponentButtonIcon so it looks in the profile if there is a sprite set for this button.

brean:
I tried implementing this but I don't know how to get the icon name (e.g. "Adjust" to do the icon lookup in the profile) in the SetIconName-function of CompoundButtonIcon.

brean:
OK, I have a working version now, as described with a simple checkbox "Use Prefab from Profile" as well as ~10 lines of code. It is an extension to the CompoundButtonIcon as well as the AppBar. 
I can provide a PR, I am not sure on what branch? (I am currently working on june18_dev)

StephenHodgson:
Thanks @brean, I think it would be a wonderful addition. Looking forward to your PR.

ThomasAstrakhan:
> I tried implementing this but I don't know how to get the icon name (e.g. "Adjust" to do the icon lookup in the profile) in the SetIconName-function of CompoundButtonIcon. (The name is always "ObjectCollectionScatter", with the comment Replace with your custom icon texture name in HolographicButton prefab so this should be replaced in code? - But then I have to provide my own AppBar.cs

I had to modify my Appbar.cs because of this. How did you manage to solve this issue ?

PR #2314, opened 2018-06-17
Better example for custom Icons for AppBar

brean:
Overview
---
Some icons for the AppBar instead of the generic ObjectCollectionScatter, to look a bit more like the version from the Holograms app

Changes
---
- Fixes: #2310 .


Assets/HoloToolkit-Examples/UX/Icons/AdjustIcon.svg
```diff
@@ -0,0 +1,236 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
```
StephenHodgson: why svg instead of png?

StephenHodgson: Oh I see you included both.

brean: Yeah, I added the original files, makes it easier for others to either modify existing icons or add new icons based on these.

StephenHodgson: I think we'll def use TextMeshPro to make the default icon pack for vNEXT.

davidkline-ms:
@cre8ivepark should review the look of these new icons

cre8ivepark:
Thanks! @brean We were not able to include icon glyph font files and images of the glyphs since we cannot publish them in the open-source projects. (font license/design patent issue). 
We provide HoloLens MDL font from Dev Center with proper EULA file. Developers can manually import into Unity and generate the texture from this font file. 
https://docs.microsoft.com/en-us/windows/uwp/design/downloads/

I agree that we need better documentation for this & tools. (something that can automatically generate icon texture from the font file - we had this feature in MixedRealityDesignLabs. Needs porting work)

I'll clone your repo and check out your icon design today. Thank you!
Yoon

StephenHodgson:
@cre8ivepark if we were to generate the icons with a tool like TextMeshPro without including the source font itself, would that satisfy the licensing constraints?

The end result would just be a texture.

cre8ivepark:
@brean I checked out the design. Since your design is very close to the original design, I think it would be better to just use the original design from HoloLens MDL font. I'll apply the updated texture to your repository this weekend. Thanks!

cre8ivepark:
@brean , I applied changes to your repository:
- Updated textures to the original design.
- Changed the location of the textures to existing HoloToolkit/UX/Textures/Icons. Textures under HoloToolkit-Examples folder will break the app bar when the user doesn't have HoloToolkit-Examples folder.
- Removed SVG files. People can use existing icon texture (512x512) png file and [HoloLens MDL font](https://docs.microsoft.com/en-us/windows/uwp/design/downloads/) as a template.
- Updated icon font material to use emission color. (icon was shaded depending on the direction, by the lighting condition)

2018-06-25 06_54_07-unity 2017 2 1p2 64bit - boundingboxgizmoexample unity - mrtk-breen - universa
2018-06-25 06_54_24-unity 2017 2 1p2 64bit - boundingboxgizmoexample unity - mrtk-breen - universa
2018-06-25 06_54_34-unity 2017 2 1p2 64bit - boundingboxgizmoexample unity - mrtk-breen - universa

PR #2315, opened 2018-06-17
vNEXT Input System Refactoring and Cleanup

StephenHodgson:
- Refactored InputSystem to accept custom InputAction types.
    - Now all raised events require an InputAction as a parameter.
- Cleaned up IMixedRealityInputSystem.
- Removed the default pointer from the input action configuration profile.

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
+++ b/Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
@@ -21,7 +21,7 @@ MonoBehaviour:
   - id: 3
     description: Grip
     axisConstraint: 7
-  pointerAction:
-    id: 0
+  defaultPointerAction:
+    id: 1
     description: 
     axisConstraint: 0
--- a/Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
+++ b/Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
@@ -21,7 +21,3 @@ MonoBehaviour:
   - id: 3
     description: Grip
     axisConstraint: 7
-  defaultPointerAction:
-    id: 1
-    description: 
-    axisConstraint: 0
--- a/Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
+++ b/Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
@@ -134,6 +134,7 @@ namespace Microsoft.MixedReality.Toolkit.InputSystem.Cursors
         public virtual void OnSourceDetected(SourceStateEventData eventData)
         {
 #if UNITY_WSA
+            // TODO Figure out a better way to know what type of source is found / lost.
             InteractionSourceKind sourceKind;
             if (InteractionInputSources.TryGetSourceKind(eventData.SourceId, out sourceKind) &&
                 sourceKind == InteractionSourceKind.Hand)
@@ -155,6 +156,7 @@ namespace Microsoft.MixedReality.Toolkit.InputSystem.Cursors
         public virtual void OnSourceLost(SourceStateEventData eventData)
         {
 #if UNITY_WSA
+            // TODO Figure out a better way to know what type of source is found / lost.
             InteractionSourceKind sourceKind;
             if (InteractionInputSources.TryGetSourceKind(eventData.SourceId, out sourceKind) &&
                 sourceKind == InteractionSourceKind.Hand)
--- a/Assets/MixedRealityToolkit/_Core/Definitions/Utilities/AxisType.cs
+++ b/Assets/MixedRealityToolkit/_Core/Definitions/Utilities/AxisType.cs
@@ -32,14 +32,14 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities
         /// <summary>
         /// Position only Axis analogue input.
         /// </summary>
-        ThreeDoFPosition,
+        ThreeDofPosition,
         /// <summary>
         /// Rotation only Axis analogue input.
         /// </summary>
-        ThreeDoFRotation,
+        ThreeDofRotation,
         /// <summary>
         /// Position AND Rotation analogue input.
         /// </summary>
-        SixDoF
+        SixDof
     }
 }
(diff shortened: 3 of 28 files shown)


Jarodshow:
Could there be an InputAction.None? Or would null represent no action?

StephenHodgson:
Null represents no action.

StephenHodgson:
But for the most part, you'll want to use an Input Action if you're using the Input System.

It helps abstract the input you'll be getting from other input sources, controllers, etc.  That way you don't have to wire up unique events for every controller input.

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
```diff
@@ -21,7 +21,7 @@ MonoBehaviour:
   - id: 3
     description: Grip
     axisConstraint: 7
-  pointerAction:
-    id: 0
+  defaultPointerAction:
```
SimonDarksideJ: Still not sure we need this separate action.

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
```diff
@@ -134,6 +134,7 @@ public bool IsVisible
         public virtual void OnSourceDetected(SourceStateEventData eventData)
         {
 #if UNITY_WSA
+            // TODO Figure out a better way to know what type of source is found / lost.
             InteractionSourceKind sourceKind;
```
SimonDarksideJ: Note, Multi-VR breaking issue

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
```diff
@@ -155,6 +156,7 @@ public virtual void OnSourceDetected(SourceStateEventData eventData)
         public virtual void OnSourceLost(SourceStateEventData eventData)
         {
 #if UNITY_WSA
```
SimonDarksideJ: Multi-VR breaking issue

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
StephenHodgson: Not exactly a breaking issue, just doesn't currently conform to our standards. Let's add an issue and get back to this later. It will compile.

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
StephenHodgson: I def think we should. I was actually contemplating adding a default up/down/and click action for all pointers.

StephenHodgson:
> Need to abstract out / remove the #IF UNITY_WSA code to avoid future multi-vr breaking

That doesn't exactly pertain to this PR. Let's open an issue and address that specifically.

This PR only addresses the updates to Input Actions and their usage in the Input System

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
StephenHodgson: The reason _why_ we specifically need them is because of the way the Unity Canvas system works and how our need to raycast against the canvases and physical objects.

Take a look at the `MixedRealityInputManager`'s Raise Pointer methods.  In each we have to make sure we've got a valid `GraphicInputEventData` before we can properly execute the pointer up/down/click.

```
public void RaisePointerUp(IMixedRealityPointer pointer, Handedness handedness, IMixedRealityInputAction inputAction)
{
    // Create input event
    pointerEventData.Initialize(pointer.InputSourceParent, handedness, inputAction);

    if (MixedRealityManager.Instance.ActiveProfile.InputActionsProfile.DefaultPointerAction.Id == inputAction.Id)
    {
        ExecutePointerUp(HandlePointerUp(pointer));
    }
}
```

SimonDarksideJ: Then this should be a static in the InputSystem that derives from the Currently configured pointer action.  Unless you feel this is a different action to the one I would configure on the spatialPointer

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
SimonDarksideJ: Ok, agreed. Create a task for an #IF UNITY pass to remove these.

but if I switched to standalone, this wouldn't work.

StephenHodgson: Well you wouldn't be using hand input on standalone without a Magic Leap right? 😉

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
StephenHodgson: That's a possibility.  Let me chew on this a bit more and see if I can come up with a better solution.

StephenHodgson: Ended up removing for now until we need it again. After testing this out, I concur, we don't really need it

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/AxisType.cs
```diff
@@ -32,14 +32,14 @@ public enum AxisType
         /// <summary>
         /// Position only Axis analogue input.
         /// </summary>
-        ThreeDoFPosition,
+        ThreeDofPosition,
```
davidkline-ms: why the casing changes?

StephenHodgson: Thought I put this in the description. Must have missed that.

So that it renders properly in the inspector.

old: `Three Do F Position`
new: `Three Dof Position`

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
```diff
@@ -134,6 +134,7 @@ public bool IsVisible
         public virtual void OnSourceDetected(SourceStateEventData eventData)
         {
 #if UNITY_WSA
+            // TODO Figure out a better way to know what type of source is found / lost.
```
StephenHodgson: @SimonDarksideJ My next PR will address this.

PR #2318, opened 2018-06-19
update class visibility and add runtime assert/logging

davidkline-ms:
The controller classes (deriving from IMixedRealityController) were marked as public. Since apps should not have access to them and should be getting their data via the interface, therefore I have marked them as internal.

The device classes (deriving from IMixedRealityDevice) had inconsistent visibility. Some were private and others public. Similar to the controllers, there is no need for these classes to be available to applications. Since WindowsMixedRealityDevice had the most restrictive visibility and is what we have successfully used, I have made them all private.

While not strictly related, the intent of adjusting the visibility of classes is to enable use of asserts during development and free the release/master builds from needing to check for situations that should not be possible. Since Unity's Debug.Assert does not provide runtime assertions (when an app has been built in VS and not running in the editor) I added an editor check and switched which assert (and logging method) is called.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Devices/Simulator/SimulatedDevice.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/Simulator/SimulatedDevice.cs
@@ -16,7 +16,7 @@ using UnityEngine.XR.WSA.Input;
 namespace Microsoft.MixedReality.Toolkit.Internal.Devices.Simulator
 {
     // TODO - Implement
-    public class SimulatedDevice : IMixedRealityDevice
+    class SimulatedDevice : IMixedRealityDevice
     {
         /// <summary>
         /// Dictionary to capture all active controllers detected
--- a/Assets/MixedRealityToolkit/_Core/Devices/Simulator/SimulatedDevice.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/Simulator/SimulatedDevice.cs
@@ -16,7 +16,7 @@ using UnityEngine.XR.WSA.Input;
 namespace Microsoft.MixedReality.Toolkit.Internal.Devices.Simulator
 {
     // TODO - Implement
-    class SimulatedDevice : IMixedRealityDevice
+    internal class SimulatedDevice : IMixedRealityDevice
     {
         /// <summary>
         /// Dictionary to capture all active controllers detected
--- a/Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
@@ -15,7 +15,7 @@ using UnityEngine.XR.WSA.Input;
 
 namespace Microsoft.MixedReality.Toolkit.Internal.Devices.WindowsMixedReality
 {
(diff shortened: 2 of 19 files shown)


StephenHodgson:
I'm also not keen on wrapping default unity api's

Assets/MixedRealityToolkit/_Core/Devices/Simulator/SimulatedDevice.cs
```diff
@@ -16,7 +16,7 @@
 namespace Microsoft.MixedReality.Toolkit.Internal.Devices.Simulator
 {
     // TODO - Implement
-    public class SimulatedDevice : IMixedRealityDevice
+    class SimulatedDevice : IMixedRealityDevice
```
keveleigh: `internal`? (There are some other cases of this too, but I'll leave one comment for discussion.)

davidkline-ms: this seems reasonable, mostly to avoid too much churn if we determine that the input system assembly needs to access these classes.

davidkline-ms:
@StephenHodgson, if we do not provide a runtime assert, we will miss easy to catch bugs when running platform specific code. I have hit a few of these when debugging the master branch releases in the past few months. They wont make production code for apps, but is handy for us to use when developing / testing the core.

davidkline-ms:
@StephenHodgson, why not wrap the Unity APIs in this case? Makes it super easy for folks to get the correct assert behavior regardless of environment.

I **will** remove the calls to System.Diagnostics.Debug.WriteLine() as they are not necessary (we get the messages even when not in the editor)

StephenHodgson:
> why not wrap the Unity APIs in this case? Makes it super easy for folks to get the correct assert behavior regardless of environment.

Because it messes with how those are raised internal in the editor.  People who have other solutions that rely on the editor logging functionality will not properly get these raised.

davidkline-ms:
> Because it messes with how those are raised internal in the engine. People who have other solutions that rely on the engine logging functionality will not properly get these raised.

Fair enough. We still need to have a solution for getting runtime asserts so we can catch platform specific bugs.

StephenHodgson:
Be sure that you have development build enabled in the Generic Unity Build window and see if they're thrown then.

davidkline-ms:
> Be sure that you have development build enabled in the Generic Unity Build window and see if they're thrown then.

They are shown in the app's window. Unfortunately, unlike .net's asserts, the app does not halt. I have removed the Application.inEditor check to ensure that we get them at runtime.

StephenHodgson:
Ugh I didn't want this

davidkline-ms:
what parts? the debug wrapper was reverted.

davidkline-ms:
I can undo the visibility changes... The DebugUtils class was pre-existing.

StephenHodgson:
We shouldn't be messing with the default debug stuff.

davidkline-ms:
I can remove the ( pre-existing :) ) DebugUtils class if we think that is the right thing to do.  The visibility change (and some doc comments) is what went in

StephenHodgson:
Also the internal stuff will be changed in my next PR but it has conflicts;

davidkline-ms:
i'll file an issue to pull DebugUtils and we can take your internal changes. Sound good?

StephenHodgson:
I'm working it out.

PR #2321, opened 2018-06-21
updated custom assembly definitions.

StephenHodgson:
Ensured that EnforceEditorSettings is 3.5 compliant so it will successfully compile and reload.

Still waiting on a fix from Unity for the correct assembly reloading.
See my [forum post here](https://forum.unity.com/threads/custom-assemblies-are-not-reloaded-if-there-is-a-compile-error.537145/).
See my [bug report here](https://fogbugz.unity3d.com/default.asp?1032070_2h7qac6ck84v34ee).

I chose to make the default scripting backend target 4.6 so it's easier for people to use the project switching to write in the native windows namespace.


Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityToolkit.SDK.Input.asmdef
```diff
@@ -1,6 +1,7 @@
 {
     "name": "Microsoft.MixedReality.Toolkit.SDK.Input",
```
SimonDarksideJ: Daft question, but do we really need to subdivide the SDK project, other than editor code?
This seems overkill as the SDK layer is a "Take it or Leave it" type of project.

Assets/MixedRealityToolkit/_Core/BuildAndDeploy/BuildDeployPreferences.cs
```diff
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using Microsoft.MixedReality.Toolkit.Internal.Utilities.Editor.Setup;
+using Microsoft.MixedReality.Toolkit.Internal.Utilities.Editor;
```
SimonDarksideJ: Why remove "Setup" as it was a good clear distinction from the Editor setup code and the rest of the project?

StephenHodgson: It wasn't removed.

StephenHodgson: The namespace was refactored and these classes are not part of setup.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityToolkit.SDK.Input.asmdef
StephenHodgson: This is mainly to ensure that the assemblies are built in the right order.

StephenHodgson: The custom assembly definitions give us clear dependencies, and complication order.

SimonDarksideJ: Not for the SDK, it's not needed.  It's meant to be the simplest of projects, just drop in and use.   Doesn't need the same level of complexity as the main project.

StephenHodgson: If you'd like to break the out later to make it work be my guest. 

But you'll likely run into compile issues

StephenHodgson: Done

Issue #2347, opened 2018-06-25
Task vNext - Remove PositionOnly and RotationOnly states. Add position accuracy check


davidkline-ms:
## Overview
From past (DirectInput) experience and a lengthy discussion with the platform PM team, we believe that most applications are going to run code that looks like:
```
if (controllerState == Tracked) 
{ DoSomething(); }
else
{ ControllerLostHandling(); }
```

This would cause the PositionOnly and RotationOnly states to be largely ignored.

WMR (and others?) provides approximate positioning when the controller leaves the tracking sensor's view, so the controller is always Tracked (once initial tracking occurs). To provide applications that wish to take advantage of the knowledge that the controller is "behind the head", we propose adding IsPositionApproximate() to the controller interface and allow specific controller classes to implement as appropriate.

## Requires
* Removal of ControllerState.PositionOnly
* Removal of ControllerState.OrientationOnly
* Update of any code that refences these states to instead use ControllerState.Tracked
* Addition of "bool IsPositionApproximate()" method to IMixedRealityController interface
* Default implementation of IsPositionApproximate in BaseController class
* Full implementation of IsPositionApproximate in WindowsMixedRealityController class
* Documentation for IsPositionApproximate

## Acceptance Criteria
Verify ControllerState is:
- [x] NotTracked until the first time the controller is seen by the tracking sensors
- [x] Tracked from that point forward (even when not visible to the sensors)

Confirm that new IsPositionApproximate() method returns:
- [ ] True if the controller is "behind the head"
- [ ] False when the controller is visible to the HMD controller tracking sensor

StephenHodgson:
I'm not sure we will need this change. I added approximate tracking to the controller state Enum.

StephenHodgson:
I'm not sure I understand exactly what you're trying to propose

davidkline-ms:
I am proposing to simplify the ControllerStates to essentially NotTracked and Tracked (with None and Other) and to allow advanced scenarios to inquire as to if the position data returned is approximate (this matches the underlying WMR platform and possibly others).

Apps that use this would look like:

if (controller.State == ControllerState.Tracked)
{
    if (controller.IsPositionApproximate())
    { HandleApproximatePositioning(); }
    else
    { HandlePrecicePositioning(); }
}

davidkline-ms:
[...]

PR #2350, opened 2018-06-25
Mrtk v next input action rollback2

StephenHodgson:
Rolling back the interface implementation to fix issues with profiles.

PR #2351, opened 2018-06-26
More Device changes.

StephenHodgson:
Overview
---
- Added a really good reason to keep controller types public.
- Added Tracking Events to Input System.
- Renamed ControllerState to TrackingState (Cause it can be for headsets too).
- Added changed check to all WMR Controller device updates so we don't spam the Input System unless something actually changes.
- Misc cleanup.
- Addresses some concerns with #2347


Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
```diff
@@ -155,13 +154,11 @@ public virtual void OnSourceDetected(SourceStateEventData eventData)
...
+            var controller = eventData.Controller as WindowsMixedRealityController;
 #if UNITY_WSA
-            // TODO Figure out a better way to know what type of source is found / lost.
-            InteractionSourceKind sourceKind;
-            if (InteractionInputSources.TryGetSourceKind(eventData.SourceId, out sourceKind) &&
-                sourceKind == InteractionSourceKind.Hand)
+            if (controller?.LastSourceStateReading.source.kind == InteractionSourceKind.Hand)
             {
-                visibleHandsCount--;
+                visibleHandsCount++;
```
keveleigh: OnSourceLost seems like it should still decrement here.

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
```diff
@@ -483,40 +486,51 @@ private void RemoveSource(IMixedRealityInputSource source)
...
-            sourcePositionEventData.Initialize(source, position);
+            sourcePositionEventData.Initialize(source, controller, rotation);
+
+            // Pass handler through HandleEvent to perform modal/fallback logic
+            HandleEvent(sourcePositionEventData, OnSourcePositionChangedEventHandler);
+        }
+
+        /// <inheritdoc />
+        public void RaiseSourcePositionChanged(IMixedRealityInputSource source, IMixedRealityController controller, SixDof position)
+
```
keveleigh: Nit: stray empty line

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
```diff
@@ -37,46 +44,57 @@ public class SourcePositionEventData : SourceStateEventData
...
         /// </summary>
         /// <param name="inputSource"></param>
+        /// <param name="controller"></param>
         /// <param name="position"></param>
-        public void Initialize(IMixedRealityInputSource inputSource, SixDof position)
+        public void Initialize(IMixedRealityInputSource inputSource, IMixedRealityController controller, SixDof position)
         {
-            Initialize(inputSource);
+            Initialize(inputSource, controller);
             SixDofPosition = position;
```
keveleigh: `SixDofPosition` seems misleadingly named. Also, does it give us anything that querying `ThreeDofPosition` and `ThreeDofRotation` (perhaps in addition to TrackingState) wouldn't?

StephenHodgson:
@keveleigh This pr is blocked by the prev one. I'm going to close it so we don't get side tracked.

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
StephenHodgson: How is the name misleading?

The next name I thought of was Pose, but I thought people wouldn't understand right away.

StephenHodgson: > . Also, does it give us anything that querying ThreeDofPosition and ThreeDofRotation (perhaps in addition to TrackingState) wouldn't?

I don't understand the question. Nothing here is being "queried", as it were. We're simply raising an event that the data in question was changed, for all the registered handler's to utilize.

keveleigh: Just the "position" part of it, to me, since it's position + rotation. If that's commonly referred to as "six dof position", then we're good. I feel like I've normally heard it referred to having "six degrees of freedom", with the inference being position + rotation.

StephenHodgson: >I feel like I've normally heard it referred to having "six degrees of freedom", with the inference being position + rotation.

Agreed, but Pose is also often used as a form of both Position & Rotation without scale.  Hmm

keveleigh: > We're simply raising an event that the data in question was changed

Right, but do we get anything out of raising an event with a SixDof that we wouldn't get from raising an event with the two ThreeDofs? Seems like we're tracking duplicate data.

keveleigh: > Pose is also often used as a form of both Position & Rotation without scale

Ooh, I personally like using pose

StephenHodgson: I'm just afraid that most people won't know what it is right away.

StephenHodgson: >Right, but do we get anything out of raising an event with a SixDof that we wouldn't get from raising an event with the two ThreeDofs? Seems like we're tracking duplicate data.

Better to raise a single event instead of two.

StephenHodgson: >Seems like we're tracking duplicate data.

Not sure I understand.  The 3dof data isn't being set when we raise a sixdof event.

keveleigh: > Not sure I understand

Possibly me neither! I'll take a better look tomorrow

StephenHodgson:
Blocked by #2350

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
```diff
@@ -428,10 +430,10 @@ public IMixedRealityInputSource RequestNewGenericInputSource(string name, IMixed
         #region Input Source State Events
 
         /// <inheritdoc />
-        public void RaiseSourceDetected(IMixedRealityInputSource source)
+        public void RaiseSourceDetected(IMixedRealityInputSource source, IMixedRealityController controller = null)
```
davidkline-ms: When would null be valid for controller? (in both Detected and Lost events)

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
```diff
@@ -484,40 +486,50 @@ private void RemoveSource(IMixedRealityInputSource source)
...
-            sourcePositionEventData.Initialize(source, rotation);
+            sourcePositionEventData.Initialize(source, controller, position);
 
             // Pass handler through HandleEvent to perform modal/fallback logic
             HandleEvent(sourcePositionEventData, OnSourcePositionChangedEventHandler);
         }
 
         /// <inheritdoc />
-        public void RaiseSourcePositionChanged(IMixedRealityInputSource source, SixDof position)
+        public void RaiseSourcePositionChanged(IMixedRealityInputSource source, IMixedRealityController controller, Quaternion rotation)
```
davidkline-ms: RaiseSourceRotationChanged()

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
```diff
@@ -484,40 +486,50 @@ private void RemoveSource(IMixedRealityInputSource source)
...
         {
             // Create input event
-            sourcePositionEventData.Initialize(source, position);
+            sourcePositionEventData.Initialize(source, controller, rotation);
+
+            // Pass handler through HandleEvent to perform modal/fallback logic
+            HandleEvent(sourcePositionEventData, OnSourcePositionChangedEventHandler);
```
davidkline-ms: HandleEvent(sourceRotationEventData, OnSourceRotationChangedEventHandler);

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
```diff
@@ -484,40 +486,50 @@ private void RemoveSource(IMixedRealityInputSource source)
...
+        }
+
+        /// <inheritdoc />
+        public void RaiseSourcePositionChanged(IMixedRealityInputSource source, IMixedRealityController controller, SixDof position)
```
davidkline-ms: RaiseSourcePoseChanged()
position -> Pose

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
```diff
@@ -484,40 +486,50 @@ private void RemoveSource(IMixedRealityInputSource source)
...
+        {
+            // Create input event
+            sourcePositionEventData.Initialize(source, controller, position);
 
             // Pass handler through HandleEvent to perform modal/fallback logic
             HandleEvent(sourcePositionEventData, OnSourcePositionChangedEventHandler);
```
davidkline-ms: HandleEvent(sourceTransformEventData, OnSourceTransformChangedEventHandler)

Assets/MixedRealityToolkit/_Core/Definitions/Devices/TrackingState.cs
```diff
@@ -18,6 +18,10 @@ public enum ControllerState
         /// </summary>
         Other,
         /// <summary>
+        /// The controller is tracked, but has approximate positioning.
```
davidkline-ms: Recommend removing this in favor of the bool in #2347 to simplify checking if a device is tracked.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -104,11 +105,27 @@ public void UpdateController(InteractionSourceState interactionSourceState)
         private void UpdateControllerData(InteractionSourceState interactionSourceState)
         {
             LastSourceStateReading = interactionSourceState;
+            var lastState = TrackingState;
 
-            isControllerTracked = interactionSourceState.sourcePose.TryGetPosition(out currentControllerPosition);
-            ControllerState = isControllerTracked ? ControllerState.Tracked : ControllerState.NotTracked;
+            switch (interactionSourceState.sourcePose.positionAccuracy)
```
davidkline-ms: Recommend  removing this until we land the decision on #2347

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
```diff
@@ -37,46 +44,57 @@ public class SourcePositionEventData : SourceStateEventData
...
         }
 
         /// <summary>
         /// Populates the event with data.
         /// </summary>
         /// <param name="inputSource"></param>
+        /// <param name="controller"></param>
         /// <param name="rotation"></param>
-        public void Initialize(IMixedRealityInputSource inputSource, Quaternion rotation)
+        public void Initialize(IMixedRealityInputSource inputSource, IMixedRealityController controller, Quaternion rotation)
```
davidkline-ms: I'd rather see Rotation be it's own event / data to avoid confusion (orientation != position)

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
```diff
@@ -37,46 +44,57 @@ public class SourcePositionEventData : SourceStateEventData
...
         }
 
         /// <summary>
         /// Populates the event with data.
         /// </summary>
         /// <param name="inputSource"></param>
+        /// <param name="controller"></param>
         /// <param name="position"></param>
-        public void Initialize(IMixedRealityInputSource inputSource, SixDof position)
+        public void Initialize(IMixedRealityInputSource inputSource, IMixedRealityController controller, SixDof position)
```
davidkline-ms: I'd rather see SixDof/Pose be it's own event / data as it is more than just position.

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
davidkline-ms: I recomment using Transform as that is what SixDof is, the position AND rotation.

davidkline-ms: Though it is lacking scale that is commonly in a transform, so maybe pose is the best name :)

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSystem.cs
```diff
@@ -115,41 +116,55 @@ public interface IMixedRealityInputSystem : IMixedRealityEventSystem
         /// Raise the event that the Input Source was detected.
         /// </summary>
         /// <param name="source">The detected Input Source.</param>
-        void RaiseSourceDetected(IMixedRealityInputSource source);
+        /// <param name="controller"></param>
+        void RaiseSourceDetected(IMixedRealityInputSource source, IMixedRealityController controller = null);
```
davidkline-ms: Dont see how you can detect / lose a null controller

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
StephenHodgson: Yes for input sources that are not tied to a controller.

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
StephenHodgson: Will be renamed to `HandleEvent(sourceRotationEventData, OnSourcePoseChangedEventHandler);`

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
StephenHodgson: Will be renamed to 
```
HandleEvent(sourceTransformEventData, OnSourcePoseChangedEventHandler)
```

Assets/MixedRealityToolkit/_Core/Definitions/Devices/TrackingState.cs
StephenHodgson: I disagree

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
StephenHodgson: I disagree

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
StephenHodgson: I disagree.  This event data class will be renamed to 
```
SourcePoseEventData.cs
```

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
StephenHodgson: I disagree, I don't want users to have lots of interlaces to implement to get similar data.

StephenHodgson:
@davidkline-ms does it make sense to rename `SixDof` to `Pose`?

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/IMixedRealityInputSystem.cs
StephenHodgson: For example, voice does not have a controller.

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
StephenHodgson: I disagree, I don't want users to have lots of interlaces to implement to get similar data.

StephenHodgson: although I am open to this idea a bit.

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePoseEventData.cs
```diff
@@ -32,52 +39,63 @@ public class SourcePositionEventData : SourceStateEventData
...
         /// <param name="inputSource"></param>
+        /// <param name="controller"></param>
         /// <param name="position"></param>
-        public void Initialize(IMixedRealityInputSource inputSource, SixDof position)
+        public void Initialize(IMixedRealityInputSource inputSource, IMixedRealityController controller, SixDof position)
         {
-            Initialize(inputSource);
-            SixDofPosition = position;
+            Initialize(inputSource, controller);
+            SixDofPose = position;
```
keveleigh: Going back to my comments from last night:

This event data currently has 4 things it might contain:
1. `TwoDofPosition`
2. `ThreeDofPosition`
3. `ThreeDofRotation`
4. `SixDofPose`

What's the benefit of providing `SixDofPose` at all, instead of using the existing `ThreeDofPosition` and `ThreeDofRotation`. Additionally, how do I even know which value to look at?

Assets/MixedRealityToolkit/_Core/EventDatum/Input/PoseInputEventData.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.MixedReality.Toolkit.Internal.EventDatum.Input
     /// <summary>
     /// Describes an Input Event that involves an Input Source's spatial position AND rotation.
     /// </summary>
-    public class SixDoFInputEventData : ThreeDoFInputEventData
+    public class PoseInputEventData : ThreeDofInputEventData
```
keveleigh: What is the difference between this and `SourcePoseEventData`?

Assets/MixedRealityToolkit/InputSystem/MixedRealityInputManager.cs
davidkline-ms: Do we get detected | lost events in those cases?

Assets/MixedRealityToolkit/_Core/EventDatum/Input/PoseInputEventData.cs
StephenHodgson: This is for actions, where as the SourcePoseEventData is only for input sources

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePoseEventData.cs
StephenHodgson: I didn't want to break up the interface implementation into more things users have to implement to get the data.

That way they can handle all types of source pose changes.

I'm open to splitting them up, but I didn't want a lot of interfaces for handlers.

keveleigh: I think I understand and am pretty sure I agree with not wanting to split them up, but I still have those questions:

1. How do I know which values in the event data I should care about? 
2. The benefit of SixDof vs using the two ThreeDof values. This would also reduce the number of places for me to check for populated data

StephenHodgson: You can implement all the interfaces and if you happen to get the data you need, then you process it.

StephenHodgson:
Looks like I didn't have to make a new type for Pose (SixDof)

https://docs.unity3d.com/2018.1/Documentation/ScriptReference/Pose.html

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePositionEventData.cs
davidkline-ms: I see where you are renaming PositionEventData to PoseEventData. I like that!

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePoseEventData.cs
davidkline-ms: What if we made the data in the event args nullables? We can then respond to the pose changed event and check by saying:
if (twoDofData.HasValue) { … }
if (threeDofPositionData.HasValue) { … }
etc.

keveleigh: Agreed. There has to be some way to actually know 

>if you happen to get the data you need

keveleigh: Agreed. There has to be some way to actually know 

>if you happen to get the data you need

Assets/MixedRealityToolkit-Tests/InteractionDefinitionTests.cs
```diff
@@ -439,5 +439,265 @@ public void Test14_TestSixDofNoChange()
...
+            Assert.IsFalse(initialValue.Changed);
+
+            initialValue.SetBoolValue(testValue1);
+
+            Assert.IsTrue(initialValue.Changed);
+
+            var setValue1 = interactions[0];
+
+            Assert.IsNotNull(setValue1);
+            Assert.IsTrue(setValue1.GetBooleanValue());
```
SimonDarksideJ: Can we standardise on the Bool get / set.
Currently it is:
* GetBooleanValue
* SetBoolValue

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
```diff
@@ -3,6 +3,7 @@
 
 using Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem;
 using Microsoft.MixedReality.Toolkit.Internal.Definitions.Physics;
+using Microsoft.MixedReality.Toolkit.Internal.Devices.WindowsMixedReality;
```
SimonDarksideJ: WOAH, why is there a WindowsMixedReality reference in Cursors.  
Only the devices space and the MR manager should have ANY knowledge of the Devices

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
```diff
@@ -133,11 +134,9 @@ public bool IsVisible
         /// <param name="eventData"></param>
         public virtual void OnSourceDetected(SourceStateEventData eventData)
         {
+            var controller = eventData.Controller as WindowsMixedRealityController;
 #if UNITY_WSA
```
SimonDarksideJ: Should only be using the input system (I know this is on the TODO)

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePoseEventData.cs
```diff
@@ -32,52 +39,63 @@ public class SourcePositionEventData : SourceStateEventData
         /// <summary>
         /// The new position and rotation of the input source.
         /// </summary>
-        public SixDof SixDofPosition { get; private set; } = SixDof.ZeroIdentity;
+        public SixDof SixDofPose { get; private set; } = SixDof.ZeroIdentity;
```
SimonDarksideJ: Do we need SixDof now that you've found this Unity "Pose" type?

Assets/MixedRealityToolkit-Tests/InteractionDefinitionTests.cs
StephenHodgson: Def. I might update that when I rename SixDof to MixedRealityPose

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
StephenHodgson: When we ask the controller if it's a hand below

SimonDarksideJ: Then we need a way for the framework to realise it.  NOT pull it from the device platform.

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
SimonDarksideJ: We can get the data directly VIA the interactionMapping!!!!

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePoseEventData.cs
SimonDarksideJ: Strike a pose

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePoseEventData.cs
StephenHodgson: I don't want to make it complicated with nullable types. I'd rather split them up into separate events before we do that.

StephenHodgson: >>Agreed. There has to be some way to actually know

>if you happen to get the data you need

Looking back at the handler, I noticed we only have a single method call `OnSourcePoseChanged`

I propose we add specific methods to know which data types got updated:

```
OnSourceChanged(); // Gets raised for them all.
OnSource2DofPositionChanged(); // Only gets raised when a Vector2 change happens
OnSource3DofPositionChanged(); // Only gets raised when a Vector3 change happens
OnSourceRotationChanged(); // Only gets raised when a Rotation change happens
OnSourcePoseChanged(); // Only gets raised when a Pose change happens.
```

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
StephenHodgson: Oh, fun fact - Source State changes do not have actions.

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
```diff
@@ -133,11 +133,9 @@ public bool IsVisible
         /// <param name="eventData"></param>
         public virtual void OnSourceDetected(SourceStateEventData eventData)
         {
+            var controller = eventData.Controller as WindowsMixedRealityController;
```
davidkline-ms: shouldnt this line be inside the UNITY_WSA condition?

Assets/MixedRealityToolkit/_Core/Definitions/Devices/TrackingState.cs
```diff
@@ -4,34 +4,26 @@
...
-        /// <summary>
-        /// The controller is currently only returning position data.
-        /// </summary>
-        PositionOnly,
-        /// <summary>
-        /// The controller is currently only returning orientation data.
+        /// Reserved, for systems that provide alternate tracking.
         /// </summary>
-        OrientationOnly
+        Other,
```
davidkline-ms: Other should follow None for easier expansion if needed

Assets/MixedRealityToolkit/_Core/Definitions/Devices/TrackingState.cs
```diff
@@ -4,34 +4,26 @@
...
+        /// The controller is currently not tracked.
         /// </summary>
-        None = 0,
+        NotTracked = 0,
         /// <summary>
-        /// Reserved, for systems that provide alternate tracking.
+        /// The controller is tracked, but has approximate positioning.
         /// </summary>
-        Other,
+        Approximate,
```
davidkline-ms: Not needed with #2347 (we may need to support approximate orientation on some platform)

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/MixedRealityPose.cs
```diff
@@ -8,14 +8,14 @@
 namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities
 {
     [Serializable]
-    public struct SixDof : IEqualityComparer
+    public struct MixedRealityPose : IEqualityComparer
```
davidkline-ms: I like this!

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/MixedRealityPose.cs
```diff
@@ -25,7 +25,7 @@ public SixDof(Vector3 position, Quaternion rotation)
         /// Constructor.
         /// </summary>
         /// <param name="position"></param>
-        public SixDof(Vector3 position)
+        public MixedRealityPose(Vector3 position)
```
davidkline-ms: Recommend 
public MixedRealityPose(Vector3 position) : this(position, Quaternion.identity) { }

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/MixedRealityPose.cs
```diff
@@ -35,7 +35,7 @@ public SixDof(Vector3 position)
         /// Constructor.
         /// </summary>
         /// <param name="rotation"></param>
-        public SixDof(Quaternion rotation)
+        public MixedRealityPose(Quaternion rotation)
```
davidkline-ms: Recommend 
public MixedRealityPose(Quaternion rotation) : this(Vector3.zero, rotation) { }

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/MixedRealityPose.cs
```diff
@@ -47,7 +47,7 @@ public SixDof(Quaternion rotation)
         /// <returns>
         /// <see cref="Vector3.zero"/> and <see cref="Quaternion.identity"/>.
         /// </returns>
-        public static SixDof ZeroIdentity { get; } = new SixDof(Vector3.zero, Quaternion.identity);
+        public static MixedRealityPose ZeroIdentity { get; } = new MixedRealityPose(Vector3.zero, Quaternion.identity);
```
davidkline-ms: Not fond of the name "ZeroIdentity" can we call it "Neutral" or something similar?

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
```diff
@@ -90,16 +85,16 @@ public void SetupConfiguration(Type controllerType)
         /// Load the Interaction mappings for this controller from the configured Controller Mapping profile
         /// </summary>
         /// <param name="mappings">Configured mappings from a controller mapping profile</param>
-        protected void AssignControllerMappings(MixedRealityInteractionMapping[] mappings)
+        public void AssignControllerMappings(MixedRealityInteractionMapping[] mappings)
```
davidkline-ms: Do we also need/want AddControllerMapping() and RemoveControllerMapping()?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -104,11 +105,27 @@ public void UpdateController(InteractionSourceState interactionSourceState)
...
+            {
+                case InteractionSourcePositionAccuracy.None:
+                    TrackingState = TrackingState.NotTracked;
+                    break;
+                case InteractionSourcePositionAccuracy.Approximate:
```
davidkline-ms: Recommend removing as #2347 handles this and we anticipate there may be a platform that has approximate orientation as well (which would break the hierarchy implied here).

Assets/MixedRealityToolkit/_Core/Interfaces/InputSystem/Handlers/IMixedReality3DoFInputHandler.cs
```diff
@@ -6,14 +6,20 @@
...
     /// </summary>
     public interface IMixedReality3DoFInputHandler : IMixedRealityInputHandler
     {
         /// <summary>
-        /// 3 Degree of Freedom input update.
+        /// Raised when the input source's position has changed.
         /// </summary>
         /// <param name="eventData"></param>
-        void On3DoFInputChanged(ThreeDoFInputEventData eventData);
+        void OnPositionChanged(PositionInputEventData eventData);
```
davidkline-ms: 3DoF is position or orientation. This feels more like a 6DoF handler. Can we rename this method as On3DoFDataChanged and remove OnRotationChanged? I think  we may want to add a hint to the eventData object.

davidkline-ms: Please ignore the above... in taking offline with @keveleigh, I get the value of the two methods.

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
StephenHodgson: Yes, if someone decides they want to use a new controller mapping at runtime.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/MixedRealityPose.cs
StephenHodgson: Unity calls theirs `Pose.identity` Why their public properly is pascalCase I'll never know,

Assets/MixedRealityToolkit/_Core/Definitions/Devices/TrackingState.cs
StephenHodgson: I'd still like to keep it in case we will need it on other platforms.  I'm fine with removing position and rotation only, but this will stay.

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
StephenHodgson: No, because we've already wrapped it in the controller class.

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePoseEventData.cs
StephenHodgson: Guys take a look at my most recent updates. This should fix a lot of the issues you guys had with this.

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
SimonDarksideJ: Ok, we'll create a task to review these requirements later

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/MixedRealityPose.cs
SimonDarksideJ: Let's go with "Identity" as that is the common standard.  Although I personally liked ZeroIdentity as it did exactly what it said on the tin :D (A Zero Vector 3 with an Identity Quaternion)

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
SimonDarksideJ: I think we should ***add a task for this later***.  The mappings are meant to be static for the MVP.  it's a very limited use case (in any VR solution at the moment) where you would want to dynamically change mappings on the fly.

StephenHodgson: I remember this being a selling point with @Railboy

Assets/MixedRealityToolkit/InputSystem/Cursors/BaseCursor.cs
davidkline-ms: I can live with that, it was just a touch confusing at first

PR #2354, opened 2018-06-28
Button.cs eventData fixes

### LINKED ISSUE #2295: Button class does not block events from reaching fallbackInputStack
## Overview
Button class (HoloToolkit.Unity.Buttons.Button) does not mark received Event Data as used. This causes events to be handled twice, once by the button callbacks, and once by handlers inside InputManager.fallbackInputStack.

## Expected Behavior
From [/Assets/HoloToolkit/Input/README.md](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit/Input/README.md)

> handle various types of input and send them to any game object being currently gazed at, **or** any fallback object.

This means that events sent to the button shouldn't be sent to the fallback object.

## Actual Behavior
Pressing a button also activates both the fallback object's method and the callback on the button.

## Steps to reproduce

1. `git clone https://github.com/SujeathPareddy/MixedRealityToolkit-Unity.git`
2. `git checkout origin/june18_dev`
3. `git checkout 9a44898316c89` (This is an intermediate commit before I fixed the bug)
3. Open Assets/repro scene
4. Press play
5. Shift + left click on the circular button
6. Observe that two types of callbacks are emitted in the console. (relevant code is in attachtobutton.cs). The callback registered directly with the button is executed twice, possibly related to #1879.

## Fix

The bug is in Button.cs. The passed eventData parameter should call EventData.Use() before returning. 

The patched version is [here](https://raw.githubusercontent.com/SujeathPareddy/MixedRealityToolkit-Unity/june18_dev/Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs).

Demo of potentially fixed code here:
`git checkout origin/june18_dev`

Note on behaviour: If the button has no registered callbacks, then the event is not marked as used. This is a matter of opinion. I believe this makes the most sense in my app's context. Perhaps there should be a flag on the Button class.

## Unity Editor Version
2017.3 and 2017.4

2017.4.0.0

SujeathPareddy:
Overview
---

Fixed #2295. Marked eventData objects as used inside Button to ensure events do not propagate to downstream handlers.

Changes
---
- Fixes: # 2295.

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
+++ b/Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
@@ -246,8 +246,8 @@ namespace HoloToolkit.Unity.Buttons
         /// </summary>
         public void OnFocusExit(PointerSpecificEventData eventData)
         {
-             if (!m_disabled) // && FocusManager.Instance.IsFocused(this))
-             {
+            if (!m_disabled) // && FocusManager.Instance.IsFocused(this))
+            {
                 if (ButtonState == ButtonStateEnum.Pressed)
                 {
                     DoButtonCanceled();
--- a/Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
+++ b/Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
@@ -141,6 +141,7 @@ namespace HoloToolkit.Unity.Buttons
                     // Set state to Pressed
                     ButtonStateEnum newState = ButtonStateEnum.Pressed;
                     this.OnStateChange(newState);
+                    eventData.Use();
                 }
             }
         }
@@ -156,6 +157,7 @@ namespace HoloToolkit.Unity.Buttons
                 if (ButtonPressFilter == InteractionSourcePressInfo.None || ButtonPressFilter == eventData.PressType)
                 {
                     DoButtonReleased();
+                    eventData.Use();
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


keveleigh:
Looks like the Button changes weren't included in this PR.

SujeathPareddy:
Apologies, I added the metafile twice by accident. It's pushed now.

Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
```diff
@@ -240,7 +247,7 @@ public void OnFocusEnter(PointerSpecificEventData eventData)
         public void OnFocusExit(PointerSpecificEventData eventData)
         {
              if (!m_disabled) // && FocusManager.Instance.IsFocused(this))
-            {
+             {
```
keveleigh: Extra space here (there might be one on the above `if` statement, which is what was throwing this line off).

Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs.meta
```diff
@@ -1,13 +1,13 @@
 fileFormatVersion: 2
-guid: da8f80956d894644bb62bc848d89ede8
-timeCreated: 1500484825
-licenseType: Free
+guid: 1fb6040b8ecac06468282404a86299e4
```
keveleigh: Meta changes still here.

Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
SujeathPareddy: Doesn't the original line have a missing space? I just thought adding the space makes it conform.

SujeathPareddy:
All changes made.

davidkline-ms:
@SujeathPareddy 

Thanks for contributing!

Issue #2040, opened 2018-05-03
PopupMenu: 'IsModal' and 'CloseOnNonTargetedTap' options don't work correctly


ivan2007:
## Overview
'IsModal' and 'CloseOnNonTargetedTap' options of the 'PopupMenu' component don't work correctly
## Expected Behavior
The options should work
## Actual Behavior
The options don't work correctly in all the situations
## Steps to reproduce
InputManagerTest scene
IsModal:
Click the modal Popup button and then click the cylinder on the left
CloseOnNonTargetedTap:
Click the cylinder on the left (only if you remove the option 'CloseOnTargetedTap' the example works correctly).
## Unity Editor Version
2017.4.2f2
## Mixed Reality Toolkit Release Version
2017.2.1.4

psytgm:
The situation of triggering twice with one click also occurred when scene switched between each other.

keveleigh:
Ahh, so it looks like PopupMenu is only consuming `OnInputDown` and `OnInputUp` events, while the cylinder uses `OnInputClicked`. It doesn't make sense to me for PopupMenu to respond to those events over clicked (and it's responding identically to both, strangely). I'll put out an update and test these settings.

keveleigh:
I just opened #2358 with a potential fix. Please take a look and let me know if something still doesn't work as expected.

ivan2007:
Hi @keveleigh, I made some test with InputManagerTest scene:
1. When you open the modal window you are still able to move the movable objects (I don't know if this is outside the scope of PopupMenu).
2. The button actions of Shrink and Grow are set with reset.
3. If you place a CheckBoxSet inside the Popup window (IsModal is true) you can't change the state of the check boxes. The modal input stack of InputManager stop the event only if it is marked with used (see HandleEvent).

PR #2358, opened 2018-06-28
PopupMenu options fix

### LINKED ISSUE #1535: Crash calling Dismiss() on PopupMenu if rootAnimator is null
## Overview
Calling Dismiss on PopupMenu with no Animator
## Expected Behavior
No crash
## Actual Behavior
Crash
## Steps to reproduce
Call Dismiss on PopupMenu with no Animator
The problem is inside Dismiss():

            if (rootAnimator.isInitialized)
            {
                rootAnimator.SetTrigger("Dehydrate");
            }
            else
            {
                gameObject.SetActive(false);
            }

## Unity Editor Version
2017.2.1.f1
Master, Dev, Stabilization

keveleigh:
Overview
---
The PopupMenu was consuming `OnInputDown` and `OnInputUp` events, while it seemed like it should consume (and other similar components were using) `OnInputClicked`. I reworded the example scene to be clearer, and also decreased the number of places states were being handled. The PopupMenu routes its "close" input handling through a button, but was also receiving its own input. Now, everything goes through the button.

Changes
---
- Fixes: #2040, fixes #1535

### DIFF
--- a/Assets/HoloToolkit-Examples/Input/Scripts/PopupMenu.cs
+++ b/Assets/HoloToolkit-Examples/Input/Scripts/PopupMenu.cs
@@ -35,6 +35,8 @@ namespace HoloToolkit.Unity.InputModule.Tests
         /// </summary>
         private Action deactivatedCallback;
 
+        private int dehydrateButtonId;
+
         public PopupState CurrentPopupState = PopupState.Closed;
 
         public enum PopupState { Open, Closed }
@@ -42,6 +44,11 @@ namespace HoloToolkit.Unity.InputModule.Tests
         private void Awake()
         {
             gameObject.SetActive(false);
+
+            if (dehydrateButtonId == 0)
+            {
+                dehydrateButtonId = Animator.StringToHash("Dehydrate");
+            }
         }
 
         private void OnEnable()
@@ -118,9 +125,9 @@ namespace HoloToolkit.Unity.InputModule.Tests
             }
 
             // Deactivates the game object
-            if (rootAnimator.isInitialized)
+            if (rootAnimator != null && rootAnimator.isInitialized)
             {
-                rootAnimator.SetTrigger("Dehydrate");
+                rootAnimator.SetTrigger(dehydrateButtonId);
             }
             else
             {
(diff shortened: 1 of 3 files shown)


Assets/HoloToolkit-Examples/Input/Scripts/PopupMenu.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace HoloToolkit.Unity.InputModule.Tests
 {
-    public class PopupMenu : MonoBehaviour, IInputHandler
+    public class PopupMenu : MonoBehaviour
```
StephenHodgson: Why are we removing the input handler?

keveleigh: The pop-up menu was only using the input handler to direct modal input in. It additionally had a handler for when the close button is pressed (which essentially did the same thing). Instead, I'm reusing the same event, which means that only the button needs to handle the input events.

ivan2007:
Hi @keveleigh, I made some test with InputManagerTest scene:
1. When you open the modal window you are still able to move the movable objects (I don't know if this is outside the scope of PopupMenu).
2. The button actions of Shrink and Grow are set with reset.
3. If you place a CheckBoxSet inside the Popup window (IsModal is true) you can't change the state of the check boxes. The modal input stack of InputManager stop the event only if it is marked with used (see HandleEvent<T>).

keveleigh:
Hey @ivan2007:

1. Yeah, at the moment it only consumes clicks. If we set it to be the `OverrideFocusedObject` instead of just modal, it should make it consume everything, which does appear to be the original intent.
2. I'm not sure what this is referring to.
3. The PopupMenu as provided in the Examples folder doesn't seem to be written in a way to really be used outside of the example. You might be able to rework it to allow input to get to the CheckBoxSet if that's what's focused. Sounds useful, and perhaps generalizing this component and graduating it to the main Toolkit UX components would be a good addition.

ivan2007:
Hey @keveleigh,
2. I am referring to this:
buttonaction
perhaps it is already fixed.
3. Yes it will be very useful having a general component of this type. In any case PopupMenu is a good starting point.

Issue #1879, opened 2018-03-26
Button.OnButtonClicked is triggered twice on clicks


thiemok:
## Overview
`OnButtonClicked` is triggered twice on input clicks. Once from the InputClickedEvent and once from InputDown.
This happens spread over some frames and thus causes unintended side effects with dialogs and button actions that spawn gameObjects infront of the user. For example the first `InpudClicked` event will open a Dialog and the second will imediately dismiss it.

![buttonclicktrace](https://user-images.githubusercontent.com/5272379/37900764-750ff88c-30ef-11e8-8e49-06799e1e7b05.PNG)

## Expected Behavior
`OnInputClicked` is only fired once

## Actual Behavior
`OnInputClicked` is only fired twice

## Steps to reproduce
* Create empty scene
* Add default MRTK stuff
* Add Button
* Trace `OnButtonClicked`

## Unity Editor Version
2017.3.1f1

## Mixed Reality Toolkit Release Version
dev branch

braden-smith:
Similar with the HolographicButton prefab in dev working branch. Unity 2017.2.1p4

https://forums.hololens.com/discussion/6999/ui-button-on-click-event-firing-twice#latest

thiemok:
I'm currently solving this by simply debouncing the invocation of `OnButtonClicked` a few hundered milliseconds, but haven't had the time to test if this has any unintended side effects.

StephenHodgson:
`eventData.Use();`

thiemok:
`Button.OnButtonClicked` recieves a `GameObject` as payload, no eventData. the trace i attached also shows that the calls originate from two different event sources (`InputDown` and `InputClicked`).

StephenHodgson:
You'll have to show me how you've got it setup in the inspector and in code.

thiemok:
Unfortunately i just left my windows machine at work, but i will try to flesh out my description a bit.

# Reproduction
This seems to be not reproducible in the Unity Editor, it only fails on the HoloLens itself.
* Create an empty scene
* Add MRTK defaults (InputManager, Cursor, Camera, ...)
* Add an instance of the HolographicButton prefab (or any other button that uses the Button script)
* Add the following script to the button (Please forgive smaller errors, since i'm tiping this of the top of my head)
```csharp
using UnityEngine;
using MixedRealityToolkit.UX.Buttons;

class Test : MonoBehaviour {
  int frame = 0;

  private void Start() {
[...]

PR #2359, opened 2018-06-28
Button.cs documentation and formatting updates

keveleigh:
Overview
---
Cleans up Button.cs a bit, in preparation for a fix for #1879 incoming after merge.

Changes
---
- First PR towards a fix for #1879 (splitting up into smaller PRs for review ease).

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
+++ b/Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
@@ -134,7 +134,7 @@ namespace HoloToolkit.Unity.Buttons
         {
             if (enabled && !m_disabled)
             {
-                if(ButtonPressFilter == InteractionSourcePressInfo.None || ButtonPressFilter == eventData.PressType)
+                if (ButtonPressFilter == InteractionSourcePressInfo.None || ButtonPressFilter == eventData.PressType)
                 {
                     DoButtonPressed();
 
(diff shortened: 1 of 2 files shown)


Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
```diff
@@ -93,29 +97,55 @@ public bool RequireGaze
         /// <summary>
         /// Last state of hands being visible
         /// </summary>
-        private bool _bLastHandVisible = false;
+        private bool lastHandVisible = false;
```
StephenHodgson: Just keep in mind that renaming these will break the previous serialization people have in their projects.

StephenHodgson: [There is an attribute we should be using to make sure it doesn't break](https://docs.unity3d.com/ScriptReference/Serialization.FormerlySerializedAsAttribute.html).
```
[FormerlySerializedAs("myValue")]
private string myValue;
```

keveleigh: What serialization does this break? None of the names I changed have the `[SerializeField]` tag, or are you referring to something else?

keveleigh: Good to know about that `[FormerlySerializedAs]` tag! That seems very useful.

StephenHodgson: Oh, I thought some were public before, then it's no worries.

PR #2369, opened 2018-06-29
add IsPosition/RotationAvailable and IsPositionApproximate

davidkline-ms:
Overview
---
As discussed at shiproom (6/27), tracking states do not fall into a neat hierarchy; you can be tracked positionally, or rotationally, or both. 

Windows Mixed Reality, and potentially other platforms, provides a concept of position accuracy for tracked controllers. To allow developers to customize application behavior based on the accuracy of tracking, a IsPositionApproximate bool has been added to IMixedRealityController.

Changes
---
The TrackingState enum has been updated to have NotApplicable (device does not support tracking), NotTracked and Tracked. Tracked here means either position or rotation (platforms can decide if position and rotation are independent or hierarchical).

To let the developer know if a device has data for position and rotation, properties for IsPositionAvailable and IsRotationAvailable have been added to the Headset struct and the IMixedRealityController interface.

The WindowsMixedRealityController has been updated to report these new properties.

- Fixes: #2347 .


Assets/MixedRealityToolkit/_Core/Definitions/Devices/TrackingState.cs
```diff
@@ -4,26 +4,33 @@
...
     /// <summary>
-    /// The Tracking State defines how a controller or headset is currently being tracked.
-    /// This enables developers to be able to handle non-tracked situations and react accordingly
+    /// The Tracking State defines how a device is currently being tracked.
+    /// This enables developers to be able to handle non-tracked situations and react accordingly.
     /// </summary>
+    /// <remarks>
+    /// Tracking is being defined as receiving sensor (positional and/or rotational) data from the device.
+    /// </remarks>
     public enum TrackingState
```
StephenHodgson: I'd like to either use the new enum and get rid of this one, or not make a enum.

SimonDarksideJ: Can you translate that please @StephenHodgson
I think for not, these changes should be fine and we can refine later once we have a working implementation and can evolve.

StephenHodgson: I just think both of the Enums do the same job.  I find it a bit confusing about which to use.

davidkline-ms: I can see removing TrackingState in favor of reporting PositionAccuracy and RotationAccuracy.

This would allow for Stephen's desire to have a hierarchical enum as well as keep positional and rotational tracking separate.

I can spin up a commit today.

StephenHodgson: Thanks!

davidkline-ms: Folding TrackingState and TrackingAccuracy is causing more churn than I had hoped... I'm working my way through it. It's not bad, just a little deeper than anticipated.

davidkline-ms: Simplified this A LOT. By removing accuracy levels in favor of a bool (controller position only) for isApproximate.

PR #2371, opened 2018-06-30
Resolves null references if the input system is disabled with components relying on Input System

SimonDarksideJ:
Overview
---
Currently, if a scene starts that previously was using the InputSystem and the Input System was then disabled.  This caused a bucketload of Null reference checks / errors.

I expect review to consider an alternate way of Null testing or EnableInputSystem testing to ensure this should not generate errors in any manager use.

Changes
---
- Fixes: Null Reference Errors in Getmanager due to the manager not existing when called from GazeProvider
- Fixes: Null Reference Errors in Getmanager due to the manager not existing when called from FocusProvider
- Fixes: Null Reference Errors in Getmanager due to the manager not existing when called from GenericPointer
- Fixes: Input System Null Reference Errors in GenericPointer due to the manager not existing or inputsystem disabled

### DIFF
--- a/Assets/MixedRealityToolkit/InputSystem/Focus/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit/InputSystem/Focus/FocusProvider.cs
@@ -21,7 +21,7 @@ namespace Microsoft.MixedReality.Toolkit.InputSystem.Focus
     public class FocusProvider : MonoBehaviour, IMixedRealityFocusProvider
     {
         private IMixedRealityInputSystem inputSystem = null;
-        public IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
+        public IMixedRealityInputSystem InputSystem => inputSystem ?? (MixedRealityManager.Instance.ActiveProfile.EnableInputSystem ? inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>() : null);
 
         /// <summary>
         /// Maximum distance at which the pointer can collide with an object.
@@ -229,7 +229,7 @@ namespace Microsoft.MixedReality.Toolkit.InputSystem.Focus
         private void Start()
         {
             // Register the FocusProvider as a global listener to get input events.
-            InputSystem.Register(gameObject);
+            InputSystem?.Register(gameObject);
         }
 
         private void Update()
--- a/Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
@@ -138,7 +138,7 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Devices.WindowsMixedReality
             interactionSourceState.sourcePose.TryGetPosition(out currentPointerPosition, InteractionSourceNode.Pointer);
             interactionSourceState.sourcePose.TryGetRotation(out currentPointerRotation, InteractionSourceNode.Pointer);
 
-            if (CameraCache.Main.transform.parent != null)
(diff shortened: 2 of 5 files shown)


Assets/MixedRealityToolkit/InputSystem/Focus/FocusProvider.cs
```diff
@@ -21,7 +21,7 @@ namespace Microsoft.MixedReality.Toolkit.InputSystem.Focus
     public class FocusProvider : MonoBehaviour, IMixedRealityFocusProvider
     {
         private IMixedRealityInputSystem inputSystem = null;
-        public IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
+        public IMixedRealityInputSystem InputSystem => inputSystem ?? (MixedRealityManager.Instance.ActiveProfile.EnableInputSystem ? inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>() : null);
```
StephenHodgson: The Focus provider relies on our implementation of the input system. So if it isn't there, this should really break.

SimonDarksideJ: Bad @StephenHodgson 
If we put something in a scene, then we should either remove it or ensure it doesn't break if something we provide is turned off.  So no, it shouldn't break.  At worst you should get a proper warning

StephenHodgson: I thought I had cleaned this up better.

Do you have a proper repro?

SimonDarksideJ:
Lol, my implementation.  But my Bad I should have added replication steps.

1. Click on the "DefaultMixedRealityConfigurationProfile"
2. MR Manager added to scene
3. Run
4. Reality the inputsystem is on and I didn't want that thing!
5. Turn off the Input System in the config profile
6. Run,  hay  what are all these errors for things I didn't add in the scene !

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -138,7 +138,7 @@ private void UpdatePointerData(InteractionSourceState interactionSourceState, Mi
             interactionSourceState.sourcePose.TryGetPosition(out currentPointerPosition, InteractionSourceNode.Pointer);
             interactionSourceState.sourcePose.TryGetRotation(out currentPointerRotation, InteractionSourceNode.Pointer);
 
-            if (CameraCache.Main.transform.parent != null)
+            if (CameraCache.Main?.transform.parent != null)
```
StephenHodgson: This should never be null, unless you forgotten to tag your camera as `MainCamera`.
image

davidkline-ms: Which is easy for developers to do. I like having this check.

StephenHodgson: Let's put an assert in CameraCache.Main telling people to tag their camera properly.

StephenHodgson: That's the real fix.  This null check should never have to happen.

StephenHodgson:
I can live with an assert that it's not null at initialization time
But those systems are driven by their creators, so it shouldn't be null if the thing creating them is the thing we're checking.

MR Manager makes the Input System.
The Input System makes the Gaze/Focus providers.
If their input system ref is null or the MR manager ref is null they were implemented or created wrong.

StephenHodgson:
My last PR should have fixed the issue, so this PR can be closed. I opened a new PR for the camera cache.

PR #2376, opened 2018-07-02
Added asserts for things

StephenHodgson:
- Added null check assert for main camera tag in `CameraCache.Main`.
    - (This should probably be backported into master as well.)
- Added asserts in Gaze and Focus Providers for MR Manager and Input System.

### DIFF
--- a/Assets/MixedRealityToolkit/InputSystem/Focus/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit/InputSystem/Focus/FocusProvider.cs
@@ -228,6 +228,9 @@ namespace Microsoft.MixedReality.Toolkit.InputSystem.Focus
 
         private void Start()
         {
+            Debug.Assert(MixedRealityManager.IsInitialized, "No Mixed Reality Manager found in the scene.  Be sure to run the Mixed Reality Configuration.");
+            Debug.Assert(InputSystem != null, "No Input System found, Did you set it up in your configuration profile?");
+
             // Register the FocusProvider as a global listener to get input events.
             InputSystem.Register(gameObject);
         }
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit/InputSystem/Focus/FocusProvider.cs
```diff
@@ -228,6 +228,9 @@ private void Awake()
 
         private void Start()
         {
+            Debug.Assert(MixedRealityManager.IsInitialized, "No Mixed Reality Manager found in the scene.  Be sure to run the Mixed Reality Configuration.");
+            Debug.Assert(InputSystem != null, "No Input System found, Did you set it up in your configuration profile?");
```
davidkline-ms: Is this really an assertable condition? Arent we building in the flexibility for NOT using the input system?

StephenHodgson: > Arent we building in the flexibility for NOT using the input system?

We did, but in this case this is a component in _our_ implementation of the input system.

StephenHodgson: > Is this really an assertable condition?

And yes.

davidkline-ms: Then is it even possible to get into this state? Sorry if I am being dense today :)

StephenHodgson: Never. unless someone puts this component on by hand (Which shouldn't happen as we do it for them)

davidkline-ms: Thanks for clarifying... looks good to me.

PR #2387, opened 2018-07-03
Fixed errors when stopping editor playmode.

StephenHodgson:
- Disabled focus provider and added a check in the `UIRaycastCamera` creation to return if it's disabled.
- Added a flag to `GetOrAddController` to skip updating the controller when we remove it.


StephenHodgson:
@davidkline-ms just moved a line down.

Issue #1712, opened 2018-02-12
Slider range not properly working


Cycov:
## Overview
The slider minimum value is set to 0 and the `MinSliderValue` variable is used subtracted from maximum value
## Expected Behavior
When proprieties `MinSliderValue` and `MaxSliderValue` are set, the value of the slider with-the knob to left-most should be `MinSliderValue` and to the right-most should be `MaxSliderValue`
## Actual Behavior
`MinSliderValue` becomes 0 and `MaxSliderValue` becomes `MaxSliderValue - MinSliderValue`. After that it works as intended.
## Steps to reproduce
Add a simple slider and set interval to 0.5-5.0, formatting 0.00.
## Unity Editor Version
2017.2.1f1
## Mixed Reality Toolkit Release Version
2017.2.1.0

KansaiTraining:
Is this issue not being addressed? I got the same problem.

keveleigh:
This is fixed in the June dev branch and will be included in the next release.

PR #2388, opened 2018-07-03
Add support for non-0 min values in SliderGestureControl

keveleigh:
Overview
---
Currently, a non-centered SliderGestureControl assumes it will always have a min value of 0. If the min value is not zero, it will slide the range down until it starts at 0, decreasing the max value by the same amount.

This change also re-exposes a starting SliderValue in the editor, which was (inadvertently?) removed in #816, when the public float was turned into a property without serializing its backing field.

Changes
---
- Fixes: #1712

### DIFF
--- a/Assets/HoloToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
+++ b/Assets/HoloToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
@@ -7,7 +7,7 @@ using UnityEngine;
 namespace HoloToolkit.Examples.InteractiveElements
 {
     /// <summary>
-    /// updates slider UI based on gesture input
+    /// Updates slider UI based on gesture input
     /// </summary>
     public class SliderGestureControl : GestureInteractiveControl
     {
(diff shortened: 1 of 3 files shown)


Assets/HoloToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
```diff
@@ -83,6 +85,20 @@ protected override void Awake()
         {
             base.Awake();
 
+            if (MinSliderValue >= MaxSliderValue)
+            {
+                Debug.LogError("Your SliderGestureControl has a min value that's greater than or equal to its max value.");
+                Destroy(this);
```
StephenHodgson: Why don't we just disable it instead of destroying it?

keveleigh: Ahh, good question.
Disabling just the script didn't do anything, since it was driven through events. I updated this line to disable the entire GameObject, which prevents events being directed via focus.

keveleigh:
> I meant disabling the component.

I know, but see my comment above. Disabling the component doesn't do anything, since it's driven through events.

Assets/HoloToolkit-Examples/UX/Scripts/Controls/SliderGestureControl.cs
keveleigh: Okay so, the way this script is currently written, there's a GestureInteractive component that drives events into the SliderGestureControl. SliderGestureControl doesn't know about the GestureInteractive, so I'd have to search for it and disable both.

You were right about `enabled = false;`, if this was written as a single script responding to the input system's events (it feels like it probably could be, but that's outside the scope of this PR/issue).

Issue #2389, opened 2018-07-03
vNext Task: Port boundary support to camera profile system


davidkline-ms:
## Overview
This task adds an initial implementation for Boundary support for VR devices.

Modifications include:
- Removal of floor prefab instantiation in favor of the app creating it's own / context appropriate floor.
- Exposing the boundary volume so that applications can do custom rendering / etc.

## Requires
- Port of Boundary component of MixedRealityCameraParent to the new MixedRealityCameraProfile system

## Acceptance Criteria
- [x] User can configure the boundary profile
- [x] Supports specifying boundary height (feature parity)
- [x] An application can request the boundary volume

PR #2402, opened 2018-07-05
vNext: Implement MixedRealityBoundaryManager and Profile

davidkline-ms:
Overview
---
Defined the IMixedRealityBoundarySystem interface contents and implemented MixedRealityBoundaryManager.

Added MixedRealityBoundaryProfile/Inspector and DefaultMixedRealityBoundaryProfile.

Updated MixedRealityConfigurationProfile/Inspector and MixedRealityManager to support the boundary profile settings.

- Fixes: #2389, #2420


Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -11,45 +15,112 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Managers
...
-            // TODO Update stuff 
+            if (!Boundary.configured)
+            {
+                // The user has not configured a playspace boundary on this device.
+                return;
+            }
+
+            if (XRDevice.GetTrackingSpaceType() != TrackingSpaceType.RoomScale)
+            {
+                Debug.Log("Boundaries are supported for Room Scale experiences only.");
```
StephenHodgson: Do we wanna spam people every time this happens?

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -11,45 +15,112 @@ namespace Microsoft.MixedReality.Toolkit.Internal.Managers
...
+        /// Not all platforms support specifying whether or not to render the playspace boundary.
+        /// For platforms without boundary rendering control, the default behavior will be unchanged 
+        /// regardless of the value provided.
+        /// </remarks>
+        private void SetPlatformBoundaryVisibility()
         {
-            // TODO Destroy stuff 
+            if (Boundary.configured)
+            {
+#if !UNITY_WSA
```
StephenHodgson: We should probably avoid platform specific calls in our profiles if we can.

davidkline-ms: Agreed. Just not sure what the correct path is for platforms that lack specific features... Is it Unity's job our ours to handle these types of things.

SimonDarksideJ: If we are providing a boundary mechanism. Then we need to provide a facility to mock it, or at least not break if the platform doesn't provide it

SimonDarksideJ:
My only issues with having an entire profile for the PR are as follows:
* There are far too few settings to warrant an entire profile.  Maybe they would be better as additional options on the main profile (with the settings hidden if boundary disabled)
* Are we likely going to need multiple boundary profiles?  Surely this is a one time thing to setup and react to the runtime environment.

I just feel this may be over complicating the setup.

However the boundary implementation is a good start.

davidkline-ms:
>This is a good point. Do we need multiple boundary profiles?

My original plan was to add it to the Camera profile. I am perfectly fine folding it into the overall config profile.

SimonDarksideJ:
I think as there are only a few settings, adding it to the main MRConfigProfile is fine.   Just make sue the settings are hidden if boundary is disabled.

davidkline-ms:
Added DNM while I update based on feedback.

davidkline-ms:
This change is ready again. Thanks for your patience :)

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealityConfigurationProfileInspector.cs
```diff
@@ -69,7 +77,11 @@ private void OnEnable()
             enableControllerProfiles = serializedObject.FindProperty("enableControllerProfiles");
             controllersProfile = serializedObject.FindProperty("controllersProfile");
             enableBoundarySystem = serializedObject.FindProperty("enableBoundarySystem");
-            boundaryProfile = serializedObject.FindProperty("boundaryProfile");
+            boundaryExperienceScale = serializedObject.FindProperty("boundaryExperienceScale");
+            boundaryHeight = serializedObject.FindProperty("boundaryHeight");
+            enablePlatformBoundaryRendering = serializedObject.FindProperty("enablePlatformBoundaryRendering");
+
+            testThis = serializedObject.FindProperty("testThis");
```
StephenHodgson: What are we testing?

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealityConfigurationProfileInspector.cs
```diff
@@ -14,16 +15,24 @@ public class MixedRealityConfigurationProfileInspector : MixedRealityBaseConfigu
...
         private SerializedProperty enableInputSystem;
         private SerializedProperty inputSystemType;
         private SerializedProperty inputActionsProfile;
         private SerializedProperty enableSpeechCommands;
         private SerializedProperty speechCommandsProfile;
         private SerializedProperty enableControllerProfiles;
         private SerializedProperty controllersProfile;
+        // Boundary system properties
         private SerializedProperty enableBoundarySystem;
+        private SerializedProperty boundaryExperienceScale;
```
StephenHodgson: We should also add a system type for users so choose which boundary class to use. (Similar to the input system type choice.

davidkline-ms: To be sure I understand this feedback, adding this will let users select from alternative implementations (perhaps their own). Am I thinking right?

davidkline-ms:
>Looking good, although will need some Unit Tests for the boundary manager.

Yes. My next boundary PR will add a test scene and i will look at a good way to create unit tests that simulate actual boundaries.

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealityConfigurationProfileInspector.cs
SimonDarksideJ: Agreed, Like the InputSystem, the property in the Main MRTK Config Profile needs to use a Systemtype bound to the IMixedRealityBoundarySystem interface.  Then anyone can create their own boundary implementation based off out interface and use it.
Base principle of vNext with any Core component

SimonDarksideJ:
Added a new PR to your branch @davidkline-ms to add the requested changes :D  should be good now.

davidkline-ms:
Added Simon's changes. This should be ready for review again.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -180,8 +180,27 @@ public MixedRealityControllerMappingProfile ControllersProfile
         /// </summary>
         public bool EnableBoundarySystem
         {
-            get { return enableBoundarySystem; }
-            private set { enableBoundarySystem = value; }
+            get
+            {
+                return boundarySystemType != null &&
```
Jarodshow: Redundant first check?

SimonDarksideJ: Unity will let you select NONE as the SystemType, so we have to check if the user has selected something.
This is the same code as the InputSystem config uses.

SimonDarksideJ: Thoughts @StephenHodgson ? As this was cloned from the Input System code

Jarodshow: Sorry, I meant the null conditional operator below it, `boundarySystemType` is being checked null even though the operator below does that check.

davidkline-ms: It looks like @Jarodshow is on to something:

Console.WriteLine("TestObject != null              : {0}", to != null);
Console.WriteLine("TestObject?.InnerObject != null : {0}", to?.InnerObject != null);

Results in:

TestObject != null              : False
TestObject?.InnerObject != null : False

SimonDarksideJ:
Struggling to understand this new "ExperienceScale" and what it's intention is?
How is this different to the TrackingType, and why the change?

Also, why separate Seated / Standing, which imho, are the same thing in all SDK's.  Surely these should be one option (an experience that operates without a boundary)

Just what is this intended to be used for and how / why / where would a dev take advantage of it, when the SDK's don't.

SimonDarksideJ:
Ok, that explains a lot.   However, why do you need a separate PR to move the Experience scale?  When this PR is ADDING the boundary with the scale.
Surely #2420 should just be covered as part of this PR dealing with Boundary

davidkline-ms:
>However, why do you need a separate PR to move the Experience scale? When this PR is ADDING the boundary with the scale.
Surely #2420 should just be covered as part of this PR dealing with Boundary

I was trying to not "bloat" this one :) I'll make the change today.

SimonDarksideJ:
Thanks. 
And it doesn't bloat it, if it's about introducing the thing :D

Issue #2067, opened 2018-05-10
0 frame animation warnings from default cursor


davidkline-ms:
## Steps to reproduce
- Create a new project
- Import the toolkit
- Apply scene settings (add input mananger and cursor)
- Open the Console window and observe warnings re:
    - Cursor_cross_root
    - CursorRing

## Unity Editor Version
i've seen this with 2017.2.1p2, 2017.4.2f2

## Mixed Reality Toolkit Release Version
2017.2.*

StephenHodgson:
Which cursor are you using?

wiwei:
Closing issues older than 180 days. If this is still an issue, please reactivate with recent information.

PR #2414, opened 2018-07-06
Ported default cursor into SDK to test cursor visibility & functionality 

StephenHodgson:
Fixed a few issues with Focus Provider.

- Wasn't updating the default gaze transform on start before pointer was created which broke stuff.
- Added a few more assertions for easier debugging.

blocked by #2412


SimonDarksideJ:
I take it were also waiting on an actual script in the SDK to present the cursor. I don't see one.
Note, the MRTK shouldn't render the cursor, the SDK component for visualizing the cursor should.

davidkline-ms:
Do these animations have the same 0 frame length issue that master has? #2067. If so, we should fix or remove before merging

SimonDarksideJ:
This PR is still showing changes from the previous PR?  does the branch need refreshing?

StephenHodgson:
> I take it were also waiting on an actual script in the SDK to present the cursor. I don't see one.
Note, the MRTK shouldn't render the cursor, the SDK component for visualizing the cursor should.

It's been in there for a while.  I suppose you're asking to move the cursor implementation from the core to the SDK layer.

StephenHodgson:
>This PR is still showing changes from the previous PR? does the branch need refreshing?

Yeah I haven't merged in the changes from the dev branch yet. I currently don't have internet so that might have to wait until tomorrow.

SimonDarksideJ:
That's correct @StephenHodgson , the SDK should be responsible for rendering the cursor, utilising the InputSystem.  Not Core.

StephenHodgson:
> Do these animations have the same 0 frame length issue that master has?

I didn't get any warnings when I imported it.

Take a pull and find out.

StephenHodgson:
> the SDK should be responsible for rendering the cursor

The renderer component is rendering it 😉.

You mean to say that the cursor class that determines the cursor implementation should be in the SDK, correct?

SimonDarksideJ:
> You mean to say that the cursor class that determines the cursor implementation should be in the SDK, correct?

That's correct @StephenHodgson 
Helpers can be in core, but the actual vizualization and implementation should be in the SDK.

SimonDarksideJ:
On further review, I'm happy with the direction this SDK prefab has taken.
Sent a PR to @StephenHodgson to remove the #IF WSA checks in cursor (as it's related)

PR #2424, opened 2018-07-10
Shader updates from Dev

keveleigh:
Overview
---
From original PR:
>Integrating the latest work to improve the MRTK/Standard shader. Improvements include:
>
>- Support for non-D3D11 APIs (tested on iOS and Android devices using OpenGL ES and Metal)
>- Support for three hover lights (useful with IHMDs with multi-pointers)
>- Improved rounded corner algorithm
>- Closer match to the Unity/Standard shader
>- Improved shader based anti-aliasing support
>- Minor optimizations
>
>
>aa
>(Shader based anti-aliasing example.)

Changes
---
- Migrates over #2129.
- Part of #2423.

PR #2428, opened 2018-07-10
Port shader updates from Dev_Working_Branch and June

keveleigh:
Overview
---
Brings over the partial branching removal from June and general improvements from Dev/June.

Changes
---
- Migrates #2363
- Migrates #2424

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Resources/Fonts/selawk.ttf.meta
+++ b/Assets/MixedRealityToolkit/_Core/Resources/Fonts/selawk.ttf.meta
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 8d7ea1fb71806ef478d2e1a118298e31
+guid: 86574c70442309b45be5a1c37a37a40b
 TrueTypeFontImporter:
   externalObjects: {}
   serializedVersion: 4
@@ -12,8 +12,8 @@ TrueTypeFontImporter:
   fontNames:
   - Selawik
   fallbackFontReferences:
-  - {fileID: 12800000, guid: 55651ae5ac8b2804f94b620dd47797c3, type: 3}
-  - {fileID: 12800000, guid: 03d8727af681c3342a00825e0d19c780, type: 3}
+  - {fileID: 12800000, guid: 2a056e2bb89e0134daaf49e5f183e5dc, type: 3}
+  - {fileID: 12800000, guid: e3d8348b7d66bae4aa4b1f3ada3ef5fd, type: 3}
   customCharacters: 
   fontRenderingMode: 0
   ascentCalculationMode: 1
--- a/Assets/MixedRealityToolkit/_Core/Resources/Fonts/selawkl.ttf.meta
+++ b/Assets/MixedRealityToolkit/_Core/Resources/Fonts/selawkl.ttf.meta
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: dc91258b82040ef4294056d89d500b9f
+guid: b8a8324793a77d14fad6ef1dc06c55ed
 TrueTypeFontImporter:
   externalObjects: {}
   serializedVersion: 4
@@ -12,10 +12,10 @@ TrueTypeFontImporter:
   fontNames:
   - Selawik
   fallbackFontReferences:
-  - {fileID: 12800000, guid: 55651ae5ac8b2804f94b620dd47797c3, type: 3}
-  - {fileID: 12800000, guid: c54a83c68f4a6b04283d122671f79647, type: 3}
-  - {fileID: 12800000, guid: 03d8727af681c3342a00825e0d19c780, type: 3}
-  - {fileID: 12800000, guid: 8d7ea1fb71806ef478d2e1a118298e31, type: 3}
+  - {fileID: 12800000, guid: 2a056e2bb89e0134daaf49e5f183e5dc, type: 3}
+  - {fileID: 12800000, guid: 98d9db324dda3a44e93fbf3b7ce9eca9, type: 3}
+  - {fileID: 12800000, guid: e3d8348b7d66bae4aa4b1f3ada3ef5fd, type: 3}
+  - {fileID: 12800000, guid: 86574c70442309b45be5a1c37a37a40b, type: 3}
   customCharacters: 
   fontRenderingMode: 0
   ascentCalculationMode: 1
--- a/Assets/MixedRealityToolkit/_Core/Utilities/UX/Billboard.cs.meta
+++ b/Assets/MixedRealityToolkit/_Core/Utilities/UX/Billboard.cs.meta
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 606b6837de4801e40886220804b02c54
+guid: ac8d5b128a1d8204fb76c86f47b75912
 MonoImporter:
   externalObjects: {}
   serializedVersion: 2
--- a/Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
+++ b/Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Collections.Generic;
 using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Internal.Utilities.UX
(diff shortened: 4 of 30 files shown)


Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
```diff
@@ -92,21 +100,87 @@ private void OnDrawGizmosSelected()
...
+                }
+            }
+
+            Shader.SetGlobalVectorArray(_HoverLightDataID, hoverLightData);
+
+            lastHoverLightUpdate = Time.frameCount;
         }
+    }
+
+#if UNITY_EDITOR
```
SimonDarksideJ: Editors cannot exist inside a runtime script. They must be broken out in to their own files in the inspectors folder.
"/Assets/MixedRealityToolkit/_Core/Inspectors "

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
```diff
@@ -92,21 +100,87 @@ private void OnDrawGizmosSelected()
             Gizmos.DrawIcon(transform.position + Vector3.back * Radius, string.Empty, false);
         }
 
-        private void Initialize()
+        private static void AddHoverLight(HoverLight light)
```
SimonDarksideJ: Is there a reason why we are introducing so many static functions?
If there is need for a single instance of the script, then it should probably be a runtime component.

Can we clarify why these need to be static @Cameron-Micka ?

Cameron-Micka: None of these "need" to be static. I just like to mark them static to hint to other programmers that these functions will likely only operate on what is passed in, or other static members. It's also probably a premature optimization haha.

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
Cameron-Micka: I can certainly do this but, I probably won't have time until next week.

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
SimonDarksideJ: Ok, can we change these back to non-static and we can review as required @keveleigh

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
SimonDarksideJ: It's ok @Cameron-Micka , it's for the vNext PR only.  @keveleigh , I'll send you a PR with the updates. (benefits of being ahead of time :D)

SimonDarksideJ:
Submitted PR to @keveleigh PR branch with vNext standards updates
https://github.com/keveleigh/HoloToolkit-Unity/pull/4

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
StephenHodgson: O think it makes sense that they're statics seeing as we only have three lights to work with at any time.

SimonDarksideJ: Already removed the statics in the PR @StephenHodgson, also note these are private statics.
Lets review once this starts getting used, unless you feel the statics should remain?

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
```diff
@@ -19,9 +20,13 @@ public class HoverLight : MonoBehaviour
...
-        private int hoverColorId;
+        // Three hover lights are supported at this time.
+        private const int hoverLightCount = 3;
+        private const int hoverLightDataSize = 2;
+        private static List<HoverLight> activeHoverLights = new List<HoverLight>(hoverLightCount);
+        private static Vector4[] hoverLightData = new Vector4[hoverLightCount * hoverLightDataSize];
+        private static int _HoverLightDataID;
+        private static int lastHoverLightUpdate = -1;
 
         public float Radius
```
SimonDarksideJ: Change to c# 6 auto property :D (example in the old PR)
```public float Radius => radius;```

Also change ordering of fields, Public and then private
Same changes for Color property
(Note, you'll need to change the use of the public properties inside the script to use the private ones)

File: Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs (hunk 1 above)
Comment (SimonDarksideJ) [reply Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs]: As PR cancelled, ensure to move the Editor code to it's own class

File: Assets/MixedRealityToolkit/_Core/Utilities/UX/Billboard.cs.meta [hunk 4]
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 606b6837de4801e40886220804b02c54
+guid: ac8d5b128a1d8204fb76c86f47b75912
```
SimonDarksideJ: Also ensure to take in the changes to the BillBoard script

Assets/MixedRealityToolkit/_Core/Resources/Fonts/selawk.ttf.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 8d7ea1fb71806ef478d2e1a118298e31
+guid: 86574c70442309b45be5a1c37a37a40b
```
StephenHodgson: Which meta file is correct?

Assets/MixedRealityToolkit/_Core/Resources/Fonts/selawkl.ttf.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: dc91258b82040ef4294056d89d500b9f
+guid: b8a8324793a77d14fad6ef1dc06c55ed
```
StephenHodgson: Why did the meta files for all these fonts change?

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
```diff
@@ -51,28 +56,31 @@ public Color Color
...
         }
 
         private void OnDisable()
         {
-            UpdateHoverLight();
+            RemoveHoverLight(this);
+            UpdateHoverLights(true);
         }
 
+#if UNITY_EDITOR
```
StephenHodgson: Why did this get reverted?

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
```diff
@@ -51,28 +56,31 @@ public Color Color
...
         private void Update()
         {
-            if (Application.isPlaying || !Application.isEditor)
```
StephenHodgson: Why did this get reverted?

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
```diff
@@ -92,21 +100,87 @@ private void OnDrawGizmosSelected()
...
+            }
+
+            if (!forceUpdate && (Time.frameCount == lastHoverLightUpdate))
+            {
+                return;
+            }
+
+            if (activeHoverLights.Count > 1)
+            {
+                Shader.EnableKeyword("_MULTI_HOVER_LIGHT");
```
StephenHodgson: Let's make this a const string seeing as we use it in multiple places.

We could do the same of for all the keywords.

davidkline-ms:
Retargeted this PR to mrtkDevelopment branch.

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
davidkline-ms: if we are using it externally, let's make it a static readonly string so that the reference is fetched and not the literal constant.

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
davidkline-ms: This is based on the changes that went into Dev_Working_Branch and back into june18_dev. Part of the gap analysis / keeping things in sync. If this change is incorrect, we should definitely fix it here and in june18_dev.

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
Cameron-Micka: This string is only used by this file at the moment!

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
StephenHodgson: Please remove compiler directive in favor of checking for `Application.isEditor`

Assets/MixedRealityToolkit/_Core/Utilities/UX/HoverLight.cs
SimonDarksideJ: Personally, considering this is ONLY required in the editor, a compiler directive is better than an if, ensures it is not run and no performance check when running in prod

PR #2442, opened 2018-07-13
Fixed a couple of input system bugs

StephenHodgson:
- Pose was executing the wrong event data handler.
- Pointer event data needed to call Initialize overload instead of base initialize.

_It's the little things that matter._


StephenHodgson:
> Are we sure these are all of them?

Idk, but I tested the repros you gave me and it was fixed.
Just let me know if you run into something else and we'll fix it.  I only have WMR controllers to test with.

PR #2445, opened 2018-07-15
Ported a collection of MRTK SDK assets 

SimonDarksideJ:
Overview
---
Pulled over a collection of useful scene assets from HoloToolkit in to the MRTK SDK.  In preparation for creating example scenes


SimonDarksideJ:
These are not examples, but core assets to be available through the SDK.

We should not have just "example" assets for the sake of them. The examples should be teaching exercises, like the Academy.

I will cull pack this PR to focus on it's original intent of the SDK requirements for Bounding Box / Button and Sphere assets (and anything else generically useful)

davidkline-ms:
@SimonDarksideJ, is this PR still not to be merged? If so, will it be or is it being superceded by another PR?

SimonDarksideJ:
Will redo for Alpha

PR #2449, opened 2018-07-16
Some Input System Feature SDK components.

StephenHodgson:
Overview
---
- Added input system global listener helper component to SDK.
- Added Pointer Click Handler component to SDK.
- Updated Speech Input Handler SDK component.
- Updated FocusDetails to use properties.
- Updated MixedRealityInputActions to use equality and == operator for easier comparison.
- Updated Quality Settings from Unity Editor update

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/BaseInputHandler.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/BaseInputHandler.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Input
+{
+    public class BaseInputHandler : InputSystemGlobalListener
+    {
+        [SerializeField]
+        [Tooltip("Is Focus required to receive input events on this GameObject?")]
+        private bool isFocusRequired = false;
+
+        /// <summary>
+        /// Is Focus required to receive input events on this GameObject?
+        /// </summary>
+        public bool IsFocusRequired
+        {
+            get { return isFocusRequired; }
+            set { isFocusRequired = value; }
+        }
+
+        #region Monobehaviour Implementation
+
+        protected override void OnEnable()
+        {
+            if (!isFocusRequired)
+            {
+                base.OnEnable();
+            }
+        }
+
+        protected override void OnDisable()
+        {
+            if (!isFocusRequired)
+            {
+                base.OnDisable();
+            }
+        }
+
+        #endregion Monobehaviour Implementation
+    }
+}
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
@@ -13,14 +13,14 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
     public class InputSystemGlobalListener : MonoBehaviour
     {
         private IMixedRealityInputSystem inputSystem = null;
-        private IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
+        protected IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
 
-        private void OnEnable()
+        protected virtual void OnEnable()
         {
             InputSystem.Register(gameObject);
         }
 
-        private void OnDisable()
+        protected virtual void OnDisable()
         {
             InputSystem.Unregister(gameObject);
         }
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+using Microsoft.MixedReality.Toolkit.Internal.Managers;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Input
+{
+    /// <summary>
+    /// This component ensures that all input events are forwarded to this <see cref="GameObject"/> when focus or gaze is not required.
+    /// </summary>
+    public class InputSystemGlobalListener : MonoBehaviour
+    {
+        private IMixedRealityInputSystem inputSystem = null;
+        private IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
+
+        private void OnEnable()
+        {
+            InputSystem.Register(gameObject);
+        }
+
+        private void OnDisable()
+        {
+            InputSystem.Unregister(gameObject);
+        }
+    }
+}
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
@@ -16,48 +16,48 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
     {
         [SerializeField]
         [Tooltip("The input actions to be recognized on pointer up.")]
-        private InputActionEventPair onPointerUpActionEvents;
+        private InputActionEventPair onPointerUpActionEvent;
 
         [SerializeField]
         [Tooltip("The input actions to be recognized on pointer down.")]
-        private InputActionEventPair onPointerDownActionEvents;
+        private InputActionEventPair onPointerDownActionEvent;
 
         [SerializeField]
         [Tooltip("The input actions to be recognized on pointer clicked.")]
-        private InputActionEventPair onPointerClickedActionEvents;
+        private InputActionEventPair onPointerClickedActionEvent;
 
         #region IMixedRealityPointerHandler Implementation
 
         /// <inheritdoc />
         public void OnPointerUp(MixedRealityPointerEventData eventData)
         {
-            if (onPointerUpActionEvents.InputAction == MixedRealityInputAction.None) { return; }
+            if (onPointerUpActionEvent.InputAction == MixedRealityInputAction.None) { return; }
 
-            if (onPointerUpActionEvents.InputAction == eventData.MixedRealityInputAction)
+            if (onPointerUpActionEvent.InputAction == eventData.MixedRealityInputAction)
             {
-                onPointerUpActionEvents.UnityEvent.Invoke();
+                onPointerUpActionEvent.UnityEvent.Invoke();
             }
         }
 
         /// <inheritdoc />
         public void OnPointerDown(MixedRealityPointerEventData eventData)
         {
-            if (onPointerDownActionEvents.InputAction == MixedRealityInputAction.None) { return; }
+            if (onPointerDownActionEvent.InputAction == MixedRealityInputAction.None) { return; }
 
-            if (onPointerDownActionEvents.InputAction == eventData.MixedRealityInputAction)
+            if (onPointerDownActionEvent.InputAction == eventData.MixedRealityInputAction)
             {
-                onPointerDownActionEvents.UnityEvent.Invoke();
+                onPointerDownActionEvent.UnityEvent.Invoke();
             }
         }
 
         /// <inheritdoc />
         public void OnPointerClicked(MixedRealityPointerEventData eventData)
         {
-            if (onPointerClickedActionEvents.InputAction == MixedRealityInputAction.None) { return; }
+            if (onPointerClickedActionEvent.InputAction == MixedRealityInputAction.None) { return; }
 
-            if (onPointerClickedActionEvents.InputAction == eventData.MixedRealityInputAction)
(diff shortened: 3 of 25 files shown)


SimonDarksideJ:
If you are touching the MixedRealityInputActions , can you test / fix the issue where you can't select the last item, and the index lookup seems out of sync by 1?  Possible index array issue.

Replication stops:
* Select the ControllerInputActions profile
* Change the Input Action to the last item
* Observer index error

* Raise an event using an interaction mapping
* Check receiving event action
* Notice it is the previous InputAction.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Input
+{
+    /// <summary>
+    /// This component ensures that all input events are forwarded to this <see cref="GameObject"/> when focus or gaze is not required.
+    /// </summary>
+    public class InputSystemGlobalListener : MonoBehaviour
+    {
+        private IMixedRealityInputSystem inputSystem = null;
+        private IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
```
SimonDarksideJ: Maybe we should put this code in a generic base class?  As we seem to be copying it everywhere.

StephenHodgson:
Will do, thanks for the repro steps

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
StephenHodgson: Maybe, it's just the easiest way to get the input system reference.

StephenHodgson: I'm not sure making a base input class would be good for the SDK components. I just wanted then to be simple and self contained.

ProjectSettings/QualitySettings.asset
```diff
@@ -29,6 +29,12 @@ QualitySettings:
     vSyncCount: 0
     lodBias: 0.3
     maximumLODLevel: 0
+    streamingMipmapsActive: 0
```
SimonDarksideJ: Best to revert these .asset changes as they aren't part of this PR

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
```diff
@@ -0,0 +1,95 @@
...
+        [Tooltip("The input actions to be recognized on pointer up.")]
+        private InputActionEventPair onPointerUpActionEvents;
+
+        [SerializeField]
+        [Tooltip("The input actions to be recognized on pointer down.")]
+        private InputActionEventPair onPointerDownActionEvents;
+
+        [SerializeField]
+        [Tooltip("The input actions to be recognized on pointer clicked.")]
+        private InputActionEventPair onPointerClickedActionEvents;
```
SimonDarksideJ: I'm guessing this will need it's own property drawer to render properly in the inspector.  Not entirely sure I like this approach, but let's see where it leads us.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
SimonDarksideJ: Nothings more self contained that a MonoBehavjour base class which implements the same code for all dependant classes.
Also make it a heck of a lot easier for people to create their own classes as there is less to copy.  They just use the new base class instead of MonoBehaviour

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
StephenHodgson: It doesn't need a property drawer

ProjectSettings/QualitySettings.asset
StephenHodgson: I'll add this in it's own or

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
StephenHodgson: BaseInputHandler it is then.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
SimonDarksideJ: So you only expect this to be accessible via code?  Or do you have a screenshot of how this field looks like in the editor?
It's not simpler if the user can't use the editor too.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/BaseInputHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            get { return isFocusRequired; }
+            set { isFocusRequired = value; }
+        }
+
+        #region Monobehaviour Implementation
+
+        protected override void OnEnable()
+        {
+            if (!isFocusRequired)
```
davidkline-ms: I originally got confused as to why nothing is done when focus IS required (until I read the doc comments for InputSystemGlobalListener). Perhaps a quick comment here would help clarify for others.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// This component ensures that all input events are forwarded to this <see cref="GameObject"/> when focus or gaze is not required.
+    /// </summary>
+    public class InputSystemGlobalListener : MonoBehaviour
+    {
+        private IMixedRealityInputSystem inputSystem = null;
+        protected IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
+
+        protected virtual void OnEnable()
```
davidkline-ms: doc comments

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/BaseInputHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+        /// </summary>
+        public bool IsFocusRequired
+        {
+            get { return isFocusRequired; }
+            set { isFocusRequired = value; }
+        }
+
+        #region Monobehaviour Implementation
+
+        protected override void OnEnable()
```
davidkline-ms: doc comments

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    /// <summary>
+    /// This component handles pointer clicks from all types of input sources.<para/>
+    /// i.e. a primary mouse button click, motion controller selection press, or hand tap.
+    /// </summary>
+    public class PointerClickHandler : BaseInputHandler, IMixedRealityPointerHandler
+    {
+        [SerializeField]
+        [Tooltip("The input actions to be recognized on pointer up.")]
+        private InputActionEventPair onPointerUpActionEvents;
```
davidkline-ms: This is described as action eventS, yet there is only one pair defined... do we *really* want to support registering multiple events on a single action or should this be a collection

(same with the other two IAEP fields)

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
davidkline-ms: @SimonDarksideJ, these look to be ONLY accessible via the inspector.
@StephenHodgson, should these be accessible via script?

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+        [SerializeField]
+        [Tooltip("The input actions to be recognized on pointer clicked.")]
+        private InputActionEventPair onPointerClickedActionEvents;
+
+        #region IMixedRealityPointerHandler Implementation
+
+        /// <inheritdoc />
+        public void OnPointerUp(MixedRealityPointerEventData eventData)
+        {
+            if (onPointerUpActionEvents.InputAction == MixedRealityInputAction.None) { return; }
```
davidkline-ms: this implies that there IS only one event per action... onPointerUpActionEvents should be singular

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/SpeechInputHandler.cs
```diff
@@ -24,37 +25,19 @@ public class SpeechInputHandler : MonoBehaviour, IMixedRealitySpeechHandler
         [Tooltip("The keywords to be recognized and optional keyboard shortcuts.")]
         private KeywordAndResponse[] keywords = new KeywordAndResponse[0];
```
davidkline-ms: I know this is not new to this PR, but do we really NOT want folks to be able to add keywords via script? Also, why an array instead of a List<>?

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/SpeechInputHandler.cs
```diff
@@ -24,37 +25,19 @@ public class SpeechInputHandler : MonoBehaviour, IMixedRealitySpeechHandler
...
 
-        [SerializeField]
-        [Tooltip("Is Gaze required for the keyword to raise the action?")]
-        private bool isGazeRequired = false;
-
         [SerializeField]
         [Tooltip("Keywords are persistent across all scenes.  This Speech Input Handler instance will not be destroyed when loading a new scene.")]
```
davidkline-ms: Should be changable via script.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/BaseInputHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            if (!isFocusRequired)
+            {
+                base.OnEnable();
+            }
+        }
+
+        protected override void OnDisable()
+        {
+            if (!isFocusRequired)
```
davidkline-ms: Shouldnt there be handlers for focus enter/exit?

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputAction.cs
```diff
@@ -51,5 +52,48 @@ public MixedRealityInputAction(uint id, string description, AxisType axisConstra
...
+        }
+
+        int IEqualityComparer.GetHashCode(object obj)
+        {
+            return obj is MixedRealityInputAction ? ((MixedRealityInputAction)obj).GetHashCode() : 0;
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
davidkline-ms: Let's do a custom GetHashCode implementation...

Ex:
return string.Format("{0}.{1}", Id, AxisConstraint).GetHashCode();

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
StephenHodgson: The opposite actually.  The events will be accessible and setup in the editor only.

image

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/BaseInputHandler.cs
StephenHodgson: Why? It's a Unity Monobehavior Method

davidkline-ms: duh... sorry

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
davidkline-ms: was reading quickly.... sorry

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/BaseInputHandler.cs
SimonDarksideJ: Users should be referring to our training docs for using this handler, Code documentation is only a fall back for advanced users.  I don't think it's needed in this specific case

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
SimonDarksideJ: @davidkline-ms We don't (and likely never will) document the Unity MonoBehaviour events.
Only standard we would impose was to ensure Unity methods are wrapped in a Region called

```#region MonoBehaviour Implementation```

As we have done in other areas of the codebase.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputAction.cs
SimonDarksideJ: Is there a particular reason for that @davidkline-ms ?

in any case, we have may implementations with hashes so far, so if this is required, there should be a separate task to discuss and agree the way forward before changing the current implementations.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/PointerClickHandler.cs
StephenHodgson: >  should these be accessible via script?

That is correct

PR #2455, opened 2018-07-18
SDK Asset Organization.

StephenHodgson:
- Moved a good amount of things to the SDK and organized cursor assets.
- Fixed BaseFocusHandler to require focus by default.
- Updated documentation.
- Updated basic focus handling.
- Rolled back Editor to 2018.1.8f1
- Refactored usage of getting Input System references into base classes.


Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/MixedRealityInputManager.cs
```diff
@@ -19,7 +17,7 @@
 using UnityEngine;
 using UnityEngine.EventSystems;
 
-namespace Microsoft.MixedReality.Toolkit.InputSystem
+namespace Microsoft.MixedReality.Toolkit.SDK.Input
```
SimonDarksideJ: Not sure the Input manager should move to the SDK, as it is one of our core features.
Granted you currently have dependencies on Focus / Gaze, which moved to core.  but this should mean refactoring the implementation rather than moving it.

Assets/MixedRealityToolkit-SDK/Features/UX/HoverLight.cs
```diff
@@ -4,7 +4,7 @@
 using System.Collections.Generic;
 using UnityEngine;
 
-namespace Microsoft.MixedReality.Toolkit.Internal.Utilities.UX
+namespace Microsoft.MixedReality.Toolkit.SDK.UX
```
SimonDarksideJ: Need to replicate this namespace update for the Billboard class.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/MixedRealityInputManager.cs
StephenHodgson: I think it makes sense after our conversation during the ship room meeting.

StephenHodgson: It's our specific implementation.

SimonDarksideJ: OK, I'll let it go for now.  But that would ring true for ANY component we build?  Making the Core just definitions, interfaces and the MRManager. (meaning out of the box, the core does nothing but start an empty scene)

davidkline-ms: >  (meaning out of the box, the core does nothing but start an empty scene)

I'm with @SimonDarksideJ on this one... I believe we will want to move our implementations back into the core.

davidkline-ms: There are very likely going to be devs that want our base implementations and *nothing else* from the SDK packages.

StephenHodgson: >There are very likely going to be devs that want our base implementations and nothing else from the SDK packages.

I can see this being the case for a few, but at that point they might as well be writing their own implementation and can just cherry pick what they want via the SDK.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
```diff
@@ -16,20 +15,14 @@ public class InputSystemGlobalListener : MonoBehaviour
         private static IMixedRealityInputSystem inputSystem = null;
         protected static IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
 
-        private readonly WaitUntil waitForInputSystem = new WaitUntil(() => MixedRealityManager.IsInitialized);
-
-        protected virtual async void OnEnable()
+        protected virtual void OnEnable()
         {
-            await waitForInputSystem;
```
SimonDarksideJ: I take it you found another way so this didn't look for the managers before the MRManager was fully initialised?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
```diff
@@ -16,6 +17,9 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Cursors
     /// </summary>
     public class BaseCursor : MonoBehaviour, IMixedRealityCursor
     {
+        private static IMixedRealityInputSystem inputSystem = null;
```
SimonDarksideJ: Why not use one of the new base classes?

StephenHodgson: This -is_ the base class 😅

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/InputSystemGlobalListener.cs
StephenHodgson: That is correct

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/MixedRealityInputManager.cs
davidkline-ms: except that the sdk packages will likely contain much more than they want

davidkline-ms: let's keep the default managers (input, boundary, etc) in the core and everything else in an sdk package

StephenHodgson: So then should we also move the default profiles as well?

davidkline-ms: that sounds like a great idea.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
SimonDarksideJ: lol, but you know what I mean.  Like InputSystemGlobalListener, which the exact same code as used here.

Assets/MixedRealityToolkit-SDK/Features/Input/Scripts/MixedRealityInputManager.cs
SimonDarksideJ: I say our "default profiles" are a different thing entirely, as they are "quick start" things.

They are not crucial to someone downloading the Core project to get running as they can quickly create their own.  

The profiles in the SDK are aimed at making the SDK easy to pick up and get running only.

davidkline-ms: @SimonDarksideJ, that makes complete sense.

StephenHodgson: I'm inclined to keep the manager implenetations and the profiles together.

I don't see them as mutually exclusive things.

davidkline-ms: @StephenHodgson, as does your comment...

I am leaning more and more to moving the default configuration files to _Core to live beside the default managers.

SimonDarksideJ: I'm losing my hair.  Our "default" configuration profiles are merely examples. Even in the SDk, users are going to customise them and create their own.  Our "defaults" are the settings we configure in the scripable objects in Core,  so there isn't any need for the .asset file to live in core.

So scriptable config definitions are in core (as they are now), the "example" profiles should live in the SDK as defaults.  One could even argue they don't even need to exist at all, but they are good defaults.

davidkline-ms: How about we table this PR until after Alpha and then spend the appropriate amount of design / discussion time to ensure we get it right for Beta.

SimonDarksideJ: I think the PR as a whole is fine.  We can always shuffle again later before we hit the Alpha button
Trying to focus on the visualizer and I'm also adding the Rift config now (as that's what is plugged in), also sorted out multi-select controls now for OpenVR.

Going to need to do another "controllers" refactor to simplify the code, later.

davidkline-ms: That makes sense. Let's get this in and shuffle next week (or right after Alpha). We are a monolithic package for this release anyhow :)

StephenHodgson:
Okay Simon. I refactored a few more classes to inherit from a base with the input system in a single place.  Couldn't do it everywhere, but got the easiest ones first.

StephenHodgson:
This is ready for another review

Assets/MixedRealityToolkit/InputSystem/Pointers/GenericPointer.cs
```diff
@@ -38,7 +39,13 @@ public GenericPointer(string pointerName, IMixedRealityInputSource inputSourcePa
         public string PointerName { get; set; }
 
         /// <inheritdoc />
-        public virtual IMixedRealityInputSource InputSourceParent { get; protected set; }
+        public virtual IMixedRealityInputSource InputSourceParent
```
keveleigh: Is there a benefit to not using the auto-property here?

StephenHodgson: Not sure how that got changed. I'll be sure to look at this real quick. Probably a mistake.

StephenHodgson: Yes, it's because it's called in the constructor and because it's virtual can cause issues.

StephenHodgson: https://www.jetbrains.com/help/resharper/2018.1/VirtualMemberCallInConstructor.html

StephenHodgson:
I want to merge the devices refactor before this

davidkline-ms:
> I want to merge the devices refactor before this

Sounds good.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
```diff
@@ -82,11 +83,21 @@ public class BaseCursor : MonoBehaviour, IMixedRealityCursor
...
 
-        #region IMixedRealityCursor Implementation
-
         /// <summary>
-        /// The pointer that this cursor should follow and process input from.
+        /// Indicates if the cursor should be visible
         /// </summary>
+        public bool IsVisible
+        {
+            set
```
davidkline-ms: Should have a getter also...

davidkline-ms: to me, a set only property is bad form

StephenHodgson: Isn't this the same in master?
But yea I agree we should be able to get as well.

davidkline-ms: to be honest, I didn't check :)

if we want to take this and file an issue to fix, I would be ok with that.

PR #2457, opened 2018-07-18
Billboard camera tolerance

PJBowron:
Overview
---
Two small changes to help Billboard cope with the lack of, or late setup of, a cached camera instance. 

Changes
---
1. null check in OnEnable
2. per-frame attempt to obtain the main camera transfer within the existing null-transform test

### DIFF
--- a/Assets/HoloToolkit/Utilities/Scripts/Billboard.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/Billboard.cs
@@ -53,7 +53,7 @@ namespace HoloToolkit.Unity
         {
             if (TargetTransform == null)
             {
-                TargetTransform = CameraCache.Main.transform;
+                TargetTransform = CameraCache.Main?.transform;
             }
         }
 
@@ -64,6 +64,7 @@ namespace HoloToolkit.Unity
         {
             if (TargetTransform == null)
             {
+                TargetTransform = CameraCache.Main?.transform;
                 return;
             }
 
--- a/Assets/HoloToolkit/Utilities/Scripts/Billboard.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/Billboard.cs
@@ -53,19 +53,28 @@ namespace HoloToolkit.Unity
         {
             if (TargetTransform == null)
             {
-                TargetTransform = CameraCache.Main?.transform;
+                if (CameraCache.Main != null)
+                {
+                    TargetTransform = CameraCache.Main.transform;
+                }
             }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


PJBowron:
In my current use case, my project has a scene manager controlling various scenes that are loaded and dropped at runtime. Under normal circumstances all the core elements, including the camera, are brought in first. But during development I can run a subscene independently - the scene manager loads the core elements but the scene content gets spun up first: currently, Billboards break due to the late camera init.

davidkline-ms:
> during development I can run a subscene independently - the scene manager loads the core elements but the scene content gets spun up first

I can see that being an issue. Is this something that should be done inside an #if UNITY_EDITOR block?

PJBowron:
Could be, but I don't really see why someone shouldn't be allowed to set up their camera after their billboard at runtime if they want to :)

Assets/HoloToolkit/Utilities/Scripts/Billboard.cs
```diff
@@ -53,7 +53,7 @@ private void OnEnable()
         {
             if (TargetTransform == null)
             {
-                TargetTransform = CameraCache.Main.transform;
+                TargetTransform = CameraCache.Main?.transform;
```
keveleigh: image

The master project isn't compatible with this syntax.

davidkline-ms: I'm not sure ?. is available with the legacy C# compiler (supporting .net 3.5). please change this to a traditional null check (in your Update() change as well).

thx!

davidkline-ms: haha! you beat me to it!

PJBowron:
Cool; getting late so I'll fix this up tomorrow. Thanks all!

PJBowron:
Sure fair enough, a billboard requires a camera, but that doesn't help someone like me who has to load their camera after the billboard initializes - should I make my own billboard-alike for every new project? 😄

PJBowron:
One thing noted in my commit but wanted to highlight: in testing earlier today I changed the calculations to run in LateUpdate rather than Update - it fixes a one-frame lag particularly noticeable on objects close and tightly lerped or - shudder - locked to the camera gaze. Thought I'd call it out in case there was a particular reason this hadn't already been done.

StephenHodgson:
Let's be sure to port this to vNEXT

davidkline-ms:
> Let's be sure to port this to vNEXT

Added to the tracking issue (#2465)

PR #2459, opened 2018-07-18
Fix bug with enableOneHandMovement in TwoHandManipulatable

### LINKED ISSUE #2021: TwoHandManipulatable script moves objects even if OneHandMovement option is unchecked
## Overview
I hope this is not a duplicate. I checked and it seems that nobody has reported this behaviour yet.
As you can get from the title, there is a bug in TwoHandManipulatable script. It's possible to move objects using one hand even if OneHandMovement option is unchecked.

## Expected Behavior
While manipulating objects in one of TwoHand state (Scaling, Rotating, MovingScaling, RotatingScaling, MovingRotatingScaling), if I lose track of one of my hands I expect the action (move, scale, rotate) that I was doing to stop.

## Actual Behavior
However, if I lose track of one of my hands the script automatically enters into the Moving state, allowing me to move the object I was manipulating with one hand.

## How I solved it
Basically, I sorted out the problem by changing line 290 of TwoHandManipulatable script.
From:
newState = State.Moving;
To:
newState = State.Start;

Now everything seems to work as expected.

## Unity Editor Version
Unity 2017.2.0f3
2017.2.1.4 Patch Release

keveleigh:
Overview
---
Previously, a lost hand would revert to one-handed move even if one-handed move wasn't enabled. Now, it falls back to `None` in that case.

Changes
---
- Fixes #2021

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/TwoHandManipulatable.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Utilities/Interactions/TwoHandManipulatable.cs
@@ -237,9 +237,9 @@ namespace HoloToolkit.Unity.InputModule.Utilities.Interactions
                     {
                         newState = ManipulationMode.None;
                     }
-                    else if (handsPressedCount == 1 && enableOneHandMovement)
+                    else if (handsPressedCount == 1)
                     {
-                        newState = ManipulationMode.Move;
+                        newState = enableOneHandMovement ? ManipulationMode.Move : ManipulationMode.None;
                     }
                     else if (handsPressedCount > 1)
                     {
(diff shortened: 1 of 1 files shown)


davidkline-ms:
> We should probably port this in vNEXT.

Most definitely.  Adding to the list @keveleigh is keeping.

keveleigh:
Doesn't look like this script exists in vNext yet.

StephenHodgson:
It does

keveleigh:
I see the helper scripts, but not this one. Do you have a link?

StephenHodgson:
Looking now, one minute... I think you're right.  I only found the logic physics helpers.

PR #2466, opened 2018-07-19
Add gizmo models/prefabs, rename SDK\Resources -> SDK\StandardAssets

davidkline-ms:
Overview
---
Add fbx model of a coordinate gizmo with; X increasing to the left, Y increasing upward, Z increasing away from the origin.

Add prefab for gizmo that scales and replaces the built-in materials with MRTK_Standard materials.

Rename Resources folder to StandardAssets as that is the true purpose and since Unity treats Resources as a special folder (all contents built into projects so that they can be runtime loaded) it is advisable to not use that name and allow developers to control the *true* Resources folder.


SimonDarksideJ:
Definitely need the other hand for the Gizmo controller model to complete this PR.

I suspect we'll need subfolders in the prefabs to make finding things like controllers easier.  Especially when we have prefab for all the different SDK types.

In my visualizer PR, I add subfolders for each SDK type.

davidkline-ms:
>  Especially when we have prefab for all the different SDK types.

If I read this correctly, the StandardAssets package will contain optional assets for ALL sdk packages

StephenHodgson:
What do they look like?

davidkline-ms:
I'll post an image of the right hand model this afternoon.

StephenHodgson:
If it's just the arrows you posed in the Slack channel, why not just use the built in gizmos in the editor?

Assets/MixedRealityToolkit-SDK/StandardAssets/Models/GizmoRight.fbx.meta
```diff
@@ -0,0 +1,132 @@
...
+    400038: transform9
+    2100000: phong2
+    2100002: phong4
+    2100004: phong3
+    2300000: pCone6
+    3300000: pCone6
+    4300000: pCone6
+  externalObjects: {}
+  materials:
+    importMaterials: 1
```
StephenHodgson: let's turn off import materials and use the standard materials we've setup.

Assets/MixedRealityToolkit-SDK/StandardAssets/Models/GizmoRight.fbx.meta
```diff
@@ -0,0 +1,132 @@
...
+    isReadable: 1
+  meshes:
+    lODScreenPercentages: []
+    globalScale: 1
+    meshCompression: 0
+    addColliders: 0
+    importVisibility: 1
+    importBlendShapes: 1
+    importCameras: 1
+    importLights: 1
```
StephenHodgson: we probably don't wanna import camera's or lights either.

Assets/MixedRealityToolkit-SDK/StandardAssets/Models/GizmoRight.fbx.meta
davidkline-ms: I can do that

Assets/MixedRealityToolkit-SDK/StandardAssets/Models/GizmoRight.fbx.meta
davidkline-ms: sounds good

davidkline-ms:
> If it's just the arrows you posed in the Slack channel, why not just use the built in gizmos in the editor?

These were requested so that they could be used as "Controller" models.

davidkline-ms:
Left gizmo added. @StephenHodgson's meta file changes made. Ready to review. Thx

davidkline-ms:
>Would still prefer the Gizmo controller in a sub folder for better organisation, but that can be organised later when I add the other controller models.

That's what I figured :)

PR #2467, opened 2018-07-19
Add "Add the Spatial Mapping Prefab" to the scene configuration window

### LINKED ISSUE #2203: Proposal:We were wondering if you could add the checkbox:"Add the Spatial Mapping prefab" in "Apply Mixed Reality Scene Setting" dialog?
## Overview

"Apply Mixed Reality Project Setting"  has useful to add "Set Default Spatial Mapping Layer",but we might misunderstand that "Spatial Mapping" is become to available.
In actuality,we must set "SpatialMapping.prefab" to the scene, so we ware wonderring if you could add the checkbox:"Add the Spatial Mapping prefab" in "Apply Mixed Reality Scene Setting" dialog?

image

we can set both project and scene from configuration dialog,this should come in useful.

## Unity Editor Version
Unity 2017.4.3f1

MRTK 2017.4.0.0-rc1

keveleigh:
Overview
---
Adds an optional setting to the scene configuration window to quickly add the SpatialMapping prefab to the scene.

Changes
---
- Fixes #2203

### DIFF
--- a/Assets/HoloToolkit/Utilities/Scripts/Editor/SceneSettingsWindow.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/Editor/SceneSettingsWindow.cs
@@ -140,12 +140,14 @@ namespace HoloToolkit.Unity
             if (Values[SceneSetting.AddSpatialMapping])
             {
                 var spatialMappingManagers = FindObjectsOfType<SpatialMappingManager>();
-                foreach (var spatialMappingManager in spatialMappingManagers)
+                if (spatialMappingManagers.Length > 0)
                 {
-                    DestroyImmediate(spatialMappingManager.gameObject.GetParentRoot());
+                    Debug.LogWarning("There's already a SpatialMappingManager in the scene. Did not add the SpatialMapping prefab.");
                 }
-
-                PrefabUtility.InstantiatePrefab(AssetDatabase.LoadAssetAtPath<GameObject>(AssetDatabase.GUIDToAssetPath(SpatialMappingPrefabGUID)));
+                else
+                {
+                    PrefabUtility.InstantiatePrefab(AssetDatabase.LoadAssetAtPath<GameObject>(AssetDatabase.GUIDToAssetPath(SpatialMappingPrefabGUID)));
+                }
             }
 
             EditorSceneManager.MarkSceneDirty(SceneManager.GetActiveScene());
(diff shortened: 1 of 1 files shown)


Assets/HoloToolkit/Utilities/Scripts/Editor/SceneSettingsWindow.cs
```diff
@@ -129,6 +137,17 @@ protected override void ApplySettings()
                 FindObjectOfType<InputManager>().GetComponent<SimpleSinglePointerSelector>().Cursor = FindObjectOfType<Cursor>();
             }
 
+            if (Values[SceneSetting.AddSpatialMapping])
+            {
+                var spatialMappingManagers = FindObjectsOfType<SpatialMappingManager>();
+                foreach (var spatialMappingManager in spatialMappingManagers)
+                {
+                    DestroyImmediate(spatialMappingManager.gameObject.GetParentRoot());
```
StephenHodgson: Do you want to destroy the GameObject?

What if there's other components ppl are using on it?

davidkline-ms: destroying the parent seems risky.

should we just leave what we find alone if it

also, we should warn users if we find more than one pre-existing spatial mapping manager

keveleigh: Ahh yeah, I copied the line above which destroys the cursor's parent: https://github.com/keveleigh/HoloToolkit-Unity/blob/9525e0803aacc26dfa13b493a0f44300c656137d/Assets/HoloToolkit/Utilities/Scripts/Editor/SceneSettingsWindow.cs#L132

I can update it to instead just destroy its own GameObject, like the InputSystem does (multiple times): https://github.com/keveleigh/HoloToolkit-Unity/blob/9525e0803aacc26dfa13b493a0f44300c656137d/Assets/HoloToolkit/Utilities/Scripts/Editor/SceneSettingsWindow.cs#L92

or should all of this behavior be updated?

davidkline-ms: I'd rather have a warning of too many spatial mapping managers and leave the existing one(s) untouched

Issue #2401, opened 2018-07-05
vNext Task: Add inscribed rectangle/volume support to Boundary Manager


davidkline-ms:
## Overview
For platforms (ex: Windows Mixed Reality) that support non-rectangular boundaries, it can be useful to provide a rectangular space that fits within the boundary volume.

## Requires
Boundary edges to be exposed, so that applications can perform custom operations (ex: rendering) on the actual boundary volume.
Port of the inscribed rectangle feature that is being added to the june18_dev branch.
Port of the boundary test scene from june18_dev

## Acceptance Criteria
- [x] IMixedRealityBoundarySystem.BoundsEdges describes the actual playspace bounds.
- [x] The rectangular playspace is reported via IMixedRealityBoundarySystem.InscribedRectangle
- [x] Verify ported boundary test scene behaves as expected

PR #2469, opened 2018-07-20
Add boundary inscribed rectangle and visualization demo scene

davidkline-ms:
Overview
---
Ported june18_dev inscribed rectangle feature.
Added demo scene that illustrates the boundary geometry (tracked area) and inscribed rectangle (play area).

Changes
---
- Replaced OutscribedVolume with GeometryBounds
- Replaced InscribedVolume with InscribedRectanglarBounds
- Added InscribedRectangle class
- Added EdgeUtils class
- Implemented Contains method to determine if a point is within the specified area
- Fixes: #2401  .


Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -66,35 +69,120 @@ public override void Reset()
...
             // Get the boundary geometry.
             List<Vector3> boundaryGeometry = new List<Vector3>(0);
-            if (Boundary.TryGetGeometry(boundaryGeometry))
+            List<Edge> boundaryEdges = new List<Edge>(0);
+
+            if (Boundary.TryGetGeometry(boundaryGeometry, Boundary.Type.TrackedArea))
             {
+                // FloorHeight starts out as null. Use a very high value for the floor to ensure
+                // that we do not accidentally set it too low.
+                float floorHeight = 1000f;
```
StephenHodgson: Should this be configurable?

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Materials.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
keveleigh: This meta doesn't have a corresponding folder.

image

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Prefabs.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
keveleigh: This meta doesn't have a corresponding folder.

image

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,603 @@
...
+        /// True if the point is within the rectangle's bounds, false otherwise.
+        /// </returns>
+        /// <exception cref="InvalidOperationException">The rectangle is not valid.</exception>
+        public bool IsInsideBoundary(Vector2 point)
+        {
+            if (!IsValid)
+            {
+                throw new InvalidOperationException("A point cannot be within an invalid rectangle.");
+            }
+
```
StephenHodgson: Extra line

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealityConfigurationProfileInspector.cs
```diff
@@ -102,34 +102,34 @@ public override void OnInspectorGUI()
...
                 case ExperienceScale.World:
-                    scaleDesciription = "The user is free to move about the world. Relies upon knowledge of the environment (Spatial Anchors and Spatial Mapping).";
+                    scaleDescription = "The user is free to move about the world. Relies upon knowledge of the environment (Spatial Anchors and Spatial Mapping).";
                     break;
             }
-            if (scaleDesciription != string.Empty)
+            if (scaleDescription != string.Empty)
             {
                 GUILayout.Space(6f);
                 EditorGUILayout.LabelField("Description:", EditorStyles.label);
```
StephenHodgson: Let's replace this with an into help box. It'll look a bit nicer and conform to the pattern we're using in the other profile inspectors for describing their function.

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        /// <summary>
+        /// Checks to see if a point is valid.
+        /// </summary>
+        /// <param name="point">The point to check.</param>
+        /// <returns>True if the point is valid, false otherwise./returns>
+        /// <remarks>
+        /// A point is considered invalid if any one of it's coordinate values are:
+        /// * Not a number
```
StephenHodgson: Is `*` valid for bullets in xml comments?

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
```diff
@@ -0,0 +1,102 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Utilities
+{
+    public static class EdgeUtils
```
StephenHodgson: EdgeUtility?

davidkline-ms: Matches some of our other utils classes (ex: MathUtils), but I am happy to change

StephenHodgson: I don't mind updating those other ones.

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: It's just a high number (the floor should never be 1000m above the origin), don't think it really needs to be configurable. I *could* make the number bigger :)

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
davidkline-ms: I got this from the june18_dev branch version. i can reformat the comment

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
davidkline-ms: sounds good. i'll change this one.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Materials.meta
davidkline-ms: missed deleting this after I moved files around..

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Prefabs.meta
davidkline-ms: missed deleting this after I moved files around..

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealityConfigurationProfileInspector.cs
davidkline-ms: thx for the suggestion. i'll take a look

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
SimonDarksideJ: Yeah, should be a remark, if it's to highlight something

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: extra line has been removed, not sure why GitHub is showing it as unchanged

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,113 @@
...
+        private Material inscribedRectangleMaterial = null;
+
+        /// <summary>
+        /// The configured boundary system manager.
+        /// </summary>
+        private IMixedRealityBoundarySystem boundaryManager = null;
+
+        private void Start()
+        {
+            MixedRealityManager mixedRealityManager = MixedRealityManager.Instance;
```
SimonDarksideJ: This needs to use the standard approach for getting a manager with the private property above, not in start, e.g.

```        private IMixedRealityInputSystem inputSystem = null;
        protected IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
```

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -66,35 +69,120 @@ public override void Reset()
...
+
+            // Boundaries are supported for Room Scale experiences only.
             if (XRDevice.GetTrackingSpaceType() != TrackingSpaceType.RoomScale)
             {
-                // Boundaries are supported for Room Scale experiences only.
                 return;
             }
 
-            OutscribedVolume = new Bounds();
+#if !UNITY_WSA
```
SimonDarksideJ: Why is this here?  it's not even needed.

Managers should not have platform specific code, that is the devices layer's job

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -136,11 +228,17 @@ private void SetTrackingSpace()
         /// </remarks>
         private void SetPlatformBoundaryVisibility()
         {
-            if (Boundary.configured)
+#if !UNITY_WSA
```
SimonDarksideJ: same here, remove any platform specific #IF's

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
StephenHodgson: Yeah if it's an instance there's no reason to cache this reference to the MR Manager.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -38,7 +43,7 @@ private void Start()
         /// </summary>
         private void AddQuad()
         {
-            IMixedRealityBoundarySystem boundaryManager = MixedRealityManager.Instance?.GetManager<IMixedRealityBoundarySystem>();
+            boundaryManager = boundaryManager ?? MixedRealityManager.Instance?.GetManager<IMixedRealityBoundarySystem>();
```
StephenHodgson: Instead of using `MixedRealityManager.Instance` let's confirm the MR Manager is active else we bail early.

StephenHodgson: Look at the ways we get the input system and have a property with that checks if the boundary system is cached in a similar way.

davidkline-ms: Here's how I see the input system being accessed...

>+        private IMixedRealityInputSystem inputSystem = null; 
+        protected IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>()); 

This looks pretty close to what I originally had (minus the ?? check).

>Instead of using MixedRealityManager.Instance? let's confirm the MR Manager is active else we bail early.

Other than checking the instance, how do you suggest checking that the manager is active? do you mean if the monobehavior is enabled?

StephenHodgson: We're not using monobehaviours for our managers, so that might be a bit difficult  😅 
But there is a way to check if the system is enabled.

davidkline-ms: Question: Should SDK/Example scripts be checking the scene for the MixedRealityManager component? Isnt it more efficient to check the instance parameter (is static, would be null if not in the scene).

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -124,8 +123,7 @@ private bool IsBoundarySystemEnabled()
                 return false;
             }
 
-            MixedRealityConfigurationProfile profile = MixedRealityManager.Instance?.ActiveProfile;
-            return (bool)(profile?.EnableBoundarySystem);
+            return (bool)(MixedRealityManager.Instance?.ActiveProfile?.EnableBoundarySystem);
```
Jarodshow: as bool?

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,129 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces;
+using Microsoft.MixedReality.Toolkit.Internal.Managers;
+using UnityEngine;
+using UnityEngine.Experimental.XR;
+
```
keveleigh: extra line

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -66,33 +69,109 @@ public override void Reset()
             InitializeInternal();
         }
 
+        public bool Contains(Vector3 location)
+        {
+            return Contains(location, Boundary.Type.TrackedArea);
+        }
+
+        /// <inheritdoc/>
+        public bool Contains(Vector3 location, Boundary.Type boundaryType)
```
keveleigh: The original `ContainsObject` method [transformed the provided location into the camera's frame of reference](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/june18_dev/Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs#L175) before checking it. Do you do that anywhere?

This is important, because it accounts for the case where you've teleported and the boundary follows you. Without this change, the `Contains` method always checks as if the boundary is wherever it was at app launch.

It also [changed the rectangle params when they're requested](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/june18_dev/Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs#L237).

davidkline-ms: missed that I believe.... i'll have to fix that

davidkline-ms: i thought i ported most of it, though... hopefully i'm not missing too much

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
davidkline-ms: nope. compiler is not happy with that. :)

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityBoundarySystem.cs
```diff
@@ -36,15 +37,45 @@ public interface IMixedRealityBoundarySystem : IMixedRealityManager
...
         /// <summary>
-        /// A three dimensional volume as described by the largest rectangle that
-        /// is contained withing the playspace and the configured height.
+        /// Indicates the height of the floor, in relation to the coordinate system origin.
         /// </summary>
-        Bounds InscribedVolume { get; }
+        /// <remarks>
+        /// If a floor has been located, FloorHeight.HasValue will be true, otherwise it will be false.
+        /// </remarks>
+        float? FloorHeight { get; }
```
SimonDarksideJ: Does this need to be a nullable float?  Surely 0 would suffice without the overhead. (or -1)

keveleigh: 0 and -1 are valid `FloorHeight`s, since it exists along the Y axis.

davidkline-ms: There's not THAT much overhead and -1, in theory is a valid floor offset. I'd rather keep the nullable so that we can know that no floor was found

StephenHodgson: I think nullable makes sense here.

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: Should be handled now.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
StephenHodgson: This line should go away if we're following the pattern above.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityBoundarySystem.cs
```diff
@@ -77,5 +72,19 @@ public interface IMixedRealityBoundarySystem : IMixedRealityManager
...
+        /// <summary>
+        /// Returns the description of the inscribed rectangular bounds.
+        /// </summary>
+        /// <param name="center">The center of the rectangle.</param>
+        /// <param name="angle">The orientation of the rectangle.</param>
+        /// <param name="width">The width of the rectangle.</param>
+        /// <param name="height">The height of the rectangle.</param>
+        /// <returns></returns>
+        bool TryGetRectangularBounds(
+            out Vector2 center,
```
StephenHodgson: these are short enough to put on one line.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <summary>
+        /// Displays the boundary as a quad primitive.
+        /// </summary>
+        private void AddQuad()
+        {
+            if (!IsBoundarySystemEnabled())
+            {
+                return;
+            }
+            boundaryManager = boundaryManager ?? MixedRealityManager.Instance?.GetManager<IMixedRealityBoundarySystem>();
```
StephenHodgson: We can get rid of this once the cache pattern is established.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,148 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
```
StephenHodgson: we're not using this

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,124 @@
...
+        [SerializeField]
+        [Tooltip("Material used to draw the area outside of the boundary geometry.")]
+        private Material outOfBoundsMaterial = null;
+
+        [SerializeField]
+        [Tooltip("Material used to draw the area inside of the inscribed rectangle.")]
+        private Material inscribedRectangleMaterial = null;
+
+        /// <summary>
+        /// Boundary system implementation.
```
StephenHodgson: Do we need documentation for this?

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityBoundarySystem.cs
```diff
@@ -36,15 +37,54 @@ public interface IMixedRealityBoundarySystem : IMixedRealityManager
...
         /// </summary>
-        Bounds InscribedVolume { get; }
+        /// <param name="center">The center of the rectangle.</param>
+        /// <param name="angle">The orientation of the rectangle.</param>
+        /// <param name="width">The width of the rectangle.</param>
+        /// <param name="height">The height of the rectangle.</param>
+        /// <returns></returns>
+        bool TryGetRectangularBoundsParams(
+            out Vector2 center,
+            out float angle,
```
StephenHodgson: these aren't that long, could we put them on a single line?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+            point -= Center;
+            point = RotatePoint(point, Vector2.zero, MathUtils.DegreesToRadians(-Angle));
+
+            bool inWidth = Mathf.Abs(point.x) <= (Width / 2.0f);
```
StephenHodgson: ` width * 0.5f`?

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -136,11 +251,7 @@ private void SetTrackingSpace()
         /// </remarks>
         private void SetPlatformBoundaryVisibility()
```
StephenHodgson: If we're setting shouldn't that also mean we're passing a parameter?

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -66,35 +71,141 @@ public override void Reset()
...
+            }
+
+            // Not in either boundary type.
+            return false;
+        }
+
+        /// <inheritdoc/>
+        public bool TryGetRectangularBoundsParams(
+            out Vector2 center, 
+            out float angle,
```
StephenHodgson: These could be a single line.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// bounds are a different color.
+        /// </summary>
+        private void AddIndicators()
+        {
+            // Get the rectangular bounds.
+            Vector2 centerRect = EdgeUtilities.InvalidPoint;
+            float angleRect = 0f;
+            float widthRect = 0f;
+            float heightRect = 0f;
+            bool haveBoundary = (bool)(BoundaryManager?.TryGetRectangularBoundsParams(out centerRect, out angleRect, out widthRect, out heightRect));
```
StephenHodgson: Let's bail and throw an error if the manager is null instead.

StephenHodgson:
Do we also want to move `MixedRealityBoundarySystem.cs` into the SDK to match our pattern with the reorg?

StephenHodgson:
Do we also want to move `MixedRealityBoundarySystem` into the SDK to conform to our SDK reorg?

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
davidkline-ms: i wanted to show something when there was no boundary (or system) but at this point, if it will get the change approved.... fine

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: i find it harder to read that way, but siure

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: i'll rename to update

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: pretty suire the compiler does the right thing, but ok

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
davidkline-ms: didnt think so, but i can add one

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityBoundarySystem.cs
davidkline-ms: as you wish

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
StephenHodgson: Ah I see, Just curious. That makes sense then.

davidkline-ms:
> Do we also want to move MixedRealityBoundarySystem into the SDK to conform to our SDK reorg?

I prefer all default systems/managers be in the _Core. That way someone can install ONLY the core and be able to succeed. otherwise, we have to have too many packages or the dev has to prune the ones we provide to get rid of the prefabs/etc that they may not want.

This was my understanding of the architecture from Simon's original diagram. Otherwise, the core seems almost unnecessary.

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: removed complete.y instead

StephenHodgson:
>This was my understanding of the architecture from Simon's original diagram. Otherwise, the core seems almost unnecessary.

Well it's the framework.  I was just curious if we wanted to follow a specific pattern we agreed on in the reorg.

davidkline-ms:
> pattern we agreed on in the reorg.

I dont think we actually agreed. I recall the plan being to revisit early Aug.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,133 @@
...
+            float angleRect = 0f;
+            float widthRect = 0f;
+            float heightRect = 0f;
+            if (!(bool)(BoundaryManager?.TryGetRectangularBoundsParams(out centerRect, out angleRect, out widthRect, out heightRect)))
+            {
+                // If we have no boundary manager or rectangular bounds we will show no indicators
+                return;
+            }
+
+                const int indicatorCount = 20;
```
StephenHodgson: formatting seems off here.

StephenHodgson:
> I don't think we actually agreed.

Then why did we approve the reorg PR?

davidkline-ms:
> Then why did we approve the reorg PR?

Because much of it was good and we were going to revisit in August.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// Displays the boundary as a quad primitive.
+        /// </summary>
+        private void AddQuad()
+        {
+            // Get the rectangular bounds.
+            Vector2 center = EdgeUtilities.InvalidPoint;
+            float angle = 0f;
+            float width = 0f;
+            float height = 0f;
+            if (!(bool)(BoundaryManager?.TryGetRectangularBoundsParams(out center, out angle, out width, out height)))
```
StephenHodgson: Why do we need to cast to a bool if the return type is a bool?

I'm also seeing a possible invalid operation exception

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealityBoundarySystem.cs
```diff
@@ -36,15 +37,50 @@ public interface IMixedRealityBoundarySystem : IMixedRealityManager
...
+        /// Determines if a location is within the specified area of the boundary space.
+        /// </summary>
+        /// <param name="location">The location to be checked.</param>
+        /// <returns>True if the location is within the specified area of the boundary space.</returns>
+        /// <remarks>
+        /// Use:
+        /// * Boundary.Type.PlayArea for the inscribed volume
+        /// * Boundary.Type.TrackedArea for the area defined by the boundary edges.
+        /// </remarks>
+        bool Contains(Vector3 location, Boundary.Type boundaryType);
```
StephenHodgson: missing boundaryType xml parameter description

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,133 @@
...
+using Microsoft.MixedReality.Toolkit.Internal.Utilities;
+using UnityEngine;
+using UnityEngine.Experimental.XR;
+
+namespace Microsoft.MixedReality.Toolkit.Examples.Demos
+{
+    /// <summary>
+    /// Demo class to show different ways of using the boundary API.
+    /// </summary>
+    public class BoundaryVisualizer : MonoBehaviour
```
StephenHodgson: wouldn't we want something like this in the SDK?

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
```diff
@@ -0,0 +1,133 @@
...
+            for (int xIndex = 0; xIndex < indicatorCount; ++xIndex)
+            {
+                for (int yIndex = 0; yIndex < indicatorCount; ++yIndex)
+                {
+                    Vector3 offset = new Vector3(xIndex * indicatorDistance, 0.0f, yIndex * indicatorDistance);
+                    Vector3 position = corner + offset;
+                    GameObject marker = GameObject.CreatePrimitive(PrimitiveType.Sphere);
+                    marker.transform.SetParent(transform);
+                    marker.transform.position = position;
+                    marker.transform.localScale = new Vector3(0.1f, 0.1f, 0.1f);
```
StephenHodgson: `marker.transform.localScale = Vector3.one * 0.1f;`?

Is there any particular reason we're using 0.1f?

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -66,35 +71,137 @@ public override void Reset()
             InitializeInternal();
         }
 
+        public bool Contains(Vector3 location)
+        {
+            return Contains(location, Boundary.Type.TrackedArea);
```
StephenHodgson: Why not just make boundaryType = tacked area by default instead of having this overload?

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -66,35 +71,137 @@ public override void Reset()
...
+            else
+            {
+                // TODO: How do we determine if we have a floor without boundaries
+                Debug.LogWarning("Failed to calculate boundary bounds.");
             }
         }
 
+        /// <summary>
+        /// Creates the two dimensional volume described by the largest rectangle that
+        /// is contained withing the playspace geoometry and the configured height.
```
StephenHodgson: geoometry -> geometry

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -66,35 +71,137 @@ public override void Reset()
...
+        /// </summary>
+        private void CreateInscribedBounds()
+        {
+            // We always use the same seed so that from run to run, the inscribed bounds are
+            // consistent.
```
StephenHodgson: single line?

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
StephenHodgson: playspace -> play space?

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
davidkline-ms: i almost wrote that code originally :)

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
davidkline-ms: goldilocks spheres (not to big not too small... just right :) )

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: same pattern as the june18_dev branch. i can make the parameter have a default value

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizer.cs
davidkline-ms: not in this form... but we should consider one that shows the playarea quad (inscribed rect)

davidkline-ms: renaming this script and class to be clear it is a demo

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -27,10 +27,15 @@ public class MixedRealityBoundaryManager : BaseManager, IMixedRealityBoundarySys
         public bool EnablePlatformBoundaryRendering { get; set; } = true;
```
StephenHodgson: Getting some warnings about `Scale`, `BoundaryHeight`, and `EnablePlatformBoundaryRendering` are all set in the constructor and their values are ignored during initialization.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+    /// <summary>
+    /// The InscribedRectangle class defines the largest rectangle within an
+    /// arbitrary shape.
+    /// </summary>
+    public class InscribedRectangle
+    {
+        /// <summary>
+        /// Total number of starting points randomly generated within the boundary.
+        /// </summary>
+        private static readonly int randomPointCount = 30;
```
StephenHodgson: could this be const?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+
+        /// <summary>
+        /// The total amount of height, in meters,  we want to gain with each binary search
+        /// change before we decide that it's good enough.
+        /// </summary>
+        private const float minimumHeightGain = 0.01f;
```
StephenHodgson: What's our naming conventions for private const values?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+
+        /// <summary>
+        /// The rotation angle, in degrees, of the inscribed rectangle.
+        /// </summary>
+        public float Angle { get; private set; } = 0f;
+
+        /// <summary>
+        /// Is the described rectangle valid?
+        /// </summary>
+        public bool IsValid
```
StephenHodgson: `public bool IsValid => EdgeUtilities.IsValidPoint(Center);`?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="geometryEdges"></param>
+        /// <param name="randomSeed"></param>
+        public InscribedRectangle(Edge[] geometryEdges, int randomSeed)
+        {
+            FindInscribedRectangle(geometryEdges, randomSeed);
```
StephenHodgson: Why not just make `FindInscribedRectangle` public?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+        /// drawn through those points. The midpoints of those lines are
+        /// used as the center of various rectangles, using a binary search to
+        /// vary the size, until the largest fit-able rectangle is found.
+        /// This is then repeated for predefined angles (0-180 in steps of 15)
+        /// and aspect ratios (1 to 15 in steps of 0.5).
+        /// </summary>
+        /// <param name="geometryEdges">The boundary geometry.</param>
+        /// <param name="randomSeed">Random number generator seed.</param>
+        /// <remarks>
+        /// For the most reproducable results, use the same randomSeed value
```
StephenHodgson: reproducable -> reproducible

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: that's odd... let me investigate

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+                    do
+                    {
+                        candidatePoint.x = ((float)random.NextDouble() * (maxX - minX)) + minX;
+                        candidatePoint.y = ((float)random.NextDouble() * (maxY - minY)) + minY;
+                    }
+                    while (!EdgeUtilities.IsInsideBoundary(geometryEdges, candidatePoint));
+
+                    startingPoints[i] = candidatePoint;
+                }
+
```
StephenHodgson: extra line

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: since it is private, yes

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: i don't think we have established one. i took this from the june18_dev PR i ported

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+            }
+
+            // Angles to use for fitting the rectangle within the boundary.
+            float[] angles = { 0, 15, 30, 45, 60, 75, 90, 105, 120, 135, 150, 165 };
```
StephenHodgson: couldn't we just declare this at the class scope so we don't allocate every time we look for  a new rectangle?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+                    FindSurroundingCollisionPoints(
+                        geometryEdges, 
+                        startingPoints[pointIndex], 
+                        angleRadians,
+                        out topCollisionPoint, 
+                        out bottomCollisionPoint, 
+                        out leftCollisionPoint, 
+                        out rightCollisionPoint);
+
+                    float newWidth = 0;
```
StephenHodgson: why set to 0 if the values are just going to get overwritten?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: this is how the original code was written. you generally arent going to do it more than once per instance of the app (boundaries are system setup thing)

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: vs needs a spell checker :)

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+                !EdgeUtilities.IsValidPoint(rightCollisionPoint))
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Rotate a two dimensiopnal point about another point by the specified angle.
```
StephenHodgson: dimensiopnal -> dimensional

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+        /// </summary>
+        /// <param name="point">The point to be rotated.</param>
+        /// <param name="origin">The point about which the rotation is to occur.</param>
+        /// <param name="angleRadians">The angle for the rotation, in radians</param>
+        /// <returns>
+        /// The coordinates of the rotated point.
+        /// </returns>
+        private Vector2 RotatePoint(Vector2 point, Vector2 origin, float angleRadians)
+        {
+            if (0.0f == angleRadians)
```
StephenHodgson: `if (angleRadians.Equals(0f))` for floating point precision

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: this was how the original code was written, i didnt want to tamper too much with it. easy enough chanbge

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: was in the original code. i left as much unchanged as possible

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+                if (EdgeUtilities.IsValidPoint(EdgeUtilities.GetIntersectionPoint(geometryEdges[i], topEdge)) ||
+                    EdgeUtilities.IsValidPoint(EdgeUtilities.GetIntersectionPoint(geometryEdges[i], rightEdge)) ||
+                    EdgeUtilities.IsValidPoint(EdgeUtilities.GetIntersectionPoint(geometryEdges[i], bottomEdge)) ||
+                    EdgeUtilities.IsValidPoint(EdgeUtilities.GetIntersectionPoint(geometryEdges[i], leftEdge)))
+                {
+                    return false;
+                }
+            }
+
+            // No collsions found with the rectangle. Success!
```
StephenHodgson: collsions -> collisions

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: we should update the june18_dev code then too

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+            Vector2 centerPoint,
+            float angleRadians,
+            float minArea,
+            out float width,
+            out float height)
+        {
+            width = 0.0f;
+            height = 0.0f;
+
+            float[] aspectRatios = {
```
StephenHodgson: Could be declared in the class scope so we don't have to allocate this every time we call this method.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
```diff
@@ -0,0 +1,570 @@
...
+
+            // For each aspect ratio we do a binary search to find the maximum rectangle that fits, 
+            // though once we start increasing our area by minimumHeightGain we call it good enough.
+            for (int i = 0; i < aspectRatios.Length; i++)
+            {
+                // The height is limited by the width. If a height would make our width exceed maxWidth, it can't be used
+                float searchHeightUpperBound = Mathf.Max(maxHeight, maxWidth / aspectRatios[i]);
+
+                // Set to the min height that will out perform our previous area at the given aspect ratio. This is 0 the first time.
+                // Derived from biggestAreaSoFar=height*(height*aspctRatio)
```
StephenHodgson: aspctRatio -> aspectRatio

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        /// <summary>
+        /// Determines if the specified point is within the provided geometry.
+        /// </summary>
+        /// <param name="geometryEdges">The geometry for which we are checking the point.</param>
+        /// <param name="point">The point being checked.</param>
+        /// <returns>
+        /// True if the point falls within the geometry, false otherwise.</returns>
+        public static bool IsInsideBoundary(
+            Edge[] geometryEdges,
```
StephenHodgson: why wrap this if we don't wrap the summary tags?

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+
+            return (intersections & 1) == 1;
+        }
+
+        /// <summary>
+        /// Checks to see if a point is valid.
+        /// </summary>
+        /// <param name="point">The point to check.</param>
+        /// <returns>True if the point is valid, false otherwise./returns>
```
StephenHodgson: missing a `<` on /returns

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <remarks>
+        /// A point is considered invalid if any one of it's coordinate values are infinite or not a number.
+        /// </remarks>
+        public static bool IsValidPoint(Vector2 point)
+        {
+            return ((!float.IsInfinity(point.x) && !float.IsInfinity(point.y)) &&
```
StephenHodgson: that's a lot of parenthesis

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                return false;
+            }
+
+            // Check if a ray to the right (X+) intersects with an 
+            // odd number of edges (inside) or an even number of edges (outside)
+            var rightEdge = new Edge(point, new Vector2(maxWidth, point.y));
+
+            int intersections = 0;
+            foreach (Edge edge in geometryEdges)
```
StephenHodgson: for loop? (might see perf improvements, but I don't think mobile platforms have bounds)

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
```diff
@@ -0,0 +1,99 @@
...
+            float sB_y = edgeB.PointB.y - edgeB.PointA.y;
+
+            float s = (-sA_y * (edgeA.PointA.x - edgeB.PointA.x) + sA_x * (edgeA.PointA.y - edgeB.PointA.y)) / (-sB_x * sA_y + sA_x * sB_y);
+            float t = (sB_x * (edgeA.PointA.y - edgeB.PointA.y) - sB_y * (edgeA.PointA.x - edgeB.PointA.x)) / (-sB_x * sA_y + sA_x * sB_y);
+
+
+            if ((s >= 0) && (s <= 1) && (t >= 0) && (t <= 1))
+            {
+                // Collision detected
+                return new Vector2(edgeA.PointA.x + (t * sA_x), edgeA.PointA.y + (t * sA_y));
```
StephenHodgson: depending on how often we get the intersection point, we may see a perf increase if we use a cached vector2 to do the calc.

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
davidkline-ms: was how it was originally coded... i missed updating this one

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
davidkline-ms: this is how the june18_dev branch is coded. i honestly dont know this code well enough to want to change it too much

StephenHodgson: example: 
```
private static Vector2 calcVector = Vector2.zero;

/// <summary>
/// Returns the point at which two <see cref="Edge"/> values intersect.
/// </summary>
/// <param name="edgeA">The first edge</param>
/// <param name="edgeB">The second edge</param>
/// <returns>
/// A Vector2 representing the point at which the two edges intersect, InscribedRectangleDescription.InvalidPoint otherwise.
/// </returns>
public static Vector2 GetIntersectionPoint(Edge edgeA, Edge edgeB)
{
    float sA_x = edgeA.PointB.x - edgeA.PointA.x;
    float sA_y = edgeA.PointB.y - edgeA.PointA.y;
    float sB_x = edgeB.PointB.x - edgeB.PointA.x;
    float sB_y = edgeB.PointB.y - edgeB.PointA.y;

    float s = (-sA_y * (edgeA.PointA.x - edgeB.PointA.x) + sA_x * (edgeA.PointA.y - edgeB.PointA.y)) / (-sB_x * sA_y + sA_x * sB_y);
    float t = (sB_x * (edgeA.PointA.y - edgeB.PointA.y) - sB_y * (edgeA.PointA.x - edgeB.PointA.x)) / (-sB_x * sA_y + sA_x * sB_y);


    if ((s >= 0) && (s <= 1) && (t >= 0) && (t <= 1))
    {
        // Collision detected
        calcVector.x = edgeA.PointA.x + (t * sA_x);
        calcVector.y = edgeA.PointA.y + (t * sA_y);
        return calcVector;
    }

    return InvalidPoint;
}
```

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
davidkline-ms: still not used to this new syntax

Assets/MixedRealityToolkit/_Core/Utilities/MathUtils.cs
```diff
@@ -419,5 +419,25 @@ public static Vector3 NearestPointToLinesLeastSquares(IEnumerable<Ray> rays)
...
+        }
+
+        /// <summary>
+        /// Convert radians to degrees.
+        /// </summary>
+        /// <param name="radians">Angle, in radians.</param>
+        /// <returns>Angle, in degrees.</returns>
+        public static double RadiansToDegrees(float radians)
+        {
+            return (double)(radians * Mathf.Rad2Deg);
```
StephenHodgson: Does casting to double give you better precision here even though the multiplied values are float types?

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: where are you seeing these? i dont see them in the unity console window or in VS

davidkline-ms: and they are definitely being referenced in the implementation

Assets/MixedRealityToolkit/_Core/Utilities/MathUtils.cs
davidkline-ms: doubtful. i can make it return a float

Assets/MixedRealityToolkit/_Core/Boundary/MixedRealityBoundarySystem.cs
StephenHodgson: The values are overwritten by the constructor. So it's redundant to set them in the property initializer.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
StephenHodgson: Resharper ftw

Assets/MixedRealityToolkit/_Core/Utilities/MathUtils.cs
StephenHodgson: Well the method above it is the same. So we can probably just remove the duplicate

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/InscribedRectangle.cs
StephenHodgson: Let's get rid of the redundant public method and just rename the private one to match.

StephenHodgson:
Pretty sure the _Core/Boundary folder should be a sibling of _Core like the InputSystem, Sharing, and Spatial feature folders.

davidkline-ms:
> Pretty sure the _Core/Boundary folder should be a sibling of _Core like the InputSystem, Sharing, and Spatial feature folders.

Let's file a task to move it. That should be a super fast PR.

keveleigh:
Seems more in scope of the PR that put it in this folder, which looks to have been your #2028.

Rearranging an entire feature area doesn't seem in scope for a PR adding a subset of that feature area.

StephenHodgson:
It's a single file in a single folder.

SimonDarksideJ:
I would have to agree with @StephenHodgson on this.
If we are updating / adding a feature, then the scope includes any move to place the changes in the correct location and update namespaces.  Doing outside the PR seems a bit superfluous.

The original folder was only a suggestion, like input.  We since agreed and moved it to it's own feature folder.  As Boundary is a Feature, is should move as Stephen suggests.

What will need discussing later, is whether the visualizer itself is a demo, or (more likely) an SDK component.  That we can review for Beta.

StephenHodgson:
I made the change request easy with a PR against your branch.

davidkline-ms:
> I made the change request easy with a PR against your branch.

Thank you, sir!

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Displays the boundary as a quad primitive.
+        /// </summary>
+        private void AddQuad()
+        {
+            // Get the rectangular bounds.
+            Vector2 center = EdgeUtilities.InvalidPoint;
+            float angle = 0f;
+            float width = 0f;
+            float height = 0f;
+            if ((BoundaryManager == null) || !BoundaryManager.TryGetRectangularBoundsParams(out center, out angle, out width, out height))
```
SimonDarksideJ: Daft sticking point, if this call is to prevent the function running.  Then it should be the first thing in the call.
Don't declare values unless they are intended to be used or you are just generating garbage.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// bounds are a different color.
+        /// </summary>
+        private void AddIndicators()
+        {
+            // Get the rectangular bounds.
+            Vector2 centerRect = EdgeUtilities.InvalidPoint;
+            float angleRect = 0f;
+            float widthRect = 0f;
+            float heightRect = 0f;
+            if ((BoundaryManager == null) || !BoundaryManager.TryGetRectangularBoundsParams(out centerRect, out angleRect, out widthRect, out heightRect))
```
SimonDarksideJ: Same here, This should be the first thing in the call.

Assets/MixedRealityToolkit/Boundary/MixedRealityBoundarySystem.cs
```diff
@@ -0,0 +1,235 @@
...
+                default:
+                    trackingSpace = TrackingSpaceType.Stationary;
+                    Debug.LogWarning("Unknown / unsupported ExperienceScale. Defaulting to Stationary tracking space.");
+                    break;
+            }
+
+            bool trackingSpaceSet = XRDevice.SetTrackingSpaceType(trackingSpace);
+            if (!trackingSpaceSet)
+            {
+                // TODO: Implement ExperienceScale fallback logic
```
SimonDarksideJ: Can we add a comment to the header of this class to detail it's capabilities, if some are not intended to be implemented until later.
Also ensure the boundary documentation reflects this.

Assets/MixedRealityToolkit/_Core/Utilities/EdgeUtilities.cs
```diff
@@ -0,0 +1,107 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Utilities
+{
+    public static class EdgeUtilities
```
SimonDarksideJ: Summary tag

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
davidkline-ms: they are out params to the TryGet function... they must be defined.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
davidkline-ms: they are out params to the TryGet function... they must be defined.

Assets/MixedRealityToolkit/Boundary/MixedRealityBoundarySystem.cs
davidkline-ms: Reworded the todo. You are correct that we need to properly document / expose the proposed fallback. Hopefully this will pass muster :)

            if (!trackingSpaceSet)
            {
                // TODO: consider how best to handle this situation
            }

Issue #1858, opened 2018-03-22
[vNext] Move MRTK versioning from Unity release based to SemVer


davidkline-ms:
Our current, Unity release based versioning system is not NuGet friendly. To remedy this, we will be changing the MRTK version to be SemVer (http://www.semver.org) based.

The current plan is to use year.month.revision (ex: 2018.06.0) to version the MRTK. This will necessitate consumers of the toolkit to consult release notes and/or README.md to determine the supported Unity version(s) for the package.

PR #2472, opened 2018-07-20
Add version files (semver) and propogate license to package roots

davidkline-ms:
Overview
---
Added Version.txt file containing the following: Mixed Reality Toolkit <package subname> <version> to each root level package folder (ex: Assets\MixedRealityToolkit).

Add license.txt from the repository root to the packages roots.

- Fixes: #1858  .

### DIFF
--- a/Assets/MixedRealityToolkit-Examples/License.txt
+++ b/Assets/MixedRealityToolkit-Examples/License.txt
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017 Microsoft Corporation
+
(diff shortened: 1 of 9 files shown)


StephenHodgson:
Did we want three different license text files for each folder?

StephenHodgson:
But they're essentially the same exact file? 

We could write an exporter that can update the version and copy the version text and license into the proper folder.

davidkline-ms:
true, the license files are the same. the version files are not since we MIGHT support limited mixing and matching of the packages.

StephenHodgson:
This PR was also missing meta files for these assets.

PR #2481, opened 2018-07-24
Fix in TryGetBoundaryRectanglePoints after teleport

keveleigh:
Overview
---
After a teleport, the boundary APIs still return positions around the origin, as though the camera has not moved.

I previously updated the RectangleParams method, but did not transform the RectanglePoints method. This fixes that.

Changes
---
- Related to #2469


keveleigh:
@StephenHodgson 
@davidkline-ms and I made sure it was accounted for in #2469 (this method ended up not existing there as written here, so we're good).

PR #2490, opened 2018-07-26
pose property drawer

StephenHodgson:
image

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/MixedRealityPosePropertyDrawer.cs
+++ b/Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/MixedRealityPosePropertyDrawer.cs
@@ -1,0 +1,60 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+using UnityEditor;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Inspectors.PropertyDrawers
+{
+    [CustomPropertyDrawer(typeof(MixedRealityPose))]
+    public class MixedRealityPosePropertyDrawer : PropertyDrawer
+    {
+        private readonly GUIContent positionContent = new GUIContent("Position");
+        private readonly GUIContent rotationContent = new GUIContent("Rotation");
+        private Quaternion rotationCache = Quaternion.identity;
+        private Vector3 rotations = Vector3.zero;
+
+        public override float GetPropertyHeight(SerializedProperty property, GUIContent label)
+        {
+            return EditorGUIUtility.singleLineHeight * 3f;
+        }
+
+        public override void OnGUI(Rect position, SerializedProperty property, GUIContent label)
+        {
+            bool lastMode = EditorGUIUtility.wideMode;
+            EditorGUIUtility.wideMode = true;
+            EditorGUI.BeginProperty(position, label, property);
+            EditorGUI.PrefixLabel(position, GUIUtility.GetControlID(FocusType.Passive), label);
+            EditorGUI.indentLevel++;
+
+            var fieldHeight = position.height / 3f;
+            var positionRect = new Rect(position.x, position.y + fieldHeight, position.width, fieldHeight);
+            var rotationRect = new Rect(position.x, position.y + fieldHeight * 2, position.width, fieldHeight);
+
+            EditorGUI.PropertyField(positionRect, property.FindPropertyRelative("position"), positionContent);
+
+            var rotationProperty = property.FindPropertyRelative("rotation");
+            rotationCache = rotationProperty.quaternionValue;
+            Debug.Assert(rotationCache.w.Equals(1f));
+            EditorGUI.BeginChangeCheck();
+            rotations.x = rotationCache.x;
+            rotations.y = rotationCache.y;
+            rotations.z = rotationCache.z;
+            rotations = EditorGUI.Vector3Field(rotationRect, rotationContent, rotations);
+
+            if (EditorGUI.EndChangeCheck())
+            {
+                rotationCache.w = 1f;
+                rotationCache.x = rotations.x;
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/MixedRealityPosePropertyDrawer.cs
```diff
@@ -0,0 +1,60 @@
...
+    public class MixedRealityPosePropertyDrawer : PropertyDrawer
+    {
+        private readonly GUIContent positionContent = new GUIContent("Position");
+        private readonly GUIContent rotationContent = new GUIContent("Rotation");
+        private Quaternion rotationCache = Quaternion.identity;
+        private Vector3 rotations = Vector3.zero;
+
+        public override float GetPropertyHeight(SerializedProperty property, GUIContent label)
+        {
+            return EditorGUIUtility.singleLineHeight * 3f;
```
davidkline-ms: I presume the 3f is to indicate we need space for the title plus lines for each of the pose components. just in case we decide to change the layout (and to establish a solid pattern), should we, instead, define a constant so that when it is used later (ex: var fieldHeight = position.height / 3f;) we can skip the hard-coded constant?

StephenHodgson: Unfortunately we have to set the property height here only per Unity.

Yes, it's space for the property name, position, and rotation lines.

StephenHodgson: I'm open to moving the 3f to a local field and replacing it in the OnGUI below if that's what you're referring to.

SimonDarksideJ: Given this is us configuring the editor layout to draw correctly, I see no reason to have this configurable.  It's not something a user would touch.

SimonDarksideJ:
Ok, the error is called because the field initialises as a zero length quaternion, even though the base property is defined as a ZeroIdentity.

Since we SET the W property to 1F anyway, is this assert even needed?

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/MixedRealityPosePropertyDrawer.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            var fieldHeight = position.height / 3f;
+            var positionRect = new Rect(position.x, position.y + fieldHeight, position.width, fieldHeight);
+            var rotationRect = new Rect(position.x, position.y + fieldHeight * 2, position.width, fieldHeight);
+
+            EditorGUI.PropertyField(positionRect, property.FindPropertyRelative("position"), positionContent);
+
+            var rotationProperty = property.FindPropertyRelative("rotation");
+            rotationCache = rotationProperty.quaternionValue;
+            Debug.Assert(rotationCache.w.Equals(1f));
```
SimonDarksideJ: Remove Assert, not needed as we SET it below.

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/MixedRealityPosePropertyDrawer.cs
```diff
@@ -0,0 +1,60 @@
...
+            rotations.z = rotationCache.z;
+            rotations = EditorGUI.Vector3Field(rotationRect, rotationContent, rotations);
+
+            if (EditorGUI.EndChangeCheck())
+            {
+                rotationCache.w = 1f;
+                rotationCache.x = rotations.x;
+                rotationCache.y = rotations.y;
+                rotationCache.z = rotations.z;
+                rotationProperty.quaternionValue = rotationCache;
```
SimonDarksideJ: Change this line to the following and remove the rotationCache lines above it.
```rotationProperty.quaternionValue = Quaternion.Euler(rotations.x, rotations.y, rotations.z);```

File: Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/MixedRealityPosePropertyDrawer.cs [hunk 4]
```diff
@@ -0,0 +1,60 @@
...
+            var positionRect = new Rect(position.x, position.y + fieldHeight, position.width, fieldHeight);
+            var rotationRect = new Rect(position.x, position.y + fieldHeight * 2, position.width, fieldHeight);
+
+            EditorGUI.PropertyField(positionRect, property.FindPropertyRelative("position"), positionContent);
+
+            var rotationProperty = property.FindPropertyRelative("rotation");
+            rotationCache = rotationProperty.quaternionValue;
+            Debug.Assert(rotationCache.w.Equals(1f));
+            EditorGUI.BeginChangeCheck();
+            rotations.x = rotationCache.x;
```
SimonDarksideJ: Change these three lines as follows:
```            rotations.x = rotationCache.eulerAngles.x;
            rotations.y = rotationCache.eulerAngles.y;
            rotations.z = rotationCache.eulerAngles.z;```

Devs only think in eulars, Not a Quaternion expressed in a V3 :P

StephenHodgson: There's a very a very specific reason _why_ I didn't do this.

StephenHodgson: Go ahead and update your version to the Euler and then start changing values in the inspector. You'll find a nasty surprise.

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/MixedRealityPosePropertyDrawer.cs
SimonDarksideJ: If we are still moving ahead with this, then this Assert will still need to come out.

PR #2492, opened 2018-07-26
Initial Check in for the Controller Visualization functionality

SimonDarksideJ:
Overview
---
Added basic controller visualization for the Mixed Reality Toolkit Alpha
More improvements needed in beta for:
* Debug overlay
* Animation support
* Controller model node support

Implements
---
* Basic controller visualization
* Support for both Global and Specific controller Models
* Offset support for controller rendering
* Documentation

PR #2496, opened 2018-07-27
Misc Input System cleanup tasks

StephenHodgson:
- removed asserts in global listener registration and just opted for a return if it's already registered.
    - prevents spam in console, since we moved to a new pattern for the global listener assignment.
- cleaned up attach to controller a bit and renamed it to `ControllerPoseSynchronizer`
- updated some old terminology in focus provider.
- moved input handler inspectors into their own folders and namespaces.
- `SourcePoseEventData` pose event also sets the position and rotation, just in case someone wants it.


Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/MotionController/AttachToController.cs
```diff
@@ -1,47 +1,37 @@
...
 namespace Microsoft.MixedReality.Toolkit.SDK.UX.MotionController
 {
     /// <summary>
-    /// Waits for a controller to be initialized, then attaches itself to a specified element
+    /// Waits for a controller to be initialized, then attaches itself to a specified hand.
     /// </summary>
     public class AttachToController : InputSystemGlobalListener, IMixedRealitySourcePoseHandler
     {
         [SerializeField]
-        protected Vector3 PositionOffset = Vector3.zero;
```
keveleigh: Why are these removed now? These are offsets from the controller, not offsets for the controller. Where's that now handled by the visualizer?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/MotionController/AttachToController.cs
```diff
@@ -1,47 +1,37 @@
...
+        private Handedness handedness = Handedness.Left;
```
davidkline-ms: do we want to initialize to Handedness.None?

StephenHodgson: Pretty sure the default (even in master) is left. What's the point of attaching to a controller of it doesn't have a hand to attach to?

StephenHodgson: Pretty sure the default (even in master) is left. What's the point of attaching to a controller of it doesn't have a hand to attach to?

keveleigh: Default in master is left; default in June is unknown.

https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/june18_dev/Assets/HoloToolkit/Input/Scripts/Utilities/ControllerFinder.cs#L42

This was changed to facilitate spawning AttachToController/ControllerFinders at runtime, so it doesn't try to attach to a controller in OnEnable (which happens on Instantiate) and then get its handedness changed in Start (which happens after settings can be changed on the script).

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/MotionController/AttachToController.cs
keveleigh: Where? It's a massive PR. The only thing I see are controller offsets.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/MotionController/ControllerPoseSynchronizer.cs
```diff
@@ -0,0 +1,141 @@
...
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem.Handlers;
+using Microsoft.MixedReality.Toolkit.SDK.Input;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.MotionController
+{
+    /// <summary>
+    /// Waits for a controller to be initialized, then synchronizes its transform position to a specified handedness.
+    /// </summary>
+    public class ControllerPoseSynchronizer : InputSystemGlobalListener, IMixedRealitySourcePoseHandler
```
SimonDarksideJ: If we are agreed with this rename.  This should be done in the visualization PR, as that has already made changes to this (and the inspector) classes.
So please revert these and I'll rename in the visualizer PR

StephenHodgson: I still need these changes for the pointer viz, which is why I decided to do it separately.

SimonDarksideJ: Can we then move this file to another folder then, say handlers? (including it's inspector)

StephenHodgson: Yes

Assets/MixedRealityToolkit-SDK/Inspectors/Input/Handlers/ControllerPoseSynchronizerInspector.cs
```diff
@@ -0,0 +1,47 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.﻿
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+using Microsoft.MixedReality.Toolkit.SDK.Input.Handlers;
+using UnityEditor;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Inspectors.Input.Handlers
+{
+    [CustomEditor(typeof(ControllerPoseSynchronizer))]
+    public class ControllerPoseSynchronizerInspector : Editor
```
keveleigh: Since this is being renamed as (as far as I could tell from our conversation yesterday) redone slightly differently than AttachToController, can we change the meta's GUID to not prevent porting the original AttachToContoller over at some point?

StephenHodgson: I purposefully kept the GUID the same, but I can regenerate it if you'd like.

StephenHodgson: Done

keveleigh: If we think it'll eventually do everything the original script did, we can keep them. It feels like it's going for a slightly different use case though.

Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -0,0 +1,140 @@
...
+        #region Monobehaviour Implementation
+
+        protected override void OnEnable()
+        {
+            // Subscribe to interaction events
+            base.OnEnable();
+
+            if (DisableChildren)
+            {
+                gameObject.SetChildrenActive(false);
```
davidkline-ms: this is a bit confusing, why aren't we activating children on enable?

keveleigh: In master/June, the children are set inactive at first, then a check is made to try to find a controller. If one is found, they're immediately made active again in the callback.

Here, they're set active after SourceDetected, but that misses the case where a controller was detected before the prefab is instantiated.

This and ControllerFinder seem to have been ported incompletely or edited after the fact feature-wise, so I'm going to do a re-port soon.

Assets/MixedRealityToolkit-SDK/Inspectors/Input/Handlers/ControllerPoseSynchronizerInspector.cs
StephenHodgson: agreed.

Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
keveleigh: Good call; I missed that.

Assets/MixedRealityToolkit-SDK/Inspectors/Input/Handlers/ControllerPoseSynchronizerInspector.cs
StephenHodgson: Regenerating the guids will help in your "re-port". (Although, I think you should address that in the controller viz in #2492)

Assets/MixedRealityToolkit/_Core/EventDatum/Input/SourcePoseEventData.cs
```diff
@@ -91,11 +91,13 @@ public void Initialize(IMixedRealityInputSource inputSource, IMixedRealityContro
...
         /// <param name="inputSource"></param>
         /// <param name="controller"></param>
-        /// <param name="position"></param>
-        public void Initialize(IMixedRealityInputSource inputSource, IMixedRealityController controller, MixedRealityPose position)
+        /// <param name="pose"></param>
+        public void Initialize(IMixedRealityInputSource inputSource, IMixedRealityController controller, MixedRealityPose pose)
         {
             Initialize(inputSource, controller);
-            MixedRealityPose = position;
+            ThreeDofPosition = pose.Position;
```
keveleigh: General, non-PR question: is the source pose grip pose, position pose, or both? Do we expose them both somewhere? I couldn't find it yesterday.

StephenHodgson: This is only for the source position. Does not give grip or pointer poses.

You have to implement the IMixedRealityInputPoseHandler to get those.

keveleigh: But...where is the "source position" centered? It has to be one of them.

MixedRealityInputPoseHandler didn't seem to distinguish between grip and pointer.

StephenHodgson: > MixedRealityInputPoseHandler didn't seem to distinguish between grip and pointer.

Sure it does, that's what the InputActions are for. You compare the action id to the one you setup via the  event data.

> But...where is the "source position" centered?

That's up to the controller implementation.

keveleigh: What if I don't want to use InputActions?

StephenHodgson: Then get the data directly from the controller.

PR #2497, opened 2018-07-27
Adding custom cursor modifier support

shubhamarora28:
Overview
---
**Issue in existing cusor system:**
Cursor class uses CursorModifier in the following two methods:
- OnActiveModifier(CursorModifier modifier)
- OnPointerSpecificFocusChanged(IPointingSource pointer, GameObject oldFocusedObject, GameObject newFocusedObject)

because of which, if anyone writes their own cursor modifier(implementing ICursorModifier), their cursor modifier code will not be called/executed as OnPointerSpecificFocusChanged will not be able to get the custom cursor modifier component as it is not of the same type as expected by the Cursor class.

Solution:
- Changed OnActiveModifier(ICursorModifier modifier) signature to use ICursorModifier as the input parameter instead of CursorModifier
- Changed OnPointerSpecificFocusChanged default implementation to use ICursorModifier instead of CursorModifier

Implication:
All classes that derive from Cursor abstract class will need to change OnActiveModifier method parameter type

Diagram images of the existing system and proposed changes to the system are attached.

**Current Cursor System:**
cursorsystemdiagram_current

**Proposed changes to cursor system:**
cursorsystemdiagram_proposed

P.S.: Edited the images with MS Paint. Please don't judge. :D
We were not able to generate the class diagram using VS 2017 Professional. Any help on this will be appreciated.

Changes
---
- Modified Cursor.cs to use ICursorModifier instead of CursorModifier
- Modified AnimatedCursor.cs
- Added CustomCursorModifierExample scene
- Added CubeCursor prefab
- Added TestCursor.cs
- Added TestCursorModifier.cs

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Cursor/AnimatedCursor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Cursor/AnimatedCursor.cs
@@ -69,19 +69,27 @@ namespace HoloToolkit.Unity.InputModule
         /// Override to set the cursor animation trigger
         /// </summary>
         /// <param name="modifier"></param>
-        protected override void OnActiveModifier(CursorModifier modifier)
+        protected override void OnActiveModifier(ICursorModifier cursorModifier)
         {
-            base.OnActiveModifier(modifier);
+            base.OnActiveModifier(cursorModifier);
 
-            if (modifier != null)
+            if (cursorModifier != null)
             {
-                if ((modifier.CursorParameters != null) && (modifier.CursorParameters.Length > 0))
+                if (cursorModifier is CursorModifier)
                 {
-                    OnCursorStateChange(CursorStateEnum.Contextual);
-                    foreach (var param in modifier.CursorParameters)
+                    CursorModifier modifier = cursorModifier as CursorModifier;
+                    if ((modifier.CursorParameters != null) && (modifier.CursorParameters.Length > 0))
                     {
-                        SetAnimatorParameter(param);
+                        OnCursorStateChange(CursorStateEnum.Contextual);
+                        foreach (var param in modifier.CursorParameters)
+                        {
+                            SetAnimatorParameter(param);
+                        }
                     }
+                }
+                else
+                {
+                    OnCursorStateChange(CursorStateEnum.None);
                 }
             }
             else
--- a/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
@@ -306,9 +306,27 @@ namespace HoloToolkit.Unity.InputModule
         /// Override function when a new modifier is found or no modifier is valid
         /// </summary>
         /// <param name="modifier"></param>
-        protected virtual void OnActiveModifier(ICursorModifier modifier)
+        [System.Obsolete("Use OnActiveModifier(ICursorModifier modifier) instead.")]
+        protected virtual void OnActiveModifier(CursorModifier modifier)
         {
             TargetedCursorModifier = modifier;
+        }
+
+        /// <summary>
+        /// Override function when a new modifier is found or no modifier is valid
+        /// </summary>
+        /// <param name="modifier"></param>
+        protected virtual void OnActiveModifier(ICursorModifier modifier)
+        {
+            CursorModifier cursorModifier = modifier as CursorModifier;
+            if (cursorModifier != null)
+            {
+                OnActiveModifier(cursorModifier);
+            }
+            else
+            {
+                TargetedCursorModifier = modifier;
+            }
         }
 
         /// <summary>
--- a/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
@@ -294,9 +294,9 @@ namespace HoloToolkit.Unity.InputModule
             {
                 TargetedObject = newFocusedObject;
 
-                CursorModifier newModifier = (newFocusedObject == null)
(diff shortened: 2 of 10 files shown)


shubhamarora28:
@StephenHodgson Brought back the old method and marked it obsolete.

Regarding the test scripts and example scene, MRTK does not have any example of a custom cursor modifier which is why I created the scene. It has three game objects, one without any modifier, one with CursorModifier and one with a custom cursor modifier(TestCursorModifier).

Being a consumer of MRTK, I myself faced some issues while implementing my own cursor related functionality. I think the example scene will be helpful for some people.

Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
```diff
@@ -306,11 +306,29 @@ protected virtual void OnPointerSpecificFocusChanged(IPointingSource pointer, Ga
         /// Override function when a new modifier is found or no modifier is valid
         /// </summary>
         /// <param name="modifier"></param>
+        [System.Obsolete("Use OnActiveModifier(ICursorModifier modifier) instead.")]
         protected virtual void OnActiveModifier(CursorModifier modifier)
         {
             TargetedCursorModifier = modifier;
```
StephenHodgson: Shouldn't this now be calling the new method with the interface parameter?

Assets/HoloToolkit/Input/Scripts/Cursor/AnimatedCursor.cs
```diff
@@ -69,20 +69,28 @@ public override void OnInputDisabled()
         /// Override to set the cursor animation trigger
         /// </summary>
         /// <param name="modifier"></param>
-        protected override void OnActiveModifier(CursorModifier modifier)
```
StephenHodgson: Let's keep the old method and use it to call the method with the interface implementation and mark the old one obsolete.

davidkline-ms:
the july (june18_dev) branch is in stabilization, re-targeting this PR to the htk_development branch

Assets/HoloToolkit/Input/Scripts/Cursor/AnimatedCursor.cs
shubhamarora28: In line with the comment I have made above, if we keep the old method, it'll receive null as a parameter when gaze is on the game object with any other implementation of ICursorModifier(e.g., TestCursorModifier) and hence, the new method with the interface implementation will also get null.

Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
shubhamarora28: Following is the OnPointerSpecificFocusChanged method:
onpointerspecificfocuschanged

Here, if I call the old OnActiveModifier(CursorModifier modifier) method instead of the new one with interface parameter, I need to cast the ICursorModifier object to CursorModifer which will return null in all the cases when newFocusedObject has an implementation of ICursorModifier which is not CursorModifier.
Hence, the old OnActiveModifier(CursorModifier modifier) method will receive null as a parameter and pass null to the new implementation of the method.

In order to avoid this issue, I am calling the new OnActiveModifier with the interface parameter from OnPointerSpecificFocusChanged since ICursorModifier is the parent here and from inside of this method, I am checking if this implementation of ICursorModifier is CursorModifier, only then I call the old OnActiveModifier.
new_onactivemodifier


This keeps the old functionality intact and will not give any errors or produce any unexpected behavior for those users who have custom classes inherited from these.

shubhamarora28:
@StephenHodgson Updated TestCursorModifier to not throw `NotImplementedExceptions`.

Assets/HoloToolkit/Boundary/Scripts/BoundaryManager.cs
```diff
@@ -215,7 +215,7 @@ public Vector3[] TryGetBoundaryRectanglePoints()
             var positions = new Vector3[points2d.Length];
             for (int i = 0; i < points2d.Length; ++i)
             {
-                positions[i] = new Vector3(points2d[i].x, boundaryFloor, points2d[i].y);
+                positions[i] = CameraCache.Main.transform.parent.TransformPoint(new Vector3(points2d[i].x, boundaryFloor, points2d[i].y));
```
StephenHodgson: This change isn't relevant to custom cursors. Please, either make a new PR and revert this change, or add it to the description.

shubhamarora28: This change was made by @keveleigh.

StephenHodgson: Oh, I see, you merged from a different branch into this one?

shubhamarora28: Yes. I took june18_dev branch as base.

shubhamarora28: Sure. Will do that.

StephenHodgson: Thanks.

shubhamarora28: The changes made in #2518 are already there in my branch.

Assets/HoloToolkit/Input/Scripts/Cursor/Cursor.cs
shubhamarora28: @StephenHodgson Are we good with this?

SimonDarksideJ:
Can we get an update on this @keveleigh / @StephenHodgson as it's getting stale.

wiwei:
Closing issues older than 180 days. If this is still an issue, please reactivate with recent information.

Issue #2478, opened 2018-07-22
vNext Task: Performance review - ALPHA


SimonDarksideJ:
## Overview
Prior to the Alpha cutoff, we need a final Performance / review check on each Alpha target platform

## Requires
Performance Review on UWP platform
Performance Review on OPENVR platform

## Acceptance Criteria
- [x] No perceptible lag through input system on UWP
- [x] No perceptible lag through input system on OPENVR
- [x] No unnecessary logging
- [x] Any remaining performance leaks documented and tasks created for Beta for review

davidkline-ms:
I am seeing the lag that @SimonDarksideJ mentions when running via OpenVR. Looking into it.

PR #2501, opened 2018-07-27
fix controller lag on openvr

davidkline-ms:
Overview
---
When running on OpenVR (either a Vive or Windows Mixed Reality via the Steam driver), controllers had hitches and lag. Overall performance was fine (frame rate 90fps).

When the controller data was being updated, the new reading was being transformed based on the camera position and then stored as the previous reading. The current reading was being stored untransformed. It appears that when the glitch/lag was occurring, previous and current *may* have been evaluating as equivalent (floating point error).

Changes
---
Change the OpenVR UpdateControllerData and UpdatePointerData to more closely match the Windows Mixed Reality implementation.

- Fixes: #2478 .


SimonDarksideJ:
Does this change actually improve anything?
This was originally how it was layed out and the lag still existed.

We broke it up between controller and pointer to better aling with how the rest of the framework ran.

Thoughts Stephen?

keveleigh:
@SimonDarksideJ

https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/_Core/Devices/OpenVR/GenericOpenVRController.cs#L199

We set the transformed new position and rotation inside currentControllerPose, then save off currentControllerPose as the "last" pose on line 209 (which, it's now not the last pose). If the transformed poses are unchanged (perhaps floating point math sometimes will stray enough to send events sometimes) from what we set on lines 210 and 211, we wouldn't send a pose updated event.

Moving line 209 before line 185 fixes this, so we save off the last pose before we touch the currentControllerPose variable.

SimonDarksideJ:
So does this cure or improve the performance jitter

keveleigh:
@SimonDarksideJ 

>So does this cure or improve the performance jitter

Cures it completely.

PR #2512, opened 2018-07-29
vNEXT - Misc Bug Fixes

StephenHodgson:
- Updated some issues with setting world space on canvas and no MR Manager in scene.
- Updated Input System initialization to be a bit smarter.
- Updated usages of `ActiveProfile` null checks  to `HasActiveProfile`
- Updated usages of individual feature profile checks to `Enable<FeatureName>`

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Inspectors/CanvasEditorExtension.cs
+++ b/Assets/MixedRealityToolkit/_Core/Inspectors/CanvasEditorExtension.cs
@@ -8,7 +8,7 @@ using Microsoft.MixedReality.Toolkit.Internal.Utilities;
 using UnityEditor;
 using UnityEngine;
 
-namespace Microsoft.MixedReality.Toolkit.Inspectors
+namespace Microsoft.MixedReality.Toolkit.Inspectors.Utilities
 {
     /// <summary>
     /// Helper class to assign the UIRaycastCamera when creating a new canvas object and assigning the world space render mode.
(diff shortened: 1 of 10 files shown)


Assets/MixedRealityToolkit/_Core/Inspectors/CanvasEditorExtension.cs
```diff
@@ -27,17 +27,27 @@ private void OnEnable()
...
             {
                 inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>();
+                CheckCanvasSettings();
             }
         }
 
         public override void OnInspectorGUI()
         {
-            if (inputSystem == null)
+            if (!MixedRealityManager.IsInitialized || !MixedRealityManager.Instance.ActiveProfile.EnableInputSystem)
```
SimonDarksideJ: What if the ActiveProfile is null?
Shouldn't happen, but if the user deletes their current profile, it will be null

StephenHodgson:
> looks good. just handle ActiveProfile == null

Literally just pushed that change 😆

StephenHodgson:
We already do something similar in `HasActiveProfile`:

```
/// <summary>
/// Checks if there is a valid instance of the MixedRealityManager, then checks if there
/// is there a valid Active Profile on this manager.
/// </summary>
public static bool HasActiveProfile
{
    get
    {
        if (!IsInitialized)
        {
            return false;
        }

        return Instance.activeProfile != null;
    }
}
```

I'm not going to change `IsInitialized`, but instead use `HasActiveProfile` to check. I'll also update the usages in the project of `ActiveProfile == null` and `ActiveProfile != null` accordingly.

StephenHodgson:
Okay guys that's the last of the changes.

I also did a quick pass and updated any usages of Profile != null and Profile == null in the whole project.

PR #2519, opened 2018-07-30
A bit of cleanup to Managers

StephenHodgson:
Overview
---
- Fixed an issue with assembly reloading not disabling/destroying managers properly.
- Fixed issues starting/stopping playing in the editor and the MR Manager state would get corrupted.
- Updated the Mixed Reality Manager's `Enable<Feature>` to be `Is<Feature>Enabed`
    - Double checked to also make sure these bools check if the profile is not null.
- Updated Handedness flags because `Both` and `Any` were the same flag value of 3.  `Any` now properly includes `Other`.
- Sorted the order of the configuration profiles in the creation context menu.
- Removed unnecessary ping object in input system initialization.
- Fixed an issue with the Input Action property drawer.
- Updated Boundary Manager to properly initialize values in it's initialization method instead of the constructor.
- Updated Boundary Viz Demo scene 
    - Added default cursor
    - Fixed a null ref when no HMD detected


Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
```diff
@@ -46,7 +46,7 @@ protected IMixedRealityInputSystem InputSystem
         {
             get
             {
-                if (inputSystem == null && MixedRealityManager.Instance.ActiveProfile.EnableInputSystem)
+                if (inputSystem == null && MixedRealityManager.Instance.ActiveProfile.IsInputSystemEnabled)
```
SimonDarksideJ: Needs to also have a check for "MRManager.HasActiveProfile"

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
```diff
@@ -84,7 +84,7 @@ protected IMixedRealityInputSystem InputSystem
         /// <param name="controllerType"></param>
         public void SetupConfiguration(Type controllerType)
         {
-            if (MixedRealityManager.Instance.ActiveProfile.EnableControllerMapping)
+            if (MixedRealityManager.Instance.ActiveProfile.IsControllerMappingEnabled)
```
SimonDarksideJ: Needs to also have a check for "MRManager.HasActiveProfile"

Should this be a standard?

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/InputActionPropertyDrawer.cs
```diff
@@ -19,16 +19,17 @@ public class InputActionPropertyDrawer : PropertyDrawer
 
         public override void OnGUI(Rect rect, SerializedProperty property, GUIContent content)
         {
-            if (!MixedRealityManager.IsInitialized)
+            if (!MixedRealityManager.ConfirmInitialized() || !MixedRealityManager.HasActiveProfile)
```
SimonDarksideJ: We don't need both, as ConfirmInitialised is called within HasActiveProfile

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
StephenHodgson: Still trying to figure that out.

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
StephenHodgson: If  we didn't have an active profile, how would we ever have created a controller?

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
StephenHodgson: I don't think we'll need to call `HasActiveProfile` in classes we know is created by a profile

StephenHodgson:
Found a few more issues I want to address.

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealityConfigurationProfileInspector.cs
```diff
@@ -149,22 +165,20 @@ public override void OnInspectorGUI()
             {
                 EditorGUILayout.PropertyField(inputSystemType);
                 RenderProfile(inputActionsProfile);
+                //RenderProfile(pointerProfile);
```
davidkline-ms: why is this commented out?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Handedness.cs
```diff
@@ -16,27 +16,27 @@ public enum Handedness : byte
...
         /// The controller is identified as being either left and/or right handed.
         /// </summary>
         Both = Left | Right,
         /// <summary>
-        /// Global catchall, used to map actions to any controller (provided the controller supports it)
-        /// *Note, by default the specific hand actions will override settings mapped as both
+        /// Reserved, for systems that provide alternate hand state.
         /// </summary>
-        Any = None | Left | Right,
+        Other = 4,
```
davidkline-ms: dont care for this. what is an Other hand state? doesnt feel needed.... except to differentiate any from both.

both may actually be something very specific in the future (might not be decomposable into left and right parts).

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealityConfigurationProfileInspector.cs
StephenHodgson: Cause I didn't add it yet.  (Must have been left over when I ported it over) I'll remove it.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Handedness.cs
StephenHodgson: Well right now `Any` and `Both` are `3`, which is a problem

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/ExperienceScale.cs
```diff
@@ -22,7 +22,7 @@ public enum ExperienceScale
         /// </summary>
         Standing,
         /// <summary>
-        /// An experience designed to support movement thoughtout a room. The coordinate system origin is at floor level.
+        /// An experience designed to support movement thought out a room. The coordinate system origin is at floor level.
```
keveleigh: throughout

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -97,64 +97,61 @@ private void InitializeInternal()
...
 #if UNITY_EDITOR
-
-            var eventSystems = UnityEngine.Object.FindObjectsOfType<EventSystem>();
-            var standaloneInputModules = UnityEngine.Object.FindObjectsOfType<StandaloneInputModule>();
-
-            CameraCache.Main.transform.position = Vector3.zero;
-            CameraCache.Main.transform.rotation = Quaternion.identity;
-
-            if (eventSystems.Length == 0)
+            if (!UnityEditor.EditorApplication.isPlaying)
```
keveleigh: Does InitializeInternal get called when the editor isn't playing?

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -97,64 +97,61 @@ private void InitializeInternal()
...
+                    bool raiseWarning;
 
-                if (raiseWarning)
-                {
-                    Debug.LogWarning("Found an existing event system in your scene. The Mixed Reality Input System requires only one, and must be found on the focus provider.");
-                }
-            }
+                    if (eventSystems.Length == 1)
+                    {
+                        raiseWarning = eventSystems[0].gameObject != focusProvider.gameObject;
```
keveleigh: Why does the event system need to be on the same GameObject as the `FocusProvider`? None of the existing example scenes are set up this way.

StephenHodgson: Because the existing example scenes were setup incorrectly.

It needs to be there because the FocusProvider depends on it.

StephenHodgson: I think the event system is added to the UIRaycastCamera that is found under the camera.

I'll update the debug log to reflect that fact.

StephenHodgson: Mainly this is so we don't have to call `FindObjectOfType<T>` anywhere. We keep track of the event system and input module.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/ExperienceScale.cs
StephenHodgson: huh, I'm getting a type warning in my spell checker with `throughout`

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
keveleigh: Where does it get called while the editor isn't playing?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/ExperienceScale.cs
```diff
@@ -22,7 +22,7 @@ public enum ExperienceScale
         /// </summary>
         Standing,
         /// <summary>
-        /// An experience designed to support movement thought out a room. The coordinate system origin is at floor level.
+        /// An experience designed to support movement thoughtout a room. The coordinate system origin is at floor level.
```
keveleigh: thoughtout -> throughout

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: > Where does it get called while the editor isn't playing? 

Any time you access the instance.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/ExperienceScale.cs
keveleigh: through out is fine too if that plays nicer with spell check, just not "thought out"

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Handedness.cs
davidkline-ms: true. maybe define Both as a discrete value and Any as Left | Right | Both

StephenHodgson: Then what's the point of using the bit flags if we assign both as a discrete value?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/ExperienceScale.cs
StephenHodgson: I get it now. Sorry 😅

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
keveleigh: > I think the event system is added to the UIRaycastCamera that is found under the camera.

Right, but we're checking against a different GameObject here. Should this be `focusProvider.UIRaycastCamera.gameObject`, to match the `FocusProvider.UIRaycastCamera.gameObject.EnsureComponent<EventSystem>()` line above? Otherwise, we're adding to one GameObject and checking against a different one.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Handedness.cs
SimonDarksideJ: Any should also include None, as it relates simply to ANY input.
As for both, it relates to "Both hands" effectively.

Other is a special case for new / custom controllers that we simply don't have a definition for, but still allow to function and participate in the framework until it evolves to include it.

Other could be included in Any, but I don't think so due to it's custom nature, it needs to be treated in a special case.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -87,18 +77,17 @@ public MixedRealityCameraProfile CameraProfile
...
         /// Enable and configure the Input System component for the Mixed Reality Toolkit
         /// </summary>
-        public bool EnableInputSystem
+        public bool IsInputSystemEnabled
         {
             get
             {
-                return inputActionsProfile != null &&
-                       enableInputSystem;
+                return inputActionsProfile != null && enableInputSystem;
```
keveleigh: I think we need to add `inputSystemType.Type != null` here.
Otherwise: 
image

Matches line 183 and the boundary manager checks.

StephenHodgson: We should update them all then. Cause the boundary might be the only done doing it correctly

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -97,64 +97,61 @@ private void InitializeInternal()
...
+                var standaloneInputModules = UnityEngine.Object.FindObjectsOfType<StandaloneInputModule>();
 
-                if (eventSystems.Length == 1)
+                CameraCache.Main.transform.position = Vector3.zero;
+                CameraCache.Main.transform.rotation = Quaternion.identity;
+
+                if (eventSystems.Length == 0)
                 {
-                    raiseWarning = eventSystems[0].gameObject != focusProvider.gameObject;
+                    FocusProvider.UIRaycastCamera.gameObject.EnsureComponent<EventSystem>();
```
keveleigh: Do we have guidelines on using the public vs the private inside the same class?
I feel like I've seen @SimonDarksideJ comment about using the private.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
StephenHodgson: Oh we only check the type for two

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: How does that apply here?

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/ExperienceScale.cs
keveleigh: Haha all good!

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
keveleigh: Yeah, it looks like those are the only two with `SystemTypes`

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
keveleigh: `FocusProvider` vs `focusProvider`.

Your previous commit changed which one was being used.

image

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Handedness.cs
```diff
@@ -30,13 +30,13 @@ public enum Handedness : byte
...
-        /// <summary>
         /// Global catchall, used to map actions to any controller (provided the controller supports it)
         /// </summary>
         /// <remarks>Note, by default the specific hand actions will override settings mapped as both</remarks>
-        Any = None | Other | Left | Right,
+        Any = None | Left | Right,
+        /// <summary>
+        /// Reserved, for systems that provide alternate hand state.
+        /// </summary>
+        Other = 4,
```
davidkline-ms: Still don't think we need an Other. Both also seems superfluous until we encounter a platform that has a controller that requires both hands.

Let's just go with None, Left, Right and Any

davidkline-ms: BTW, Or-ing in a value of 0 still makes Both == Any

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: Depends on if the property does anything extra when accessing it. I can revert if you'd like.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Handedness.cs
StephenHodgson: @SimonDarksideJ ?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Utilities/InteractibleHighlight.cs
```diff
@@ -18,15 +18,15 @@ public class InteractableHighlight : BaseFocusHandler
         [Flags]
         public enum HighlightedMaterialStyle
         {
-            None = 0,
+            None = 0 << 0,
             /// <summary>
             /// A highlight to indicate focus.
             /// </summary>
-            Highlight = 1,
+            Highlight = 1 << 0,
```
davidkline-ms: this looks really weird... shouldn't it be
Highlight = 1 << 0,
Overlay = 1 << 1,

StephenHodgson: Isn't that what it is?

Issue #2384, opened 2018-07-03
Keyboard does not warn that XAML is unsupported.


SujeathPareddy:
## Overview
Keyboard.cs does not warn that XAML is not supported, instead throws a hard to diagnose error message. I believe this is connected to #2169 and the resulting fix on commit f56fa76da80.

Essentially, f56fa76da80 tries to force cast -1 to an Enum. At some indeterminate time, Unity tries to cast the Enum to Int32 inside KernelBase. On D3D, nothing bad happens. On XAML builds, this causes an OverflowException which cannot be examined in the debugger and can send newcomers (like me) down a rabbit hole.

From the comments on #1940, I understand that XAML is not supported. It would be nice to receive a helpful message such as: 
>XAML builds are not supported with MRTK. Please switch to D3D inside Build Settings.

I think some helpful messages are already part of the language of MRTK, like this one in InputManager:  
> Cursor hasn't been explicitly set on "InputManager.SimpleSinglePointerSelector". We'll search for a cursor in the hierarchy, but that comes with a performance cost, so it would be best if you explicitly set the cursor.

## Expected Behavior
Print helpful message about XAML not being supported.

## Actual Behavior
Inside **[Keyboard->keyboard_Background->search->InputField=>SliderInputField]**
```
OverflowException: Value was either too large or too small for a UInt32.
   at System.Convert.ThrowUInt32OverflowException()
   at System.Convert.ToUInt32(Int32 value)
   at System.Int32.System.IConvertible.ToUInt32(IFormatProvider provider)
   at System.Convert.ToUInt32(Object value, IFormatProvider provider)
   at System.Enum.System.IConvertible.ToUInt32(IFormatProvider provider)
   at System.Convert.ToUInt32(Object value)
   at UnityEngine.TouchScreenKeyboard..ctor(String text, TouchScreenKeyboardType keyboardType, Boolean autocorrection, Boolean multiline, Boolean secure, Boolean alert, String textPlaceholder)
   at UnityEngine.TouchScreenKeyboard.Open(String text, TouchScreenKeyboardType keyboardType, Boolean autocorrection, Boolean multiline)
   at UnityEngine.UI.InputField.ActivateInputFieldInternal()
   at UnityEngine.UI.InputField.LateUpdate()
   at UnityEngine.UI.InputField.$Invoke72LateUpdate(Int64 instance, Int64* args)
[...]

PR #2522, opened 2018-07-30
Minor keyboard fixes

keveleigh:
Overview
---
Instead of trying to case a negative int and failing on XAML, now we cast a large, still invalid, number.

Also, disables the Toolkit keyboard on platforms that support a system keyboard.

Changes
---
- Fixes: #2384

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/Keyboard.cs
+++ b/Assets/HoloToolkit/UX/Scripts/Keyboard.cs
@@ -253,7 +253,7 @@ namespace HoloToolkit.UI.Keyboard
 
             // Setting the keyboardType to an undefined TouchScreenKeyboardType,
             // which prevents the MRTK keyboard from triggering the system keyboard itself.
-            InputField.keyboardType = (TouchScreenKeyboardType)(-1);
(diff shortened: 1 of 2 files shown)


Assets/HoloToolkit/UX/Scripts/Keyboard.cs
```diff
@@ -253,7 +253,7 @@ protected override void Awake()
 
             // Setting the keyboardType to an undefined TouchScreenKeyboardType,
             // which prevents the MRTK keyboard from triggering the system keyboard itself.
-            InputField.keyboardType = (TouchScreenKeyboardType)(-1);
+            InputField.keyboardType = (TouchScreenKeyboardType)(int.MaxValue);
```
fieldsJacksonG: MinValue?  (technically you might be able to connect a *lot* of keyboards, but you can't connect negative keyboards!)

keveleigh: This was changed because, when building for UWP XAML, it tries to convert the value into a uint and throws an error. See: #2384

Issue #2524, opened 2018-07-31
OnButtonClicked is triggered twice on an AirTap


ei2kpi:
## Overview
OnButtonClicked is called twice since DoButtonPressed is called OnInputDown and OnInputClicked in [Button.cs](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs)

## Expected Behavior
OnButtonClicked is called only when OnInputClicked is received by the Button

## Actual Behavior
OnButtonClicked is called twice since DoButtonPressed is called OnInputDown and OnInputClicked

## Steps to reproduce
Register for OnButtonClicked event on any button and then tap on the button. OnButtonClicked is triggered twice

## Unity Editor Version
2017.4.4f1

## Mixed Reality Toolkit Release Version
2017.4.0.0

PR #2525, opened 2018-07-31
OnButtonClicked event is called twice on every tap

ei2kpi:
OnButtonClicked is called twice since DoButtonPressed is called OnInputDown and OnInputClicked. Splitting out Clicked into it's own DoButtonClicked function prevents this from happening.

Overview
---

Changes
---
- Fixes: #2524  .


Assets/HoloToolkit/UX/Scripts/Buttons/Button.cs
```diff
@@ -292,6 +292,25 @@ public void OnSourceLost(SourceStateEventData eventData)
         }
         #endregion
 
+        /// <summary>
+        /// Called when button is "Clicked".
+        /// </summary>
+        protected void DoButtonClicked(bool bRelease = false)
```
StephenHodgson: virtual?

ei2kpi:
Closing since this was already fixed for next release...

PR #2533, opened 2018-08-01
vNEXT Basic Controller Pointers.

### LINKED ISSUE #2541: GlobalListeners list get's mutated as we iterate the list.
## Overview
Found a bug where the list of global listeners isn't always guaranteed to be executed if that list changes during the execution of that event.

## Expected Behavior
Global Listener Events should always be executed.

## Actual Behavior
If a GameObject registers or un registers during the ExecuteEvent loop, some objects later in the list will be skipped because the list index has changed.

MRTK vNEXT and Legacy HTK

StephenHodgson:
## Feature port of Controller Pointers

Added the ability to easily setup pointers via Pointer Profile:
image

## Pointer profile:
- Assigning a pointer to a specific controller using the controller type.
    - If no controller type is assigned, then the pointer is assigned to every controller that supports spatial pointers.
- Assigning a pointer to specific hands
    - Left & Right for assigning unique pointers per hand
    - Both for assigning a pointer to both hands
        - **Doesn't support** controllers that have no specific handedness
    - Any for assigning a pointer to both hands
        - **Supports** controllers that support spatial pointers but don't have a specific handedness (i.e. HoloLens hands & clicker)

## How to create a new pointer:
1. create an empty game object in the scene view and name it
2. add a `SimpleLineDataProvider` to the new game object
3. add a` MixedRealityLineRenderer` to the game object
4. add a `LinePointer` to the game object.
5. customize all the components to your hearts desire
6. drag the game object into the project window to make it a prefab
7. delete the pointer from your scene
8. select or create a pointer profile
9. add a new pointer item in the pointer profile
10. drag your new pointer prefab into the game object field
11. select the controller type you want to use the prefab for
    - Keeping the type `None` will assign it to every controller type
12. choose the handedness of the pointer
13. save and play in editor

TODO
---
- [x] Add option for selecting controller pointer vs grip attachment.
- [x] Write up steps for pointer prefab creation

Additional Changes
---
- updated default cursor and cursor scripts
- updated input action property drawer to properly assign the axis constraint and description
- FIxed #2541
- Updated Mixed Reality Manager to disable and destroy all managers when application wants to quit.


Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -48,90 +41,77 @@ public bool DisableChildren
         /// <summary>
         /// Is the controller this Synchronizer is registered to currently tracked?
         /// </summary>
-        public bool IsTracked { get; private set; } = false;
+        public bool IsTracked { get; protected set; } = false;
```
davidkline-ms: shouldn't this return (TrackingState == TrackingState.Tracked)?

davidkline-ms: also, do we even need this with TrackingState being public?

Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -48,90 +41,77 @@ public bool DisableChildren
...
+
+        public TrackingState TrackingState { get; protected set; } = TrackingState.NotTracked;
```
davidkline-ms: summary tag

Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -48,90 +41,77 @@ public bool DisableChildren
...
 
-        private TrackingState lastTrackingState = TrackingState.NotTracked;
+        public virtual IMixedRealityController Controller
```
davidkline-ms: summary tag

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
```diff
@@ -16,77 +19,116 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
     /// <summary>
     /// Base Pointer class for pointers that exist in the scene as GameObjects.
     /// </summary>
-    public abstract class BaseControllerPointer : ControllerPoseSynchronizer, IMixedRealityInputHandler, IMixedRealityPointer
+    [DisallowMultipleComponent]
```
davidkline-ms: Does this mean we cannot have a pointer for each controller / hand in the scene?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
```diff
@@ -16,77 +19,116 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
...
+            set
+            {
+                pointerOrientation = Mathf.Clamp(value, 0f, 360f);
+            }
+        }
 
         [SerializeField]
-        [Range(0.5f, 50f)]
-        private float pointerExtent = 2f;
+        [Tooltip("Should the pointers position be driven from the source pose or from input handler?")]
```
davidkline-ms: pointer's

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
```diff
@@ -109,52 +151,29 @@ protected virtual void Start()
...
-        /// </summary>
-        public virtual void OnSelectPressed()
-        {
-            SelectPressed = true;
-            SelectPressedOnce = true;
-        }
+        IMixedRealityInputSystem IMixedRealityPointer.InputSystem => InputSystem;
 
-        public virtual void OnSelectReleased()
+        public override IMixedRealityController Controller
```
davidkline-ms: summary tag

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/LinePointer.cs
```diff
@@ -0,0 +1,198 @@
...
+        [SerializeField]
+        private BaseMixedRealityLineDataProvider lineBase;
+
+        public BaseMixedRealityLineDataProvider LineBase => lineBase;
+
+        [SerializeField]
+        [Tooltip("If no line renderers are specified, this array will be auto-populated on startup.")]
+        private BaseMixedRealityLineRenderer[] lineRenderers;
+
+        public BaseMixedRealityLineRenderer[] LineRenderers;
```
davidkline-ms: … LineRenderers => lineRenderers

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Pointers/BaseControllerPointerInspector.cs
```diff
@@ -0,0 +1,64 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.﻿
+
+using Microsoft.MixedReality.Toolkit.SDK.Inspectors.Input.Handlers;
+using Microsoft.MixedReality.Toolkit.SDK.UX.Pointers;
+using UnityEditor;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Inspectors.UX.Pointers
+{
+    [CustomEditor(typeof(BaseControllerPointer))]
```
davidkline-ms: summary tag

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Pointers/LinePointerInspector.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.﻿
+
+using Microsoft.MixedReality.Toolkit.SDK.UX.Pointers;
+using UnityEditor;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Inspectors.UX.Pointers
+{
+    [CustomEditor(typeof(LinePointer))]
+    public class LinePointerInspector : BaseControllerPointerInspector
```
davidkline-ms: summary tag

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityPointerProfile.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Configuration profile settings for setting up controller pointers.
+    /// </summary>
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Mixed Reality Pointer Profile", fileName = "MixedRealityInputPointerProfile", order = 3)]
+    public class MixedRealityPointerProfile : ScriptableObject
+    {
+        [SerializeField]
+        private PointerOption[] pointerOptions = new PointerOption[0];
+
+        public PointerOption[] PointerOptions => pointerOptions;
```
davidkline-ms: summary tag

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/LinePointer.cs
```diff
@@ -0,0 +1,198 @@
...
+        protected Gradient LineColorLockFocus = new Gradient();
+
+        [Range(2, 100)]
+        [SerializeField]
+        protected int LineCastResolution = 25;
+
+        [SerializeField]
+        private BaseMixedRealityLineDataProvider lineBase;
+
+        public BaseMixedRealityLineDataProvider LineBase => lineBase;
```
davidkline-ms: summary tags for the public properties

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityPointerProfileInspector.cs
```diff
@@ -0,0 +1,101 @@
...
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem;
+using UnityEditor;
+using UnityEditorInternal;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Inspectors
+{
+    [CustomEditor(typeof(MixedRealityPointerProfile))]
+    public class MixedRealityPointerProfileInspector : MixedRealityBaseConfigurationProfileInspector
```
davidkline-ms: summary tag

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
StephenHodgson: Do, it means we can only have a simple controller component on a single `GameObject`.

We can have as many pointers for as many hands as we'd like.

Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
StephenHodgson: I made it protected for the time being.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -115,12 +111,48 @@ private WindowsMixedRealityController GetOrAddController(InteractionSourceState
                     throw new ArgumentOutOfRangeException();
             }
 
-            var inputSource = InputSystem?.RequestNewGenericInputSource($"Mixed Reality Controller {controllingHand}");
+            var pointers = new List<IMixedRealityPointer>();
+
+            if (MixedRealityManager.HasActiveProfile &&
+                MixedRealityManager.Instance.ActiveProfile.IsInputSystemEnabled &&
+                MixedRealityManager.Instance.ActiveProfile.PointerProfile != null)
```
StephenHodgson: I should probably also check if the controller supports pointing here.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
```diff
@@ -179,35 +181,71 @@ private GenericOpenVRController GetOrAddController(XRNodeState xrNodeState)
...
+                default:
+                    Debug.LogError($"Unsupported controller type detected.");
+                    return null;
             }
 
+            var pointers = new List<IMixedRealityPointer>();
+
+            if (MixedRealityManager.HasActiveProfile &&
+                MixedRealityManager.Instance.ActiveProfile.IsInputSystemEnabled &&
+                MixedRealityManager.Instance.ActiveProfile.PointerProfile != null)
```
StephenHodgson: I suppose we're assuming that all Open VR Controllers support pointing?

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Pointers/BaseControllerPointerInspector.cs
davidkline-ms: fair enough

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
davidkline-ms: good question. I have limited experience with their controllers

StephenHodgson: @SimonDarksideJ would know, but he's on holiday

StephenHodgson:
@davidkline-ms sorry just a small update to make sure the WMR controller supports pointing. should be good now.

davidkline-ms:
> @davidkline-ms sorry just a small update

No apology required :)

Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -48,90 +41,83 @@ public bool DisableChildren
...
+                {
+                    gameObject.SetChildrenActive(true);
+                }
+            }
         }
 
         /// <inheritdoc />
         public virtual void OnSourceLost(SourceStateEventData eventData)
         {
-            if (eventData.Controller.ControllerHandedness == Handedness)
```
StephenHodgson: I should be checking if the source id is the same here.

StephenHodgson:
Found an issue when working with multiple pointers. Fix in the worx

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -115,12 +111,49 @@ private WindowsMixedRealityController GetOrAddController(InteractionSourceState
                     throw new ArgumentOutOfRangeException();
             }
 
-            var inputSource = InputSystem?.RequestNewGenericInputSource($"Mixed Reality Controller {controllingHand}");
+            var pointers = new List<IMixedRealityPointer>();
```
SimonDarksideJ: Question, could this be part of the inputSource request?
rather than put this in every device we create, effectively copying the same code.
Or failing that, move this to a static function to return configured pointers for a given controller / type and hand

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
```diff
@@ -140,12 +143,10 @@ private static SupportedControllerType CurrentControllerType
                         case "Vive Knuckles - Left":                                            // TODO: Yet to test
                         case "Vive Knuckles- Right":                                            // TODO: Yet to test
                             return SupportedControllerType.ViveKnuckles;
-                        default:
-                            break;
                     }
                 }
 
-                return SupportedControllerType.GenericOpenVR;
+                return SupportedControllerType.None;
```
SimonDarksideJ: why change this to None, it makes no sense?  The generic is the Fallback device, so any new OpenVR controller will be supported by default using the default options until it's added.

Please revert this.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
```diff
@@ -179,35 +181,71 @@ private GenericOpenVRController GetOrAddController(XRNodeState xrNodeState)
...
-            IMixedRealityInputSource inputSource = InputSystem?.RequestNewGenericInputSource($"{CurrentControllerType} Controller {controllingHand}");
-
-            GenericOpenVRController detectedController = null;
+            Type controllerType;
 
-            // Initialize the controller base on the detected type
             switch (CurrentControllerType)
             {
                 case SupportedControllerType.GenericOpenVR:
-                    detectedController = new GenericOpenVRController(TrackingState.NotTracked, controllingHand, inputSource);
```
SimonDarksideJ: why have you changed this, you have now broken the specific device implementation, meaning there is only one controller.
As discussed previously, we need the separate definitions to allow for variations between the controllers as required, input mappings, SDK models (when we support them) and others.

Please revert this back

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
```diff
@@ -179,35 +181,71 @@ private GenericOpenVRController GetOrAddController(XRNodeState xrNodeState)
...
+                                pointers.Add(pointer);
+                            }
+                        }
+                    }
+                }
+            }
+
+            IMixedRealityInputSource inputSource = InputSystem?.RequestNewGenericInputSource($"{CurrentControllerType} Controller {controllingHand}", pointers.ToArray());
+
+            var detectedController = new GenericOpenVRController(TrackingState.NotTracked, controllingHand, inputSource);
```
SimonDarksideJ: Concerned by this change, as we are not creating each by their type.
Yes, we have generalized the input, but other things as we move forward will be specific.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
SimonDarksideJ: OpenVR controllers only support pointing, e.g. one pose.
Testing however, will prove whether each needs a different offset pose for forward.  
They don't have a grip pose, which I proposed we may have to simulate later.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityPointerProfileInspector.cs
SimonDarksideJ: We've agreed not to do documentation on inspectors for now, as they are fairly self explanatory.
Also, they are not part of the API, they are internal to unity.

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/InputActionPropertyDrawer.cs
```diff
@@ -19,7 +19,7 @@ public class InputActionPropertyDrawer : PropertyDrawer
 
         public override void OnGUI(Rect rect, SerializedProperty property, GUIContent content)
         {
-            if (!MixedRealityManager.HasActiveProfile)
+            if (!MixedRealityManager.IsInitialized || !MixedRealityManager.HasActiveProfile)
```
SimonDarksideJ: just use HasActiveProfile, as IsInitialized is checked within HasActiveProfile

StephenHodgson:
Fixed: #2541

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/InputActionPropertyDrawer.cs
StephenHodgson: I tried that but still got a null ref on the MR Manager so I added it back.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
StephenHodgson: Whoops, yeah this is a bad change. Sorry.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
StephenHodgson: Will do. I thought I was being clever only needing to get the type ahead of time.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
StephenHodgson: Bc any controller would be an open VR controller, not just open VR controller. 

I tried to get Xbox controller setup, and it came out as open vr.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
StephenHodgson: Possibly I'll see what I can do.

StephenHodgson:
> The default for OpenVR controllers is the Generic controller NOT none. Generic covers any new device added that we haven't added a specific implementation for yet.

Yeah but that also includes controller's that aren't open VR controllers.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
StephenHodgson: Doesn't look like we can embed this into the input source request. There's too many controller specific things that could end up happening while we register our pointers. \

First being that we need to know the type of controller.

Second, in WMR we can query if the input source pose supports pointing or not.

We would also need to know the handedness of the controller and check against the list of pointers in our pointer profile that match the handedness of our controllers.

We could make all those parameters that the input source request makes, but then I think that breaks the OOP nature of the setup, and makes the request do more than it's responsible for.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
StephenHodgson: So the nice thing is that pointers can choose to get their input data from the device's pose, or listen to input actions from that device, which makes choosing the source easy.

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/InputActionPropertyDrawer.cs
SimonDarksideJ: odd, might need to check that function then

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
SimonDarksideJ: then you need to add an Xbox controller class to handle it differently, not default none as that breaks all other new controllers we haven't defined

StephenHodgson: It doesn't break them, but just warns that they're unsupported.

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/InputActionPropertyDrawer.cs
StephenHodgson: I think it's bc the MR Manager is in the process of initializing and this get's raised from a different thread in the editor.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
StephenHodgson: I figured something out, I think you'll like the result.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
```diff
@@ -179,35 +181,62 @@ private GenericOpenVRController GetOrAddController(XRNodeState xrNodeState)
...
+                    controllerType = typeof(ViveKnucklesController);
+                    break;
+                case SupportedControllerType.OculusTouch:
+                    controllerType = typeof(OculusTouchController);
+                    break;
+                case SupportedControllerType.OculusRemote:
+                    controllerType = typeof(OculusRemoteController);
+                    break;
+                default:
+                    Debug.LogError($"Unsupported controller type detected.");
```
StephenHodgson: @SimonDarksideJ https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2533/files#r207372143

StephenHodgson:
Okay guys, I think I'm done messing with this PR. One last review is in order.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
keveleigh: WinMR headset with OpenVR:

image

Previously this correctly fell back to `GenericOpenVR`.

StephenHodgson: That's because it's not in the list, and appropriately shows that it's unsupported

StephenHodgson: Agreed, but then how do we get around fact that even non open VR controllers get added?

StephenHodgson: And it's not like we're hard coding anything, but adding the name of the controller to the list

keveleigh: >Agreed, but then how do we get around fact that even non open VR controllers get added?

I keep trying, but I've yet to see an Xbox controller trigger this code. Do you have repro steps?

>And it's not like we're hard coding anything, but adding the name of the controller to the list

Yes, a hard-coded list of strings.

StephenHodgson: I've got an idea about how to solve this.

In the controller configuration we can add a text field for people to put the name of the controller whenever they set it to generic, and update this switch to fall back to generic when a match is found. But we should definitely keep this line as none.

StephenHodgson: > Yes, a hard-coded list of strings

It doesn't have to be.

keveleigh: Where would we get the controller names from?

StephenHodgson: > In the controller configuration we can add a text field for people to put the name of the controller whenever they set it to generic, and update this switch to fall back to generic when a match is found. But we should definitely keep this line as none.

In short, from the user's controller configuration file.

StephenHodgson: @keveleigh I went a head and reverted this for now. We can pick it back up in beta.

Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs
```diff
@@ -74,6 +74,9 @@ protected override void SetPointInternal(int pointIndex, Vector3 point)
         {
             switch (pointIndex)
             {
+                case 0:
+                    endPoint.Position = point;
```
keveleigh: Should this be `startPoint.Position` to match https://github.com/StephenHodgson/MixedRealityToolkit-Unity/blob/c2e196261d24ed8774589ba3b073b963e4ab2388/Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs#L63?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -115,12 +110,20 @@ private WindowsMixedRealityController GetOrAddController(InteractionSourceState
                     throw new ArgumentOutOfRangeException();
             }
 
-            var inputSource = InputSystem?.RequestNewGenericInputSource($"Mixed Reality Controller {controllingHand}");
+            IMixedRealityPointer[] pointers = interactionSourceState.source.supportsPointing ? RequestPointers(typeof(WindowsMixedRealityController), controllingHand) : null;
+            var inputSource = InputSystem?.RequestNewGenericInputSource($"Mixed Reality Controller {controllingHand}", pointers);
             var detectedController = new WindowsMixedRealityController(TrackingState.NotTracked, controllingHand, inputSource);
+
             detectedController.SetupConfiguration(typeof(WindowsMixedRealityController));
-            detectedController.UpdateController(interactionSourceState);
```
keveleigh: Why don't we update the controller data here anymore? Now the data in a `SourceDetected` event will be empty, or is it properly filled in elsewhere?

Assets/MixedRealityToolkit/InputSystem/Pointers/GenericPointer.cs
```diff
@@ -92,13 +106,13 @@ public virtual void OnPreRaycast()
             Ray pointingRay;
             if (TryGetPointingRay(out pointingRay))
             {
-                Rays[0].CopyRay(pointingRay, (PointerExtent ?? InputSystem.FocusProvider.GlobalPointingExtent));
```
keveleigh: What happened to the `GlobalPointingExtent`? Would be nice to have an option to not have to set the same extent on every pointer.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
SimonDarksideJ: Update is already called in the GetOrAddCotroller, so this was in effect calling it twice.

Assets/MixedRealityToolkit/InputSystem/Pointers/GenericPointer.cs
StephenHodgson: It's still there. We set it when we create the pointer now.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
keveleigh: This _is_ `GetOrAddController`

Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs
StephenHodgson: This fixes a bug.

StephenHodgson:
> In addition to the comments, do I have to do anything to get this PR to work? The pointer spawns but never moves.

What platform are you on? I didn't light anything up for Open VR

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -130,7 +133,7 @@ private WindowsMixedRealityController GetOrAddController(InteractionSourceState
         /// <param name="interactionSourceState">Source State provided by the SDK to remove</param>
         private void RemoveController(InteractionSourceState interactionSourceState)
         {
-            var controller = GetOrAddController(interactionSourceState, false);
```
SimonDarksideJ: Why change/ remove this.  Had to add this as it caused issues on startup, unless you've resolve them?

Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs
keveleigh: What was the bug? Why would `case 0` set one thing (`endPoint`) but get a different thing (`startPoint`)?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
StephenHodgson: Haha, I think he meant after we call GetOrAddCotroller we usually update it too.

We were running into issues, and this fixes it.
Essentially we were updating the controller before we raised the source found.

Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs
StephenHodgson: Because we need to set the first point, and we weren't

StephenHodgson: I don't understand what you're getting at

StephenHodgson: Oh I see, this should be setting the start point. haha. (why didn't you just say that?) 😆

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
keveleigh: Shouldn't be raising input changes, no.
Should be filling in the state though, in case a dev wants to query it in `SourceDetected`. Is that happening?

StephenHodgson: Yes. after we get the controller.

Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs
keveleigh: > (why didn't you just say that?)

>Should this be **startPoint.Position** to match https://github.com/StephenHodgson/MixedRealityToolkit-Unity/blob/c2e196261d24ed8774589ba3b073b963e4ab2388/Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs#L63?

🙃

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
```diff
@@ -179,35 +181,62 @@ private GenericOpenVRController GetOrAddController(XRNodeState xrNodeState)
                     break;
             }
 
-            IMixedRealityInputSource inputSource = InputSystem?.RequestNewGenericInputSource($"{CurrentControllerType} Controller {controllingHand}");
+            Type controllerType;
+
+            switch (CurrentControllerType)
```
SimonDarksideJ: Seems crazy to have two separate identical switch statements, any change we can reorganize it to make it tidier?

Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs
StephenHodgson: 😆 Wow all I saw was the link

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
keveleigh: In here? https://github.com/StephenHodgson/MixedRealityToolkit-Unity/blob/c2e196261d24ed8774589ba3b073b963e4ab2388/Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs#L149

Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/SimpleLineDataProvider.cs
StephenHodgson: Fixed

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
StephenHodgson: Let me add an update there and test it out to see if we're still getting that race condition.

StephenHodgson: Doesn't seem to have the race condition, so I added the update after raise detected

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
StephenHodgson: I resolved them.

StephenHodgson: Besides this parameter was not being used.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
StephenHodgson: I wish we could but there's things that need to happen in a specific order.

1. The pointers need to know the controller type (hence the first switch)
2. The controllers need to request a new Input Source using the generated pointer array
3. We need to actually create the concrete controller reference (hence the second switch)\
4. Then call setup config on the new controller
5. Then set the new controller on each of the new pointers

StephenHodgson: Unless you know of a way to create a concrete instance from a type reference

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
keveleigh: It was being used here:

image

StephenHodgson: Yup, that was the second Update we were talking about that gets called.

StephenHodgson: Getting the controller shouldn't update it.

It should be a discrete step

keveleigh: Agreed.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
Jarodshow: Would `Activator.CreateInstance` not work for this?

StephenHodgson: Good question. I'll try it out.

StephenHodgson: @Jarodshow how would you pass the constructor parameters in this case?

Jarodshow: [CreateInstance(Type, Object[])](https://msdn.microsoft.com/en-us/library/wcxyzt4d(v=vs.110).aspx), Examples are on the page :)

PR #2514, opened 2018-07-29
Creation of the first example scene

SimonDarksideJ:
Overview
---
Basic first example scene, aiming to mimic the WMR startup experience and Start-up view
Aims to demonstrate the primitive basics of vNext operation such as:

* Look
* Pointers
* Basic interaction (still missing interaction scripts)
* Teleporting

> This scene should use **NO additional scripts** where possible. It's aim is to demonstrate the "out of the box" experience the SDK provides.  New profiles (input actions / controller config) are ok, to show the full extent of what is needed to build a project.

image

Aim is to ensure it is a professional looking scene that is engaging as well as educational

Changes
---
- Added scene start
- Added some additional scene assets from HTK
- Also ported the Bounding Box assets as they will be needed later

TODO
---
- [ ] Styling on Scene - help please
- [ ] Lighting / shader effects in interactables - glow on focus
- [ ] Interact sound on click, effect on click
- [x]  Background audio from scene and scene objects required, similar to WMR experience
- Tutorial script / audio (pushed to Beta)
- [ ] Tutorial / example documentation, walking through creation and use of the scene (example panel?)
- [ ] Input config / profiles
- [ ] Input interactions to Activate / Grab and Drop objects in the scene.
- [x] Visualize floor and play area (inscribed within boundary)

Replaces PR #2511 with a branch against main development

PR #2538, opened 2018-08-02
Example scene 1 updates

davidkline-ms:
Overview
---
- Added floor plane and play area rendering.
- Added background (non-spatialized) audio.
- Deleted audio files (bell and chime) that are not being used and, honestly, needed improvement.
- Added Boundary SDK feature folder with play area visualization script.
    - Updated the Boundary demo to use this script as well as the "show spheres that are in specific areas" script

This addresses some issues documented in #2514, which will also be updated with current todo items.


StephenHodgson:
> Added floor plane and play area rendering.

I think we should have a boundary profile with floor plane prefabs that we setup on a per scene basis.

See Controller Pointers PR for an example of how to implement.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scenes/BoundaryVisualization.unity
```diff
@@ -412,6 +412,7 @@ GameObject:
   m_Component:
   - component: {fileID: 353698135}
   - component: {fileID: 353698136}
+  - component: {fileID: 353698137}
```
StephenHodgson: Shouldn't we be updating this on the main dev branch since it's not related to the example scene?

davidkline-ms: I can split it out.

davidkline-ms:
> I think we should have a boundary profile with floor plane prefabs that we setup on a per scene basis.

A little confused here... Didn't we decide to NOT have a boundary profile?

StephenHodgson:
> Didn't we decide to NOT have a boundary profile?

When?

davidkline-ms:
> There's a lot of unrelated changes outside of the example scene.

I am splitting out the boundary sdk / demo changes. The audio file deletion is part of the example work as they went into std assets with the intent of using in the example(s)

davidkline-ms:
>Didn't we decide to NOT have a boundary profile?
When?

When I first implemented the system. Unless you are talking about a different type of profile.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scenes/BoundaryVisualization.unity
StephenHodgson: Thanks 😄

StephenHodgson:
> Unless you are talking about a different type of profile.

That is correct. It'll be more along the lines of the pointer prefab setup, where we have a list of boundary prefabs to spawn based on the platform, the scene, and other settings we decide.

davidkline-ms:
Closing this for now. Will reopen after the boundary sdk PR is complete.

PR #2545, opened 2018-08-03
vNEXT Teleportation Feature Port

StephenHodgson:
Overview
---
Feature port of teleportation in the MRTK.

Teleportation builds off of Controller Pointers and is configured and utilized in the exact same way.

Currently only supports instantaneous teleportation, but support for teleportation over time will come in beta.

TODO
---
- [x] Test on WMR Platform
- [x] All the events for teleportation are setup but the handlers need to be completed.

Other Changes
---
- Fixed an issue with incorrectly transforming the controller position
- Fixed an issue with pointer profile not rendering correctly when no MR Manager instance is found.


SimonDarksideJ:
Liking what I'm seeing so far, however the Standard Asset stuff should be pulled out separately to avoid dirtying this pr

StephenHodgson:
It's getting there, bit still a ways to go.

The standard asset changes could be pulled out. Mostly I just added better distinctions between different gray values.

StephenHodgson:
Okay, this PR is ready for review.

Haven't been able to test with Open VR because it seems we need to double check how we're mapping the buttons to actions.  Setting the actions doesn't seem to do anything for our inputs. (Might have to address in a separate PR just for OpenVR refactor.)

davidkline-ms:
> Haven't been able to test with Open VR

We can cover that.

StephenHodgson:
>>Haven't been able to test with Open VR

>We can cover that.

I'm setup to test with Open VR, but the issues with the implementation needs to be addressed before we can properly test.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/ParabolicTeleportPointer.cs
```diff
@@ -0,0 +1,96 @@
...
+        #region Monobehaviour Implementation
+
+        protected override void OnValidate()
+        {
+            base.OnValidate();
+            EnsureSetup();
+
+            if (parabolicLineData.LineTransform == transform)
+            {
+                Debug.LogWarning("Missing Parabolic line helper.\nThe Parabolic Teleport Pointer requires an empty GameObject child for calculating the parabola arc.");
```
davidkline-ms: Why are we making the dev create the empty child? Seems we should dynamically create / remove automatically

StephenHodgson: We are dynamically creating the child. I just added the warning here just in case something goes wrong.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
```diff
@@ -0,0 +1,263 @@
...
+            }
+
+            // Make sure the camera is at the scene origin.
+            CameraCache.Main.transform.parent.transform.position = Vector3.zero;
+            CameraCache.Main.transform.localPosition = Vector3.zero;
+
+#if UNITY_EDITOR
+            if (!UnityEditor.EditorApplication.isPlaying)
+            {
+
```
davidkline-ms: nit: extra line

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
```diff
@@ -0,0 +1,263 @@
...
+        public override void Unregister(GameObject listener)
+        {
+            base.Unregister(listener);
+        }
+
+        #endregion IEventSystemManager Implementation
+
+        #region IMixedRealityTeleportSystem Implementation
+
+        public float TeleportDuration
```
davidkline-ms: summary tag. should define duration units

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
```diff
@@ -0,0 +1,263 @@
...
+                {
+                    Debug.LogWarning("Couldn't change teleport duration. Teleport in progress.");
+                    return;
+                }
+
+                teleportDuration = value;
+            }
+        }
+
+        private float teleportDuration = 0.25f;
```
davidkline-ms: nit: define above the property (better matches the rest of the code base)

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
```diff
@@ -0,0 +1,263 @@
...
+            };
+
+        /// <inheritdoc />
+        public void RaiseTeleportComplete(IMixedRealityPointer pointer, IMixedRealityTeleportHotSpot hotSpot)
+        {
+            // Check to make sure no one from outside the Teleport System called this method.
+            // Other implementations may have a different way of processing requests.
+            if (isProcessingTeleportRequest)
+            {
+                Debug.LogError("Calls to this method from outside the Teleport System is not allowed in this implementation.");
```
davidkline-ms: Error message does not appear to actually describe the situation.

what sets/clears isProcessingTeleportRequest?

davidkline-ms: also, there isn't a similar check in RaiseTeleportRequest or the other related events

davidkline-ms: This may be what you are looking for: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/caller-information

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
```diff
@@ -266,6 +272,26 @@ public float PointerExtent
         /// <inheritdoc />
         public float SphereCastRadius { get; set; } = 0.1f;
 
+        [SerializeField]
+        [Range(0f, 360f)]
+        [Tooltip("The Y orientation of the pointer - used for touchpad rotation and navigation")]
```
davidkline-ms: could this not be used with thumbstick rotation?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
```diff
@@ -0,0 +1,318 @@
...
+        [SerializeField]
+        protected Gradient LineColorHotSpot = new Gradient();
+
+        [SerializeField]
+        [Tooltip("Layers that are considered 'valid' for navigation")]
+        protected LayerMask ValidLayers = 1 << 0; // Default
+
+        [SerializeField]
+        [Tooltip("Layers that are considered 'invalid' for navigation")]
+        protected LayerMask InvalidLayers = 1 << 2; // Ignore raycast
```
davidkline-ms: Doesnt Unity have a way to access these by name?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
```diff
@@ -0,0 +1,318 @@
...
+        /// <inheritdoc />
+        public override bool IsInteractionEnabled => !IsTeleportRequestActive && teleportEnabled;
+
+        /// <inheritdoc />
+        public override float PointerOrientation
+        {
+            get
+            {
+                if (
+                    TeleportHotSpot != null &&
```
davidkline-ms: nit: can you move this up to the same line as the (?

Assets/MixedRealityToolkit-SDK/StandardAssets/Materials/MRTK_Standard_Emerald_Transparent.mat
```diff
@@ -0,0 +1,144 @@
+%YAML 1.1
```
davidkline-ms: should we add transparent versions of all of the std materials?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
StephenHodgson: It is also used with thumbstick rotation. 

I copied as is. I can update.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
StephenHodgson: In our specific implementation, we're handling the complete event ourselves from within the teleport manager. Other implementations may choose to do it in other ways.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
StephenHodgson: I actually need to remove this. It's currently not being used.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
StephenHodgson: yes, but they're expensive, and a simple bit flag is enough. I simply copied from master for this one and left as is.

Assets/MixedRealityToolkit-SDK/StandardAssets/Materials/MRTK_Standard_Emerald_Transparent.mat
StephenHodgson: Should we?  I just did this one for the teleporter cursor

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
StephenHodgson: yes, I'm not sure how that got passed me. lol

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
davidkline-ms: it still feels like we are not doing what the comment advertises... I am ok with cleaning this up in beta

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
StephenHodgson: Scratch that. It might be used in the future. It's defined as part of the interface, and I'd rather keep it for now.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
StephenHodgson: I don't understand.

Assets/MixedRealityToolkit-SDK/StandardAssets/Materials/MRTK_Standard_Emerald_Transparent.mat
StephenHodgson: Probably a task we can log for later.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
davidkline-ms: #2557

StephenHodgson: I don't think I understand, but okay, let's do that.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
keveleigh: >I simply copied from master for this one and left as is.

Where did you copy from master for this?

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
davidkline-ms: other files in our project define backing fields before the property. ex:

private bool exampleField = false;
public bool ExampleProperty
{ … }

Assets/MixedRealityToolkit-SDK/StandardAssets/Materials/MRTK_Standard_Emerald_Transparent.mat
davidkline-ms: > Should we? I just did this one for the teleporter cursor

Perhaps a rename is more reasonable: MRTK_Default_Teleport_Cursor.mat

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
StephenHodgson: > also, there isn't a similar check in RaiseTeleportRequest or the other related events

That's because it's expected that they come from Teleport Pointers.

StephenHodgson: In our specific implementation of they system, the teleport manager is the one telling everyone listening that the teleport has completed. (line 260)

Other people using different implementations may choose to do it differently.

We have to satisfy the interface implementation of `RaiseTeleportCompleted`, and so too do other implementations.

Assets/MixedRealityToolkit-SDK/StandardAssets/Materials/MRTK_Standard_Emerald_Transparent.mat
StephenHodgson: I just wanted to use a standard asset so we wouldn't make too many one off materials like this.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
davidkline-ms: I'm going to need to understand the intent / system better. Either way, the work item to change this is scheduled for beta :)

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
StephenHodgson: private field before property. Got it.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
StephenHodgson: Nvm, couldn't find it. I think it was from HUX actually. I can't remember exactly where I got it.

davidkline-ms: let's file a beta issue to read the layer values (by name) at startup and cache their mask values.... just in case they change in a future version of Unity

StephenHodgson: > just in case they change in a future version of Unity

I highly doubt that, but sure.

StephenHodgson: > layer values (by name) at startup

Sounds expensive.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -163,12 +163,6 @@ private void UpdateControllerData(InteractionSourceState interactionSourceState)
 
                 // Devices are considered tracked if we receive position OR rotation data from the sensors.
                 TrackingState = (IsPositionAvailable || IsRotationAvailable) ? TrackingState.Tracked : TrackingState.NotTracked;
-
-                if (CameraCache.Main.transform.parent != null)
```
keveleigh: We definitely need this, otherwise the controller position/rotation won't be in world space. Not everything that needs this info is parented under the camera parent, so we should transform into world space.

StephenHodgson: I'm not keen on transforming if over and over again.

Let's annotate the property letting people know that it's local position, and or add a local position property to access it.

StephenHodgson: > Not everything that needs this info is parented under the camera parent

Can you give a specific example we'd be using in beta?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
keveleigh: >Sounds expensive

One time at startup?

davidkline-ms: #2564

StephenHodgson: I think you guys are being a bit overboard here.

The inspector for this is user friendly and shows all the layers in the project in the dropdown.
image

StephenHodgson: The only thing I've done here is set it to the default and ignore raycast layers which will most likely never change (nor have they in 10 years that unity has used them)

StephenHodgson: They're even blocked out from being editable:
image

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
StephenHodgson: I ended up just making RaiseTeleportCompleted private and removed it from the interface.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
StephenHodgson: I found a nice const value for a substitute 👍 

`Physics.DefaultRaycastLayers` and `Physics.IgnoreRaycastLayer`

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
keveleigh: There are/will be solvers that depend on the positions of sources. I guess we can address it then.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
```diff
@@ -20,8 +23,11 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
     /// Base Pointer class for pointers that exist in the scene as GameObjects.
     /// </summary>
     [DisallowMultipleComponent]
-    public abstract class BaseControllerPointer : ControllerPoseSynchronizer, IMixedRealityPointer, IMixedRealitySpatialInputHandler
+    public abstract class BaseControllerPointer : ControllerPoseSynchronizer, IMixedRealityPointer, IMixedRealitySpatialInputHandler, IMixedRealityTeleportHandler
     {
+        private static IMixedRealityTeleportSystem teleportSystem = null;
```
keveleigh: A lot of these additions doesn't seem like things that would be in a "`BaseControllerPointer`" and are teleport-specific. Why are they here and not in specific pointer implementations?

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
StephenHodgson: I'm open to the idea of  adding a Position/Rotation property for world space and a LocalPosition/LocalRotation proptery for local space.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
StephenHodgson: They are here because pointers still need to know about teleportation. That's how we're hiding/showing the pointers and their cursors when a teleport request is raised.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
```diff
@@ -348,24 +374,6 @@ public override int GetHashCode()
 
         #region IMixedRealitySourcePoseHandler Implementation
 
-        /// <inheritdoc />
-        public override void OnSourceLost(SourceStateEventData eventData)
```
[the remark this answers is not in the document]
StephenHodgson: I moved that into a base class.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
keveleigh: Right, but hiding/showing pointers depending on the state of other pointers seems like a common scenario and could get out of hand if every pointer needs to be written into the `BaseControllerPointer`. Teleportation isn't necessarily a special case, though I suppose we can take another look in beta.

StephenHodgson: Indeed, and maybe that can be abstracted out a bit more.

PR #2553, opened 2018-08-06
Create Body prefab and add to all demo scenes

keveleigh:
Overview
---
All of the scenes had the same hierarchy of Body/Head for the camera. This now refactors that out into a prefab and adds the controller visualizer script to the prefab (and thus all scenes).


keveleigh:
Closed until some other PRs get in.

StephenHodgson:
We shouldn't need prefabs anyway.

PR #2559, opened 2018-08-06
Refactored Controller Viz

StephenHodgson:
Overview
---
Controller viz wasn't working in all scenes by default, without any additional setup.  This fixes that issue.

- Added new interface `IMixedRealityControllerPoseSynchronizer` and did a bit of refactoring for the controller pointers, as they will inherit from a class that specifically specializes in syncing controller pose data.
- Refactored `MixedRealityControllerVisualizer ` to inherit from `ControllerPoseSynchronizer`.
    - Base Controller class is now responsible for creating any needed controller models when it starts up based on the config profile, similar to how pointers initialize.


keveleigh:
Do we need a full refactor to get controllers to spawn in the scene without any additional setup, or can the scope of the fix PR be smaller?

StephenHodgson:
We're still in alpha, what's the apprehension coming from? Things are undoubtedly going to shift and move until they're where we want them to be.

I'm only aligning it with the paradigm of "works in any scene without additional setup".

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
```diff
@@ -84,12 +85,19 @@ protected IMixedRealityInputSystem InputSystem
         /// <param name="controllerType"></param>
         public void SetupConfiguration(Type controllerType)
         {
+
             if (MixedRealityManager.Instance.ActiveProfile.IsControllerMappingEnabled)
             {
+                if (MixedRealityManager.Instance.ActiveProfile.ControllerMappingProfile.RenderMotionControllers)
+                {
+                    TryRenderControllerModel(controllerType);
```
SimonDarksideJ: Good that we are bringing this back in to the base controller.
However, it only covers creation, what about destruction of the controller when source is lost?

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
```diff
@@ -141,5 +152,50 @@ public void AssignControllerMappings(MixedRealityInteractionMapping[] mappings)
...
+            {
+                controllerModel = MixedRealityManager.Instance.ActiveProfile.ControllerMappingProfile.GlobalLeftHandModel;
+            }
+            else if (ControllerHandedness == Handedness.Right && MixedRealityManager.Instance.ActiveProfile.ControllerMappingProfile.GlobalRightHandModel != null)
+            {
+                controllerModel = MixedRealityManager.Instance.ActiveProfile.ControllerMappingProfile.GlobalRightHandModel;
+            }
+
+            // If a specific controller template wants to override the global model, assign that instead.
+            if (MixedRealityManager.Instance.ActiveProfile.IsControllerMappingEnabled &&
```
SimonDarksideJ: Move this section above the previous, we should check for an override first and THEN use the global if none found

SimonDarksideJ:
Overall I believe this is a good move, my main concern is that instead of one visualizer, you would need one per hand?  and I can't see how they are added.

Also, creation / destruction seems disjointed.

But as long as this works for now, it's good for Alpha

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
StephenHodgson: Destruction is handled in the pose synchronizer

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
StephenHodgson: Will do.

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
StephenHodgson: Now that I look back the interface already has you implement `DestroyOnSourceLost` to satisfy the requirements of it handling destruction of the object

SimonDarksideJ: OK,  I missed the requirement that this implementation required prefabs, so that should be noted in the documentation

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Controllers/MixedRealityControllerVisualizer.cs
```diff
@@ -1,113 +1,19 @@
...
-    /// To Start, ensure there is a "Controllers Profile" configured against the main Mixed Reality Configuration Profile and that Controller Rendering is enabled in that profile, as well as configuring at least one controller for each platform you intend to support.
-    /// Once ready, then simply add this script to a GameObject in the scene, for example the Camera / head.
-    /// </example>
-    /// <remarks>
-    /// For Alpha, this visualizer only does basic rendering and management of controllers.  In future versions, it will expose specific transforms on a controller model (if it has them) to be able to attach items to any part of the model
-    /// It will also support animation of controller parts and possibly fading in and our of controllers.
-    /// </remarks>
     /// <seealso cref="Internal.Definitions.Devices.MixedRealityControllerMappingProfile"/>
-    public class MixedRealityControllerVisualizer : InputSystemGlobalListener, IMixedRealitySourcePoseHandler, IMixedRealityInputHandler
+    public class MixedRealityControllerVisualizer : ControllerPoseSynchronizer
```
SimonDarksideJ: With the refactor, we can remove this class now as it doesn't do anything in ALpha

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
StephenHodgson: Added doc comments in interface.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Controllers/MixedRealityControllerVisualizer.cs
StephenHodgson: I think it's okay to have a stub for future planning.

Besides I think David already decided this wasn't making it in alpha, but stabilization.

StephenHodgson:
Huh, where did those come from?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Controllers/MixedRealityControllerVisualizer.cs
davidkline-ms: Nothing is 100% decided.... let's at least get a beta task to do a cleanup pass

Issue #2563, opened 2018-08-07
Open VR input mappings are incorrect


StephenHodgson:
## Overview
Not all generic input mappings are correct for Open vr which affects input for different systems.

Impacts #2562

Issue #2562, opened 2018-08-06
vnext bug: cannot teleport using OpenVR layer


davidkline-ms:
## Overview
Teleportation is working on Windows Mixed Reality, not OpenVR (on Vive or via Windows Mixed Reality Steam driver)

## Expected Behavior
Functional teleportation

## Actual Behavior
Stationary user

## Unity Editor Version
2018.1.9f1

## Mixed Reality Toolkit Release Version
vNext Alpha

StephenHodgson:
Mostly bc the open vr layer's input mappings are wrong.  Did we have a task to fix that open already?

davidkline-ms:
not sure. just filing as I find....

davidkline-ms:
just checked.... we didnt have a bug filed

StephenHodgson:
I'll do that now.

Issue #2549, opened 2018-08-03
Open VR controllers do not raise some UnityEditor.XR.Node events


StephenHodgson:
## Overview
Some UnityEditor.XR.Node events do not get raised properly.

When turning a controller off, the tracking lost and node removed events do not get raised by the editor.

## Steps to reproduce
Add the following to the Unity Interaction Events region in `OpenVRDeviceManger.cs`

```


        #region Unity InteractionManager Events

        private void InputTracking_nodeAdded(XRNodeState nodeState)
        {
            if (IsNodeTypeSupported(nodeState))
            {
                Debug.Log("Node added:" + nodeState.nodeType);
                GetOrAddController(nodeState);
            }
        }

        private void InputTracking_trackingAcquired(XRNodeState nodeState)
        {
            if (IsNodeTypeSupported(nodeState))
            {
                if (activeControllers.ContainsKey(nodeState.nodeType))
                {
                    Debug.Log("Tracking acquired:" + nodeState.nodeType);
                    var controller = GetOrAddController(nodeState);
                    InputSystem?.RaiseSourceDetected(controller?.InputSource, controller);
                }
            }
        }

        private void InputTracking_trackingLost(XRNodeState nodeState)
        {
            if (IsNodeTypeSupported(nodeState))
            {
                Debug.Log("Tracking lost:" + nodeState.nodeType);
                var controller = GetOrAddController(nodeState);
                InputSystem?.RaiseSourceLost(controller?.InputSource, controller);
            }
        }

        private void InputTracking_nodeRemoved(XRNodeState nodeState)
        {
            Debug.Log("node removed:" + nodeState.nodeType);
            activeControllers.Remove(nodeState.nodeType);
        }

        #endregion Unity InteractionManager Events
```

Start editor with open vr settings.
Turn controllers on, then off.

## Unity Editor Version
2018.1.9f1

## Mixed Reality Toolkit Release Version
vNEXT Alpha

keveleigh:
I can repro this with Vive as well, on 2018.1.9f1. Might not be WinMR-specific.

SimonDarksideJ:
Ok, so possibly an issue in the editor. What about an actual build?

And for Alpha, is this really an issue? SO long as controllers are detected and work. 
We do need to look for better "non Unity" methods of getting controller data

keveleigh:
> What about an actual build?

Same thing.

[...]

Issue #2548, opened 2018-08-03
Open VR Controller Instances raise assertions when controller is on before playing in editor


StephenHodgson:
## Overview
Open VR controllers raise assertions about being registered with the Input Manger if the controllers are on before the editor starts.

## Expected Behavior
No assertions

## Actual Behavior
Assertions

## Steps to reproduce
Launch any scene with Open VR settings and both controllers enabled.

## Mixed Reality Toolkit Release Version
vNEXT Alpha

SimonDarksideJ:
I had tested this and found no issues previously. What specific assertions are being raised
(And is it your implementation :P )

StephenHodgson:
Haha, I haven't changed anything with the implementation of the open VR controllers.

It says they're being registered twice.

keveleigh:
Probably a side-effect of no lost being fired. Might be a new Unity bug that we weren't seeing in previous builds?

PR #2568, opened 2018-08-07
Added Generic Unity Input Devices & Refactored OVR Support

StephenHodgson:
Overview
---
- Refactored OVR Device Manager to inherit from a Generic Unity Device Manager
- Refactored Generic OVR Controller to inherit from new Generic Unity Controller
- Updated axis utility to properly check if axes are set when MR Manager is initialized in editor if input system is enabled.
- Updated Controller Mapping Profile Inspector
    - Added ability to reset the controller's mappings if generic controller is used.
    - Added invert toggle for dual axis mappings.
    - Added null check for valid action profile.
- Added Windows Mixed Reality Open VR Controller Type
- Added Xbox Controller Type
- Added Oculus Remote mappings

Changes
---
- Fixes: #2563
- Fixes: #2562
- Fixes: #2549
- Fixes: #2548
- Fixes: #2474
- Fixes: #2577
- Fixes: #2335

Todo
---
- [x] Fixed: #2489 Controllers initialize with Controller profiles disabled
- [x] Filter input actions by axis constraints
- [x] Test dual axis vertical inputs are all normalized
    - Up is positive, down is negative


Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
```diff
@@ -191,19 +197,23 @@ public static MixedRealityInteractionMapping[] GetMappingsForControllerType(Syst
             {
                 return WindowsMixedRealityController.DefaultInteractions;
             }
-            else if (controllerType == typeof(OculusTouchController))
```
keveleigh: Why change this?

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
```diff
@@ -142,7 +146,7 @@ public static InputManagerAxis[] GetOpenVRInputManagerAxes
         {
             get
             {
-                return new InputManagerAxis[]
```
keveleigh: Why change this?

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/GenericOpenVRController.cs
```diff
@@ -66,6 +72,8 @@ public GenericOpenVRController(TrackingState trackingState, Handedness controlle
             // Oculus Touch Controller - Button.Four Press
             // Valve Knuckles Controller - Left Controller Outer Face Button
             new MixedRealityInteractionMapping(9, "Unity Button Id 3", AxisType.Digital, DeviceInputType.ButtonPress, MixedRealityInputAction.None, KeyCode.JoystickButton3),
+            new MixedRealityInteractionMapping(10, "WMR Touchpad Touch", AxisType.Digital, DeviceInputType.TouchpadTouch, MixedRealityInputAction.None, KeyCode.JoystickButton18),
```
keveleigh: I see the WMR touchpad touch coming through `new MixedRealityInteractionMapping(6, "Trackpad-Thumbstick Touch", AxisType.Digital, DeviceInputType.TouchpadTouch, MixedRealityInputAction.None, KeyCode.JoystickButton16)` as expected. Do we need this?

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/ViveWandController.cs
```diff
@@ -30,8 +30,8 @@ public ViveWandController(TrackingState trackingState, Handedness controllerHand
         {
             new MixedRealityInteractionMapping(0, "Spatial Pointer", AxisType.SixDof, DeviceInputType.SpatialPointer, MixedRealityInputAction.None),
             new MixedRealityInteractionMapping(1, "Trigger Position", AxisType.SingleAxis, DeviceInputType.Trigger, MixedRealityInputAction.None, KeyCode.None, ControllerMappingLibrary.MIXEDREALITY_AXIS9),
-            new MixedRealityInteractionMapping(2, "Trigger Touch", AxisType.Digital, DeviceInputType.TriggerTouch, MixedRealityInputAction.None, KeyCode.JoystickButton14),
-            new MixedRealityInteractionMapping(3, "Trigger Press", AxisType.Digital, DeviceInputType.TriggerPress, MixedRealityInputAction.None, KeyCode.None, ControllerMappingLibrary.MIXEDREALITY_AXIS9),
+            new MixedRealityInteractionMapping(2, "Trigger Press", AxisType.Digital, DeviceInputType.TriggerPress, MixedRealityInputAction.None, KeyCode.JoystickButton14),
```
keveleigh: What is "Press" vs "Touch"?

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
StephenHodgson: cause we return in every other swtich. Pretty much redundant else.

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
StephenHodgson: code clean up of redundant things

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/ViveWandController.cs
StephenHodgson: Touch is from the axis, where press is from the button

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/GenericOpenVRController.cs
StephenHodgson: Huh, I'll double check. I might have missed this one.

StephenHodgson: image

StephenHodgson: Oh I see what I did here, I got it confused with the axes haha.

StephenHodgson: Well, it is touchpad press for WMR but the other devices have it categorized as touch.

keveleigh: I think these mappings are only valid for UWP. I'm not getting anything anything through these new mappings on Standalone/OpenVR. Are you able to run this?

StephenHodgson: I tested recently with good results, but I'm going down a different path for this. You'll see in an update soon.

StephenHodgson:
>Looks good, however, I would break out the interaction mappings in to their own custom editor window for the screenshot below.
>For that much configuration, you are going to need a bigger screen!

Agreed, and that is a future task I have planned. For now, only the generic controllers will have the custom mappings, where you can fine tune everything. All the other controllers have action assignments only.

davidkline-ms:
>  that is a future task I have planned.

Have you created the task? I do not want to lose it.

Also, please coordinate with @keveleigh on this one as he has spent significant time looking at the mappings recently.

StephenHodgson:
> Also, please coordinate with @keveleigh on this one as he has spent significant time looking at the mappings recently.

I'm happy to accept PRs.

And yes [we already have a task open](2474).

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
```diff
@@ -141,6 +141,9 @@ private static SupportedControllerType CurrentControllerType
                         case "Vive Knuckles - Left":                                            // TODO: Yet to test
                         case "Vive Knuckles- Right":                                            // TODO: Yet to test
                             return SupportedControllerType.ViveKnuckles;
+                        case "OpenVR Controller(WindowsMR: 0x045E/0x065B/0/1) - Left":
+                        case "OpenVR Controller(WindowsMR: 0x045E/0x065B/0/2) - Right":
```
keveleigh: image

Mine comes through slightly differently (Samsung controllers, which might be why?).

StephenHodgson: I suppose we could do a partial match here.

StephenHodgson: I wonder what the hex values refer to.

keveleigh: Good question

keveleigh: >I suppose we could do a partial match here.

Possibly, but we might need to take into account future controllers with different button configurations that might match our partial match.

StephenHodgson: I wonder what is building this string... Hopefully when we switch to the native plugin for input we'll get better info about the controllers

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/GenericOpenVRController.cs
```diff
@@ -124,44 +149,27 @@ public void UpdateController(XRNodeState xrNodeState)
...
             {
-                switch (Interactions[i].InputType)
+                switch (Interactions[i].AxisType)
                 {
-                    case DeviceInputType.None:
+                    case AxisType.None:
                         break;
-                    case DeviceInputType.SpatialPointer:
-                        UpdatePointerData(Interactions[i]);
+                    case AxisType.Digital:
```
keveleigh: With this change, [the line in `UpdateButtonData`](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/8e6554e86d8627c13251544af9463de6d0486d50/Assets/MixedRealityToolkit/_Core/Devices/OpenVR/GenericOpenVRController.cs#L358) that allows for converting a `SingleAxis` to an `OnInputDown/Up` event will never be called.

StephenHodgson: I'll take a look. Trying to get controller registration solid before going in depth on input.

StephenHodgson: Fixed.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/WindowsMixedRealityOpenVRController.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// The Generic OpenVR Controller default interactions.
+        /// </summary>
+        /// <remarks>A single interaction mapping works for both left and right controllers.</remarks>
+        public new static readonly MixedRealityInteractionMapping[] DefaultLeftHandedInteractions =
+        {
+            new MixedRealityInteractionMapping(0, "Spatial Pointer", AxisType.SixDof, DeviceInputType.SpatialPointer, MixedRealityInputAction.None),
+            new MixedRealityInteractionMapping(1, "Spatial Grip", AxisType.SixDof, DeviceInputType.SpatialGrip, MixedRealityInputAction.None),
+            new MixedRealityInteractionMapping(2, "Trigger Position", AxisType.SingleAxis, DeviceInputType.Trigger, ControllerMappingLibrary.MIXEDREALITY_AXIS9),
+            new MixedRealityInteractionMapping(3, "Trigger Press (Select)", AxisType.Digital, DeviceInputType.TriggerPress,  KeyCode.JoystickButton14),
+            new MixedRealityInteractionMapping(4, "Grip Button", AxisType.Digital, DeviceInputType.ButtonPress, KeyCode.JoystickButton4),
```
keveleigh: Nothing is currently coming across many of these axes in Unity 2018. Adding this breaks WMR usability in OpenVR, until Unity fixes their bug.

StephenHodgson: Isn't there an alt keycode to use?

The means generic OVR input would be broken as well, correct?

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/GenericOpenVRController.cs
```diff
@@ -232,11 +235,26 @@ protected void UpdateControllerData(XRNodeState state)
...
-            // TODO: configure an offset pointer position for each OpenVR Controller?
-            // Update the interaction data source
-            interactionMapping.PoseData = currentControllerPose; // Currently no way to get pointer specific data, so we use the last controller pose.
+            if (interactionMapping.InputType == DeviceInputType.SpatialPointer)
+            {
+                // TODO Currently no way to get pointer specific data, so we use the last controller pose.
+                // TODO: configure an offset pointer position for each OpenVR Controller?
+
+                pointerOffsetPose = currentControllerPose;
+                pointerOffsetPosition = currentControllerPose.Rotation.eulerAngles;
```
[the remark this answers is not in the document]
StephenHodgson: Sounds good. I think having it as a vector3 made me put position for some reason haha.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/WindowsMixedRealityOpenVRController.cs
keveleigh: The generic axes are still working, it's just the WMR-specific axes. We could use those here if we want, though touchpad position doesn't seem to be coming through any axis at the moment, since it's the component that's specific to WMR (a thumbstick and a touchpad together).

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
```diff
@@ -34,8 +34,6 @@ public override IMixedRealityController[] GetActiveControllers()
 
         public override void Enable()
         {
-            InputTracking.nodeAdded += InputTracking_nodeAdded;
```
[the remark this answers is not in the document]
keveleigh: >these are controllers attached but not tracked yet

We should be triggering `SourceDetected` events off those then, and drop the tracking ones. `SourceDetected` is independent of and distinct from the tracked state.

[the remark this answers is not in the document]
StephenHodgson: The tracking state doesn't trigger when the controller moves in and out of your view anyway.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/WindowsMixedRealityOpenVRController.cs
StephenHodgson: No Bueno

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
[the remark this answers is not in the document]
StephenHodgson: Okay I flipped this for now and only raise source detected and lost on node added/removed.

But I went ahead and removed the tracking state events, cause we try to get the tracking data in the controller update anyway.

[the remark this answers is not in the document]
StephenHodgson: This might be related to a unity bug. Let's file an issue and get back to it when they fix it. 

Right now it's working reliably enough.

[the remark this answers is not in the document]
StephenHodgson: Don't forget my next PR is refactoring this to use a generic unity input base device manager, that'll be where we get the source detected and lost from.

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/GenericUnityController.cs
```diff
@@ -0,0 +1,232 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Devices;
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+using Microsoft.MixedReality.Toolkit.Internal.Interfaces.InputSystem;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Devices.UnityInput
+{
+    public class GenericUnityController : BaseController
```
Jarodshow: Should `BaseController` become `GenericUnityController`? What are some examples of devices that would utilize `BaseController` and not `GenericUnityController`?

StephenHodgson: Not all controllers will be getting input from the Unity Input System, so no.

We already have the Windows Mixed Reality Controllers getting input directly from the Interaction Manager, instead of unity input.

Assets/MixedRealityToolkit/_Core/Devices/OpenVR/OpenVRDeviceManager.cs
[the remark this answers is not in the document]
StephenHodgson: Completely removed all the XR Node events from the open vr controller bc we assign the node, and during the update, we get the node's status, which updates the controller's tracked state

jessemcculloch:
Not sure if this is the right place to submit this, but when I am in a scene using WMR Immersive and I click backwards on the thumbstick, my position moves back, but also up, and the upward movement appears to be exponential

StephenHodgson:
I don't think those changes are part of this PR.

I'd it the branch with the teleport updates?

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/OpenVR/UnityOpenVRDeviceManager.cs
```diff
@@ -0,0 +1,127 @@
...
+                    controllerType = typeof(UnityOculusTouchController);
+                    break;
+                case SupportedControllerType.OculusRemote:
+                    controllerType = typeof(UnityOculusRemoteController);
+                    break;
+                case SupportedControllerType.WindowsMixedReality:
+                    controllerType = typeof(UnityWindowsMixedRealityUnityOpenVRController);
+                    break;
+                default:
+                    return null;
```
SimonDarksideJ: Again, why is the default NULL?
If we don't recognise it in OpenVR, it should be the generic controllr NOT NULL

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/UnityDeviceManager.cs
```diff
@@ -0,0 +1,164 @@
...
+        protected virtual SupportedControllerType GetCurrentControllerType(string joystickName)
+        {
+            if (string.IsNullOrEmpty(joystickName) ||
+                joystickName.Contains("OpenVR") ||
+                joystickName.Contains("Spatial"))
+            {
+                return SupportedControllerType.None;
+            }
+
+            if (joystickName.Contains("Xbox Controller") ||
```
SimonDarksideJ: Why not just use "Xbox"?  Seems a little convoluted.

SimonDarksideJ: Ad what about generic gamepads?  They wouldn't be supported here.

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/UnityDeviceManager.cs
```diff
@@ -0,0 +1,164 @@
...
+                    {
+                        InputSystem?.RaiseSourceDetected(controller.InputSource, controller);
+                    }
+                }
+            }
+
+            LastDeviceList = joystickNames;
+        }
+
+        protected virtual GenericUnityController GetOrAddController(string joystickName)
```
SimonDarksideJ: Have you tested connecting both gamepad and WMR controllers are the same time?
This virtual / override seems to indicate only one can work at a time.
Plus we never add just the UnityDeviceManager and the UnityOpenVRDeviceManager anywhere, unless you have also made changes to the MR manager?  Plus if both are active, are they not going to conflict (I don't think they would, but just asking)

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/OpenVR/UnityOpenVRDeviceManager.cs
StephenHodgson: because when null is returned, no input source is found and no input source detected is raised.

StephenHodgson: If the controller is a generic ovr controller, when it'll already be assigned the type before we get here.

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/UnityDeviceManager.cs
StephenHodgson: Because Unity names the joysticks by specific strings, anything else would fail to recognize the xbox controller by this joystick name.

StephenHodgson: > And what about generic gamepads? They wouldn't be supported here.

That's not true, anything else falls back to being a generic unity controller.

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/UnityDeviceManager.cs
StephenHodgson: Yes, I've tested with both the WMR controllers, and xbox controller connected at the same time. 

No conflicts there.

StephenHodgson:
@SimonDarksideJ I suppose I could update the names to be less confusing. (like Unity VR Device Manager or something)

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/OpenVR/UnityOpenVRDeviceManager.cs
StephenHodgson: See [Line 46](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2568/files/528aa8eb742b881f4c5094b282c3ef854ea9595a#diff-42d2182336b6426682583fc7e7a29408R46)

In the `GetCurrentControllerType` method we should get a valid open vr controller type if the joystick name contains "OpenVR". If it comes back null, then it's not an open vr device.

Assets/MixedRealityToolkit/_Core/Devices/UnityInput/UnityDeviceManager.cs
StephenHodgson: > Plus we never add just the UnityDeviceManager and the UnityOpenVRDeviceManager anywhere, unless you have also made changes to the MR manager?

The `OpenVRDeviceManager` was already defined in the MRManager. I added the `UnityDeviceManager` next to it.

StephenHodgson:
> we should abstract underneath where the input comes from.

I actually disagree here. We should have concrete controller types for each input source.
Just like how for the WMR Motion controllers we have both a open vr implementation and a uwp implementation.  This isn't different from how we've been patterning our systems. Essentially we're using the same pattern there as well, but here, we have a lot more types to deal with.

It's not too convoluted in the setup:
image

As you can see each namespace gets put into it's own dropdown nicely, and I'd like to keep the unity input as an option for devs who want to use it.

I have every intention to implement the open vr devices using the native dll (same as WMR) but again we're limited by Unity bc we can't update to 2018.2+ yet (as always bc of known bugs).

When we get further along that path, we'll see the OpenVR come back on that list.
For now, I'm okay with removing the Unity prefix from the controller classes, but not the open vr device manager.

SimonDarksideJ:
Do not want "Unity input" listed anywhere for the user to select.
Whether we get input from the Unity input system or the native drivers should be hidden behind the controller. 
They are selecting openvr for input, not multiple implementations of it

StephenHodgson:
> Do not want "Unity input" listed anywhere for the user to select.

Sorry, but that needs to be a valid selection.

StephenHodgson:
> Whether we get input from the Unity input system or the native drivers should be hidden behind the controller.

I disagree because each controller is a specific implementation of that driver's representation of said controller.  It's no different from how we implement each of our feature areas.

keveleigh:
>we should abstract underneath where the input comes from.

**_Strongly_** agree.

>We should have concrete controller types for each input source.

Why? What's the actual benefit? Either way, we're using OpenVR, so why should a dev care if it's coming from Unity input vs the DLL?

StephenHodgson:
> we should abstract underneath where the input comes from.

The whole point of the PR is to to _exactly_ that.

StephenHodgson:
> Why? What's the actual benefit? Either way, we're using OpenVR, so why should a dev care if it's coming from Unity input vs the DLL?

It matters because of the path that data takes to get there based on the specific implementation of that controller.

keveleigh:
>The whole point of the PR is to do exactly that.

image

What are you disagreeing with then?

StephenHodgson:
> What's the new thing in 2018.2 that will allow us to do that while 2018.1 doesn't?

Shows items in the package manager and opens up access to the open vr dll.

keveleigh:
>It matters because of the path that data takes to get there based on the specific implementation of that controller.

When will a controller be defined differently because of where the OpenVR input comes from? It still has the same buttons.

>opens up access to the open vr dll.

Do you have a link to this documentation or something? I haven't heard about this.

StephenHodgson:
> What are you disagreeing with then?

The way Simon wants to implement it by only using a single controller type for both native open vr support and unity support.

StephenHodgson:
> Do you have a link to this documentation or something? I haven't heard about this.

Just open up the project in a later version and you'll see the package folder with all the contents.

Hang on while I grab a link.

StephenHodgson:
> When will a controller be defined differently because of where the OpenVR input comes from? It still has the same buttons.

Sure the controller has the same physical buttons, but how we wire up and send actions/event by those buttons are drastically different.

StephenHodgson:
My real question is have you guys even taken the time to pull and try out the branch/

keveleigh:
>The way Simon wants to implement it by only using a single controller type for both native open vr support and unity support.

So then...the input isn't abstracted? There are then specific implementations and nothing is abstracted for the dev.

StephenHodgson:
> So then...the input isn't abstracted? There are then specific implementations and nothing is abstracted for the dev.

It is abstracted, because it's implemented in different ways based on the concrete type.

keveleigh:
>My real question is have you guys even taken the time to pull and try out the branch/

Yup.

StephenHodgson:
The point I'm trying take is that input from the unity input system and input from the native drivers should be done in entirely different implementations.

Each controller has an implementation, just like each system does.

We may want to do something completely different in the open VR implementation of the controllers, that we didn't wanna do for the unity ones.

I don't want to see a single controller implementation getting input from two different sources.

keveleigh:
>We may want to do something completely different in the open VR implementation of the controllers, that we didn't wanna do for the unity ones.

Do you have an example?

This conversation has got me thinking, do we even need different definitions for controllers on different platforms? It feels like we might even be able to abstract this out further and have definitions based solely on the controller type, while the inputs/platforms are completely abstracted away from the dev. For example, one Windows motion controller mapping that works on both OpenVR and WMR UWP. The controller still has the same buttons, so why would I want the buttons to do different things on different platforms?

StephenHodgson:
> so why would I want the buttons to do different things on different platforms?

You wouldn't., And this PR doesn't do that.

> do we even need different definitions for controllers on different platforms?

Yes we do bc unity input is different from getting the input directly from native dlls, which is the point I'm trying to make.

> abstract this out further and have definitions based solely on the controller type

That's exactly what this PR address and does.

StephenHodgson:
> Do you have an example?

The best example is the WMR controller vs it's open VR counterpart.

StephenHodgson:
Tbh, if I had my way we, we wouldn't be getting input from Unity input system at all.

StephenHodgson:
Let's move forward with this PR.

When we finally decide if we want to get input directly from open vr we will decide then if we want to just completely remove the unity inputs.

For now this works great, and all the input is coming from Unity's input system and their open vr wrapper.

keveleigh:
Would be great if this was broken up into smaller, more reviewable PRs. Multiple things have been refactored and renamed, and it's tough to follow the reasoning and the result.

keveleigh:
> I'm okay with reverting the names of the controllers for now.

That would definitely help work towards resolving my previous comment :)

SimonDarksideJ:
Still not agreed on the naming sorry.
The source of the input should be behind the implementation, at most it should be configuration

Should be like:
* openvrmanager
  - > generic openvr controller (plus the sub types) 
        - >  Unity input
        - >  Native input

The choice of input should be determined by configuration and running platform, or if the user has selected a preference.

It should not be named selections in the controller drop down

I plan to use the same model for UWP as well, since the Windows spacial stuff is not allowed when running in the editor, so the code would force using unity input in the editor and use Windows when in Runtime.

StephenHodgson:
@SimonDarksideJ I reverted the namespace changes. Let me know if there's anything else.

Jarodshow:
@StephenHodgson Do you think it's a good idea to have the directions on the D-Pad be their own InteractionMapping for the Oculus Remote?

StephenHodgson:
@Jarodshow I don't understand, could you open a PR showing the suggested implementation?

StephenHodgson:
@SimonDarksideJ he did, but I already addressed his concerns https://github.com/StephenHodgson/MixedRealityToolkit-Unity/pull/27

StephenHodgson:
> Not entirely sure about Unity input becoming its own device but I can live with it for beta

Agreed, but once we can get access to open vr dll we're gonna drop that so fast!

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
```diff
@@ -15,204 +14,220 @@ public static class ControllerMappingLibrary
...
 
+#if UNITY_EDITOR
+
         /// <summary>
         /// Get the InputManagerAxis data needed to configure the Input Mappings for a controller
         /// </summary>
-        /// <param name="type">The type of controller to retrieve configuration for</param>
         /// <returns></returns>
-        public static InputManagerAxis[] GetOpenVRInputManagerAxes
+        public static InputManagerAxis[] UnityInputManagerAxes => new[]
```
davidkline-ms: Willing to have multiple arrays (one per known controller type + a generic one)?
We can make a very readable UI by allowing users to select their controller type and then see the controls available for mapping. Ping me if you have questions on this feedback.

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
```diff
@@ -15,204 +14,220 @@ public static class ControllerMappingLibrary
...
         /// <summary>
-        /// TOUCHPAD-THUMSTICK_LEFT_CONTROLLER_HORIZONTAL
+        /// Mouse: Position Horizontal Movement<para/>
+        /// HTC Vive Controller: Left Controller Trackpad (2) Horizontal Movement<para/>
+        /// Oculus Touch Controller: Axis2D.PrimaryThumbstick Horizontal Movement<para/>
+        /// Valve Knuckles Controller: Left Controller Trackpad Horizontal Movement<para/>
+        /// Windows Mixed Reality Motion Controller: Left Thumbstick Horizontal Movement<para/>
+        /// Xbox Controller: Left Thumbstick Horizontal Movement<para/>
+        /// </summary>
+        public const string AXIS_1 = "AXIS_1";
```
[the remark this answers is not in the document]
StephenHodgson: I specifically didn't give it an alias because it's for multiple different devices.

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
StephenHodgson: > We can make a very readable UI by allowing users to select their controller type and then see the controls available for mapping. Ping me if you have questions on this feedback.

We already do this. I don't understand how this would change anything in the UI

davidkline-ms:
> I don't think we should make aliases for the axis definitions. They say exactly what they are and need to be generic across all devices.

Looking at these definitions, it is nearly impossible to know what axis maps to what control without having documentation right by a dev's side. I don't see why we couldn't have separate definitions for the control set on known controllers and aliases to make them easy to understand. Would likely make developers more productive without having to search for mappings.

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
davidkline-ms: I am not at a system that i can try this out. Are you saying that I would see "WMR Left Controller Thumbstick Horizontal Axis" instead of AXIS_1 in the UI?

StephenHodgson:
> Looking at these definitions, it is nearly impossible to know what axis maps to what control without having documentation right by a dev's side. I don't see why we couldn't have separate definitions for the control set on known controllers and aliases to make them easy to understand. Would likely make developers more productive without having to search for mappings.

The devs will be configuring actions only, most of the time.  So there's no worry there.
Much of the time the're going to see a controller layout and action dropdowns.

I'm working on the next PR that updates the interaction mapping UI to reflect this.

Now you may be referring to the generic controller's UI, which is quite a bit more complex (with good reason, it gets into the knitty gritty stuff, which not everyone will need to do all the time).

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
StephenHodgson: No, you should def only see AXIS_1 only.

StephenHodgson: There's no point in defining the same axis with a different name, it's redundant.

keveleigh:
>Now you may be referring to the generic controller's UI, which is quite a bit more complex (with good reason, it gets into the knitty gritty stuff, which not everyone will need to do all the time).

It is not currently that way (unless you choose custom mappings, which is already an option), and I don't see a good reason to make it less usable than it is right now. What does two touchpad input types in the same list even mean to a developer who just wants to use our mappings and doesn't want to have to look up what each axis means? Without showing the descriptions like the generic OpenVR mapping does currently, this screen becomes less usable.

I also think the image you show with the xbox controller adds even more unnecessary clutter to the mappings menu.

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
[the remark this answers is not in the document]
SimonDarksideJ: Without trying to get to crass, heck no.

These aren't readable config elements and we are not trying to do unitys job for them by documenting their API. 

Currently the dev / user does not see any of this, they see controller input x maps to action a, that is all.  We have already done the hard work translating the Unity docs to resolve these. 

When / if a dev wants to create their own controller (using the Unity input system, which we wouldn't advise) they will need to confer with the Unity docs as they do today. The same can be said when we enable the dll method, you will need to use Steams/Oculus/whatever to build the underlying mapping.

Let's not overenginer this when the vendor already documents this and devs should already be used to it. We can handle this with our own documentation

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
SimonDarksideJ: > We can make a very readable UI by allowing users to select their controller type and then see the controls available for mapping

Again, no, far to complicated. These are to support the MRTK default configuration. We are already going to supply a custom option for users to build their own, so changing this works not help and is overly complex.

They either use what we give them (most users) and advanced users will either use the custom option, or more likely create their own controller to our spec

SimonDarksideJ:
> What does two touchpad input types in the same list even mean to a developer

This is the advanced custom view and the developer may want to map an input to multiple actions, so we are not doing them. 
I expect most devs to either use our default mappings or create their own controller class. 

The custom screen is a catch all for those that prefer a uioption, to ensure all bases are covered

keveleigh:
>This is the advanced custom view 

@SimonDarksideJ Right, that's my point. This PR removes the custom view from other controllers, and makes the custom view the _only_ view for the generic OpenVR controller. I like the idea of allowing for an optional, advanced custom view, but this PR appears to remove that option.

StephenHodgson:
> makes the custom view the only view for the generic OpenVR controller.

That's the point. A generic controller should be completely customizable. There's no one size fits all here. 

Every known controller has a type and simplified view. I never intended the generic controller to have the simple view, and I never intended the controllers to have an optional advanced view. This PR finishes up the work I was doing there before you guys locked down alpha on us.

StephenHodgson:
> Is there still a way to map an input (say, the touchpad on the WMR controller) to multiple actions

Why would you want to do that?
One input == one action

You can't map multiple inputs and expect that to work correctly. If you want to have a different action for the touchpad make another controller profile and map them. One the the things we wanted to do was make it easy to swap them out at runtime.

If you want the advanced view, specify a generic controller. If you want the nice, prebuilt mapping use the concrete type.

It's clear that you don't understand exactly what's going on here in this menu, (and maybe it's partly our fault for not disabling it outright to begin with) but you should be using the concrete types instead of the generic ones. If you know what your contollers are going to be, then map them with the concrete types.

The only reason the generic is there, is if you're working with a controller that's doesn't have a concrete types (like a prototype or something that just isn't defined yet like a switch or PS4 controller).

SimonDarksideJ:
We discussed on the Slack channels and the pr that added the custom view @keveleigh and agreed it wasn't ready and should be pulled for beta.
The custom view is just far too weighty and needs more discussion to implement properly. 

For beta, to create your custom controller, you works need to clone an existing controller and implement it your way if you needed something more advanced. 

As to a single input doing multiple things, that is not supported yet and needs practical discussion (projects with a need to do so) 
Currently each input maps to a single action, when is then interrogated in code. You can also map the same action to multiple inputs (provided they are of a compatible axis type)

Assets/MixedRealityToolkit/_Core/Definitions/Devices/ControllerMappingLibrary.cs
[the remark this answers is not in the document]
davidkline-ms: > Without trying to get to crass, heck no.

That's not crass :)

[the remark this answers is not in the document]
davidkline-ms: I see your point @SimonDarksideJ .

davidkline-ms:
>Is there still a way to map an input (say, the touchpad on the WMR controller) to multiple actions
Why would you want to do that?
One input == one action
You can't map multiple of the same inputs and expect that to work correctly.

Example, based on the image above:
Xbox controller D-Pad. Left, Right, Up and Down should be able to have individual actions assigned (as many xbox games implement)

Jarodshow:
> Example, based on the image above:
Xbox controller D-Pad. Left, Right, Up and Down should be able to have individual actions assigned (as many xbox games implement)

That is another issue itself #2610 

Still conforms to one input, one action

StephenHodgson:
I thought about adding #2610 to this, but I thought it would delay it even more. We should be able to add it cleanly later.

SimonDarksideJ:
Agreed, we don't have a d-pad or menu style input handler at present and works be a worthy extension for both d-pads and touch pads alike.

So let's create a task for it later, as it's effectively a layer on top of the raw input.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -95,7 +101,7 @@ private WindowsMixedRealityController GetOrAddController(InteractionSourceState
             Handedness controllingHand;
             switch (interactionSourceState.source.handedness)
             {
-                case InteractionSourceHandedness.Unknown:
+                default:
```
davidkline-ms: feels weird to have default come first, but don't change it because of me :)

StephenHodgson: I don't mind reverting, or moving this to the bottom, if it makes more sense.

PR #2571, opened 2018-08-08
Cleaned up event system if teleport system creates one.

StephenHodgson:
Overview
---
Fixes an issue where the teleport system doesn't properly clean up after itself.


StephenHodgson:
>  Seems overkill to get each manager to manage this (no pun in tended)

What to do you mean? If we're sticking to OOP only they should be handling that.

davidkline-ms:
> But let's add a task to Beta to review the event system registration / destruction. Seems overkill to get each manager to manage this (no pun in tended)

@SimonDarksideJ / @StephenHodgson 
Kurtis and I are doing demos all day today, can one of you two fine folk add the task?

PR #2573, opened 2018-08-08
Fix pointer line material not being included in the build

keveleigh:
Overview
---
Unity doesn't include shaders that aren't used in materials either referenced in a scene or in a Resources folder. This PR creates a material for the pointer lines with the same settings as the code was doing, instead of trying to create one at runtime and failing when not in the Editor.


keveleigh:
The MRTK_Standard materials? The script was creating a script that uses the Particles/Alpha Blended shader, so that's the one I used here. I can experiment with the MRTK Standard shader and see if I can replicate the look and feel with it, but, to start, I simply migrated the runtime-created material into an asset.

StephenHodgson:
No worries, I'd like to see if we could get something nicer from the standard materials.

Also trying to keep misc one off materials out of the SDK where possible.

davidkline-ms:
> Also trying to keep misc one off materials out of the SDK where possible.

Yes! If it hasn't been done already, I plan on doing a pass to remove the UIDarkGray material and replace with the Std dark gray

PR #2580, opened 2018-08-10
Focus Provider Update

StephenHodgson:
Overview
---
- Removed duplicate faulty logic when getting focused object via event data.
- Removed duplicate calls to register pointers. (they were already being registered by the input source's on detected event)
- Refactored a bit to make the pattern of getting pointer data the same, by passing a pointer, and getting data out.  This makes getting the focused object safer, by first attempting to get the pointer data, if any exists.

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
@@ -315,13 +315,11 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
         /// <inheritdoc />
         public bool TryGetFocusDetails(IMixedRealityPointer pointer, out FocusDetails focusDetails)
         {
-            foreach (var pointerData in pointers)
+            PointerData pointerData;
+            if (TryGetPointerData(pointer, out pointerData))
             {
-                if (pointerData.Pointer.PointerId == pointer.PointerId)
-                {
-                    focusDetails = pointerData.Details;
-                    return true;
-                }
+                focusDetails = pointerData.Details;
+                return true;
             }
 
             focusDetails = default(FocusDetails);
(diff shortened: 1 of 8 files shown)


Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -313,12 +308,15 @@ public GameObject GetFocusedObject(IMixedRealityPointer pointingSource)
         /// <inheritdoc />
         public bool TryGetFocusDetails(IMixedRealityPointer pointer, out FocusDetails focusDetails)
         {
-            foreach (var pointerData in pointers)
+            if (pointer != null)
             {
-                if (pointerData.Pointer.PointerId == pointer.PointerId)
+                foreach (var pointerData in pointers)
```
davidkline-ms: foreach -> for

StephenHodgson: Pointers is a collection. For is not possible.

keveleigh: vaguegrippingantipodesgreenparakeet-size_restricted

keveleigh: But probably not advised :P

StephenHodgson: https://stackoverflow.com/questions/9625270/what-is-the-fastest-safest-method-to-iterate-over-a-hashset

>A foreach loop has a small amount of addition overhead on an indexed collections (like an array). This is mostly because the foreach does a little more bounds checking than a for loop.
>
>HashSet does not have an indexer so you have to use the enumerator.
>
>In this case foreach is efficient as it only calls MoveNext() as it moves through the collection.

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -385,7 +393,7 @@ private void CreateUiRaycastCamera()
         }
 
         /// <summary>
-        /// Helper for assigning world space canvases event cameras.
+        /// Helper for assigning world space canvases event cameras.<para/>
```
keveleigh: What is `<para/>` and why?

StephenHodgson:
Moving this over to my repo and reopening.

Issue #2554, opened 2018-08-06
vNext Task - Boundary visualization profile should have configurable default "show" options


davidkline-ms:
## Overview
Current boundary profile auto-creates and displays the floor and play area. Not all apps want these created by default.

## Requires
Add show options to profile
Manager needs to read and respect at startup

## Acceptance criteria
- [x] Can suppress creation of default visualizations
- [x] When turned on later, the objects are created

Issue #2552, opened 2018-08-06
vNext Task: Handle floor / boundary z-fighting better


davidkline-ms:
## Overview
Alpha ships with a small fudge factor to keep the play area quad from fighting with the floor plane. There are better ways to handle this (configurable offsets, render order). Beta should handle this better with fewer "magic numbers" baked into the code.

## Requires
Implementing the chosen solution
Testing with multiple boundary visualization combinations

## Acceptance criteria
- [ ] No z-fighting with the floor and any other boundary visualizations

Issue #2544, opened 2018-08-02
vNext Task Add visualization options for tracked area


davidkline-ms:
## Overview
Alpha has (will have) visualization for the floor and for the rectangular play area. Some experiences may wish to visualize the raw tracked area, including to the height configured in the MRConfigurationProfile.

We should provide a simple solution for:
- tracked area outline
- tracked area "walls"

## Requirements
Create visualization script
Update BoundaryVisualizationProfile to add new asset (material, line style, etc) configuration

## Acceptance criteria
- [ ] Can successfully show / hide tracked area border as an outline
- [ ] Can successfully show / hide tracked area "walls" at the configured height

PR #2583, opened 2018-08-11
Boundary system updates and fixes

davidkline-ms:
Overview
---
Boundary system updates and fixes

Changes
---
- [x] Replace quad with a cube for floor
- [x] Specify floor dimensions using a Vector2 instead of a Vector3
- [x] Add default render options for boundary components (ex: floor, play area) to the profile
- [x] Implement display of tracked area
- [x] Implement display of boundary walls
- [x] Implement display of boundary ceiling (at BoundaryHeight above floor)
- [x] Make boundary visualizations (*except the floor*) follow user when teleporting
    - [x] Play Area
    - [x] Tracked Area
    - [x] Walls
    - [x] Ceiling
- [x] Update Boundary Visualization Demo
   - [x] Teleport markers with user
   - [x] Add toggles for tracked area, walls and ceiling

Tasks
---
- Fixes: #2544 
- Fixes: #2552 
- Fixes: #2554 
- Fixes: #2565


StephenHodgson:
>  Make boundary visualizations (except the floor) follow user when teleporting

Sounds like you could just parent them to the body and call it a day, correct?

davidkline-ms:
> Sounds like you could just parent them to the body and call it a day, correct?

I was thinking parenting was part of the solution. My concern with parenting to the Body is that they may move when the user walks vs teleports, which would not be desired.  I will definitely investigate.

Thx!

SimonDarksideJ:
> My concern with parenting to the Body is that they may move when the user walks vs teleports, which would not be desired

Agreed, the boundary shouldn't move when the user walks, only when they teleport.
So the boundary should listen for the teleport events and then reposition.  

Could possibly attach it to the MRManager object, as it is just a "Manager" ?

StephenHodgson:
> Could possibly attach it to the MRManager object, as it is just a "Manager" ?

I don't think we should be attaching anything to the MR Manager GameObject.  People got pretty upset in the last version that things didn't get destroyed. Anything on this object will persist between scenes.

davidkline-ms:
Ready for review. Thanks!

Assets/MixedRealityToolkit/_Core/Definitions/BoundarySystem/MixedRealityBoundaryVisualizationProfile.cs
```diff
@@ -27,20 +83,45 @@ public class MixedRealityBoundaryVisualizationProfile : ScriptableObject
...
+
+        [SerializeField]
+        private Material boundaryCeilingMaterial = null;
+
+        /// <summary>
+        /// The material to use for displaying the boundary ceiling.
+        /// </summary>
+        public Material BoundaryCeilingMaterial => boundaryCeilingMaterial;
+
+        #region Boundary ceiling settings
```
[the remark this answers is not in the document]
davidkline-ms: yep.  thanks!

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
```diff
@@ -27,10 +29,27 @@ public class BoundaryVisualizationDemo : MonoBehaviour, IMixedRealityBoundaryHan
...
+        private bool showBoundaryWalls = true;
+
+        [SerializeField]
+        private bool showBoundaryCeiling = true;
+
         #region MonoBehaviour Implementation
 
+        private void Awake()
+        {
+            markerParent = new GameObject();
```
StephenHodgson: Why not just put this in the scene since it's for the demo?

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
```diff
@@ -128,7 +150,7 @@ private void AddMarkers()
                     {
                         GameObject marker = GameObject.CreatePrimitive(PrimitiveType.Sphere);
                         marker.name = "Boundary Demo Marker";
-                        marker.transform.SetParent(transform);
+                        marker.transform.parent = markerParent.transform;
```
StephenHodgson: I wonder that the difference is here...

Assets/MixedRealityToolkit-SDK/Features/Boundary/System/MixedRealityBoundarySystem.cs
```diff
@@ -38,14 +38,22 @@ private void InitializeInternal()
         {
             boundaryEventData = new BoundaryEventData(EventSystem.current);
 
-            Scale = MixedRealityManager.Instance.ActiveProfile.TargetExperienceScale;
+            boundaryVisualizationParent = new GameObject("Boundary System");
```
StephenHodgson: `Boundary System Visualization`?

StephenHodgson: I think leaving off the last part is a bit misleading, and might have devs looking for a boundary system component on it, when most likely none will exist.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
davidkline-ms: parenting this way for teleportation support.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
davidkline-ms: unity says SetParent is obsolete.

Assets/MixedRealityToolkit-SDK/Features/Boundary/System/MixedRealityBoundarySystem.cs
```diff
@@ -178,26 +256,36 @@ public int GetHashCode(object obj)
...
+
+        /// <summary>
+        /// Parent <see cref="GameObject"/> which will encapsulate all of the teleportable boundary visualizations.
+        /// </summary>
+        private GameObject boundaryVisualizationParent;
+
+        /// <summary>
+        /// Layer used to tell the (non-floor) boundary objects to not accept raycasts
+        /// </summary>
+        private readonly int ignoreRaycastLayer = LayerMask.NameToLayer("Ignore Raycast");
```
StephenHodgson: if we know the layer index, why not just assign it here since it's private? 

```
/// <summary>
/// Layer used to tell the (non-floor) boundary objects to not accept raycasts
/// </summary>
/// <remarks>
/// Default is Ignore Raycast layer.
/// </remarks>
 private readonly int ignoreRaycastLayer = 2; 
```

Assets/MixedRealityToolkit-SDK/Features/Boundary/System/MixedRealityBoundarySystem.cs
```diff
@@ -269,15 +280,30 @@ public int GetHashCode(object obj)
...
-        private GameObject boundaryVisualizationParent;
+        private GameObject BoundaryVisualizationParent => boundaryVisualizationParent ?? (boundaryVisualizationParent = CreateBoundaryVisualizationParent());
+
+        private int ignoreRaycastLayerValue = -1;
 
         /// <summary>
         /// Layer used to tell the (non-floor) boundary objects to not accept raycasts
         /// </summary>
-        private readonly int ignoreRaycastLayer = LayerMask.NameToLayer("Ignore Raycast");
+        private int IgnoreRaycastLayerValue
```
StephenHodgson: wondering if this should be a `LayerMask` instead of an `int`

Assets/MixedRealityToolkit-SDK/Features/Boundary/System/MixedRealityBoundarySystem.cs
```diff
@@ -182,12 +179,26 @@ public override void Destroy()
...
 
+        /// <summary>
+        /// Creates the parent for boundary visualization objects that need to teleport with the user.
+        /// </summary>
+        /// <returns>
+        /// The <see cref="GameObject"/> to which boundary visualizations will be parented.
+        /// </returns>
+        private GameObject CreateBoundaryVisualizationParent()
+        {
+            GameObject visualizationParent = new GameObject("Boundary System");
```
StephenHodgson: `Boundary System Visualization`?

I think leaving off the last part is a bit misleading, and might have devs looking for a boundary system component on it, when most likely none will exist.

davidkline-ms: makes sense... i'll update it

Assets/MixedRealityToolkit-SDK/Features/Boundary/System/MixedRealityBoundarySystem.cs
davidkline-ms: setting the layer on a GameObject needs the value, not the mask.

davidkline-ms: understood. I needed to set the mask of the boundary system visualizations so that they would not interfere with the cursor / teleporting. at least it's a one time lookup.

StephenHodgson: Do we want this to be configurable, like with focus?

image

davidkline-ms: I don't believe so. The boundary visualizations should behave "transparently", meaning that they should not interfere with cursors or other objects in the scene (like the auto-boundary in the shell)

davidkline-ms: Future proofing, in case the layer ever changes.

StephenHodgson: I seriously doubt that, but then we could just change it, if it ever does change.

davidkline-ms: This is a one time cost that ensures that we do not have to update.

davidkline-ms: Defensive coding :)

StephenHodgson: But at the cost of runtime perf.  Idk, I think in this case perf > once in a lifetime event

davidkline-ms: how does this hurt perf when it is a "done once" task?

davidkline-ms: we may be over-indexing a little on perf when we are discussing startup / one time lookups

StephenHodgson: It has a cumulative impact, death by a thousand cuts.

keveleigh: In this case, death by...one cut?

davidkline-ms: I worry about how hard it would be to find these types of bugs if Unity opts to change but sure...

StephenHodgson:
Is everything supposed to be all pink?

StephenHodgson:
Also, I noticed that the floor doesn't seem to move with you when you teleport. Is this intended?

If we're making the ceiling and boundaries stationary to the user, shouldn't we do the same for the floor?

davidkline-ms:
>Is everything supposed to be all pink?
Could we update the boundary wall and ceiling materials to be transparent blue or cyan?

Sounds like your profile does not have materials configured. the one in the PR *should* have  been setup.

davidkline-ms:
>Also, I noticed that the floor doesn't seem to move with you when you teleport. Is this intended?
If we're making the ceiling and boundaries stationary to the user, shouldn't we do the same for the floor?

That was by design as the floor is likely to define the *experience's* bounds, rather than the physical boundary.

StephenHodgson:
@davidkline-ms I noted that the inscribed rectange struct is located in _Core.Definitions.Utilities, but should probably live in _Core.Definitions.BoundarySystem.

Could you add that to this PR?

davidkline-ms:
>@davidkline-ms I noted that the InscribedRectangle and Edge structs are located in _Core.Definitions.Utilities, but should probably live in _Core.Definitions.BoundarySystem.
Same with ExperienceScale enum
Could you add that to this PR?

I am open to moving Edge and InscribedRectangle as they are (currently) most affinitized to the Boundary System. 

ExperienceScale is going to be used in a broader sense (Boundaries are merely the first component to use it).  I want to leave that one in Utilties.

StephenHodgson:
Sounds good.

davidkline-ms:
Edge and InscribedRectangle moved and namespaces updated

PR #2584, opened 2018-08-12
Focus Provider Update w/Pointer Fixes

StephenHodgson:
Overview
---
- Removed duplicate faulty logic when getting focused object via event data.
- Removed duplicate calls to register pointers. (they were already being registered by the input source's on detected event)
- Refactored a bit to make the pattern of getting pointer data the same, by passing a pointer, and getting data out.  This makes getting the focused object safer, by first attempting to get the pointer data, if any exists.
- Removed virtual member call in gaze pointer constructor.


Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -298,38 +292,52 @@ public bool TryGetPointingSource(BaseInputEventData eventData, out IMixedReality
         /// <inheritdoc />
         public GameObject GetFocusedObject(IMixedRealityPointer pointingSource)
         {
+            if (pointingSource == null)
```
keveleigh: I think this should be below the `OverrideFocusedObject` check. That should override everything.

Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
```diff
@@ -150,6 +150,15 @@ public InternalGazePointer(GazeProvider gazeProvider, string pointerName, IMixed
             /// <inheritdoc />
             public override IMixedRealityInputSource InputSourceParent { get; protected set; }
 
+            private float pointerExtent;
+
+            /// <inheritdoc />
+            public override float PointerExtent
```
keveleigh: What's the benefit of overriding here, when we don't do anything differently than the base class?

StephenHodgson: because we're setting it in the constructor, and because it's virtual it'll cause some unintended side effects.

keveleigh: Isn't that just for virtual methods called in base constructors?

StephenHodgson: Any virtual member call in any constructor.

Essentually if you set a virtual member in a constructor, it's calling on a derived member that might not be initialized yet.

StephenHodgson: https://stackoverflow.com/questions/119506/virtual-member-call-in-a-constructor

Issue #2593, opened 2018-08-15
vNEXT Task: Ensure All Input Sources are being implemented


StephenHodgson:
## Overview
Double check and make sure the following input sources are being utilized:

- [x] Dictation Input
- [x] Speech Input
- [x] Touch Screen Input

PR #2594, opened 2018-08-15
Ensured Generic Input Sources were implemented

StephenHodgson:
Overview
---
Made sure input that doesn't come from controllers are implemented.

- Updated generic input sources to be disposable, and updated the MRTK's input manager to create them and dispose of them when needed.
- Added a Input Source Option Profile for setting input actions for generic input sources.
- Updated the profile inspectors namespace to match the folder they live in.
- Added Gesture Input Source.

Changes
---
- Fixes: #2593


SimonDarksideJ:
Going to need to review this a bit more fully when I get home at the weekend.
My main concern is abstracting the physical source of the input from the logical handling of the input. 

Gesture, speech and navigation can be sourced differently from from different devices, so we don't want the solution hard chided to how the HoloLens/win10 does it

StephenHodgson:
Speech is essentially the same at the moment.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -70,6 +70,26 @@ public class MixedRealityInputManager : MixedRealityEventManager, IMixedRealityI
...
+
+        /// <summary>
+        /// Current Dictation Input Source.
+        /// </summary>
+        public DictationInputSource DictationInputSource { get; private set; }
+
+        /// <summary>
+        /// Current Touch Screen Input Source.
+        /// </summary>
+        public TouchscreenInputSource TouchscreenInputSource { get; private set; }
```
davidkline-ms: A question was raised by a customer this week wrt the automatic implementation of inputs like touch screen. They would like the ability to completely shut off input sources that they never intend to use.

Seems like we could leverage our configuration profiles to add check boxes (defaulted to checked) specifying each input type (mouse, touch screen, controller, etc.) and allow the dev to disable specific source types.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -188,6 +208,31 @@ private void InitializeInternal()
...
 #endif // UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
+
+            if (MixedRealityManager.Instance.ActiveProfile.IsSpeechCommandsEnabled)
+            {
+                SpeechInputSource = new SpeechInputSource(MixedRealityManager.Instance.ActiveProfile.SpeechCommandsProfile.SpeechCommands);
+            }
+
+            DictationInputSource = new DictationInputSource();
+            TouchscreenInputSource = new TouchscreenInputSource();
+            GestureInputSource = new GestureInputSource(MixedRealityManager.Instance.ActiveProfile != null && MixedRealityManager.Instance.ActiveProfile.InputSourceOptions.UseRailsNavigation);
```
davidkline-ms: Why is rails navigation checked here?

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
```diff
@@ -0,0 +1,191 @@
...
+            navigationGestureRecognizer = new GestureRecognizer();
+
+            navigationGestureRecognizer.NavigationStarted += NavigationGestureRecognizer_NavigationStarted;
+            navigationGestureRecognizer.NavigationUpdated += NavigationGestureRecognizer_NavigationUpdated;
+            navigationGestureRecognizer.NavigationCompleted += NavigationGestureRecognizer_NavigationCompleted;
+            navigationGestureRecognizer.NavigationCanceled += NavigationGestureRecognizer_NavigationCanceled;
+
+            if (useRailsNavigation)
+            {
+                navigationGestureRecognizer.SetRecognizableGestures(GestureSettings.NavigationRailsX | GestureSettings.NavigationRailsY | GestureSettings.NavigationRailsZ);
```
davidkline-ms: what if the dev wants a subset of the axes? i recommend we surface gesture axis selection to the profile

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
```diff
@@ -0,0 +1,191 @@
...
+            gestureRecognizer.HoldStarted += GestureRecognizer_HoldStarted;
+            gestureRecognizer.HoldCompleted += GestureRecognizer_HoldCompleted;
+            gestureRecognizer.HoldCanceled += GestureRecognizer_HoldCanceled;
+
+            gestureRecognizer.ManipulationStarted += GestureRecognizer_ManipulationStarted;
+            gestureRecognizer.ManipulationUpdated += GestureRecognizer_ManipulationUpdated;
+            gestureRecognizer.ManipulationCompleted += GestureRecognizer_ManipulationCompleted;
+            gestureRecognizer.ManipulationCanceled += GestureRecognizer_ManipulationCanceled;
+
+            gestureRecognizer.SetRecognizableGestures(GestureSettings.Tap | GestureSettings.ManipulationTranslate | GestureSettings.Hold);
```
davidkline-ms: seems as if we are forcing apps to register for all of the gestures... they should be allowed to limit their selection (saves the recognizer unneeded work)

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: Great idea, I'll be sure to add that

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
davidkline-ms: also, when switching to the navigation recognizer, the dev wil lose hold and tap recognition in this implementation

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
```diff
@@ -0,0 +1,191 @@
...
+            if (useRailsNavigation)
+            {
+                navigationGestureRecognizer.SetRecognizableGestures(GestureSettings.NavigationRailsX | GestureSettings.NavigationRailsY | GestureSettings.NavigationRailsZ);
+            }
+            else
+            {
+                navigationGestureRecognizer.SetRecognizableGestures(GestureSettings.NavigationX | GestureSettings.NavigationY | GestureSettings.NavigationZ);
+            }
+
+            if (gestureRecognizer != null && !gestureRecognizer.IsCapturingGestures())
```
davidkline-ms: allow the developer to defer the start of the recognizer

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                gestureRecognizer.StartCapturingGestures();
+            }
+
+            if (navigationGestureRecognizer != null && !navigationGestureRecognizer.IsCapturingGestures())
```
davidkline-ms: this can cause BOTH recognizers to be running at the same time as it is possible tyo have the recognizers created and not started at the same time.

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
StephenHodgson: no, either the system is enabled, or it's not. 

Same goes for the conversation we were having in the spatial awareness construct.

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
StephenHodgson: I'm not sure I understand your comment

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: We're passing in the value assigned to it.

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
davidkline-ms: it is possible for gestureRecognizer to be != null and not recognizing at the same time as the navRecognizer is in the same exact state.

by the code here, each would be started and be running at the same time... which would result in both manipulation and navigation gestures being recognized from the same single gesture

davidkline-ms: in other words, we should never allow gesture recognizers with mutually exclusive settings (manipulation and navigation) to be recognizing at the same time

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
davidkline-ms: i disagree here. there are times where the developer needs multiple recognizers to exist, but to be selective about which is running when (see Holograms 211 for a great example)

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
davidkline-ms: it would.

davidkline-ms: in this case, i would say the HTK has a bug

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
davidkline-ms: i see now

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
StephenHodgson: Ah, I see. Then why does the HTK do it?

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
StephenHodgson: Haha of course you do :p

davidkline-ms: I believe we are having some (likely because of only talking via typing) confusion about what it means for a system vs a feature of that system to be enabled / activated. we should chat via phone or in shiproom (sometimes easier that way)

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
StephenHodgson: I agree, we should make this configurable as well.

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        private readonly GestureRecognizer gestureRecognizer;
+        private readonly GestureRecognizer navigationGestureRecognizer;
+
+        #region Raise GestureRecognizer Events
+
+        private void GestureRecognizer_Tapped(TappedEventArgs args)
+        {
+            InputSystem.RaisePointerClicked(InputSystem.GazeProvider.GazePointer, (Handedness)args.source.handedness, PointerAction, args.tapCount);
```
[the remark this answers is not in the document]
StephenHodgson: Because it's a pointer event not a key press.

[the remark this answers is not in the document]
StephenHodgson: If they do trigger this event, then we should filter them out. Clicks from motion controllers are handled in the controller implemention.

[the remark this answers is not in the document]
keveleigh: Where are clicks handled in the controller implementation? We definitely should not filter them out here.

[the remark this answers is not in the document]
keveleigh: It's a trigger click, if you're using a controller.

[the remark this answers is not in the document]
StephenHodgson: When they press select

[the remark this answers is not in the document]
StephenHodgson: We'll be sure to filter the controllers out here, then.

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
davidkline-ms: >  Then why does the HTK do it?

I honestly don't know,

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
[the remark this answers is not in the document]
StephenHodgson: Why the down thumbs? 😅 I thought that's what you were suggesting.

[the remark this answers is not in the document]
StephenHodgson: Anything that is handled by the controller should happen in the controller implementation, not in this class, so it would be valid to remove anything not related to gesture control.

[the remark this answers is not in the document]
StephenHodgson: Anything handled by the controller should be in the contoller's concrete class implementation, not here.

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
StephenHodgson: Is this similar to how the dictation and keyword recognizers cannot run at the same time?

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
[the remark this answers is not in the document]
keveleigh: >I thought that's what you were suggesting.

Nope! I think everything that comes through this recognizer is valid and should be turned into the appropriate events, without filtering by type. 

I don't think the gesture recognizer being its own input source makes sense. The events are coming from the same input source as the source detected / input down / etc events from the controller implementation, so it makes sense to me that these should reference the same Controller/source.

I also think pointer clicked for the UWP WMR controller implementation makes more sense to happen here, since it's a better defined "click" event than using an input up event, and provides things like `tapCount`.

I started an implementation of the gesture recognizer a week or so ago that maintains the references to the input source and their pointers and whatnot. What do we gain by making this separate from the controller implementation?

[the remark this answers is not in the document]
StephenHodgson: I figured out a way to get the controller references from the windows device manager. Not sure how @SimonDarksideJ will like this, but I thought it was an interesting fix to the issue without filtering out the controllers

StephenHodgson:
For dictation input, I'm not entirely convinced we need to set input actions, seeing as we get specific events for each stage of the process. What do you guys think?

Assets/MixedRealityToolkit/InputSystem/Sources/GestureInputSource.cs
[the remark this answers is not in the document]
StephenHodgson: > I don't think the gesture recognizer being its own input source makes sense. The events are coming from the same input source as the source detected / input down / etc events from the controller implementation, so it makes sense to me that these should reference the same Controller/source.

I do agree with you here, but @SimonDarksideJ made it very clear he wanted to separate these out.

> I started an implementation of the gesture recognizer a week or so ago

Well where's that at, and why is it the first time we're hearing about it?

StephenHodgson:
I'm also tempted to remove the gesture input source until we've collectively decided how we want to handle that.

SimonDarksideJ:
OK, been reading through the comments and code a bit. This should definitely not be the final version but will be "good enough" for now.

It's main limitation will be its exclusivity to Windows 10 and HoloLens for now, for feature parity with htk. Just need to make sure the docs are clear on that. 

We can add a task to refactor this to support other platforms later.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -16,6 +16,7 @@
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.EventSystems;
+using UnityEngine.Windows.Speech;
```
SimonDarksideJ: Please strap this in a Windows #if, 
This should Not be in the core input manager. Will need to refactor this out to another layer later. 

We ideally should never have platform specific dependencies in a manager

SimonDarksideJ: Just wrap it for now, good enough for summit. Won't be enough time now to make it fully cross platform.

cre8ivepark:
Tested on HoloLens but still shows the same symptom described in #2626

StephenHodgson:
@cre8ivepark The gesture input source isn't really implemented in this branch either. I was waiting to get more input/feedback from everyone about where the gesture recognizer's should live.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -188,6 +203,26 @@ private void InitializeInternal()
...
+            }
+
+            if (MixedRealityManager.Instance.ActiveProfile.IsDictationEnabled)
+            {
+                DictationInputSource = new DictationInputSource();
+            }
+
+            if (MixedRealityManager.Instance.ActiveProfile.IsTouchScreenInputEnabled)
+            {
+                TouchscreenInputSource = new TouchscreenInputSource();
```
StephenHodgson: @cre8ivepark This area would be where we create our gesture input source, but I've removed it for now.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityTouchInputProfile.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem
+{
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Mixed Reality Touch Input Profile", fileName = "MixedRealityTouchInputProfile", order = (int)CreateProfileMenuItemIndices.TouchInput)]
+    public class MixedRealityTouchInputProfile : ScriptableObject { }
```
StephenHodgson: We're not using this class. I should remove it

StephenHodgson:
@davidkline-ms sorry I accidentally corrupted the config profile.

SimonDarksideJ:
Getting some unused references errors in the latest commit, can you please confirm and update @StephenHodgson

StephenHodgson:
Confirmed:
image
On tick

PR #2595, opened 2018-08-15
Config Profile SystemType null check.

StephenHodgson:
Overview
---
Made sure we're checking if the `SystemType` is not null before attempting to check if that `SystemType`'s resolved Type was not null.

This was causing our unit tests to fail.

Unit Tests now pass with flying colors.


keveleigh:
Does the [InputSystem check](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/b55bed358775b1c111cd495bec5163bca7913880/Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs#L94) need to be updated too?

StephenHodgson:
Wasn't preventing us from passing our unit tests so it fell through the cracks.

Nice find!

PR #2599, opened 2018-08-16
Update BoundingBox.cs

aarrnndd:
There is a bug in the calculation of the size for the BoundingBox/BoundingBoxRig to be shown around a GameObject target.
To fix it, I removed a code line in method RefreshTargetBounds() that performed a scale transformation which may be correct in some special, simple cases. But in general, if the target object, its parent and the bounding box had different scales, the transformation would be incorrect.
The replacement for the removed line, a scale transformation that works in any case, is implemented in method UpdateScaleTransform(), where it is needed and so that the variable name targetBoundsLocalScale does not have to be changed to remain appropriate.


Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBox.cs
```diff
@@ -442,6 +441,8 @@ protected virtual void UpdateScaleTransform()
             // Get position of object based on renderers
             transform.position = targetBoundsWorldCenter;
             Vector3 scale = targetBoundsLocalScale;
+            scale = target.transform.TransformVector(scale);  // transform the target's local scale to world scale
+            scale = transform.InverseTransformVector(scale);  // transform world scale to the local scale of the bounding box
```
StephenHodgson: Not sure how transforming it from then back to local space fixes this.

Isn't this redundant?

aarrnndd: It's not redundant, because in general the target's local scale is not the same as the local scale of the bounding box.

Please look where variable targetBoundsLocalScale is used:
It is calculated (each Update()) in RefreshTargetBounds(), as the target's local size.
Afterwards it is needed in UpdateScaleTransform(), where it has to be assigned to scaleTransform.localScale.
Since scaleTransform is a child of BoundingBox (see BoundingBox.CreateTransforms(), last line),
targetBoundsLocalScale has to be transformed to the local scale of the BoundingBox.

In my project, I want to apply the BoundingBox/BoundingBoxRig dynamically to GameObjects that are within a complex hierarchy of objects where parents and childs have different scales.
The original code in BoundingBox.cs does not work correctly, it calculates the bounding box about 5 or 10 times too large. (Depending on the scale differences between parents/childs in my GameObject hierarchy.)
But with my changes in BoundingBox.cs, the code works perfectly.

SimonDarksideJ:
Can you refresh and update your PR from htk_development @aarrnndd to avoid conflicting merges and validate this PR please.

wiwei:
Closing issues older than 180 days. If this is still an issue, please reactivate with recent information.

PR #2229, opened 2018-06-01
Fixing a bug the causes FaceOrigin collections to face the wrong direction when collections are rotated (and some bonuses)

Ecnassianer:
This is a pull request for the completion of my proposal:
https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2227

PR #2606, opened 2018-08-17
MRTK SDK UX - ObjectCollection

radicalad:
Overview:
Porting ObjectCollection to the vNext branch. Allow users to easily arrange items in a grid in various volumes (planes, spheres, cylinders).

Changes:
Abstracted original ObjectCollection to a BaseCollection class. This makes for a good pattern to extend from in the future. Additionally, the custom editor script is now inherited by all children.

To use BaseCollection:
Inherit BaseCollection and override the abstract method LayoutChildren(). Easy as that.


Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
```diff
@@ -0,0 +1,134 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    public abstract class BaseCollection : MonoBehaviour
+    {
+        #region public members
+        /// <summary>
+        /// Action called when collection is updated
+        /// </summary>
+        public Action<BaseCollection> OnCollectionUpdated;
```
StephenHodgson: Should we update this to be an event routed through the MRTK?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <summary>
+        /// List of objects with generated data on the object.
+        /// </summary>
+        [HideInInspector]
+        public List<CollectionNode> NodeList = new List<CollectionNode>();
+
+        /// <summary>
+        /// Whether to treat inactive transforms as 'invisible'
+        /// </summary>
+        public bool IgnoreInactiveTransforms = true;
```
StephenHodgson: fields should not be public. Please update them to be private and serialized. If a field needs to be accessed outside of the scope of the class, please add a property accessor.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/LayoutTypeEnum.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    /// <summary>
+    /// Collection layout type enum
+    /// </summary>
+    public enum LayoutTypeEnum
+    {
+        ColumnThenRow,          // Sort by column, then by row
```
StephenHodgson: These comments could be summary tags instead.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/SurfaceTypeEnum.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    /// <summary>
+    /// The type of surface to map the collect to.
+    /// </summary>
+    public enum SurfaceTypeEnum
```
StephenHodgson: Some of these enums should probably be defined in the mrtk core

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// Type of sorting to use.
+        /// </summary>
+        [Tooltip("Type of sorting to use")]
+        public SortTypeEnum SortType = SortTypeEnum.None;
+        #endregion
+
+
+        public virtual void UpdateCollection()
+        {
+
```
davidkline-ms: nit: extra line

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionNode.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    /// <summary>
+    /// Collection node is a data storage class for individual data about an object in the collection.
+    /// </summary>
+    [Serializable]
+    public class CollectionNode
```
StephenHodgson: Should this be a struct?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
```diff
@@ -0,0 +1,134 @@
...
+            // Check for empty nodes and remove them
+            List<CollectionNode> emptyNodes = new List<CollectionNode>();
+
+            for (int i = 0; i < NodeList.Count; i++)
+            {
+                if (NodeList[i].transform == null || (IgnoreInactiveTransforms && !NodeList[i].transform.gameObject.activeSelf) || NodeList[i].transform.parent == null || !(NodeList[i].transform.parent.gameObject == this.gameObject))
```
davidkline-ms: code style: I don't believe we have been using "this." in the toolkit.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/Editor/BaseCollectionEditor.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using UnityEditor;
+
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
```
StephenHodgson: `namespace Microsoft.MixedReality.Toolkit.SDK.UX.Inspectors.Collections`

Also, please update the editor folder name to be `Inspectors` instead.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/Editor/BaseCollectionEditor.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    [CustomEditor( typeof(BaseCollection), true )]
+    public class CollectionEditor : Editor
+    {
+        public override void OnInspectorGUI()
+        {
+            // Draw the default
+            base.OnInspectorGUI();
```
StephenHodgson: Let's go ahead and grab all the properties off the object so we can have a clean inspector.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridCollection.cs
```diff
@@ -0,0 +1,451 @@
...
+    /// </summary>
+    public class GridCollection : BaseCollection
+    {
+        #region public members
+
+        /// <summary>
+        /// Type of surface to map the collection to.
+        /// </summary>
+        [Tooltip("Type of surface to map the collection to")]
+        public SurfaceTypeEnum SurfaceType = SurfaceTypeEnum.Plane;
```
StephenHodgson: I'm pretty sure some of these enums are already defined in core.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
```diff
@@ -0,0 +1,134 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    public abstract class BaseCollection : MonoBehaviour
```
StephenHodgson: Should we have a collection interface?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
```diff
@@ -0,0 +1,134 @@
...
+            LayoutChildren();
+
+            if (OnCollectionUpdated != null)
+            {
+                OnCollectionUpdated.Invoke(this);
+            }
+        }
+
+        /// <summary>
+        /// Internal function to check if a node exists in the NodeList.
```
davidkline-ms: Don't need to mention the function is internal. It is marked private

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
```diff
@@ -0,0 +1,134 @@
...
+                    {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Protected Internal function for laying out all children when UpdateCollection is called.
```
davidkline-ms: Similar to the earlier comment. No need to specify the visibility in the comments

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/LayoutTypeEnum.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    /// <summary>
+    /// Collection layout type enum
+    /// </summary>
+    public enum LayoutTypeEnum
```
davidkline-ms: Rename to ObjectCollectionLayoutType (or similar)?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/OrientTypeEnum.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    /// <summary>
+    /// Orientation type enum for collections
+    /// </summary>
+    public enum OrientTypeEnum
```
davidkline-ms: ObjectCollectionOrientationType? Or, if more generic than that, OrientationType and move to a common location

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/SortTypeEnum.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    /// <summary>
+    /// Sorting type for collections
+    /// </summary>
+    public enum SortTypeEnum
```
davidkline-ms: Rename to ObjectCollectionSortType?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/SurfaceTypeEnum.cs
davidkline-ms: ObjectOrientationSurfaceType?

radicalad: Any guidance on organizational structure? _Core.Utilities.Enums? SortType and OrientType seem to be the most useful.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionNode.cs
davidkline-ms: ObjectCollectionNode?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/Editor/BaseCollectionEditor.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using UnityEditor;
+
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    [CustomEditor( typeof(BaseCollection), true )]
+    public class CollectionEditor : Editor
```
davidkline-ms: BaseObjectCollectionInspector?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/Editor/BaseCollectionEditor.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using UnityEditor;
+
+
```
davidkline-ms: nit: extra line

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridCollection.cs
```diff
@@ -0,0 +1,451 @@
...
+
+        /// <summary>
+        /// Should the objects in the collection face the origin of the collection
+        /// </summary>
+        [Tooltip("Should the objects in the collection be rotated / how should they be rotated")]
+        public OrientTypeEnum OrientType = OrientTypeEnum.FaceOrigin;
```
davidkline-ms: code style: public property with private backing field

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridCollection.cs
radicalad: I didn't see any existing enums when I did a search -> related to the previous comment. Which should we pull in?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/SurfaceTypeEnum.cs
StephenHodgson: `_Core.Definitions.Utilities`

StephenHodgson: That's where most of our enums are defined.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridCollection.cs
StephenHodgson: No worries, I just wanted to double check. I couldn't find any either. The only one that looks similar is PivotAxis, but I think these area all good here.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
radicalad: What are you thinking would be included in it vs the base class? LayoutChildren()?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionNode.cs
StephenHodgson: If we end up changing this to a struct we should also move it to the core/definitions/utilities folder

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
StephenHodgson: Anything that defines a collection.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/SurfaceTypeEnum.cs
radicalad: Ah okay, the naming convention changed. Got it.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/Editor/BaseCollectionEditor.cs
StephenHodgson: Folder location should also reflect the namespace I provided above.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
keveleigh: I'm not what we'd gain that'd be worth the perf impact of routing local events like this through the event system. I'm sure @radicalad knows more about the use of this, it seems to me like mostly the objects in the collection or related objects would care about this event, instead of routing it to every global/modal/etc listener as well.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/OrientTypeEnum.cs
StephenHodgson: Sounds pretty generic.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
StephenHodgson: I was mostly just curious. I don't mind keeping the action, but also adding an event.

radicalad: I'm kinda into it purely for an excuse to get to know the new system better. As far as use case, I get why its there in theory, but I've never used it myself, I almost scrapped it when I revisited it for this PR.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionNode.cs
radicalad: I realized I have null checks for the CollectionNode list. I switched it back to a class.

StephenHodgson: Why would a node be null?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseCollection.cs
radicalad: Old habits die hard. 😄

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/Editor/BaseCollectionEditor.cs
radicalad: Not sure what you mean here. I'm just adding a custom button.

StephenHodgson: Just grab all the serialized properties, and implement them in the gui update. 

If you'd like I can open a PR that does this, if it's still unclear.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -103,21 +117,19 @@ public virtual void UpdateCollection()
...
             for (int i = 0; i < NodeList.Count; i++)
             {
                 if (NodeList[i] != null)
                 {
+
                     if (NodeList[i].transform == node)
                     {
                         return true;
                     }
+
```
StephenHodgson: extra lines?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Protected Internal function for laying out all children when UpdateCollection is called.
+        /// </summary>
+        protected virtual void LayoutChildren() { }
```
StephenHodgson: abstract ?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,482 @@
...
+            float yAngle = -(source.y / circumferenc) * 360f;
+
+            Quaternion rot = Quaternion.Euler(yAngle, xAngle, 0.0f);
+            newPos = rot * newPos;
+
+            return newPos;
+        }
+
+        // Internal function for getting the relative mapping based on a source Vec3 and a radius for cylinder mapping.
+        // source to be mapped to cylinder, Radius for the radius of the cylinder
```
StephenHodgson: Should these be xml format?

```
/// <summary>
/// Comments
/// </summary>
```

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,482 @@
...
+                    {
+                        IterateScatterPacking(NodeList, Radius);
+                    }
+                    break;
+            }
+        }
+
+        // Internal function for getting the relative mapping based on a source Vec3 and a radius for spherical mapping.
+        // The source to be mapped to sphere, radius the radius of the sphere</param>
+        private Vector3 SphericalMapping(Vector3 source, float radius)
```
StephenHodgson: Seems like a nice vector 3 math extension...

radicalad:
Updates made. Left CollectionNode is a class -> had some value assignment issues with structs being immutable. Left the public Action alone as well as the protected virtual method LayoutChildren. The base class approach will give more for free (not needing to implement UpdateCollection() etc) vs an interface.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,482 @@
...
+        public float Width
+        {
+            get { return width; }
+        }
+
+        public float Height
+        {
+            get { return height; }
+        }
+        #endregion
```
StephenHodgson: we've been adding the name of the region at the beginning and end to help identify them better.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
radicalad: good catch.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,482 @@
...
+            }
+
+            switch (SurfaceType)
+            {
+                case ObjectOrientationSurfaceType.Plane:
+                    for (int i = 0; i < NodeList.Count; i++)
+                    {
+                        newPos = nodeGrid[i];
+                        NodeList[i].transform.localPosition = newPos;
+                        switch (OrientType)
```
StephenHodgson: With all these nested switch statements, let's call a method instead and pass any values it needs to it.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,482 @@
...
+                    {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+
+        // Internal function for randomized mapping based on a source Vec3 and a radius for randomization distance.
+        private Vector3 ScatterMapping(Vector3 source, float radius)
```
StephenHodgson: This could also be a handy vector 3 extension.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,482 @@
...
+        }
+
+
+        // Internal function to pack randomly spaced nodes so they don't overlap
+        // Usually requires about 25 iterations for decent packing
+        private void IterateScatterPacking(List<ObjectCollectionNode> nodes, float radiusPadding)
+        {
+            // Sort by closest to center (don't worry about z axis)
+            // Use the position of the collection as the packing center
+            nodes.Sort(delegate (ObjectCollectionNode circle1, ObjectCollectionNode circle2) {
```
StephenHodgson: let's drop the bracket onto the next line.

StephenHodgson: Could we make this a method instead of a delegate or lamda?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,482 @@
...
+            {
+                case LayoutType.ColumnThenRow:
+                default:
+                    for (int c = 0; c < columns; c++)
+                    {
+                        for (int r = 0; r < Rows; r++)
+                        {
+                            if (cellCounter < NodeList.Count)
+                            {
+                                nodeGrid[cellCounter] = new Vector3((c * CellWidth) - startOffsetX + halfcell.x, -(r * CellHeight) + startOffsetY - halfcell.y, 0f) + (Vector3)((NodeList[cellCounter])).Offset;
```
StephenHodgson: We probably don't need to assign a new vector here, seeing as we create a new grid array anyway.

StephenHodgson: Just assign the x,y,z values

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,482 @@
...
+                    break;
+
+                case LayoutType.RowThenColumn:
+                    for (int r = 0; r < Rows; r++)
+                    {
+                        for (int c = 0; c < columns; c++)
+                        {
+                            if (cellCounter < NodeList.Count)
+                            {
+                                nodeGrid[cellCounter] = new Vector3((c * CellWidth) - startOffsetX + halfcell.x, -(r * CellHeight) + startOffsetY - halfcell.y, 0f) + (Vector3)((NodeList[cellCounter])).Offset;
```
StephenHodgson: We probably don't need to assign a new vector here, seeing as we create a new grid array anyway.
We should be able to just set the x,y,z values.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
radicalad: @davidkline-ms marked a change to drop them since they're private/internal. I figured I'd leave the explanation - just removed the XML format.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/OrientationType.cs
```diff
@@ -0,0 +1,32 @@
...
+        FaceOriginReversed,
+        /// <summary>
+        /// Zero rotation
+        /// </summary>
+        FaceFoward,
+        /// <summary>
+        /// Zero rotation + 180 degrees
+        /// </summary>
+        FaceForwardReversed,
+    }
```
Ecnassianer: Erm, it looks like this is not being ported from the latest version of the object collection. Here are very significant changes I made for the June/July release:
https://github.com/Ecnassianer/MixedRealityToolkit-Unity/commit/945401107c48493cd1e649c5a92090cf4710a7b8#diff-67828c8d64af75261de12e4344488104

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
davidkline-ms: I was just looking for "Internal function for getting the ..." to be changed to "Get the ..."

There is no real need to specify the visibility in the comment (it's easy to see in the method signature :)

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/ObjectOrientationSurfaceType.cs
```diff
@@ -11,6 +11,7 @@ public enum ObjectOrientationSurfaceType
         Cylinder,
         Plane,
         Sphere,
-        Scatter,
+        Radial,
+        Scatter
```
Ecnassianer: Was this reordering intentional? I generally try to avoid renumbering enums when I add a new one, in part because it breaks existing serialized data. But more generally a philosophy of "if it's not broken, don't touch it". :)

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/OrientationType.cs
```diff
@@ -28,5 +28,22 @@ public enum OrientationType
         /// Zero rotation + 180 degrees
         /// </summary>
         FaceForwardReversed,
+        /// <summary>
+        /// Parent Relative Up
+        /// </summary>
+        FaceParentUp,
```
Ecnassianer: We should take the htk->vNext port as an opportunity to rename FaceForward and FaceForwardReversed to FaceParentForward and FaceParentBack.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
```diff
@@ -5,6 +5,7 @@
 using Microsoft.MixedReality.Toolkit.Internal.EventDatum.Boundary;
 using Microsoft.MixedReality.Toolkit.Internal.Interfaces.BoundarySystem;
 using Microsoft.MixedReality.Toolkit.Internal.Managers;
+using Microsoft.MixedReality.Toolkit.Internal.Utilities;
```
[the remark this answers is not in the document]
radicalad: They showed up when I did a merge with the latest mrtk_development branch.

Assets/MixedRealityToolkit/_Core/EventDatum/Boundary/BoundaryEventData.cs
```diff
@@ -11,15 +11,6 @@ namespace Microsoft.MixedReality.Toolkit.Internal.EventDatum.Boundary
...
     {
-        ///// <summary>
-        ///// Is the ceiling being visualized by the boundary system.
-        ///// </summary>
-        ///// <remarks>
-        ///// The boundary system defines the ceiling as a plane set at <see cref="IMixedRealityBoundarySystem.BoundaryHeight"/> above the floor.
-        ///// </remarks>
-        // todo: coming in Beta
-        // public bool IsCeilingVisualized { get; private set; }
-
```
Ecnassianer: I think perhaps this should not appear in this diff?

Assets/MixedRealityToolkit/_Core/Utilities/MathUtilities.cs
```diff
@@ -439,5 +439,19 @@ public static float RadiansToDegrees(float radians)
...
+
+        /// <summary>
+        /// Calculates the angle (at pointA) between two, two-dimensional points.
+        /// </summary>
+        /// <param name="pointA">The first point.</param>
+        /// <param name="pointB">The second point.</param>
+        /// <returns>
+        /// The angle between the two points.
+        /// </returns>
+        public static float GetAngleBetween(Vector2 pointA, Vector2 pointB)
```
Ecnassianer: Is this different from Vector2.Angle? https://docs.unity3d.com/ScriptReference/Vector2.Angle.html If so, include the difference in the summary text, since it's not immediately apparent to me what the difference is.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SortType.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities
+{
+    /// <summary>
+    /// Sorting type for collections
+    /// </summary>
+    public enum SortType
```
Ecnassianer: This is too generic of a name. "SortType" seems like it should have entries for QuickSort, BubbleSort, MergeSort etc. This should be something like:
3DSortType
SpatialSortType
VolumetricSortType
PositionalSortingType
GameObjectSortType

Ecnassianer: Does vNext end enum names with "Enum"? i.e, SortTypeEnum. I've seen both in htk (SortType and SortTypeEnum), but I'm not sure what the guideline is for vNext.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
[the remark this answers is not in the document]
Ecnassianer: I think the best way to not have this happen is this:
1) Always keep your fork's mrtk_development free of any changes.
2) When you want to update, merge Microsoft's mrtk_development into your fork's, and it will automatically fast forward with no merge to deal with. Beautiful clean.
3) Create your changes in a branch based on your fork's mrtk_development, don't ever* merge into it.
4) Whenever you get an update from Microsoft's fork, *rebase* your branch to the new tip of your fork's updated mrtk_development branch.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SortType.cs
StephenHodgson: Type is the suffix we've been going with. But I'm fine with either as long as we make a decision.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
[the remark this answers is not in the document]
radicalad: Yep, that's what I followed, not sure why all of this stuff showed up. I'll try to revert and see what I can do.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SortType.cs
radicalad: Thats what I thought too, but going through everything, I didn't find any enum suffixes so i've since renamed everything. Personally, I feel having the suffix helps with self-documentation.

radicalad: How About CollationOrderType. Technically this isn't for sorting - you need an order in which to sort.

radicalad:
Okay, sorry about the crazy merge history. We should be good now. I might have missed something as I rolled WAY back to get to a clean spot. Notable change: added the ...Enum suffix to any enum I'm modifying or adding.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SortType.cs
Ecnassianer: That feels like a synonym for "sort" to me? And it doesn't actually clarify what these are used for. =/

Maybe we're barking at the wrong tree by calling it a sort. Should it be something like SpatialLayoutType instead? SpatialPatternType? 3DLayoutType?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,446 @@
...
+            startOffsetY = (rows * 0.5f) * CellHeight;
+            halfCell = new Vector2(CellWidth * 0.5f, CellHeight * 0.5f);
+
+            // First start with a grid then project onto surface
+            ResolveGridLayout(nodeGrid, startOffsetX, startOffsetY, layout);
+
+            switch (SurfaceType)
+            {
+                case ObjectOrientationSurfaceTypeEnum.Plane:
+                    for (int i = 0; i < NodeList.Count; i++)
```
Ecnassianer: Would this inheritance be more useful if each SurfaceType called an overridable function? That way you would just override the behaviors you wanted differently, and could leave the rest of the LayoutChildren() function alone?

Ecnassianer: Just seems like this is a huge block of functionality to override. :) (I am lazy)

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,446 @@
...
+                            // Make the radius a default value
+                            // TODO move this into a public field ?
+                            NodeList[i].Radius = 1f;
+                        }
+                        NodeList[i].transform.localPosition = newPos;
+                        UpdateNodeFacing(NodeList[i], orientType, newPos);
+                    }
+
+                    // Iterate [x] times
+                    // TODO move center, iterations and padding into a public field
```
Ecnassianer: Is this TODO still relevant? I'm not sure what it's referring to

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/CollationOrderTypeEnum.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// Collation order type used for sorting
+    /// </summary>
+    public enum CollationOrderTypeEnum
+    {
+        /// <summary>
+        /// Don't sort, just display in order received
+        /// </summary>
+        None,
+        /// <summary>
+        /// Sort by transform order
```
Ecnassianer: I haven't used this, so I'm not sure what "transform order" is, but this documentation doesn't help me understand. If you know what this order does, can you improve the summary?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <summary>
+        /// List of objects with generated data on the object.
+        /// </summary>
+        [HideInInspector]
+        public List<ObjectCollectionNode> NodeList = new List<ObjectCollectionNode>();
+        #endregion
+
+        #region private fields
+        /// <summary>
+        /// Whether to treat inactive transforms as 'invisible'
```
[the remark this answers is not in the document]
radicalad: Agreed.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/CollationOrderTypeEnum.cs
radicalad: I'm not sure, I've never used it before - this is leftover from the original enum.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
radicalad: That assumes SurfaceType needs to be implemented - which it doesn't. This Collection just happens to implement LayoutChildren with Surface types. The BaseObjectCollection contains the abstract LayoutChildren method - so more likely you would be inheriting the base collection and writing from scratch - not the grid collection. This is actually the main update I wanted to get into MRTK, as it is, the ObjectCollection is a great control, but since all of its specific functionality is embedded in the single class, its not very extensible.

Ecnassianer: Aha! This all click into sense now that we've talked!

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    public abstract class BaseObjectCollection : MonoBehaviour
+    {
+        #region public members
+        /// <summary>
+        /// Action called when collection is updated
+        /// </summary>
+        public Action<BaseObjectCollection> OnCollectionUpdated;
```
StephenHodgson: Needs to be a property if public

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// <summary>
+        /// List of objects with generated data on the object.
+        /// </summary>
+        [HideInInspector]
+        public List<ObjectCollectionNode> NodeList = new List<ObjectCollectionNode>();
```
StephenHodgson: Also needs to be a property.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <summary>
+        /// Type of sorting to use.
+        /// </summary>
+        [Tooltip("Type of sorting to use")]
+        [SerializeField]
+        protected CollationOrderTypeEnum SortType = CollationOrderTypeEnum.None;
+        #endregion
+
+        #region public accessors
+        public bool IgnoreInactiveTransforms
```
StephenHodgson: missing summary tags

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            get { return ignoreInactiveTransforms; }
+            set { ignoreInactiveTransforms = value; }
+        }
+
+        #endregion
+
+
+        public virtual void UpdateCollection()
```
StephenHodgson: missing summary tags

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,446 @@
...
+
+        #region public accessors
+        /// <summary>
+        /// Type of surface to map the collection to.
+        /// </summary>
+        public ObjectOrientationSurfaceTypeEnum SurfaceType
+        {
+            get { return surfaceType; }
+            set { surfaceType = value; }
+
```
StephenHodgson: extra space

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,446 @@
...
+    /// <summary>
+    /// A Grid Object Collection is simply a set of child objects organized with some
+    /// layout parameters.  The collection can be used to quickly create 
+    /// control panels or sets of prefab/objects.
+    /// </summary>
+    public class GridObjectCollection : BaseObjectCollection
+    {
+        #region private fields
+        [Tooltip("Type of surface to map the collection to")]
+        private ObjectOrientationSurfaceTypeEnum surfaceType = ObjectOrientationSurfaceTypeEnum.Plane;
```
StephenHodgson: properties and their corresponding fields should be co-located.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,446 @@
...
+            }
+
+            for (int i = 0; i < iMax; i++)
+            {
+                for (int j = 0; j < jMax; j++)
+                {
+                    if (cellCounter < NodeList.Count)
+                    {
+                        grid[cellCounter].Set(
+                            ((i * CellWidth) - offsetX + halfCell.x) + NodeList[cellCounter].Offset.x, (-(j * CellHeight) + offsetY - halfCell.y) + NodeList[cellCounter].Offset.y, 0.0f);
```
StephenHodgson: Let's put each x,y,z value assignments on new lines here. Might make it a bit easier to read.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,446 @@
...
+        /// <param name="orientType"></param>
+        /// <param name="newPos"></param>
+        private void UpdateNodeFacing(ObjectCollectionNode node, OrientationTypeEnum orientType, Vector3 newPos = default(Vector3))
+        {
+            Vector3 centerAxis;
+            Vector3 pointOnAxisNearestNode;
+            switch (OrientType)
+            {
+                case OrientationTypeEnum.FaceOrigin:
+                    node.transform.rotation = Quaternion.LookRotation(node.transform.position - this.transform.position, this.transform.up);
```
StephenHodgson: We can probably remove all the redundant `this`'s as well.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,446 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Update the facing of a node given the nodes new position for facing orign with node and orientation type
+        /// </summary>
+        /// <param name="node"></param>
+        /// <param name="orientType"></param>
+        /// <param name="newPos"></param>
+        private void UpdateNodeFacing(ObjectCollectionNode node, OrientationTypeEnum orientType, Vector3 newPos = default(Vector3))
```
StephenHodgson: why not just vector3.zero?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/ObjectCollectionNode.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+    /// <summary>
+    /// Collection node is a data storage class for individual data about an object in a collection.
+    /// </summary>
+    [Serializable]
+    public class ObjectCollectionNode
```
StephenHodgson: pretty sure we can probably make this a struct. Where are we checking if it's null?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Inspectors/Collections/BaseCollectionEditor.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Inspectors.Collections
+{
+    [CustomEditor( typeof(BaseObjectCollection), true )]
+    public class CollectionEditor : Editor
+    {
+        public override void OnInspectorGUI()
+        {
+            // Draw the default
+            base.OnInspectorGUI();
```
StephenHodgson: Let's go ahead and grab all the properties and show them here so we can a nice clean inspector.

Assets/MixedRealityToolkit/_Core/Extensions/VectorExtensions.cs
```diff
@@ -161,5 +161,86 @@ public static bool IsValidPosition(this Vector3 position)
...
+        /// Get the relative mapping based on a source Vec3 and a radius for spherical mapping.
+        /// </summary>
+        /// <param name="source">The source <see cref="Vector3"/> to be mapped to sphere</param>
+        /// <param name="radius">This is a <see cref="float"/> for the radius of the sphere</param>
+        /// <returns></returns>
+        public static Vector3 SphericalMapping(Vector3 source, float radius)
+        {
+            float circ = 2f * Mathf.PI * radius;
+
+            Vector3 newPos = new Vector3(0f, 0f, radius);
```
StephenHodgson: why make a new vector here? couldn't we just assign it back to the source?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/ObjectCollectionNode.cs
radicalad: Its not the null check (it was unnecessary so I removed it), ObjectCollectionNode needs to be mutable. Switching to a struct broke basically every line where properties were being modified.

Assets/MixedRealityToolkit/_Core/Extensions/VectorExtensions.cs
```diff
@@ -161,5 +161,86 @@ public static bool IsValidPosition(this Vector3 position)
...
+        /// <summary>
+        /// Get the relative mapping based on a source Vec3 and a radius for cylinder mapping.
+        /// </summary>
+        /// <param name="source">The source <see cref="Vector3"/> to be mapped to cylinder</param>
+        /// <param name="radius">This is a <see cref="float"/> for the radius of the cylinder</param>
+        /// <returns></returns>
+        public static Vector3 CylindricalMapping(Vector3 source, float radius)
+        {
+            float circ  = 2f * Mathf.PI * radius;
+            Vector3 newPos = new Vector3(0f, source.y, radius);
```
StephenHodgson: same thing here. We probably don't need a new vector and can just update the source, then return it.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Inspectors/Collections/BaseCollectionEditor.cs
radicalad: I'm not sure what you mean by this? I don't typically do custom inspectors.

StephenHodgson: If you'd like I can open a PR with the suggested changes. Shouldn't take but a few minutes.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
Ecnassianer: Dejavu :D 
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2229#discussion_r192543633

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+
+                }
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Protected Internal function for laying out all children when UpdateCollection is called.
+        /// </summary>
+        protected abstract void LayoutChildren()
```
StephenHodgson: missing a `;`

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Inspectors/Collections/BaseCollectionEditor.cs
Ecnassianer: I think it is much better to rely on the base's default behavior. By overriding that behavior and rebuilding it here, we just make it harder to add properties to the class (e.g, they now need to be added in two places to show up in the inspector). From past discussions, I think @keveleigh is of the same opinion.

Assets/MixedRealityToolkit/_Core/Extensions/VectorExtensions.cs
Ecnassianer: It looks like some of Kurtis' changes got lost in the port, you can see this was fixed in the RadialMapping() function in ObjectCollection.cs here:
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2246/files/64f46d85b93b59faf2fd93f69a18deb9f706cdb0

Assets/MixedRealityToolkit/_Core/Extensions/VectorExtensions.cs
Ecnassianer: I didn't bother to find this line in the previous PR, but this was probably also fixed there. (See my comment above)

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/CollationOrderTypeEnum.cs
radicalad: It compares sibling index, I'll rename to better describe the behavior.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
StephenHodgson: Seems as though orientType and newPos are unused in the method. We can just remove them altogether.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/ObjectCollectionNode.cs
StephenHodgson: All you'd need to do is get a local copy, assign the values, then assign your modified copy back to the array. Like so:

```
case ObjectOrientationSurfaceTypeEnum.Scatter:
// Get randomized planar mapping
// Calculate radius of each node while we're here
// Then use the packer function to shift them into place
for (int i = 0; i < NodeList.Count; i++)
{
    var node = NodeList[i];
    newPos = VectorExtensions.ScatterMapping(nodeGrid[i], Radius);
    var nodeCollider = node.transform.GetComponentInChildren<Collider>();

    if (nodeCollider != null)
    {
        // Make the radius the largest of the object's dimensions to avoid overlap
        Bounds bounds = nodeCollider.bounds;
        node.Radius = Mathf.Max(Mathf.Max(bounds.size.x, bounds.size.y), bounds.size.z) * 0.5f;
    }
    else
    {
        // Make the radius a default value
        // TODO move this into a public field ?
        node.Radius = 1f;
    }

    node.transform.localPosition = newPos;
    UpdateNodeFacing(node);
    NodeList[i] = node;
}
```

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Inspectors/Collections/BaseCollectionEditor.cs
StephenHodgson: Let's not be lazy and do it right.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+
+            switch (SortType)
+            {
+                case CollationOrderTypeEnum.None:
+                    break;
+
+                case CollationOrderTypeEnum.Transform:
+                    NodeList.Sort(delegate (ObjectCollectionNode c1, ObjectCollectionNode c2) { return c1.transform.GetSiblingIndex().CompareTo(c2.transform.GetSiblingIndex()); });
```
StephenHodgson: These could be lambda expressions. might shorten it down a bit and make it a tad easier to read.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,143 @@
...
+                    break;
+
+                case CollationOrderTypeEnum.Alphabetical:
+                    NodeList.Sort(delegate (ObjectCollectionNode c1, ObjectCollectionNode c2) { return c1.Name.CompareTo(c2.Name); });
```
StephenHodgson: let's be sure to use ordinal comparison here. and on the apha reversed as well.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
StephenHodgson: Where are we even creating and assigning this action?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Inspectors/Collections/BaseCollectionEditor.cs
radicalad: Sure. whatever is easiest, is there an existing example somewhere in MRTK?

StephenHodgson:
Let's be sure to make all the icons the same for the collection components.
image

StephenHodgson:
Also please be sure to assign the editor inspector icon and rename this class to `BaseCollectionInspector`
image

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Inspectors/Collections/BaseCollectionEditor.cs
Ecnassianer: What I'm asserting is that the base.OnInspectorGUI() or DrawDefaultInspector() methods is "doing it right". It removes code duplication and creates a clearer and easier to maintain inspector for future developers.

Think of it in a more generalized way: say we had a class that provided a virtual DrawMe() function, and then we got a PR to bring in a child class that overrode DrawMe() by copy pasting the body of the parent class's DrawMe() and then added a text label at the bottom. We'd ask for it to be refactored to a call to the parent function followed by adding a the text label.

This is the same, we're inheriting behavior from the default inspector. It's bad practice to duplicate that behavior.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
radicalad: in the UpdateCollection() method. This action gets called every time a user presses the custom inspector button. Like I said earlier, I haven't actually ever subscribed to this action in practice.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
radicalad: Does List<> behave as expected when wrapped in a property? I have the change in a commit -> just not sure about the setter working as one would expect.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/ObjectCollectionNode.cs
radicalad: Ah, thanks @StephenHodgson. 😄 Just curious, is there a perf reason for switching to a struct?

radicalad:
Okay, I think I addressed everything. @StephenHodgson as far as the icon inconsistency, it makes sense to me for the actual UX component script to have the UX icon. Everything else is a back end script (and has the script icon). but the GridObjectCollection script is what you actually apply in a unity scene. I would expect the icon to reflect that.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,419 @@
...
+    /// <summary>
+    /// A Grid Object Collection is simply a set of child objects organized with some
+    /// layout parameters.  The collection can be used to quickly create 
+    /// control panels or sets of prefab/objects.
+    /// </summary>
+    public class GridObjectCollection : BaseObjectCollection
+    {
+        #region public accessors
+        [Tooltip("Type of surface to map the collection to")]
+        private ObjectOrientationSurfaceTypeEnum surfaceType = ObjectOrientationSurfaceTypeEnum.Plane;
```
SimonDarksideJ: Please ensure you use the correct spacings in the script. 
Properties both public and private should be spaced between them.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,419 @@
...
+
+                case OrientationTypeEnum.FaceParentDown:
+                    node.transform.forward = transform.rotation * Vector3.down;
+                    break;
+
+                case OrientationTypeEnum.None:
+                    break;
+
+                default:
+                    throw new ArgumentOutOfRangeException();
```
SimonDarksideJ: Don't use Throw, as it can case Unity to crash which makes it really hard to debug.

Use 
```Debug.LogWarning`` or ```Debug.LogError`` instead (depending on the seriousness of the error.
Include help text in the error where possible to ease diagnosis

File: Assets/MixedRealityToolkit/_Core/Definitions/Utilities/CollationOrderTypeEnum.cs [hunk 59]
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Collation order type used for sorting
+    /// </summary>
+    public enum CollationOrderTypeEnum
+    {
+        /// <summary>
+        /// Don't sort, just display in order received
+        /// </summary>
+        None,
```
SimonDarksideJ: Initialise with
```None = 0,```
As per the enum standards.

File: Assets/MixedRealityToolkit/_Core/Definitions/Utilities/OrientationTypeEnum.cs [hunk 60]
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Orientation type enum
+    /// </summary>
+    public enum OrientationTypeEnum
+    {
+        /// <summary>
+        /// Don't rotate at all
+        /// </summary>
+        None,
```
SimonDarksideJ: Initialise with
```None = 0,```
As per the enum standards.

File: Assets/MixedRealityToolkit/_Core/Definitions/Utilities/OrientationTypeEnum.cs [hunk 61]
```diff
@@ -0,0 +1,50 @@
...
+        FaceParentDown,
+        /// <summary>
+        /// Lay flat on the surface, facing in
+        /// </summary>
+        FaceCenterAxis,
+        /// <summary>
+        /// Lay flat on the surface, facing out
+        /// </summary>
+        FaceCenterAxisReversed,
+
```
davidkline-ms: nit: extra spaces

radicalad:
@SimonDarksideJ added a demo scene! Hopefully we're good to go. Your columns warning was actually a symptom of a larger bug - at some point I deleted the line that sets columns, kinda problematic considering its required for layout... 😄

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/OrientationType.cs
Ecnassianer: bump

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/CollectionEnums/ObjectOrientationSurfaceType.cs
radicalad: Works for me,  I put it at the bottom of the list for preference.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/OrientationType.cs
radicalad: Sure. Done.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/SortType.cs
radicalad: This actually has nothing to do with spatial layout. I'd suggest sticking with Collation here as it has to do with naming of objects vs when they appear in the hierarchy, conceptually it makes sense.

Ecnassianer: I think this now has both type suffixes XD

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -149,8 +175,8 @@ public Mesh CylinderMesh
         #region private fields
 
         private int columns;
-
-        private float circumference;
+        
+        //private float circumference;
```
SimonDarksideJ: If it's not needed, then simply remove it (if you need it later, GIT will remember :D)

radicalad:
@Ecnassianer scatter is arguably still a "grid" layout, but its always felt like the odd man out. I've gone ahead and moved it to its own script, to minimize the amount of refactoring and duplicate code, I have it extending the *GridObjectCollection* and changed a few of the access modifiers in *GridObjectCollection* to make that work. Other than that, I'm with you on the suffixes, I prefer *enum* over *type*.

radicalad:
@Ecnassianer I just reread your comment - you're suggesting *type* over *enum*! @davidkline-ms @StephenHodgson @SimonDarksideJ what is the call here? I'm happy to go with either.

davidkline-ms:
> How come you approved this @davidkline-ms with a reported warning and your "nit" comment?
Especially can't have PR's generating warnings in the project at this stage

because an extra blank line or two is a nit-pick and not, in my book, a merge blocker.  I didn't notice the missing initialization, besides that is far less of a concern than breaking changes at this point in the project.

keveleigh:
@radicalad We've typically been using `Type` as the suffix for enums in vNext. Not sure where/when/if an official decision was made (@davidkline-ms?), but I believe that's been case.

radicalad:
'Type' it is then. Just want to be consistent.

davidkline-ms:
> Not sure where/when/if an official decision was made (@davidkline-ms?), but I believe that's been case.

I honestly do not recall. I believe we are not 100% consistent (not all enums have Type or any other suffix)

Ecnassianer:
Seems like this thread has critical mass to make an official decision?
Enum Suffix: 0
Type Suffix: 5 (David, Kurtis, Eric, Stephen Addison)

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/ScatterObjectCollection.cs
```diff
@@ -0,0 +1,116 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Internal.Extensions;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Collections
+{
+
+
```
StephenHodgson: missing summary

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/ScatterObjectCollection.cs
```diff
@@ -0,0 +1,116 @@
...
+            // Now lets lay out the grid
+            columns = Mathf.CeilToInt( (float)NodeList.Count / Rows );
+            startOffsetX = ( columns * 0.5f ) * CellWidth;
+            startOffsetY = ( Rows * 0.5f ) * CellHeight;
+            halfCell = new Vector2( CellWidth * 0.5f, CellHeight * 0.5f );
+
+            // First start with a grid then project onto surface
+            ResolveGridLayout( nodeGrid, startOffsetX, startOffsetY, Layout );
+
```
StephenHodgson: extra line

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/BaseObjectCollection.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// Type of sorting to use.
+        /// </summary>
+        public CollationOrderType SortType
+        {
+            get { return sortType; }
+            set { sortType = value; }
+        }
+        #endregion public members
+
+
```
StephenHodgson: extra line

StephenHodgson:
Just some nitpicky stuff but looks good.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -0,0 +1,377 @@
...
+            }
+
+            for (int i = 0; i < iMax; i++)
+            {
+                for (int j = 0; j < jMax; j++)
+                {
+                    if (cellCounter < NodeList.Count)
+                    {
+                        grid[cellCounter].Set
+                            (
```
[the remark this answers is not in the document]
radicalad: Is there a convention I should be following?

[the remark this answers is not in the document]
StephenHodgson: Yeah generally `(`'s are on the same line, but parameters can break.

PR #2609, opened 2018-08-17
Fixed some null refs when no input action profile exists

StephenHodgson:
Overview
---

- Speech input source would throw a null ref if created out of order and no input system nor input action profile existed. Added error dialog and safe returns.
- Fixed an issue making a new profile via add button in main config.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealitySpeechCommandsProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealitySpeechCommandsProfileInspector.cs
@@ -31,6 +31,9 @@ namespace Microsoft.MixedReality.Toolkit.Inspectors.Profiles
                 return;
             }
 
+            if (!MixedRealityManager.Instance.ActiveProfile.IsInputSystemEnabled ||
+                MixedRealityManager.Instance.ActiveProfile.InputActionsProfile == null) { return; }
+
             speechCommands = serializedObject.FindProperty("speechCommands");
             actionLabels = MixedRealityManager.Instance.ActiveProfile.InputActionsProfile.InputActions.Select(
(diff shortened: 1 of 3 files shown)


StephenHodgson:
I can't repro with your steps, can you verify?

jessemcculloch:
There was an error in my testing, it's working just fine.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealitySpeechCommandsProfileInspector.cs
```diff
@@ -50,6 +53,18 @@ public override void OnInspectorGUI()
 
             EditorGUILayout.HelpBox("Speech Commands are any/all spoken keywords your users will be able say to raise an Input Action in your application.", MessageType.Info);
 
+            if (!MixedRealityManager.Instance.ActiveProfile.IsInputSystemEnabled)
+            {
+                EditorGUILayout.HelpBox("No input system is enabled, or you need to specify the type in the main configuration profile.", MessageType.Error);
+                return;
+            }
+
+            if (MixedRealityManager.Instance.ActiveProfile.InputActionsProfile == null)
```
keveleigh: Why can't I use speech without input actions? The `SpeechEventData` provides the recognized text, so there are other ways to tell what was recognized.

StephenHodgson: Because Input Actions are required for using the mrtk input system

SimonDarksideJ: I think it's fair to say, some speech is pass through, words to be displayed.
Some are commands, which are actions to be performed. 

I think that is what you are getting at @keveleigh, correct?

If so, that works be a different pr, as this is just bug fixes and what you are suggesting would be a change.

Issue #2591, opened 2018-08-14
bug: it is possible to teleport to "nowhere"


davidkline-ms:
## Overview
On vNext, teleportation does not appear to look for a valid position before moving the user. For example, I was able to go >1750m below the floor in the boundary visualization demo.

## Expected Behavior
By default, to block the user from teleporting outside of the experience's world.

## Actual Behavior
Was able to teleport so far away, I could not see the floor object.

## Steps to reproduce
Load the Boundary Visualization Demo scene, teleport off of the side of the floor platform.

## Unity Editor Version
2018.1.9

## Mixed Reality Toolkit Release Version
mrtk_development branch as of 8/14/2018

StephenHodgson:
Interesting that we're getting a raycast "hit" so far away. The teleport position shouldn't be valid unless we focus a GameObject with a collider and the normal of that raycast is not too steep.

PR #2619, opened 2018-08-20
Fixed teleporting when no hit is found.

StephenHodgson:
Overview
---
Updated valid calculation up direction
Updated distance calculation used when rendering line.

Changes
---
- Fixes: #2591

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
@@ -180,6 +180,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
                 // If we hit something
                 if (Result.CurrentPointerTarget != null)
                 {
+                    BaseCursor?.SetVisibility(true);
                     // Check if it's in our valid layers
                     if (((1 << Result.CurrentPointerTarget.layer) & ValidLayers.value) != 0)
                     {
@@ -195,7 +196,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
                         {
                             // If it's NOT a hotspot, check if the hit normal is too steep 
                             // (Hotspots override dot requirements)
-                            TeleportSurfaceResult = Vector3.Dot(Result.StartPoint, Vector3.up) < upDirectionThreshold
+                            TeleportSurfaceResult = Vector3.Dot(Result.Details.LastRaycastHit.normal, Vector3.up) > upDirectionThreshold
                                 ? TeleportSurfaceResult.Valid
                                 : TeleportSurfaceResult.Invalid;
                         }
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
```diff
@@ -195,7 +196,7 @@ public override void OnPostRaycast()
                         {
                             // If it's NOT a hotspot, check if the hit normal is too steep 
                             // (Hotspots override dot requirements)
-                            TeleportSurfaceResult = Vector3.Dot(Result.StartPoint, Vector3.up) < upDirectionThreshold
+                            TeleportSurfaceResult = Vector3.Dot(Result.Details.LastRaycastHit.normal, Vector3.up) > upDirectionThreshold
```
davidkline-ms: can devs override this threshold? for example if they want to make an experience where the user climbs steep inclines?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
```diff
@@ -350,7 +352,12 @@ public override void OnPositionInputChanged(InputEventData<Vector2> eventData)
                 {
                     canTeleport = false;
                     teleportEnabled = false;
-                    TeleportSystem.RaiseTeleportStarted(this, TeleportHotSpot);
+
+                    if (TeleportSurfaceResult == TeleportSurfaceResult.Valid ||
```
davidkline-ms: I see distance being calculated but it is not being considered here... is that intentional? is distance a factor in whether or not a teleport should be allowed?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
StephenHodgson: Yes by setting the value in the inspector.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
StephenHodgson: At this point we've already calculated a valid point. So I'm not sure what you mean here.

davidkline-ms: was just wondering if distance was involved in determining a valid point.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
```diff
@@ -224,7 +225,7 @@ public override void OnPostRaycast()
                             }
 
                             // Only add the distance between the start point and the hit
-                            clearWorldLength += Vector3.Distance(Result.StartPoint, Result.StartPoint);
+                            clearWorldLength += Vector3.Distance(Result.StartPoint, Result.Details.Point);
```
StephenHodgson: @davidkline-ms this distance calculation is only for properly clamping the line renderer to properly render to the end of the parabola. It is not used in calculating the validity of the teleport result and is outside the scope of this PR.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
StephenHodgson: It is not

StephenHodgson:
Bump

Issue #2610, opened 2018-08-18
vNext Task: Add input action rules to controller events


StephenHodgson:
## Overview
It was suggested that we may also want to call actions for when inputs reach their max/min values or have a specific value.

For example the Xbox Dpad outputs a single Vector 2 value, where as Users want to be able to assign different actions based on the Dpad up / DPad down / etc

This however translate differently based on what the input is, so the user might want a different action for the directions on a touchpad, e.g. N / NW / W / SW / S / SE / E / NE

Currently this is only way to process the directions is in the handler, as shown below. 

```
public void OnPositionInputChanged(InputEventData<Vector2> eventData)
{
    if(eventData.InputData.x == 1)
    {
        // Right Press
    }

    if(eventData.InputData.x == -1)
    {
        // Left Press
    }

    if(eventData.InputData.y == 1)
    {
        // Up Press
    }

    if(eventData.InputData.y == -1)
    {
        // Down Press
    }
}
```

Ideally, there should be an editor configuration screen to be able to add the ability to configure these translations / re-directions of input actions

## Requires
* A new scriptable profile to define rules for input actions (Input Action X with Value Y raises Input Action Z)
* Input Handlers should read from the rules for any configured rules and raise the corresponding event.
* Code documentation
* Supporting user documentation

## Acceptance Criteria
- [x] I can configure 4 or more different input actions for each direction, from a input action received from a Dpad (Dual Axis)
- [x] I can configure 8 or more different input actions for each direction, from a input action received from a Thumbstick (Dual Axis)
- [x] I can configure 8 or more different input actions for each direction, from a input action received from a Touchpad (Dual Axis)
- [x] I can configure different Input actions for pressure received from a single axis / float input action
- [x] I can configure different Input actions for rotations received from a 3Dof or 6Dof input action
- [x] I can configure different Input actions for positions received from a 3Dof or 6Dof input action

SimonDarksideJ:
Updated Task to be a little more generic.  Aim is to provide a "Helper" editor component for users who "don't code", but want a way to generate new actions based on an existing input actions value

SimonDarksideJ:
Revised task based of (post)shiproom discussions on the implementation.

PR #2620, opened 2018-08-21
Add InvertXAxis 

Jarodshow:
## Overview
- Added `InvertXAxis` property

Helps with #2610

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Editor/InvertAxesEnum.cs
+++ b/Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Editor/InvertAxesEnum.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities.Editor
+{
+    /// <summary>
+    /// Used for the editor enum flag field.
+    /// </summary>
+    public enum InvertAxesEnum
+    {
+        Nothing = 0,
+        X = 1,
+        Y = 2,
+        Both = ~0
+    }
+}
--- a/Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityControllerMappingProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityControllerMappingProfileInspector.cs
@@ -4,6 +4,7 @@
 using Microsoft.MixedReality.Toolkit.Internal.Definitions.Devices;
 using Microsoft.MixedReality.Toolkit.Internal.Definitions.InputSystem;
 using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities;
+using Microsoft.MixedReality.Toolkit.Internal.Definitions.Utilities.Editor;
 using Microsoft.MixedReality.Toolkit.Internal.Devices.OpenVR;
 using Microsoft.MixedReality.Toolkit.Internal.Devices.UnityInput;
 using Microsoft.MixedReality.Toolkit.Internal.Managers;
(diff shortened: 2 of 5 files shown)


SimonDarksideJ:
Looks good, but does this also need to go in to the inspectors? 
We have "InvertYAxis" in the inspector, so should this also be added?  @StephenHodgson

StephenHodgson:
Indeed, it would need to get added into the inspector. But seeing as I'm updating that in an ongoing pr, I could make sure it gets updated later.

Jarodshow:
@StephenHodgson 
How does this look?

invert_1


invert_2

Or do you have your own implementation?

StephenHodgson:
Yeah that's awesome. I agree with Simon about the names.

SimonDarksideJ:
On reflection "Both" would be the best choice (invert both axis)
So if you can make that update, I'm more than happy to approve this.

Jarodshow:
I'm using `EditorGUILayout.EnumFlagsField` to render the dropdown content, but the IMGUI is making the logic pretty strenuous, can't seem to get it to work.

Jarodshow:
Someone feel free to contribute if they have a solution

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/Editor/InvertAxesEnum.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Used for the editor enum flag field.
+    /// </summary>
+    public enum InvertAxesEnum
+    {
+        Nothing = 0,
+        X = 1,
+        Y = 2,
+        Both = ~0
```
StephenHodgson: probably need to specify both x & y here instead of the bit assignment bc you didn't specify flags attribute. But also Unity is picky about how it'll render it in the inspector.

Jarodshow: Specifying it as `Both = X & Y` with or without flags yields this

everything

StephenHodgson: Could always define it as 3

Jarodshow: [EditorGUILayout.EnumFlagsField](https://docs.unity3d.com/ScriptReference/EditorGUILayout.EnumFlagsField.html) says:
>Note: This method only supports enums whose underlying types are supported by Unity's >serialization system (sbyte, short, int, byte, ushort, or uint). For enums backed by an unsigned ?>type, the "Everything" option should have the value corresponding to all bits set (i.e. ~0 in an >unchecked context or the MaxValue constant for the type).

so 

```
[System.Flags]
public enum InvertAxesEnum : uint
{
    Nothing = 0,
    X = 1,
    Y = 2,
    Both = uint.MaxValue
}
```

is valid, same as

```
[System.Flags]
public enum InvertAxesEnum
{
    Nothing = 0,
    X = 1,
    Y = 2,
    Both = ~0
}
```

StephenHodgson: Idk, I usually just just the IntPopup

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityControllerMappingProfileInspector.cs
```diff
@@ -546,12 +549,23 @@ private void RenderInteractionList(SerializedProperty list, bool useCustomIntera
...
+                            EditorGUIUtility.labelWidth = defaultLabelWidth;
+                            EditorGUIUtility.fieldWidth = defaultFieldWidth;
+                        } else if ((AxisType)axisType.intValue == AxisType.DualAxis)
+                        {
+                            EditorGUIUtility.labelWidth = customLabelWidth * 0.5f;
+                            EditorGUIUtility.fieldWidth = 8f;
+                            
+                            InvertAxesEnum invertAxesEnum = InvertAxesEnum.Nothing;
+
+                            var modifedValue = (InvertAxesEnum)EditorGUILayout.EnumFlagsField(InvertAxes, invertAxesEnum, GUILayout.Width(customLabelWidth));
```
keveleigh: It doesn't look like this value gets used to write the settings back to the properties anywhere. I've opened a PR with some fixes, assuming I'm not misunderstanding how this is being used.

Jarodshow:
Thanks @keveleigh !

PR #2636, opened 2018-08-23
Corrected assignment of the initial rotation in BoundingBoxGizmoHandle

aarrnndd:
Variable initialRotation is used only in calculations where the rotation coordinate system is local.
So it needs to be initialized with the local instead of the global rotation.
(The previous code version works only as long as none of the parents of transformToAffect were rotated.)

### DIFF
--- a/Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBox.cs
+++ b/Assets/HoloToolkit/UX/Scripts/BoundingBoxes/BoundingBox.cs
@@ -370,7 +370,6 @@ namespace HoloToolkit.Unity.UX
 
             // Store the local scale of the target bb
             targetBoundsLocalScale = localTargetBounds.size;
-            targetBoundsLocalScale.Scale(target.transform.localScale);
 
             // Now check our flatten behavior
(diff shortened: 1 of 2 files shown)


davidkline-ms:
Closing as this is the same fix (#2599) that is already targeted at htk_development. 

Note: GitHub pointed this out when I attempted to retarget the PR.

aarrnndd:
This is not the same fix as #2599, since it contains an addition to it (BoundingBoxGizmoHandle.cs, line 390).
This git stuff is a bit confusing for me, sorry, don't really know what to do now.
But would like to make a suggestion:
In general, the BoundingBox code does not seem to distinguish always cleanly between local and global coordinates.
These problems become apparent when the BoundingBox is applied to dynamically created objects whose parents have a different scale or rotation.
So it would be a task for someone who really knows the code (the original author?) to check all the BoundingBox code for this.
Also it might be useful to create a test and to expand the example accordingly.
These should also contain the creation of the BoundingBox via script (gameObject.AddComponent<BoundingBoxRig>() ...), because there are further minor problems with accessibility of properties.

PR #2660, opened 2018-08-29
Update ---mrtk-task.md

SimonDarksideJ:
Overview
---
Updated the New Task template to include some more required elements and refined the language used

Changes
---
- Updates - vNext Task github template


.github/ISSUE_TEMPLATE/---mrtk-task.md
```diff
@@ -2,16 +2,27 @@
 name: "\U0001F4DD Mixed Reality Toolkit Task"
 about: File an outstanding task for MRTK vNext.
 ---
+> Necessary checks for creating a new vNext Task:
+> * Prefix the Title of the Task with "vNext Task: "
```
StephenHodgson: Why when we can just add the label?

.github/ISSUE_TEMPLATE/---mrtk-task.md
```diff
@@ -2,16 +2,27 @@
...
+> * Remove any defaulted test below, only include new information you are supplying for the task
```
davidkline-ms: test -> text.

also, should likely be two sentences.

.github/ISSUE_TEMPLATE/---mrtk-task.md
SimonDarksideJ: hahahaha, I tested.

WE can add labels, normal users cannot.
And as with the other vNext tasks, we have pre-fixed them for now to make them clear vNext Tasks and not issues or suggestions.

.github/ISSUE_TEMPLATE/---mrtk-task.md
```diff
@@ -2,16 +2,27 @@
...
+> * Ensure requirements detail everything the feature will be required to do
+> * Ensure acceptance criteria (tests) follow the SOLID principles, to ensure when testing, the task passes or fails
+> 
+> *Please remove this section before clicking "Submit new issue" :D
 
 **Overview**
 A clear and concise description of what the outstanding task is.
+Include any reference material as required to properly define why this feature is required (helps with prioritization)
 
 **Requirements**
```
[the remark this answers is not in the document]
SimonDarksideJ: Agreed, will update the sections

SimonDarksideJ:
Updated

.github/ISSUE_TEMPLATE/---mrtk-task.md
SimonDarksideJ: Fixed

StephenHodgson:
Is this targeted to the right branch?

.github/ISSUE_TEMPLATE/---mrtk-task.md
keveleigh: Yeah, that's what we're saying. We'll be the ones adding the labels, since only users with write permissions can. Whether it's prefixed or not, we'll be adding the labels, so the prefix seems redundant.

SimonDarksideJ:
Yes, this is to Master. All the templates will also need replicating to MRTK_Release once the default branch is changed

.github/ISSUE_TEMPLATE/---mrtk-task.md
SimonDarksideJ: It's been the standard for all the vNext tasks so far to make them stand out. :D

keveleigh: Right, and this is a request to change the way things are done going forward. We don't have to do the same thing forever!

Issue #2659, opened 2018-08-29
vNext Task: Update vNext branch homepage


SimonDarksideJ:
**Overview**
Update the vNext homepage to be more relevant following recent development. 
Also update the HTK homepage to be able to easily navigate users to the updated vNext homepages (release and dev)

**Requirements**
* Ensure the information is concise and up to date
* Update details related to releases and release strategy
* Update the Build information? Ensure this works
* Include references to the Task list for outstanding work

**Acceptance Criteria**
- [x] All links on the vNext home page work
- [x] All information is current and works (e.g. the build info)
- [x] A user can quickly find / access the Dev / Release pages from the current HTK homepage
- [x] A user can readily read the current status of the project

PR #2662, opened 2018-08-29
Update master Roadmap

keveleigh:
Overview
---
Updates the roadmap in the master branch, updating the release list and the MRTK vNext plan and branch/release strategy.

Also increments the version file, which was slightly lagging behind the actual release.

Changes
---
- Related to #2659.


Roadmap.md
```diff
@@ -1,40 +1,47 @@
...
+
+For the [master](https://github.com/Microsoft/MixedRealityToolkit-Unity/tree/master) branch releases, development is being done in [htk_development](https://github.com/Microsoft/MixedRealityToolkit-Unity/tree/htk_development). Please refer to the [Upcoming releases table](#upcoming-releases), or the [Projects](https://github.com/Microsoft/MixedRealityToolkit-Unity/projects) page for specifics.
 
 Please see [Branch management plan](#branch-management-plan) for more details.
 
+## MRTK Version Next
 
-# MRTK Version Next
-MixedRealityToolkit Version Next aims to further extend the capabilities of the toolkit and also introduce new features, including the capability to support more VR/AR/XR platforms beyond Microsoft's own Windows Mixed Reality platform.
+MixedRealityToolkit Version Next aims to further extend the capabilities of the toolkit and also introduce new features, including the capability to support more VR / AR / XR platforms beyond Microsoft's own Windows Mixed Reality platform.
```
StephenHodgson: This might be a silly request but I think the order of the realities is important:

` AR / XR / VR ` to reflect the spectrum better.

Ecnassianer: Do you see XR as a segment of the spectrum, rather than the entire spectrum itself?

StephenHodgson: Usually I refer to the whole spectrum as Mixed Reality not XR.

Ecnassianer: I've always read XR as "<Wildcard> Reality", thus the whole spectrum. But you know, we're all making this stuff up as we go, that's the fun part. :)

keveleigh: I also tend to use MR instead of XR.

I updated the order of this list.

Ecnassianer: I only see spaces your commit. :D

PR #2679, opened 2018-08-30
Updated Touch Screen Input Source

StephenHodgson:
Overview
---
Added check for touch support before running.


Assets/MixedRealityToolkit/InputSystem/Sources/TouchscreenInputSource.cs
```diff
@@ -47,7 +47,7 @@ public TouchPointer(string name, Touch touch, Ray ray, IMixedRealityInputSource
         /// </summary>
         public TouchscreenInputSource() : base("Touch Screen Input Source")
         {
-            if (Application.isPlaying)
+            if (Application.isPlaying && Input.touchSupported)
```
keveleigh: Should we instead check and never create the input source in the first place, if touch isn't supported?

PR #2695, opened 2018-09-03
Focus Provider Registration Fix

StephenHodgson:
Overview
---
Fixed an issue where Focus Provider was not registering pointers for input sources that were already detected by the time our Focus Provider is started.

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
@@ -231,6 +231,30 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
                                  "To create a UIRaycastCamera in your scene, find this Focus Provider GameObject and add one there.");
                 CreateUiRaycastCamera();
             }
+
+            foreach (var inputSource in InputSystem.DetectedInputSources)
+            {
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -432,6 +437,23 @@ public bool RegisterPointer(IMixedRealityPointer pointer)
             return true;
         }
 
+        private void RegisterPointers(IMixedRealityInputSource inputSource)
```
SimonDarksideJ: Can we make these RegisterPointer(s) calls more generic and put them in the InputSystem?
Seeing this duplicated / replicated around the project.

StephenHodgson: That's a much larger breaking change than a simple bug fix.

SimonDarksideJ: Fair enough, make a task for it later :D

StephenHodgson: Much appreciated. I do agree we can refine this class a bit more and possibly make it a full proper manager. For now I want to get the rest of the input devices working first.

PR #2701, opened 2018-09-04
GazeProvider Start() visibility to protected for derived providers

wassx:
Overview
---
Broadened visibility for Start() to allow derived providers to add e.g. a GazeFuse to the DefaultCursor after init.


keveleigh:
Should this have been changed to `protected virtual`?

Issue #2630, opened 2018-08-22
vNEXT Task: HoloLens Gesture Support


StephenHodgson:
## Overview
Add HoloLens gesture support: Tap, Hold, Manipulate, & Navigate.
(Hand Simulation to be done separately, I'd like to discuss how to handle mouse support via profile)

## Acceptance criteria
- [x] User should be able to configure gesture recognizer from Input System Config profile
- [x] User should be able to configure input actions for Tap, Hold, Manipulate, & Navigate in Input System Config Profile.
- [x] When user raises hand to the ready position, the cursor should update to ready state.
- [x] When user taps on a GameObject, the Input System raises a Click event.
- [x] When user taps the cursor should animate properly with interaction states & cursor shapes in HoloLens:
![2018-08-22 17_36_24-yoonpark_unityuniteberlin2018_mixedrealitytoolkit_20180621 pptx - powerpoint](https://user-images.githubusercontent.com/13754172/44498198-fcd18d00-a631-11e8-90ee-cb30c454f650.png)
- [x] When user taps and holds a GameObject, the Input System raises a Hold Started.
- [x] User should be able to enable/disable Navigation and Manipulation gesture recognizers
    - [x] System should intelligently disable a recognizer if it's sibling has been enabled, and the recognizer is running.
- [x] When user taps, holds, and moves their hand, the Input System raises the appropriate Manipulation events when Manipulation Gesture Recognizer is running.
- [x] When the user taps, holds, and moves their hand, the Input System raises the appropriate Navigation events when Navigation Gesture Recognizer is running.

Ecnassianer:
It's probably worth calling out that we should do this in the context of some platform agnostic gesture handling system. I know Leap Motion had gesture recognition support for a while (though I think it didn't make it in the new SDK rewrite (yet?)). I wouldn't be surprised if Magic Leap has a similar event API either. We're likely going to see numerous Tap/Hold/etc gesture event APIs as more hand tracking solutions find wider adoption.

Though it's outside the scope of this work item, I think we'll eventually want to implement our own simple gesture recognizer as well, for devices that offer articulated hand positions but not gesture recognition specifically.

StephenHodgson:
That's def planned. But I think we're also trying to figure out if we wanna include the WMR Gestures in the WMR Controller implementation or not bc they're so tightly coupled.

Ecnassianer:
[...]

PR #2705, opened 2018-09-05
Windows Mixed Reality Gesture Support

StephenHodgson:
Overview
---
- Added Windows Mixed Reality Gesture Support.
- Added windows gesture settings to gesture profile.
image

Changes
---
- Fixes: #2630


SimonDarksideJ:
Very confused by the changes so far in this PR:
* There's a bunch of changes to the input system, which should be a separate PR IMHO
* A new GestureSettings enum is introduced but not actually used??
* Don't see anything added for gestures?

If we really are expecting to reuse the WMR controller for gestures, I don't see how you are going to be able to configure the input actions for each of those gestures on the controller picture.

StephenHodgson:
I removed the input actions for events like hold, navigation, and manipulation so we didn't have time add anymore actions to the controller, but I'm still not sold on these changes.

One of the main reasons I went this route is because we need to keep track of the controllers and the interaction source id is the same from the gesture recognizer and the interaction manager. Without the interaction manager to raise source detected/lost the gesture controllers would stick around indefinitely.

I did think about pulling out the input system changes but I needed them to happen to make it all work.

In either case it works on HoloLens and WMR no problem. 

I suppose for other gestures on other platforms they'll have their own managers.

SimonDarksideJ:
Agreed, other platforms will have their own gesture controllers, which is another reason it should be it's own type of controller or device.  Really don't like having two independent controllers mashed together like this.

StephenHodgson:
> Really don't like having two independent controllers mashed together like this.

But they're not. They're literally using the same interaction source id.

As for the other platforms having gestures separate from the Controllers, it'll be likely be up to how they have it exposed in their API as to how we will address it.

SimonDarksideJ:
Retracting my comments based off the Shiproom meeting. Full gesture implementation to be reviewed for 2019 build

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/GestureSettings.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Definitions.InputSystem
+{
+    /// <summary>
+    /// Copy of Unity's XR.WSA.GestureSettings.
+    /// </summary>
+    [Flags]
+    public enum GestureSettings
```
davidkline-ms: Rename to WindowsMixedRealityGestureSettings? Or do we expect to add additional platform gestures here in the future?

StephenHodgson: Needs to be cross platform. This isn't platform specific so I won't be renaming it.

StephenHodgson: I can see us adding to this in the future as more platforms get supported.

Ecnassianer:
> As for the other platforms having gestures separate from the Controllers, it'll be likely be up to how they have it exposed in their API as to how we will address it.

Since we're sourcing gestures from controllers here, we should probably be disambiguating gesture api's back onto our controller classes so the MRTK API is uniform across platforms. The more I see gesture work go in, the more I believe we should create virtual controllers in the case of physical-controllerless hardware like HoloLens or LeapMotion.

StephenHodgson:
@Ecnassianer see #2708

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputSystemProfile.cs
```diff
@@ -63,7 +64,7 @@ public MixedRealityPointerProfile PointerProfile
         /// <summary>
         /// Is the speech Commands Enabled?
         /// </summary>
-        public bool IsSpeechCommandsEnabled => speechCommandsProfile != null && MixedRealityManager.Instance.ActiveProfile.IsInputSystemEnabled && SpeechSystem != null;
+        public bool IsSpeechCommandsEnabled => speechCommandsProfile != null && SpeechSystem != null && MixedRealityManager.Instance.ActiveProfile.IsInputSystemEnabled;
```
SimonDarksideJ: This should not be in the Input System. It was removed in the speech PR

StephenHodgson: I don't understand. All I did was flip the order of the check here.

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/GestureSettings.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        /// <summary>
+        ///   <para>Disable support for gestures.</para>
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        ///   <para>Enable support for the tap gesture.</para>
+        /// </summary>
+        Tap = 1,
```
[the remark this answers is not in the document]
StephenHodgson: Sure thing!

I just copied the original from the WSA gesture settings but I figured we'd expand this to be our own when new gestures are supported.  I'm not as good at converting these to bit flags, could you help?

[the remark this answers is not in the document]
StephenHodgson: I mainly didn't want to change them bc the file was an exact copy of what Unity's gesture settings enum was. Didn't know if changing the type would affect the transformation.

StephenHodgson:
I should also provide a way to set the gesture settings in the gesture profile.

StephenHodgson:
Okay guys, just need to test this out on the HoloLens then I'll let you know when it's ready for review.

davidkline-ms:
will test Monday am to confirm approval

davidkline-ms:
> It's looking a lot better, but I'm still running into an issue where the pointer is raising two events for some reason.

Is this something that we should delay merging on?

StephenHodgson:
I think the breaking changes getting in are pretty important compared to the duplicate pointer events.

That fix will be easier to merge later.

SimonDarksideJ:
I'ts more critical to get these breaking changes in so we do not have any large impacts later @davidkline-ms else any projects updating later will have larger impacts.
Small things are easier to fix later, plus several other PR's depend on this

StephenHodgson:
@davidkline-ms Okay figured it out. It's not a duplicate event. It's a separate event that's triggered by touch amount being greater than zero. It maps to the WMR controller as Trigger Touched.

StephenHodgson:
Okay just need to fix one little thing with the teleport cursors showing on start and this is good to go

StephenHodgson:
Okay, everything looks good now.

StephenHodgson:
@keveleigh any thoughts on this?

StephenHodgson:
Bippity Boppity Bump

davidkline-ms:
When I test this, I can see Hold Action messages, but when I move my hand, I do not get any Manipulation events (that I can see).

Also, I seem to get Hand Lost a lot...

StephenHodgson:
>Also, I seem to get Hand Lost a lot...

That's because your hand went out of view.

StephenHodgson:
> When I test this, I can see Hold Action messages, but when I move my hand, I do not get any Manipulation events (that I can see).

Strange, I thought I only set it up to use `Manipulation Translate` only.

davidkline-ms:
image

davidkline-ms:
this is what is in the default profile for me (I didn't change it that I can recall)

davidkline-ms:
> That's because your hand went out of view.

Actually, my hand was right in front of me. I barely moved it.

StephenHodgson:
I wonder if the bit shifting is off (haha, it's my weaknesss)

StephenHodgson:
Okay this now correctly handles the events. I had an issue where I was raising the wrong event data

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -418,9 +418,24 @@ private void UpdateTriggerData(InteractionSourceState interactionSourceState, Mi
                             if (interactionSourceState.selectPressed)
                             {
                                 InputSystem?.RaiseOnInputDown(InputSource, ControllerHandedness, interactionMapping.MixedRealityInputAction);
+
+                                for (int i = 0; i < InputSource.Pointers.Length; i++)
+                                {
+                                    InputSystem?.RaisePointerDown(InputSource.Pointers[i], ControllerHandedness, interactionMapping.MixedRealityInputAction);
```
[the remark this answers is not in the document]
StephenHodgson: I'll be sure to update.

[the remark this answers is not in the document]
keveleigh: I'm not sure how the touch pointer/controller works, but it's probably worth double checking there that this is handled too.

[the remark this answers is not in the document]
StephenHodgson: I'm pretty sure It's handled correctly.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -418,9 +418,24 @@ private void UpdateTriggerData(InteractionSourceState interactionSourceState, Mi
...
+                                }
                             }
                             else
                             {
+                                for (int i = 0; i < InputSource.Pointers.Length; i++)
+                                {
+                                    InputSystem?.RaisePointerClicked(InputSource.Pointers[i], ControllerHandedness, interactionMapping.MixedRealityInputAction, 1);
```
keveleigh: The more I think about this change, the less certain I am. It changes the way pointer events work.

At the moment, pointers can be assigned an action to fire their pointer events. This becomes useful for, for example, a sphere-based grab pointer using the grip action. Now, only the trigger can fire those events, as we're explicitly firing them based on the `selectPressed` state.

At the very least, this change is unrelated to gestures, and I'd prefer not to block gestures on it.

StephenHodgson:
Bump

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
StephenHodgson: >**Removed Pointer Up/Down from all scripts raising events outside their respective device manager/controller classes.**

StephenHodgson:
Removed pointer changes and broke out breaking changes into https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2781 which now blocks this PR

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -87,19 +353,22 @@ public override void Disable()
         /// Retrieve the source controller from the Active Store, or create a new device and register it
         /// </summary>
         /// <param name="interactionSourceState">Source State provided by the SDK</param>
+        /// <param name="addController">Should the Source be added as a controller if it isn't found?</param>
         /// <returns>New or Existing Controller Input Source</returns>
-        private WindowsMixedRealityController GetOrAddController(InteractionSourceState interactionSourceState)
+        private WindowsMixedRealityController GetController(InteractionSource interactionSourceState, bool addController = true)
```
keveleigh: Should update the parameter name too, to match the new type.

StephenHodgson: Will do!

PR #2715, opened 2018-09-06
Shader demo texture file size reduction

maluoi:
Reduced file sizes of textures in the StandardShader examples from **76.5mb down to 4.5mb** with little perceptual difference. All assets should be visible from the MaterialGallery scene.

Switched a couple of .tga files to .png, reduced dimensions from 2048 down to smaller sizes where appropriate, and used web image compression tools to get better filesizes from the pngs. There is loss of detail, but there's still plenty enough for this context, and I'll be surprised if you can spot it! Materials are all hooked up to new textures appropriately.

This significantly reduces the project's overall filesize, and should make development experience nicer in certain contexts!


StephenHodgson:
Awesome! The only concern here is that all the texture guids changed.

maluoi:
I did string search the StandardShader folder for guid references to ensure they were changed correctly, no old references etc. Didn't even think about doing guid surgery on the new meta files though, does that actually work? I can do that if you need it.

StephenHodgson:
I think it'll be okay as long as people were just using the materials. If people were using the textures directly then it'll break. We're still early in development, so I think this is okay.

maluoi:
Kewl, I'll keep that in mind for later then, that's actually a pretty nice idea/trick! :)

Cameron-Micka:
@maluoi Thanks Nick for these changes!

PR #2727, opened 2018-09-07
[Solver PR 1] Add some prototyping models from HTK

keveleigh:
Overview
---
I'm breaking down my solver port into several smaller PRs. This is PR 1.

This brings over a subset of the prototyping models we shipped in the HTK examples folder, to be used in the Solvers example scene.


StephenHodgson:
Do we need to import all these things?
image

Assets/MixedRealityToolkit-Examples/StandardAssets/Models/Prototyping/Torus.fbx.meta
```diff
@@ -45,8 +45,8 @@ ModelImporter:
     addColliders: 0
     importVisibility: 1
```
StephenHodgson: We can probably get rid of blend shapes and visibility as well.

PR #2728, opened 2018-09-07
[Solver PR 2] Move solvers from Core into SDK

keveleigh:
Overview
---
Moves the solvers from Core into the SDK layer. This seems like a more natural fit for them, as they're MonoBehaviours and are components that build off of the Core layer.


keveleigh:
>We need to keep the solvers in Core

Why do we need to keep solvers in Core?

StephenHodgson:
Because the Lines depend on them, and they also should stay in Core, as Pointers depend on Lines.

keveleigh:
Which lines depend on solvers?

StephenHodgson:
> Which lines depend on solvers?

All of them. I suppose the real question is why do you want to move it into the SDK?

keveleigh:
>All of them.

Can you show me the lines of code where they depend on Solvers? I did not have to update any code, and the demo scene and pointers all still work.

>I suppose the real question is why do you want to move it into the SDK?

They make more sense in the SDK. These are MonoBehaviour scripts that developers will need to put into their scene in order to use. That's basically the definition of the SDK layer, not the internal/core MRTK layer. If the Core is distributed as a DLL, is it possible to consume MonoBehaviours that have been included?

StephenHodgson:
> If the Core is distributed as a DLL, is it even possible to consume MonoBehaviours that have been included?

Yes, but it's not ideal.

StephenHodgson:
> Can you show me the lines of code where they depend on Solvers? 

I'm surprised you didn't get compile errors after moving them

StephenHodgson:
Oh wait... maybe it was distorters I was thinking of.

StephenHodgson:
While, yes these are generally going to be MonoBehaviours, I think it makes sense to keep them in core as they're used in important physics calculations but I suppose they could potentially be made into an interface if we want to abstract it out more.

keveleigh:
>Yes, but it's not ideal.

Then it sounds like we shouldn't include any MonoBehaviours in the Core that developers need to add to the scene. We need to make the Toolkit as easy as possible to be consumed.

>I think it makes sense to keep them in core as they're used in important physics calculations

If we want to bring out certain physics calculations into the core, that sounds like a great future task to talk about. As-is, nothing in the core depends on or uses Solvers, and I'm not sure it makes sense for them to.

StephenHodgson:
> We need to make the Toolkit as easy as possible to be consumed.

Agreed That's def my mantra.

StephenHodgson:
> If we want to pull some primitive physics pieces back into the Core.

But it's already in the core. I have a feeling this is a bad idea. Can we at least keep the physics namespace?

Issue #2716, opened 2018-09-06
vNextTask: Add debug panel in app


ryzngard:
# Overview
Provide a panel for developers to use to gain valuable information for debugging at runtime (e.g fps) 

# Requirements
* Display while running in app
* Can be toggled on/off
* Does not require a prefab
* Displays fps
* Investigate other useful information

# Acceptance Criteria
A list of the specific things that must be done for approval / acceptance of the task in checklist form. 
This may correspond with the above requirements.

- [x] Does not require scene modification
- [x] Easily toggleable
- [ ] Minimum performance impact

PR #2740, opened 2018-09-10
Add scaffolding for diagnostic profile and display

ryzngard:
Overview
---

The intent is to provide the layout for what configuration of diagnostic display will be. This does not include any UI elements yet, but still has the layout there.

Currently supported diagnostic information:

* FPS
* Memory
* CPU

Each diagnostic is only allowed to be turned on or off. Granularity of information will be determined as implementation for the display is added. 

Changes
---
- #2716


Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
```diff
@@ -0,0 +1,26 @@
...
+using UnityEngine;
+
+namespace Assets.MixedRealityToolkit_SDK.Features.Diagnostics
+{
+    public class DiagnosticBehavior : MonoBehaviour
+    {
+        // Use this for initialization
+        void Start()
+        {
+
```
StephenHodgson: Let's remove empty methods

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,243 @@
...
+            {
+                return showCpu;
+            }
+
+            set
+            {
+                if (value != showCpu)
+                {
+                    showCpu = value;
+                    var component = GetDiagnosticVisualization()?.GetComponent<DiagnosticBehavior>();
```
StephenHodgson: can we cache some of these `GetComponent` calls? They're pretty expensive, and we can always cache it when we set it up in enable

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -215,6 +217,18 @@ public float TeleportDuration
         }
 
         [SerializeField]
+        [Tooltip("Profile for wiring up diagnostic assets.")]
+        private MixedRealityDiagnosticsProfile diagnosticsProfile;
+
+        /// <summary>
+        /// Active profile for diagnostic configuration
+        /// </summary>
+        public MixedRealityDiagnosticsProfile DiagnosticsProfile
```
StephenHodgson: Do we want to add Diagnostics to the core managers?

Or should we make it an optional component?

StephenHodgson:
Looks good so far! great work!

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Assets.MixedRealityToolkit_SDK.Features.Diagnostics
+{
+    public class DiagnosticBehavior : MonoBehaviour
```
StephenHodgson: shouldn't this have a `IMixedRealityDiagnosticsHandler`?

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
ryzngard: Definitely. I'll make the component a member variable too to save the reference. Just need to be careful with lifetime I think

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,243 @@
...
+                    if (component != null)
+                    {
+                        component.ShowMemory = value;
+                    }
+
+                    RaiseDiagnosticsChanged();
+                }
+            }
+        }
+        #endregion
```
StephenHodgson: let's add the name of the region to the end as well please

`#endregion IMixedRealityDiagnosticsManager`

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
ryzngard: Should it? I assumed since it's an internal behavior that the manager driving it was sufficient. Maybe it makes sense to hook up with the interface though.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            if (!Visible)
+            {
+                // Don't create a gameobject if it's not needed
+                return null;
+            }
+
+            diagnosticVisualization = GameObject.CreatePrimitive(PrimitiveType.Quad);
+            diagnosticVisualization.name = "Diagnostics";
+            diagnosticVisualization.layer = 2; // magic?
```
StephenHodgson: you can also set it using `Physics.<LayerConstantValue>`

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
StephenHodgson: Well I noticed you're just setting it outright. bc you have a reference to it. 

But that would also remove the need to keep a cached reference to this component after you add it to your Quad

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
StephenHodgson: If you use the handler, you won't need to keep a cached value. Just raise the event.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
davidkline-ms: this is a scaffolding PR, I am ok with leaving them until we know if they are needed.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
davidkline-ms: given the value to devs, core makes sense.

considering that it will likely be disabled for retail builds, i can see it being optional

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
StephenHodgson: They're just the default Unity ones. Let's take'em out.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
StephenHodgson: How are we disabling it for production builds?

Assets/MixedRealityToolkit/_Core/Interfaces/Diagnostics/IMixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public interface IMixedRealityDiagnosticsManager : IMixedRealityEventSystem, IMixedRealityEventSource
+    {
+        /// <summary>
+        /// Enable / disable the diagnostic display
+        /// </summary>
+        bool Visible { get; set; }
```
StephenHodgson: We prob don't want people setting these values, let's remove all the setters for the interface.

Assets/MixedRealityToolkit/_Core/Interfaces/Diagnostics/IMixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Enable / disable memory profiling when the diagnostic panel is visible. 
+        /// </summary>
+        bool ShowMemory { get; set; }
+
+        /// <summary>
+        /// Gets the <see cref="GameObject"/> that represents the diagnostic visualization
+        /// </summary>
+        /// <returns>The diagnostic visualization object, or null if one does not exist</returns>
+        GameObject GetDiagnosticVisualization();
```
StephenHodgson: This could be a property

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
ryzngard: Got it, that seems much cleaner.

Assets/MixedRealityToolkit/_Core/Interfaces/Diagnostics/IMixedRealityDiagnosticsManager.cs
ryzngard: It fits the pay-to-play mentality. Diagnostics aren't free (although likely not that expensive), so if you only want to see CPU usage or FPS it would be best to only allow showing that. I'd prefer to keep the knobs open if possible

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
ryzngard: If it's left in production, the overhead is memory and binary size right? Without enabling the profile, there should be no work being done. 

Could add an `#if DEBUG` check, or the equivalent for whatever works in unity.

Assets/MixedRealityToolkit/_Core/Interfaces/Diagnostics/IMixedRealityDiagnosticsManager.cs
StephenHodgson: Agreed, but we do keep the knobs open by allowing devs to write their own implementation using the interface. We don't want another system outside of the diagnostics chaining values on us.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
StephenHodgson: `# UNITY_EDITOR`

StephenHodgson: That's partly why I wondered if we should just register this as a `IMixedRealityComponent`

Assets/MixedRealityToolkit/_Core/Interfaces/Diagnostics/IMixedRealityDiagnosticsManager.cs
ryzngard: I think I misunderstood the data binding between UI and manager implementations. Will set to readonly. Agree we don't want random systems messing with this

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,243 @@
...
+                return diagnosticVisualization;
+            }
+
+            if (!Visible)
+            {
+                // Don't create a gameobject if it's not needed
+                return null;
+            }
+
+            diagnosticVisualization = GameObject.CreatePrimitive(PrimitiveType.Quad);
```
StephenHodgson: Quad's aren't usually aligned properly, to face the camera (cause they're forward is perpendicular to their face). Are we going to add something for this later?

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
ryzngard: I don't really see a _bad_ reason to allow people to enable this in production though. For some games there's fps, memory, and cpu indicators accessible to users to help them monitor what's going wrong if something does go wrong. While I don't think it's our job to tell devs what information is best for this, having this available as an easy entry point doesn't hurt. 

It's good to know we can disable it completely for release builds. I'm not sure on policy of should we, but my gut says to leave it open and reduce surface area later if needed.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
ryzngard: Probably. This is my first dive into unity, so I just put something in that looked like it would work. Is there a primitive that makes more sense?

StephenHodgson: Nah a quad is fine. I think it's okay for now until we do the UX pass. Just need to keep it in mind when we cross that bridge.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
```diff
@@ -1,26 +1,24 @@
...
-
-        }
-
-        // Update is called once per frame
-        void Update()
-        {
-
-        }
-
         public bool ShowCpu { get; set; }
```
StephenHodgson: Do these need to be public still? or set able?

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
```diff
@@ -31,7 +31,7 @@ private void InitializeInternal()
 
             if (Visible)
             {
-                GetDiagnosticVisualization();
+                DiagnosticVisualization.SetActive(Visible);
```
StephenHodgson: `DiagnosticVisualization?.SetActive(Visible);`?

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
ryzngard: Nope, good catch

ryzngard: Actually*: Needed to set initial values, since I don't see a way to pass as constructor arguments. Looking through documentation now.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
ryzngard: Shouldn't be needed, although it's not clearly outlined why. If visible is true then the getter will always create if it's null. I'll add a comment to clarify the expectation in behavior.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
StephenHodgson: Mono's don't have constructors. you can set it inline:

`public bool ShowCpu { get; set; } = false;`

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
StephenHodgson: A few lines below you were also checking null ability. might want to remove that one then as well?

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,228 @@
...
+                component.ShowMemory = ShowMemory;
+
+                return diagnosticVisualization;
+            }
+        }
+        #endregion IMixedRealityDiagnosticsManager
+
+        #region IMixedRealityEventSource
+        /// <inheritdoc />
+        public uint SourceId => 0;
```
StephenHodgson: 0 means it's unset. I'm not sure, but we may want to get a unique id for this somehow.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
ryzngard: I'm talking about from the manager on creation.

```
var component = diagnosticVisualization.GetComponent<DiagnosticBehavior>();
component.ShowCpu = ShowCpu;
component.ShowFps = ShowFps;
component.ShowMemory = ShowMemory;

```


Maybe it would be better to rely on the eventing to pass those values? I believe everywhere the property could be created is followed by a `RaiseDiagnosticsChanged()`

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,228 @@
...
+            {
+                return visible;
+            }
+
+            set
+            {
+                if (value != visible)
+                {
+                    visible = value;
+                    DiagnosticVisualization?.SetActive(value);
```
StephenHodgson: here.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
StephenHodgson: Yeah as long as you call `RaiseDiagnosticsChanged` after every place you could create it, you should be okay.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
ryzngard: Good point. It was originally designed for if going from true > false, but that would imply that the default value (false) was switched to true before then. I'll remove the check here.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
ryzngard: I'm going to resolve this and keep as is for now. Feel free to reopen if you feel there's more to talk about.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
ryzngard: This whole if block can be removed. The visibility property will correctly handle setting active, no need to do it again.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MixedRealityDiagnosticsManager.cs
ryzngard: @davidkline-ms recommended using `SourceName.GetHashCode()` as a good way to tie a unique enough value

StephenHodgson: Sounds good to me

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Core.EventDatum.Boundary;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.Diagnostics;
+using UnityEngine;
+
+namespace Assets.MixedRealityToolkit_SDK.Features.Diagnostics
+{
+    public class DiagnosticBehavior : MonoBehaviour, IMixedRealityDiagnosticsHandler
```
StephenHodgson: I forgotten one thing. We need to register this game object to get events

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityConfigurationProfile.cs
```diff
@@ -214,6 +217,43 @@ public float TeleportDuration
...
+
+        public bool IsDiagnosticsSystemEnabled
+        {
+            get { return enableDiagnosticsSystem && DiagnosticsSystemSystemType?.Type != null; }
+            private set { enableDiagnosticsSystem = value; }
+        }
+
+        [SerializeField]
+        [Tooltip("Diagnostics System Class to instantiate at runtime.")]
+        [Implements(typeof(IMixedRealityDiagnosticsManager), TypeGrouping.ByNamespaceFlat)]
```
StephenHodgson: We should also probably add this to the IsCoreManager check in the MixedRealityManager class

PR #2741, opened 2018-09-10
simplify the platform check logic, make it easier to extend

davidkline-ms:
Overview
---
The existing platform check had an early out that didn't quite feel right to me after going back and looking at it again.

There was also multiple very specific checks for individual flags values.

Modified the code to be easier to manage in the future

- add to the SupportedPlatforms enum
- update the conversion methods

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Utilities/PlatformUtility.cs
+++ b/Assets/MixedRealityToolkit/_Core/Utilities/PlatformUtility.cs
@@ -8,59 +8,78 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities
 {
     public static class PlatformUtility
     {
-        public static bool IsPlatformSupported(this RuntimePlatform runtimePlatform, SupportedPlatforms platform)
+        public static bool IsPlatformSupported(this RuntimePlatform runtimePlatform, SupportedPlatforms platforms)
         {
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/_Core/Utilities/PlatformUtility.cs
```diff
@@ -8,59 +8,78 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities
...
+        private static SupportedPlatforms GetSupportedPlatformMask(RuntimePlatform runtimePlatform)
+        {
+            SupportedPlatforms supportedPlatforms = 0;
 
             switch (runtimePlatform)
             {
                 case RuntimePlatform.WindowsPlayer:
                 case RuntimePlatform.WindowsEditor:
-                    return (platform & SupportedPlatforms.WindowsStandalone) != 0;
+                    supportedPlatforms |= SupportedPlatforms.WindowsStandalone;
```
ryzngard: Is there ever a case where a runtimeplatform maps to more than one supported platform? If not, why not just a return without a local variable here?

StephenHodgson: No, unity only ever has one platform set at a time.

PR #2742, opened 2018-09-10
Fix GetManager(s) calls in MixedRealityManager

keveleigh:
Overview
---
After #2718, only one GetManager call was updated to actually check all interfaces. The others simply kept checking against the registered type, which is now IMixedRealityComponent for all components (causing them to return empty).

Basically, this is an expansion of the code added in e710233caeadf9238fb180a061e589fa61dfeadb to all component/manager getters.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
+++ b/Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
@@ -724,48 +724,11 @@ namespace Microsoft.MixedReality.Toolkit.Core.Managers
                 // If no name provided, return all components of the same type. Else return the type/name combination.
                 if (string.IsNullOrWhiteSpace(managerName))
                 {
-                    for (int i = 0; i < mixedRealityComponentsCount; i++)
-                    {
-                        if (MixedRealityComponents[i].Item1.Name == type.Name ||
-                            MixedRealityComponents[i].Item2.GetType().Name == type.Name)
-                        {
-                            managers.Add(MixedRealityComponents[i].Item2);
-                            continue;
-                        }
-
-                        var interfaces = MixedRealityComponents[i].Item2.GetType().GetInterfaces();
-                        for (int j = 0; j < interfaces.Length; j++)
-                        {
-                            if (interfaces[j].Name == type.Name)
-                            {
-                                managers.Add(MixedRealityComponents[i].Item2);
-                                break;
-                            }
-                        }
-                    }
+                    GetComponentsByType(type, ref managers);
                 }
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -942,6 +967,17 @@ private void GetComponentByTypeAndName(Type type, string managerName, out IMixed
                     manager = MixedRealityComponents[i].Item2;
                     break;
                 }
+
+                var interfaces = MixedRealityComponents[i].Item2.GetType().GetInterfaces();
```
StephenHodgson: Seems like the same lines over and over again. Could we make this into it's own thing?

keveleigh: Probably

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -955,33 +893,67 @@ private void GetComponentByType(Type type, out IMixedRealityManager manager)
...
-                    MixedRealityComponents[i].Item2.Name == managerName)
+                if (CheckComponentMatch(type, managerName, MixedRealityComponents[i]))
                 {
                     manager = MixedRealityComponents[i].Item2;
                     break;
                 }
+            }
+        }
+
+        private void GetComponentsByType(Type type, ref List<IMixedRealityManager> managers)
```
StephenHodgson: if we modify the list while we get a component, will that cause problems?

keveleigh: The managers list being passed in? It's created in the original method call and not passed on until the return. I'm not sure where else it could be modified.

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -930,22 +893,67 @@ private void GetComponentByType(Type type, out IMixedRealityManager manager)
...
+            else
+            {
+                var interfaces = managerTuple.Item2.GetType().GetInterfaces();
+                for (int j = 0; j < interfaces.Length; j++)
+                {
+                    if (interfaces[j].Name == type.Name &&
+                        (string.IsNullOrEmpty(managerName) ? true :
+                        managerTuple.Item2.Name == managerName))
+                    {
+                        return true;
```
StephenHodgson: Wait, do we wanna break here?

PR #2755, opened 2018-09-11
removed spatial mouse for now

StephenHodgson:
Overview
---
spatial mouse still needs some work. Opt-in for now.


StephenHodgson:
I do agree though, we should show a notification on play mode start that mouse control can be regained by pressing "Esc"

PR #2758, opened 2018-09-11
Proposing requiring all enums to end with the Type suffix

Ecnassianer:
Added a rule indicating that enums should end with Type, in response to the discussion here:
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2606#issuecomment-420322751

Updated the other examples to follow this rule.


Ecnassianer:
Oops, thought that's what I was doing. My mistake.

PR #2760, opened 2018-09-11
Proposing requiring all enums to end with the Type suffix

Ecnassianer:
Added a rule indicating that enums should end with Type, in response to the discussion here:
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2606#issuecomment-420322751

Updated the other examples to follow this rule.

### DIFF
--- a/CodingGuidelines.md
+++ b/CodingGuidelines.md
@@ -249,7 +249,7 @@ To ensure all Enum's are initialized correctly starting at 0, .NET gives you a t
 ### Don't:
 
 ```c#
-public enum MyEnum
+public enum Value
 {
     Value1, // <- no initializer
     Value2,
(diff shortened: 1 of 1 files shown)


Ecnassianer:
XD

davidkline-ms:
mrtk_development… htk is not going to be changing.

sorry for the confusion

Ecnassianer:
I cannot read today.

CodingGuidelines.md
```diff
@@ -315,6 +315,36 @@ public enum AudioFormat
...
+    Second,
+    Third
+}
+```
+### Do:
+```c#
+[Flags]
+public enum OrderingType
+{
+    First = 0,
```
StephenHodgson: Wait, if we have flags don't we want to use bit shifting?

PR #2771, opened 2018-09-12
vNEXT: Input Action Rules

StephenHodgson:
Overview
---
Input Action Rules for transforming actions.

image

It's important to note, that rules do not "re raise" or "re throw" events, and that they simply replace the action raised with the new action defined in the rule, and do not modify the event data in any other way.

For example, events raised for `Player Movement` will only replace `Player Movement` action with the new action defined (such as Up, Down, Left, or Right), and do not re raise the event. As such rules for Up/Down/Left/Right need to raise the event for the rule to come into effect.

Changes
---
- Fixes: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2610

TODO
---
- [x] Update Inspector to be more friendly
- [x] Filter input action rules by Base Rule Axis Constraint


Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityInputActionRulesInspector.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            if (GUILayout.Button("Back to Input Profile"))
+            {
+                Selection.activeObject = MixedRealityManager.Instance.ActiveProfile.InputSystemProfile;
+            }
+
+            EditorGUILayout.Space();
+            EditorGUILayout.LabelField("Input Action Rules Profile", EditorStyles.boldLabel);
+            EditorGUILayout.HelpBox("The Input Action Rules...", MessageType.Info);
```
StephenHodgson: Looking for a better description.  Suggestions welcome.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -1702,5 +1743,156 @@ public void RaiseDictationError(IMixedRealityInputSource source, string dictatio
...
+        {
+            if (CurrentInputActionRulesProfile != null)
+            {
+                var digitalRules = CurrentInputActionRulesProfile.InputActionRulesSingleAxis;
+
+                for (int i = 0; i < digitalRules.Length; i++)
+                {
+                    if (digitalRules[i].BaseAction == inputAction && digitalRules[i].Criteria.Equals(criteria))
+                    {
+                        if (digitalRules[i].RuleAction == inputAction)
```
SimonDarksideJ: Don't think we need this check.  It could be possible to configure a rule with the same action.
It's up to the user at the end of the day how they configure it.

Maybe only log a warning, just to inform them.   But better that is done in the inspector rather than here.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -1702,5 +1743,156 @@ public void RaiseDictationError(IMixedRealityInputSource source, string dictatio
...
+                    }
+                }
+            }
+
+            return inputAction;
+        }
+
+        private MixedRealityInputAction ProcessRules(MixedRealityInputAction inputAction, Quaternion criteria)
+        {
+            if (CurrentInputActionRulesProfile != null)
```
SimonDarksideJ: A lot of this is the same code for each type. Possibly there is a way to genericise it, maybe?

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/InputActionRuleDigital.cs
```diff
@@ -0,0 +1,50 @@
...
+using System;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Definitions.InputSystem
+{
+    /// <summary>
+    /// Generic Input Action Rule for raising actions based on specific criteria.
+    /// </summary>
+    [Serializable]
+    public struct InputActionRuleDigital : IInputActionRule<bool>
```
SimonDarksideJ: I love the reuse of our original pattern :D

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionRulesProfile.cs
```diff
@@ -0,0 +1,60 @@
...
+using Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Definitions.InputSystem
+{
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Mixed Reality Input Action Rules Profile", fileName = "MixedRealityInputActionRulesProfile", order = (int)CreateProfileMenuItemIndices.InputActionRules)]
+    public class MixedRealityInputActionRulesProfile : ScriptableObject
+    {
+        [SerializeField]
+        private InputActionRuleDigital[] inputActionRulesDigital;
```
SimonDarksideJ: Will need to check these in the inspector.  But would be better if this was hidden, like with the interaction definitions.  So the user simply selects an action and the rule is stored based on the axis of that action maybe?

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: Well, if we don't check it here and someone sets it up grammatically it's an infinite loop

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: I tried, but unfortunately none that was immediately obvious. If you figure out a way I'm happy to accept a PR

Assets/MixedRealityToolkit/_Core/Definitions/InputSystem/MixedRealityInputActionRulesProfile.cs
StephenHodgson: I'm not sure I understand.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: You shouldn't be raising an action you're consuming.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
Jarodshow: oops, should this get reverted?

StephenHodgson: I'm not even sure what changed. It should be okay.

StephenHodgson:
Correct. Sorry I took a long weekend to unplug. I'll try to update this soon, but it is functional. Just needs some TLC on the inspector.

Jarodshow:
@StephenHodgson Any update on this? I have an inspector implementation if you'd like to see it.

StephenHodgson:
I've been super busy as of late. Open a PR on my branch and let's take a look!

Jarodshow:
[StephenHodgson#33](https://github.com/StephenHodgson/MixedRealityToolkit-Unity/pull/33)

davidkline-ms:
What is the status on this? Beta is getting clos!! :o

Jarodshow:
almost done!

StephenHodgson:
This is ready for review. @SimonDarksideJ @Jarodshow @davidkline-ms

StephenHodgson:
>can we align the fields to make them more uniform across all the different types.

They're already aligned. I'm not sure why Unity's default label width for dinner are different. Fixing it isn't trivial.

> If I can't configure it, then maybe the action should be filtered from the list. I found this confusing

How is it confusing? It tells you exactly why. But I agree, we should just filter it out.

>Needs to add a comment in the Help Text, that mapping can only be done between actions of the same axis 

I figured it'd be pretty obvious since we filter them out

> Able to create circular actions

I'm not sure how we could possible check against that. Seems rather obvious you shouldn't do that though.

> Would it be possible to have a separator / group per Base Input Action? 

That's why I added the fold outs. 😅

StephenHodgson:
@SimonDarksideJ After looking into the Circular action references, this is actually not a problem. We don't "re-throw" the action, but instead replace the action in the Raise methods. So it's actually not circular at all.

In fact, if you wanted to do a "reverse" transform rule, it could be helpful.

StephenHodgson:
>Duplicate checking is not testing value. E.G. I should be possible to have a rule that uses the same output action for a different value.

What are your repro steps?

I made it so you couldn't register actions with the same base and rule.

StephenHodgson:
Oh I see, nvm. Yeah at the moment we're only checking if the actions are the same.

Issue #2719, opened 2018-09-07
Proposal: Update rename "Body" to "MixedRealityPlayspace", and add an accessor to MixedRealityManager


Ecnassianer:
## Overview

Right now MRTK creates a "Body" game object which contains a camera. The camera's position is determined by a combination of headset tracking and teleporting. Though the name "Body" seems to imply that it is either a virtual body (positioned beneath the headset) or a tracked object that represents the body (e.g, positioned by a vive tracker puck), but it is neither. The "Body" position is determined by teleporting. Furthermore, it seems to mostly exist sitting around somewhere on the real world ground, something that "Bodies" usually don't do.

## Step 1) Give Body a better name

We should rename it MixedRealityPlayspace.

**Feel free to skip to the next section.** Here's an interesting philosophy to think about. The MixedRealityPlayspace GameObject actually represents a tracker agnostic reference point to convert from arbitrary tracker spaces into Unity World Space Coordinates. It essentially defines a "Reality Space Coordinate System". 

In the case of a stationary tracker, like a Vive Lighthouse, 0,0,0 in tracker space happens to always line up exactly with Reality Space 0,0,0. It would also line up exactly with Unity World Space 0,0,0, except that when you teleport they get offset by the distance you teleported. So why do we need a Reality Space Coordinate System if always lines up with the position the tracker is reporting?

Consider non-stationary trackers, like Oculus Go, HoloLens, or a headset mounted Leap Motion. These devices might start up anywhere in their playspace, and they're always tracking relative to something else (in most cases, relative to the headset). In the case of non-visible WMR controllers, they're essentially tracked relative to where they were last seen by the headset. So we have to combine a whole host of inconsistent tracker spaces; and these tracker spaces may even move relative to the Unity World Space. (Picture graph paper pinned to a leap motion which is attached to a headset as it moves through Unity World Space).

None of these tracker spaces are aware of each other, or share any kind of common reference. So we have to choose an arbitrary point (or really, we let the platform choose, but from an MRTK user's perspective it's arbitrary), and calibrate each tracker to that point. That way controllers, hands, cameras, bodies, pucks, and world anchors all appear to align correctly, and when we teleport, they all receive the teleport's offset uniformly.

## Step 2) Add an accessor to MixedRealityManager

[...]

PR #2773, opened 2018-09-12
replacing "Body" with MixedRealityPlayspace

Ecnassianer:
Proposal here:
https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2719

I touched several systems I'm not at all familiar with, so if you want to give this one a little extra attention, I'd appreciate it.

This has a bit of a weird quirk that if an existing scene has a Body that contains a Camera, it's basically ignored. I figured this was acceptable behavior in order for existing content wouldn't break. It's just strange because now there's a Body object hanging around doing nothing.

I haven't updated the example scenes that use a Body in the scene; I'm going to do that in a follow up PR. 

Teleporting is intermittently broken due to the ParabolicTeleportPointer has not been registered bug. This is unrelated to my code, and some combination of restarting makes it go away. There's an issue/PR tracking that problem somewhere else.


Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -91,12 +91,6 @@ public override void Initialize()
 
         private void InitializeInternal()
         {
-            if (CameraCache.Main.transform.parent == null)
```
StephenHodgson: Please revert. 

Manager's should take care of setting up what they need, if it's not done so for them already.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
```diff
@@ -40,16 +40,6 @@ public override void Initialize()
 
         private void InitializeInternal()
         {
-            if (CameraCache.Main.transform.parent == null)
-            {
-                var cameraParent = new GameObject("Body");
```
StephenHodgson: Same thing here.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
```diff
@@ -44,7 +44,7 @@ private void Awake()
         {
             markerParent = new GameObject();
             markerParent.name = "Boundary Demo Markers";
-            markerParent.transform.parent = CameraCache.Main.transform.parent;
+            markerParent.transform.parent = MixedRealityManager.Instance.MixedRealityPlayspace.transform;
```
StephenHodgson: Why not just make the `MixedRealityPlayspace` a transform?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -341,6 +341,36 @@ private void InitializeInternal()
             }
         }
 
+        private GameObject mixedRealityPlayspace;
+
+        /// <summary>
+        /// Returns the MixedRealityPlayspace for the local player
+        /// </summary>
+        public GameObject MixedRealityPlayspace
```
StephenHodgson: Let's make the a transform instead as we use it the most.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: Or at least call the property to make sure it creates what it needs.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
```diff
@@ -40,16 +40,6 @@ public override void Initialize()
...
-                CameraCache.Main.transform.SetParent(cameraParent.transform);
-            }
-
-            // Make sure the camera is at the scene origin.
-            CameraCache.Main.transform.parent.transform.position = Vector3.zero;
```
StephenHodgson: Do we ensure this someplace else?

Ecnassianer: This was the question I was asking about our assumptions of whether tracker space 0,0,0 always overlaps world space 0,0,0 when the scene starts.
https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2719#issuecomment-420449199

I found a problem where the controllers were offset improperly and fixed it. I wasn't able to find any other problems with not moving the headset, so I removed this line. We will definitely hit problems if we zero out the offset between the headset and the Playspace.

StephenHodgson: In either case the camera itself should start at the origin. There's lots of documented cases that if it doesn't start there, things break with WMR specifically.

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
SimonDarksideJ: Agreed

Ecnassianer:
Sorry I haven't had any time to revisit this, I got pulled away on some urgent work last week and it hasn't been on my mind. I'll update again soon.

Assets/MixedRealityToolkit-SDK/Features/Teleportation/MixedRealityTeleportManager.cs
Ecnassianer: Right, that's why I brought up the assumption in the other thread. We should be doing this somewhere else, but I don't know where.

It is totally reasonable for us to require that Camera to be at 0,0,0 in the scene before everything starts, I'm not disagreeing with that, but once we've offset the camera via tracking we can't arbitrarily zero it out in the TeleportManager.

If the scene starts and the physical headeset is 5 meters to the left of the center of the tracking space it should not be at world space 0,0,0 on the first frame its view renders. It should be at something like 500,0,0. (That doesn't preclude it being at 0,0,0 before tracking moves it the first time though).

Look at it the other way: If a developer doesn't want to use our teleport manager, they disable it, this code never runs. That doesn't mean we still don't need to enforce a 0,0,0 camera in the scene, so we need that code to live somewhere else. If we're enforcing it somewhere else, we don't need it here.

StephenHodgson: Maybe in the mixed reality manager, when we setup the camera profile?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -341,6 +349,36 @@ private void InitializeInternal()
...
+        {
+            get {
+                AssertIsInitialized();
+                if (mixedRealityPlayspace)
+                {
+                    return mixedRealityPlayspace;
+                }
+                else
+                {
+                    mixedRealityPlayspace = new GameObject("MixedRealityPlayspace").transform;
```
keveleigh: Previously, we c[hecked to see if the parent was null](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2773/files#diff-0f24968827a5d17ade2a044a8cddf8bcL43) before creating a new parent.

We should probably do that here too, otherwise we might create a second gameobject and leave the original parent elsewhere in the scene.

Is that accounted for anywhere I missed?

Ecnassianer: It wasn't, though it was sort of by design. My intention was to follow up with another PR that handles situations like that more robustly, perhaps even giving a warning when compiling. Would you be satisfied merging this if I created a task to track better handle incorrect parents and worked on it next week?

Ecnassianer: Went ahead and created the task whether it goes in this PR or not:
https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2850

StephenHodgson: This is a pretty important feature to our one click setup.

I am definitely not for just merging it and revisiting it next week.

SimonDarksideJ:
If we have a PR visiting this area, we should attempt to resolve any outstanding queries with that feature (while the operation is still underway)

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -229,6 +230,13 @@ private void Initialize()
             isMixedRealityManagerInitializing = false;
         }
 
+        private void EnsureMixedRealityRequirements()
```
StephenHodgson: I feel like this method is a bit too vague to be in the Mixed Reality manager.

Issue #2503, opened 2018-07-28
vNext Task: Port Audio Influencers from HTK


davidkline-ms:
## Overview
As part of bringing vNext to feature parity with the HoloToolkit, port the Audio Influencers scripts and demo scenes.

## Requires
Creating an SDK Feature folder for Audio.
Updating the existing HoloToolkit scripts (namespaces, remove obsolete methods, etc)
Example audio files (not as standard assets) that illustrate the specific influencer effect
Updated demo scenes (use standard asset materials, improve description panel, etc)

## Acceptance Criteria
- [x] Scripts have appropriate names (ex: AudioInfluencerController)
- [x] Demo scenes use standard assets where appropriate

Issue #2592, opened 2018-08-14
vNext Task: Spatial Audio System


davidkline-ms:
## Overview
Different platforms require different spatializers. Some platforms are supported by more than one spatializer. Some may need additional configuration beyond simply selecting a plugin.

Developers should be presented with an intelligent default and the option of selecting the spatializer of their choice.

## Requires
- Interface to define the Spatial Audio System
- Default manager implementation
- Mixed Reality Manager updates to enable/disable the system and select the appropriate profile

## Acceptance criteria
- [ ] User can select and configure their chosen spatializer via a profile
- [ ] Correct Spatializer is used for the appropriate platform.

StephenHodgson:
Isn't there already a task for this?

davidkline-ms:
> Isn't there already a task for this?

Nope. The one you are thinking of is the replacement for UAudioManager.  This is to implement a system that choses intelligent defaults for the spatializer plugin (ex: you cannot use MS HRTF on non-Windows 10 platforms).

davidkline-ms:
moving to 2019 as it is a simple process to select the spatializer before buiding. and also hoping that Unity implements a build type configurable spatializer setting

davidkline-ms:
this link should be helpful when implementing this task

https://docs.unity3d.com/ScriptReference/AudioSettings.html

Yoyozilla:
This will be cover by future audio extension

PR #2775, opened 2018-09-12
port audio influencers and effects from HTK

davidkline-ms:
Overview
---
This brings in the audio influencer scripts, lo fi and occlusion effects from the HoloToolkit. Also adds example scenes for each effect.

Changes
---
- Add audio effect scripts to SDK\Features\Audio
- Update scripts to meet the coding guidelines
- Add example scenes
- Set MSHRTF as the default audio spatializer (supports Windows 10 platforms).
- Fixes: #2503


Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,297 @@
...
+            nativeLowPassCutoffFrequency = frequency;
+        }
+
+        /// <summary>
+        /// Gets the native cutoff frequency of the attached high pass filter.
+        /// </summary>
+        /// <returns>
+        /// The native cutoff frequency, or a neutral frequency if there is no high pass filter attached.
+        /// </returns>
+        public float GetNativeHighPassCutoffFrequency()
```
StephenHodgson: Also can be a property

Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/LoFiFilterSelection.cs
```diff
@@ -0,0 +1,117 @@
...
+
+    private AudioLoFiEffect loFiEffect;
+
+    void Start()
+    {
+        // Get the attached AudioLoFiEffect script.
+        loFiEffect = gameObject.GetComponent<AudioLoFiEffect>();
+        if (loFiEffect == null)
+        {
+            Debug.LogError("LoFiFilterSelection requires an AudioLoFiEffect to be attached to the game object.");
```
StephenHodgson: Why not add require this component attribute?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,297 @@
...
+                // Get the audio influencers that should apply to the audio source.
+                List<IAudioInfluencer> influencers = GetInfluencers();
+                foreach (IAudioInfluencer influencer in influencers)
+                {
+                    // Apply the influencer's effect.
+                    influencer.ApplyEffect(gameObject);
+                }
+
+                // Find and remove the audio influencers that are to be removed from the audio source.
+                List<IAudioInfluencer> influencersToRemove = new List<IAudioInfluencer>();
```
StephenHodgson: Do we want to make a new list each time?

Or clear the one we have and fill it again?

ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2018.1.9f1
+m_EditorVersion: 2018.1.9f2
```
StephenHodgson: Do we want this?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// <summary>
+    /// Class that implements <see cref="IAudioInfluencer"/> to provide an audio occlusion effect, similar
+    /// to listening to sound from outside of an enclosed space.
+    /// </summary>
+    /// <remarks>
+    /// Ensure that all sound emitting objects have an attached <see cref="AudioInfluencerController"/>. 
+    /// Failing to do so will result in the desired effect not being applied to the sound.
+    /// </remarks>
+    [DisallowMultipleComponent]
+    public class AudioOccluder : MonoBehaviour, IAudioInfluencer
```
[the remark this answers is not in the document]
StephenHodgson: We can probably make this a `IMixedRealityComponent`

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <summary>
+        /// Applies the audio effect.
+        /// </summary>
+        /// <param name="soundEmittingObject">The GameObject on which the effect is to be applied.</param>
+        public void ApplyEffect(GameObject soundEmittingObject)
+        {
+            if (!isActiveAndEnabled)
+            { return; }
+
+            AudioSource audioSource = soundEmittingObject.GetComponent<AudioSource>();
```
StephenHodgson: Why don't we cache these components on the `gameObject` and pass the `AudioSource`, `AudioLowPassFilter`, and any other components through as parameters to `ApplyEffect`?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            sourceQualityFilterSettings.Add(
+                AudioLoFiSourceQualityType.FullRange,
+                new AudioLoFiFilterSettings(10, 22000));
+            sourceQualityFilterSettings.Add(
+                AudioLoFiSourceQualityType.NarrowBandTelephony,
+                new AudioLoFiFilterSettings(300, 3400));
+            sourceQualityFilterSettings.Add(
+                AudioLoFiSourceQualityType.WideBandTelephony,
+                new AudioLoFiFilterSettings(50, 7000));
```
StephenHodgson: magic numbers?

ProjectSettings/ProjectVersion.txt
SimonDarksideJ: Revert this for now. Granted most of us are using f2, but we didn't agree to update it in the project yet. Plus it's not part of this PR

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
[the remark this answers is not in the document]
davidkline-ms: It needs to be a MonoBehaviour since it is attached to a game object.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: since callers iterate over what is returned, we should not change the contents. can cause exceptions if the caller uses foreach and potentially walking off the end of the collection / missing items in the case of for loops.

ProjectSettings/ProjectVersion.txt
davidkline-ms: do we seriously need a special PR to rev to the latest 2018.1? I can see a wider discussion for moving to .3

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
davidkline-ms: This was the pattern in HTK. i used to pass the audiosource and this ended up being simpler (more encapsulated / less prone to error)

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
davidkline-ms: they are specific frequency ranges for the audio source. there is no real reason to make consts out of them for this useage.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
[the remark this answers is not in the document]
davidkline-ms: the influencer manager asks each object between the user and the sound emitter what occlusion settings should be applied and to apply them

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
davidkline-ms: i added comments. i assumed (incorrectly?) that the quality type specified above clarified the question.

i dont think consts are that useful here since it is specific to the lo fi effect.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,297 @@
...
+        /// properties.
+        /// <summary>
+        private AudioLowPassFilter lowPassFilter;
+        private float nativeLowPassCutoffFrequency;
+        private AudioHighPassFilter highPassFilter;
+        private float nativeHighPassCutoffFrequency;
+
+        private void Awake() 
+        {           
+            audioSource = gameObject.GetComponent<AudioSource>();
```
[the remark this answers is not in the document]
davidkline-ms: Per Unity docs, OnValidate is editor only. We need these values cached for the player as well

"This function is called when the script is loaded or a value is changed in the inspector (Called in the editor only).
Use this function to validate the data of your MonoBehaviours. This can be used to ensure that when you modify data in an editor that the data stays within a certain range."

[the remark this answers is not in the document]
StephenHodgson: Right, but in the editor we would be serializing the value that the required component attached is. We can still check if it's null here before attempting to get it.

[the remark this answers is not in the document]
keveleigh: As long as the field is serialized and doesn't reset itself, it will embed the reference in the scene/prefab. It's safe for some cases, and this looks like one of those cases.

Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/LoFiFilterSelection.cs
```diff
@@ -0,0 +1,117 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Core.EventDatum.Input;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem.Handlers;
+using Microsoft.MixedReality.Toolkit.SDK.Audio.Influencers;
+using UnityEngine;
+
```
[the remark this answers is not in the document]
davidkline-ms: namespace added, GitHub did not auto-resolve

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: turned into a property... do not know why github did not autoresolve

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: this is especially problematic if a project is using a Coroutine and the collection gets changed during a yield

StephenHodgson: Let's tend to use async instead of coroutines... maybe we should write that up as well.

davidkline-ms: it is not for us to decide how client applications are written, :)

StephenHodgson: of course! lol,

but, internally we can def keep to using async over corountines

StephenHodgson:
Just need to pull this down and test.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// </summary>
+        /// <remarks>
+        /// MaxObjects can only be set in the Unity Inspector.
+        /// The MaxObjects range is 1 to 25, inclusive.
+        /// The default value is 10.
+        /// </remarks>
+        [Tooltip("Maximum number of objects that will be considered when looking for influencers.")]
+        [Range(1, 25)]
+        [SerializeField]
+        private int MaxObjects = 10;
```
StephenHodgson: pascalCase

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// sound emitter.
+        /// </summary>
+        public float NativeLowPassCutoffFrequency
+        {
+            get { return nativeLowPassCutoffFrequency; }
+            set
+            {
+                if (nativeLowPassCutoffFrequency != value)
+                {
+                    value = nativeLowPassCutoffFrequency;
```
StephenHodgson: Do we need to check it before setting it?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// sound emitter.
+        /// </summary>
+        public float NativeHighPassCutoffFrequency
+        {
+            get { return nativeHighPassCutoffFrequency; }
+            set
+            {
+                if (nativeHighPassCutoffFrequency != value)
+                {
+                    value = nativeHighPassCutoffFrequency;
```
StephenHodgson: Same question here.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+                }
+            }
+        }
+
+        private void Awake() 
+        {
```
StephenHodgson: there's some extra white space on this line

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
StephenHodgson: Just pulled your last commit, seems to still be a method

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+                if (influencer != null)
+                {
+                    influencers.Add(influencer);
+                }
+            }
+
+            return influencers;
+        }
+
+        /// Gets the native cutoff frequency of the attached high pass filter.
```
StephenHodgson: missing beginning summary tag.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,208 @@
...
+    [DisallowMultipleComponent]
+    public class AudioLoFiEffect : MonoBehaviour
+    {
+        /// <summary>
+        /// The quality level of the simulated audio source (ex: AM radio).
+        /// </summary>
+        [Tooltip("The quality level of the simulated audio source.")]
+        [SerializeField]
+        private AudioLoFiSourceQualityType sourceQuality;
+        public AudioLoFiSourceQualityType SourceQuality
```
StephenHodgson: missing tooltip summary

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,208 @@
...
+                LowPassCutoff = lowPassCutoff;
+            }
+
+            /// <summary>
+            /// Checks to see if two FilterSettings objects are equivalent.
+            /// </summary>
+            /// <returns>True if equivalent, false otherwise.</returns>
+            public static bool operator ==(AudioLoFiFilterSettings a, AudioLoFiFilterSettings b)
+            {
+                return a.Equals(b);
```
StephenHodgson: Boxing here. Did we implement our own equality check for `AudioLoFiFilterSettings` ?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,208 @@
...
+                    return false;
+                }
+
+                if (!(obj is AudioLoFiFilterSettings))
+                {
+                    return false;
+                }
+
+                AudioLoFiFilterSettings other = (AudioLoFiFilterSettings)obj;
+                if ((other.LowPassCutoff != LowPassCutoff) ||
```
StephenHodgson: `!other.LowPassCutoff.Equals(LowPassCutoff)` so we get proper floating point checks. (same for the line below)

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,208 @@
...
+                return true;
+            }
+
+            /// <summary>
+            /// Generates a hash code representing this FilterSettings.
+            /// </summary>
+            /// <returns></returns>
+            public override int GetHashCode()
+            {
+                string s = $"[{GetType().ToString()}] Low: {LowPassCutoff}, High: {HighPassCutoff}";
```
StephenHodgson: Is it okay that we're using non read only values to compute the hash code?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiSourceQualityType.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <summary>
+    /// Source quality options, used by the AudioLoFiEffect, that match common telephony and
+    /// radio broadcast options.
+    /// </summary>
+    public enum AudioLoFiSourceQualityType
+    {
+        /// <summary>
+        /// Narrow frequency range telephony.
+        /// </summary>
+        NarrowBandTelephony,
```
StephenHodgson: Initialize enum

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <remarks>
+        /// Chaining occluders will result in the lowest of the cutoff frequencies being applied to the sound.
+        /// The CutoffFrequency range is 0.0 - 22000.0 (0 - 22kHz), inclusive.
+        /// The default value is 5000.0 (5kHz).
+        /// </remarks>
+        [Tooltip("Frequency above which sound will not be heard after applying occlusion.")]
+        [Range(10.0f, 22000.0f)]
+        [SerializeField]
+        private float cutoffFrequency = 5000.0f;
+        public float CutoffFrequency
```
StephenHodgson: tooltip text in summary

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// VolumePassThrough is cumulative. It is applied to the current volume of the object at the time
+        /// the effect is applied.
+        /// The VolumePassThrough range is from 0.0 - 1.0 (0-100%), inclusive.
+        /// The default value is 1.0.
+        /// </remarks>
+        [Tooltip("Percentage of the audio source volume that will be heard after applying occlusion.")]
+        [Range(0.0f, 1.0f)]
+        [SerializeField]
+        private float volumePassThrough = 1.0f;
+        public float VolumePassThrough
```
StephenHodgson: We can move the summary down above the public property accessor

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+    {
+        /// <summary>
+        /// Frequency above which sound will not be heard after applying occlusion.
+        /// Setting this value to 22000.0 effectively disables the effect.
+        /// </summary>
+        /// <remarks>
+        /// Chaining occluders will result in the lowest of the cutoff frequencies being applied to the sound.
+        /// The CutoffFrequency range is 0.0 - 22000.0 (0 - 22kHz), inclusive.
+        /// The default value is 5000.0 (5kHz).
+        /// </remarks>
```
StephenHodgson: Oh I see this needs to get moved down to be above the public property.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,208 @@
...
+            sourceQualityFilterSettings.Add(
+                AudioLoFiSourceQualityType.FmRadio,
+                new AudioLoFiFilterSettings(30, 15000));    // Frequency range: 30 Hz - 15 kHz
+        }
+
+        /// <summary>
+        /// Settings for the filters used to simulate a low fidelity sound source.
+        /// </summary>
+        /// <remarks>
+        /// This struct is soley for the private use of the AudioLoFiEffect class.
```
StephenHodgson: soley ? Typo?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+        [Range(10.0f, 22000.0f)]
+        [SerializeField]
+        private float cutoffFrequency = 5000.0f;
+        public float CutoffFrequency
+        {
+            get { return cutoffFrequency; }
+            set
+            {
+                // set cutoffFrequency and enforce the specified range
+                if (value < 10.0f)
```
StephenHodgson: I think we can write this a bit cleaner:
`cutoffFrequency = Mathf.Clamp(value, 10.0f, 22000.0f);`

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/IAudioInfluencer.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Audio.Influencers
```
StephenHodgson: Shouldn't this be in core interfaces?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiSourceQualityType.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Audio.Influencers
```
StephenHodgson: We should probably move this into core as well.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+            {
+                influencer.RemoveEffect(gameObject);
+            }
+        }
+
+        /// <summary>
+        /// Finds the IAudioInfluencer objects that are to be applied to the audio source.
+        /// </summary>
+        /// <returns>Collection of IAudioInfluencers between the user and the game object.</returns>
+        private List<IAudioInfluencer> GetInfluencers()
```
StephenHodgson: This could also be a property.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+            audioSource = gameObject.GetComponent<AudioSource>();
+        }
+
+        /// <summary>
+        /// Removes the effects applied by specified audio influencers.
+        /// </summary>
+        /// <param name="influencers">Collection of IAudioInfluencer objects on which to remove the effect.</param>
+        private void RemoveInfluencers(List<IAudioInfluencer> influencers)
+        {
+            foreach (IAudioInfluencer influencer in influencers)
```
StephenHodgson: for loop?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+                List<IAudioInfluencer> influencers = GetInfluencers();
+                foreach (IAudioInfluencer influencer in influencers)
+                {
+                    // Apply the influencer's effect.
+                    influencer.ApplyEffect(gameObject);
+                }
+
+                // Find and remove the audio influencers that are to be removed from the audio source.
+                List<IAudioInfluencer> influencersToRemove = new List<IAudioInfluencer>();
+                foreach (IAudioInfluencer prev in previousInfluencers)
```
StephenHodgson: for loop?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+            DateTime now = DateTime.Now;
+
+            // Audio influences are not updated every frame.
+            if ((UpdateInterval * 1000.0f) <= (now - lastUpdate).Milliseconds)
+            {
+                audioSource.volume = initialAudioSourceVolume;
+
+                // Get the audio influencers that should apply to the audio source.
+                List<IAudioInfluencer> influencers = GetInfluencers();
+                foreach (IAudioInfluencer influencer in influencers)
```
StephenHodgson: for loop?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+        private RaycastHit[] hits;
+
+        /// <summary>
+        /// The collection of previously applied audio influencers.
+        /// </summary>
+        private List<IAudioInfluencer> previousInfluencers = new List<IAudioInfluencer>();
+
+        /// <summary>
+        /// Potential effects manipulated by an audio influencer.
+        /// <summary>
```
StephenHodgson: missing a /

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+    public class AudioInfluencerController : MonoBehaviour
+    {
+        /// <summary>
+        /// Frequency below the nominal range of human hearing.
+        /// </summary>
+        /// <remarks>
+        /// This frequency can be used to set a high pass filter to allow all
+        /// audible frequencies through the filter.
+        /// </remarks>
+        public static readonly float NeutralLowFrequency = 10.0f;
```
StephenHodgson: const?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        /// <summary>
+        /// Frequency above the nominal range of human hearing.
+        /// </summary>
+        /// <remarks>
+        /// This frequency can be used to set a low pass filter to allow all
+        /// audible frequencies through the filter.
+        /// </remarks>
+        public static readonly float NeutralHighFrequency = 22000.0f;
```
StephenHodgson: const?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// <summary>
+    /// Class which supports components implementing <see cref="IAudioInfluencer"/> being used with audio sources.
+    /// </summary>
+    /// <remarks>
+    /// AudioInfluencerController requires an <see cref="AudioSource"/> component. If one is not attached, it will be added automatically.
+    /// Each sound playing game object needs to have an AudioInfluencerController attached in order to have it's audio influenced.
+    /// </remarks>
+    [RequireComponent(typeof(AudioSource))]
+    [DisallowMultipleComponent]
+    public class AudioInfluencerController : MonoBehaviour
```
StephenHodgson: For some reason I'm getting a strong vibe that we should be making an interface for this and putting it in core so people can implement their own kinds of influencers.

Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/LoFiFilterSelection.cs
```diff
@@ -0,0 +1,122 @@
...
+        [SerializeField]
+        private Material FullRange = null;
+
+        [Tooltip("Material used when the emitter is set to an unknown quality")]
+        [SerializeField]
+        private Material UnknownQuality = null;
+
+        private AudioLoFiEffect loFiEffect = null;
+        // Component.renderer has been deprecated. It is safe to hide it and reuse the name.
+        private new Renderer renderer = null;
```
StephenHodgson: let's rename it anyway so we don't get build warnings

StephenHodgson: (I'm also not keep on using pragma to hide it)

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// This frequency can be used to set a low pass filter to allow all
+        /// audible frequencies through the filter.
+        /// </remarks>
+        public static readonly float NeutralHighFrequency = 22000.0f;
+
+        /// <summary>
+        /// Time, in seconds, between audio influence updates.
+        /// </summary>
+        /// <remarks>
+        /// The UpdateInterval range is betweel 0.0 and 1.0, inclusive.
```
StephenHodgson: between

Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/LoFiFilterSelection.cs
StephenHodgson: image

StephenHodgson:
> Set MSHRTF as the default audio spatializer (supports Windows 10 platforms).

Where is this change?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            AudioSource audioSource = soundEmittingObject.GetComponent<AudioSource>();
+            if (audioSource == null)
+            {
+                Debug.LogWarning("The specified emitter does not have an attached AudioSource component.");
+                return;
+            }
+
+            // Audio occlusion is performed using a low pass filter.                
+            AudioLowPassFilter lowPass = soundEmittingObject.GetComponent<AudioLowPassFilter>();
```
StephenHodgson: ` var lowPass = soundEmittingObject.EnsureComponent<AudioLowPassFilter>();`

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+            }
+        }
+
+        // Update is not used, but is kept so that this component can be enabled/disabled.
+        private void Update() { }
+
+        /// <summary>
+        /// Applies the audio effect.
+        /// </summary>
+        /// <param name="soundEmittingObject">The GameObject on which the effect is to be applied.</param>
```
StephenHodgson: We could move the docs to the interface and then inherit them here.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
```diff
@@ -0,0 +1,145 @@
...
+            lowPass.cutoffFrequency = Mathf.Min(lowPass.cutoffFrequency, CutoffFrequency);
+
+            // Unlike the cutoff frequency, volume pass-through is cumulative.
+            audioSource.volume *= VolumePassThrough;
+        }
+
+        /// <summary>
+        /// Removes the previously applied audio effect.
+        /// </summary>
+        /// <param name="soundEmittingObject">The GameObject from which the effect is to be removed.</param>
```
StephenHodgson: Same thing about inheriting the docs from the interface

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
StephenHodgson: I'm not keen on this as `GetComponent` is rather expensive and we're calling this a lot in an Update loop. Perhaps we should change this into an event based system with an `IAudioInfluencerHandler`

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
keveleigh: As I understand it, this is an optional component, not a required one.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
davidkline-ms: This happens on human scales of time, this is NOT every frame. The occluder cannot cache the emitter as it may change when the user moves around the scene,

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/IAudioInfluencer.cs
davidkline-ms: no

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiSourceQualityType.cs
davidkline-ms: no. this is SDK only. it is 100% addon

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: we have discussed this before. we agreed on this for HTK. why change it now?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: they are SDK constructs folks already CAN implement their own influencers. and the controller has to be a monobehavior

Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/LoFiFilterSelection.cs
davidkline-ms: what do you suggest i call it other than renderer

davidkline-ms:
> Where is this change?

forgot to check it in

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
StephenHodgson: If it's optional why do we attempt to get it then add one if it doesn't exist?

Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/LoFiFilterSelection.cs
StephenHodgson: `loFiSelectionRenderer` or similar I suppose. Other suggestions welcome.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: somehow missed deleting htese

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,313 @@
...
+        [Range(1.0f, 50.0f)]
+        [SerializeField]
+        private float maxDistance = 20.0f;
+        public float MaxDistance
+        {
+            get { return maxDistance; }
+            set
+            {
+                // set maxDistance and enforce the specified range
+                if (value < 1.0f)
```
StephenHodgson: we could do the same here with clamp function.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
StephenHodgson: > The occluder cannot cache the emitter as it may change when the user moves around the scene,

Sounds like we should be using our event system to handle these kinds of things.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/IAudioInfluencer.cs
StephenHodgson: Why not? Seems like something that should be. What if a developer whats to make their own influencers?

What's the advantage of using the interface if we're not going to let people use their own concrete implementations?

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
StephenHodgson: Link?

I'm not sure that this is something we'll ever change, so I think const is okay here.

Again I'll leave it up to your discretion.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
davidkline-ms: yes, get only property.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
StephenHodgson: Let's move it into core and use the event system. We should stick to our patterns we've set for the MRTK and not worry about how we've done things in HTK.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
StephenHodgson: No worries! Just wanted to double check :)

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: technically, but it feels more like a method based on the work it is doing

StephenHodgson: It is def a property `get` in the name says it all.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: this is not intended to be a core system it never has been.

davidkline-ms: the event system does not fit this scenario.

StephenHodgson: I disagree, as we're using an interface and using the event data would help keep us from having to get components (which is rather costly)

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioOccluder.cs
davidkline-ms: the low pass filter is not on the occluder, it is on the object containing the AudioInfluencerController. If it does not have this object, we add it

            AudioLowPassFilter lowPass = soundEmittingObject.GetComponent<AudioLowPassFilter>();
            if (lowPass == null)
            { 
                lowPass = soundEmittingObject.AddComponent<AudioLowPassFilter>();
            }

StephenHodgson: Correct, so then we can use the proposed `EnsureComponent`

StephenHodgson: Under the hood the method does exactly what's written:

```
        public static T EnsureComponent<T>(this GameObject gameObject) where T : Component
        {
            T foundComponent = gameObject.GetComponent<T>();
            if (foundComponent == null)
            {
                return gameObject.AddComponent<T>();
            }

            return foundComponent;
        }
```

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: should unity have implemented GetComponents as a property?  there are plenty of APIs that both have properties and Get* methods.

my bar is that properties should not do very much work (set a value, get a value, validate parameters). this is due to properties looking and feeling like a field... there should be minimal if any side effects

davidkline-ms: this is how the .net framework was written and is a reasonably common engineering pattern

StephenHodgson: > my bar is that properties should not do very much work (set a value, get a value, validate parameters). this is due to properties looking and feeling like a field... there should be minimal if any side effects

That seems to check all the boxes in this case. I'll leave it up to you.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/IAudioInfluencer.cs
davidkline-ms: they CAN add their own concrete implementatioons…. in the SDK like AudioOccluder is a concrete implementation

this is one of our "use it if you like" components

StephenHodgson: Then what's the apprehension to adding it to core?

davidkline-ms: because it has never been a core feature. it is not necessary to 99% of apps

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
davidkline-ms: costly when done tons of times every frame. at a human scale (4 times per second) it is not that expensive

StephenHodgson: I suppose we'll have to see when the GC kicks in and we get stalls on HL or another mobile device. 

In either case, we should at least take a look at what this might look like instead of dismissing it bc it's different than the way it was originally written. Remember the MRTK is not the HTK.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/IAudioInfluencer.cs
StephenHodgson: I would argue that audio is a pretty core feature.

keveleigh: This is advanced audio. Most apps will be fine with Unity's built in audio support and not need to add these features on top of it.

StephenHodgson: Indeed, but in that case, shouldn't we be registering it as a `IMixedRealityComponent` so we know we're adding it to the correct platforms?

I don't see us checking SDK's or plugins in this PR, but we do set it, but what if we're not targeting WMR platform?

keveleigh: Which scripts in this PR depend on a specific platform or plugin?

StephenHodgson: Good question.

keveleigh: The spatializer is being set in the Unity project for the example scene. This PR does not handle that, but is instead a port of audio influencers and effects.

For setting the spatializer, see #2592.

StephenHodgson: Okay, well what about devs who import the MRTK into their project? Will they have to manually set it as well? Do we have a way to warn them if it's unset? or if their platform is unsupported?

StephenHodgson: How do we handle other spatializers for other platforms? would this demo still apply?

keveleigh: Again, for most of that, see #2592.

This PR uses stock Unity components. If the resulting audio ends up spatialized or not, it doesn't really matter. These simply add effects on top of the audio.

StephenHodgson: Because they might not have the SDK. Remember, the SDK is only an interpretation of the interfaces, which are all part of the framework.

keveleigh: If they don't have the SDK, there's nothing to talk to the interfaces. Why do I need the interfaces if the script that consumes them doesn't exist?

I'm aware of what the SDK is.

StephenHodgson: I don't think you guys understand the point I'm trying to make. 

Either we use an interface and we move it to core, or we keep a concrete type and remove the interface.

keveleigh: I understand that that's what you're saying.

I'm saying that that's not a requirement. It's fine as-is.

StephenHodgson: It's a requirement per the architectural agreement we laid out before we started.

keveleigh: Can you point me to that requirement? The interface layer mentions devices, interactions, and input, none of which this seems to fall under.

keveleigh: It is not a catch-all layer, nor should it be, as the size of the core should be kept as minimal as possible.

StephenHodgson: https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/MRTK-vNext.md#objective

StephenHodgson: > It is not a catch-all layer, nor should it be, as the size of the core should be kept as minimal as possible.

Agreed, which is why it's only and interface defining the structure, not the concrete implementation.

StephenHodgson: Essentially this class belongs in the "interface layer"

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -0,0 +1,297 @@
...
+            // Preallocate the array that will be used to collect RaycastHit structures.
+            hits = new RaycastHit[MaxObjects];
+        }
+    
+        private void Update() 
+        {
+            DateTime now = DateTime.Now;
+
+            // Audio influences are not updated every frame.
+            if ((UpdateInterval * 1000.0f) <= (now - lastUpdate).Milliseconds)
```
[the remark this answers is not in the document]
davidkline-ms: update interval is a fraction of a second. the bug you point out is a corner case when the user specifies 1 for the interval. i'll switch to totalms.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
```diff
@@ -0,0 +1,198 @@
...
+                return !(a.Equals(b));
+            }
+
+            /// <summary>
+            /// Checks to see if a object is equivalent to this FilterSettings.
+            /// </summary>
+            /// <returns>True if equivalent, false otherwise.</returns>
+            public override bool Equals(object obj)
+            {
+                if (obj == null)
```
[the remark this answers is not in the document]
davidkline-ms: I should have thought of that

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/IAudioInfluencer.cs
davidkline-ms: fine. i'll move the interface. i still do not agree that EVERY interface belongs in the core, but i dont want to argue anymore about an SDK component port.

davidkline-ms: there are likely to be more PRs where there are sdk layer specific interfaces...

StephenHodgson: I just want to make it clear that we're not just dragging and dropping code from the htk.

We need to be thoughtful of the way we've structured things, and make sure we're sticking to what we agreed on with the initial proposed architecture.

Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs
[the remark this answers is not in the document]
davidkline-ms: non-nullable value types require is and an explicit cast

PR #2779, opened 2018-09-13
vNEXT: Updated Coding Guidelines

StephenHodgson:
Overview
---
- Added expression-bodied member and auto property guidelines.
- Fixed a few spelling errors.
- Added tooltip guidelines
- Added color to the code with c# syntax

### DIFF
--- a/CodingGuidelines.md
+++ b/CodingGuidelines.md
@@ -473,8 +473,13 @@ public float MyValue;
 
  ```
  // Enable field to be configurable in the editor and available externally to other scripts (field is correctly serialized in Unity)
- [SerializeField] 
- private float myValue; // <- Notice we co-located the backing field above our corrisponding property.
+ [SerializeField]
+ [ToolTip("If using a tooltip, the text should match the public property's summary documentation, if appropriate.")]
+ private float myValue; // <- Notice we co-located the backing field above our corresponding property.
+
+ /// <summary>
+ /// If using a tooltip, the text should match the public property's summary documentation, if appropriate.
+ /// </summary>
  public float MyValue
  {
      get{ return myValue; }
--- a/CodingGuidelines.md
+++ b/CodingGuidelines.md
@@ -8,7 +8,7 @@ This document outlines the recommended coding guidelines for the Mixed Reality T
 
 All scripts posted to the MRTK should have the standard License header attached, exactly as shown below:
 
-```
+```c#
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 ```
@@ -19,7 +19,7 @@ Any script files submitted without the license header will be rejected
 
 All public classes, structs, enums, functions, properties, fields posted to the MRTK should be described as to it's purpose and use, exactly as shown below:
 
-```
+```c#
     /// <summary>
     /// The Controller definition defines the Controller as defined by the SDK / Unity.
     /// </summary>
@@ -43,7 +43,8 @@ The vNext structure adheres to a strict namespace culture of mapping the namespa
 ![](/External/ReadMeImages/MRTK-NameSpaceExample.png)
 
 ### Do:
-```
+
+```c#
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
@@ -68,7 +69,7 @@ Additionally, ensure that spaces are added for conditional / loop functions like
 
 ### Don't:
 
-```
+```c#
 private Foo () // < - space between Foo and ()
 {
     if(Bar==null) // <- no space between if and ()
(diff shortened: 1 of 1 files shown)


SimonDarksideJ:
Given this [issue](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2780)
Can we also add something about events to the coding guidelines. So we are not updating this doc multiple times.

StephenHodgson:
Sure, I'll update it in a few hours.

CodingGuidelines.md
```diff
@@ -100,21 +101,46 @@ private Foo()
 }
  ```
 
+## Spacing
+
+Please be sure not to add additional spaces between square brackets and parenthesis:
```
keveleigh: I'm not sure I intended my PR review to mean that we should add this to the guidelines.

I've never created coding standards for a shared repo before, but this seems like a nitpicky thing to enforce. Perhaps it's necessary though, and I'd appreciate others weighing in.

StephenHodgson: Well it's part of MSDN as as we clearly say follow MSDN when it's not explicitly called out, I decided add it to make it clear.

StephenHodgson: We could also enforce this by adding a vs configuration file, but I feel that's a bit extreme for editor default settings.

keveleigh: Do you have a link to the MSDN guideline related to this?

StephenHodgson: [Sure](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#layout-conventions)

> Use the default Code Editor settings

StephenHodgson: Now let's get back to reviewing some important PRs like gestures. We really need to get that in sooner rather than later.

keveleigh: "Use the default Code Editor settings" doesn't say anything explicitly about this rule, but okay.

>Now let's get back to reviewing some important PRs like gestures

Larger PRs require more targeted time to review.

Additionally, I have an outstanding change request from before you most recently "bump"ed that PR...

SimonDarksideJ: Please reword the header to 
"Do not add additional spaces between square brackets and parenthesis:"
Old statement was too passive

CodingGuidelines.md
```diff
@@ -38,16 +38,17 @@ This ensures documentation is properly generated and disseminated for all all cl
...
 
 ### Do:
-```
+
+```c#
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-namespace Microsoft.MixedReality.Toolkit.Internal.Definitons
+namespace Microsoft.MixedReality.Toolkit.Internal.Definitions
```
davidkline-ms: Internal -> Core

CodingGuidelines.md
```diff
@@ -126,17 +152,29 @@ Always declare an access modifier for all fields, properties and methods.
 
 >All Unity API Methods should be `private` by default, unless you need to override them in a derived class. In this case `protected` should be used.
 
+>Fields should always be `private`, with `public` or `protected` property accessors.
```
davidkline-ms: Does this imply that we will not allow readonly fields (essentially only allow get-only properties)? Feels a bit heavy handed.

CodingGuidelines.md
```diff
@@ -212,43 +250,43 @@ public class MyClass
...
     private MyStruct myStructreference;
     private MyEnumType myEnumReference;
 }
  ```
 
-## Initilize Enums.
+## Initialize Enums.
 
-To ensure all Enum's are initialized correctly starting at 0, .NET gives you a tidy shortcut to automatically initilize the enum by just adding the first (starter) value.
+To ensure all Enum's are initialized correctly starting at 0, .NET gives you a tidy shortcut to automatically initialize the enum by just adding the first (starter) value.
```
davidkline-ms: Other than readability, what is the reason to require this? .net automatically initializes the first emum value to 0 unless otherwise specified.

CodingGuidelines.md
SimonDarksideJ: With the architecture of the framework, there should not be much need for readonly fields, especially as most things are made configurable or extendable.
But a lot of the guidance was formed based from issues found in the HTK and to avoid them in the future.

CodingGuidelines.md
SimonDarksideJ: because it's good practice, this was one of the first guidelines added when we started the document.  To ensure consitancy across the solution. (also recommended in MSDN I believe from the discussions we had back then.)

CodingGuidelines.md
SimonDarksideJ: Good catch :D

CodingGuidelines.md
StephenHodgson: Not sure read only fields apply here. I think those are fine to be public/protected as they're not being changed at runtime.

The principal goal here is ensuring proper encapsulation.

StephenHodgson:
@davidkline-ms anything else for this PR?

PR #2782, opened 2018-09-13
Pointers shouldn't raise input events

StephenHodgson:
Overview
---
Updated pointer events to only raised from the controller/device classes as originally intended.


StephenHodgson:
Bump. Not sure why [we're going over a week to review PRs](
https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/CONTRIBUTING.md#pull-request)

>We aim to review all changes within three business days.

StephenHodgson:
Awesome, I'll be sure to address that shortly :)

StephenHodgson:
Yeah I'm still flip flopping on if we should only have pointer raise pointer events or devices only raise pointer events.

@keveleigh  You bring up a good point about the pointer actions. Hmm.

@SimonDarksideJ How would you feel if instead we do it the other way around and only have pointers raise pointer events?

StephenHodgson:
Okay looking back, we don't raise pointer events in the device layer except for touch and mouse (which I think is okay. They have their own specialized pointers anyway)

keveleigh:
Sounds good to me!

Issue #2608, opened 2018-08-17
vNext: Refactor pose update events into passive transforms


Ecnassianer:
## Overview

Right now tracked objects fire pose events every time they move. I understand the expectation to be that every object that wants to know about the position of a controller, tracker, headset, etc implements those pose events and updates their transforms accordingly. An event based movement model makes a lot of sense for mice, touch screens, or other traditional devices, which spend the vast majority of their time static. However, because of the nature of tracking devices, they are firing pose updates every frame; even if you hold your head as perfectly still as possible, it's still going to be in a new position every time it polls. 

Each class that cares about tracked positions handles an increasing number of pose events every frame. A class might find itself updating a pointer pose for each controller, a grip pose for each, plus a headset pose. @keveleigh and I easily imagined a world where future controllers add additional tracked points. Analog sticks tilt independently of tracking. Finger tracking like Knuckles, Leap Motion, or ManuVR add an arbitrary number of additional position events (over 20 per hand in the case of Leap Motion!). A flip up WMR headset may offer the position of the display in addition to the position of the user's head. A unity class concerned with these things now implements and maintains a massive list of event handles, all firing every frame.

There is nothing explicitly wrong with this approach, but it is not very "Unityish". Unity convention presents constantly updating positions as Transforms which offer a suite of developer convenience like parenting and local offsets, in addition to a host of invisible optimizations such as network quantization.

### Adding position Transforms for all tracked objects

I'm proposing that we adopt a method of exposing tracked positions as transforms. A user's code can request a transform for the Left Motion Controller, parent itself with an offset, and trust that it will always follow the tracked device. Additionally that transform can be cached in Start() and read from at a later date, not having to worry about keeping it updated for some future calculation.

### Promote Transforms as the primary method of accessing position data

I believe this is the the most intuitive method for Unity users, and we should encourage the use of Transforms as the primary method of accessing tracked position data. I do not believe there is any case where an event based position update is superior for continuously tracked objects. This means updating example maps and documentation to use this method.

### Optional: Removing pose events

I think we would be best served by not offering pose events at all, since they only offer a less ideal choice and demand an additional maintenance burden. 

However, these are relatively low costs compared to offering the familiarity of event driven positions for developers who are used to developing on traditional position devices like mice. This also would have been a better change to make before Alpha, but in the bigger picture, we're still very early in the infancy of vNext.

Do you know of a time when pose events are a better choice than transforms?

[...]

Issue #2465, opened 2018-07-19
Migrate master branch fixes to mrtk_development


davidkline-ms:
List of known fixes that need to be migrated from master -> mrtk_development.

This is a work in progress, *DO NOT* close this issue until vNext release 1 is complete.

- [x] #2245
- [ ] #2306
- [ ] #2457 
- [ ] #2459 
- [ ] #2487

wiwei:
Closing issues older than 180 days. If this is still an issue, please reactivate with recent information.

PR #2784, opened 2018-09-13
[Solver PR 4] Add controller-tracking support to solvers

keveleigh:
Overview
---
Adds controller finder to the solvers. Now, left and right motion controllers join the head as valid Solver targets. In order to facilitate this, I had to add the controller transform to the base MRController definition in order to expose it.

This also ports over the custom inspectors for SolverHandler and ControllerFinder.

Changes
---
Related to #2465

Related to #2608 (adding passive controller transforms)

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
@@ -1,0 +1,164 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities;
+using Microsoft.MixedReality.Toolkit.Core.EventDatum.Input;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.Devices;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem.Handlers;
+using Microsoft.MixedReality.Toolkit.Core.Managers;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
+{
+    /// <summary>
+    /// ControllerFinder is a base class providing simple event handling for getting/releasing MotionController Transforms.
+    /// </summary>
+    public abstract class ControllerFinder : MonoBehaviour, IMixedRealitySourceStateHandler
+    {
+        //[SerializeField]
+        //private ControllerElement trackedControllerElement = ControllerElement.PointingPose;
+
+        //public ControllerElement TrackedControllerElement
+        //{
+        //    get { return trackedControllerElement; }
+        //    set { trackedControllerElement = value; }
+        //}
+
+        [SerializeField]
+        private Handedness handedness = Handedness.None;
+
+        public Handedness Handedness
+        {
+            get { return handedness; }
+            set
+            {
+                // We need to refresh which controller we're attached to if we switch handedness.
+                if (handedness != value)
+                {
+                    handedness = value;
+                    RefreshControllerTransform();
+                }
+            }
+        }
+
+        //public Transform ElementTransform { get; private set; }
+
+        protected Transform ControllerTransform;
+
+        private IMixedRealityDeviceManager BaseDeviceManager => baseDeviceManager ?? (baseDeviceManager = MixedRealityManager.Instance.GetManager<IMixedRealityDeviceManager>());
+        private IMixedRealityDeviceManager baseDeviceManager = null;
+
+        #region MonoBehaviour Implementation
+
+        protected virtual void OnEnable()
+        {
+            if (BaseDeviceManager == null)
+            {
+                // The base device manager has not been set up yet.
+                return;
+            }
+
+            // Look if the controller has loaded.
+            RefreshControllerTransform();
+        }
+
+        #endregion MonoBehaviour Implementation
+
+        #region IMixedRealitySourceStateHandler Implementation
+
+        public void OnSourceDetected(SourceStateEventData eventData)
+        {
+            if (eventData.Controller.ControllerHandedness == handedness)
+            {
+                AddControllerTransform(eventData.Controller);
+            }
+        }
+
+        public void OnSourceLost(SourceStateEventData eventData)
+        {
+            if (eventData.Controller.ControllerHandedness == handedness)
+            {
+                RemoveControllerTransform();
+            }
+        }
+
+        #endregion IMixedRealitySourceStateHandler Implementation
+
+        /// <summary>
+        /// Looks to see if the controller model already exists and registers it if so.
+        /// </summary>
+        protected virtual void TryAndAddControllerTransform()
+        {
+            // Look if the controller was already loaded. This could happen if the
+            // GameObject was instantiated at runtime and the model loaded event has already fired.
+            if (BaseDeviceManager == null)
+            {
+                // The BaseDeviceManager could not be found.
+                return;
+            }
+
+            IMixedRealityController[] controllers = BaseDeviceManager.GetActiveControllers();
+
+            for (int i = 0; i < controllers.Length; i++)
+            {
+                if (controllers[i].ControllerHandedness == handedness)
+                {
+                    AddControllerTransform(controllers[i]);
+                    return;
+                }
+            }
+        }
+
+        protected virtual void AddControllerTransform(IMixedRealityController newController)
+        {
+            if (newController.ControllerHandedness == handedness && !newController.Transform.Equals(ControllerTransform))
+            {
+                //Transform elementTransform;
+                //if (!newController.TryGetElement(trackedControllerElement, out elementTransform))
+                //{
+                //    Debug.LogError("Unable to find element of type " + trackedControllerElement + " under controller " + newController.ControllerParent.name + "; not attaching.");
+                //    return;
+                //}
+
+                ControllerTransform = newController.Transform;
+
+                //// Update ElementTransform for consumption
+                //ElementTransform = elementTransform;
+
+                OnControllerFound();
+            }
+        }
+
+        protected virtual void RemoveControllerTransform()
+        {
+            OnControllerLost();
+
+            ControllerTransform = null;
+            //ElementTransform = null;
+        }
+
+        protected virtual void RefreshControllerTransform()
+        {
+            if (ControllerTransform != null)
+            {
+                RemoveControllerTransform();
+            }
+
+            TryAndAddControllerTransform();
+        }
+
+        /// <summary>
+        /// Override this method to act when the correct controller is actually found.
+        /// This provides similar functionality to overriding AddControllerTransform,
+        /// without the overhead of needing to check that handedness matches.
+        /// </summary>
+        protected virtual void OnControllerFound() { }
+
+        /// <summary>
+        /// Override this method to act when the correct controller is actually lost.
+        /// This provides similar functionality to overriding AddControllerTransform,
+        /// without the overhead of needing to check that handedness matches.
+        /// </summary>
+        protected virtual void OnControllerLost() { }
+    }
+}
--- a/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
@@ -15,15 +15,6 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
     /// </summary>
     public abstract class ControllerFinder : MonoBehaviour, IMixedRealitySourceStateHandler
     {
-        //[SerializeField]
-        //private ControllerElement trackedControllerElement = ControllerElement.PointingPose;
-
-        //public ControllerElement TrackedControllerElement
-        //{
-        //    get { return trackedControllerElement; }
-        //    set { trackedControllerElement = value; }
-        //}
-
         [SerializeField]
         private Handedness handedness = Handedness.None;
 
@@ -40,8 +31,6 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
                 }
             }
         }
-
-        //public Transform ElementTransform { get; private set; }
 
         protected Transform ControllerTransform;
 
@@ -111,19 +100,9 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
 
         protected virtual void AddControllerTransform(IMixedRealityController newController)
         {
-            if (newController.ControllerHandedness == handedness && !newController.Transform.Equals(ControllerTransform))
+            if (newController.ControllerHandedness == handedness && newController.Transform != null && !newController.Transform.Equals(ControllerTransform))
             {
-                //Transform elementTransform;
-                //if (!newController.TryGetElement(trackedControllerElement, out elementTransform))
-                //{
-                //    Debug.LogError("Unable to find element of type " + trackedControllerElement + " under controller " + newController.ControllerParent.name + "; not attaching.");
-                //    return;
-                //}
-
                 ControllerTransform = newController.Transform;
-
-                //// Update ElementTransform for consumption
-                //ElementTransform = elementTransform;
 
                 OnControllerFound();
             }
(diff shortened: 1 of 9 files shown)


SimonDarksideJ:
From an architectural point of View, i object to this PR.
We refactored out the "ControllerFinder" when we put in the new visualisation system (@StephenHodgson can comment more on the work he did there)

Although I agree, we need to link the visualizer with the solver. However, this is not the way to do it.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
```diff
@@ -0,0 +1,168 @@
...
+using Microsoft.MixedReality.Toolkit.Core.Managers;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
+{
+    /// <summary>
+    /// ControllerFinder is a base class providing simple event handling for getting/releasing MotionController Transforms.
+    /// </summary>
+    public abstract class ControllerFinder : MonoBehaviour, IMixedRealitySourceStateHandler
```
StephenHodgson: Isn't this the same as the ControllerPoseSynchonizer?

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
```diff
@@ -0,0 +1,168 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// The Transform of the currently found controller.
+        /// </summary>
+        protected Transform ControllerTransform;
+
+        private IEnumerable<IMixedRealityManager> BaseDeviceManagers => baseDeviceManagers ?? (baseDeviceManagers = MixedRealityManager.Instance.GetManagers(typeof(IMixedRealityDeviceManager)));
```
StephenHodgson: Not sure this is a good idea.

Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityController.cs
```diff
@@ -59,5 +60,10 @@ public interface IMixedRealityController
         /// Mapping definition for this controller, linking the Physical inputs to logical Input System Actions
         /// </summary>
         MixedRealityInteractionMapping[] Interactions { get; }
+
+        /// <summary>
+        /// The transform of this controller, used for tracking and observing the controller if it's tracked.
+        /// </summary>
+        Transform Transform { get; }
```
StephenHodgson: The device layer shouldn't know about anything in the scene.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
SimonDarksideJ: That was it, I recall rejecting bring it in and conceded slightly when we changed the implementation in line with the MRTK approach.

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
```diff
@@ -97,6 +97,9 @@ protected IMixedRealityInputSystem InputSystem
         /// <inheritdoc />
         public MixedRealityInteractionMapping[] Interactions { get; private set; } = null;
 
+        /// <inheritdoc />
+        public Transform Transform { get; protected set; }
```
StephenHodgson: Same comment here.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -11,8 +11,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
     /// <summary>
     /// This class handles the solver components that are attached to this <see cref="GameObject"/>
     /// </summary>
-    [DisallowMultipleComponent]
-    public class SolverHandler : MonoBehaviour
+    public class SolverHandler : ControllerFinder
```
StephenHodgson: If anything we should be inheriting form the pose synchronizer.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
SimonDarksideJ: Yikes, missed that.  I cannot think of any good reason why any component outside of the device layer would need to enumerate the managers, especially the devices.
The abstraction is there for a reason to ensure maximum compatibility.

StephenHodgson: Thanks!

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
keveleigh: It's not, no. The ControllerPoseSynchonizer moves an object to a specific pose based on the events. This exposes a transform that can be used to follow or do math on or whatever.

This additionally diverges from ControllerPoseSynchonizer when specific controller attachment points come online.

SimonDarksideJ:
Where a fair amount of code can be copied from HTK (like the solver logic, etc)
The runtime implementation of a lot of features has radically changed.

Especially when it interacts with the 4 layer architecture which is the foundation of the new MRTK.  One key principle is that ONLy the device layer talks to devices and it purely (with the exception of Haptics) and output factory.  Everything else only reacts to what come out of the device layer, it should not try and interact  with it.
Also we need to ensure we avoid some of the old bad practices where functionality tried to interact with the platform direct (thinking of the old two handed implementation which used the interaction manager and Unity Input).

I'll need to sleep on this and what it's trying to achieve to see if there is a better way forward with the requirement (although I've not seen a task listing the requirements)

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
keveleigh: As I mentioned in the PR where you created that pose synchronizer, it does not fulfill the requirements of the original ControllerFinder. It serves a similar purpose, but is incompatible with this case.

Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityController.cs
StephenHodgson: Sure, I'll open a PR

davidkline-ms:
> (although I've not seen a task listing the requirements)

There haven't been tasks for quite a few PRs that have been considered *critical*. If we are going to block one PR because there isn't a task, we should block ALL non-bugfix PRs if there isn't a task. Consistency is going to be key to expanding the set of contributors.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
StephenHodgson: Easy, use the pose synchronizer

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
StephenHodgson: Those requirements being?

StephenHodgson:
Working in a PR that satisfies your requirements in your task without breaking the architecture. Keep an eye out for a PR soon.

Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityController.cs
keveleigh: >The device layer shouldn't know about anything in the scene.

Can you point me to the requirement in the approved architecture that states this?

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
keveleigh: >The abstraction is there for a reason to ensure maximum compatibility.

Exactly, that's why I'm only asking for and calling methods contained in the base device interface abstraction. Complete compatibility regardless of the platform I'm on. Other SDK components talk to systems using `GetManager` as well.

What's a scenario where this breaks or does something "wrong"?

StephenHodgson: Exactly the scenario you're proposing.

keveleigh: Nothing breaks and nothing wrong happens. Please be more specific instead of restating what you've already said.

StephenHodgson: I'm not sure how I can get more specific as we've clearing pointed out everything wrong in other sections of the comments and issues related to this. I'm going to politely ask you not to work on this anymore, as I'm not entirely confident you understand the implications of the changes you're requesting to make.

keveleigh: I'm making one change to the device layer, and the rest are SDK changes to solvers only. Feel free to enumerate the implications in this thread.

StephenHodgson:
The fact that we're also needing to change the device layer to get solvers to work, confirms my suspicion that they should have stayed in core, where we had originally placed them.

keveleigh:
If solvers stayed in core, we'd still have needed to add a way to access the transform. The layer they live in doesn't change that, and solvers do not make sense in the core. That discussion was already resolved in a separate PR.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -176,20 +174,57 @@ protected void OnDestroy()
...
             }
         }
 
         private void TrackTransform(Transform newTrackedTransform)
         {
-            TransformTarget = RequiresOffset ? MakeOffsetTransform(newTrackedTransform) : newTrackedTransform;
+            transformTarget = RequiresOffset ? MakeOffsetTransform(newTrackedTransform) : newTrackedTransform;
+        }
+
+        private void UpdateOffsetTransform()
```
StephenHodgson: Why add this if we're not using it?

StephenHodgson:
Please cherry pick  your changes not related to the device layer and controller finder.

StephenHodgson:
> That discussion was already resolved in a separate PR.

I'd argue that it wasn't but was done anyway with knowledge and warning that it wasn't a good idea.

keveleigh:
Can you explain, in detail, how solvers being in the core would change this PR?

SimonDarksideJ:
Ok, to simplify things.  Give me the weekend to propose an alternative solution that should work better with the new architecture in MRTK.

Breaking things aside, in the Architecture diagram that is used for the basis of most of the discussions.

* The Devices job is to transport and expose current information about the controller and it's state.  It has no presence within a project, it is simply a data pump.
* The DeviceManager coordinates data from the device and constructs the "inputSource" data that is attached to all events sent out from the device.  
* The inputsource includes things like pointers, data values and internal representations of the information about the controller (e.g. Handedness, Index, capabilities)
* The Input System (when used) consumes events thrown by the device and outputs / routes actions and data for use within an internal project

Each component in the architecture flows down, ensuring there is no impact or dependency in the project on any previous layer and each can be easily swapped out.

The current ControllerPoseSynchroniser for instance. Registers and listens for data from the Input system and then with the visualiser, presents the controllers on screen.  It does interact with the controller configuration, but only to read out what has been defined internally to describe the device.

If (assuming I'm reading the task requirements right) the Solver needs "something" to attach to, it should be coming from something generated within the project, the "Thing" or "position" or GameObject that is being presented in the scene that you want to attach to.

This way it doesn't matter which device / input or "thing" is putting the visualisation in the scene and is not another requirement to be factored in to every device, as it is just a representation in the scene.

I hope that helps clear up a little more of the thinking.  But let me come up with an alternate proposal for now.

keveleigh:
The "something" is a Transform. The controller definition already creates a GameObject/Transform when it spawns the controller model. The only change is to actually expose that model/Transform.

I'm calling methods defined in the base interfaces. Not all device managers need to (or do) implement everything. If no controller representation exists, then no Transform exists, then this falls back gracefully.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
davidkline-ms: > I'm not sure how I can get more specific as we've clearing pointed out everything wrong in other sections of the comments and issues related to this. I'm going to politely ask you not to work on this anymore, as I'm not entirely confident you understand the implications of the changes you're requesting to make.

There have been quite a few "this is wrong and vague comments on this PR relating to lack of confidence in various contributor's understandings of the architecture and the intent of these changes, as well as requests for elaboration on architecture.

I do agree that we need clarity (tasks?) for feature additions and I would like to understand why this "is not compatible with the way the MRTK is architected,". Documentation beyond the layer diagrams would go a long way towards helping here.

davidkline-ms: Let's try to come to a common understanding on what the goal is and why things need to be done in certain ways (or not).

My goal is to avoid heated discussions over async text as they can rapidly devolve.

StephenHodgson:
All I'm simply asking is that you attach the solver handler to the `IMixedRealityControllerPoseSynchronizer` (usually in the pointer prefab) and using that transform the synchronizer is attached to. I've got a PR I'm working on to demonstrate the changes.

davidkline-ms:
>The Devices job is to transport and expose current information about the controller and it's state. It has no presence within a project, it is simply a data pump.
The DeviceManager coordinates data from the device and constructs the "inputSource" data that is attached to all events sent out from the device.
The inputsource includes things like pointers, data values and internal representations of the information about the controller (e.g. Handedness, Index, capabilities)
The Input System (when used) consumes events thrown by the device and outputs / routes actions and data for use within an internal project

This is *precisely* the type of information that is good to have written down (and reviewed) in our architecture documentation!

Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityController.cs
StephenHodgson: see https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2784#issuecomment-421424132

SimonDarksideJ:
I'll get to work on that writeup @davidkline-ms , Good abstraction is hard to build and far too easy to break.  The devices especially need to remain isolated to ensure it can handle all the demands of the various hardware we need to support in the future.
Objects from the scene should only receive data from the device layer or discover properties / registration / configuration from the MR manager through the profiles.

Granted the current visualization is not ideal, but it was a halfway house to meet the Alpha

SimonDarksideJ:
Closing until #2837 is merged which will change the implementation anyway and require a new PR

SimonDarksideJ:
It would be cleaner @keveleigh to do a fresh PR on the current dev branch, else we risk dirtying the change history with the back and to.  Especially as it's dependant on the Visualization PR, which still isn't merged.
Can we focus on that and then get a fresh PR to update just the solver handler PR.

keveleigh:
I've been rebasing all my changes/PRs to create a clean, descriptive history (this guideline used to be in our contributing doc, but looks to have been removed), and I'll do the same with this one.

keveleigh:
Rebased on top of the current mrtk_development, with the new way of accessing visualizers.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/ControllerFinder.cs
```diff
@@ -0,0 +1,151 @@
...
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem.Handlers;
+using Microsoft.MixedReality.Toolkit.Core.Managers;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
+{
+    /// <summary>
+    /// ControllerFinder is a base class providing simple event handling for getting/releasing MotionController Transforms.
+    /// </summary>
+    public abstract class ControllerFinder : MonoBehaviour, IMixedRealitySourceStateHandler
```
StephenHodgson: Just to clarify this class is meant to be used on GameObjects that are looking for controllers, not the controller itself, correct?

keveleigh: Correct

PR #2788, opened 2018-09-14
Cherry picked Solver changes from Part 4 PR

StephenHodgson:
Overview
---
These are the changes I do agree with from #2784

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
@@ -29,8 +29,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
                 if (trackedObjectToReference != value)
                 {
                     trackedObjectToReference = value;
-                    TransformTarget = null;
-                    AttachToNewTrackedObject();
+                    RefreshTrackedObject();
                 }
             }
         }
@@ -48,7 +47,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
             set
             {
                 additionalOffset = value;
-                TransformTarget = MakeOffsetTransform(TransformTarget);
+                transformTarget = MakeOffsetTransform(transformTarget);
             }
         }
 
@@ -65,7 +64,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
             set
             {
                 additionalRotation = value;
-                TransformTarget = MakeOffsetTransform(TransformTarget);
+                transformTarget = MakeOffsetTransform(transformTarget);
             }
         }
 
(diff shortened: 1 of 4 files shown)


keveleigh:
These changes are fine to remain in the other PR. Cherry-picking changes you agree with out of someone else's PR is not how this process works. Feel free to discuss on #2784.

StephenHodgson:
> Cherry-picking changes you agree with out of someone else's PR is not how this process works.

Sorry I was just trying to help get some of those changes out of the way so we can focus on the main ones.

StephenHodgson:
> These changes are fine to remain in the other PR.

I disagree, as they can be broken out like I've done here, or else I wouldn't have proposed the PR.

davidkline-ms:
I am generally in favor of targeted, smaller PRs (from all of us). One thing I would like to understand is what the benefit of this change in isolation provides outside of the larger PR (that has already been broken down from an even larger one).

Thx

davidkline-ms:
I'd also like to understand the overwhelming objection to the larger PR.

SimonDarksideJ:
I hoped I had detailed the root cause of the architectural issues with the PR in the following comment
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2784#issuecomment-421424132

Ultimately, a Controller does not register itself with a scene, neither does a device.  Only the Visualiser creates an entity / object in the scene which represents the active controller. The two are not directly linked except by the data that the device outputs.

keveleigh:
>Ultimately, a Controller does not register itself with a scene, neither does a device. Only the Visualiser creates an entity / object in the scene which represents the active controller. 

This is not true. See the `BaseController` for more info:

https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs#L206

StephenHodgson:
>Only the Visualiser creates an entity / object in the scene which represents the active controller.

>This is not true. See the BaseController for more info:

That's exactly what's going on here, how is that not true?  That's exactly the transform you should be using.

SimonDarksideJ:
You will notice, that that function is not called anywhere in the controller classes. It is consumed by the pose synchroniser.  It isn't the thing that does it, just for hosting the code.

Ultimately that is only there temporarily for now until we have the full visualizer framework, which would do the rendering.

keveleigh:
>That's exactly what's going on here, how is that not true? 

I don't see any visualizer creating the model object. I see the controller creating the model object, as I linked above. It's literally code inside the controller, not a visualizer.

>That's exactly the transform you should be using

That's exactly the transform I am using.

keveleigh:
>You will notice, that that function is not called anywhere in the controller classes. 

It is called in the BaseController's SetupConfiguration:

https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs#L112

StephenHodgson:
Let's leave the discussed changes about implementation in the other thread, please, as it's off topic here.

PR #2791, opened 2018-09-15
[Solver PR 3.5] SolverHandler bug fixes

keveleigh:
Overview
---
Removed `[DisallowMultipleComponent]` as it shouldn't have been added in the first place and the InBetween solver depends on multiple handlers.
Refactored out `RefreshTrackedObject()` as a public method to facilitate runtime refreshes.
Fixed a bug where a solver's parent's scale wasn't being accounted for in the desired solver offset, so offsets could be larger or smaller in world space than expected.

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
@@ -11,7 +11,6 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
     /// <summary>
     /// This class handles the solver components that are attached to this <see cref="GameObject"/>
     /// </summary>
-    [DisallowMultipleComponent]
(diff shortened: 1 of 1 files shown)


StephenHodgson:
> Removed [DisallowMultipleComponent] as it shouldn't have been added in the first place and the 

I disagree here, as in between solver should have been adding the second handler component to the other tracked object if it didn't already have one.

keveleigh:
>I disagree here, as the in between solver should have been adding the second handler component to the other tracked object

This is not how solvers work. There is a single object that contains solvers and a solver handler. There is typically one solver handler for any number of solvers, but the InBetween solver works slightly differently, as it has two transforms to track.

The tracked objects are simply a Transform and do not need to have any specific scripts on them.

StephenHodgson:
Oh I see, so in this case we were just tracking two objects, correct?

StephenHodgson:
So why can't the single handler accommodate tracking multiple n+ objects?

Issue #2790, opened 2018-09-14
There appears to be no easy way for a title to get the controller object from the manager


davidkline-ms:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
Based on the MixedRealityManager, I do not appear to be able to access the controller game object. My scenario needs to be able to perform situational tasks on the object (ex; highlight features, hide during portions where they would distract or no input is posssible).

**To Reproduce**
_(Links to sample GitHub project preferred)_
1. Create a scene with the MRManager
2. Try to find the controller object(s).

**Expected behavior**
A simple method / property to return me the collection of controller game objects

**Actual behavior**
Cant find one

StephenHodgson:
`System.object`? `UnityEngine.Object`? `UnityEngine.GameObject`? I don't think I understand the use of terminology here, but I'll assume the latter `GameObject` as the object you're referring to here.

Shouldn't this have been a proposal or a task? 

Seems like kinda a stretch to call this a bug when there wasn't support for getting the controller `GameObject` references.

StephenHodgson:
I'm going to close this in favor of #2799 as this isn't technically a bug and lists out all the requirements.

Please add any additional details or requirements you'd like to see added

PR #2795, opened 2018-09-16
Alternate implementation for Solver Handler Scene Object attachment

SimonDarksideJ:
Overview
---
This implementation enables the Mixed Reality manager to keep a track of any Scene Objects created in the scene by devices and makes them available for Scene components.

This is a preliminary check-in for review

It's critical that devices are not polled or "part" of a scene.  The Current MixedRealityControllerVisualizer is a representation of a controller in a scene, the GameObject of the controller.
To this the device layer appends the SourcePoseController to enable input events from the controller to manipulate the Controller Game object.

This implementation has enabled the Visualizer to register it's existence with the MixedRealityManager, so that other scripts can easily query for the gameobjects for the controllers in the scene.  Scene objects interacting with scene objects.

*Note* **No parts of the device layer have been modified for this implementation**.

Please review and comment where further changes are needed, 

> Additionally, this solves issue #2790 

Changes
---
- Adds new Scene Object Registration system to the Mixed Reality manager (currently "limited to" MotionControllerVisualizers)
 - Adds functionality to the SolverHandler to grab the correct handed controller from the Scene Object Registration
- Added the ISourceStatehandler implementation to the SolverHandler, so it can refresh the solver when a controller is attached / detached.

TODO
---
This is a preliminary implementation, Ideally the Scene Object Registry should be able to handle any game object, but this requires a fair amount of additional design.

# Registry notes

This brings the registry count up to 3, the distinct registries and their purpose are as follows:

* Core manager Registry - Unique registry for managers where there can only ever ne a single instance in operation (e.g. InputSystem, BoundarySystem). Each manager receives all Unity events coordinated through the MR Manager.
* Component Registry - Emulates the managers registry, except allows multiple components to be added.  Each manager receives all Unity events coordinated through the MR Manager.
* Scene Object Registry - List of referencable Scene objects, to avoid use of "FindObjectOfType" lookups.  Used for regularly queries scene objects critical to a project (e.g. VR Controllers represented in a scene)

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
@@ -3,10 +3,11 @@
 
 using Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities;
 using Microsoft.MixedReality.Toolkit.Core.EventDatum.Input;
-using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem;
 using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem.Handlers;
 using Microsoft.MixedReality.Toolkit.Core.Managers;
 using Microsoft.MixedReality.Toolkit.Core.Utilities;
+using Microsoft.MixedReality.Toolkit.Core.Utilities.Async;
+using Microsoft.MixedReality.Toolkit.SDK.Input.Handlers;
 using Microsoft.MixedReality.Toolkit.SDK.UX.Controllers;
 using System.Collections.Generic;
 using UnityEngine;
@@ -16,7 +17,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
     /// <summary>
     /// This class handles the solver components that are attached to this <see cref="GameObject"/>
     /// </summary>
-    public class SolverHandler : MonoBehaviour, IMixedRealitySourceStateHandler
+    public class SolverHandler : BaseInputHandler, IMixedRealitySourceStateHandler
     {
         [SerializeField]
         [Tooltip("Tracked object to calculate position and orientation from. If you want to manually override and use a scene object, use the TransformTarget field.")]
@@ -127,21 +128,23 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
 
         protected readonly List<Solver> solvers = new List<Solver>();
 
+        /// <inheritdoc />
+        public override bool IsFocusRequired => false;
+
         private float lastUpdateTime;
         private GameObject transformWithOffset;
 
         #region MonoBehaviour Implementation
 
-        private void Awake()
+        private async void Awake()
         {
-            //Register this GameObject to receive Input System Events.
-            MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>()?.Register(this.gameObject);
-
             GoalScale = Vector3.one;
             AltScale = new Vector3Smoothed(Vector3.one, 0.1f);
             DeltaTime = 0.0f;
 
             solvers.AddRange(GetComponents<Solver>());
+
+            await WaitUntilInputSystemValid;
 
             // TransformTarget overrides TrackedObjectToReference
             if (trackedObjectToReference != TrackedObjectType.None)
@@ -219,13 +222,16 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
 
         private MixedRealityControllerVisualizer GetVisualizerForHand(Handedness hand)
         {
-            foreach (MixedRealityControllerVisualizer controller in MixedRealityManager.Instance.MixedRealitySceneObjects)
+            for (var i = 0; i < MixedRealityManager.Instance.MixedRealitySceneObjects.Count; i++)
             {
-                if (controller.Handedness == hand)
+                var controller = MixedRealityManager.Instance.MixedRealitySceneObjects[i] as MixedRealityControllerVisualizer;
+
+                if (controller != null && controller.Handedness == hand)
                 {
                     return controller;
                 }
             }
+
             return null;
         }
 
--- a/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
@@ -2,7 +2,12 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities;
+using Microsoft.MixedReality.Toolkit.Core.EventDatum.Input;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem.Handlers;
+using Microsoft.MixedReality.Toolkit.Core.Managers;
 using Microsoft.MixedReality.Toolkit.Core.Utilities;
+using Microsoft.MixedReality.Toolkit.SDK.UX.Controllers;
 using System.Collections.Generic;
 using UnityEngine;
 
@@ -11,11 +16,11 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
     /// <summary>
     /// This class handles the solver components that are attached to this <see cref="GameObject"/>
     /// </summary>
-    public class SolverHandler : MonoBehaviour
+    public class SolverHandler : MonoBehaviour, IMixedRealitySourceStateHandler
     {
         [SerializeField]
         [Tooltip("Tracked object to calculate position and orientation from. If you want to manually override and use a scene object, use the TransformTarget field.")]
-        private TrackedObjectType trackedObjectToReference = TrackedObjectType.Head;
+        private TrackedObjectType trackedObjectToReference = TrackedObjectType.None;
 
         /// <summary>
         /// Tracked object to calculate position and orientation from. If you want to manually override and use a scene object, use the TransformTarget field.
@@ -129,6 +134,9 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
 
         private void Awake()
         {
+            //Register this GameObject to receive Input System Events.
+            MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>()?.Register(this.gameObject);
+
             GoalScale = Vector3.one;
             AltScale = new Vector3Smoothed(Vector3.one, 0.1f);
             DeltaTime = 0.0f;
@@ -136,7 +144,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
             solvers.AddRange(GetComponents<Solver>());
 
             // TransformTarget overrides TrackedObjectToReference
-            if (!TransformTarget)
+            if (trackedObjectToReference != TrackedObjectType.None)
             {
                 AttachToNewTrackedObject();
             }
@@ -190,8 +198,35 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
                 case TrackedObjectType.Head:
                     TrackTransform(CameraCache.Main.transform);
                     break;
-                // Other cases will come online as ControllerFinder is ported appropriately.
+                case TrackedObjectType.LeftController:
+                    var leftController = GetVisualizerForHand(Handedness.Left);
+                    if (leftController != null)
+                    {
+                        TrackTransform(leftController.gameObject.transform);
+                    }
+                    break;
+                case TrackedObjectType.RightController:
+                    var rightController = GetVisualizerForHand(Handedness.Right);
+                    if (rightController != null)
+                    {
+                        TrackTransform(rightController.gameObject.transform);
+                    }
+                    break;
+
+                    // Other cases will come online as ControllerFinder is ported appropriately.
             }
+        }
+
+        private MixedRealityControllerVisualizer GetVisualizerForHand(Handedness hand)
+        {
+            foreach (MixedRealityControllerVisualizer controller in MixedRealityManager.Instance.MixedRealitySceneObjects)
+            {
+                if (controller.Handedness == hand)
+                {
+                    return controller;
+                }
+            }
+            return null;
         }
 
         private void TrackTransform(Transform newTrackedTransform)
@@ -212,5 +247,29 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
             transformWithOffset.name = string.Format("{0} on {1} with offset {2}, {3}", gameObject.name, TrackedObjectToReference.ToString(), AdditionalOffset, AdditionalRotation);
             return transformWithOffset.transform;
         }
+
+        #region IMixedRealitySourceStateHandler implementation
+
+        public void OnSourceDetected(SourceStateEventData eventData)
+        {
+            if (trackedObjectToReference != TrackedObjectType.None)
+            {
+                AttachToNewTrackedObject();
+            }
+        }
+
+        public void OnSourceLost(SourceStateEventData eventData)
+        {
+            if (transformWithOffset != null)
+            {
+                Destroy(transformWithOffset);
+            }
+            if (trackedObjectToReference != TrackedObjectType.None)
+            {
+                TransformTarget = null;
+            }
+        }
+
+        #endregion IMixedRealitySourceStateHandler implementation
     }
 }
--- a/Assets/MixedRealityToolkit/_Core/Definitions/Utilities/TrackedObjectType.cs
+++ b/Assets/MixedRealityToolkit/_Core/Definitions/Utilities/TrackedObjectType.cs
@@ -3,19 +3,26 @@
 
 namespace Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities
 {
+    /// <summary>
+    /// Defines the types of automatic objects that can be tracked
+    /// </summary>
     public enum TrackedObjectType
     {
         /// <summary>
+        /// Do not automatically track an object, use the manual reference instead.
+        /// </summary>
+        None = 0,
+        /// <summary>
         /// Calculates position and orientation from the main camera.
         /// </summary>
-        Head = 0,
+        Head,
         /// <summary>
-        /// Calculates position and orientation from the left motion-tracked controller.
+        /// Calculates position and orientation from the left tracked controller.
         /// </summary>
-        MotionControllerLeft,
+        LeftController,
         /// <summary>
-        /// Calculates position and orientation from the right motion-tracked controller.
+        /// Calculates position and orientation from the right tracked controller.
         /// </summary>
-        MotionControllerRight
+        RightController
     }
 }
--- a/Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealitySceneObject.cs
+++ b/Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealitySceneObject.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Core.Interfaces
+{
+    /// <summary>
+    /// Generic interface for all optional Mixed Reality systems, components, or features that can be added to the <see cref="Definitions.MixedRealityComponentConfiguration"/>
+    /// </summary>
+    public interface IMixedRealitySceneObject 
+    {
+        /// <summary>
+        /// Register the gameobject with the Mixed Reality Manager Scene Object registry
+        /// </summary>
+        void RegisterSceneObject();
+
+        /// <summary>
+        /// Unregister the gameobject with the Mixed Reality Manager Scene Object registry
+        /// </summary>
+        void UnregisterSceneObject();
+    }
+}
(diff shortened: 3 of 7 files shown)


Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -114,6 +114,10 @@ public void ResetConfiguration(MixedRealityConfigurationProfile profile)
 
         #endregion Mixed Reality runtime component registry
 
+        #region Mixed Reality scene object registry
+        public List<IMixedRealitySceneObject> MixedRealitySceneObjects { get; set; } = new List<IMixedRealitySceneObject>();
```
StephenHodgson: read only list?

SimonDarksideJ: Err, no.  This is constantly updated as new Scene objects are added to / destroyed in the scene.
Unless I fundamentally misunderstand what a Readonly list does :D

StephenHodgson: Just that there's only ever one list, created when the MRM is created.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealitySceneObject.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Core.Interfaces
+{
+    /// <summary>
+    /// Generic interface for all optional Mixed Reality systems, components, or features that can be added to the <see cref="Definitions.MixedRealityComponentConfiguration"/>
+    /// </summary>
+    public interface IMixedRealitySceneObject
```
StephenHodgson: Should we add the `TrackedObjectType` to this? 

Might make it easier to determine the type of object that's registered.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -190,10 +198,37 @@ protected virtual void AttachToNewTrackedObject()
...
+                    }
+                    break;
+
+                    // Other cases will come online as ControllerFinder is ported appropriately.
             }
         }
 
+        private MixedRealityControllerVisualizer GetVisualizerForHand(Handedness hand)
+        {
+            foreach (MixedRealityControllerVisualizer controller in MixedRealityManager.Instance.MixedRealitySceneObjects)
```
StephenHodgson: for?

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -190,10 +198,37 @@ protected virtual void AttachToNewTrackedObject()
...
+            {
+                if (controller.Handedness == hand)
```
StephenHodgson: Oh I see, instead of using the `TrackedObjectType` we just get the controller's handedness.

Assets/MixedRealityToolkit/_Core/Interfaces/IMixedRealitySceneObject.cs
SimonDarksideJ: I don't think so, as the TrackedObject attachment is unique to the solvers atm.
Plus, (eventually) this could just contain regular scene objects that don't care whether they are tracked.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
SimonDarksideJ: I thought about it.  However, this call is used so infrequently I didn't think we needed to optimise it.
Open to review of course :D (as per standards)

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -129,14 +134,17 @@ public bool UpdateSolvers
 
         private void Awake()
         {
+            //Register this GameObject to receive Input System Events.
+            MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>()?.Register(this.gameObject);
```
StephenHodgson: Why not inherit from `BaseInputHandler`?

Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
StephenHodgson: Not so keen on the name either. `MixedRealitySceneGameObjects`?

`object != Object != GameObject`

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
StephenHodgson: We should probably be checking if the controller is null as well, just in case. changing to a for loop might make that a bit simpler.

keveleigh: This doubles down on the current requirement that model overrides must have a pose synchronizer on them.

This has already been given as feedback that it adds an unnecessary requirement (as in, there are ways to do this without adding a script; see: HTK) to adding arbitrary prefabs as overrides, as well as likely being incompatible once we load platform controller models and create GameObjects for them at runtime, since the core has no knowledge of the concrete pose synchronizers.

keveleigh:
>It's critical that devices are not polled or "part" of a scene.

Again, why is this critical? Not because of your definition of what a device is, but what actual effect does this have on an application or the Toolkit?

StephenHodgson:
> It's critical that devices are not polled or "part" of a scene.

> Again, why is this critical? Not because of your definition of what a device is, but what actual effect does this have on an application or the Toolkit?

Good question. I'm not so sure about the polling being crucial as we already give devs the ability to get the current info from the controller at any time, so we're not preventing something like this from happening, **but** it _is_ device specific, so you'll need a concrete type to poll from, which breaks cross platform support, and why we encourage the use of getting data from the input system via `IMixedRealityControllerPoseSynchronizer`.

As for not being part of the scene? That's because the device layer is Encapsulated. All it knows about is the device and the data being read from the device. The "Scene" part of the controller is the `IMixedRealityControllerPoseSynchronizer` which is being driven by the `IMixedRealityController`.

We're separating our concerns here and making sure we're sticking to good OOP design.

> This has already been given as feedback that it adds an unnecessary requirement (as in, there are ways to do this without adding a script; see: HTK)

For one, this isn't the HTK, so we need to stop trying to compare it. Second, where is this feedback? Third, without the `IMixedRealityControllerPoseSynchronizer` you won't know anything about the controller unless you implement it. So yes, it's a requirement for scene based controllers, as it's the representation of that controller.

That being said, this doesn't mean it's a requirement for _all_ tracked objects by the SolverHandler.

keveleigh:
>but it is device specific, so you'll need a concrete type to poll from, which breaks cross platform support

I do not poll from any concrete type. I only call methods from the base interfaces, and this has been tested and works completely cross platform.

>For one, this isn't the HTK, so we need to stop trying to compare it. 

No, but why would we make things worse? Adding a requirement to add a script to an object to obtain the same result as the HTK is objectively not as good.

>Second, where is this feedback? 

I've heard it internally, and you also told me there was confusion at the Summit when arbitrary prefabs didn't "just work".

keveleigh:
>Second, where is this feedback?

And you're right, it should have been filed as an issue. I've been working on a proposal for a real controller visualization solution and was going to bring it up at that time.

StephenHodgson:
> I do not poll from any concrete type. I only call methods from the base interfaces, and this has been tested and works completely cross platform.

I know you don't I was agreeing with you 😅 That was more or less for @SimonDarksideJ.

> No, but why would we make things worse?

Seems rather subjective, but I agree we should try to reach similar results to the the HTK in regards to feature parity.

> Adding a requirement to add a script to an object to obtain the same result

I don't understand why this is objectively a bad idea, as we've broken out requirements and responsibilities of the scripts to be more encapsulated and do single task. That's just good OOP design.

keveleigh:
>I don't understand why this is objectively a bad idea

Think about it from the point of view of a consumer of this product: a developer working on their app.

In the HTK, they dragged a prefab somewhere and it became the controller model, tracked with the controller, and worked.

In the MRTK, if they do the exact same thing, nothing happens. They now have to add a script onto their prefab to obtain the same results. That prefab might be used elsewhere and shouldn't be modified. They're also now doing an extra step on top of the HTK steps to obtain the same result.

StephenHodgson:
That's because you need to use the configuration profile. You can't just add these things to the scene, as the Mixed Reality Manager handles all of that. (It's how we were able to get away from using Singletons).

That's just part of the documentation we're currently writing up.

keveleigh:
>Also, that's what the [RequireComponent(type)] attribute is for.

What problem does this solve in this scenario?

keveleigh:
>That's because you need to use the configuration profile. You can't just add these things to the scene

Yes, they did not just drag them into the scene in the HTK either. In both cases, the prefabs are being set in the inspector.

StephenHodgson:
> In the HTK, they dragged a prefab somewhere and it became the controller model, tracked with the controller, and worked.

I think most of the frustration here is that this entire system _isn't like the HTK at all_, sans the input system events. I will reiterate again, _this is not the HTK_.

> They're also now doing an extra step on top of the HTK steps to obtain the same result.

I don't see it that way at all.

To setup a controller visualization you just need to create your prefab, attach the visualization script (which isn't even supply fleshed out yet, so maybe they also assign all their transforms who knows) and then assign their controller prefab in the controller mapping profile. No adding anything to your scene.

SimonDarksideJ:
Can you simply list out your issues with the proposal @keveleigh ?
It has achieve all the requirements set out from your taks and even addressed @davidkline-ms 's issue. As well as starting an extensible registration system for managing any scene object.

All without compromising the base architecture and dirtying the Device layer with runtime / scene data.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -190,10 +198,37 @@ protected virtual void AttachToNewTrackedObject()
                 case TrackedObjectType.Head:
                     TrackTransform(CameraCache.Main.transform);
                     break;
-                // Other cases will come online as ControllerFinder is ported appropriately.
+                case TrackedObjectType.LeftController:
+                    var leftController = GetVisualizerForHand(Handedness.Left);
```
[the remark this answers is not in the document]
StephenHodgson: Eric, I think you nailed it right on the head. I'm down to update the terminology to better reflect mvvm.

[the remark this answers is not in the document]
SimonDarksideJ: Agree, an interesting way to re"visualize" the concept to another pattern.
* A Device is the "Model" - data
* The DeviceManager/MRManager are the View - connecting a model to a vm
* The scene "things" are the view models - providing cuntionality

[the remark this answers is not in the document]
Ecnassianer: Hmmm, maybe I don't understand the role of the DeviceManager/MRManager, @SimonDarksideJ ?

>  the View - connecting a model to a vm

In MVVM, a View doesn't connect a model to a VM, in fact, it may not necessarily even correspond to a specific model of the data (though it does correspond to the *kind* of data that's stored in that model [a birthdate View can't connect to a Model of a audio clip]). The ViewModel is the class responsible for translating things like Unix Time to a string, and the View handles how strings look on the screen. A robust View Model might support several views (VM.TimeAsAudio(), VM.TimeASSentence(), etc). A View just wants a string or a number, some data to bind to. You could rewrite the model from a binary tree into an SQL database into a decryption tool and the View wouldn't need to change at all, it only talks to the ViewModel.

My understanding was that the Managers are sort of utility classes that help the Scene Representations stay coordinated/initialized with the device models, and helps the Views (User Abstraction Layer) get references to the right Scene Representations. I could see that described as part of the View Model, or alternatively simply a utility outside the MVVM paradigm (like a DNS service isn't specifically a part of the client-server paradigm, but is a necessary utility for implementing a client-server paradigm).

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -190,10 +201,40 @@ protected virtual void AttachToNewTrackedObject()
...
             }
         }
 
+        private MixedRealityControllerVisualizer GetVisualizerForHand(Handedness hand)
+        {
+            for (var i = 0; i < MixedRealityManager.Instance.MixedRealitySceneObjects.Count; i++)
+            {
+                var controller = MixedRealityManager.Instance.MixedRealitySceneObjects[i] as MixedRealityControllerVisualizer;
+
+                if (controller != null && controller.Handedness == hand)
```
Ecnassianer: I am a little concerned about the assumption that there is only one right handed motion controller at a particular time. There's nothing that prevents a person from connecting multiple bluetooth controllers. I might be a fine choice to just give them the first right handed controller the same way GetComponent<>() just gives the first component it finds, but I wish it was a little more explicit (or we should also offer a GetVisualizersForHand() that returns an array of all of them the same way there is a GetComponents<>()?)

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -11,11 +17,11 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Utilities.Solvers
     /// <summary>
     /// This class handles the solver components that are attached to this <see cref="GameObject"/>
     /// </summary>
-    public class SolverHandler : MonoBehaviour
+    public class SolverHandler : BaseInputHandler, IMixedRealitySourceStateHandler
     {
         [SerializeField]
         [Tooltip("Tracked object to calculate position and orientation from. If you want to manually override and use a scene object, use the TransformTarget field.")]
-        private TrackedObjectType trackedObjectToReference = TrackedObjectType.Head;
+        private TrackedObjectType trackedObjectToReference = TrackedObjectType.None;
```
Ecnassianer: None doesn't seem like a good default here, since it essentially means when it gets added it is in a non-functional state. Head is a better safe default.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -122,21 +128,26 @@ public bool UpdateSolvers
...
+        /// <inheritdoc />
+        public override bool IsFocusRequired => false;
+
         private float lastUpdateTime;
         private GameObject transformWithOffset;
 
         #region MonoBehaviour Implementation
 
-        private void Awake()
+        private async void Awake()
```
Ecnassianer: Doesn't making Awake asychronous violate Unity's order of operations contract?
https://docs.unity3d.com/Manual/ExecutionOrder.html

Though the Awake() docs don't specifically say Awake cannot be async, it does say it can't be a coroutine:
https://docs.unity3d.com/ScriptReference/MonoBehaviour.Awake.html

It seems safer to stick with the spirit of Awake() and finish up before any Start()s get called.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/TrackedObjectType.cs
```diff
@@ -3,19 +3,26 @@
...
 {
+    /// <summary>
+    /// Defines the types of automatic objects that can be tracked
+    /// </summary>
     public enum TrackedObjectType
     {
+        /// <summary>
+        /// Do not automatically track an object, use the manual reference instead.
+        /// </summary>
+        None = 0,
```
Ecnassianer: I like this addition, for completeness.

Ecnassianer:
Aside from the comments I left in the review, I don't have any concerns with this code itself, but I still feel pretty vague about how someone actually uses this stuff in Unity.

A particular mystery to me comes up in this line:
foreach (MixedRealityControllerVisualizer controller in MixedRealityManager.Instance.MixedRealitySceneObjects

Where do MixedRealityControllerVisualizers come from? How is MixedRealityManager.Instance.MixedRealitySceneObjects populated? If this is the implementation we go with, I want that list to be automatically populated by the MixedRealityManager or something.

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
StephenHodgson: It does't change the timing of the order of operations.

StephenHodgson: But if it is awaited, then yes, it'll delay that call.

StephenHodgson:
>Where do MixedRealityControllerVisualizers come from? How is MixedRealityManager.Instance.MixedRealitySceneObjects populated? If this is the implementation we go with, I want that list to be automatically populated by the MixedRealityManager or something.

That's what I want to address with https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2799

Ecnassianer:
We should consider this user story as a requirement for this work:

As a content creator, I can get a solver working with not more steps than this:
1) Import MRTK into a fresh project.
2) Run MRTK Configure in my scene.
3) Add a cube to the scene.
4) Apply a Radial Solver to the cube.
5) In the Solver Handler (which has been added for me), I set the tracked object to Motion Controller Right
6) In the Radial Solver I set the min distance to 0.2, and the max distance to 0.7
7) I hit play and the cube is following my motion controller in the scene

keveleigh:
>I still feel pretty vague about how someone actually uses this stuff in Unity

That's probably my main concern as well (though for a slightly different reason, I think). This seems like a generic answer searching for additional problems to solve. What other scenario would use this new scene object registry? It seems like, if we do visualization right, we could obtain the same information by asking the visualization system explicitly for a right/left controller instead of needing to iterate through a list filled with some number of objects (that aren't all controller visualizations), casting them all to find something that's both a controller visualization and the correct handedness.

Ecnassianer:
>That's what I want to address with #2799

Cool, I'm looking forward to seeing that PR, I think it will help clarify a lot. :)

Ecnassianer:
>It seems like, if we do visualization right, we could obtain the same information by asking the visualization system explicitly for a right/left controller instead of needing to iterate through a list filled with some number of objects (that aren't all controller visualizations), casting them all to find something that's both a controller visualization and the correct handedness.

Now that you mention that, it definitely feels like a "Gimme The Left Controller" utility function that some manager should provide rather than being in this class, rather than writing it in every User Abstraction class that needs access to controllers. (Whether that function loops over a list of everything or keeps controllers separate is a different question that I have less opinions on.)

Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -190,10 +198,37 @@ protected virtual void AttachToNewTrackedObject()
...
+                    {
+                        TrackTransform(rightController.gameObject.transform);
+                    }
+                    break;
+
+                    // Other cases will come online as ControllerFinder is ported appropriately.
             }
         }
 
+        private MixedRealityControllerVisualizer GetVisualizerForHand(Handedness hand)
```
Ecnassianer: This isn't specific to SolverHandlers, and should be a utility function offered by another class/library. Otherwise we'll write it over and over again. Is there already a good place to move it?

SimonDarksideJ: Agreed, I tried to make this is as generic as possible, but I take your point that "hand" probably isn't the correct verb here.

SimonDarksideJ:
Given the discussions, I agree that #2799 should be resolved before this is merged.
We seem to have consensus on the pattern but need a better visualization manager.  It may be even possible to drop the SceneObject registry if the visualizer also offers a way to query for connected controllers/hands.

Then the Solver update would be redirected / updated to use the new Vizualizer logic

Ecnassianer:
>Then the Solver update would be redirected / updated to use the new Vizualizer logic

Since #2799 is still a bit off, how do you guys feel about merging in https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2784 (or whichever of the proposed implementations) with the understanding that it's just temporary until the Visualizer logic is done? I and other members of my team would like to start using Solvers in MRTK. :)

SimonDarksideJ:
Well, this PR is more in line with the direction that seems to have the most agreement, although I'd like some more feedback on the general approach.

Understand you need solvers asap (although I have raised another PR - #2798 ,  to update Solvers to a more MRTK configurable approach, which would also be good to get feedback on) so we do need to get a  general consensus.

StephenHodgson:
I'm not in favor of merging #2784

StephenHodgson:
I'll have #2799 finished tonight

Ecnassianer:
>I'll have #2799 finished tonight

Sweet, thank you for prioritizing it. :)

SimonDarksideJ:
I've been pitching in and giving @StephenHodgson a hard time with a PR on his PR 
https://github.com/StephenHodgson/MixedRealityToolkit-Unity/pull/32
Feel free to jump in and comment @Ecnassianer , It'll help to carve the way forward

Ecnassianer:
Sorry, I'm blocking a big initiative and need to crunch on some project code today. I won't be able to look at this until tomorrow. :(

SimonDarksideJ:
Closing as implementation is out of date with current plans

Ecnassianer:
We spoke about this in another thread which I can't find at the moment, but can we change the name from "Visualizer" to something else. This PR seems like the place to make that change. Here are some things that I think would more clearly reveal the purpose of these classes:

SceneRepresentation (i.e, ControllerSceneRepresentation or  ControllerRepresentation)
DeviceAvatar (i.e, ControllerDeviceAvatar)
SceneObject
TrackingProxy
TrackedProxy
MixedRealityProxy
RealityProxy

Let me know if you want a recap of why I think Visualizer is a misleading name.

Issue #2744, opened 2018-09-11
When unloading and loading a new scene (Non additive), managers do not reinitialize properly


wassx:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
mrtk_development

**Describe the bug**
GazeInput is not registered on Start. RaiseSourceDetected is called in GazeProvider#OnStart but `await WaitUntilInputSystemValid` does not continue.
When quitting application Assertion fails:
```
Assertion failed: Gaze was never registered with the Input Manager!
UnityEngine.Debug:Assert(Boolean, String)
Microsoft.MixedReality.Toolkit.SDK.Input.MixedRealityInputManager:RaiseSourceLost(IMixedRealityInputSource, IMixedRealityController) (at Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs:521)
Microsoft.MixedReality.Toolkit.SDK.Input.GazeProvider:OnDisable() (at Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs:343)
```

**To Reproduce**
1. Start a scene with MixedRealityManager and a GazeProvider with Default Cursor
2. Cursor is not visible
3. Cursor does not appear on object with collider

**Actual behavior**
Cursor stays in the middle of the head and is inactive

**Unity Editor Version**
2018.1.9f2

**Mixed Reality Toolkit Release Version**
vNext (mrtk_development)
Rev: 7c960a6e46c331171900f24e09e3935e6ccc583a

wassx:
Tested in a freshly created project with nothing than a cube in the scene, and there it works.

wassx:
Retested and could reproduce in multiple scene scenarios (Networkmanager loads next scene).
This is related to: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2653
Is there a quick workaround please? (e.g. remove the wait and let gracefully re-register the input?)

StephenHodgson:
Did you mark the camera as persistent in your camera profile?
![image](https://user-images.githubusercontent.com/13334553/45380035-88499880-b5b6-11e8-851f-2d2970462b1d.png)

wassx:
Yes, tick is set and camera is not destroyed.

StephenHodgson:
You only have one Mixed Reality Manager and no camera's in your other scenes?

wassx:
Yes. Additionally there is an error:
[...]

PR #2796, opened 2018-09-17
Gazeprovider input system issue due to scene loads

wassx:
Overview
---
WaitUntil() InputSystem is active did not work for multiple scene setup. It did not continue despite the InputSystem was !=null from the beginning.
Workaround for: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2744

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Collections;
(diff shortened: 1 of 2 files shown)


SimonDarksideJ:
Agree with @StephenHodgson here.  Although Coorutines are more Unity friendly, we should probably asses the multi-scene case with the current Async method which is a little more code robust.

Is there a test scene / replication instructions for testing this issue?

StephenHodgson:
> Is there a test scene / replication instructions for testing this issue?

Yes: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2744#issuecomment-420374532

StephenHodgson:
I don't think it has anything to do with async vs coroutine specifically as it's unity's network manager that's forcing the scene to unload everything. we just need to make sure we support this edge case with the Mixed Reality Manager.

Issue #876, opened 2017-08-18
Thoughts on the Input system for MRToolkit


witian:
![longbow_input_diagram](https://user-images.githubusercontent.com/5750532/29450618-4d18be7c-83f7-11e7-9abe-5f984e17a707.PNG)
[InputSandbox.zip](https://github.com/Microsoft/MixedRealityToolkit-Unity/files/1233768/InputSandbox.zip)

Hello!

I'm from the Horizon team in London (Microsoft) and I've recently joined this repo as a contributor. :) We (Horizon) had developed an Input module for Unity over a year ago that is simple yet elegant. However, in order to avoid duplicating systems we've decided to start using the MRToolkit's input system. We would encourage you to have a look at it (attached) if you are interested. We've taken the approach to make input work "out-of-the-box" keeping inline with Unity's style of input. Simply add "HoloLensInputModuleEx" (bad name, I know...) to your EventSystem, set your cursor, and you're done! Uses IPointerHandlers for the gaze handlers. Doing this enables you to take advantage of using built in systems in Unity without adding additional scripts on your objects. It also simplifies input interaction. Note that you can click on the "Help" field in the component in the Inspector to see how to use the in-Editor simulation.

Documentation:
- Docs/Input/html/index.html

Samples:
- Gaze Input: Assets/Longbow/Input/Gaze/Samples
- Hand Input: Assets/Longbow/Input/Hands/Samples
- Speech Input: Assets/Longbow/Input/Speech/Samples

A few good takeaways would be a few extra features we have in the input module:
- When running in the editor you can tilt the camera left and right using "q" and "e" (You need to press "shift+cntrl" first to initiate camera movement).
- You can change the depth of the hand positions in the Editor by using the scroll wheel on your mouse.
- You can do two-hand simulations by pressing '1' (left hand) + '2' (right hand) or simply '3'.
- Input Module doesn't require special handlers. Works with the Canvas without adding any extra handlers or listeners to UI components.
- HandRecognizer can run independently from the input module. It also will allow simulation when running in the editor.
- Speech Recognition is simple. Just attach the "Contextual Speech Recognition" component to any object in your scene and you get instant speech reco! It also understand gaze context.

Cheers,
William

witian:
BTW, the attached unity project was built for Unity 5.5.4

StephenHodgson:
Any plans for support with the new `Gaze and point` / `Gaze and Commit` APIs that are coming out with XR namespace changes?

[...]

PR #2800, opened 2018-09-17
vNEXT: Adding Interaction and button components to UX features

killerantz:
Overview
---

Changes
---
- Fixes: # .


StephenHodgson:
Any way we can break this up into smaller pieces to review?

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scenes/ButtonTest.unity
```diff
@@ -0,0 +1,973 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
```
StephenHodgson: Test scenes shouldn't be in the SDK

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Events/OnClickReceiver.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            bool hadDown = hasDown;
+            hasDown = state.GetState(InteractableStates.InteractableStateEnum.Pressed).Value > 0;
+
+            bool focused = state.GetState(InteractableStates.InteractableStateEnum.Focus).Value > 0;
+
+            if (hadDown && !hasDown && focused && clickTimer < ClickTime)
+            {
+                Debug.Log("Click!");
+                uEvent.Invoke();
```
StephenHodgson: Not sure we should be using Unity Events. We've got a pretty solid event system we're already using that doesn't require uGUI.

StephenHodgson:
I think it would be helpful to add the materials, icons, and textures in a first pass, then any scripts in a multi pass review, then the scenes last. 

I'm going to also suggest that the test scenes be moved over to examples, and test be dropped from the name.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scenes/ButtonTest.unity
killerantz: For me this is still a work in progress until we get all the functionality, like down and up on HoloLens working. I agree with your other comment below about moving the scene to examples. I may end up replacing it at some point or at least repurposing it to contain all the button components.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Events/OnClickReceiver.cs
killerantz: The UnityEvents are for designers to use in editor, unless we have an event system with a property drawer. I'm happy to change how the events work, but the main point is that designers can hook them up without code.

StephenHodgson: >the main point is that designers can hook them up without code.

Def agree! Let's sync and discuss how we can accommodate this.

My other goal as well is to try and not be dependent on unity gui as well. That's mainly for 2d screens and wasn't really designed with 3d in mind.

StephenHodgson: Oh, I thought that because we were using unity events that they it was strictly uGUI based. In that case, we can def try to stick to the handlers we've setup (very similar to old HTK input). We should still be able to setup events in the inspector without any code. (take a look at PointerClickHandler and SpeechInputHandler classes as examples). We can still use unity events in a small capacity where needed, but I'd like to make sure we're just exposing the fields we need to reference when handling events.

StephenHodgson: Take a look at [my comments from this issue related to Unity Events](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/876#issuecomment-323417651)

killerantz: For sure, I expose the basic click event because if follows the patterns on button, but there are a ton of other events that the designer can add as needed and they are very specific. In the last version of InteractiveElements I thought this button could be used in 5 ways, so I exposed 5 UnityEvents. That just makes things complicated and quite frankly it was easier to building custom inspectors. Now only one event is shown at default and if the design wants another event, they can add that specific event. The same goes for an engineer that would not like to use the inspector, but stick to code, they can target a pattern that defines an event. This part still needs some work though. I know there's a ton here. Thanks for taking the time.

StephenHodgson: I def love where this is going, and there's a ton of awesome work you (and I'm sure others) have put into this. I'd like to dig in a bit more before I add any additional feedback.

SimonDarksideJ:
We should also ensure to check what materials, icons, and textures can fit in the Standard Assets pack in the SDK and what are "Demo" specific.  Aiming to have good clean reusable assets that can be shared.

cre8ivepark:
@SimonDarksideJ Definitely the materials used in foundational UI components such as HolographicButton.prefab should go into SDK.

killerantz:
I'm closing this and will resubmit in pieces after doing some more cleanup. There's some redundant code and getting all the base assets in.

PR #2801, opened 2018-09-17
Found an error when using MRTK in other projects

StephenHodgson:
Overview
---
Added ability to get the Mixed Reality Toolkit's folder directory path.


Assets/MixedRealityToolkit/_Core/Utilities/Editor/Setup/EnforceEditorSettings.cs
```diff
@@ -148,9 +166,38 @@ private static bool IsNewSession()
             return false;
         }
 
+        private static bool FindDirectory(string directoryPathToSearch, string directoryName, out string path)
```
StephenHodgson: I suppose I could make this an extension, but this class is supposed to be self sufficient as it runs before all other assemblies are compiled by Unity.

SimonDarksideJ: I think this is fine here.

PR #2804, opened 2018-09-18
vNext: InteractionReceiver base class

radicalad:
### Overview
Porting the original InteractionReceiver from HTK. This script registers itself with the `InputSystem` and listens for all events available for GameObjects in the Interactables list. To extend, simply override the virtual methods. This is a core component used by many others (namely App bar). Unfortunately, since InteractionReceiver is not necessarily a "UX" component, I can't make a demo scene - this is more foundational and unblocks other components.

### Changes
-- **InteractionReceiver** now implements `IMixedRealityGestureHandler<T>`

### Note
I put the InteractionReceiver in the `SDK.UX.Receivers` namespace, but it might make more sense in `SDK.Input.Receivers`

### Interaction Receiver in HoloToolkit
https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit-Examples/UX/Readme/README_InteractableObjectExample.md
receiver
https://medium.com/@dongyoonpark/open-source-building-blocks-for-windows-mixed-reality-experiences-hololens-mixedrealitytoolkit-28a0a16ebb61

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
@@ -123,7 +123,6 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
             interactables.Add(interactable);
         }
 
-#if UNITY_EDITOR
         /// <summary>
         /// When selected draw lines to all linked interactables
         /// </summary>
@@ -157,7 +156,6 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
                 }
             }
         }
-#endif
 
         /// <summary>
         /// Function to remove an interactable from the linked list.
@@ -237,17 +235,14 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
 
         public void OnFocusChanged( FocusEventData eventData )
         {
-            FocusEventData newData = new FocusEventData( EventSystem.current );
-            newData.Initialize( eventData.Pointer );
-
             if ( eventData.NewFocusedObject != null && IsInteractable( eventData.NewFocusedObject ) )
             {
-                FocusEnter( eventData.NewFocusedObject, newData );
+                FocusEnter( eventData.NewFocusedObject, eventData );
             }
 
             if ( eventData.OldFocusedObject != null && IsInteractable( eventData.OldFocusedObject ) )
             {
-                FocusExit( eventData.OldFocusedObject, newData );
+                FocusExit( eventData.OldFocusedObject, eventData );
             }
 
             CheckLockFocus( eventData.Pointer );
--- a/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
@@ -60,15 +60,14 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
 
         #endregion
 
-        #region private variables
         /// <summary>
         /// When true, this interaction receiver will draw connections in the editor to Interactables and Targets
         /// </summary>
         [Tooltip( "When true, this interaction receiver will draw connections in the editor to Interactables and Targets" )]
         [SerializeField]
         private bool drawEditorConnections = true;
-        
-        #endregion
+
+        #region MonoBehaviour implementation
 
         /// <summary>
         /// On enable, set the BaseInputHandler's IsFocusRequired to false to receive all events.
@@ -77,20 +76,6 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
         {
             IsFocusRequired = false;
             base.OnEnable();
-        }
-
-        /// <summary>
-        /// Register an interactable with this receiver.
-        /// </summary>
-        /// <param name="interactable">takes a GameObject as the interactable to register.</param>
-        public virtual void RegisterInteractable( GameObject interactable )
-        {
-            if ( interactable == null || interactables.Contains( interactable ) )
-            {
-                return;
-            }
-
-            interactables.Add( interactable );
         }
 
         /// <summary>
@@ -130,6 +115,21 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
             }
         }
 
+        #endregion
+
+        /// <summary>
+        /// Register an interactable with this receiver.
+        /// </summary>
+        /// <param name="interactable">takes a GameObject as the interactable to register.</param>
+        public virtual void RegisterInteractable( GameObject interactable )
+        {
+            if ( interactable == null || interactables.Contains( interactable ) )
+            {
+                return;
+            }
+
+            interactables.Add( interactable );
+        }
 
         /// <summary>
         /// Function to remove an interactable from the linked list.
@@ -166,6 +166,8 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
             return ( interactables != null && interactables.Contains( interactable ) );
         }
 
+        #region Global Listener Callbacks
+
         public void OnBeforeFocusChange( FocusEventData eventData ) { /*Unused*/ }
 
         public void OnFocusChanged( FocusEventData eventData )
@@ -180,8 +182,6 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
                 FocusExit( eventData.OldFocusedObject, eventData );
             }
         }
-
-        #region Global Listener Callbacks
 
         public void OnGestureStarted( InputEventData eventData )
         {
@@ -244,6 +244,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
             }
 
         }
+
         public void OnGestureCompleted( InputEventData<Vector2> eventData )
         {
             if ( IsInteractable( eventData.selectedObject ) )
@@ -270,7 +271,6 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
             }
 
         }
-
         public void OnGestureCanceled( InputEventData eventData )
         {
             if ( IsInteractable( eventData.selectedObject ) )
@@ -278,6 +278,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
                 GestureCanceled( eventData.selectedObject, eventData );
             }
         }
+
         public void OnInputUp( InputEventData eventData )
         {
             if ( IsInteractable( eventData.selectedObject ) )
--- a/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
@@ -385,6 +385,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
         #endregion
 
         #region Protected Virtual Callback Functions
+
         protected virtual void FocusEnter(GameObject obj, FocusEventData eventData ) { }
         protected virtual void FocusExit(GameObject obj, FocusEventData eventData ) { }
 
--- a/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
@@ -1,0 +1,414 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using UnityEngine.EventSystems;
+using System.Collections.Generic;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem.Handlers;
+using Microsoft.MixedReality.Toolkit.Core.EventDatum.Input;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.Devices;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.InputSystem;
+using Microsoft.MixedReality.Toolkit.Core.Managers;
+using Microsoft.MixedReality.Toolkit.SDK.Input;
+using Microsoft.MixedReality.Toolkit.SDK.Input.Handlers;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Receivers
+{
+    /// <summary>
+    /// An interaction receiver is simply a component that attached to a list of interactable objects and does something
+    /// based on events from those interactable objects.  This is the base abstract class to extend from.
+    /// </summary>
+    public abstract class InteractionReceiver : BaseInputHandler,
+        IMixedRealityFocusChangedHandler,
+        IMixedRealityGestureHandler,
+        IMixedRealityInputHandler,
+        IMixedRealityGestureHandler<Vector2>,
+        IMixedRealityGestureHandler<Vector3>,
+        IMixedRealityGestureHandler<Quaternion>
+
+    {
+        #region Public Members
+        /// <summary>
+        /// List of linked interactable objects to receive events for
+        /// </summary>
+        [SerializeField]
+        [ Tooltip("Target interactable Object to receive events for")]
+        private List<GameObject> interactables = new List<GameObject>();
+
+        public List<GameObject> Interactables
+        {
+            get { return interactables; }
+            set { value = interactables; }
+        }
+
+        /// <summary>
+        /// List of linked targets that the receiver affects
+        /// </summary>
+        [Tooltip("Targets for the receiver to ")]
+        private List<GameObject> targets = new List<GameObject>();
+
+        public List<GameObject> Targets
+        {
+            get { return targets; }
+            set { value = targets; }
+        }
+
+        [Tooltip( "If true, this object will remain the prime focus while select is held" )]
+        [SerializeField]
+        private bool lockFocus = false;
+
+        /// <summary>
+        /// Flag for locking focus while selected
+        /// </summary>
+        public bool LockFocus
+        {
+            get
+            {
+                return lockFocus;
+            }
+            set
+            {
+                lockFocus = value;
+                CheckLockFocus(_selectingFocuser);
+            }
+        }
+        
+        #endregion
+
+        #region Private and Protected Members
+
+        /// <summary>
+        /// Protected focuser for the current selecting focuser
+        /// </summary>
+        protected IMixedRealityPointer _selectingFocuser;
+        
+        #endregion
+
+        /// <summary>
+        /// On start subscribe to all interaction events on elements in the interactables list.
+        /// </summary>
+        protected override void OnEnable()
+        {
+            base.OnEnable();
+            //Still need to register since focus isn't required
+            if(InputSystem != null && InputSystem.IsInputEnabled)
+            {
+                InputSystem.Register( gameObject );
+            }
+        }
+
+        /// <summary>
+        /// On disable remove all linked interactables from the delegate functions
+        /// </summary>
+        protected override void OnDisable()
+        {
+            //Clean up our event subscription
+            if ( InputSystem != null && InputSystem.IsInputEnabled )
+            {
+                InputSystem.Unregister( gameObject );
+            }
+            base.OnDisable();
+        }
+
+        /// <summary>
+        /// Register an interactable with this receiver.
+        /// </summary>
+        /// <param name="interactable">takes a GameObject as the interactable to register.</param>
+        public virtual void Registerinteractable(GameObject interactable)
+        {
+            if (interactable == null || interactables.Contains(interactable))
+            {
+                return;
+            }
+            interactables.Add(interactable);
+        }
+
+#if UNITY_EDITOR
+        /// <summary>
+        /// When selected draw lines to all linked interactables
+        /// </summary>
+        protected virtual void OnDrawGizmosSelected()
+        {
+            if (this.interactables.Count > 0)
+            {
+                GameObject[] bioList = this.interactables.ToArray();
+
+                for (int i = 0; i < bioList.Length; i++)
+                {
+                    if (bioList[i] != null)
+                    {
+                        Gizmos.color = Color.green;
+                        Gizmos.DrawLine(this.transform.position, bioList[i].transform.position);
+                    }
+                }
+            }
+
+            if (this.Targets.Count > 0)
+            {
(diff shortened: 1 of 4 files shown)


radicalad:
@killerantz I'd recommend extending this for your "hook up buttons with no code" approach.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,415 @@
...
+                return;
+            }
+            interactables.Add(interactable);
+        }
+
+#if UNITY_EDITOR
+        /// <summary>
+        /// When selected draw lines to all linked interactables
+        /// </summary>
+        protected virtual void OnDrawGizmosSelected()
```
StephenHodgson: `OnDrawGizmosSelected` only gets called in the editor anyway. Do we have editor specific code in this ifdef?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,415 @@
...
+                _selectingFocuser.IsFocusLocked = false;
+                _selectingFocuser = null;
+            }
+        }
+
+        public void OnBeforeFocusChange( FocusEventData eventData ) { /*Unused*/ }
+
+        public void OnFocusChanged( FocusEventData eventData )
+        {
+            FocusEventData newData = new FocusEventData( EventSystem.current );
```
StephenHodgson: This doesn't look good. Event data should only be created by the input system.

radicalad: I suppose we can pass the existing FocusEventData into the virtual methods. The original created a new pointer event to pass along. To be honest, I'm not entirely sure why since the new FocusEvent is effectively a clone of the original.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,410 @@
...
+        }
+        
+        #endregion
+
+        #region Private and Protected Members
+
+        /// <summary>
+        /// Protected focuser for the current selecting focuser
+        /// </summary>
+        protected IMixedRealityPointer _selectingFocuser;
```
StephenHodgson: this could be multiple pointers, correct?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,410 @@
...
+        /// <summary>
+        /// On start subscribe to all interaction events on elements in the interactables list.
+        /// </summary>
+        protected override void OnEnable()
+        {
+            base.OnEnable();
+            //Still need to register since focus isn't required
+            if(InputSystem != null && InputSystem.IsInputEnabled)
+            {
+                InputSystem.Register( gameObject );
```
StephenHodgson: The base input handler already does this kind of stuff. We could make `IsFocusRequired` virtual in the base input handler and override it in this class, then just call the base.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,410 @@
...
+        }
+
+        /// <summary>
+        /// When selected draw lines to all linked interactables
+        /// </summary>
+        protected virtual void OnDrawGizmosSelected()
+        {
+            if (this.interactables.Count > 0)
+            {
+                GameObject[] bioList = this.interactables.ToArray();
```
StephenHodgson: bioList? haha

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,410 @@
...
+            {
+                PositionInputChanged( eventData.selectedObject, eventData );
+            }
+        }
+
+        #endregion
+
+        #region Protected Virtual Callback Functions
+
+        protected virtual void FocusEnter(GameObject obj, FocusEventData eventData ) { }
```
StephenHodgson: Not sure how I feel about making "clone" methods of the handlers. What's the advantage here?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
radicalad: This is what you mentioned yesterday. That sounds much more elegant. I can add it to this PR.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
radicalad: 😉 naming artifact from HTK.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
radicalad: This is how it worked previously, do you see a better approach? The important thing is the interactable list check. If the extending class overrides the original handler method, it will git every event call from `InputSystem` - where we really only want extending classes to get the event if interactable list check passes.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
radicalad: Since `IsFocusedRequired` is a public accessor, can't we just set it to false before we call `base.onEnable()`? Do we need to override the accessor here?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
StephenHodgson: I don't see why we can't just handle everything in the handler implementations. No need to duplicate methods.

We can make all the handler methods virtual and override them just like the ones you've got here.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
StephenHodgson: Take a look at the [`BaseFocusHandler`](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/BaseFocusHandler.cs)

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
StephenHodgson: We could make the `IsFocusedRequired`'s set accessor protected

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
StephenHodgson: So we're removing the check if the event is valid, then passing that on to the derived classes, correct?

radicalad: Currently (and desired): The handler contains the interactable check, then if valid, passes that on to the derived classes via the protected methods.

StephenHodgson: Alright, I think that's okay, but the methods seem redundant (as you could still do the same things with a base call to check the validity). I'm interested in what others think as well.

radicalad:
Talked with @paseb about the original intent for InteractionReceivers, LockFocus/focus tracking really belongs as a separate feature, especially now that `IMixedRealityPointer` supports it better than in HTK. Lets keep this one simple.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -41,10 +41,23 @@ public List<GameObject> Interactables
             set { value = interactables; }
         }
 
+        public override bool IsFocusRequired
```
StephenHodgson: Let's be sure to add the summary statement as well ( I know it's kinda redundant, but it's part of our coding guidelines )

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,349 @@
...
+        protected override void OnEnable()
+        {
+            IsFocusRequired = false;
+            base.OnEnable();
+        }
+
+        /// <summary>
+        /// On disable remove all linked interactables from the delegate functions
+        /// </summary>
+        protected override void OnDisable()
```
keveleigh: Do we need this override that just calls the base method?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,349 @@
...
+        {
+            base.OnDisable();
+        }
+
+        /// <summary>
+        /// Register an interactable with this receiver.
+        /// </summary>
+        /// <param name="interactable">takes a GameObject as the interactable to register.</param>
+        public virtual void Registerinteractable(GameObject interactable)
```
keveleigh: Capitalization for clarity: `Registerinteractable` -> `RegisterInteractable`

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
radicalad: We do not.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
SimonDarksideJ: So can we rename it to something more like what is actually is?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
SimonDarksideJ: Yeah, I'm also not sure about "translating" the events here. I can understand having a "isInteracted" check before doing something but not sure this is the correct path.
Maybe if there was an "IMixedRealityInteractable" interface maybe...  But still not sure.

StephenHodgson: Not really sure we need another interface here. It'd just be a copy of all the handers anyway

cre8ivepark: FYI, added a diagram from HoloToolkit readme. The script which inherits InteractionReceiver can selectively implement event handlers and handle those types of events coming from multiple objects.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,341 @@
...
+        /// List of linked interactable objects to receive events for
+        /// </summary>
+        [SerializeField]
+        [ Tooltip("Target interactable Object to receive events for")]
+        private List<GameObject> interactables = new List<GameObject>();
+
+        public List<GameObject> Interactables
+        {
+            get { return interactables; }
+            set { value = interactables; }
```
StephenHodgson: I'm not sure we want to allow people to set a new list. We can make this a getter only as it doesn't prevent people from modifying the list.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,341 @@
...
+        /// <summary>
+        /// List of linked targets that the receiver affects
+        /// </summary>
+        [Tooltip("Targets for the receiver to affect")]
+        private List<GameObject> targets = new List<GameObject>();
+
+        public List<GameObject> Targets
+        {
+            get { return targets; }
+            set { value = targets; }
```
StephenHodgson: same idea here

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
cre8ivepark: @StephenHodgson Setting this list of Interactables from the inspector by dragging & dropping objects from the scene is the crucial elements of the Receiver. Same for the 'Targets' 
You can find the example in the 'InteractableObjectExample' scene.

2018-09-18 22_13_27-settings


2018-09-18 22_12_27-settings

StephenHodgson: You should still be able to add/remove items from the list as well. Having a get only accessor just means someone can't make a new list and set it.

cre8ivepark: Right, using List.Add() / Remove(). Yes, I can't think about an example of replacing the list with a new one.

radicalad: This is really good to know. I always thought any modification of a list required the setter as well. I’d be happy to change.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,343 @@
...
+
+        #region private variables
+        /// <summary>
+        /// When true, this interaction receiver will draw connections in the editor to Interactables and Targets
+        /// </summary>
+        [Tooltip( "When true, this interaction receiver will draw connections in the editor to Interactables and Targets" )]
+        [SerializeField]
+        private bool drawEditorConnections = true;
+        
+        #endregion
```
StephenHodgson: do we need a region with only one field?

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,343 @@
...
+
+        /// <summary>
+        /// On enable, set the BaseInputHandler's IsFocusRequired to false to receive all events.
+        /// </summary>
+        protected override void OnEnable()
```
StephenHodgson: I'd much rather sort the regions by monobehaviors and interface implementations

radicalad: Fair enough - just trying to be consistent. Do we have region recommendations in the code guidance?

StephenHodgson: Hmm, no but maybe it's something we should add.

radicalad: fixed.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,344 @@
...
+        [Tooltip( "Targets for the receiver to affect" )]
+        private List<GameObject> targets = new List<GameObject>();
+
+        public List<GameObject> Targets
+        {
+            get { return targets; }
+            private set { value = targets; }
+        }
+
+        #endregion
```
StephenHodgson: let's be sure to add the region's name to the end as well.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
StephenHodgson: I'd take a look at any of the pointers to see how we've broken out the regions by interface

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,344 @@
...
+            }
+
+        }
+        public void OnGestureCompleted( InputEventData<float> eventData )
+        {
+            if ( IsInteractable( eventData.selectedObject ) )
+            {
+                GestureCompleted( eventData.selectedObject, eventData );
+            }
+
```
StephenHodgson: extra line

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        public void OnGestureCompleted( InputEventData<Vector2> eventData )
+        {
+            if ( IsInteractable( eventData.selectedObject ) )
+            {
+                GestureCompleted( eventData.selectedObject, eventData );
+            }
+
```
StephenHodgson: extra line

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        public void OnGestureCompleted( InputEventData<Vector3> eventData )
+        {
+            if ( IsInteractable( eventData.selectedObject ) )
+            {
+                GestureCompleted( eventData.selectedObject, eventData );
+            }
+
```
StephenHodgson: extra line

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        public void OnGestureCompleted( InputEventData<Quaternion> eventData )
+        {
+            if ( IsInteractable( eventData.selectedObject ) )
+            {
+                GestureCompleted( eventData.selectedObject, eventData );
+            }
+
```
StephenHodgson: extra line

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,344 @@
...
+        /// <param name="interactable"></param>
+        /// <returns></returns>
+        protected bool IsInteractable( GameObject interactable )
+        {
+            return ( interactables != null && interactables.Contains( interactable ) );
+        }
+
+        #region Global Listener Callbacks
+
+        public void OnBeforeFocusChange( FocusEventData eventData ) { /*Unused*/ }
```
StephenHodgson: I know this is a bit nit picky but let's be sure to keep the spacing consistent as well in our project. 

Please, also add the `/// <Ineritdoc/>` xml tag for any interface implementations.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,362 @@
...
+        /// <param name="interactable"></param>
+        /// <returns></returns>
+        protected bool IsInteractable( GameObject interactable )
+        {
+            return ( interactables != null && interactables.Contains( interactable ) );
+        }
+
+        #region Global Listener Callbacks
+
+        /// <Ineritdoc/>
```
keveleigh: This should be `/// <inheritdoc />` (this happens in several places)

StephenHodgson:
I think one last formatting pass is needed, and I'll check off on this.

Please remove the extra spaces between the parenthesis

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
StephenHodgson: Good catch

StephenHodgson:
After reviewing this more, I'm not sure we need this class to do a lot of the functions I believe we're trying to achieve. 

We can achieve the same effect by using the `MixedRealityInputActions`

radicalad:
Is there an implementation example of `MixedRealityInputActions`?

StephenHodgson:
Yes, one tick

StephenHodgson:
https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2811#issuecomment-423000107

StephenHodgson:
Although, I do see the need and desire to centralize all the input into one place, instead on a bunch of different objects.

I think it's fine to keep for now.

Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Receivers/InteractionReceiver.cs
```diff
@@ -0,0 +1,358 @@
...
+        #endregion Global Listener Callbacks
+
+        #region Protected Virtual Callback Functions
+
+        protected virtual void FocusEnter(GameObject obj, FocusEventData eventData) { }
+        protected virtual void FocusExit(GameObject obj, FocusEventData eventData) { }
+
+        protected virtual void InputDown(GameObject obj, InputEventData eventData) { }
+        protected virtual void InputUp(GameObject obj, InputEventData eventData) { }
+        protected virtual void InputClicked(GameObject obj, InputEventData eventData) { }
```
StephenHodgson: unused?

radicalad:
Thanks @StephenHodgson !

StephenHodgson:
No, thank you!

Awesome work. I'm excited to see what you and @cre8ivepark have in store for us.

Issue #2799, opened 2018-09-17
vNext Task: Improve and better implement Controller Model Visualization


StephenHodgson:
# Overview
In reviewing the migration of the Solver framework, it has highlighted that controller visualization needs to be better supported.

We also need a way to easily get the controller visualization's `GameObject` or `Transform` references for other components to use.

Currently the controller models are assigned in the Controller Mapping Profile, I'd like to propose that the model assignments should be their own profile, separate from input mapping.

# Requirements

* Controller visualization needs to have it's own profile in the Input System's configuration.
* Controller visualization needs to have it's own interface to define the requirements of a visualized controller. `IMixedRealityControllerVisualizer` ?
* Controller visualization profile needs to be able to assign a concrete implementation that is automatically added to the prefab at runtime, if it was not already added.
* Controller visualization profile needs to have a global prefab for left/right hands
* Controller visualization profile needs to have a prefab option for each supported platform's controller.
* Controller visualization's `GameObject` or `Transform` references need to be easily accessible.

# Acceptance Criteria

- [x] I can create and assign a Controller Visualization Profile to the Input System Configuration.
- [x] I can configure a profile to use a specific concrete implementation of a controller visualizer using the `IMixedRealityControllerVisualizer`.
- [x] I can configure a profile to use a prefab with no additional scripts (auto assigned from concrete implementation).
- [x] I can configure a profile to use a prefab with a previously attached script that implements `IMixedRealityControllerVisualizer`.
- [x] I can get the Controller `GameObject` reference from the `IMixedRealityControllerVisualizer`
- [x] I can get a list of currently detected controllers from the `IMixedRealityInputSystem`
- [ ] I can configure a profile to use the default controller model loaded form glTF or other sources to use the concrete implementation defined in the profile.

SimonDarksideJ:
I would argue, that if we have a Visualization Manager, then it could manage the list of "connected" visualized objects / controllers. Negating the need for the SceneObjects registry.

If you can query the visualization config through the MR Manager, then it would server the same purpose.

SimonDarksideJ:
Note, this requirement

> Controller visualization profile needs to have a global prefab for left/right hands

Is already provided by the Controller configuration, as both a global and a per controller model.

However, the new logic needs to extract the current model lookup currently sitting in the BaseController class.

StephenHodgson:
glTF support to come later after 2807

Issue #2785, opened 2018-09-13
Solvers need to track controller transforms


keveleigh:
# Overview
The solvers in vNext are currently head-only. The also need to support tracking motion controllers.

> Review (SJ)
> Solvers are components added to GameObjects in a scene. These Solvers then need a "target" to bind to which may be the users head or could be a physical controller

> Question
> Are ONLY controller entities (head / controller / hand / finger) bound to? Or could it in theory be bound to ANY other GameObject?

# Requirements

* Solvers can have a Transform reference to the tracked controller to "solve" against.
  * This means there must be a way to ask something for the currently tracked controllers ([which exists](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityDeviceManager.cs#L16), but shouldn't be used?)
  * This does not mean listening for source pose events, as that breaks the paradigm of solver independence and adds yet another set of global listeners.

# Acceptance Criteria
A list of the specific things that must be done for approval / acceptance of the task in checklist form. 
This may correspond with the above requirements.

- [x] Solvers can be configured to obtain and solve towards motion controller transforms.
- [x] Doesn't add another global listener to the scene.

keveleigh:
Proposed solution:

MRControllers add a Transform reference to their own transform, so a script can access it either in SourceDetected or by asking the controller manager for the specifically handed controller.

keveleigh:
Additionally, asking the controller manger for its controllers was made more difficult as more "device managers" came online. Previously, the OpenVRDeviceManager or the WindowsMixedRealityDeviceManager would be the only active "device managers" in the scene. Now, there's a whole list of them, but only one of them will have tracked motion controllers.

StephenHodgson:
The solvers just need to handle events for the controller they're assigned to. 

Look at the pointer implementation for details.

You should be able to fulfill the requirements with `SourcePoseSynchonizer`

keveleigh:
Solvers have their own update cycle and track towards a specific transform that gets set once. This is important, because a benefit of solvers is that they can interact with each other in a known moment, state, and order.

Additionally, once attachment points are added (again, as passive transforms that get set once), the pose synchronizer is incompatible.

StephenHodgson:
[...]

PR #2807, opened 2018-09-19
Implementation of Controller Model Visualization Task

StephenHodgson:
Overview
---
I've broken out the visualization of the controllers into their own profile, to give more fine grained control over the controller models. Part of the reason for this change was the fact that the input mapping profile didn't allow us to expose override models and other settings because of the pop out windows. 

image

I've added a `IMixedRealityControllerVisualizer` that supports getting the controller's `GameObejct` reference without breaking the architecture design patterns we've established. (You'll notice that I updated the [`BaseController`](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2807/files#diff-cfedb4662811658603ef2dfdf35c85a6R210) class to properly reference this interface, and not the transform or game object reference).

I've enhanced the controller visualization profile's model object fields to [ensure we're assigning valid prefabs only](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2807/files#diff-226a65267f295f1911d642c8ac1aa0d6R209) (and not just mesh assets) as well as adding the user defined concrete type that implements `IMixedRealityControllerVisualizer`.

Here's a little snippet example of how to get the controller visuals from the input system:
(You'll notice that we can get any number of detected controllers)

```C#

public class MyClass : MonoBehaviour, IMixedRealitySourceStateHandler
{
    private List<Transform> leftHandControllerTransforms = new List<Transform>();
    private List<Transform> rightHandControllerTransforms = new List<Transform>();

    private void Start()
    {
        foreach (var visualizer in MixedRealityManager.VisualizationSystem.DetectedVisualizers)
        {
            // Skip the controllers that don't have visuals
            if (visualizer == null) { continue; }

            if (visualizer.Controller is WindowsMixedRealityController)
            {
                // Do something for a specific controller type
            }

            switch (visualizer.Handedness)
            {
                case Handedness.Left:
                    leftHandControllerTransforms.Add(visualizer.VisualizerGameObjectReference.transform);
                    break;
                case Handedness.Right:
                    rightHandControllerTransforms.Add(visualizer.VisualizerGameObjectReference.transform);
                    break;
            }
        }
    }

    {
        if (eventData.Controller?.Visualization != null)
        {
            switch (eventData.Controller.ControllerHandedness)
            {
                case Handedness.Left:
                    leftHandControllerTransforms.Add(eventData.Controller.Visualization.VisualizerGameObjectReference.transform);
                    break;
                case Handedness.Right:
                    rightHandControllerTransforms.Add(eventData.Controller.Visualization.VisualizerGameObjectReference.transform);
                    break;
            }
        }
    }

    {
        if (eventData.Controller?.Visualization != null)
        {
            switch (eventData.Controller.ControllerHandedness)
            {
                case Handedness.Left:
                    leftHandControllerTransforms.Remove(eventData.Controller.Visualization.VisualizerGameObjectReference.transform);
                    break;
                case Handedness.Right:
                    rightHandControllerTransforms.Remove(eventData.Controller.Visualization.VisualizerGameObjectReference.transform);
                    break;
            }
        }
    }
}
```

Changes
---
- Fixes: #2799 
- Fixes: #2785 By giving the ability to get a controller's visualization and in extension it's transform.

TODO
---
- [ ] [Test type passed correctly](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2807/files#diff-0dfbf575be4a6bde52ea31ae7718e31cR37)

Breaking Changes
---
- Removed unused id's from interaction mapping (we usually use the array item's position in the array as the id)
- Broke out controller visualization options in Controller Mapping profile into its own profile


StephenHodgson:
@SimonDarksideJ Can you provide an example of how we would get the controllers with your CR on my PR?

SimonDarksideJ:
As it's now a manager, the call would look like any other manager call, e.g.
```
IMixedRealityVisualizationSystem visualizer = MixedRealityManager.Instance.GetManager<IMixedRealityVisualizationSystem>();
            foreach (MixedRealityVisualizer controller in Visualizer.DetectedVisualizers)
            {
                if (controller.Handedness == hand)
                {
                    return controller;
                }
            }
```
Obviously like any other system provided by a Manager (including the Input System) you'd want to cache the reference and in the case of the Solver handler, make it a Static reference so it's the same across all handlers.

StephenHodgson:
Still not sure about how I feel about a viz manager, but I'd like to hear @keveleigh and @Ecnassianer's input as well. What do you guys think?

SimonDarksideJ:
Well I also disagree with putting the visualization elements within the InputSystem (mostly for the same reasons it shouldn't be in the Device layer) as we don't want to introduce that dependency.

It could almost work as a Static utility, except we want people to be able to write their own without having to change the MRTK itself.

radicalad:
I'd expect to get it from the event "SourceDetected." Accessing controllers models via a manager sounds overly complicated for what it is (up to two handSource GameObjects).

StephenHodgson:
@radicalad That's an awesome idea!

StephenHodgson:
Although I do think we should still be able to get access to the visualized models through a manager

keveleigh:
>I'd expect to get it from the event "SourceDetected."

Yep, that's exactly why I implemented it the way I did in [my original PR](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2784/files#diff-59a2991baea506ac4fdeab262189e84fR67), so the controller transform/model was passed along with the event.

StephenHodgson:
Let me see what I can do. 

The original (before the viz manager) could have passed it along with the detected events.

But now I need to dissect @SimonDarksideJ's changes a bit to see how I can still accommodate something like that.

radicalad:
Could they be accessed directly by `MixedRealityManager`? The controller visualizations feel much closer to `CameraCache` as far as accessibility. Having a hand visualization manager doesn't seem very discoverable.

StephenHodgson:
The `CameraCache` is a static utility for something that will always be in the scene (typically) and improves functionality of existing unity engine functions.

radicalad:
Oh I know, I just mean its relatively easy to find. (aside from HoloLens) getting the controller visualizations is a task "most devs will do all the time". I'm not sure if you're familiar with that 2x2 matrix concept.

StephenHodgson:
> 2x2 matrix concept

I'm not actually. Wanna PM me on slack and tell me about it?

StephenHodgson:
@SimonDarksideJ Sorry, but I really think we should add the `IMixedRealityVisualizer` to the `IMixedRealityController` so that devs can easily access it in `OnSourceDetected` like @keveleigh wants.

SimonDarksideJ:
Sorry to still be a stick in the mud, but we simply add the visualizer to the controller, still for the same reasons:

* You are creating a circular reverence from Controller to Visualizer to Controller, etc.
* Can't remove the controller reference from the visualizer as you need to know which controller it came from.

If the Visualizer is a thing that is needed in the SourceDetected event, then simply add the visualizer to the EVENT not the controller (using the interface).

```
    {
        if (eventData.Visualizer != null)
        {
            switch (eventData.Visualizer.Handedness)
            {
                case Handedness.Left:
                    leftHandControllerTransforms.Add(eventData.Visualizer.VisualizerGameObjectReference.transform);
                    break;
                case Handedness.Right:
                    rightHandControllerTransforms.Add(eventData.Visualizer.VisualizerGameObjectReference.transform);
                    break;
            }
        }
    }
```

## Question
If we do move the reference to the visualizer, do we even need the "DetectedVisualizers" hashset any more?  Would GameObjects being able to discover the visualizer from the Event be enough to meet @davidkline-ms requirement of discovering visualizers?

SimonDarksideJ:
on another point, we should probably close all of the final SolverHanlder PR's until this PR is solved (no pun intended) as the implementation will have to change to meet the output of this PR anyway

StephenHodgson:
We pretty much do the same thing with pointers. I know the circular reference isn't ideal, but I think it solves the problem without a complex solution.

My favorite circular reference in unity is: 
`gameObject.transform.gameObect.transform...`
🤣

StephenHodgson:
And yes, we'd still need the hashset to query against in the event a source is detected before our listener was subscribed.

StephenHodgson:
If we were to add the visualizer to the source detected event, the list of visualizers would need to move into the input system. (Which I think is okay, seeing as the profile and viz manager both depend on the input system)

SimonDarksideJ:
No need to move the list of detected Visualisers to the Input System, simply passing the reference down in the event is enough.  
Still don't want a hard dependency with other components and the input system. Else those who do want to (for what ever reason) will get a lot more than they bargained for and then not bother.

StephenHodgson:
> Still don't want a hard dependency with other components and the input system.

You can't know the position of a controller without the input system, and without the input system the Controller's rendered model isn't going to be where it needs to be. Unless you made the visualization part of the input system or the controller itself (but we already agreed the transform position shouldn't be part of the device layer so we can port the system to other engines).

StephenHodgson:
In either case, my main goal is to have this function pretty much exactly like pointers, where we set a prefab and have it get created as needed.

Ecnassianer:
We spoke about this in another thread which I can't find at the moment, but can we change the name from "Visualizer" to something else. This PR seems like the place to make that change. Here are some things that I think would more clearly reveal the purpose of these classes:

SceneRepresentation (i.e, ControllerSceneRepresentation or ControllerRepresentation)
DeviceAvatar (i.e, ControllerDeviceAvatar)
SceneObject
TrackingProxy
TrackedProxy
MixedRealityProxy
RealityProxy

Let me know if you want a recap of why I think Visualizer is a misleading name.

StephenHodgson:
How is it misleading?

Ecnassianer:
"Visualizer" seems to imply that the focus of the class is in rendering a visual representation of the controller. It's about pushing pixels to the screen. But what I understand now is the "visualizer" class is all the data that the device puts into the scene, not just the visuals.

Consider approaching this problem from a naive perspective: I don't want to show mechanical controllers in my fantasy game. Instead I want to use information about the controller state (button states, location, etc) to draw some kind of particle fairies around where the controllers would be. Given that prompt, a "Controller Visualizer" isn't nearly the first thing I would look for, since I explicitly don't want to visualize controllers, I just want to get scene data about the controller.

The more we build out the "Visualization System", the lower percentage of it actually has to do with visualizing controllers. We're already seeing this class grow in a non-visualization direction, we should anticipate that further growth with a more generalized name.

StephenHodgson:
I'm cool with "proxy".

StephenHodgson:
@SimonDarksideJ I wasn't able to easily overload the on source detected like we wanted to do and pass the visualizer (proxy?) controller through with it.

Any other ideas?

At the moment it's pretty much identical to how pointers are setup and used (which was the goal, seeing as they serve virtually the same function, sans the controller proxy itself should be unique to the controller, but there could be multiple pointers).

I'm not so keen on having our own Visualization System as it's completely driven by the input system, and our goal was to have systems not be depend on each other. All that manager does is just keep a list of the device visualizers and I'm not sure why the Input System can't just facilitate that.

If we wanted to make it its own full blown system, we should also add event handlers when visualization is added/removed, etc

davidkline-ms:
> (but we already agreed the transform position shouldn't be part of the device layer so we can port the system to other engines).

Except that other engines are going to have different requirements... I wouldn't worry about porting at this point.

StephenHodgson:
> I wouldn't worry about porting at this point.

haha isn't that the truth! 😆

StephenHodgson:
Okay I can't do a hard reset on this, so I'll need to open a new PR. I'll be sure to link this one.

SimonDarksideJ:
This statement is not really valid
> I'm not so keen on having our own Visualization System as it's completely driven by the input system, and our goal was to have systems not be depend on each other. All that manager does is just keep a list of the device visualizers and I'm not sure why the Input System can't just facilitate that.

This was only the start of the (controller) visualization system, who responsibility was to maintain and control the visualized elements in the scene.

Moving yet another function in to the Input system is a mistake imho, as it breaks the rule of only having individual component systems responsible for a single area / function.  Now in order to replace the input system (granted not something we advise people to do, they should contribute to this one), they are going to have to rewrite:

- Input routing
- Input source handling
- Focus
- Gaze
- Pointers
- Pointer visualization (as this is a scene task)
and now Controller visualization

A single system should not be this big and have responsibility over so many features, which was one of the fundamental issues with the old HTK that we were trying to break free from.  
This PR provided the right level of abstraction and only needed to address a final requirement of making the data/visualizer reference available in the source detected event.

Really not happy with having an input system also being responsible for scene management of features (pointer visualization was a very fine line as it already generated pointers) doing both the job of routing and now UX, but that seems to be the consensus for the approach.

StephenHodgson:
😅🤔 I know you are correct.

But I'm not sure how to facilitate the visual parts with being driven by input itself. Seems like the visualizer system would be too dependant on the input system, which was another requirement, as there shouldn't be dependencies on other systems.

I'm trying to figure out a middle ground where we can all agree so we can move forward.

PR #2809, opened 2018-09-19
Add UI for diagnostics

ryzngard:
Overview
---

Add the UI for diagnostics. This is a simple UI to start. We should evaluate https://github.com/Microsoft/VisualProfiler-Unity as a UI option since there's been a lot of good work there, but this current iteration will do well for a baseline.


ryzngard:
image

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
```diff
@@ -3,22 +3,122 @@
 
 using Microsoft.MixedReality.Toolkit.Core.EventDatum.Boundary;
 using Microsoft.MixedReality.Toolkit.Core.Interfaces.Diagnostics;
+using System;
+using System.Text;
 using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.SDK.DiagnosticsSystem
 {
     public class DiagnosticBehavior : MonoBehaviour, IMixedRealityDiagnosticsHandler
```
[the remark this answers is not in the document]
SimonDarksideJ: Nice catch @davidkline-ms :D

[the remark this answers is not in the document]
ryzngard: Woops. Behavor sounds like a dinosaur 😂

ryzngard:
@SimonDarksideJ I think the idea that they are stateful means they should be classes instead of public static functions. They should all be internal though, so I can definitely fix that.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticBehavor.cs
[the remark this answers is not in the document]
StephenHodgson: I've always felt that the "Behaviour" suffix seemed redundant.

`DiagnosticsHandler` seems much more focused and describes its function.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/CpuUseTracker.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Diagnostics;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.DiagnosticsSystem
+{
+    internal class CpuUseTracker
```
StephenHodgson: Do we want to make this internal, or just sealed?

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        void UpdateIsShowingInformation()
+        {
+            isShowingInformation = ShowCpu ||
+                                   ShowFps ||
+                                   ShowMemory;
+        }
+        
+        void Update()
```
StephenHodgson: missing private accessors per coding guidelines

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            UpdateIsShowingInformation();
+
+            if (!isShowingInformation)
+            {
+                return;
+            }
+
+            StringBuilder sb = new StringBuilder();
```
StephenHodgson: We're making a new string builder on each update?

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        void OnGUI()
+        {
+            if (!isShowingInformation || displayText == null)
+            {
+                return;
+            }
+
+            var style = new GUIStyle();
```
StephenHodgson: let's get rid of all these new keywords in our update loops.

We can cache and update a lot of this stuff.

StephenHodgson: we can create it, then check if it's not null after.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticsHandler.cs
StephenHodgson: Just not sure we should have any allocations in update loops.

I think keeping the same string builder around and just clearing it in each update loop is a bit better. Idk, I suppose we'd have test the comparison.

ryzngard: Ah, I see. I haven't looked at StringBuilder in a while, they added Clear in 4.0 that should do the right thing. Before, setting length would sort of do it... but it still might allocate a new string, so it wasn't guaranteed. I'll Move to a member and call Clear()

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticsHandler.cs
ryzngard: Just did a quick and dirty test to see how much stack allocation actually costs, and newing a struct was twice as expensive with int.MaxValue iterations (both were still really quick though). I'll move this to a field and remove the new here. Good catch

StephenHodgson: We also need to be aware that imgui is depreciated and it's most likely getting taken out of the core Unity Engine soon.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/CpuUseTracker.cs
davidkline-ms: sealed indicates that code (including the toolkit) cannot extend the class... I tend to fall on the side of internal.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticsHandler.cs
ryzngard: I don't see where it's marked deprecated, but I believe we should open another issue on that if it is. I was looking at https://docs.unity3d.com/2018.3/Documentation/Manual/GUIScriptingGuide.html for information.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/DiagnosticsHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+using System;
+using System.Text;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.DiagnosticsSystem
+{
+    public class DiagnosticsHandler : MonoBehaviour, IMixedRealityDiagnosticsHandler
+    {
+        private bool showCpu;
+        private bool ShowCpu
```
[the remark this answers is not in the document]
ryzngard: Sure. Coming in next iteration

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MemoryUseTracker.cs
```diff
@@ -0,0 +1,59 @@
...
+            };
+
+            memoryReadings[index] = reading;
+            index = (index + 1) % memoryReadings.Length;
+
+            var sum = memoryReadings.Aggregate(new MemoryReading(), (a, b) => a + b);
+            return sum / memoryReadings.Length;
+        }
+
+        public struct MemoryReading
```
StephenHodgson: internal?

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MemoryUseTracker.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    internal class MemoryUseTracker
+    {
+        private Process currentProcess = Process.GetCurrentProcess();
+        private MemoryReading[] memoryReadings = new MemoryReading[10];
+        private int index = 0;
+
+        public MemoryReading GetReading()
+        {
+            var reading = new MemoryReading()
```
StephenHodgson: Instead of making a new one each time, and we know we always have 10 in our array, why not just reuse the ones we have if we've already filled the array?

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MemoryUseTracker.cs
ryzngard: Doesn't need to be because it's encapsulated by the MemoryUseTrackerClass

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/MemoryUseTracker.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                VirtualMemoryInBytes = currentProcess.VirtualMemorySize64,
+                WorkingSetMemoryInBytes = currentProcess.WorkingSet64,
+                GCMemoryInBytes = GC.GetTotalMemory(false)
+            };
+
+            memoryReadings[index] = reading;
+            index = (index + 1) % memoryReadings.Length;
+
+            var sum = memoryReadings.Aggregate(new MemoryReading(), (a, b) => a + b);
```
StephenHodgson: our sum MemoryReading could also be reused each time.

PR #2810, opened 2018-09-19
Feature/mrtk sharing

mr0ng:
Overview
---
Implemented rough architecture for INetworkHandler and NetworkEventData

Updated Ong Sharing code to work with vNEXT - resolved all console errors


mr0ng:
Photon (included in this PR) is bundled with newtonsoft, hence why it was included. @StephenHodgson how do you recommend we proceed with a third party package like photon?

StephenHodgson:
>how do you recommend we proceed with a third party package like photon?

I think that's a question for when we start adding in the package manager.

mr0ng:
Closing this pull request, and will attempt again starting with only the architecture docs

PR #2814, opened 2018-09-20
Add initial documentation around package manager

ryzngard:
Overview
---

Start documentation for Package Management. This should be used as a basis for conversation in determining policy more than implementation details. I want to make sure the community has the right discussion before I start adding any interfaces, classes, or in depth implementation on how to hook in with Unity.

### DIFF
--- a/Documentation/Architecture/Package_Manager/Package_Manager.md
+++ b/Documentation/Architecture/Package_Manager/Package_Manager.md
@@ -8,11 +8,11 @@ Package vetting for showing up in the package manager UI should be done in multi
 
 The main goal is that all packages are contained within the Mixed Reality Toolkit Repo, or some separate repo mantained by the same community. This brings clarity to users about where issues can be filed, where to find more information about specific packages, and how they can get their packages added. This will explicitly not display any packages that are not contained within these constraints; while they may be good packages that work well with Mixed Reality Toolkit, they are not mantained within the community and will not be promoted in the same way. Instead, it will be up to the package owners to promote it's use in another way. We may find other promotion techniques if we as a community believe a package provides value but is not part of our ecosystem. 
 
-### Microsoft Packages
+### Release Packages
 
-Microsoft packages are promoted and mantained by Microsoft, with some promise of stability and support (TBD). These will be packages like Azure integration that we believe work well with Mixed Reality Toolkit and explicitly provide features that Microsoft is promoting. These packages will also have a strict process for submitting code changes.
+Release packages are promoted and mantained with some promise of stability and support (TBD). These will be packages like Azure integration that we believe work well with Mixed Reality Toolkit and explicitly provide features that Microsoft is promoting. These packages will also have a strict process for submitting code changes.
 
-### Community Packages
+### Preview Packages
 
 These packages contain less promise to quality and maintainance, but still exist within the ecosystem. The process for code changes will be much less stringent. However, we still want to celebrate community contributions and will surface packages like this in the package manager UI. 
 
--- a/Documentation/Architecture/Package_Manager/Package_Manager.md
+++ b/Documentation/Architecture/Package_Manager/Package_Manager.md
@@ -1,6 +1,6 @@
 # Package Manager
 
-The package manager is intended to highlight what is available from both Microsoft and the community for Mixed Reality Toolkit. It will suplement Unity's built in package management that is available by highlighting [vetted](#package-vetting) resources for developers.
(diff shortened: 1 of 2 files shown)


Documentation/Architecture/Package_Manager/Package_Manager.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Package Vetting
+
+Package vetting for showing up in the package manager UI should be done in multiple ways. Naming is subject to change, but the below outlines the general split on how packages will be presented. 
+
+The main goal is that all packages are contained within the Mixed Reality Toolkit Repo, or some separate repo mantained by the same community. This brings clarity to users about where issues can be filed, where to find more information about specific packages, and how they can get their packages added. This will explicitly not display any packages that are not contained within these constraints; while they may be good packages that work well with Mixed Reality Toolkit, they are not mantained within the community and will not be promoted in the same way. Instead, it will be up to the package owners to promote it's use in another way. We may find other promotion techniques if we as a community believe a package provides value but is not part of our ecosystem. 
+
+### Microsoft Packages
```
chbecker-ms: Input and spatial understanding I would promote as coming with a promise of stability and support, which includes from Microsoft and community contributions.

Similarly, Microsoft will contribute some packages which may be more experimental in nature and would not come with long term support. 

Since our grouping of packages together is the contract that MRTK has with an app developer, I think we would name them differently than which organization authored them.

Some ideas:
Production, Supported, Core, Retail, Release
vs
Experimental, Beta, Explorational, Prototype, Community (may still work here)

ryzngard:
I'm happy to discuss my initial investigations, but I think the final decision on that should be another revision. This initial proposal is to make sure we understand the intent behind the package manager and expectations of it. 

I believe we can leverage Unity's native package management solution, but I'm still looking into the requirements on publishing packages and what all we can automate.

Documentation/Architecture/Package_Manager/Package_Manager.md
chbecker-ms: How about "Release" and "Preview"?

Here's why: "Core" can also refer to the implementations of the base level components for MRTK. Input and Spatial Understanding are part of "MRTK Core".

"Community" I feel is overloaded to be both authorship and support level.

"Release" are components/platforms that we will have a strong contract with an app developer stating we'll fix bugs, respond to issues, service requests.

"Preview" are components/platforms that we offer to developers but are "at your own risk" to take a dependency on. This category has flexibility to attract new ideas and find signal. With enough feedback we can promote "Preview" work into the next "Release". But components can also remain "Preview" indefinitely.

ryzngard: I think that's a good start. I believe @davidkline-ms  suggest "Standard" and "Extension", but thinking about I like "Release" and "Preview" a bit more.

Documentation/Architecture/Package_Manager/Package_Manager.md
```diff
@@ -8,11 +8,11 @@ Package vetting for showing up in the package manager UI should be done in multi
...
 
-### Microsoft Packages
+### Release Packages
 
-Microsoft packages are promoted and mantained by Microsoft, with some promise of stability and support (TBD). These will be packages like Azure integration that we believe work well with Mixed Reality Toolkit and explicitly provide features that Microsoft is promoting. These packages will also have a strict process for submitting code changes.
+Release packages are promoted and mantained with some promise of stability and support (TBD). These will be packages like Azure integration that we believe work well with Mixed Reality Toolkit and explicitly provide features that Microsoft is promoting. These packages will also have a strict process for submitting code changes.
 
-### Community Packages
+### Preview Packages
```
davidkline-ms: preview doesn't feel quite right here. these are more like unity asset store packages rather than preview features. most (?) packages in this category are likely not going to be appropriate to move to the "Release" category

davidkline-ms: we should also create a category for experimental features.

ryzngard: What's the difference between experimental and (whatever community/Preview changes to)?

davidkline-ms: experimental is for features that are being test flighted that may never ship and have no promise to be functional

ryzngard: I think we've landed on nuget being our solution, so lets align with common terminology from there. "Prerelease" should be as you described experimental above.

PR #2815, opened 2018-09-21
Fixed jump in SolverBodyLock rotation tether

cdsenselab:
Overview
---
Fixed jump in SolverBodyLock rotation tether by keeping the value of "desiredRot" between updates. Before, desiredRot was reset to Quaternion.identity in every updated which caused a bigger lock area around the zero rotation, and irregular jumps when the initial angle limit was exceeded.

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/HoloToolkit/Utilities/Scripts/Solvers/SolverBodyLock.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/Solvers/SolverBodyLock.cs
@@ -37,9 +37,10 @@ namespace HoloToolkit.Unity
         public int TetherAngleSteps = 6;
         #endregion
 
+        Quaternion desiredRot = Quaternion.identity;
+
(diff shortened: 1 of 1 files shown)


Assets/HoloToolkit/Utilities/Scripts/Solvers/SolverBodyLock.cs
```diff
@@ -37,9 +37,10 @@ public enum OrientationReference
         public int TetherAngleSteps = 6;
         #endregion
 
+        Quaternion desiredRot = Quaternion.identity;
```
radicalad: add `private`

cdsenselab: Done

Ecnassianer:
I am 10 hours late anyway. XD

PR #2821, opened 2018-09-21
Ensured Input System is valid before setting cursor on Gaze Provider.

StephenHodgson:
Overview
---
Added await to start to make sure the input system is initialized before setting cursor.

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
@@ -269,9 +269,11 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
             }
         }
 
-        protected override void Start()
+        protected override async void Start()
         {
             base.Start();
+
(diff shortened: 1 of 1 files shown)


StephenHodgson:
@wassx

wassx:
Thanks, very neat solution!

StephenHodgson:
@wassx, could you review and approve this, if you agree with the change?

Did it fix your issue?

wassx:
Tested and it works! Thanks

wassx:
Sadly problem still occurs in the GazeTest sample

wassx:
I'm using unity 2018.1.9f2. I retested the Basic GazeTest example from the original issue and there is still the error message (after stopping) and no gaze cursor.

StephenHodgson:
Even tho it may not fix his problem we should still make this change.

StephenHodgson:
Hi @wassx. Sorry I've been so busy lately. I left your issue open and plan to revisit and solve it as soon as I can.

wassx:
@StephenHodgson Thank you! 👍

Issue #2832, opened 2018-09-25
Pointer does not always spawn cursor 


StephenHodgson:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
Pointer doesn't always spawn its cursor prefab.

BaseCursor's start checks if the input system is null, then awaits, but if the input system isn't null it doesn't properly create the prefab.

![image](https://user-images.githubusercontent.com/13334553/45994268-b2379c00-c047-11e8-80c3-649a461cd752.png)

Issue #2756, opened 2018-09-11
vNEXT: Spatial Mouse Add Notification to regain control


StephenHodgson:
# Overview
Add an editor notification to regain mouse control by pressing "Esc"

# Requirements
A list of the required functionality this task requires to implement.

* When user enters play mode a notification shows in the Game window describing how to get mouse control back by pressing "Esc"
* When user presses Esc mouse control is regained.

# Acceptance Criteria
A list of the specific things that must be done for approval / acceptance of the task in checklist form. 
This may correspond with the above requirements.

- [x] User can press play mode and notification shows in Game window.
- [x] User can press "Esc" to regain mouse control.

PR #2833, opened 2018-09-25
Mouse Pointer Updates

StephenHodgson:
Overview
---
- Added notification when mouse device manager is started about how to regain mouse control.
- Cached the ray in `BasePointer`
image

Changes
---
- Fixes: #2832 
- FIxes: #2756

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
@@ -147,11 +147,8 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
                 TeleportSystem.Register(gameObject);
             }
 
-            if (InputSystem == null)
-            {
-                await WaitUntilInputSystemValid;
-                SetCursor();
-            }
+            await WaitUntilInputSystemValid;
+            SetCursor();
         }
 
         protected override void OnDisable()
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
```diff
@@ -147,11 +147,8 @@ protected override async void Start()
                 TeleportSystem.Register(gameObject);
             }
 
-            if (InputSystem == null)
```
SimonDarksideJ: Not sure about removing this null check, seems quite valid

StephenHodgson: We don't need it bc we wait until the input system is valid in the line below on 

`await WaitUntilInputSystemValid;`

SimonDarksideJ: Unless the InputSystem is off and I add the BaseControllerPointer to an object manually (or other component consuming it).  Which means it will sit forever in this loop.

StephenHodgson: > Which means it will sit forever in this loop.

It's not a loop, as it's just a continuation. No overhead there.

Besides why would you disable the input system then put something that depends on the input system directly in the scene yourself unless you're asking for trouble.

SimonDarksideJ:
Conflicts @StephenHodgson , do you want to resolve so this can get approved again and merged?

StephenHodgson:
Bump @SimonDarksideJ

PR #2837, opened 2018-09-26
Controller proxy visualization profile

StephenHodgson:
Overview
---
Moved from #2807

I've broken out the visualization of the controllers into their own profile, to give more fine grained control over the controller models. Part of the reason for this change was the fact that the input mapping profile didn't allow us to expose override models and other settings because of the pop out windows. 

image

I've added a `IMixedRealityControllerVisualizer` that supports getting the controller's `GameObejct` reference without breaking the architecture design patterns we've established. (You'll notice that I updated the [`BaseController`](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2807/files#diff-cfedb4662811658603ef2dfdf35c85a6R210) class to properly reference this interface, and not the transform or game object reference).

I've enhanced the controller visualization profile's model object fields to [ensure we're assigning valid prefabs only](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2807/files#diff-226a65267f295f1911d642c8ac1aa0d6R209) (and not just mesh assets) as well as adding the user defined concrete type that implements `IMixedRealityControllerVisualizer`.

Here's a little snippet example of how to get the controller visuals from the input system:
(You'll notice that we can get any number of detected controllers)

```C#
public class MyClass : BaseInputHandler, IMixedRealitySourceStateHandler
{
    private List<Transform> leftHandControllerTransforms = new List<Transform>();
    private List<Transform> rightHandControllerTransforms = new List<Transform>();

    protected override void Start()
    {
        base.Start();

        foreach (var controller in InputSystem.DetectedControllers)
        {
            // Skip the controllers that don't have visuals
            if (controller?.Visualizer == null) { continue; }

            if (controller is WindowsMixedRealityController)
            {
                // Do something for a specific controller type
            }

            switch (controller.ControllerHandedness)
            {
                case Handedness.Left:
                    leftHandControllerTransforms.Add(controller.Visualizer.GameObjectProxy.transform);
                    break;
                case Handedness.Right:
                    rightHandControllerTransforms.Add(controller.Visualizer.GameObjectProxy.transform);
                    break;
            }
        }
    }

    {
        if (eventData.Controller?.Visualizer != null)
        {
            switch (eventData.Controller.ControllerHandedness)
            {
                case Handedness.Left:
                    leftHandControllerTransforms.Add(eventData.Controller.Visualizer.GameObjectProxy.transform);
                    break;
                case Handedness.Right:
                    rightHandControllerTransforms.Add(eventData.Controller.Visualizer.GameObjectProxy.transform);
                    break;
            }
        }
    }

    {
        if (eventData.Controller?.Visualizer != null)
        {
            switch (eventData.Controller.ControllerHandedness)
            {
                case Handedness.Left:
                    leftHandControllerTransforms.Remove(eventData.Controller.Visualizer.GameObjectProxy.transform);
                    break;
                case Handedness.Right:
                    rightHandControllerTransforms.Remove(eventData.Controller.Visualizer.GameObjectProxy.transform);
                    break;
            }
        }
    }
}
```

Changes
---
- Fixes: #2799 
- Fixes: #2785 By giving the ability to get a controller's visualization and in extension it's transform.

Breaking Changes
---
- Removed unused id's from interaction mapping (we usually use the array item's position in the array as the id)
- Broke out controller visualization options in Controller Mapping profile into its own profile


Ecnassianer:
@SimonDarksideJ said this on the old thread:
>WHat you have described is the Controller pose Synchronizer. WHich consumes the Input Pose event data to feed information from the controller state.

Elsewhere I thought we came to the conclusion that Solver Handlers and other UX things would get the Visualizer/Proxy and use the Visualizer/Proxy's transform to calculate their GoalPosition. In my mind, that's not a task that has to do with visualization of a controller. Is that no longer the intention?

If it is the intention for things like Solvers to use the Visualizer/Proxy transform, then when should a system prefer the Pose Synchronizer over the Visualizer? Only when you want to respond to every pose update, rather than just once during some other function like Update()?

StephenHodgson:
>then when should a system prefer the Pose Synchronizer over the Visualizer?

... They're the same thing 😅 the visualizer interface inherits directly from the pose synchronizer.

keveleigh:
He said "Visualizer/Proxy transform", talking about asking the visualization system for the transform, not about being a pose synchronizer.

I can't think of a case where responding to events (as a pose synchronizer) is better than attaching to or tracking towards a specific transform, but I'm interested to hear ideas and answers to his questions.

StephenHodgson:
We just listen for a source found/lost event, then everything after that can follow the proxy reference we captured (if one exists).

StephenHodgson:
>Yes, that is exactly how the ControllerFinder works.

Sure, but the way we accessed and got that data broke the dependency design patterns we were trying to break away from in HTK.

@SimonDarksideJ made a wonderful graphic to demonstrate the appropriate flow:
image
 
Where in your implementation you broke this pattern by going directly from UX/Scene Object into the device/SDK layer (which breaks some of the core principles we were trying to setup).

keveleigh:
>Sure, but the way ControllerFinder accessed and got that data broke the dependency design patterns we were trying to break away from in HTK.

That's exactly why this PR exist, to do it "correctly".

StephenHodgson:
@Ecnassianer I think I misread your question from last night. After talking with Kurtis, I just wanted to give you a better answer. 

Here's our conversation from slack (for transparency):

@keveleigh :
>With the visualizer, we'll now have access to the controller transform. That gives us the position and rotation of the controller at all times. If we parent ourselves to it, we get "controller synchronization" for free, with no additional events outside of source detected / lost.
>
>Currently, synchronization (via the pose synchronizer) happens based on pose updated events. When, if ever, would the continuous pose events be better than having the transform reference?

@StephenHodgson :
>I don't think there's much of a difference. The events update the transform as soon as they arrive.
>
>I think it'll be simpler and easier to just use the transform (like I think Eric originally wanted) Less code to write.
>
>As for actually moving and updating the "proxy", something will need to be handling the controller's events
>if that's internal to the pose synchonizer itself that should be okay if it's easy to get the controller game object reference, correct?

Ecnassianer:
I would just Ctrl-F "Visualizer" and replace with Proxy.

I'd be willing to do it, but I have never cloned from a PR before so I'm not sure how to get at your branch. Is there a link in the PR? Seems like everywhere it mentions the branch is just a label (StephenHodgson:vNEXT-ControllerProxyVisualizationProfile) without a URL to your fork. Is there an easier way to get your changes locally?

Ecnassianer:
@radicalad  helped me out, I'll get you a PR in a bit

StephenHodgson:
@Ecnassianer Sorry, I'm still not sold on Proxy. I think for now the naming convention we have is okay. 

We really need to get this in before sept 30th

davidkline-ms:
> @radicalad helped me out, I'll get you a PR in a bit

@Ecnassianer, any update on your PR?

Ecnassianer:
@davidkline-ms I submitted it later that day and it spawned another long discussion:
https://github.com/StephenHodgson/MixedRealityToolkit-Unity/pull/36

@StephenHodgson It feels like you're not really trying to address my feedback in good faith. It's not my job to defend the architecture choices you believe in. You created a PR specifically to address the confusion around how to use this architecture, and I offered advice about a confusing name. I even went so far as to write you a PR to help clarify what would make this API less confusing for a person in my seat. If you want to help clarify how solvers should talk to controllers, then do the work to make that happen. @keveleigh had working solver code almost three weeks ago, and we've given you plenty of time to come up with a way for that code to get in that you're happy with.

If you are not sold on any of the examples I have offered, then name it whatever you want. I've dismissed my review and won't be re-reviewing this PR.

StephenHodgson:
I really do appreciate the PR, Eric. I'm not trying to be dismissive, and I really do appreciate your contributions and work towards this common goal.

Naming is very important to how we want instantly convey what we're trying to describe, and some patterns have been fairly deliberate.

My only concern is that if we change the naming for the controllers, we should also address the same for boundary and pointers as they also share similar functions (and could add to the confusion if the naming convention were to change).

I didn't reject nor approve your PR until we had a better understanding and common agreement.

SimonDarksideJ:
Hi @ericahlstrom I agree with Stephen on this, we really appreciate you taking the time to put a PR in, but through that conversation we've discussed the different terms used and stated why visualization is a better fit name.

Proxy fits for something like the ControllerPoseSynchronizer, which is acting on behalf of the controller to manipulate an object in the scene.  However that name hasn't been brought in to question so far.

Visualization by definition, controls how you visualize something (in this case a controller) in the scene.  The same definition was also used for the Boundary visualizer, with places visuals in the scene to represent the boundary.

Both Stephen and I at times (as well as others in the team) have had PRs rejected because they didn't fit (Heck, I renamed the namespaces on the entire HTK at one point, weeks of work, only to see it trashed as we started vNext instead :D ).  So you're not being singled out, but the consensus so far doesn't warrant the change.

Thoughts @davidkline-ms / @keveleigh , should controller visualization be renamed to proxy and in that list also effectively renaming  Boundary and Pointer visualization?

I certainly don't want to dissuade @Ecnassianer doing more contributions, but his PR just doesn't seem a good change.

StephenHodgson:
Thanks for the review @keveleigh, I'll be sure to address that.

StephenHodgson:
Yeah seems like I also need to update it to lock as well. 

I'll also double check to make sure an empty cube prefab tracks. I'm thinking the chevron didn't get the component added correctly at runtime.

keveleigh:
>I'm thinking the chevron didn't get the component added correctly at runtime.

As far as I can tell, the component is only added at [edit time in the inspector](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2837/files#diff-226a65267f295f1911d642c8ac1aa0d6R227).

davidkline-ms:
> As far as I can tell, the component is only added at edit time in the inspector.

What is the intent for the feature? Editor time or runtime binding?

StephenHodgson:
> What is the intent for the feature? Editor time or runtime binding?

This already does both.  So when we add glTF support it'll use the runtime binding.

StephenHodgson:
@keveleigh I tested by making two cube prefabs for left and right hands and assigning them in new controller definitions.

Should add the component at edit time. 

When we add glTF and support for loading controllers at runtime we'll also add the component then too. I added a snippet above for how we could do it (should be similar to the inspector logic)

Let me know if you have any questions or concerns.

StephenHodgson:
Sorry I thought I had pushed the updates. @keveleigh it's ready for a new review.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityControllerVisualizationProfileInspector.cs
```diff
@@ -0,0 +1,271 @@
...
+
+                EditorGUIUtility.labelWidth = 64f;
+                EditorGUIUtility.fieldWidth = 64f;
+                var controllerSetting = controllerList.GetArrayElementAtIndex(i);
+                var mixedRealityControllerMappingDescription = controllerSetting.FindPropertyRelative("description");
+                bool hasValidType = thisProfile.ControllerVisualizationSettings[i].ControllerType != null &&
+                                    thisProfile.ControllerVisualizationSettings[i].ControllerType.Type != null;
+                if (hasValidType)
+                {
+                    mixedRealityControllerMappingDescription.stringValue = thisProfile.ControllerVisualizationSettings[i].ControllerType.Type.Name.ToProperCase();
```
keveleigh: Not blockers, but:

1. We should reset this to some default description if `hasValidType` is false
2. There appears to be strange spacing in the name when this is rendered (note the larger spacing between the words in the controller type vs "Right Hand"

image

StephenHodgson: yeah I also noticed that the spacing was strange, but I'm not sure what's going on because I'm only adding a single space. Maybe Unity is increasing the word spacing in the label somehow?

Do you have any suggestions on a default description? I'm more than willing to accommodate that.  Maybe `Spatial Controller <handedness> Hand`?

keveleigh: Yeah, that would work. Or, since we're notifying them that a controller type must be defined, even something like `Undefined Controller <handedness> Hand` seems fine. I'll leave it up to you, and it can always be iterated on in the future.

keveleigh: >I'm not sure what's going on because I'm only adding a single space

Ah I figured it out! You're calling `ToProperCase()` both when you set the `stringValue` and when you turn the `stringValue` into the `LabelField`. Both calls are adding a space.

StephenHodgson: Yup I just noticed that. I'll be sure to fix it as well.

StephenHodgson: Fix incoming

SimonDarksideJ:
Sadly still not working for me on the current update.  My model shows but does not synchronise.
Replication steps:

- Create new 3D object and scale appropriately
- Add new Controller Definition to the Visualizer config for WMR left
- Set model to override model property
- Run

Model appears in the scene but does not move.  Noted it does not have a visualiser attached.

This needs to check at "create" time in the scene for the presence of a visualizer component, instead of at edit time maybe?   As it's not automatically added for specific controller model overrides (it does for the global models)

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityControllerVisualizationProfileInspector.cs
```diff
@@ -0,0 +1,278 @@
...
+            }
+
+            var componentList = modelPrefab.GetComponentsInChildren<IMixedRealityControllerVisualizer>();
+
+            if (componentList == null || componentList.Length == 0)
+            {
+                if (thisProfile.ControllerVisualizationType != null &&
+                    thisProfile.ControllerVisualizationType.Type != null)
+                {
+                    modelPrefab.AddComponent(thisProfile.ControllerVisualizationType.Type);
```
keveleigh: Looks like even when the visualization script is successfully added, this still falls through to return `false` at line 275. This causes override model setting to take two steps. The first time adds the script but doesn't fill in the field.

keveleigh:
@SimonDarksideJ I wasn't able to repro your report of being able to add an override model which didn't add the visualizer script.

Following your repro steps (though needing to do step 3 twice [see my review]), I was successfully able to add a custom model. If I didn't do step 3 twice, no override model was spawned (because the field wasn't set), though the prefab did have the visualizer script successfully added anyway.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityControllerVisualizationProfileInspector.cs
StephenHodgson: I'll be sure to fix. Thanks.

StephenHodgson:
I'm going to wait until @SimonDarksideJ retests just to make sure we're good, before merging. 

@SimonDarksideJ please merge if you approve.

davidkline-ms:
> I'm going to wait until @SimonDarksideJ retests just to make sure we're good, before merging.

that's why @keveleigh and I did not merge, we discussed making sure Simon was able to get it to work.

PR #2840, opened 2018-09-26
Merge diagnostics into development

ryzngard:
Overview
---

This adds the full diagnostics feature for use in MRTK. Diagnostics shows:

* Frames per second 
* Frame time
* Cpu usage
* Memory Usage

Diagnostics are disabled by default, and each component of diagnostics is configurable to be enabled/disabled.


davidkline-ms:
Can you merge mrtk_dev into the diagnostics branch so that merge conflicts can be fixed?

StephenHodgson:
Chances are we just need to merge dev back into the feature branch to resolve

Assets/MixedRealityToolkit/_Core/Interfaces/Diagnostics/IMixedRealityDiagnosticsManager.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Enable / disable memory profiling when the diagnostic panel is visible. 
+        /// </summary>
+        bool ShowMemory { get; }
+
+        /// <summary>
+        /// Gets the <see cref="GameObject"/> that represents the diagnostic visualization
+        /// </summary>
+        GameObject DiagnosticVisualization { get; }
```
StephenHodgson: Do we want to agree on a naming convention here?

There's been talk of using proxy vs visualization

Ecnassianer: Sorry, I haven't been following this PR/issue. What is DiagnosticVisualization for? I'm not sure how this relates to ControllerProxies (or another kind of scene proxy?).

StephenHodgson: It's a naming pattern we've already established using visualization vs proxy as the game object reference.

I do like proxy a bit more, as it's just a reference to the game object, not the things being rendered (the renderer component comes to mind here)

ryzngard: I'm fine with renaming to Proxy if that's the pattern we have.

Ecnassianer: I'd only name it a Proxy if it's a game object proxying something in the device layer. If it's just a game object for visualizing diagnostic data, I think Visualization or Visuals or Visualizer are all fine names.

StephenHodgson: Just trying to figure out what pattern we want to stick with. 🤔

StephenHodgson: They're both game object references.

I just wanna make sure we're on the same page about how we want to talk about and reference these things.

keveleigh: >They're both game object references.

I don't think that was the important part of the original suggestion. It was pretty specific to the controller scenario (and probably a small number of others), but I don't see the relationship here.

StephenHodgson: Naming is pretty important to the idea of a concept or thing. 

I don't really see a difference here.
I see them as pretty much the same concepts.

We have a game object reference that doesn't actually do any "visualization" but it's a stand in (proxy) for the concept as a whole.

If we wanted to be specific that it's actually "visualizing" something the field type should reflect that by being a renderer or something that does the rendering in the scene.

keveleigh: The naming isn't about GameObject vs renderer vs transform vs anything else.

It's about physical device/object vs text.

As Eric said above:

>I'd only name it a Proxy if it's a game object proxying something that lives in the device layer. If it's just a game object for visualizing diagnostic data, I think Visualization or Visuals or Visualizer are all fine names.

StephenHodgson: That's the thing tho. It's not text. It's a game object.

keveleigh: Right, and, again, it's not about the fact that they're both GameObjects. It's about the things those GameObjects are representing, which are very different.

StephenHodgson: But they're similar in the fact that they reference a thing in the scene, and that reference doesn't actually do any "visualization" or rendering of said thing.

keveleigh: Are you proposing that all GameObject references in code should be appended with `Proxy`?

StephenHodgson: No, just the ones that have "visualization" or similar as the name.

ryzngard: I created a new PR from a branch I can edit to handle merge conflicts. I am going to abandon this one, so please continue any discussions in #2846

PR #2844, opened 2018-09-27
Moved some of the non-guidance info out of the architecture docs

Ecnassianer:
Overview
---
I'll follow up with a PR removing this from the other doc. (https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2845)

I am not opposed to deleting this list, if it's not longer relevant. I just wanted to get it out of the architecture guidance.

### DIFF
--- a/MRTK-WorkRoadmap.md
+++ b/MRTK-WorkRoadmap.md
@@ -1,0 +1,43 @@
+# Work Roadmap for early vNext development 
+
+Here are some of the goals and tasks we're pursuing during the Alpha/Beta and early release stages of MRTK vNext. Please create github issues for specific tasks, but if there's a general area that needs planning and development, feel free to add it here.
+ 
+## Fringe thoughts (dumping ground for ideas :D)
+
+ * Defining headset gameplay, allowing for sitting / standing or free roaming play.
+Has to be easy enough to enable single projects to handle both behaviors
+ * Defining headset movement, is it free roam, shifting or teleporting?
+ * With Teleporting specifically, is it cursor based or zone based (something we don’t have).  After watching the Rick and Morty retrospective, they struggled with this and movement space, something an SDK should be able to provide a starter experience for.
+ * Adding interaction mechanics to objects in the scene easily. Is it grabbable, moveable, altered by gravity etc.
+ * Helpers for 3d’ifying scene objects, 2 sided drawing or best practices
+ * Examples, Examples, Examples – Ensuring our examples are bigger than current based off the latest examples scene that came from VRTK which was well received. (showed ALL interaction options within a small house)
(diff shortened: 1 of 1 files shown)


StephenHodgson:
I'm wondering if we want to have a better break down with a table and such.

Ecnassianer:
We probably do, but I'm focused on improving the architecture doc for now. I just wanted to get this stuff saved out somewhere else.

MRTK-WorkRoadmap.md
```diff
@@ -0,0 +1,43 @@
...
+ * With Teleporting specifically, is it cursor based or zone based (something we don’t have).  After watching the Rick and Morty retrospective, they struggled with this and movement space, something an SDK should be able to provide a starter experience for.
+ * Adding interaction mechanics to objects in the scene easily. Is it grabbable, moveable, altered by gravity etc.
+ * Helpers for 3d’ifying scene objects, 2 sided drawing or best practices
+ * Examples, Examples, Examples – Ensuring our examples are bigger than current based off the latest examples scene that came from VRTK which was well received. (showed ALL interaction options within a small house)
+ * Videos – a YT / FB video channel with educational snippets dedicated to MRTK 
+
+ 
+## Proposed Tasks
+
+To ensure we have a smooth transition, we need several key tasks to be completed, to ensure a smooth transition and ensure we limit (as much as possible) any future breaking changes:
```
radicalad: how about: "To ensure we have a smooth transition, we need several key tasks to be completed and ensure we limit (as much as possible) any future breaking changes:"

MRTK-WorkRoadmap.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. Build the new frontend architecture, to enable both existing users and new to start building from, comprising of an Initial set interactable component prefabs / components:
```
radicalad: Something like "Build the new front-end architecture, enabling both new and existing users to start building from. Comprising of an initial set of interactable prefabs / components: "

MRTK-WorkRoadmap.md
Ecnassianer: Since this is just a copy paste from that other doc, do you mind if I make these changes in a separate PR. That way it's just a huge subtract from the other file and a huge word-for-word add here.

MRTK-WorkRoadmap.md
Ecnassianer: Same as above.

MRTK-WorkRoadmap.md
radicalad: works for me.

Ecnassianer:
This is my favorite github error message:
image

PR #2846, opened 2018-09-27
Feature/mrtk diagnostics

ryzngard:
Overview
---
This adds the full diagnostics feature for use in MRTK. Diagnostics shows:
* Frames per second
* Frame time
* Cpu usage
* Memory Usage

Diagnostics are disabled by default, and each component of diagnostics is configurable to be enabled/disabled.


ryzngard:
migrating from #2840

PR #2848, opened 2018-09-27
Update spatial architecture to include observer interface

davidkline-ms:
Overview
---
Adds the IMixedRealitySpatialObserver interface and updates the existing architecture files / documentation to support this model.

This change facilitates support for platform specific implementations without having platform specific code in the spatial awareness system.


Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -65,12 +66,17 @@ private GameObject CreateSecondGenerationParent(string name)
             return secondGeneration;
         }
 
-        // todo
+        private IMixedRealitySpatialAwarenessObserver spatialAwarenessObserver = null;
+
+        /// <summary>
+        /// The <see cref="IMixedRealitySpatialAwarenessObserver"/>, if any, that is active on the current platform.
+        /// </summary>
+        private IMixedRealitySpatialAwarenessObserver SpatialAwarenessObserver => spatialAwarenessObserver ?? (spatialAwarenessObserver = MixedRealityManager.Instance.GetManager<IMixedRealitySpatialAwarenessObserver>());
```
StephenHodgson: Seems like this might be a `IMixedRealityComponent`? Although I think it's fine here at the moment. Thoughts?

Assets/MixedRealityToolkit/_Core/Interfaces/SpatialAwareness/IMixedRealitySpatialAwarenessSystem.cs
```diff
@@ -106,17 +104,10 @@ public interface IMixedRealitySpatialAwarenessSystem : IMixedRealityEventSystem,
...
-        /// </summary>
-        /// <returns>Dictionary of <see cref="IMixedRealitySpatialAwarenessMeshDescription"/>s, indexed by the mesh id.</returns>
-        Dictionary<uint, IMixedRealitySpatialAwarenessMeshDescription> GetMeshes();
-
         /// <summary>
         /// Gets the collection of <see cref="GameObject"/>s being managed by the spatial awareness mesh subsystem.
         /// </summary>
-        /// <returns>Dictionary of <see cref="GameObject"/>s, indexed by the mesh id.</returns>
-        Dictionary<uint, GameObject> GetMeshObjects();
+        Dictionary<int, GameObject> MeshObjects { get; }
```
StephenHodgson: what prompted the change from `uint` to `int` type?

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -398,33 +400,30 @@ public float UpdateInterval
...
         {
-            if (!IsObserverRunning) { return; }
-
-#if UNITY_WSA
-            // todo
-#endif // UNITY_WSA
-
-            IsObserverRunning = false;
+            if (SpatialAwarenessObserver == null) { return; }
+            SpatialAwarenessObserver.StopObserving();
```
ryzngard: Can `IsObserverRunning` be true while `SpatialAwarenessObserver` is null?

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
keveleigh: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2848/files#diff-5f07c17e8951e44aaa4d837bbb50a0c7R12

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -157,68 +163,64 @@ public override void Destroy()
...
-        /// <param name="meshObject">The mesh subsystem managed <see cref="GameObject"/> for the mesh.</param>
-        private void RaiseMeshUpdated(uint meshId, Vector3 position, Mesh meshData, GameObject meshObject = null)
+        /// <param name="meshObject">The mesh <see cref="GameObject"/>.</param>
+        private void RaiseMeshAvailable(uint meshId, GameObject meshObject)
         {
             // todo
         }
 
         /// <summary>
         /// Event sent whenever a mesh is updated.
```
ryzngard: This comment is likely incorrect. It seems like it would be when the mesh is made available, nut updated.

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -157,68 +163,64 @@ public override void Destroy()
...
         /// </summary>
-        private static readonly ExecuteEvents.EventFunction<IMixedRealitySpatialAwarenessMeshHandler> OnMeshUpdated =
+        private static readonly ExecuteEvents.EventFunction<IMixedRealitySpatialAwarenessMeshHandler> OnMeshAvailable =
```
ryzngard: `OnMeshAdded` would match `OnMeshRemoved`, is there a reason we don't name it that?

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
StephenHodgson: I'm wondering if we want to get it from the profile, or the manager (as discussed in the shiproom last week)

I don't think we had an in depth discussion about it. I could go either way.

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
ryzngard: Also, could we just move to IObservable<T> implementations? Do they not work with the implementation we have?

Assets/MixedRealityToolkit/_Core/Interfaces/SpatialAwareness/IMixedRealitySpatialAwarenessSystem.cs
ryzngard: +1

ryzngard: Also, generally speaking I believe interface contracts should use interfaces rather than concrete types. `IDictionary<,>` over `Dictionary<,>` for example

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
StephenHodgson: Seems to make sense. I second that.

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
davidkline-ms: The observer will be a registered component. this is how I believe the input device managers work. Hence the design :)

Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
davidkline-ms: add/update has been restored.

PR #2852, opened 2018-09-28
Assets and some layout code for Interactables - 1 of 4

killerantz:
Overview
---

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit/_Core/Resources/Shaders/3DTextShader.shader
```diff
@@ -0,0 +1,114 @@
+﻿// Upgrade NOTE: upgraded instancing buffer 'Props' to new syntax.
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+///
+/// Basic 3D Text shader with proper z-sorting
+///
+Shader "Mixed Reality Toolkit/3DTextShader"
```
StephenHodgson: I'm not sure we want to bring this over.

Assets/MixedRealityToolkit-SDK/Features/UX/Shaders/FontShader3D.shader
```diff
@@ -0,0 +1,116 @@
+﻿// Upgrade NOTE: upgraded instancing buffer 'Props' to new syntax.
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+///
+/// Basic 3D Text shader with proper z-shorting and culling options.
+/// this shader is based on the HoloToolkit/3DTextShader
+///
+Shader "Mixed Reality Toolkit/FontShader3D"
```
StephenHodgson: More font shaders?

Do we want to put these in standard assets?

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonCollider.cs
```diff
@@ -0,0 +1,64 @@
...
+                if (capsule != null)
+                {
+                    capsule.radius = CopyFrom.transform.localScale.x * ScaleFactor.x + Expand.x;
+                    capsule.height = CopyFrom.transform.localScale.y * ScaleFactor.y + Expand.y;
+                }
+            }
+        }
+
+        // Update is called once per frame
+        void Update()
```
StephenHodgson: missing explicit access modifiers, should be `private`

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonBorder.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace Microsoft.MixedReality.Toolkit.SDK.UX
+{
+    /// <summary>
+    /// Use a Cube 3D object as a border segment relative to the scale of the AnchorTransform
+    /// </summary>
+    [ExecuteInEditMode]
+    public class ButtonBorder : MonoBehaviour
+    {
+        [Tooltip("A pixel to Unity unit conversion, Default: 2048x2048 pixels covers a 1x1 Unity Unit or default primitive size")]
+        public float BasePixelScale = 2048;
```
StephenHodgson: All fields should be private with public property accessors is needed.

radicalad:
I'm curious why the base interactable is registering itself as a global listener with the input system? This is going to be a HUGE perf risk, and I don't see the upside as all events can be consumed by the appropriate interface.

StephenHodgson:
>This is going to be a HUGE perf risk

How is this a perf risk?

StephenHodgson:
People keep saying that but I've yet to see any hard evidence that our event system creates garbage or impacts perf.

That's the whole event system is performance & data driven.

StephenHodgson:
Todd, Is there a time when we need to send an event to the  base interactable without it being focused by a pointer?

radicalad:
that is kind of my point, interactable is essentially a 1:1 `InteractionReceiver`. Interactable is subscribing to every available event interface and then filtering to see if the raised event is applicable. In the case of `InteractionReceiver`, that is intended as it is a one to many relationship. But in the case of interactable, if I have 100 buttons the event filtering is going to get out of hand quickly.

StephenHodgson:
I'm still trying to understand if we even need the `InteractionReceiver` now that we've abstracted out the input to be `Action` based.

radicalad:
I haven't seen the pattern in use so I can't comment, but I'm more concerned of the 1:1 relationship of every `Interactable` Button acting essentially as its own `InteractionReceiver`.

StephenHodgson:
I still need to dig into the PR to see how things are structured. I'm sure to read it though to get a better understanding.

Railboy:
Agree with Addison. Picture a scene with some virtual pianos that use Interactable for its keys. Even if interfaces are filtering out some events that's several hundred event calls being raised and discarded in a scenario where it intent to raise events is relatively easy to determine.

Interactable's main purpose is look / feel so it's probably fine to roll with it as-is and refine how it uses events later.

In the future an input _targeting_ system really ought be implemented to compliment the focus system, where 'global' input events are only raised in a subset of available receivers based on some kind of constraint (proximity, collision, whatever). We tried to incorporate this into the previous input system and it didn't go well. But the new input system is flexible enough to accommodate it.

StephenHodgson:
>In the future an input targeting system really ought be implemented to compliment the focus system, where 'global' input events are only raised in a subset of available receivers based on some kind of constraint (proximity, collision, whatever).

I agree, and I was looking into this a bit more, but ultimately I'm trying to understand why it needs to be registered. I'll look into that a bit more. Intractables seem like something that are pretty specific to reacting to "focused" events (which means they don't need to be registered).

killerantz:
>           Todd, Is there a time when we need to send an event to the  base interactable without it being focused by a pointer?

There could be a button that does not require focus, such as a modal. We had that functionality in the old InputManager, the Model Listener and Global listener. When we did gestures, I would register the interactive as a global listener to get the release event. Now that all events are just global, it's a lot of filtering. That filtering has to occur somewhere though, right now it's similar to polling.

I would like to get action events, we could register to the input manager with the actions we care about for each button and only those action related inputs will be routed to the Interactable. It could also register as a global or focus only listener.

The interactable basically does two things, evaluating inputs to create a simple set of states that drive scensory feedback and then routes those states and some of the input data to receivers to look for patterns and make things happen on focus, press, up, rotation, click, hold, gesture, shake or any other detected patther. Actually the receiver looks at the Interactable and grabs that information, this way the interactable only has to worry about state and feedback. All the extra crazy stuff is seperated, like multi-hand gestures and combinations of actions.

A collection of pointer and action information would be really cool. Pointers could contain an array of actions that are active on each pointer with the current action values. We could have a button that is looking for grip, while a controller has focus, the button looks at those actions to figure out state. The theme system would just react to the main actions it cares about, like select or grip. The state of the Interactable and the current list of pointers with focus is also seen by a receiver which could loop through all the actions on each controller and evaluate those values. This way, while gripping something, I can use the trackpad or stick to rotate the object or change it’s state in some way, like color or size.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonCollider.cs
killerantz: Good catch.

keveleigh:
>Now that all events are just global, it's a lot of filtering.

We still have the same event routing rules, with [global](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs#L271), [modal](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs#L283), [focused](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs#L306), and [fallback](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs#L315) listeners all still existing, just like the HTK. That part of the buttons shouldn't need to change.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonBorder.cs
killerantz: Will do, so everything is private with [SerializeField] tag?

Assets/MixedRealityToolkit/_Core/Resources/Shaders/3DTextShader.shader
killerantz: Wait, I think I added the FontShader3D under Features/UX which has the culling option. Did I add that one too? We should only have one, just let me know where to put it.

killerantz:
For sure, there are still some holes in the event information I am getting and need. Like when a controller has gaze and does a selection, using the PointerEventHandler, I get two of everything. Two downs, ups, and clicks. But with gaze and using the controller as a clicker, pulling the trigger, all the events fire properly; one time each. 

I'm sure there's some things there that would make it easier.

keveleigh:
>Like when a controller has gaze and does a selection, using the PointerEventHandler, I get two of everything. 

I think that's because you're getting both the global event and the focused event. Either removing the global registration or adding `eventData.Use()` to the event handler should resolve that.

killerantz:
I'm not sure how to register for the global events, is that how I setup the IMixedRealtyInputSystem?
`
private static IMixedRealityInputSystem inputSystem = null;
        protected static IMixedRealityInputSystem InputSystem => inputSystem ?? (inputSystem = MixedRealityManager.Instance.GetManager<IMixedRealityInputSystem>());
`
Do I not need this? I copied it from another script I found somewhere.

StephenHodgson:
@killerantz That should get a lot simpler soon.

killerantz:
That did it, not registering worked. it now works really well on MR devices and the HoloLens!

StephenHodgson:
I think it would be super helpful to break this first one up into assets and code passes as well.

Any assets you can add first, like models, materials, textures, animations, etc before any code would help us get through these a bit faster.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Materials/Background.mat
```diff
@@ -0,0 +1,178 @@
+%YAML 1.1
```
davidkline-ms: Can we use one of the MRTK standard materials here?

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Materials/ColorButton.mat
```diff
@@ -0,0 +1,141 @@
+%YAML 1.1
```
davidkline-ms: same question re: MRTK standard materials.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/InspectorFields/GenericFields.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// A set of Inspector fields for setting up properties in a
+    /// component to automatically draw in a custom inspector
+    /// </summary>
+    public class GenericFields<T>
+    {
+        public static void LoadSettings(T target, List<PropertySetting> settings)
+        {
+            Type myType = target.GetType();
+
+            foreach (PropertyInfo prop in myType.GetProperties())
```
davidkline-ms: for consistency with the rest of the MRTK, please replace foreach with for

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/InspectorFields/InspectorField.cs
```diff
@@ -0,0 +1,208 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX
+{
+    /// <summary>
+    /// A set of Inspector fields for setting up properties in a
+    /// component to automatically draw in a custom inspector
+    /// </summary>
+    public class InspectorField : Attribute
+    {
+        public enum FieldTypes { Float, Int, String, Bool, Color, DropdownInt, DropdownString, GameObject, ScriptableObject, Object, Material, Texture, Vector2, Vector3, Vector4, Curve, Quaternion, AudioClip, Event }
```
davidkline-ms: this formatting is hard to read. please put each enum value on it's own line

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonBorder.cs
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX
+{
+    /// <summary>
+    /// Use a Cube 3D object as a border segment relative to the scale of the AnchorTransform
+    /// </summary>
+    [ExecuteInEditMode]
+    public class ButtonBorder : MonoBehaviour
+    {
+        [Tooltip("A pixel to Unity unit conversion, Default: 2048x2048 pixels covers a 1x1 Unity Unit or default primitive size")]
```
davidkline-ms: please add summary tags for api doc generation.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonCollider.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Microsoft.MixedReality.Toolkit.SDK.UX
+{
+    /// <summary>
+    /// Scales the collider on one object relative to the transform's scale of another object
+    /// </summary>
+    [ExecuteInEditMode]
+    public class ButtonCollider : MonoBehaviour
+    {
+
+        [Tooltip("the object to copy the scale from")]
```
davidkline-ms: same here re: summary tags

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonLayout.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// <summary>
+    /// Place an object in space relative to another object's scale
+    /// Good for responsive buttons that can stretch and object realign
+    /// </summary>
+    [ExecuteInEditMode]
+    public class ButtonLayout : MonoBehaviour
+    {
+        [Tooltip("Where to set this object's center point in relation to the Anchor's center point")]
+        [SerializeField]
+        private Vector3 Alignment;
```
davidkline-ms: please add summary tags

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonSize.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX
+{
+    [ExecuteInEditMode]
+    public class ButtonSize : MonoBehaviour
+    {
+        [Tooltip("A pixel to Unity unit conversion, Default: 2048x2048 pixels covers a 1x1 Unity Unit or default primitive size")]
```
davidkline-ms: summary tags needed, thanks

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonSizeOffset.cs
```diff
@@ -0,0 +1,78 @@
...
+namespace Microsoft.MixedReality.Toolkit.SDK.UX
+{
+    /// <summary>
+    /// Scales an object relative the scale of the AnchorTransform
+    /// </summary>
+    [ExecuteInEditMode]
+    public class ButtonSizeOffset : MonoBehaviour
+    {
+
+        [Tooltip("A pixel to Unity unit conversion, Default: 2048x2048 pixels covers a 1x1 Unity Unit or default primitive size")]
```
davidkline-ms: summary tags, thx

Assets/MixedRealityToolkit/_Core/Resources/Shaders/3DTextShader.shader
davidkline-ms: the shaders need to be availabe to all SDK packages. please co-locate the chosen font shader with the mrtk standard shader. we will ensure (post this pr series) that they are in the correct place. thx

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Materials/Background.mat
killerantz: Do we have a set of button materials?

Assets/MixedRealityToolkit/_Core/Inspectors/Text3DShaderGUI.cs
```diff
@@ -0,0 +1,59 @@
...
+            EditorGUI.BeginChangeCheck();
+            {
+                base.OnGUI(matEditor, props);
+                ShowMainGUI(matEditor);
+            }
+        }
+
+        protected virtual void ShowMainGUI(MaterialEditor matEditor)
+        {
+            //matEditor.ShaderProperty(cullMode, Styles.cullMode);
```
SimonDarksideJ: No redundant code, either remove or implement
Same throughout this file

Assets/MixedRealityToolkit/_Core/Resources/Shaders/Text3DShader.shader
```diff
@@ -0,0 +1,116 @@
+﻿// Upgrade NOTE: upgraded instancing buffer 'Props' to new syntax.
```
SimonDarksideJ: Is this shader / inspector needed in favour of using TXText now?

StephenHodgson: Why don't we just use tmpro? It's free and looks a lot better.

StephenHodgson:
>Please squash the commit history so as not to dirty things once it's merged.

We haven't enforced this before, why now?

Assets/MixedRealityToolkit/_Core/Resources/Shaders/Text3DShader.shader
killerantz: We should use TMPro, it's the new standard now that it ships with Unity, and our team is getting ready to move over, but we have some tools that don't support it yet. This is a stop-gap until we get our tools updated or we can draw a line here to say TMPro only beyond this point.

Assets/MixedRealityToolkit/_Core/Inspectors/Text3DShaderGUI.cs
killerantz: removed unused methods

Assets/MixedRealityToolkit/_Core/Resources/Shaders/Text3DShader.shader
killerantz: @SimonDarksideJ We are using this with TextMesh. The Selawik font is currently in the repo and the materials were using the default Unity Font Shader, which is a bad practice so we added this shader in HTK. I brought in the shader to bring MRTK to HTK parody as far as the materials go. We also need to create TMPro Font assets for these fonts to push TMPro as the new standard. If we remove this shader, we should remove the font materials as well and just have the TMPro Font assets.

SimonDarksideJ:
Any progress in this area @killerantz ?
In the shiproom call, I was under the impression you were going to close these PRs, consolidate your changes in to a single local repo and then submit a PR for a new branch.  Please correct if my understanding was wrong 
/ cc @davidkline-ms

SimonDarksideJ:
Lol, no worries @killerantz , but might be best to close these original PR's to avoid confusion :D 
Will help where I can with the new branch, including PR's to help it along.

killerantz:
Reverted the icons, they were persistent and finally made it in :)

PR #2853, opened 2018-09-28
States for Interactables 2 of 4

killerantz:
Overview
---

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/InspectorFields/InspectorField.cs
```diff
@@ -0,0 +1,208 @@
...
+            setting.Name = fieldName;
+
+            UpdatePropertySetting(setting, fieldValue);
+
+            return setting;
+        }
+        
+        public static PropertySetting UpdatePropertySetting(PropertySetting setting, object update)
+        {
+            switch (setting.Type)
```
Ecnassianer: I don't know for sure that this exists, but is there not a way to do something like this:
(setting.Type)update;

Why do we need to keep the setting value in these type specific fields? Only one of them can be set at once, right?

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonLayout.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// Good for responsive buttons that can stretch and object realign
+    /// </summary>
+    [ExecuteInEditMode]
+    public class ButtonLayout : MonoBehaviour
+    {
+        [Tooltip("Where to set this object's center point in relation to the Anchor's center point")]
+        public Vector3 Alignment;
+
+        [Tooltip("A pixel to Unity unit conversion, Default: 2048x2048 pixels covers a 1x1 Unity Unit or default primitive size")]
+        public float BasePixelSize = 2048;
```
Ecnassianer: Feels weird that this appears in several places. Should it be in a single place the project profiles? Would someone really want to set it to one value on a ButtonLayout and another on a ButtonCollider?

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonSize.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX
+{
+    [ExecuteInEditMode]
+    public class ButtonSize : MonoBehaviour
+    {
+        [Tooltip("A pixel to Unity unit conversion, Default: 2048x2048 pixels covers a 1x1 Unity Unit or default primitive size")]
+        public float BasePixelScale = 2048;
```
Ecnassianer: Same as above

Ecnassianer: Though "ButtonSize" might be the class where the single reference belongs, if it's not in the profiles.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/States/InteractableStates.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// <summary>
+    /// list of Interactable states and basic comparison
+    /// </summary>
+    public class InteractableStates : StateModel
+    {
+        public enum InteractableStateEnum {
+            /// <summary>
+            /// Default state, nothing happening
+            /// </summary>
+            Default,
```
Ecnassianer: Default  = 0,

https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/CodingGuidelines.md#initialize-enums

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/States/States.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                TypeInfo info = type.GetTypeInfo();
+                if (info.BaseType.Equals(typeof(InteractableStates)) || type.Equals(typeof(InteractableStates)))
+                {
+                    stateTypes.Add(type);
+                    names.Add(type.Name);
+                }
+            }
+
+            /* works with IL2CPP but not in .NET
```
Ecnassianer: Is this code being saved for something else? If you're about to revisit it in your next PR, add a //TODO comment explaining what needs to be done, otherwise best to delete it.

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Interactable/InspectorBase.cs
```diff
@@ -0,0 +1,646 @@
...
+            bool triggered = false;
+            if (GUILayout.Button(label, smallButton, GUILayout.Width(smallButtonWidth)))
+            {
+                callback(index, prop);
+                triggered = true;
+            }
+            return triggered;
+        }
+
+        public static bool SmallButton(GUIContent label, int[] arr, MultiListButtonEvent callback, SerializedProperty prop = null)
```
Ecnassianer: Kinda unclear what "arr" is here. Is there a better name that we could use?
Also would help if all these functions had summary comments.

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Interactable/InspectorBase.cs
```diff
@@ -0,0 +1,646 @@
...
+                case InspectorField.FieldTypes.Quaternion:
+                    SerializedProperty quaternionValue = prop.FindPropertyRelative("QuaternionValue");
+                    quaternionValue.quaternionValue = (Quaternion)update;
+                    break;
+                case InspectorField.FieldTypes.AudioClip:
+                    SerializedProperty audioClip = prop.FindPropertyRelative("AudioClipValue");
+                    audioClip.objectReferenceValue = (AudioClip)update;
+                    break;
+                case InspectorField.FieldTypes.Event:
+                    //SerializedProperty uEvent = prop.FindPropertyRelative("EventValue");
```
Ecnassianer: Is this supposed to be commented out?

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Interactable/InspectorBase.cs
```diff
@@ -0,0 +1,646 @@
...
+                    SerializedProperty audioClip = prop.FindPropertyRelative("AudioClipValue");
+                    EditorGUILayout.PropertyField(audioClip, new GUIContent(label.stringValue, tooltip.stringValue), false);
+                    break;
+                case InspectorField.FieldTypes.Event:
+                    SerializedProperty uEvent = prop.FindPropertyRelative("EventValue");
+                    if (uEvent != null)
+                    {
+                        //Debug.Log("update? " + uEvent);
+                    }
+                    EditorGUILayout.PropertyField(uEvent, new GUIContent(label.stringValue, tooltip.stringValue));
```
Ecnassianer: We should probably display differently if we can't edit events. At the very least, there should be a label that they're readonly.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/InspectorFields/InspectorField.cs
killerantz: You are correct, I built it out that way to figure out how many different properties I would need for the enum and if there were any special cases which would require a switch statement. The event was the only thing I found so far that did not work as planned. I thought about condensing it during my clean up last week and feel better about it now after getting your input.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/States/States.cs
killerantz: This is the code change that fixed the errors building for .NET. The older code felt more robust, especially type.IsSubclassOf(typeof(InteractableStates)). Right now I'm only a level deep so (info.BaseType.Equals(typeof(InteractableStates)) || type.Equals(typeof(InteractableStates))) is working fine, but if things are extended more should I do a for loop until info.BaseType == null? I'm willing to delete it though, just wanted more testing and feedback first.

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Interactable/InspectorBase.cs
killerantz: Very good call. I'll get that cleaned up.

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Interactable/InspectorBase.cs
killerantz: Events are a weird situation, there was talk on an earlier PR about removing UnityEvents and replace them with a pattern being used in other features of MRTK. I haven't got around to looking at that yet. I can definitely clean this up for now.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonSize.cs
killerantz: Thinking about this further, the component does not have to use a ButtonSize element as an Anchor, so without that requirement there is no way to get a BasePixelScale if it's not set on this object. It should just be 2048, it works best for designers building UI from redlines, but if someone wanted to throw it on something for a quick layout and didn't care about pixels, then they can set it to whatever works for them.

killerantz: This code is being updated in PR #2870.

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Interactable/InspectorBase.cs
StephenHodgson: I'd really like to do that I'd we can. Hopefully I'll get some time to do a deep dive and open a PR with any suggestions there.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonSize.cs
killerantz: #2870 should go in first. I moved the code that was in #2852 (1 of 4) and moved it into #2870 (1.5 of 4), which is probably confusing. Each PR layers on top of the previous PRs because of their dependencies otherwise I can't test any changes in the later branches. I figured once one branch is through I would merge dev into the next branch is probably the cleanest path. It gets dicey though because any code changes (other than added comments) should get merged into my later branches for testing, but I'm worried about conflicts at some point.

Assets/MixedRealityToolkit-SDK/Features/UX/Interactable/Scripts/Layout/ButtonLayout.cs
killerantz: It's more than just code, it's a workflow. I like the camera settings idea, but it seems too detacted from Interactables. Maybe changing the name of these files would reposition as more than just button related code, which could make more sense in a camera resolution type setting. This layout code is more about 3D layout in general then just button code anyway, with similar functionality to RectTransforms, but in 3D. I just crammed it in here because I use it for buttons and this was the first enty point. It should be rethought. We could call it Layout (ObjectSize, ObjectAlignment, ObjectSizeOffset, ColliderCopy, ObjectBorder)

Assets/MixedRealityToolkit-SDK/Inspectors/UX/Interactable/InspectorBase.cs
killerantz: I removed the comments in this section of code. Should I resolve this because there are UnityEvent discussions going on elsewhere?

killerantz: The only pattern I've seen is more of a callback system so developers could extend a class and handle events in their specific code. That's definitely doable and a preferred method for doing this in code. There is another thread about using UnityEvents in the inspector only so designers could wire things up if they need to, which is super important because a lot of this is for designers so they can create nice buttons, but they have to be able to test them as well without code. I'll start looking into that. This code is only about what shows up in the inspector, so I will dig into the developer side soon.

PR #2855, opened 2018-09-28
Interactables complete 4 of 4

killerantz:
Overview
---

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit-SDK/Inspectors/UX/Interactable/InspectorBase.cs
```diff
@@ -631,10 +633,7 @@ public static void DisplayPropertyField(SerializedProperty prop)
                     break;
                 case InspectorField.FieldTypes.Event:
                     SerializedProperty uEvent = prop.FindPropertyRelative("EventValue");
-                    if (uEvent != null)
-                    {
-                        //Debug.Log("update? " + uEvent);
-                    }
+                    // Events do not work properly as a custom editor property, they do not get instanced properly
```
StephenHodgson: I'd like to get rid of the unity events entirely if possible, but keep the simple editor assignments

killerantz: We need a way for designers to wire things up in the editor without code. UnityEvents are the only way I know to do this, but I'm open to learning something new :) I looked around the input system and couldn't find anything. Is there something specific I can look at? I could build a better system for handling events for developers that don't need the editor, but only expose UnityEvents in the inspector. If they are not used, will they still cause issues? If they are used, then a designer has used them to wire things up and are opting in.

StephenHodgson: >We need a way for designers to wire things up in the editor without code. 

100% yes. We will definitely have a system that allows you to do very similar things.

There's a few concerns with Unity events:

- They're really hard to set and use in code in a dynamic way (like the lines below this thread)
- You can only set a single parameter in the editor
- When debugging them, it's hard to trace the full stack through the entire input system
- They have considerable overhead and performance costs
- If their static serialized references get broken they silently fail.

StephenHodgson: That's a good point.

@killerantz have you set up tasks for your interactable features yet for us to discuss requirements and acceptance criteria?

killerantz: That is a good point. With all the discussions and PRs it would help with prioritizing.

davidkline-ms: @StephenHodgson , I believe @keveleigh was asking for documentation re: "We will definitely have a system that allows you to do very similar things."

As for this PR, I do not believe we need to wholesale change @killerantz 's implementation for something that does not yet exist.  We are on a tight schedule and should not be expecting everything to be completely rewritten (not everything needs to be).

StephenHodgson: I'm not saying we should rewrite things, I'm just saying we should discuss implementations before adding it. 

Besides, from the comment on the line, it seems that it doesn't really work anyway?
So finding an alternative is preferable. Especially one that takes into account the issues I brought up above.

PR #2859, opened 2018-09-28
Add interface guidelines

ryzngard:
The coding guidelines don't currently specify interface design recommendations. This adds the recommendation that interface types should avoid concrete types where possible.

### DIFF
--- a/CodingGuidelines.md
+++ b/CodingGuidelines.md
@@ -240,6 +240,26 @@ public class MyClass
 }
  ```
 
+## Public Interface Definitions Should Avoid Concrete Types Where Possible.
+
+Interfaces are designed to provide a public contract devoid of implementation. Where possible, avoid assuming a non-primitive type. 
+
+### Don't:
+```
+public interface IFoo
+{
+    Dictionary<string, object> Objects { get; set; }
+}
+```
+
+### Do:
+```
+public interface IFoo
+{
+    IDictionary<string, object> Objects { get; set; }
+}
+```
+
 ## Initilize Enums.
 
 To ensure all Enum's are initialized correctly starting at 0, .NET gives you a tidy shortcut to automatically initilize the enum by just adding the first (starter) value.
--- a/CodingGuidelines.md
+++ b/CodingGuidelines.md
@@ -120,6 +120,10 @@ protected string MyProperty;
 private string myProperty;
  ```
 
+## Interface Naming Conventions
+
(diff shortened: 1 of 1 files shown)


StephenHodgson:
This is a great idea. 

Maybe we should also add that it should be prefixed with `IMixedReality`? (haha, this is mostly for @SimonDarksideJ )

CodingGuidelines.md
```diff
@@ -240,6 +240,26 @@ public class MyClass
...
+```
+public interface IFoo
+{
+    Dictionary<string, object> Objects { get; set; }
+}
+```
+
+### Do:
+```
+public interface IFoo
```
StephenHodgson: `public interface IMixedRealityFoo`

CodingGuidelines.md
```diff
@@ -240,6 +240,26 @@ public class MyClass
...
+{
+    IDictionary<string, object> Objects { get; set; }
```
StephenHodgson: Let's also add examples of other interfaces that are defined in the framework like `IMixedRealityController` or some other.

Also, I think only being able to get a collection is good, do maybe the example should be a getter only? We probably don't want a collection to be completely changed outside of the class implementation.

ryzngard: We maybe should add more examples about that. I think in general that should be true, but sometimes it's better (and more efficient) to completely change the underlying object instead of modifying the existing. 

Maybe another recommendation about that?

StephenHodgson: I'm okay with adding or removing an item from a collection, but I'm not as keen on replacing the whole collection with a new instance.

davidkline-ms: >  We probably don't want a collection to be completely changed outside of the class implementation.

Even if we do limit to get only collection properties, devs can change it outside of our control... unless we return them a copy of our collections (less memory efficient, but safer)

StephenHodgson: I'm fine with that as long as we're not requesting copies in hot code loops

ryzngard: I think we're getting a little sidetracked on the intent :) I'll make the example get only, but not put an explicit guideline for whether a collection should be replaceable at runtime. That's an implementation detail that might be unique to any given interface, and isn't really a coding guideline as much as an architecture guidance.

ryzngard:
> Maybe we should also add that it should be prefixed with IMixedReality? (haha, this is mostly for @SimonDarksideJ )

And so it is done 😁

Issue #2864, opened 2018-10-01
Default diagnostics profile needs to be locked


davidkline-ms:
Mixed Reality Toolkit

**Describe the bug**
The default diagnostics profile needs to be marked as a default (custom == 0)

**Expected behavior**
The default diagnostics profile is locked

**Actual behavior**
The default diagnostics profile is editable

StephenHodgson:
We should probably update the interface to be system as well to be named similar to the other system interfaces

davidkline-ms:
> We should probably update the interface to be system as well to be named similar to the other system interfaces

That should be it's own issue. Agree?

StephenHodgson:
Sounds good, Just something I noticed but didn't have time to write up.

davidkline-ms:
This profile is still unlocked... It probably needs the inspector to be updated to disable the UI

keveleigh:
![image](https://user-images.githubusercontent.com/3580640/46381189-12d36400-c65a-11e8-9564-0e68a7200b8d.png)

It's locked for me. Your editor might be caching some state.

StephenHodgson:
Be sure to check your preferences. Maybe you have it unlocked there.

Issue #2863, opened 2018-10-01
Diagnostics profile currently contains a title and help text for the Boundary system


SimonDarksideJ:
# Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?
Mixed Reality Toolkit

# Describe the bug
#2849 added a new configuration screen for managing the Diagnostics configuration.  However as you can see below, the Boundary config screen was copied but not updated for Diagnostics.
![image](https://user-images.githubusercontent.com/1793042/46297463-a17dad80-c594-11e8-9f4a-bbf8d60aaae7.png)
The Header and help text is still the boundary text.

# To Reproduce
Open the Diagnostics Default profile
Use eyes

**Expected behavior**
Header and text should relate to the Diagnostics config

**Actual behavior**
Header and text are duplicated from the Boundary config

**Screenshots**

**Unity Editor Version**
Which Unity version(s) you've been able to reproduce the bug in.

**Mixed Reality Toolkit Release Version**
Which MRTK version(s) you've been able to reproduce the bug in.

**Additional context**

SimonDarksideJ:
Resolved by #2869

PR #2869, opened 2018-10-01
a few more diagnostics changes

StephenHodgson:
Overview
---
Found a missing namespace reference (Current dev branch doesn't compile)
Fixed profile locking.

- Fixes: #2864

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityDiagnosticsSystemProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityDiagnosticsSystemProfileInspector.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.﻿
 
+using Microsoft.MixedReality.Toolkit.Core.Definitions;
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityDiagnosticsSystemProfileInspector.cs
```diff
@@ -42,6 +43,11 @@ public override void OnInspectorGUI()
                 Selection.activeObject = MixedRealityManager.Instance.ActiveProfile;
             }
 
+            if (MixedRealityPreferences.LockProfiles && !((BaseMixedRealityProfile)target).IsCustomProfile)
```
keveleigh: Should this go in the `MixedRealityBaseConfigurationProfileInspector` so we don't have to add it to each individual inspector?

keveleigh: Ah, I see what you mean about the profile button. If there's a way to do it in the base, that'd be preferable, but it works as-is and that can be addressed later (and called out in the documentation).

StephenHodgson: I tried to put it in the base initially, but it's not trivial.

SimonDarksideJ:
Resolves #2863

PR #2870, opened 2018-10-02
Asset code for Interactables - 1.5 of 4

killerantz:
Overview
---
Moved code to a new branch and PR. Incorporated feedback from previous branch and PR.

Changes
---
Added summary tags
Changed some foreach loops
Simplified propertySettings to remove the switch statement
- Fixes: # .


killerantz:
The ButtonSize components can be used for things beyond buttons. I use them for building UI panels and more advanced elements. I'm open to name changes if you have any suggestions. But then we would want to move it out of Interactables with it still being a dependency.

SimonDarksideJ:
This needs to target the Interactables Branch, not dev I believe

killerantz:
@SimonDarksideJ Should I create a new Interactables branch for all these PRs to go into, then we can move everything over to dev? I created this branch by copying the dev/interactive branch and removed all the other files. If this merges back into dev/interactive, would that remove all the files missing in this branch? I'm primarily a designer so I may need some extra guidance on how to get everything into the right place. Thanks for your patience and reviews.

SimonDarksideJ:
If we can get the new interactables feature working in the dev/Interactables branch, squash the commits and then have PR's bringing over each chunk.
But it must be finished / working in there before bringing over ideally.

@StephenHodgson points about UnityEvents are quite critical so it fits properly within the vNext architecture and doesn't duplicate features.  I think we're more than happy to pitch in on the Interactables branch, using Tasks to identify things that need updating.

PR #2872, opened 2018-10-02
Add QRTracking assets and samples in Holotoolkit-preview

chgatla-microsoft:
Overview
---Introducing QR tracking, platform driver implements the QR tracking and this sample shows how to use QR tracking.

Changes
---
- Fixes: # .


Assets/HoloToolkit-Preview/QRTracker/Plugins/QRCodesTrackerPlugin.Placeholder.dll.mdb.meta
```diff
@@ -0,0 +1,7 @@
+fileFormatVersion: 2
+guid: 9b91909175255dd40bbabf1f29de0b2e
```
StephenHodgson: Is this dll to be replaced by direct windows api calls?

Assets/HoloToolkit-Preview/QRTracker/Plugins/WSA/arm/QRCodesTrackerPlugin.dll.meta
```diff
@@ -0,0 +1,30 @@
+fileFormatVersion: 2
+guid: 980931158bfdfdd47ab8de3098f68f38
```
StephenHodgson: Is the source for this dll availible or in the native MRTK repo?

Or is this something that will be replaced by native windows api calls?

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using QRCodesTrackerPlugin;
+
+namespace HoloToolkit.Unity.QRTracking
+{
+    public class AttachToQRCode : MonoBehaviour
+    {
+        [Tooltip("Object that need to be attached to the QR code.")]
+        public GameObject gameObjectToAttach;
```
StephenHodgson: private fields with property accessors.

Will we be able to update this game object after we've attached?
If we set a new game object, do we need to do any additional processing?

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
```diff
@@ -0,0 +1,118 @@
...
+                    
+                    qrCodeId = QRCodesManager.Instance.GetIdForQRCode(qRCodeData);
+                    updatedId = true;
+                }
+            }
+        }
+
+        private System.Collections.Generic.SortedDictionary<System.Guid, GameObject> qrCodesObjectsList;
+
+        void Awake()
```
StephenHodgson: missing explicit accessor modifier `private`

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            QRCodesManager.Instance.QRCodeAdded += Instance_QRCodeAdded;
+            QRCodesManager.Instance.QRCodeUpdated += Instance_QRCodeUpdated;
+            QRCodesManager.Instance.QRCodeRemoved += Instance_QRCodeRemoved;
+        }
+
+        // Use this for initialization
+        void Start()
```
StephenHodgson: missing explicit accessor modifier `private`

We can also loose the boilerplate comment.

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            if (qrCodeId == e.Data.Id)
+            {
+                qrCodeId = new System.Guid();
+                updatedId = true;
+            }
+        }
+
+        // Update is called once per frame
+        void Update()
```
StephenHodgson: missing explicit accessor modifier `private`

We can also loose the boilerplate comment.

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        private bool updatedId = false;
+
+        private SpatialGraphCoordinateSystem coordSystem = null;
+        /// <summary>
+        /// Data of the QR code to which we want to attach the game object."
+        /// </summary>
+        [SerializeField]
+        [Tooltip("Data of the QR code to which we want to attach the game object.")]
+        private string qRCodeData = "";
```
ryzngard: nit: Should we use string.Empty as the default?

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
```diff
@@ -0,0 +1,98 @@
...
+            CodeText = "Dummy";
+            if (qrCode == null)
+            {
+                throw new System.Exception("QR Code Empty");
+            }
+
+            PhysicalSize = qrCode.PhysicalSizeMeters;
+            CodeText = qrCode.Code;
+
+            qrCodeCube = gameObject.transform.Find("Cube").gameObject;
```
StephenHodgson: Might be a better way to look for the transforms.

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
```diff
@@ -0,0 +1,118 @@
...
+                // default use the scripts object
+                gameObjectToAttach = gameObject;
+            }
+            qrCodeId = QRCodesManager.Instance.GetIdForQRCode(qRCodeData);
+            updatedId = true;
+        }
+
+        private void Instance_QRCodeAdded(object sender, QRCodeEventArgs<QRCodesTrackerPlugin.QRCode> e)
+        {
+            if (qrCodeId == new System.Guid())
```
ryzngard: This seems nefarious. Maybe use default(System.Guid) instead?

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
```diff
@@ -0,0 +1,98 @@
...
+
+            QRID.text = qrCode.Id.ToString();
+            QRText.text = CodeText;
+            QRVersion.text = "Ver: " + qrCode.Version;
+            QRSize.text = "Size:" + qrCode.PhysicalSizeMeters.ToString("F04") + "m";
+            QRTimeStamp.text = "Time:" + qrCode.LastDetectedQPCTicks;
+            Debug.Log("Id= " + qrCode.Id + " PhysicalSize = " + PhysicalSize + " TimeStamp = " + qrCode.LastDetectedQPCTicks + " QRVersion = " + qrCode.Version + " QRData = " + CodeText);
+        }
+
+        void UpdatePropertiesDisplay()
```
StephenHodgson: missing access modifier

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            // Update properties that change
+            if (qrCode != null && lastTimeStamp != qrCode.LastDetectedQPCTicks)
+            {
+                QRSize.text = "Size:" + qrCode.PhysicalSizeMeters.ToString("F04") + "m";
+#if true
```
StephenHodgson: Always runs?

What updates this?

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,160 @@
+﻿using System;
+using System.Collections;
+
+using System.Collections.Generic;
+
+using UnityEngine;
+
```
StephenHodgson: missing licence and could be cleaned up a bit.

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,160 @@
...
+    public static class QRCodeEventArgs
+    {
+        public static QRCodeEventArgs<TData> Create<TData>(TData data)
+        {
+            return new QRCodeEventArgs<TData>(data);
+        }
+    }
+
+    [Serializable]
+    public class QRCodeEventArgs<TData> : EventArgs
```
StephenHodgson: classes should be in their own files

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
```diff
@@ -0,0 +1,118 @@
...
+                    updatedId = true;
+                }
+            }
+        }
+
+        private void Instance_QRCodeRemoved(object sender, QRCodeEventArgs<QRCodesTrackerPlugin.QRCode> e)
+        {
+            if (qrCodeId == e.Data.Id)
+            {
+                qrCodeId = new System.Guid();
```
ryzngard: Guid.NewGuid()?

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
ryzngard: nit: I'd prefer a commented section of code over an always true/false pragma. At least the comment can say why it's not currently available

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                }
+#else
+                {
+                    QRTimeStamp.text = "Time:" + qrCode.LastDetectedQPCTicks;
+                }
+#endif
+
+                PhysicalSize = qrCode.PhysicalSizeMeters;
+                Debug.Log("Id= " + qrCode.Id + " PhysicalSize = " + PhysicalSize + " TimeStamp = " + qrCode.LastDetectedQPCTicks);
```
ryzngard: nit: I _think_ string interpolation does less allocations than string +

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        public bool IsTrackerRunning { get; private set; }
+        public QRCodesTrackerPlugin.QRTrackerStartResult StartResult { get; private set; }
+
+
+        public event EventHandler<QRCodeEventArgs<QRCodesTrackerPlugin.QRCode>> QRCodeAdded;
+        public event EventHandler<QRCodeEventArgs<QRCodesTrackerPlugin.QRCode>> QRCodeUpdated;
+        public event EventHandler<QRCodeEventArgs<QRCodesTrackerPlugin.QRCode>> QRCodeRemoved;
+
+        private System.Collections.Generic.SortedDictionary<System.Guid, QRCodesTrackerPlugin.QRCode> qrCodesList = new SortedDictionary<System.Guid, QRCodesTrackerPlugin.QRCode>();
```
ryzngard: I don't think you need `System.Collections.Generic` here

Assets/HoloToolkit-Preview/QRTracker/Plugins/QRCodesTrackerPlugin.Placeholder.dll.mdb.meta
davidkline-ms: eventually, when it is no longer a preview / experimental feature

Assets/HoloToolkit-Preview/QRTracker/Plugins/WSA/arm/QRCodesTrackerPlugin.dll.meta
davidkline-ms: it is currently not available. this is an experimental feature. when any final form is ready, the code will be updated

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,160 @@
...
+
+        private QRTracker qrTracker;
+
+        public System.Guid GetIdForQRCode(string qrCodeData)
+        {
+            lock (qrCodesList)
```
ryzngard: This is more of a meta discussion, but locking an object you're modifying can be tricky. Usually you want another lock object to be more explicit. I don't see any offending code here, but if we did something like
`lock (qrCodesList) { qrCodesList = new ... }` we'd hit problems

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
StephenHodgson: Old HTK does not use c# 4 language features. I think the .NET support is only 3.5

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,160 @@
...
+using System.Collections.Generic;
+
+using UnityEngine;
+
+using QRCodesTrackerPlugin;
+namespace HoloToolkit.Unity.QRTracking
+{
+    public static class QRCodeEventArgs
+    {
+        public static QRCodeEventArgs<TData> Create<TData>(TData data)
```
ryzngard: Why do we have an extra layer instead of just a new directly?

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,160 @@
...
+            }
+        }
+
+        private void QrTracker_Added(QRCodeAddedEventArgs args)
+        {
+            lock (qrCodesList)
+            {
+                qrCodesList.Add(args.Code.Id, args.Code);
+            }
+            var handlers = QRCodeAdded;
```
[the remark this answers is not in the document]
StephenHodgson: We don't have those c# language features availible in the old HTK

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
StephenHodgson: Agreed, I think the async utilities doesn't play nicely with the lock as well bc of the way [Unity's synchronization context is implemented](https://github.com/svermeulen/Unity3dAsyncAwaitUtil/issues/10) as well. (But we won't need to worry about that until we bring this over to MRTK and if we make this method asynchronous).

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesVisualizer.cs
```diff
@@ -0,0 +1,109 @@
...
+using UnityEngine;
+
+using QRCodesTrackerPlugin;
+namespace HoloToolkit.Unity.QRTracking
+{
+    public class QRCodesVisualizer : MonoBehaviour
+    {
+        public GameObject qrCodePrefab;
+
+        private System.Collections.Generic.SortedDictionary<System.Guid, GameObject> qrCodesObjectsList;
```
ryzngard: `SortedDictionary` is fine with the usings above

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesVisualizer.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            if (qrCodePrefab == null)
+            {
+                throw new System.Exception("Prefab not assigned");
+            }
+        }
+
+        private void Instance_QRCodeAdded(object sender, QRCodeEventArgs<QRCodesTrackerPlugin.QRCode> e)
+        {
+            lock (pendingActions)
```
ryzngard: Can we just use a `ConcurrentQueue`?

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesVisualizer.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        private void HandleEvents()
+        {
+            lock (pendingActions)
+            {
+                while (pendingActions.Count > 0)
+                {
+                    var action = pendingActions.Dequeue();
+                    if (action.type == ActionData.Type.Added)
```
ryzngard: nit: With enums, I think switch works better. It makes it easier to understand the flow for us mere mortals

Assets/HoloToolkit-Preview/QRTracker/Scripts/SpatialGraphCoordinateSystem.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+#if WINDOWS_UWP
+using Windows.Perception.Spatial;
+#endif
+namespace HoloToolkit.Unity
+{
+    public class SpatialGraphCoordinateSystem : MonoBehaviour
+    {
+#if WINDOWS_UWP
```
ryzngard: This class behaves so different for UWP and non-UWP. Is there a more clever way to do this without pragmas everywhere?

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
ryzngard: 🤮

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
ryzngard: Ah! I missed that this was for HTK only

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
ryzngard: The only downside to using the concurrent queue is that it technically has different behavior. Here, and in the issue you linked, the queue gets drained all at once before allowing more work to be added. We could get in a state where the thread locks waiting for the queue to drain while continually adding work. 

I wonder if a CopyTo would be the best approach if we wanted to snapshot the queue per execution call...

Queue also has a [SyncRoot](https://docs.microsoft.com/en-us/dotnet/api/system.collections.queue.syncroot?view=netframework-4.7.2), but I'm not sure that's really any better. It at least would work if someone used the Queue as a [Synchronized](https://docs.microsoft.com/en-us/dotnet/api/system.collections.queue.synchronized?view=netframework-4.7.2)

StephenHodgson: >We could get in a state where the thread locks waiting for the queue to drain while continually adding work.

That's exactly how I ended up fixing it in my own synchronization context implementation.

StephenHodgson:
```

        internal static void Post(Action task)
        {
            lock (_Actions)
            {
                _Actions.Enqueue(task);
            }
        }


        private void Update()
        {
            int actionCount;

            lock (_Actions)
            {
                actionCount = _Actions.Count;
            }

            for (int i = 0; i < actionCount; i++)
            {
                Action next;

                lock (_Actions)
                {
                    next = _Actions.Dequeue();
                }

                next();
            }
        }
```

Yoyozilla:
This is an experimental feature that just shipped with RS5 today, this pull request contains the unity plugin required to access this API (not MRTK dependent) and a simple sample to showcase this API.  We would like to give developer access of this feature ASAP. That said, we still want to make sure it is performance and reasonably sane, since developers are likely going to use this in their app... 

If we could fix the functional issues now and fixed the other "better this way" issues when we move this feature into vNext that would be appreciated! This is not meant to be the final version and we should iterate on it given the community feedback later.

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
xwipeoutx: `Guid.Empty` is the best way to do this

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
SimonDarksideJ: Agreed, should really be assigned to a property for assigning in the editor.

Assets/HoloToolkit-Preview/QRTracker/Plugins/QRCodesTrackerPlugin.Placeholder.dll.mdb.meta
chgatla-microsoft: This API is do nothing for non UWP and Editor. Unity links against this dll for that purpose.

Assets/HoloToolkit-Preview/QRTracker/Plugins/WSA/arm/QRCodesTrackerPlugin.dll.meta
chgatla-microsoft: This dll will be copied from windows latest builds ( they are built in RS6 windows builds) to the MRTK on very slow cadence ( any bugs).

SimonDarksideJ:
Sounds good @Yoyozilla , so long as we stay true to that.  It should be architected properly to introduce it to vNext once it's "working" in the HTK dev.

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
chgatla-microsoft: It is templated used for events.

Assets/HoloToolkit-Preview/QRTracker/Scripts/SpatialGraphCoordinateSystem.cs
chgatla-microsoft: This class is do-nothing for non UWP. Let me know if we have a better way to achieve that..

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
chgatla-microsoft: This script is example visualization just for a particular prefab. Not intended to be used for any other objects.

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCode.cs
chgatla-microsoft: qrCode properties change when new detection happens

Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
davidkline-ms: not 100% required for a preview feature in the HTK

Assets/HoloToolkit-Preview/QRTracker/Scripts/QRCodesManager.cs
davidkline-ms: not 100% required for a preview feature in the HTK

davidkline-ms:
for the first checkin (preview, HTK), let's resolve the lock() and ConcurrentQueue issues as well as adding license header to all files

davidkline-ms:
> It should be architected properly to introduce it to vNext once it's "working" in the HTK dev.

Before this leaves preview/experimental status, feature architecture / code cleanliness / etc will all be performed.

StephenHodgson:
I think at the minimum we should add the license headers.

keveleigh:
>I think at the minimum we should add the license headers.

Those have been added: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2872/commits/9db75ec99cf7f3de3f63f0883898f28f9afaa661

PR #2885, opened 2018-10-05
[Solver PR 5] Add Solver example scene

keveleigh:
Overview
---
Adds the Solver example scene.

To test, point and click on any of the spheres in the scene with a motion controller. The solver should attach to your controller and behave appropriately.

Changes
---
Related to #2465


Ecnassianer:
After spending time in the newer scenes, going back to this one made me realize there are zero instructions, and without the controller models as an affordance, it feels hard to figure out how I activate the solvers. Would you mind adding one of those instruction cards from the other scene? Most just instructions like "Use the motion controllers to point at a solver display and pull the trigger to activate that solver example. Aim at the sphere and click again to deactivate that example."

You could elaborate on each example, but I don't think that's strictly necessary.

I tested out all the solvers and they seem to be working fine. This color scheme/lighting is much improved, it helps see the steps a lot better.

keveleigh:
Sounds good! I'll add an instructions panel.

SimonDarksideJ:
Can we also add a screenshot of the scene to the overview of this PR, so we have a frame of reference :D

Ecnassianer:
>The magnetism examples seem to be locked to the short edge of the object. Would make more sense to be the large flat side, as this is a more common use. (also seems to flip between short and long on the stairs?)

I choose this shape to illustrate the normal of the surface. It was unclear to me when I first started using these how they were choosing rotation. The thin end of the solver is the dimension that seems clearest in every orientation, particularly at the tops and edges of the sphere, where the width-wide arrangement was too thin to clearly illustrate the relationship to the surface. While I agree the large flat side is more common, I don't think it is as useful to illustrate their behavior.

> The current "Window following" style of body lock is great (although a plane might be better than a capsule), but can the body lock also have a style that is attached/fixed to the players body/waist as I've seen this used many times in projects.

I'm not sure I understand the behavior you're asking for. Do you mean a version that turns off the Tether Angle Steps, so it always stays exactly in front of the observer instead of jumping between positions?

keveleigh:
>A completely dark scene is a bit to embolding. Might be better with something akin to the Example 3 branch, a living demo scene (albeit still basic)

I added the skybox used in other demo scenes.

>The different solver spheres (A think this is a much better demo than the one in the HTK) need to be arranged within a 110 FOV. A dev should not need to turn 360 degrees to act on them all.

I rearranged the spheres to be in front of the origin

>The magnetism examples seem to be locked to the short edge of the object. Would make more sense to be the large flat side, as this is a more common use. (also seems to flip between short and long on the stairs?)

This is by-design, as stated above by @Ecnassianer.

>Agree with other comments, does need the billboard example description to explain the scene. I was confused about the pickup / put down mechanism at first.

Added.

>Should only one lock style at a time ne in use, as I was able to "pick up" multiple different styles at a time.

This is by-design, as it can be useful to experience multiple solved objects at the same time.

>The current "Window following" style of body lock is great (although a plane might be better than a capsule), but can the body lock also have a style that is attached/fixed to the players body/waist as I've seen this used many times in projects.

I added a second orbital prefab and turned off the angle stepping feature.

>Do we want tooltips in the example scripts?

Added.

keveleigh:
@SimonDarksideJ Are you sure you pulled the latest changes? That looks no different from my original submission.

keveleigh:
The floor should be spawned from the boundary system and existed in my testing. I've noticed that sometimes Unity doesn't provide the boundary information in time for app launch when running in-editor (there's typically a warning in the console), so you might be seeing that.

SimonDarksideJ:
The boundary will provide a floor guide, but I wouldn't use that as the floor, especially if you use it in Sitting mode (which I do)

keveleigh:
The boundary system spawns a floor quad when enabled in the profile, [which it is](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityBoundaryVisualizationProfile.asset#L16) in the default profile.

Even if you use it while sitting, the quad will still align with the floor. I also often run apps while sitting.

SimonDarksideJ:
It's just my preference, I've always supplied the floor with the boundary as a guide.
Leaving that, it's just the background to the text above the spheres left.

davidkline-ms:
> Even if you use it while sitting, the quad will still align with the floor. I also often run apps while sitting.

Only in room scale mode if a boundary is configured on the platform. We should probably file an issue against the boundary system to generate a floor if there is no actual boundary configured.

davidkline-ms:
> Leaving that, it's just the background to the text above the spheres left.

That should be any easy fix. Since @keveleigh is out, let's accept this PR and file an issue to fix the background of the text (that should be an easy change). 

Agreed?

davidkline-ms:
Just noticed that Kurtis gave me access to his repository... I can make the change inline with this PR

davidkline-ms:
@SimonDarksideJ, I am not sure what you mean by "disjointedness".

SimonDarksideJ:
They aren't straight!
* Some Spheres / boards are turned
* Some sphere's / boards are forward or back
* Some Sphere's / boards are on different levels.

Can we put them in an object collection or something, or at the very least, line them up?
I'll try and do a PR but I'm stuck with 15 other things atm

davidkline-ms:
> Can we put them in an object collection or something, or at the very least, line them up?

I'll take a look. I believe @keveleigh was manually arcing them.

PR #2887, opened 2018-10-05
Remove non-events from IMixedRealityFocusHandler

keveleigh:
Overview
---
This is the only handler interface that contains more than the event handlers. `IMixedRealityFocusHandler` contains three additional properties which added unnecessary overhead for implementing the focus events and are implementation-specific.

If a developer needs access to this information, they can inherit from `BaseFocusHandler`. Either way, we don't pass `I[*]Handler` references around, so there's no need to define them there, and not all interface implementations will need them.


StephenHodgson:
@Railboy how do you feel about this change?

> unnecessary overhead

Haha wasn't the point of defining them in the interface, meaning that they were necessary? 😅 

>there's no need to define them there, and not all interface implementations will need them

Yup that makes sense to me.

PR #2897, opened 2018-10-08
fixed a case where the MixedRealityPlayspace could return null

Ecnassianer:


Ecnassianer:
I'm not sure what disabling the enabling the input system does, so I'm not sure.

The case I found was if you have a GameObject named MixedRealityPlayspace as the camera's parent, it checks to make sure we can use that, but then never actually returns it when you ask for it.

StephenHodgson:
> I'm not sure what disabling the enabling the input system does, so I'm not sure.

Then why did you change how the whole setup worked? 😅 (It's too late now I suppose)

PR #2903, opened 2018-10-09
Overlap Solver

Ecnassianer:
I keep finding myself recreating this solver in all my projects, so I figured maybe someone else would like it in the toolkit. You can get the same behavior from the right settings in other solvers, but this simplest form takes potential mistakes out of debugging, so I end up using it a lot.

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/Overlap.cs.meta
+++ b/Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/Overlap.cs.meta
@@ -5,7 +5,7 @@ MonoImporter:
   serializedVersion: 2
   defaultReferences: []
   executionOrder: 0
-  icon: {instanceID: 0}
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit-SDK/Features/Utilities/Solvers/Overlap.cs.meta
```diff
@@ -0,0 +1,11 @@
+fileFormatVersion: 2
+guid: 97026b56b4e804b468b8d53af13e3c3b
+MonoImporter:
+  externalObjects: {}
+  serializedVersion: 2
+  defaultReferences: []
+  executionOrder: 0
+  icon: {instanceID: 0}
```
StephenHodgson: Be sure to set an Icon :)

Ecnassianer: Why do we override the default script icon?

Issue #2888, opened 2018-10-06
Rethink Versioning Of MRTK: Allow users to know risk of ingesting packages


ryzngard:
While working on a solution to package MRTK in NuGet, versioning has come up a lot. 

There's 2 major points to bring up for this discussion:

1. We should document versioning in a more formal way; that way our community and consumers can all easily reference and understand what a particular version means. 

2. I believe our version should follow a different paradigm than currently.

The current version, from #1858 , should be [product].[month].[revision]. I propose we move to [product].[release].[revision].

Defining the terms as below: 

[product]: Our canonical product name. We have agreed on yearly releases and using year as the nomenclature. For MRTK, the first product name will be 2018. We also tie these to milestones 

[release]: A version release that adds new features. May be a breaking change, if documented, from a previous release on the same [product]. May also just add new features 

[revision]: Intended not to be a breaking change nor add or remove functionality. This would be bug fix releases or patching. Same as [patch] in SemVer

The important difference here is that we are encoding the usability of the package in the versioning, which closer aligns to the intent of SemVer. We are sacrificing the [major].[minor] granularity and collapsing them into a [release] version. This lets customers know that there is an inherent risk in upgrading from 18.1 to 18.2, for example. When using [month] instead of [release] we lose the ability to convey this information in order to communicate a time. I believe that we should avoid this, as release times are encapsulated with package and distribution information. 

I would love to hear you thoughts on this. I've discussed heavily with @davidkline-ms in person (and somewhat heatedly) to arrive to a small revision to our plan that would have a big impact. This also means that if we release at a cadence that is not monthly, our versioning still supports that.

Ecnassianer:
I'm a fan of using risk as the primary factor in version revving.

hridpath:
I don't mind conveying a risk via versioning. I do believe that risk should be well documented for each release. However, making it very clear that there are breaking changed to the reader is very important.
Making the realeas a value starting at 00 and incrementing by 10. However when there is a breaking change the one's position would contain a value other than 0.
I would like to see somthing 2018.21.1005 like for any release that has a breaking change  

Just a thought.

davidkline-ms:
> (and somewhat heatedly) 

C'mon... it didn't get that warm :)

StephenHodgson:
Haha, I think we've already changed it 4 times 😅

davidkline-ms:
> Haha, I think we've already changed it 4 times 😅

[...]

PR #2908, opened 2018-10-10
Update roadmap for new version scheme

ryzngard:
As discussed in #2888

### DIFF
--- a/Roadmap.md
+++ b/Roadmap.md
@@ -30,7 +30,7 @@ As needed, patch and hot fix releases will be released between the scheduled cad
 
 Starting with the May 2018 release, the MRTK version numbers are adopting [Semantic versioning](https://semver.org/) rather than aligning with a specific Unity release.
 
-The MRTK will use the first two version fields (2017.4) to indicate the recommended version is the Unity 2017 LTS version. The third field will indicate the MRTK revision number (ex: 2017.4.2). The fourth will always be zero (2017.4.2.0).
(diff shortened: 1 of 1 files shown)


StephenHodgson:
Couldn't we just use it as the unity targeting version?

For example:

Unitys version is 2018.1.9f1
MRTK version is 2018.1.[release].[revision]

ryzngard:
NuGet only supports 3 numbers, as is specified by SemVer. I'm not sure why we have 4. 

If we are strictly following SemVer, we can encapsulate Unity version as [prod].[release].[revision]-Unity2018.1

ryzngard:
[NuGet Versioning](https://docs.microsoft.com/en-us/nuget/reference/package-versioning) does have one interesting part but isn't documented well. 

```
1.00 is treated as 1.0
1.01.1 is treated as 1.1.1
1.00.0.1 is treated as 1.0.0.1
```

That seems to indicate that 4 numbers might be supported, but isn't specified in SemVer.

davidkline-ms:
Let's stick to 3 version components as it is the most commonly used and is more compatible with other systems (ex: Windows Store)

davidkline-ms:
>Couldn't we just use it as the unity targeting version?
For example:
Unitys version is 2018.1.9f1
MRTK version is 2018.1.[release].[revision]

I'd rather not as we will eventually lock on the Unity LTS release (ex: 2018.4) which would result in an essentially unused parameter (and that is a step quite a bit backward to the HTK versioning scheme).

davidkline-ms:
> Why is the fourth always zero?
I'm in favor of dropping it but didn't explicitly bring that up for consensus.

I'm fine with us only specifying 3. It's not like we are using a VersionInfo struct :)

StephenHodgson:
> It's not like we are using a VersionInfo struct 

Actually we do, but that's in the build window ;)

Issue #2834, opened 2018-09-25
MRTK vNext SpeechInputHandler NullReference


alcooper91:
I'm attempting to utilize MRTK vNext to add some speech commands to an app, but I'm hitting some issues editing my scene and then had a question about usage.  Sorry if this isn't the right place for posting, but I didn't see any docs around what I'm attempting either.

Scenario:
I'm attempting to add speech to my app, I created a Custom MixedRealityConfigurationProfile that matches the default except for a custom InputSystemProfile, that matches the default except for a custom SpeechProfile to add a few voice commands.  When I first add new VoiceCommands, I'm able to interact with the SpeechInputHandler to register new commands; but when I go back in and edit the script/functions that are being targeted by that command, I end up being told that the Handler thinks there are no keywords registered, and viewing the Handler causes the following NullReference to appear in the output:
NullReferenceException: Unable to find IMixedRealitySpeechSystem.
Microsoft.MixedReality.Toolkit.Core.Managers.MixedRealityManager.GetManager (System.Type type) (at Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs:502)
Microsoft.MixedReality.Toolkit.Core.Managers.MixedRealityManager.GetManager[T] () (at Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs:473)
Microsoft.MixedReality.Toolkit.Core.Definitions.InputSystem.MixedRealityInputSystemProfile.get_SpeechSystem () (at <5633ad60a2174c55802b45e61d12df29>:0)
Microsoft.MixedReality.Toolkit.Core.Definitions.InputSystem.MixedRealityInputSystemProfile.get_IsSpeechCommandsEnabled () (at <5633ad60a2174c55802b45e61d12df29>:0)
Microsoft.MixedReality.Toolkit.SDK.Inspectors.Input.Handlers.SpeechInputHandlerInspector+<RegisteredKeywords>c__Iterator0.MoveNext () (at Assets/MixedRealityToolkit-SDK/Inspectors/Input/Handlers/SpeechInputHandlerInspector.cs:141)
System.Linq.Set`1[TElement].UnionWith (System.Collections.Generic.IEnumerable`1[T] other) (at <839a3cb835c04d14aeb58d83bb7bc4bd>:0)
System.Linq.Enumerable+DistinctIterator`1[TSource].FillSet () (at <839a3cb835c04d14aeb58d83bb7bc4bd>:0)
[...]

PR #2913, opened 2018-10-12
Fixed a few issues with speech input handler

StephenHodgson:
Overview
---
- Made sure Windows Speech Input's event data was raised by the gaze's input source, as Focused Object is determined by the event's input source.
- Updated `SpeechInputHandler` to only check if speech commands profile was null instead of if the speech commands were enabled.
- Added an error helpbox to the speech input handler if a previously registered keyword is no longer in the speech command profile

Changes
---
- Fixes: #2834


StephenHodgson:
I'm not really liking the way I fixed the speech commands by replacing the InputSystem with the gaze directly like that.

In out generic input source creation we set the input source's pointer to be the gaze pointer if no pointers are assigned.

In the Mixed Reality Input Manager's event handler method, we should be getting the focused object based on the event data, which should give us whatever object the gaze pointer is currently focused on, but it's not.

Going to look into and test that first.

StephenHodgson:
I removed the changes to the WindowsSpeechInputDeviceManager and will address the focus bug in a separate PR

StephenHodgson:
@alcooper91 can you test this and make sure it fixes the Null Ref you were running into please?

alcooper91:
I won’t be able to check until Monday but I’ll let you know then.

StephenHodgson:
No worries :) much appreciated.

StephenHodgson:
image

StephenHodgson:
Thanks @keveleigh!

PR #2914, opened 2018-10-12
Added _HoverLightRadius for hover light radius

quakeulf:
Greetings,
I have added _HoverLightRadius to the shader to allow for finer radius control of the hover light. I have also made the gradient exponential instead of linear, so it now more closely matches the hover light effect elsewhere in the HoloLens OS. 

IMPORTANT:

This also needs the edits I have done in StandardShaderGUI.cs to show in the editor: #2915

Overview
---
It is a float in the shader that gives the user the possibility to change the radius of the hover light.

Changes
---
- Needs: #2915

### DIFF
--- a/Assets/HoloToolkit/Common/Shaders/Standard.shader
+++ b/Assets/HoloToolkit/Common/Shaders/Standard.shader
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 Shader "MixedRealityToolkit/Standard"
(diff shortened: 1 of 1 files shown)


davidkline-ms:
re-targeted this change to htk_development (code does not go directly into master)… thx

Assets/HoloToolkit/Common/Shaders/Standard.shader
```diff
@@ -261,6 +262,7 @@ Shader "MixedRealityToolkit/Standard"
 #endif
 
 #if defined(_HOVER_LIGHT)
+			fixed _HoverLightRadius;
```
Cameron-Micka: @quakeulf thank you for taking a look at this. I love the idea of making the gradient exponential. I am a bit curious about why you needed to add this since the radius can be controlled on the HoverLight component (it is packed into the 'w' component of the hover light data). Is this so the radius can be controlled on a per material basis?

quakeulf: Well, I wanted more options on the shader itself, and I per-material is a bonus I didn't think of. That's about it I can think of for reasons right now.

Assets/HoloToolkit/Common/Shaders/Standard.shader
```diff
@@ -522,7 +524,7 @@ Shader "MixedRealityToolkit/Standard"
                 {
                     int dataIndex = lightIndex * HOVER_LIGHT_DATA_SIZE;
                     fixed hoverValue = HoverLight(_HoverLightData[dataIndex], i.worldPosition.xyz, _HoverLightData[dataIndex + 1].a);
-                    pointToHover += hoverValue;
+                    pointToHover += (hoverValue * hoverValue);
```
Cameron-Micka: After reviewing this I think a good addition would be to make this feature "opt in." Much like how _HOVER_COLOR_OVERRIDE is done. That way existing materials will not change automatically on users when they get this change. Then I would change the shader parameter to something like _HoverLightRadiusOverride.

chrisfromwork:
@Cameron-Micka @quakeulf please reopen this pull request if there is still interest in integrating this change into the MRTK.

Issue #2918, opened 2018-10-13
Focus Provider not returning correct focused object for event


StephenHodgson:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
When we handle events in the `MixedRealityInputManager` we don't get the focused object for our event properly when the pointer falls back to gaze.

This affects any input source that does not assign pointers when created.

**Expected behavior**
An event handled with an input source that does not have pointers, should use the gaze pointer and its focused object to handle the event

**Actual behavior**
No focused object is returned nor used, so the gaze input fails.

keveleigh:
I believe the issue largely lies here: https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs#L288.

The gaze pointer has its own input source and this will never match the input source of the event, so it's never returned by this method and thus the focused object is never returned.

This method also has an issue if we have multiple pointers with the same input source, since it'll just return the pointer that shows up first matching the ID instead of one specifically designated for focus / the type of event (like pointing vs teleport).

It seems like, since each input source knows about its pointers, there's probably a better way to figure out the correct pointer / focused object instead of needing some other script to iterate through all registered pointers from all sources.

I was pretty sure there was a different issue about this recently (and had been thinking about how to best fix it), but I can't seem to find it anymore.

StephenHodgson:
I'm thinking we should return all the pointers and their focused objects, but I'm not sure how to do it without incurring any allocations.

PR #2920, opened 2018-10-13
Fixed issue getting focused GameObject(s) when handling input event

StephenHodgson:
Overview
---
Removed eventData methods for getting focus data. Users should instead iterate over the event data's pointers and call on the focus provider's pointer specific methods for getting focus data. This ensures there are not extra allocations while getting focus data and keeps performance in mind.

Changes
---
- Fixes: #2918

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/DragAndDropHandler.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/DragAndDropHandler.cs
@@ -135,7 +135,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input.Handlers
             currentPointer = eventData.Pointer;
 
             FocusDetails focusDetails;
-            Vector3 initialDraggingPosition = MixedRealityManager.InputSystem.FocusProvider.TryGetFocusDetails(eventData.Pointer, out focusDetails)
+            Vector3 initialDraggingPosition = MixedRealityManager.InputSystem.FocusProvider.TryGetFocusDetails(currentPointer, out focusDetails)
                     ? focusDetails.Point
                     : hostTransform.position;
 
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/DragAndDropHandler.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/DragAndDropHandler.cs
@@ -135,7 +135,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input.Handlers
             currentPointer = eventData.Pointer;
 
             FocusDetails focusDetails;
-            Vector3 initialDraggingPosition = MixedRealityManager.InputSystem.FocusProvider.TryGetFocusDetails(eventData, out focusDetails)
+            Vector3 initialDraggingPosition = MixedRealityManager.InputSystem.FocusProvider.TryGetFocusDetails(eventData.Pointer, out focusDetails)
                     ? focusDetails.Point
                     : hostTransform.position;
 
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
@@ -302,6 +302,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
                     if (modalInput != null)
                     {
                         modalEventHandled = true;
+
                         // If there is a focused object in the hierarchy of the modal handler, start the event bubble there
                         if (focusedObject != null && focusedObject.transform.IsChildOf(modalInput.transform))
                         {
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
@@ -275,37 +275,43 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
                 return;
             }
 
-            GameObject focusedObject = FocusProvider?.GetFocusedObject(baseInputEventData);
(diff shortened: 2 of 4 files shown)


Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/DragAndDropHandler.cs
```diff
@@ -135,7 +135,7 @@ void IMixedRealityPointerHandler.OnPointerDown(MixedRealityPointerEventData even
             currentPointer = eventData.Pointer;
 
             FocusDetails focusDetails;
-            Vector3 initialDraggingPosition = MixedRealityManager.InputSystem.FocusProvider.TryGetFocusDetails(eventData, out focusDetails)
+            Vector3 initialDraggingPosition = MixedRealityManager.InputSystem.FocusProvider.TryGetFocusDetails(eventData.Pointer, out focusDetails)
```
StephenHodgson:
```suggestion
            Vector3 initialDraggingPosition = MixedRealityManager.InputSystem.FocusProvider.TryGetFocusDetails(currentPointer, out focusDetails)
```

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -275,37 +280,47 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
...
-                GameObject modalInput = modalInputStack.Peek();
+                GameObject focusedObject = FocusProvider?.GetFocusedObject(baseInputEventData.InputSource.Pointers[i]);
 
-                // If there is a focused object in the hierarchy of the modal handler, start the event bubble there
-                if (focusedObject != null && modalInput != null && focusedObject.transform.IsChildOf(modalInput.transform))
+                // Handle modal input if one exists
+                if (modalInputStack.Count > 0)
                 {
-                    if (ExecuteEvents.ExecuteHierarchy(focusedObject, baseInputEventData, eventHandler) && baseInputEventData.used)
+                    GameObject modalInput = modalInputStack.Peek();
```
keveleigh: It looks like this will send the same event to the modal stack for every pointer, even though the modal object isn't going to change. Is there a reason for that? Seems like it would lead to duplicate events and event handling.

StephenHodgson: Depends on if the pointer handles the event or not. We still check if the event is used.

If it's not used, then it should go to every handler.

keveleigh: Not all handlers will mark the event as used though. Some might want to respond to an event but let it continue down to other objects.

StephenHodgson: If the developer gets duplicate events, then they need to mark the eventData as used.

StephenHodgson: >Previously, this was fine because the modal object would only receive one event. Now, it will receive one for every pointer.

This doesn't handle "modal" input, only handles the currently focused objects. So there is no impact there.  If an object is focused by two different pointers, then it should get both events.

keveleigh: This does send events to the modal listeners: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2920/files#diff-83ff2790c22f67c0d7d4a7a413be9c29R310

I agree about the two different pointers scenario. That's not what I'm talking about.

StephenHodgson: Maybe we should check if the `modalInput` is the same as the focused object?

Thoughts?

StephenHodgson:
```suggestion
                    GameObject modalInput = modalInputStack.Peek();

                    if (modalInput != focusedObject) { continue; }
```

StephenHodgson:
```suggestion
                    GameObject modalInput = modalInputStack.Peek();

                    if (modalInput != focusedObject) { continue; }
```

keveleigh: This code checks to see if the `focusedObject` is a child of `modalInput`, but can probably be removed in this new world with multiple focused objects (unless we want to get all focused objects, do this check for all focused objects, send the modal event as appropriate, then send the focused events): https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2920/files#diff-83ff2790c22f67c0d7d4a7a413be9c29R300

If we take the above suggested change, `modalInput` will never get the event unless it's also the `focusedObject`, which goes against the purpose of the modal input stack.

The event should be sent to the modal object once, probably before the pointers are iterated through.

davidkline-ms: > The event should be sent to the modal object once, probably before the pointers are iterated through.

I agree

StephenHodgson: >If we take the above suggested change, modalInput will never get the event unless it's also the focusedObject, which goes against the purpose of the modal input stack.

Yikes, yeah can't do that. 

> The event should be sent to the modal object once, probably before the pointers are iterated through.

The only way to get the focused object is with a pointer reference, and we could potentially have multiple focused objects.

I'm fairly certain that most (if not all) modal inputs consume the event.

keveleigh: >The only way to get the focused object is with a pointer reference, and we could potentially have multiple focused objects.

Yes, I agree. How does this relate to sending the event to the modal input?

StephenHodgson: I've got an idea. One tick.

PR #2921, opened 2018-10-13
Windows Dictation Input Device Update

StephenHodgson:
Overview
---
- Cleaned up the dictation input device manager, and removed allocations when starting/stopping recording.
- Also added some safely checks to ensure we're in play mode when recording.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Devices/VoiceInput/WindowsDictationInputDeviceManager.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/VoiceInput/WindowsDictationInputDeviceManager.cs
@@ -59,8 +59,14 @@ namespace Microsoft.MixedReality.Toolkit.Core.Devices.VoiceInput
         /// </summary>
         private AudioClip dictationAudioClip;
 
-        private DictationRecognizer dictationRecognizer;
-        
+        private static DictationRecognizer dictationRecognizer;
+
+        private readonly WaitUntil waitUntilPhraseRecognitionSystemHasStarted = new WaitUntil(() => PhraseRecognitionSystem.Status != SpeechSystemStatus.Stopped);
+        private readonly WaitUntil waitUntilPhraseRecognitionSystemHasStopped = new WaitUntil(() => PhraseRecognitionSystem.Status != SpeechSystemStatus.Running);
+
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/_Core/Devices/VoiceInput/WindowsDictationInputDeviceManager.cs
```diff
@@ -59,8 +59,14 @@ public class WindowsDictationInputDeviceManager : BaseDeviceManager, IMixedReali
         /// </summary>
         private AudioClip dictationAudioClip;
 
-        private DictationRecognizer dictationRecognizer;
-        
+        private static DictationRecognizer dictationRecognizer;
```
keveleigh: What's the benefit to adding `static` here?

StephenHodgson: because then I couldn't call it in the non-static methods

StephenHodgson: So that I can use it in the other static contexts where it is needed.

davidkline-ms: For posterity (ex: when someone does a search and finds this 6 months from now), can you provide and example of one of the static contexts where you need access?

thx!

StephenHodgson: Sure, two lines below, in the [readonly fields where we setup our async helpers.](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2921/files#diff-ee543afc5d05dda8c9247ad49c1ab8e7R67)

StephenHodgson: There should never be more than one dictation recognizer or `WindowsDictationInputDeviceManager` running in an application anyhow.

davidkline-ms: that's what I figured. thanks for making it clear for future readers :)

PR #2922, opened 2018-10-14
Refactored Manager concept into Services

StephenHodgson:
Overview
---
There was a bit of confusion and misconception of the `MixedRealityManager`. Its sole responsibility is to keep all the other feature systems, and components synchronized and in step. This is the only entry point for the Mixed Reality Toolkit into the user's project and into the scene. This component's sole responsibility is to be a resource locator and facilitator of each feature's communication in a decoupled and emergent way.

[More info in my blog post about the design decisions](https://medium.com/@stephen_hodgson/the-mixed-reality-framework-6fdb5c11feb2)

This Pull Request tries to re-define the concepts of this core system and break misconceptions.

No logical changes, only renames.


StephenHodgson:
I highly recommend we make this change before we go beta

StephenHodgson:
> we will likely need to be changing some layering as well.

Likely not unless something substantial comes up that wasn't previously addressed.

davidkline-ms:
> I highly recommend we make this change before we go beta

why the rush? i really want to understand the customer issue before making changes

keveleigh:
I don't think this change addresses the concerns as I understand them. We shouldn't make a breaking change until we properly understand the underlying concerns.

davidkline-ms:
> we will likely need to be changing some layering as well.
Likely not unless something substantial comes up that wasn't previously addressed.

The layering I am alluding to is updating the systems to access the manager through an interface, not the concrete class and allowing for pluggable managers.

StephenHodgson:
> The layering I am alluding to is updating the systems to access the manager through an interface, not the concrete class and allowing for pluggable managers.

Lol what? We already enforce that.

StephenHodgson:
This also isn't really a breaking change as just a rename.

StephenHodgson:
I'm tired of being hanstrung trying to be agile, make improvements, and address feedback while still in alpha.

davidkline-ms:
> The layering I am alluding to is updating the systems to access the manager through an interface, not the concrete class and allowing for pluggable managers.
Lol what? We already enforce that.

We enforce from the manager -> systems, not the other way around...

For example:
MixedRealityToolkit-SDK\Features\Input\FocusProvider.cs:239:            foreach (var inputSource in MixedRealityManager.InputSystem.DetectedInputSources)

davidkline-ms:
> I'm tired of being hanstrung trying to be agile, make improvements, and address feedback while still in alpha.

We are in the final stretch of finishing the beta... we have already agreed that breaking changes are allowed between major releases (ex: 2018 -> 2019 -> 2020).

I really want to make sure we are setup for success with both enterprise and independent customers.

davidkline-ms:
> If we don't do it now it'll cause more confusion in the future.

The confusion is why I requested (on slack) that I meet with the customer and have a deep dive discussion. I do not believe changing the name solves the underlying issue.

StephenHodgson:
> we have already agreed that breaking changes are allowed between major releases (ex: 2018 -> 2019 -> 2020).

Community members didn't have much say in the matter as it's part of the internal agenda. We really shouldn't be rushing this as much as it has been. Imo we need to release the beta _when it's ready_.

StephenHodgson:
> We enforce from the manager -> systems, not the other way around...
>
>For example:
MixedRealityToolkit-SDK\Features\Input\FocusProvider.cs:239: foreach (var inputSource in MixedRealityManager.InputSystem.

No we don't. And this is a one off example of something that needs to be fixed and already has an issue assigned to it. This is one of the few exceptions, but is identified.

StephenHodgson:
>  I do not believe changing the name solves the underlying issue.

I don't think there's an issue at all, but a complete misunderstanding and misconceptions of past experiences. Documentation and discussion will help solve those issues. If the MRTK isn't something they want to utilize that's their decision.

davidkline-ms:
> I do not believe changing the name solves the underlying issue.
I don't think there's an issue at all, but a complete misunderstanding and misconceptions of past experiences. Documentation and discussion will help solve those issues. If the MRTK isn't something they want to utilize that's their decision.

We should table this discussion until the next shiproom (tomorrow) so that we can get clarity.

StephenHodgson:
Unfortunately I will not be able to attend as I'll be 30k ft 😅

davidkline-ms:
> Unfortunately I will not be able to attend as I'll be 30k ft

We can schedule a call for when you are on the ground.

SimonDarksideJ:
I would have to agree with Stephen on this point.  As this is really just about naming and breaking down barriers to understanding, especially following the recent meetings.  It would be better to get the names right before finalising Beta.

Else, this is going to hit hard on any consumers / customer consuming the project as they'll have major refactoring to do later.

I can understand wanting to wait, but that will ultimately hurt us more upgrading to 2019, as well as invalidating all our documentation / tutorials / etc.

I personally would prefer to settle on the naming now and stick to it, rather than change later. imho.  And the new namings align more to the feedback we've been getting.

davidkline-ms:
> invalidating all our documentation / tutorials / etc.

that is a reasonable concern. I **do** want to make sure we revisit the feedback from the customer (I have a meeting tomorrow) and only change names if the value is there.

SimonDarksideJ:
Fair enough.
I'm planning an a quick start series for the beta.  But if the naming / use is going to change straight after the beta is released (e.g. for 2019) then there isn't much point, as the Beta would just be an internal stream.

davidkline-ms:
I should have a better inderstanding within 24 hours.  thanks 
for your feedback and understanding

Railboy:
Don't want to step in the name now / name later debate - I see the need for the change as well as the rationale behind waiting.

@StephenHodgson @SimonDarksideJ  `IMixedRealityService` feels much clearer. Not a huge fan of `Orchestrator` though.

`Manager` is misleading but clear. You get that important things happen there even if you're wrong about its responsibilities.

`Orchestrator` doesn't mislead but only because it puts no clear idea in mind, even if you're familiar with the technical term. At a glance I'd take it for some background service I wasn't supposed to touch.

We played with alternatives and thought `Hub` had potential. Hubs are where important things happen (or through which important things are routed) without the expectation of management / intervention. And `MixedRealityHub` is less verbose than `MixedRealityOrchestrator` which is a nice bonus.

davidkline-ms:
> I should have a better inderstanding within 24 hours.

Looks like I misread my calendar.... I meet with the customer friday.

>  Hubs are where important things happen (or through which important things are routed)

I could live with Hub... Part of me would really like to just call it MixedRealityToolkit as the component is the entry point to the MRTK.

SimonDarksideJ:
I like your thinking Lars and this is a really good debate, especially around not trying to make it too wordy.

Whilst I already suggested simply naming the scene object to "MixedRealityToolkit", there is also potential merit to also calling the "orchestrator / service locator / hub" as simply

> MixedRealityToolkit

Doesn't need to say what it does or how it works, just that it IS the heart of the toolkit (although, also now an "optional" thing maybe)

Thoughts?

davidkline-ms:
To summarize the discussion so far. I believe everyone who commented is good with the interface rename and that the blocker for this PR is the manager rename.

SimonDarksideJ:
@davidkline-ms , can you summarise the discussion in Shiproom about the "agreed" changes / standards for this PR here please.

davidkline-ms:
@SimonDarksideJ, was just about to do that :)

davidkline-ms:
From 10/17 shiproom, it was decided that the following changes should be made:

Namespace:
Microsoft.MixedReality.Toolkit.Core.Managers -> Microsoft.MixedReality.Toolkit.Core.Systems

Interfaces:
IMixedRealityManager* -> IMixedRealitySystem*

Scene object:
MixedRealityManager -> MixedRealityToolkit
(this includes the internal object name as well, so MixedRealityManager.Instance becomes MixedRealityToolkit.Instance)

Open item: What to call the registered components, since component has a Unity concept associated.
Proposed terms:
* Service
* Agent
* Provider

mrmanager1

mrmanager2

StephenHodgson:
Just to clarify `IMixedReality<Feature Name>System`, correct?

davidkline-ms:
> Just to clarify IMixedReality<Feature Name>System, correct?

Yes. Trust the shiproom notes (the multicolored handwritten images). :)

StephenHodgson:
I believe the only change that needs to be made to accommodate this with the PR is renaming the Mixed Reality Orchestrator.

You guys should have push access to my branch to do that. I won't be able to make it to a computer and time soon.

davidkline-ms:
> I believe the only change that needs to be made to accommodate this with the PR is renaming the Mixed Reality Orchestrator.

The Services namespace was also a question mark.

StephenHodgson:
They're all services

Assets/MixedRealityToolkit-Examples/Demos/Boundary/Scripts/BoundaryVisualizationDemo.cs
```diff
@@ -5,7 +5,6 @@
 using Microsoft.MixedReality.Toolkit.Core.EventDatum.Boundary;
 using Microsoft.MixedReality.Toolkit.Core.Interfaces.BoundarySystem;
 using Microsoft.MixedReality.Toolkit.Core.Services;
-using Microsoft.MixedReality.Toolkit.Core.Utilities;
```
davidkline-ms: while removing this line feels like a good cleanliness change, it doesn't feel like part of this change...

(plus other similar changes)

StephenHodgson: I heavily adjusted the namespaces and resharper cleaned them up automatically. I don't think it's worth the effort to pull them out separately.

PR #2929, opened 2018-10-16
 Add link to Azure hosted Slack invitation sender

Nakda:
The current documentation mention to join HoloLens dedicated forum, or Slack. If you are not member of HoloDevelopers on Slack, you need someone to send you an invitation.

An automatic invitation sender has been hosted on Azure for a while now, and can provide a quick and easy access to anybody without manual intervention.

Since the Contribution guide was providing the Slack address, I suggest we also provide a link to the automatic invitation sender.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ For larger submissions, we have drafted contribution guidelines to ensure a smoo
 
 All changes be they small or large, need to adhere to the [MRTK Coding Standards](/CodingGuidelines.md), so please ensure you are familiar with these while developing to avoid delays when the change is being reviewed.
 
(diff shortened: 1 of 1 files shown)


StephenHodgson:
Much appreciated!

Maybe we should do the same for htk_development as well?

Nakda:
> Maybe we should do the same for htk_development as well?

Do you only want this addition mirrored to `htk_development`, or should we replace the whole document by the one from `mrtk_development` (which seems more up-to-date) ? 

In case of replacement, statements refering to vNEXT will need to be edited accordingly.

StephenHodgson:
I think it'd be a good idea as the HTK won't be depreciated anytime soon.

PR #2935, opened 2018-10-17
Seperate editor features into their own binary

ryzngard:
Assemblies included in standalone apps cannot target the UnityEditor.dll binary. This means that when we ship binaries, the ones we intend to be included as part of the deployed app have to be independent of editor features. 

This change moves all non-editor code out of the Assets folder into a Src folder at the root. The Assets\Bin folder will contain the compiled dlls that can then be referenced. 

This changes the workflow of **contributors of MRTK** to be divided into two steps: 

1. Code that isn't asset/editor related needs to be changed in Src location and built using the sln/csproj files
2. Assets and editor scripts need to be changed in the Asset folder directly 

Right now this also adds a separate NuGet package for editor resources. We should consider including them in the right NuGet packages directly and broken up by feature as we break up packages more and more.


ryzngard:
This is one solution proposal. Another would be to have a source only NuGet that unpacks these files as script files. https://github.com/GlitchEnzo/NuGetForUnity/issues/193 is currently blocking this.

ryzngard:
As discussed in shiproom, we could include as content in NuGet packages.

StephenHodgson:
I'm not a fan of this change. Is there another way we can accomplish this?

ryzngard:
@StephenHodgson possibly, but what specifically is wrong with this change? If I understand the concern it will help me evaluate options.

StephenHodgson:
All the assets need to stay where they are.

davidkline-ms:
> All the assets need to stay where they are.

Can you elaborate, please?

StephenHodgson:
>> All the assets need to stay where they are.
> Can you elaborate, please?

I'm not sure what else to add. There's no reason to move everything outside of the MRTK folders. It will add extra overhead to making unity packages, and keeping everything together in an organized way is ideal.

davidkline-ms:
>  It will add extra overhead to making unity packages

Can you be specific on the extra overhead involved in making the packages? Based on my read, @ryzngard is proposing one possible solution that will, as a side effect, ensure that contributors test what is actually shipping.

StephenHodgson:
> ensure that contributors test what is actually shipping.

That isn't part of the change outlined in the overview. Besides we can already test before we ship anyway.

It adds a lot of overhead to edit and make changes. Everything should be in the Assets folder. Anything published to NuGet, and binaries were created that lived outside of the Assets folder.

Doing this will create a barrier to entry to editing any of those scripts.

jwittner:
An alternative that I've used here is to have a Shared Project referenced by two projects that build the same DLL - one with UNITY_EDITOR defined and one without it. Then I have Unity .meta files preconfigured for the two dlls which indicate which one should be used in the editor vs uwp/etc.


Alongside @davidkline-ms points - the big problem with the eternal code workflow for me has been the ability to debug those DLLs. I have worked around it by referencing the external projects into the Sln created by Unity and then debugging as usual - since we don't checking that Sln right now I have to remake those references locally pretty often. That part is a pain.

Upside has been that the code was much easier to keep compartmentalized ( may be alleviated by the assembly defs now available in Unity), I get the latest/greatest C# features since I don't ever assume Unity will be compiling it, and ensuring dependencies on Unity aren't accidentally taken by our packages has been vastly easier.

StephenHodgson:
Seeing as the .unityassets are already binaries, couldn't we just ship that?

StephenHodgson:
We can also just use the build assemblies that the engine creates in `\Library\ScriptAssemblies`

StephenHodgson:
Literally no extra work. Those things are ready to go out of the box.

jwittner:
That's an interesting option @StephenHodgson ! I'm interested in how the assembly defs work with editor content. Have we experimented with that?

StephenHodgson:
Yes, the new MRTK takes advantage of all those things. The editor scripts are already broken out into their own assemblies by the time they make it into that folder.

StephenHodgson:
image

StephenHodgson:
Although I do want to point out that this doesn't have any assets like textures, meshes, etc in them.

davidkline-ms:
> The SDK was going to be all source anyway.

This is something that @ryzngard and I have talked a bit about. we could improve compile time by shipping the SDK ~system~ implementations as compiled assemblies.

StephenHodgson:
> SDK system implementations as compiled assemblies.

I'm going to take a hard stance and say no. We want people to be able to copy what's in the SDK and customize it to what their needs/requirements are if it doesn't 100% align.

We already save loads of compile time just using the assembly definitions.

davidkline-ms:
>I'm going to take a hard stance and say no. We want people to be able to copy what's in the SDK and customize it to what their needs/requirements are if it doesn't 100% align.

The proposal is not to pre-compile the entire SDK, just the default system implementations. There are pros and cons for shipping source for the systems and precompiling them.

The MRTK is targeted at a wide variety of customers with varying needs and skills.  Having precompiled systems can provide a great plug and play experience for those who wish only to consume the toolkit and those who wish to use the default implementations.

This seems like a great topic for an upcoming shiproom and/or a packaging design meeting.

StephenHodgson:
That's why I'm confused why we don't have the default implementations in the core to begin with.

Idk just seems rather insignificant compared to the time it takes to import art assets of the SDK.

- There's little to no speed improvement and will likely make it harder for people to contribute fixes back.
- Harder for people to copy and customize.
- More binaries to sign and verify.

The more we have these kinds of conversations, the more I fear we don't share some core values for  our level of commitment to developers. Many of the decisions are made from either a naieve understanding of why we've made specific design choices or a lack of experience in developing software specifically for Unity.

Many developers will expect to get source code from our asset packages, unless it's very stable or highly reliable and trusted. Will we get there one day with MRTK? I hope so. But until we've had time to vet and verify I think we ought to stick with the source in the SDK.

davidkline-ms:
Closing as the specific PR is not ready and there needs to be more design discussion related to packaging and what ships pre-compiled.

Issue #2933, opened 2018-10-17
Radial Solver Z Coordinates inverted


hayalk:
I'm guessing after #2898 the radial solver wasn't updated and GameObjects attached with that script always appears behind the camera. HoloToolkit 2017.4.2.0.

keveleigh:
Ah, I must have missed one. I'll get a patch out asap.

PR #2937, opened 2018-10-17
Prevent solver targets from rotating when the target is the head

keveleigh:
Overview
---
The fix in #2898 didn't properly account for when the solver was targeting the head. In that case, we don't want to add the 180 degree rotation, as that's only needed for the platform controller models.

Changes
---
- Fixes: #2933

### DIFF
--- a/Assets/HoloToolkit/Utilities/Scripts/Editor/SolverHandlerEditor.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/Editor/SolverHandlerEditor.cs
@@ -50,13 +50,13 @@ namespace HoloToolkit.Unity.UX
 
             serializedObject.ApplyModifiedProperties();
 
-            if (trackedObjectChanged)
+            if (Application.isPlaying && trackedObjectChanged)
             {
                 solverHandler.TransformTarget = null;
                 solverHandler.AttachToNewTrackedObject();
             }
 
-            if (additionalOffsetChanged)
+            if (Application.isPlaying && additionalOffsetChanged)
             {
                 solverHandler.AdditionalOffset = additionalOffsetProperty.vector3Value;
                 solverHandler.AdditionalRotation = additionalRotationProperty.vector3Value;
--- a/Assets/HoloToolkit/Utilities/Scripts/Solvers/SolverHandler.cs
+++ b/Assets/HoloToolkit/Utilities/Scripts/Solvers/SolverHandler.cs
@@ -216,8 +216,12 @@ namespace HoloToolkit.Unity
             transformWithOffset.transform.localPosition = AdditionalOffset;
             transformWithOffset.transform.localRotation = Quaternion.Euler(AdditionalRotation);
             transformWithOffset.name = string.Format("{0} on {1} with offset {2}, {3}", gameObject.name, TrackedObjectToReference.ToString(), AdditionalOffset, AdditionalRotation);
-            // In order to account for the reversed normals due to the glTF coordinate system change, we need to provide rotated attachment points.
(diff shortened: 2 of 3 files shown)


keveleigh:
@hayalk, this should resolve your issue. Please let me know if it doesn't!

Assets/HoloToolkit/Utilities/Scripts/Editor/SolverHandlerEditor.cs
```diff
@@ -50,13 +50,13 @@ public override void OnInspectorGUI()
...
 
-            if (trackedObjectChanged)
+            if (Application.isPlaying && trackedObjectChanged)
             {
                 solverHandler.TransformTarget = null;
                 solverHandler.AttachToNewTrackedObject();
             }
 
-            if (additionalOffsetChanged)
+            if (Application.isPlaying && additionalOffsetChanged)
```
Ecnassianer: Does changing the offset while the game is running actually work? I though we disabled this field since we don't regenerate the offset transforms once they're set. Does the AdditionalOffset and AdditionaRotation accessors reinitialize the offset transforms?

Assets/HoloToolkit/Utilities/Scripts/Solvers/SolverHandler.cs
```diff
@@ -216,8 +216,12 @@ private Transform MakeOffsetTransform(Transform parentTransform)
             transformWithOffset.transform.localPosition = AdditionalOffset;
             transformWithOffset.transform.localRotation = Quaternion.Euler(AdditionalRotation);
             transformWithOffset.name = string.Format("{0} on {1} with offset {2}, {3}", gameObject.name, TrackedObjectToReference.ToString(), AdditionalOffset, AdditionalRotation);
-            // In order to account for the reversed normals due to the glTF coordinate system change, we need to provide rotated attachment points.
-            transformWithOffset.transform.Rotate(0, 180, 0);
+
+            if (TrackedObjectToReference != TrackedObjectToReferenceEnum.Head)
```
Ecnassianer: Will things like puck trackers and other accessories need to be reversed too?

keveleigh: Likely not, but those types also likely won't go into the legacy Toolkit. We'll need to address this a little more elegantly in vNext, since the legacy Toolkit assumes that the only platform controller models we'll receive are from the Windows MR APIs.

I can rework it here first if you'd prefer, but I don't think it's a scenario we'll encounter.

Assets/HoloToolkit/Utilities/Scripts/Editor/SolverHandlerEditor.cs
keveleigh: Yep, when the additional offset or additional rotation is changed at runtime, we update the tracked transform accordingly: https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit/Utilities/Scripts/Solvers/SolverHandler.cs#L69

Assets/HoloToolkit/Utilities/Scripts/Solvers/SolverHandler.cs
Ecnassianer: Ah, I missed which branch this was going into!

Assets/HoloToolkit/Utilities/Scripts/Editor/SolverHandlerEditor.cs
Ecnassianer: Awesome!

PR #2944, opened 2018-10-20
Updated Base Pointer Controller Pointer Name Property  …

StephenHodgson:
Overview
---
- Added a backing field, in case the gameObject reference is destroyed when attempting to get the pointer name

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
@@ -190,11 +190,17 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
             }
         }
 
+        private string pointerName = string.Empty;
+
         /// <inheritdoc />
         public string PointerName
         {
-            get { return gameObject.name; }
-            set { gameObject.name = value; }
+            get { return pointerName; }
+            set
+            {
+                pointerName = value;
+                gameObject.name = value;
+            }
         }
 
         /// <inheritdoc />


Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
```diff
@@ -190,11 +190,17 @@ public uint PointerId
...
         /// <inheritdoc />
         public string PointerName
         {
-            get { return gameObject.name; }
-            set { gameObject.name = value; }
+            get { return pointerName; }
+            set
+            {
+                pointerName = value;
+                gameObject.name = value;
```
keveleigh: This will still null ref if the gameObject is destroyed.

StephenHodgson: Yes, but we only set the name once we create the pointer, so I doubt that can ever really happen.

StephenHodgson:
```suggestion
                if(gameObject == null)
                {
                    Debug.LogError($"{pointerName}'s GameObject reference is missing!");
                    return;
                }
                
                gameObject.name = value;
```

StephenHodgson:
>isn't this hiding an issue? if the underlying gameobject has been destroyed, the pointer is no longer valid... how do you envision an app getting into this situation?

I've never had an issue _setting_ the name, only _getting_ it during shutdown because of the non-deterministic nature of unity's game loop. For example, if a GameObject get's destroyed before the MixedRealityToolkit ServiceLocator finished shutting down the InputSystem.

keveleigh:
>For example, if a GameObject get's destroyed before the MixedRealityToolkit ServiceLocator finished shutting down the InputSystem.

If the `gameObject` is destroyed, won't the pointer script also be destroyed? Does checking the pointer script reference for null before trying to access the name not work?

StephenHodgson:
>If the gameObject is destroyed, won't the pointer script also be destroyed? Does checking the pointer script reference for null before trying to access the name not work?

No it doesn't which is why we're getting the issue in the first place. 😅

keveleigh:
Which line is causing the issue?

StephenHodgson:
The `gameObject` can sometimes be null [when exiting play mode in the editor](
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2944/files#diff-820aa912c19ab5fb36b99791f204016cL196) because it is cleaned up before the system shuts down.

keveleigh:
Right, that part is clear from this PR. I was asking if there's a line in the Toolkit asking the pointer for its name at a time when the `gameObject` is null.

StephenHodgson:
When we're [unregistering the pointers and removing it from the list](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs#L494). Under the [hood the `Remove` function tries to get the hash cod](https://referencesource.microsoft.com/#system.core/system/Collections/Generic/HashSet.cs,289)e.  In our [hash code override](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs#L368) we access [the property](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs#L196), which returns the `gameObject.name` instead of a backed field.

davidkline-ms:
@StephenHodgson, that's a great description of the issue. Thanks for adding it. Makes the change much easier to understand.

Issue #2856, opened 2018-09-28
New config profile created in the wrong folder.


StephenHodgson:
**Describe the bug**
When importing only the MRTK core folder and creating a new configuration profile, the profile creates a 
`MixedRealityToolkit-SDK` folder.

Instead this should create just a folder named profiles in the root assets folder.

**Steps to reproduce the behavior:**
1. Download and import just the core MRTK folder.
2. Click configure.
3. Accept the dialog option, asking to create a new configuration
4. See a new folder named `MixedRealityToolkit-SDK`

**Expected behavior**
Should create a new folder named `Profiles` at the root assets folder.

**Actual behavior**
Creates a new folder named `MixedRealityToolkit-SDK`

davidkline-ms:
shouldn't profiles be created in the current folder? that's the behavior I have been seeing.

StephenHodgson:
It is, except for the main configuration profile itself.

StephenHodgson:
Best to repro this without the SDK installed in a new project.

PR #2951, opened 2018-10-21
Added Copy/Paste functionality to configuration profiles.

StephenHodgson:
Overview
---
- Added two buttons on the master configuration profile for options on how users would like to make their profiles. They can choose to copy from the default ones and customize, or create new profiles.
image
- Configuration profiles now have copy/paste functionality.
image
- Default Configuration Profiles will now show warnings with helpful hints about how to proceed.
image
- All Configuration Profiles can be copied into a completely new asset.
- Once a master configuration profile has been swapped out with a custom once, an option to replace default profiles appears.
image

Changes
---
- Fixes: #2856
- Fixed an issue where BoundarySystemIsEnabled wasn't checking if a profile was set.


Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityConfigurationProfileInspector.cs
```diff
@@ -116,7 +119,24 @@ public override void OnInspectorGUI()
...
+                EditorGUILayout.HelpBox("The Mixed Reality Toolkit's core SDK profiles can be used to get up and running quickly.\n\n" +
+                                        "You can use the default profiles provided, copy and customize the default profiles, or create your own from scratch", MessageType.Warning);
+                EditorGUILayout.BeginHorizontal();
+
+                if (GUILayout.Button("Copy & Customize"))
+                {
+                    CreateCopyProfileValues();
+                }
+
+                if (GUILayout.Button("Start from Scratch"))
```
SimonDarksideJ:
```suggestion
                if (GUILayout.Button("Create a new profile"))
```
better title

StephenHodgson:
```suggestion
                if (GUILayout.Button("Start from blank"))
```

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityConfigurationProfileInspector.cs
```diff
@@ -116,7 +119,24 @@ public override void OnInspectorGUI()
 
             if (MixedRealityPreferences.LockProfiles && !((BaseMixedRealityProfile)target).IsCustomProfile)
             {
-                EditorGUILayout.HelpBox("The Mixed Reality Toolkit's core SDK profiles can be used to get up and running quickly.\n\nYou can use the default profiles provided or create your own in the context menu:\n'Create/Mixed Reality Toolkit/...'", MessageType.Warning);
+                EditorGUILayout.HelpBox("The Mixed Reality Toolkit's core SDK profiles can be used to get up and running quickly.\n\n" +
+                                        "You can use the default profiles provided, copy and customize the default profiles, or create your own from scratch", MessageType.Warning);
```
StephenHodgson:
```suggestion
                                        "You can use the default profiles provided, copy and customize the default profiles, or create your own.", MessageType.Warning);
```

davidkline-ms:
I pulled this change and do not see the new buttons...  am I missing sonething?

StephenHodgson:
Did you have custom profiles?

davidkline-ms:
> Did you have custom profiles?

My steps were:
1. Pull your branch
2. Load the project
3. Click Mixed Reality Toolkit > Configure
-- no copy button --

Note: My profiles are unlocked.

StephenHodgson:
Lock the profiles and select the MixedRealityManager

davidkline-ms:
> Lock the profiles and select the MixedRealityManager's configuration profile.

Why do they have to be locked for this feature to work? It feels confusing. Also, shouldn't I be able to clone custom profiles?

StephenHodgson:
>Why do they have to be locked for this feature to work?

Because it only works if they're unlocked.

davidkline-ms:
> Because it's only meant to work if they're locked. That's by design.

I am not sure i agree with the design. i found it confusing why the behavior difference... i believe it will confuse our customers as well.

davidkline-ms:
what is the difference between Copy and Create Copy From in the drop down menu?

StephenHodgson:
image
This is simply a warning about the profile being one of the default ones. Users will get the choice to copy and Customize or simply create their own, or use the ones provided as is.
Every profile has this warning, but the master configuration profile is unique with these buttons because it's the only one that doesn't have a "Copy & Customize" button to press on the MixedRealityToolkit GameObject.
image

The full copy/paste exists as a context menu to the Object itself, so pressing the gear in the top right makes since cause it's the same way you'd copy and paste component values for anything else.
image

StephenHodgson:
I'm open to updating the text to be "Clone & Customize" for the button.

and updating the context menu to read as "Create a Clone from Profile Values

wassx:
Love this feature!

Assets/MixedRealityToolkit/_Core/Inspectors/MixedRealityToolkitInspector.cs
```diff
@@ -53,7 +53,7 @@ public override void OnInspectorGUI()
                                                                   "Would you like to create one now?", "OK", "Later"))
                     {
                         ScriptableObject profile = CreateInstance(nameof(MixedRealityToolkitConfigurationProfile));
-                        profile.CreateAsset();
+                        profile.CreateAsset("Assets/CustomProfiles");
```
wassx: Minor naming thing. Maybe call it something like "Assets/MixedRealityToolkit-generated/CustomProfiles"
to be aware where it belongs to in a bigger folder structure and it's generated?

StephenHodgson:
```suggestion
                        profile.CreateAsset("Assets/MixedRealityToolkit-generated/CustomProfiles");
```

keveleigh:
>No one outside of our development team will likely unlock the profiles, nor should they (unless they're modifying it for a PR against the MRTK)

We can't prevent that. Having the copy button only appear when the profile is locked is confusing, since it's the "default" profile regardless of locked state, and will be overwritten in a Toolkit update.

It's almost more important to provide information about creating a copied profile in the unlocked state, since that's the "danger" state where you might make changes that won't persist.

davidkline-ms:
> Having the copy button only appear when the profile is locked is confusing, since it's the "default" profile regardless of locked state

This is the root of my confusion. There is no easily discoverable way to set isCustomProfile to true if the Copy button is hidden when profiles are unlocked.

I would be 100% in support of this change if the copy button existed in both the locked and unlocked states.

StephenHodgson:
I disagree as the only state a new profile is in is `isCustomProfile==tue`

StephenHodgson:
> I would be 100% in support of this change if the copy button existed in both the locked and unlocked states.

Technically the buttons are always there in the context menu. But sure I'll take a PR with your suggested edits.

StephenHodgson:
@davidkline-ms added button visibility when unlocked.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityBaseConfigurationProfileInspector.cs
```diff
@@ -17,6 +22,11 @@ public abstract class MixedRealityBaseConfigurationProfileInspector : BaseMixedR
         [SerializeField]
         private Texture2D logoDarkTheme = null;
 
+        private static BaseMixedRealityProfile profile;
+        private static SerializedObject targetProfile;
+
```
StephenHodgson:
```suggestion
```

davidkline-ms:
> @davidkline-ms added button visibility when unlocked.

Great! i'll try it at my first opportunity,

StephenHodgson:
@davidkline-ms going to let you test and merge this then satisfactory :)

davidkline-ms:
Just pulled this PR and I do not see copy buttons. Tested the profiles being locked and unlocked.

davidkline-ms:
wait... I may have opened the wrong branch... let me double-check

davidkline-ms:
Yep. Wrong branch. Buttons appear in both states. Thanks!

wassx:
Will this find its way into the mrtk_development branch or will the mrtk_2018beta_stabilization be the most recent branch?

PR #2953, opened 2018-10-22
replace manager exceptions with debug.logerror and return

davidkline-ms:
The manager should not throw exceptions. They are expensive and handled automatically, so they aren't much more useful than a simple (and quicker) debug message

Changes
---
* The manager no longer throws
* Error debug log messages are displayed
* Safe (null, empty collections, etc) values are returned where an exception would have been thrown

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
+++ b/Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
@@ -271,7 +271,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Managers
         /// <summary>
         /// Returns the Singleton instance of the classes type.
         /// If no instance is found, then we search for an instance in the scene.
-        /// If more than one instance is found, we throw an error and no instance is returned.
+        /// If more than one instance is found, we log an error and no instance is returned.
         /// </summary>
         public static MixedRealityManager Instance
         {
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/_Core/Managers/MixedRealityManager.cs
```diff
@@ -786,7 +848,11 @@ public void EnableManager(Type type, string managerName)
         /// <returns>An array of Managers that meet the search criteria</returns>
         public IEnumerable<IMixedRealityManager> GetManagers(Type type)
         {
-            if (type == null) { throw new ArgumentNullException(nameof(type)); }
+            if (type == null)
+            {
+                Debug.LogWarning("Unable to get managers with a type of null.");
+                return new List<IMixedRealityManager>();
```
StephenHodgson: shouldn't this return null?

StephenHodgson: (same question for similar returns)

davidkline-ms: if it returns null, then code that expects to be able to enumerate over the collection, for example

foreach (IMixedRealityManager mgr in MixedRealityManager,Instance.GetManagers(type) 

will throw an exception, or have to be rewritten to cache the result of GetManagers before the loop.

davidkline-ms:
trying to get checks to resolve....

StephenHodgson:
Yup. Likely Github is trying to recover still.

PR #2967, opened 2018-10-24
Mrtk vNext Sample Scene Models

JonathanPalmerGD:
Overview
---Various sample scenes require the use of HTK's assets but they have not been brought over to vNext yet.

Changes
---Added most of the old Holo Toolkit models into MRTK-Example/StandardAssets
---Did not include the old button model that was used (Yoon's suggestion)
---Did not include Model_Tetra as it was just a Unity cube with un-intuitive parent hierarchy.


cre8ivepark:
Thanks @JonathanPalmerGD !

Issue #2975, opened 2018-10-25
vNEXT Task: Setup VR SDKs when configuring


StephenHodgson:
# Overview
When the user imports and configures the MRTK, the XR SDKs have not been setup and requires the user to go an manually turn these on. We would like to automate this a bit more and automatically enable XR settings after a dialog that prompts them for acceptance of the project change.

# Requirements

* When importing or configuring the MRTK show a prompt that asks the user to enable XR settings for their currently targeted platform.
* If user accepts then we update the project settings.

# Acceptance Criteria

- [x] When configuring the MRTK and XR settings is not enabled show a prompt.
- [x] If user accepts prompt update the project settings to enable XR for the current target platform.

StephenHodgson:
Looks like new Unity projects already have the required SDKs enabled, so we just need to turn on the XR Settings.

Issue #2860, opened 2018-09-28
vNext: enable "Virtual Reality Supported" automatically like htk did


Ecnassianer:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Is your feature request related to a problem? Please describe.**
I've gotten hamstrung by not having Virtual Reality Supported checked for me. I'm so used to htk project settings setup doing it for me, that I keep forgetting it in MRTK, and I send up spending 30+ minutes figuring out why my headset doesn't work.

**Describe the solution you'd like**
I'd like for this to be turned on for me, or at least to have a prompt asking if I want it turned on like the IL2CPP prompt.

StephenHodgson:
This is a good point, and surprised we haven't addressed this sooner.

One of the ideas we had was to enable/disable VR support as hardware changed/plugged in/or was removed. I'm pretty certain you can do this at runtime, but harder still is the api support to intelligently choose which sdk to support based on that hardware.

StephenHodgson:
@Ecnassianer can you please update this to reflect a task with a set of requirements and acceptance criteria?

SimonDarksideJ:
A prompt would be best.  As I find it actually quite annoying (especially while in OpenVR) that Unity "autostarts" the VR client when you open the editor.

But this check needs to verify it for ALL platforms you have enable in the project, and validate it when you have a new one.

Personally, I'd also like a "Update VR Settings" bool, so if I have it OFF, it stays off until I'm ready to test :D  
But have it on by default so existing HTK users get what they expect.

StephenHodgson:
A prompt might get annoying as not every project will be XR specific.

StephenHodgson:
Simon, I thought we discussed active SDK features where things set up based on the hardware detected?

SimonDarksideJ:
The prompt was more whether to turn on/off the VR switch easily.  To stop Unity being annoying. 
Based on the comment from the MRW team work work 90% in the editor, having SteamVR and / or MR Portal open EVERY time you open the editor would BE annoying :D 
Not saying the prompt should be ALL the time, just the first time you import the toolkit.

StephenHodgson:
Generally I'll keep the portal open and running if I'm testing in VR anyway.

Seems like this should be something that's automagickally handled for you.

If we were to get into "simulation" of said devices, I could see that being a toggle it switch somewhere, to turn in/off said simulator feature.

PR #2983, opened 2018-10-26
Consolidated setup prompts into one, and added VR setup

StephenHodgson:
Overview
---
Turns on VR support and consolidated optional first install prompts into one.

image

Changes
---
- Fixes: #2975
- Fixes: #2860


davidkline-ms:
Curious on this one... In HTK, we had a similar dialog that defaulted checkboxes to make each recommended change and allowed developers to opt out of individual settings.

Why not the same approach for vNext?

While each of these are good settings, it is really handy to be able to opt out of one or two without needing to find the underlying option in the Unity UI.

SimonDarksideJ:
The approach in Mrtk is slightly different to htk, in that the scriptable object config replaces most of what was configured on the htk screens.

The config settings here are only those that prevent the default operation of the MRTK (. Net/IL2CPP/VR on etc) but things that are optional.

I feel that this simple prompt which appears before Unity even loads fully, ensures that 90% of consumers starting a new project will benefit from a hassle free startup.
The other advanced users know these options and if they choose not to take them can sort it out later. 
Plus most of these settings are on one screen, the player settings. 

Aim is to ensure new users have the simplest setup experience and are welcomed in to the project and removing and manual setup or hinderence.

This is essential for the Beta, where most users will simply be trialling the experience and we can gather their feed back to improve this if need be for the Beta after that

keveleigh:
I'm not sure these are required for the default operation of the MRTK. Visible meta files and force text serialization were requirements in the legacy Toolkit so that contributions back were formatted correctly, but use of the Toolkit doesn't need them.

Enabling XR makes sense, and changing to IL2CPP is nice to have, but it'd be great to be able to pick which one(s) I want instead of being given all at once.

SimonDarksideJ:
Not possible to have choices on startup, else I works agree and putting them later when you sold be choosing before you begin just handicaps as "something else you need to do"

The Meta files options are just good practice and highly recommended for git sourced projects (as can be found in every tutorial for using unity with source control) 

Advanced users can simply select later and it won't prompt ever again and they can do it themselves. 

But for the majority of user, we want add smooth a process to adoption as possible with no barriers

In the beta, users can request more or less from this setuo, but at least then we will have usage feedback. 

Our aim for the first beta should require the fewest actions to hit play and see your connect in the headset as possible. This option reduces it to 3 clicks for a new project with no editor windows

SimonDarksideJ:
As an FYI, this is the current "new user" setup experience we are trying to improve.
https://drive.google.com/file/d/1OMlFHdbTeiMns5KYpydKGv446Ua2gwBh/view?usp=sharing

Ultimately we NEED to reduce this to a single message / box click, to get users started quickly

StephenHodgson:
Could you share the YouTube link directly?

StephenHodgson:
I can't build the dll until https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2988 is in.

I'd like to do that in a separate PR

StephenHodgson:
Just to reiterate, all I've done is consolidated the _optional_ editor settings to a simple dialogue.

These have always been part of mrtk/htk and *do not* relate to the "configuration" screens from HTK.

We didn't want to inundate the user with 5 popups right after import.

PR #2985, opened 2018-10-26
Added #if blocks to QR code preview feature

keveleigh:
Overview
---
The QR code plugin is only provided for UWP and the editor, so all references to it (and other Unity WSA specific APIs) needed to be wrapped with `#if UNITY_EDITOR || UNITY_WSA`. This also led to some minor clean-up to better wrap the contents.

Changes
---
- Fixes: https://github.com/Microsoft/MixedRealityToolkit/issues/183

### DIFF
--- a/Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
+++ b/Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
@@ -1,10 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
-using QRCodesTrackerPlugin;
 
 namespace HoloToolkit.Unity.QRTracking
 {
@@ -18,6 +16,7 @@ namespace HoloToolkit.Unity.QRTracking
         private bool updatedId = false;
 
         private SpatialGraphCoordinateSystem coordSystem = null;
+
         /// <summary>
         /// Data of the QR code to which we want to attach the game object."
         /// </summary>
@@ -39,9 +38,11 @@ namespace HoloToolkit.Unity.QRTracking
                     {
                         qRCodeData = string.Empty;
                     }
-                    
+
+#if UNITY_EDITOR || UNITY_WSA
                     qrCodeId = QRCodesManager.Instance.GetIdForQRCode(qRCodeData);
                     updatedId = true;
+#endif // UNITY_EDITOR || UNITY_WSA
                 }
             }
         }
@@ -50,9 +51,11 @@ namespace HoloToolkit.Unity.QRTracking
 
         private void Awake()
         {
+#if UNITY_EDITOR || UNITY_WSA
             QRCodesManager.Instance.QRCodeAdded += Instance_QRCodeAdded;
             QRCodesManager.Instance.QRCodeUpdated += Instance_QRCodeUpdated;
             QRCodesManager.Instance.QRCodeRemoved += Instance_QRCodeRemoved;
+#endif // UNITY_EDITOR || UNITY_WSA
         }
 
         private void Start()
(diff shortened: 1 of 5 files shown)


Assets/HoloToolkit-Preview/QRTracker/Scripts/AttachToQRCode.cs
```diff
@@ -39,9 +38,11 @@ public string QRCodeData
                     {
                         qRCodeData = string.Empty;
                     }
-                    
+
+#if UNITY_EDITOR || UNITY_WSA
```
StephenHodgson: what about `WINDOWS_UWP`?

keveleigh: An [editor plugin](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit-Preview/QRTracker/Plugins/QRCodesTrackerPlugin.Placeholder.dll) is provided, so it appears some parts of this (the ones not `WINDOWS_UWP`-specific) will work in the editor.

Of course, that's marked specifically as a placeholder, so...maybe not.

StephenHodgson: No worries, I was just curious.

Assets/HoloToolkit-Preview/QRTracker/Scripts/SpatialGraphCoordinateSystem.cs
```diff
@@ -1,66 +1,70 @@
...
+
 #if WINDOWS_UWP
 using Windows.Perception.Spatial;
 #endif
+
 namespace HoloToolkit.Unity
 {
     public class SpatialGraphCoordinateSystem : MonoBehaviour
     {
 #if WINDOWS_UWP
```
StephenHodgson: Here's one of those cases were we're not being consistent with the compiler define symbols. Will this affect the project?

keveleigh: These mean two different things. One is wrapping a Windows platform call, which will only run under `#if WINDOWS_UWP`. The other times, I'm either wrapping Unity's platform-specific APIs or references to a plugin that only exists in UWP/editor.

StephenHodgson: No worries, I just want to make sure we're using these correctly. Only difference from docs is that it's also targeting .NET Scripting Backend.

[Unity Docs](https://docs.unity3d.com/Manual/PlatformDependentCompilation.html)

UNITY_WSA | #define directive for Universal Windows Platform. Additionally, NETFX_CORE is defined when compiling C# files against .NET Core and using .NET scripting backend.
-- | --
UNITY_WSA_10_0 | #define directive for Universal Windows Platform. Additionally WINDOWS_UWP is defined when compiling C# files against .NET Core.

keveleigh: That description is slightly misleading, as that's not the difference: https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/2017.4.2.0/CodingGuidelines.md#use-platform-dependent-compilation-to-ensure-the-toolkit-wont-break-the-build-on-another-platform

StephenHodgson: That's exactly what I was looking for! 

Thanks :)

StephenHodgson: I don't think we have this section in vNEXT.

keveleigh: That looks to be true.

keveleigh: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/2990

davidkline-ms:
This PR has lingered for a long time. All feedback has been resolved.

Issue #2987, opened 2018-10-26
Inspector image assets don't load from relative path.


StephenHodgson:
**Describe the bug**
If the MRTK is installed to a different folder, or outside of the root Assets folder, then inspector texture assets do not load properly.

**To Reproduce**
1. Export MRTK
2. Import MRTK in a new project inside of a "ThirdParty" folder or similar.
3. Restart editor
4. MRTK Logo and Controller Mapping Icons do not load properly

PR #2988, opened 2018-10-26
Fixed editor texture asset loading

StephenHodgson:
Overview
---
If the MRTK is moved into a different folder, some editor texture assets would not work properly.

Renamed `EnforceEditorSettings` to `MixedRealityEditorSettings`

Changes
---
- Fixes: #2987

Blocked by: #2983


Assets/MixedRealityToolkit/MixedRealityToolkit.asmdef
```diff
@@ -2,7 +2,8 @@
     "name": "Microsoft.MixedReality.Toolkit",
     "references": [
         "Microsoft.MixedReality.Toolkit.Core.Utilities.Editor",
-        "Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities.Editor"
+        "Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities.Editor",
+        "Microsoft.MixedReality.Toolkit.Core.Utilities.Editor.Setup"
```
keveleigh: Does this assembly need to exist? It appears to contain a single file that would otherwise be captured by the `Microsoft.MixedReality.Toolkit.Core.Utilities.Editor` assembly (which is also an editor-only assembly).

StephenHodgson: Yes. Absolutely. Or else the rest of them wouldn't work.

keveleigh: I'm not asking "does it need to be added to this list?". I'm asking "does this separate assembly need to exist at all?".

StephenHodgson: If it didn't exist, then the rest wouldn't work.

Not much to explain.

keveleigh: Why wouldn't the rest work?

StephenHodgson: Go read the code and find out

Issue #2706, opened 2018-09-05
vNext: Incorrect type returned for the floor scale in the boundary visualization profile


davidkline-ms:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
```
[SerializeField]
[Tooltip("The dimensions of the floor, in meters.")]
private Vector2 floorScale = new Vector2(10f, 10f);

/// <summary>
/// The size at which to display the rectangular floor plane <see cref="GameObject"/>.
/// </summary>
public Vector3 FloorScale => floorScale;
```

**Expected behavior**
The FloorScale type to match the floorScale type.

**Actual behavior**
Type mismatch

**Mixed Reality Toolkit Release Version**
mrtk_development as of 2018/09/05

PR #2989, opened 2018-10-26
Make FloorScale's type match floorScale's type

keveleigh:
Overview
---
Changes the return type from Vector3 -> Vector2. We were only using `x` and `y` in the manager anyway.

Changes
---
- Fixes: #2706


davidkline-ms:
thank you, Kurtis!

Issue #2757, opened 2018-09-11
vNEXT: Intelligent mouse pointer switching


StephenHodgson:
# Overview
On platforms that use multiple inputs (like OpenVR or WMR) where mouse, keyboard, motion controllers, etc are available and valid inputs, we should be intelligently enabling and disabling input as they become active/inactive.

# Requirements
A list of the required functionality this task requires to implement.

* If mouse input isn't being actively used, but another is, we should stop rendering its pointer's cursor in the scene.
* If mouse input is being actively used, we should show its cursor if they're not already enabled.

# Acceptance Criteria
A list of the specific things that must be done for approval / acceptance of the task in checklist form. 
This may correspond with the above requirements.

- [x] User can switch from mouse input to motion controllers and mouse pointer/cursor should become inactive.
- [x] User can switch from motion controller to mouse input and motion controller's pointers/cursors should become inactive. (Motion controller visuals should continue to be rendered)

Issue #2754, opened 2018-09-11
vNext: four copies of the mouse pointer are visible in OpenVR if the mouse has not been moved


davidkline-ms:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
At app startup (with spatial mouse enabled) when running on OpenVR, there are four copies of the mouse pointer; one each at roughly the cardinal directions.

Once the mouse is moved, they consolidate into one.

**To Reproduce**
- Open create a scene
- Set build platform to Standalone and use the OpenVR SDK
- Launch the app in play mode
- Note the extra mouse pointers
- Move the mouse
- Pointers are consolidated

**Expected behavior**
There to be only one mouse pointer.

**Actual behavior**
There are four until physically moving the mouse

StephenHodgson:
Pretty sure there only ever is one mouse pointer. Likely the position is moving very quickly between these four points. I'll look into this.

StephenHodgson:
I can't seem to repro with your steps. Is there any additional info you can give me?

davidkline-ms:
it is possible that it is related to the combination of this and #2750. let's revisit this once that PR is complete

Issue #2330, opened 2018-06-23
Add Bluetooth mouse pointing source


dimatomp:
## Overview
A Bluetooth mouse as an input source could be a nice extension to Gaze and Gestures on HoloLens (and maybe also on other MR devices, but I cannot be sure about them). The logic could be as follows:
* When mouse is not used, the cursor points at center of the screen.
* When a mouse is connected and a move/button event occurs, cursor stops following the Gaze and changes its position further according to mouse, like in Windows Holographic shell.
* Move and Navigation for mouse can also be implemented.
* When the mouse appears to be inactive for >2sec, cursor gets back to center of the screen and follows the Gaze.

As far as I understood it, `CustomInputSource` is intended for Editor debugging purposes only, but if the above or similar functionality is already implemented, I will be glad to hear about it.

## Implementation details
According to my observations, `UnityEngine.Input.mousePosition` produces invalid and huge numbers on device; looks like the implementaion should use UWP classes directly.
The corresponding input source can rely on another pointing source and produce pointing input as well:
```
class MouseInputSource: BaseInputSource, IPointingSource
{
    [SerializeField] private GameObject basePointingSource;
    [SerializeField] private float inactivityTimeout = 2;

    // ...
}
```

StephenHodgson:
Mrtk vNext will definitely support spatial mouse input similar to the shell

dimatomp:
@StephenHodgson Great news, looking forward to seeing it in vNext.
Actually, I have already implemented the feature for an older version of MRTK. Will it be of any help if I port it to upstream and file a PR?

StephenHodgson:
That's odd, thanks for the heads up.

StephenHodgson:
> Actually, I have already implemented the feature for an older version of MRTK. Will it be of any help if I port it to upstream and file a PR?

Depends on if it follows our updated architectural patterns.

You're more than welcome to submit a PR and we can talk about what changes need to be made to support the new patterns.

PR #3003, opened 2018-10-28
Mouse Input Fixes

StephenHodgson:
Overview
---
- Fixed issues with Mouse in OpenVR
- Cleaned up a few input related classes with simpler logic.
- Added configurable mouse pointer properties
image
    - Hide mouse cursor on start.
    - Hide mouse after a few seconds if no input is detected from it.
    - Show mouse if any button is pressed / or above a set threshold.
        - When mouse is shown it's position is at gaze position
        - Mouse click to un-hide doesn't raise pointer click.

Changes
---
- Fixes: #2754
- Fixes: #2757
- Fixes: #2330


SimonDarksideJ:
Can we have a demo scene showing mouse input over controllers?  if it's not too much trouble.

davidkline-ms:
That pointer settings script has some great looking features that would be nice to demo.

StephenHodgson:
Every scene is a demo scene. We don't have demo scenes for the controllers.

StephenHodgson:
Should we also add demo scenes for motion controllers as well?

davidkline-ms:
> Every scene is a demo scene. We don't have demo scenes for the controllers.

We also don't have the threshold settings that make this a very compelling feature

davidkline-ms:
> Should we also add demo scenes for motion controllers as well?

Maybe, but this mouse script has some great customer focused settings that would be really nice to enable easy experimentation.

StephenHodgson:
Now that I think about it, I need to do more work to this PR by supporting an "always on" setting, in case someone wants that.

davidkline-ms:
Question: Is the intent to have the vertical axes inverted? I was initially puzzled as to why pulling back on the mouse moved the cursor up in the world.

StephenHodgson:
> Question: Is the intent to have the vertical axes inverted? I was initially puzzled as to why pulling back on the mouse moved the cursor up in the world.

Should also probably be customizable as well. I'll add that as well

davidkline-ms:
I'll wait for your update and then test again.

StephenHodgson:
Likely we want to be able to let users invert the mouse at the Controller mapping profile. Gotta think about this one a tick.

StephenHodgson:
> Should the default be forward == up? It feels like it.

Probably. Sounds correct to me.

StephenHodgson:
> I was initially puzzled as to why pulling back on the mouse moved the cursor up in the world.

When pushing the mouse up on the pad, the mouse should rise in front of your view as well. (Just like what you'd expect to see on screen).

The axis inversion might need to be a separate PR as it needs to be changed in the Controller Mapping Profile

davidkline-ms:
> Looks good to me. @davidkline-ms can you also test and approve?

Will do.

davidkline-ms:
I am getting some odd feeling behaviors, though that might be my inexperience with the feature.  The concerns I had with the previous version of this PR have been addressed.

Alexees:
@StephenHodgson Just wanted to use an eventdata's Pointer.TryGetPointerPosition on the mousecursor, but it seems it's returned to something's localspace as it's definitely not world space. Is it the camera I have to convert it from?

PR #3005, opened 2018-10-29
Add missing public properties to BoundingBoxRig

guitarrapc:
### Overview

Currently BoundingBoxRig missing several public properties for private SerializedField fields.
This cause user could not modify component when dynamically add component from Script. This commit open these parameters modifiable.


davidkline-ms:
@guitarrapc, thanks for the contribution!

PR #3014, opened 2018-10-29
Make IMixedRealityControllerVisualizer less implementation-specific

keveleigh:
Overview
---
Not all visualizers will be driven by a pose synchronizer. Implementations that want to be (like `MixedRealityControllerVisualizer`) are still able to implement both interfaces.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityControllerVisualizer.cs
+++ b/Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityControllerVisualizer.cs
@@ -5,7 +5,7 @@ using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.Devices
 {
-    public interface IMixedRealityControllerVisualizer : IMixedRealityControllerPoseSynchronizer
+    public interface IMixedRealityControllerVisualizer
     {
         /// <summary>
         /// The <see cref="GameObject"/> reference for this controller.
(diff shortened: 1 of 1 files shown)


davidkline-ms:
> Needs to have a design discussion

@keveleigh, can you describe a concrete example of how this is helpful?

Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityControllerVisualizer.cs
```diff
@@ -15,6 +15,11 @@ public interface IMixedRealityControllerVisualizer : IMixedRealityControllerPose
         /// </remarks>
         GameObject GameObjectProxy { get; }
 
+        /// <summary>
+        /// The current controller reference.
+        /// </summary>
+        IMixedRealityController Controller { get; set; }
```
SimonDarksideJ: I see no reason to attach a physical controller in to the abstract implementation for visualization.

The events already give access to the source controller / data.

And the current visualizer has dependencies on the SourcePoseSynchroniser

StephenHodgson:
Also let's focus on bug fixes for stabilization.

This isn't something we **need** to release on beta.

Assets/MixedRealityToolkit/_Core/Interfaces/Devices/IMixedRealityControllerVisualizer.cs
keveleigh: This is needed for a line in `BaseController`, which only references this interface: https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs#L197

SimonDarksideJ:
Re-targetted back to development for further discussion. Not a beta candidate

SimonDarksideJ:
@keveleigh do you think you could close this PR and start up a new task to discuss the requirements for this change, reasons why the current implementation is not sufficient and any other alternatives you've considered.
That is supposed to be the process for introducing change in to the framework as I understand it.

StephenHodgson:
Agreed. Also, visualization is going to need to be synced with the controller pose. I'm not sure this is a change we should be making.

PR #3022, opened 2018-10-29
Build Utilities update

StephenHodgson:
# Overview

- Opened up some of the build window methods so they can be called from third party libraries.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
+++ b/Assets/MixedRealityToolkit/_Core/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
@@ -234,7 +234,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Build
             return string.Empty;
         }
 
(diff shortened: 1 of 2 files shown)


davidkline-ms:
is this post beta or should we retarget the change?

StephenHodgson:
Depends on how we plan the release. Are we building packages from stabilization, or are we merging stabilization back into dev and then building packages?

We're using these changes in our active project, so I thought I'd bubble them up.

davidkline-ms:
> Are we building packages from stabilization

I intend to merge stabilization into release then tag and produce the packages

StephenHodgson:
Doesn't matter as I consume the MRTK as a submodule anyway.

SimonDarksideJ:
If you are actively using this @StephenHodgson in your project, then it seems fit to go in to the Beta as you are consuming this already.
Thoughts @davidkline-ms ?

StephenHodgson:
Seems like a trivial change as we're just making the accessors public.

StephenHodgson:
I'd rather keep bug fixes and beta critical things on stabilization

Issue #3021, opened 2018-10-29
If no tagged "Main Camera", the MRTK -> Configure fails badly


SimonDarksideJ:
# Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?
Mixed Reality Toolkit

# Describe the bug
If a scene does not have a main camera tagged as "Main Camera", the Mixed Reality Toolkit -> Configure command fails badly with script errors.
![image](https://user-images.githubusercontent.com/13334553/47684839-82713c00-db91-11e8-877c-6ee665dcd621.png)

The script should fall back to find a camera named "Main Camera" and if that fails it should output warning to the console.

# To Reproduce
1: create a brand new Unity project
2: Import the MRTK
3: Click Mixed Reality Toolkit -> Configure

(Yup, you guessed it, the default "sample scene" in a new unity project does not have its camera tagged as "Main Camera"

# Expected behavior
The scene should be configured correctly.
if there is no object tagged "Main Camera" it should look for an object named "Main Camera"
If still no camera is found, it should fail nicely with a warning

# Actual behavior
Script fails with console warnings about the playspace

# Unity Editor Version
2018.2.13f1

# Mixed Reality Toolkit Release Version
Stabilization

PR #3024, opened 2018-10-29
Fixed CameraCache fail if no camera tagged as main camera.

StephenHodgson:
Overview
---
- added camera lookup if no main camera tag is found.

Changes
---
- Fixes: #3021

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Utilities/CameraCache.cs
+++ b/Assets/MixedRealityToolkit/_Core/Utilities/CameraCache.cs
@@ -16,7 +16,48 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities
         /// <summary>
         /// Returns a cached reference to the main camera and uses Camera.main if it hasn't been cached yet.
         /// </summary>
-        public static Camera Main => cachedCamera == null ? Refresh(Camera.main) : cachedCamera;
+        public static Camera Main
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/_Core/Utilities/CameraCache.cs
```diff
@@ -16,7 +16,48 @@ public static class CameraCache
...
+                                {
+                                    mainCamera = Refresh(cameras[i]);
+                                    break;
+                                }
+                            }
+
+                            // If we still didn't find it, just set the first camera.
+                            if (mainCamera == null)
+                            {
+                                mainCamera = Refresh(cameras[0]);
```
SimonDarksideJ: Should also Debug.LogWarning to show the first camera has been set as main due to no "Main Camera" found

(Odd couldn't edit code in GitHub here?)

StephenHodgson: It will still throw the assert about the missing main camera tag. I was actually wondering if we should just add it for them?

SimonDarksideJ: Can't add it for them, in case the primary camera is not their main.
Best we can do is warn them to check.

PR #3032, opened 2018-10-30
Sealed MixedRealityToolkit

StephenHodgson:


davidkline-ms:
Question. Why does this class need to be sealed? There was no description on this PR (and no referenced issue).

Thanks!

StephenHodgson:
This is a pretty key & important class. Makes sense that we wouldn't want people to change its functionality.

davidkline-ms:
> Makes sense that we wouldn't want people to change its functionality.

Sealing means customers cannot extend it as well. Let's not make this change for beta and discuss pros/cons for RTW.

Seem fair?

StephenHodgson:
> Sealing means customers cannot extend it as well.

Exactly. This is critical infrastructure and no one should be extending it.

davidkline-ms:
Can we discuss this post beta? There have been enterprise customer requests to provide their own implementation of the toolkit component and it seems prudent to consider all possibilities.

Thanks

StephenHodgson:
If there's something missing from the implementation that is required, then they need to submit a PR to fulfill the requirements like any other contributor.

We've already provided a full suite of options for creating their own services and components that the Framework can ingest and utilize.

davidkline-ms:
The customer request has been to enable custom Toolkit component implementations.

We need to ensure that we understand the customer request and do our due diligence. That feels like a post beta task.

Thanks.

StephenHodgson:
If you're referring to the customer I think you are, that's an entirely different thing altogether.

They don't want to use this service locator pattern at all, which means likely it won't inherit or implement this class, which in turn makes your argument invalid, as they'd likely use their own service locator.

davidkline-ms:
The request is to discuss this post beta. Can we please have an architecture discussion.

Thank you

StephenHodgson:
Closing this PR. Has unrelated changes on the branch.

PR #3034, opened 2018-10-30
Updated async utilities to prevent possible deadlocks

StephenHodgson:
Overview
---
- Fixed a possible deadlocking situation with the async utilities.
    - (See https://github.com/svermeulen/Unity3dAsyncAwaitUtil/issues/10)
- Added async utilities to their own assembly and made critical classes internal.


Assets/MixedRealityToolkit/_Core/Utilities/Async/Internal/AsyncCoroutineRunner.cs
```diff
@@ -28,11 +30,13 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Async.Internal
     /// This Async Coroutine Runner is just a helper object to
     /// ensure that coroutines run properly with async/await.
     /// </summary>
-    public class AsyncCoroutineRunner : MonoBehaviour
+    internal sealed class AsyncCoroutineRunner : MonoBehaviour
```
davidkline-ms: why is this no longer public? wouldn't customer code benefit from this class?

Assets/MixedRealityToolkit/_Core/Utilities/Async/MixedRealityToolkit.Core.Utilities.Async.asmdef
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "Microsoft.MixedReality.Toolkit.Core.Utilities.Async",
```
davidkline-ms: why does the toolkit need a new assembly? all major components reference this new one...

do we get a benefit above having this class in the core assembly?

Assets/MixedRealityToolkit/_Core/Utilities/Async/Internal/AsyncCoroutineRunner.cs
StephenHodgson: Because nothing outside of this assembly should be calling on this.
No one should be using this class except for the utilities. That's why it's in the internal folder.

Assets/MixedRealityToolkit/_Core/Utilities/Async/MixedRealityToolkit.Core.Utilities.Async.asmdef
StephenHodgson: Yes, it's one less thing that has to recompile, and adds a bit of security to the async utilities, as there were a few public methods and classes that should have been internal when I initially wrote it.

Assets/MixedRealityToolkit/_Core/Utilities/Async/Internal/AsyncCoroutineRunner.cs
davidkline-ms: wouldn't there be customer benefit from having a solid / deadlock safe async runner available?

StephenHodgson: Yes and that's exactly what they get with this change.

davidkline-ms: Should have been more clear in the question... Wouldn't it benefit customer code to be able to directly use this class?

StephenHodgson: No, as they shouldn't be directly using this class, because it's an implementation detail of the overall utility.

It's to protect the integrity of this system and its function.

PR #3046, opened 2018-10-31
update version to 2018.9

davidkline-ms:
Overview
---
Updates the project and text files to the beta release version


StephenHodgson:
Just a quick question. From our new release cadence, so the second value no longer corresponds to the month release, correct?

davidkline-ms:
> Just a quick question. From our new release cadence, so the second value no longer corresponds to the month release, correct?

Correct, it is simply an incrementing value. Starting with the 2019 MRTK, we will begin at 0 and go as high as required.

To refresh, the versioning system is:

<product>.<release>.<update>

For example: 2019.1.0

Where an increase in product number carries a *high* likelihood of breaking changes, an increase in release indicates there *may* be breaking changes and an update increase is an *extremely low* change of breaking changes.

PR #3066, opened 2018-11-02
Mixed Reality Toolkit Standard Shader New Features

Cameron-Micka:
Overview
---
This is an update to the MRTK/Standard shader and systems. The primary change is the addition triplanar mapping. Triplanar mapping is a technique to programmatically texture a mesh. Often used in terrain, meshes without UVs, or difficult to unwrap shapes. This implementation supports world or local space projection, the specification of blending smoothness, and normal map support. Note, each texture used requires 3 texture samples, so please use sparingly in performance critical situations.

triplanar

A checkbox has also been added to control albedo optimizations. As an optimization albedo operations are disabled when no albedo texture is specified. To control this (as requested by this blog post: http://dotnetbyexample.blogspot.com/2018/10/workaround-remote-texture-loading-does.html) Simply check this box:

albedoassignment

The lighting model has been tweaked to match the Unity standard shader a little closer. And, the example comparison scene has been updated to reflect this:

matrixcompare

Fresnel lighting on back facing polygons should now look more correct. (Thank you Gerrit Lochmann for the fix).

Finally, normal map scale is now supported and a few extra material gallery examples have been added.

newmaterials


Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Materials/ShaderBallTriplanar.mat
```diff
@@ -6,142 +6,99 @@ Material:
   m_ObjectHideFlags: 0
   m_CorrespondingSourceObject: {fileID: 0}
   m_PrefabInternal: {fileID: 0}
-  m_Name: ShaderBallGlass
+  m_Name: ShaderBallTriplanar
```
davidkline-ms: Did you mean to remove the glass material?

Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Materials/ShaderBallTriplanar.mat
```diff
@@ -6,142 +6,99 @@ Material:
...
-        m_Scale: {x: 1, y: 1}
-        m_Offset: {x: 0, y: 0}
-    - _ParallaxMap:
-        m_Texture: {fileID: 0}
-        m_Scale: {x: 1, y: 1}
-        m_Offset: {x: 0, y: 0}
     m_Floats:
     - _AlbedoAlphaMode: 0
-    - _AlbedoAlphaSmoothness: 0
+    - _AlbedoAssignedAtRuntime: 0
```
davidkline-ms: Does this need to get updated in all of the checked in materials?

StephenHodgson: Likely yes as it was renamed in the shader

Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Scripts/MaterialMatrix.cs
```diff
@@ -66,32 +71,48 @@ public void BuildMatrix()
...
                     Material newMaterial = new Material(material);
                     newMaterial.SetFloat(firstPropertyId, firstProperty);
                     newMaterial.SetFloat(secondPropertyId, secondProperty);
 
-                    Renderer renderer = sphere.GetComponent<Renderer>();
+                    Renderer renderer = element.GetComponent<Renderer>();
+                    MeshFilter meshFilter = element.GetComponent<MeshFilter>();
 
-                    if (Application.isPlaying)
+                    if (Application.isPlaying == true)
```
StephenHodgson: Sorry, being nit picky, but do we really need to make this change? It's redundant.

Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Materials/ShaderBallTriplanar.mat
Cameron-Micka: I did remove it and replaced it with the triplanar one since I thought the glass one was a little "meh" looking. But, I'm more than happy to add it back if you think it is useful?

Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Scripts/MaterialMatrix.cs
Cameron-Micka: I agree haha, this is our teams new coding style so I must have done it subconsciously. But, I happy to revert it.

Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Materials/ShaderBallTriplanar.mat
Cameron-Micka: Oh no it should be fine. Unity's material system will just resort to the default value which is good by default. It also keep around old properties that don't have to do with newly assigned materials, which seems odd, but can be a lifesaver when you switch back to a previous shader and your settings are the same.

Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Materials/ShaderBallTriplanar.mat
davidkline-ms: the nice thing about having the glass material is it shows customers (like me :) ) who may need an example, how to achieve a glass effect,

StephenHodgson: Glass material is still there, correct? Just replaced with a better one?

Cameron-Micka: That's a very good point! I'll add it back in, with a few more material examples (to satisfy my need for uniformly organized shader balls).

Cameron-Micka: Alright, I've added the glass back with a few extra examples (please see the last gif in the overview).

Assets/MixedRealityToolkit/_Core/Resources/Shaders/MixedRealityStandard.shader
```diff
@@ -501,12 +527,33 @@ Shader "Mixed Reality Toolkit/Standard"
...
+
+#if defined(_TRIPLANAR_MAPPING)
+                // Calculate triplanar uvs and apply texture scale and offset values like TRANSFORM_TEX.
+                fixed3 triplanarBlend = pow(abs(i.triplanarNormal), _TriplanarMappingBlendSharpness);
+                triplanarBlend /= dot(triplanarBlend, fixed3(1.0, 1.0, 1.0));
+                float2 uvX = i.triplanarPosition.zy * _MainTex_ST.xy + _MainTex_ST.zw;
+                float2 uvY = i.triplanarPosition.xz * _MainTex_ST.xy + _MainTex_ST.zw;
+                float2 uvZ = i.triplanarPosition.xy * _MainTex_ST.xy + _MainTex_ST.zw;
+
+                float3 axisSign = i.triplanarNormal < 0 ? -1 : 1;
```
xwipeoutx: cg has a `sign` function which should be a bit faster and more idiomatic than this ternary http://developer.download.nvidia.com/cg/sign.html

This will have the added advantage of setting the `uv.x` values (below) to `0` when that component of the normal is 0 - essentially increasing the chance of a cache hit on the texture lookup when the normals match the basis vectors.

I've not profiled either of these :)

Cameron-Micka: Thank you for pointing this out. My gut reaction was to use sign too but I don't ever want a "axisSign" of zero, so I wasn't sure of a better way to do this?

xwipeoutx: I don't have a problem with a sign of 0 in this case - its impact on the tri planar blend is exactly nothing, so it makes sense to me for its sign to match that. Any other choice is equally arbitrary IMHO.

Especially if it does give perf benefits - add a comment if you're worried about confusion.

PR #3068, opened 2018-11-03
[vNext][UX] Feature/mrtk interactable layout3D and InspectorFields

killerantz:
Overview
Moved  layout code out of Interactables and created a demo scene
Moved InspectorFields out of Interactables and created a demo scene
Moved some inspector helper code to InspectorUIUntility
Created receiver components to allow handling interactable state changes without Unity Events


Assets/MixedRealityToolkit-Examples/Demos/UX/Layout3D/Scripts/AutoCycletLayout3DSize.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SDK.UX.Layout3D
+{
+    public class AutoCycletLayout3DSize : MonoBehaviour
```
cre8ivepark: Cyclet?

radicalad:
@killerantz  this is super cool. Can we groupt these layout tools with ObjectCollection?

Assets/MixedRealityToolkit-Examples/Demos/UX/Layout3D/Scripts/AutoCycletLayout3DSize.cs
killerantz: Thanks for catching that, I'll update shortly.

killerantz:
@radicalad The grid would make sense in ObjectCollection for sure, the other items are more about aligning elements to build an object. I'm not sure most of them fit as a collection.

cre8ivepark:
Release page shows 'Requires Unity 2018.1 or later' but Recommended version: 2018.2.14f1
Considering the fact that prefabs are broken when saved with 2018.2.x and opened with 2018.1.x, it would be great if we prepare these with 2018.1.x.

StephenHodgson:
Most of the 2018.2.x builds don't work with the MRTK. I think recommending the latest for now is a good idea.

SimonDarksideJ:
I've stated 2018.2.13f1 as a minimum in all my content so far. (which was the first 2018.2 build it actually worked.)
2018.2.1 - 2018.2.12 DO NOT WORK

killerantz:
I fixed the prefabs, there's some property names that get changed in the prefab and scene files from 2018.1 to 2018.2, so if we update the prefabs, we should only do it in the earlier versions unless we are standardizing on the latest. Confirmed they work in 2018.1.9f2 and 2018.2.14f1

Issue #3067, opened 2018-11-02
vNext Task: Add diagnostic window demo scene


davidkline-ms:
# Overview
The diagnostic window is intended to provide developers with help to determine the runtime performance of their experience without needing the overhead of a traditional profiler.

There should be a demo scene that shows how to control the data being displayed.

# Requirements
A list of the required functionality this task requires to implement.

* Demo scene provides speech commands to show the diagnostic display window.
* Demo scene provides speech commands to show:
    * CPU statistics
    * Frame Rate
    * Memory Usage

# Acceptance Criteria
- [x] Diagnostics window settings and window visibility can be changed

Issue #3055, opened 2018-11-02
Diagnostic view does not enable/disable when project is running.


SimonDarksideJ:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
If the Diagnostics are set to visible while the project is running, the view does not display

**To Reproduce**

* Enable diagnostics in the configuration profile
* Create a new Scene
* Click Mixed Reality Toolkit > Configure
* Enter play mode.
* Set Visible to True in the diagnostics profile.

**Expected behavior**
The diagnostics window will appear
(correspondingly, if turned "off" it should disappear)

**Actual behavior**
Nothing happens

**Unity Editor Version**
2018.2.13f1

**Mixed Reality Toolkit Release Version**
2018.9 Beta

davidkline-ms:
Best of my understanding is that profiles are read at system startup. Code that wishes to change values during runtime should be talking to the system through its interface.

I believe the issue here is that the diagnostic system defined read-only properties instead of allowing runtime changes.

I'll make the appropriate changes.

SimonDarksideJ:
Your basis is incorrect @davidkline-ms , all systems and profiles are designed to be able to be configured while the project is running. Also the scriptable config system has been designed to react when the user swaps out a profile at runtime.

The only exception to this is the InputActions profile, due to the ID's used can't be changed while the project is running.

davidkline-ms:
The systems that I am familiar with read the profile during Initialization and the app code references the runtime class (via the interface) properties to access. Is there an example where the system interface does not expose the profile settings via a property and requires client code to access the profile directly?

It feels like that is an undue burden on the clients when the systems can easily expose the profile properties (similar to the spatial awareness system).

StephenHodgson:
> The systems that I am familiar with read the profile during Initialization and the app code references the runtime class (via the interface) properties to access. 

This isn't true... We explicitly load from profile data. We're data driven by design.

keveleigh:
Do you have an example of a system implementation that continues to read from the profile after initialization?

davidkline-ms:
Is there an example where the system interface does not expose the profile settings via a property and requires client code to access the profile directly?

davidkline-ms:
One thing to note, if the systems/client code make changes and directly access the profile at runtime, that profile is altered which could lead to incorrect state the next time the app is run....

[...]

PR #3079, opened 2018-11-05
update diagnostics system interface to add setters, add diagnostics demo scene

davidkline-ms:
Overview
---
This change adds the ability to customize the diagnostics window display at runtime by adding setters to the interface properties (the concrete implementation already had setters implemented),

Added diagnostics demo scene to illustrate how to toggle the display components. This scene uses custom profiles for diagnostics visualization, input actions and speech commands.

Interface change is being marked as a breaking change since it was possible to have implemented a concrete class without the property setters.

Changes
---
- Fixes: #3055 
- Fixes: #3067  .


StephenHodgson:
Wasn't sure how this was a breaking change. Does the namespace need to be updated with the move?

davidkline-ms:
> Wasn't sure how this was a breaking change. Does the namespace need to be updated with the move?

It is technically breaking in that any code that did not previously implement setters would no longer compile.  No namespace changes were needed.

Issue #3077, opened 2018-11-05
vNext customer feedback: too much debug log spam when a system isnt running


davidkline-ms:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
We have received direct feedback that when a system (ex: Teleport) is disabled, there can be a great deal of "debug log spam" when system and/or client code asks for the system (whenever they are requested, they attempt to lazy-initialize).

**To Reproduce**
1. Disable the Teleport system in the configuration profile
2. Enter play mode

**Expected behavior**
Minimal Debug.Log entries regarding the lack of a teleport system.

**Actual behavior**
An over abundance of Debug.Log entries that can, at times, hide log messages from the application.

**Screenshots**

**Unity Editor Version**
Which Unity version(s) you've been able to reproduce the bug in.

**Mixed Reality Toolkit Release Version**
Which MRTK version(s) you've been able to reproduce the bug in.

**Additional context**

StephenHodgson:
Not sure this constitutes a bug report, but yes systems shouldn't complain if something isn't immediately turned on.

Do we have a list of debug logs that need to be addressed?

StephenHodgson:
> Customer feedback needs to be captured. :)

Agreed. Just surprised to see feedback as a bug report. 

Likely we'll need to add a flag in the static system property accessors that can silence after the first null return, and reset the flag after a valid one is found.

StephenHodgson:
The other issue is that they failed to also remove any pointers/assets that are using the Teleport system.

StephenHodgson:
Makes sense that they're getting logs about a system that doesn't exist if assets and things in the scene are searching for them.

PR #3086, opened 2018-11-06
Updated Service Locator logging

StephenHodgson:
Overview
---
- Reduced service searching logs.
- Misc formatting.
- Added better handling for searching for services when shutting down.

Changes
---
- Fixes: #3077

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
@@ -30,6 +30,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
 
         private bool isInitializing = false;
 
+        private static bool isApplicationQuitting = false;
+
         /// <summary>
         /// Checks if there is a valid instance of the MixedRealityToolkit, then checks if there is there a valid Active Profile.
         /// </summary>
@@ -374,6 +376,12 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
 #if UNITY_EDITOR
                 UnityEditor.EditorApplication.playModeStateChanged += playModeState =>
                 {
+                    if (playModeState == UnityEditor.PlayModeStateChange.ExitingEditMode ||
+                        playModeState == UnityEditor.PlayModeStateChange.EnteredEditMode)
+                    {
+                        isApplicationQuitting = false;
+                    }
+
                     if (playModeState == UnityEditor.PlayModeStateChange.ExitingEditMode && activeProfile == null)
                     {
                         UnityEditor.EditorApplication.isPlaying = false;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -593,7 +611,7 @@ public IMixedRealityService GetService(Type type)
                 GetService(type, out service);
             }
 
-            if (service == null)
+            if (service == null && showLogs)
             {
                 Debug.Log($"Unable to find {type.Name}.");
```
StephenHodgson:
```suggestion
                Debug.LogWarning($"Unable to find {type.Name}.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -550,21 +560,28 @@ public bool RegisterService(Type type, IMixedRealityService service)
...
         /// <returns>The Mixed Reality Toolkit of the specified type</returns>
-        public IMixedRealityService GetService(Type type)
+        public IMixedRealityService GetService(Type type, bool showLogs = true)
         {
+            if (isApplicationQuitting)
+            {
+                return null;
+            }
+
             if (ActiveProfile == null)
```
davidkline-ms: need to check showLogs here and other similar calls

StephenHodgson: Why? These are much more important than the other where we check. This sort of thing is pretty serious and we should def always show this error no matter what.

davidkline-ms: Is that not the purpose of the showLogs argument? If a customer passes in "showLogs == false" it seems as if the method should not log.

This is where something like the logging level task/proposal can address the "much more important". It gives customers a finer grain control over the types of messages received.

davidkline-ms:
> Proposed changes are outside the scope of the PR and have much bigger implications.

How so? A customer asking for no logging should receive no logging.

davidkline-ms:
The contract specified in this PR appears to be that setting showLogs to false will stop log messages. As implemented this is not the case, hence the confusion. How is this outside of the scope?

Let's take this offline and discuss in a smaller group. Historically. giving customers control over which specific log messages to see has been a net positive. It would be helpful to be able to pro/con an "all off" vs "log level" solution.

StephenHodgson:
> How is this outside of the scope?

Because this PR is addressing the specific issue of log "spamming" as described in #3077. Anything outside of that is a further discussion, proposal, and task.

StephenHodgson:
I'm tired of dragging out Pull Requests like this out over something we should be discussing and formally proposing and creating tasks for. We can circle back to the concerns you have after we've agreed on the best path forward. Let's not bog down the development process with petty politics.

StephenHodgson:
Let's get this pull request completed and address the logging levels in a separate subsequent pull request.

Issue #3098, opened 2018-11-08
MixedRealitySpatialAwarenessSystem inappropriately calls Destroy on SpatialAwarenessObserver


StephenHodgson:
**Describe the bug**
The `MixedRealitySpatialAwarenessSystem` inappropriately calls Destroy on `SpatialAwarenessObserver` inside its [Destroy method](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs#L182).

Destroy should only be called from the service locator as it cleans up the extension services.

PR #3099, opened 2018-11-08
Removed inappropriate destroy call from spatial awareness system

StephenHodgson:
Overview
---
Removed inappropriate destroy call from spatial awareness system

Changes
---
- Fixes: #3098


Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -177,9 +177,6 @@ public override void Destroy()
                     }
                     surfaceParent = null;
                 }
-
-                // Tell the observer to clean up
-                SpatialAwarenessObserver?.Destroy();
```
davidkline-ms: Presumably this is not required since the core toolkit component will call Destroy on the observer, correct?

StephenHodgson: That is correct.

PR #3102, opened 2018-11-08
Easier Extension Service Implementations and a bit of cleanup

StephenHodgson:
Overview
---
Added `BaseExtensionService` class for easy creation of extension service implementations.
Cleaned up the existing extension services with new BaseExtensionService class.

Always feels good to get rid of duplicated code :)

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Services/BaseService.cs
+++ b/Assets/MixedRealityToolkit/_Core/Services/BaseService.cs
@@ -9,10 +9,10 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
     public class BaseService : Interfaces.IMixedRealityService
     {
         /// <inheritdoc />
-        public virtual string Name { get; set; }
+        public virtual string Name { get; protected set; }
 
         /// <inheritdoc />
-        public virtual uint Priority { get; set; } = 5;
+        public virtual uint Priority { get; protected set; } = 5;
 
         /// <inheritdoc />
         public virtual void Initialize() { }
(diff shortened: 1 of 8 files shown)


StephenHodgson:
> before we start properly documenting how to create services

Exactly :)

Assets/MixedRealityToolkit/_Core/Services/BaseService.cs
```diff
@@ -9,10 +9,10 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
     public class BaseService : Interfaces.IMixedRealityService
     {
         /// <inheritdoc />
-        public virtual string Name { get; set; }
+        public virtual string Name { get; protected set; }
 
         /// <inheritdoc />
-        public virtual uint Priority { get; set; } = 5;
+        public virtual uint Priority { get; protected set; } = 5;
```
davidkline-ms: Curious as to why 5.

davidkline-ms: Should we consider making an enum that defines specific priorities? Similar to how we have defined the menu index enum?

StephenHodgson: I'd rather not use an enum here as it removes the flexibility of the developer to determine the priority in a fine tuned, or fine grained way.

I'm not sure why we choose 5, but it was a happy middle ground between 1 (Which the input system is bc it's critically important that it's initialize first) and 10, which is what all of the extension services are registered at. 

This priority determines which order Initialize, Enable, Update, Disable, and Destroy are called on each service.

davidkline-ms: Seems reasonable.

Issue #2702, opened 2018-09-04
vNEXT: Refactor Focus Provider to be a service


StephenHodgson:
# Overview
If it's possible to make the Focus Provider it's own manager implementation, it would help with making sure it is initialized before the input system so that Pointer Registration is cleaner.

# Alt1
Remove the class completely and have the Input Manager absorb it entirely.

PR #3113, opened 2018-11-10
Refactored focus provider to service

StephenHodgson:
Overview
---
- Refactored the Focus Provider into a registered service.
- Added Focus provider selector to Input System Profile.
image
- Moved Focus/Gaze provider settings to Pointer Profile settings (as these settings directly affect pointers specifically)
image
- Updated all scenes to ensure they work properly with the new focus provider service.

Changes
---
- Fixes: #2702
- Fixes: #3125
- Fixes: #3168

Breaking Changes
---
- `IMixedRealityFocusHandler` no longer depends on the `IMixedRealityFocusChangedHandler`
- The old `FocusProvider` component in the scene will now be missing from the main camera game object. This will need to be manually removed in the scene.
- `FocusProvider` service will need to be added to any custom extension service profiles.
- `EventSystem` check was moved from Input System to the MixedRealityToolkit, because we use a generic base event system for all our messaging from most of our services. This just makes sure to put it a standardized location. The component was moved from the ui raycast camera to the main camera.
- `StandaloneInputModule` was also moved to the main camera from the ui raycast camera.
- Users who have custom pointer profiles will need to update them with new settings for gaze provider.

Steps to address braking changes
---
1. Open the scene with old Focus Provider
2. Select the main camera game object
3. Remove the `FocusProvider` (empty component, with a warning on it)
4. Delete the `UiRaycastCamera` GameObject if one exists under the main camera's GameObject.
5. Select the `MixedRealityToolkit` GameObject, then the configuration profile.

### If using a custom profile do the additional steps:

1. Navigate to the Registered Extension Services profile
2. Register a new service.
3. Name it `Focus Provider`
4. Select the dropdown and set the `Microsoft.MixedReality.Toolkit.SDK.Input.FocusProvider`
5. Save changes

A few lingering questions
---
- Should the Focus Provider be a core system? (As we require that there only be one)
    - Even though this would be a core system, it's requires the input system, so the type set should be in the input system profile with the focus profile.
   - Added [Proposal](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3127) for a way to specify extension services that cannot be registered more than once.

Blocked by
---
- [x] #3112
- [x] #3123
- [x] #3126 

TODO
---
- [x] Update profile configuration docs with reflected changes.
- [x] #3216
- [x] #3218


Assets/MixedRealityToolkit-SDK/Features/Diagnostics/System/MixedRealityDiagnosticsManager.cs
```diff
@@ -17,12 +17,8 @@ public class MixedRealityDiagnosticsManager : BaseEventSystem, IMixedRealityDiag
 
         public override void Initialize()
         {
-            base.Initialize();
-            InitializeInternal();
-        }
+            if (!Application.isPlaying) { return; }
```
SimonDarksideJ: Are these changes from another PR?

StephenHodgson:
Need to split out changes to unreleated systems and core service locator.

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/System/MixedRealityDiagnosticsManager.cs
SimonDarksideJ: Please revert changes to the Diagnostics system as they are not in the scope of a Focus Provider change.  Please resubmit as a separate PR.

Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Materials/ShaderBallClippingPlane.mat
```diff
@@ -4,12 +4,12 @@
 Material:
   serializedVersion: 6
   m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
```
SimonDarksideJ: Please revert changes to these materials, we should not include auto-updated files by Unity. They need to be tested in the scope of a Standard Shader PR

Assets/MixedRealityToolkit-Examples/Demos/StandardShader/Materials/ShaderBallClippingPlaneInner.mat
```diff
@@ -4,12 +4,12 @@
 Material:
   serializedVersion: 6
   m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
```
SimonDarksideJ: Please revert changes to these materials, we should not include auto-updated files by Unity. They need to be tested in the scope of a Standard Shader PR

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -220,34 +198,81 @@ public override int GetHashCode()
...
             if (uiRaycastCamera == null)
             {
-                Debug.LogWarning("No UIRaycastCamera assigned! Creating a new UIRaycastCamera.\n" +
-                                 "To create a UIRaycastCamera in your scene, find this Focus Provider GameObject and add one there.");
-                CreateUiRaycastCamera();
+                EnsureUiRaycastCameraSetup();
             }
 
-            await WaitUntilInputSystemValid;
+            if (MixedRealityToolkit.InputSystem == null)
```
SimonDarksideJ: Can we break these calls in to a generic function, I hate duplicated code :D

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -220,34 +198,81 @@ public override int GetHashCode()
...
                 RegisterPointers(inputSource);
             }
         }
 
-        private void Update()
+        /// <inheritdoc />
+        public override void Update()
         {
-            if (MixedRealityToolkit.InputSystem == null) { return; }
+            if (MixedRealityToolkit.InputSystem == null)
```
SimonDarksideJ: Can we break these calls in to a generic function, I hate duplicated code :D

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
StephenHodgson: If you can figure out a way to do it, sure. I'm accepting PRs.

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
StephenHodgson: Same comment as above

Assets/MixedRealityToolkit-SDK/Features/Diagnostics/System/MixedRealityDiagnosticsManager.cs
StephenHodgson: Already planning on it.

StephenHodgson:
Question for everyone. 

I'm considering instead of making a new "Focus" profile, just moving all these settings into the pointer profile.

They all are pointer specific settings, such as global pointer extent and debug settings...

StephenHodgson:
Okay this PR is ready for review.

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -61,25 +96,36 @@ public Camera UIRaycastCamera
...
+                {
+                    Debug.LogError($"Unable to start {Name}. An Input System Profile is required for this feature.");
+                    return false;
+                }
 
-        [SerializeField]
-        private Color[] debugDrawPointingRayColors = null;
+                if (MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.PointerProfile == null)
+                {
+                    Debug.LogError($"Unable to start {Name}. An Focus Profile is required for this feature.");
```
StephenHodgson:
```suggestion
                    Debug.LogError($"Unable to start {Name}. An Pointer Profile is required for this feature.");
```

davidkline-ms:
If a service is optional, removing it should not result in exceptions.

davidkline-ms:
Scenario relating to removing the focus service:

An app is an on-rails experience that implements only global controller / speech command inputs.

StephenHodgson:
The service isn't optional if the input system is enabled.

StephenHodgson:
That's part of the question on if we wanted to make it a core system or not and disabling the input system if a focus provider isn't registered.

I also thought about having the input system completely absorb the focus provider and have it register the dependency on startup.

StephenHodgson:
Your test is the equivalent of removing the focus provider component on the main camera in the current version. Which would likely also break.

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealitySpatialObserver.cs.meta
```diff
@@ -5,7 +5,7 @@ MonoImporter:
   serializedVersion: 2
   defaultReferences: []
   executionOrder: 0
-  icon: {instanceID: 0}
+  icon: {fileID: 2800000, guid: 961230b29c294bb780054c5d02eb6180, type: 3}
```
StephenHodgson:
```suggestion
  icon: {instanceID: 0}
```

davidkline-ms:
> The service isn't optional if the input system is enabled.

If that is the case, then why should it be a service? Are there platform specific implementation needs?

davidkline-ms:
> disabling the input system if a focus provider isn't registered.

There are scenarios where input is used and focus is not required. Should there be such a strong tie between focus and input?

StephenHodgson:
> There are scenarios where input is used and focus is not required

No, and it was intentionally designed to be this way.

> Should there be such a strong tie between focus and input?

Absolutely. Yes. How else would you be able to select a GameObject?

Event unity's base pointer has a `selectedObject` property, and the only way we set it is via pointer events via the focus provider.

Without the focus provider the input system will not function. at. all. The current version (both in the HTK and MRTK) both rely on some sort of way to provide focus via raycasting and without it would also fail.

A "Rails" type of scenario would likely disable the input system completely.

davidkline-ms:
> No, and it was intentionally designed to be this way.

Why not? We have had multiple instances of feedback from customers indicating that if they do not use a portion of our system(s) they would prefer them to be completely disabled. 

> How else would you be able to select a GameObject?
Event unity's base pointer has a selectedObject property, and the only way we set it is via pointer events via the focus provider.
Without the focus provider the input system will not function. at. all. 

The described scenario uses only global input and does not rely upon a game object receiving focus

> A "Rails" type of scenario would likely disable the input system completely.

Not necessarily, a rails scenario could involve voice and controller input that is global in nature (ex: press the menu button to select a new scene.)

davidkline-ms:
Please do not misunderstand the feedback, I am 100% supportive of making focus a system.

davidkline-ms:
In fact, I am considering making gaze an optional system as well.

keveleigh:
>Without the focus provider the input system will not function. at. all. 

The input system has four types of objects to route events to (global, modal, focused, fallback). Focused objects are only one of the four, and the system will function without them.

StephenHodgson:
>  Focused objects are only one of the four,

Right and if we're missing one of the four modal inputs then it won't work.

StephenHodgson:
Focus has never been "optional" is a much bigger breaking change that I would like to discuss as a separate proposal.

Imho this isn't something we can make optional with our input system. That's why the `IMixedRealityFocusProvider` and `IMixedRealityGazeProvider` are both part of the `IMixedRealityInputSystem` because they're _required_.

Making this change has far reaching implications I do not with to try to address in this PR.

StephenHodgson:
One important reason why we can't remove the focus provider is because _every_ input source needs to have at least one pointer (gaze is always the fallback if a pointer isn't assigned) the pointers require the focus provider to update their states by sending the focus events to them.

davidkline-ms:
> One important reason why we can't remove the focus provider is because every input source needs to have at least one pointer

Does the system require there to be at least one input source? Seems it should be able to handle situations where all sources are lost or none have yet been found.

StephenHodgson:
> Does the system require there to be at least one input source? 

None that I'm aware of, but none the less, the gaze input source is always availible because we guarantee it as a fallback pointer. It's always known here the forward vector from the main camera is any any given moment, so it's the one input source we can guarantee 100% of the time to have.

wassx:
I love how the customization of an own gaze provider works 👍 
Still I have troubles adding the focus provider. It does not appear in the list of component types (drop down) 🤔

StephenHodgson:
The focus provider service type drop-down was moved to the input system's configuration profile.

It should now be the first field.

Let me know if it's now showing up there

StephenHodgson:
The gaze provider type is in the pointer profile.

wassx:
Ok thx, found both and set the focus provider and the custom gaze provider. Still I can't see any focus cursor. Where will the focus provider be in the scene? Can't find it anymore. Gaze provider is on the head again.

StephenHodgson:
The gaze cursor can also be set in the pointer profile.

StephenHodgson:
The gaze provider is correctly set in the scene, using the type you specified in the pointer profile

StephenHodgson:
Thanks Kurtis, I'll check these issues out

StephenHodgson:
I'll just add the unit tests later

davidkline-ms:
@StephenHodgson, looks like some merge conflicts have arisen... Can you take a look? Reviewing the breaking change PRs today.

StephenHodgson:
Will do

StephenHodgson:
I would like to try and add some unit tests to this now that it's been merged in.


Any feedback and review now would still be helpful.

StephenHodgson:
Also need to address #3218 as it's only utilizing this PR

StephenHodgson:
Okay fixed the unit tests. We can start reviewing this PR as well.

StephenHodgson:
Okay fixed

Issue #2540, opened 2018-08-02
vNext Task: Structure for SDK packages


davidkline-ms:
## Overview
The decision was made to limit the contents of the Core package to assemblies and code that define interfaces and types (plus the Mixed Reality Manager) as managers and profiles need to live in the same layer.

SDK packages should be structured as to make it clear what can safely be unchecked when importing.

## Proposed package layout
- MixedRealityTookit-SDK
     - Features
        - Input
            - System (implementation files, handlers, inspectors, etc)
            - Components
               - Prefabs
               - Scripts

## Requires
Moving systems / managers from Core to their appropriate SDK feature folder
Formatting SDK feature folders per proposed layout (after discussion and agreement)
Prototype Core and SDK packages (one per SDK feature area)

## Acceptance criteria
- [ ] Project can be built and does not break when minimal SDK packages installed

PR #3114, opened 2018-11-11
SDK Feature re-org

StephenHodgson:
Overview
---
- removed unnecessary folders in spatial awareness and diagnostics feature folders
- updated system file icons


StephenHodgson:
Because we're not doing that literally anywhere else. This is for consistency, and the folder is for all the system level scripts anyway. Anything else shouldn't be in the folder to begin with.

davidkline-ms:
It had been previously proposed (#2540) and discussed in shiproom (and originally checked in for Boundary) to have a structure that resembled:

Feature
  \ Prefabs
  \ Scripts
  \ < any additional folders >
  \ System

For the SDK. This is intended to help make it clear for customers importing SDK packages which folders mapped the higher level components and which were the systems as well has representing each SDK package with a single folder containing everything available for that feature (easy import / easy delete).

Issue #2957, opened 2018-10-23
Global event listeners and speech input handlers


Bazilikum:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
HoloToolkit 2017.4.1.0 

**Describe the bug**
I'm not sure this is a real bug, but I would like to discuss this issue.
A speech global listener that is also a button will be a global listener also for all click events.

**To Reproduce**
Create an object that has a component which implements IInputClickHandler.
Attach a SpeechHandler to it and set it to be a global listener.
The click handler is also global and responds to any click in the scene.

**Expected behavior**
Be a global speech handler but not a global click handler.

**Actual behavior**
Object responds to all events globally.

**Unity Editor Version**
2018.2.2f1

**Mixed Reality Toolkit Release Version**
HoloToolkit 2017.4.1.0 

**Additional context**
When a button is required to also have a voice command to trigger it, and the button might be instanced during the runtime of the scene.
We solved the problem by separating the global speech handler, making it a child of the button and connecting the speech events to the triggering of the button.
This solution is a bit messy.
It also makes me wonder about how events are handled in relation to focused objects and if it is the responsibility of the event system or the event receiver to check if the object is being focused.
If I'm not mistaken usually in unity if you want to check for a mouse click or keyboard status its usually the object's responsibility to check "focus".
I would love to hear your opinion and why you chose the way you did.

PJBowron:
Unfortunately, by adding itself as a global listener all a component does is simply say: "give me first crack at ALL events from now on!"
A lot of things believe they should be global listeners when actually they'd operate more considerately as fallback handlers :)

StephenHodgson:
Yeah, and unfortunately there's not a way to specify a _specific_ component as the event is [routed on via GameObject reference](https://docs.unity3d.com/Manual/MessagingSystem.html).

Bazilikum:
Thank you guys for the information and help!
I'm wondering what is the benefit of a system that is designed this way?
Is there a way for us to change it or propose something to mitigate this?
Thanks again!

Bazilikum:
Based on this issue, I would be interested to hear your opinions about an update to the event system where global listeners are seperated by the handler type, so an object can be a global listener on one handler and not on another (for example voice commands vs. click).
As far as I understand how the system is designed we could update the global listeners list to be a dict where the key is the type.
Is that something you people believe makes sense?
Thanks!

[...]

Issue #2607, opened 2018-08-17
vNext Task: New Interactable Object


cre8ivepark:
## Overview
Add new button UI control system which unifies old [CompoundButton ](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit-Examples/UX/Readme/README_InteractableObjectExample.md)and [InteractiveButtonComponents](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/master/Assets/HoloToolkit-Examples/UX/Readme/README_InteractiveButtonComponents.md)

# Requirements
A list of the required functionality this task requires to implement.

* A interaction system that is easy for designers to create buttons and other interactables with sensory feedback.
* An easy to use inspector that ties everything together
* States for interactables are configurable and easily extended.
* Profiles are a collection of themes paired to game objects.
* Themes allow the designer to assign object property values to states.
* Events are easily extendable at usable in the Inspector.
* Events can also be used by developers using code in optimized way.
* Events may care about states that themes do not.
* Toggle interaction and theming is supported
* Action filtering
* Keyword or voice commands
* Global listening can be enabled
* A collection of basic themes
* A collection of basic events
* A collection of button prefabs to show examples of interaction types

# Acceptance Criteria

- [ ] Interactable assets including models, materials, textures, animations
- [ ] Inspector base code for layout, buttons and formatting
- [ ] InspectorFields, used by Event system and themes to allow for adding custom properties to a subclass that are automatically rendered in the inspector.
- [ ] Layout code to make building 3D and responsive buttons easier for designers. 
- [ ] State management system using ScriptableObjects and a state logic abstract class allows for easily choosing only the states a designer cares about. A custom logic class can be created and assigned to handle unique edge cases.
- [ ] States are easily configured using the Unity Editor inspectors
- [ ] Profiles in the interactable is a list of Themes paired with the gameObjects they control
- [ ] Themes read in states and allow a designer to assign property values for each state
- [ ] Themes should manipulate transform, Text and shader properties
- [ ] Themes will be saved as ScriptableObjects for easy reuse
- [ ] Themes will have an easy to use inspector
- [ ] Event system that provides some basic functionality through the inspector, but is easily customizable and extendable. For instance default expectation is a click event, but could be extended to handle two handed gestures.
- [ ] Events can monitor Interactable states as well as all registered input events, the event system may care about states that themes do not.
[...]

PR #3117, opened 2018-11-11
Fixed issues with double click events and see-it/say-it functionality

killerantz:
Overview
Fixed issues with double click events, especially when isGlobal is set. Fixed SeeIt-SayIt functionality, General bug fixes and filled in missing themes and receivers. Expanded custom receiver example in demo scene

Feature Details
 - IsGlobal : When set to true the Interactable registers as a global listener meaning it will get all events from Input manager even if it's not in focus. Use for things like Menu button on the controller or when the Interactable should be triggered by a speech command action. (will trigger with or without focus) MR headsets do not seem to trigger speech commands that are setup in the speech commands profile.

 - VoiceCommand: a way to assign a voice command to an Interactable directly. This feature was not hooked up previously, sorry for the confusion. Add a string to the Voice Command field. Check Requires Gaze if the event should only trigger when the interactable has focus, otherwise it will trigger globally. This feature has nothing to do with the Speech Commands profile. Examples in the demo scene include the Radial One, Radial Two, Radial Three global commands for the radial buttons and Button and Toggle button voice commands for Toggle Button ("Toggle Button") and Holographic Toggle Button ("Button" as the tool tip says). An advanced feature of the VoiceCommands is the ability to add multiple voice commands for toggle buttons or multi-state controls by adding a comma, with no spaces around commas.

 - Receivers: The demo scene has an example of a Component that looks at an Interactable and can contain a list of receivers that extend ReceiverBase. The example shows how to extend receiver base and turn off UnityEvents for custom receivers. The component (InteractableReceiverList) is an example of building a component that can receive updates from an Interactable for custom events. There are two main ways to handle Interactable events with code. 1. Extend ReceiverBase, follow the example in the demo. Now click New Event on the Interactable and choose your custom receiver from the drop-down list. This is the same Pattern for InteractableReceiverList and InteractableReceiver components. 2. Extend RecieveBaserMonoBehavior, This is a component that can listen to an interactable state changes and register as a handler. All receiver types will get OnClick and OnVoiceCommand event calls. OnVoiceCommand events fire using the VoiceCommands field on the Interactable or setting IsGlobal to true and selecting an Input Action that has a keyword setup in the Speech Commands Profile.

Updates
 - Fixed double event firing when an Interactive has IsGlobal set to true.
 - verified consistent functionality on MR Headset and HoloLens.
 - Added texture, material, audio, string (labels) and rotation themes
 - Added an audio OnClick receiver to make it easy to add audio to click events.
 - Removed Layout3D example scene and most of the code. Moved some of the code back into Interactables and the grid list into Collections. A member of our team is working on something that does the same thing better.


StephenHodgson:
> IsGlobal

maybe `IsFocusRequired`? That's really all that's going on here.

killerantz:
@StephenHodgson "IsFocusRequired ?". That's a good question. IsGlobal registers the interactable to the input system to ignore focus. There's another property that's similar that we should address as well, RequiresGaze, which is a sub property of VoiceCommand. IsGlobal means that the interactable will receive events when not in focus, IsFocusedRequired seems like a practical term for that. RquiresGaze does not follow the pointer, focus pattern, but would be a very similar term, should we make this one VoiceCommandRequiresFocus? There would be a case where interacting with a pointer should require focus, but a voice command does not. The Radial buttons in the Example scene are setup this way and show both properties. I would not want them to be confusing.

StephenHodgson:
See this conversation https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2957

killerantz:
I have pattern working, we just need to make sure the terms are correct and make sense. To get an idea for what all is going on in this interactable, to get everything to work smoothly and allow events for grip and menu to fire correctly, as well as select. There are two systems, one is the input system which I listen to for global events, not select events and speech commands setup in the SpeechCommands profile. For the VoiceCommands that are setup on the Interactable directly by entering a string into the VoiceCommand field, it used it's own keyword recognizer. For the global stuff though, there are issues when the Input System and Pointer System overlap and they over lap differently on the HoloLens compared to MR Headsets. There are some redundancy check that could be optimized with more work, but it's working on both HoloLens and the headsets I tested this weekend.

PJBowron:
@killerantz Hi Todd! Were these the components we discussed last week? Considering adding provision for registering with the input system as a fallback handler as well as (or instead of) a global listener?

killerantz:
@PJBowron , yes, but I haven't added anything for marking an event as used or registering as a fallback. I guess, there's a little of that happening internally to avoid the Pointer and input events causing double OnClick events to be fired.

cre8ivepark:
@killerantz Could you please add a README file for this? I think we can include [your diagrams from the task](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/2607). (add to External\ReadMeImages) 
It would be great if you can include quick summary of script files and recommended usage. (e.g. What is  custom Receivers? How can I use it? What should I do if I want to handle events in the script, not on the inspector panel?)

https://user-images.githubusercontent.com/22085553/37225808-afa6a7cc-238c-11e8-9cc2-fbd5dd7e3656.jpg

https://user-images.githubusercontent.com/22085553/37225790-9e23938e-238c-11e8-81a0-2b48661a3938.jpg

StephenHodgson:
I think we should maybe we should have some meetings (similar to what we're doing with sharing) about the design and implementation of this feature.

PR #3121, opened 2018-11-12
Updated core editor setup classes.

### LINKED ISSUE #3118: Update the "Apply Settings" dialog with choices on import (based on SteamVR dialog)
## Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?
Mixed Reality Toolkit

## Describe the solution you'd like
The current "Configure" screen when first importing the MRTK is an "all or nothing" prompt.

However, while investigating the latest SteamVR support, I found their dialog a lot more useful, as seen below.

image

This would likely be a better solution to the current approach and enable the flexibility to "opt in" to the features the user would like enabled on import.

StephenHodgson:
Overview
---
- Added ignore to settings dialog popup.
image
- Customized message for each setting that will be changed.
    - Only settings that need to be updated will be shown.
- Added preference item to disable ignore if needed.
image
- Changed icons to blue and disabled icon updater.
image

Changes
---
- Related to, but does not address #3118


StephenHodgson:
@SimonDarksideJ I did like the steam vr sdk settings window, but I didn't want to steal it, lol.

Issue #3082, opened 2018-11-06
vNext Proposal: Implement specifying logging level


davidkline-ms:
# Overview
Per #3077, customers have a desire to configure the debug output that the toolkit components emit. In past experience, it has been helpful to be able to opt-in /out to specific classes of messages.

This task is to add a set of logging level flags and allow the app developer to specify the desired message types via the MixedRealityToolkit component. Other components can check the current logging level and determine whether or not to log.

This level will be configurable in the inspector and can be changed at runtime (another VERY handy option I have found).

Fair warning: this will impact a large number of files in a simple and predictable way.

Proposed flags:

- Informational (aka Debug.Log)
- Assert (Debug.Assert)
- Warning (Debug.LogWarning)
- Error (Debug.LogError)
- CriticalError (Debug.LogError with significant functionality impact, ex: specified system failed)

# Requirements

* Application developer can customize the types of log messages to be emitted.
* Built-in MRTK components, including core and registered services honor the logging level
* Logging level can be changed (via script and the inspector) and be honored by MRTK components.

# Acceptance Criteria

- [x] Logging level configurable through the MRTK configuration profile
- [x] Logging level can be changed at runtime
- [x] Supporting logging level is as easy as calling DebugUtilities.Log*
- [x] Multiple logging levels (MRTK and app specific) can be supported by passing a custom logging level to the logging methods
- [x] Toolkit component honors the logging level
- [ ] Existing services honor the logging level

StephenHodgson:
I'm not keen on making our own Logging system, when this is something Unity already handles on its own.

Maybe we should be opening a feature request to Unity?

SimonDarksideJ:
i can understand having a log filtering system to prevent us logging when users don't want to.  But I don't think we should be changing how we log.  At least not in this release.

We would need significant data from the uses of logging by users / consumers / etc before even considering something like this, else it has a severe risk of being over-engineered

StephenHodgson:
Agreed, also this seems much more like a proposal than a task.

davidkline-ms:
This is directly related to a customer issue / request. Having the ability to filter debug output is often key to being able to understand issues with client specific code.

davidkline-ms:
> i can understand having a log filtering system to prevent us logging when users don't want to. But I don't think we should be changing how we log. At least not in this release.

This is not changing how we log, it is adding to the configuration profile (default value is Everything) and adding, for example:

[...]

PR #3143, opened 2018-11-14
Add support for specifying debug logging levels - Part 1 (MixedRealityToolkit component support)

davidkline-ms:
Overview
---
This change addresses the core aspects of #3082, by implementing the infrastructure for enabling developer customers to configure the types of debug log messages they wish to see. This setting is set in the MixedRealityToolkitConfigurationProfile.

As part of this, the DebugUtilities methods were updated / added to enable code to participate by honoring the toolkit's configured level. For example:

    DebugUtilities.LogWarning("My warning message");

If an application developer wishes to maintain a separate level (i.e. one for the toolkit and one for the application) they can use the following pattern:

    DebugUtilities.LogWarning(appSpecificLoggingLevel, "My warning message");

To keep this change small (and to address the first part of #3077), this PR focuses only on the LoggingLevels enum, the DebugUtilities class, config profile support and the MixedRealityToolkit object. Core and extension system updates will follow.

Fixes: #3144


Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityToolkitConfigurationProfile.cs
```diff
@@ -36,6 +36,20 @@ public class MixedRealityToolkitConfigurationProfile : BaseMixedRealityProfile
...
 
+        [EnumFlags]
+        [SerializeField]
+        [Tooltip("The types of debug messages to log to the output window.")]
+        private LoggingLevels loggingLevel = (LoggingLevels)(-1);
+
+        /// <summary>
+        /// The desired types of debug messages to log to the output window.
+        /// </summary>
+        public LoggingLevels LoggingLevel
```
StephenHodgson: What if users want to supply their own custom logging levels?

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -371,8 +375,8 @@ private void InitializeInstance()
                 {
                     if (instance != null)
                     {
-                        Debug.Assert(instance.transform.parent == null, "The MixedRealityToolkit should not be parented under any other GameObject!");
-                        Debug.Assert(instance.transform.childCount == 0, "The MixedRealityToolkit should not have GameObject children!");
+                        DebugUtilities.Assert(instance.transform.parent == null, "The MixedRealityToolkit should not be parented under any other GameObject!");
```
StephenHodgson: I hope we don't have to go through each feature area and replace all our debug statements.

I'd almost want to say each feature area should have it's own logging level, so some can be more verbose than others.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -391,14 +395,6 @@ private void InitializeInstance()
         // ReSharper disable once StaticMemberInGenericType
         private static bool searchForInstance = true;
 
-        /// <summary>
-        /// Expose an assertion whether the MixedRealityToolkit class is initialized.
-        /// </summary>
-        public static void AssertIsInitialized()
```
StephenHodgson: This cannot be removed and needs to stay.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -966,7 +962,11 @@ public List<IMixedRealityService> GetActiveServices(Type type, string serviceNam
         private void InitializeAllServices()
         {
             //If the Mixed Reality Toolkit is not configured, stop.
-            if (activeProfile == null) { return; }
+            if (ActiveProfile == null)
+            {
+                DebugUtilities.LogCriticalError("Unable to intialize services. There is no active profile configured.");
```
StephenHodgson: Fairly certain we could never get this far anyway. But I suppose it doesn't hurt to have it.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -984,7 +984,11 @@ private void InitializeAllServices()
         private void ResetAllServices()
         {
             //If the Mixed Reality Toolkit is not configured, stop.
-            if (activeProfile == null) { return; }
+            if (ActiveProfile == null)
+            {
+                DebugUtilities.LogCriticalError("Unable to reset services. There is no active profile configured.");
```
StephenHodgson: Seems like a lot of duplicated code, couldn't we just log this critical error in the `ActiveProfile` property?

Assets/MixedRealityToolkit/_Core/Utilities/DebugUtilities.cs
```diff
@@ -1,56 +1,168 @@
...
+        /// object's currently configured logging level.
+        /// </summary>
+        /// <param name="message">The message to display.</param>
+        /// <remarks>
+        /// To help differentiate the output of this method from that of LogError, the
+        /// string "Critical: " is prepended to the supplied message.
+        /// </remarks>
+        public static void LogCriticalError(string message)
+        {
+//            LogCriticalError(MixedRealityToolkitLoggingLevel, message);
```
StephenHodgson: Why is this commented out?

Assets/MixedRealityToolkit/_Core/Utilities/DebugUtilities.cs
```diff
@@ -1,56 +1,168 @@
...
 
         /// <summary>
-        /// Logs a warning message.
+        /// Logs an informational message, using the Mixed Reality Toolkit
+        /// object's currently configured logging level.
+        /// </summary>
+        /// <param name="message">The message to display.</param>
+        public static void LogInformation(string message)
+        {
+//            LogInformation(MixedRealityToolkitLoggingLevel, message);
```
StephenHodgson: same question

Assets/MixedRealityToolkit/_Core/Utilities/DebugUtilities.cs
```diff
@@ -1,56 +1,168 @@
...
+        /// Logs a warning message, using the Mixed Reality Toolkit
+        /// object's currently configured logging level.
         /// </summary>
-        /// <param name="condition">The message to log.</param>
-        public static void DebugLogWarning(string message)
+        /// <param name="message">The message to display.</param>
+        public static void LogWarning(string message)
         {
-            Debug.LogWarning(message);
+//            LogWarning(MixedRealityToolkitLoggingLevel, message);
```
StephenHodgson: also commented out.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: It was called from one location internal to the toolkit object (and it was a simple wrapper around an assert statement. What makes it required? Consistency of message?

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: > I hope we don't have to go through each feature area and replace all our debug statements.

We do because static classes (ex: UnityEngine.Debug) cannot have extension methods. I am signing myself up to update the calls.

> I'd almost want to say each feature area should have it's own logging level, so some can be more verbose than others.

The implementation makes that very doable. Each system would use the DebugUtilities overload that takes in a LoggingLevels parameter.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityToolkitConfigurationProfile.cs
davidkline-ms: > What if users want to supply their own custom logging levels?

Are you asking about runtime users or developer users? Developers can set their desired level in the MixedRealityToolkitConfigurationProfile and that will control the core toolkit object and the services.

If you are asking about custom application code, this is why there are DebugUtilities overloads that take the logging level as a parameter.  It allows any code to send custom levels to the log methods.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: > Seems like a lot of duplicated code, couldn't we just log this critical error in the ActiveProfile property?

That sounds like a good optimization. This change was to match the pattern used in several other methods (ex: GetService).

Assets/MixedRealityToolkit/_Core/Utilities/DebugUtilities.cs
davidkline-ms: I was testing something and missed uncommenting it... thanks for the catch!

Assets/MixedRealityToolkit/_Core/Utilities/DebugUtilities.cs
davidkline-ms: I was testing something and missed uncommenting it... thanks for the catch!

Assets/MixedRealityToolkit/_Core/Utilities/DebugUtilities.cs
davidkline-ms: I was testing something and missed uncommenting it... thanks for the catch!

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
StephenHodgson: Because it would be a breaking change.

We need to at least mart it obsolete.

davidkline-ms: Fair enough. I put it back in my recent commit.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityToolkitConfigurationProfile.cs
StephenHodgson: Sounds good. Let's also add that virtual property to the `IMixedRealityService` interface, and update each Service profile with the logging level. 

For the extension services we can add the logging level right under the Runtime Platform settings.

davidkline-ms: > Let's also add that virtual property to the IMixedRealityService interface, and update each Service profile with the logging level.

Only concern here is that customers will be required to change the logging level for ALL of the services individually.... that can be quite a few.  Any thoughts on a good way to have a "set all to <level>" feature? Perhaps on the toolkit menu?

ryzngard:
Could this be done with IoC in mind? Ex: define `ILoggerProvider` that each class can take in and call `ILoggerProvider.CreateLogger<T>` similar to [asp.net](https://github.com/aspnet/Extensions/tree/master/src/Logging/Logging/src). This allows us to have some custom configuration, easy extensibility, and per instance configuration.

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityToolkitConfigurationProfile.cs
davidkline-ms: > Let's also add that virtual property to the IMixedRealityService interface,

Propose making this change in a subsequent PR (when the services are updated to honor the setting). Sound fair?

StephenHodgson: Oh, I see that was your part 2 PR?

davidkline-ms:
> Could this be done with IoC in mind? Ex: define ILoggerProvider that each class can take in and call ILoggerProvider.CreateLogger<T> similar to asp.net. This allows us to have some custom configuration, easy extensibility, and per instance configuration.

Interesting thought. Not sure per instance buys much when many (most?) services are single instance. If we add to the service interface, each profile can provide a custom level.

StephenHodgson:
@ryzngard That's what I was thinking as well, but I think the virtual property on each service's profile will provide all the custom configuration we will need.

StephenHodgson:
When registering multiple services they can each have their own logging level, so it should be okay.

StephenHodgson:
We could do logging as a service, but I think that'll dramatically add to that registry. It's already feeling a bit crowded in that list haha.

davidkline-ms:
> for extensibility.

What if we looked at this as the 2018 implementation of logging will inform any needed changes for the 2019 product. 

While it is not ideal to plan for breaking changes, we would gain the benefit of user feedback... Thoughts?

Assets/MixedRealityToolkit/_Core/Definitions/MixedRealityToolkitConfigurationProfile.cs
davidkline-ms: yep :)

StephenHodgson:
I think the extensibility already is there in the form of an extension class and logging level in the config profile. I think it's fine as is, but a bit overboard if we have to make a logging service for each service that needs one.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: Updated ActiveProfile to always log a critical error if there isn't one configured

SimonDarksideJ:
I think @StephenHodgson and @ryzngard are correct @davidkline-ms , this approach is far to limiting.
If we are to build an alternate logging system, then it needs to follow the componentized approach, allowing developers to replace the "logging" system with their own implementation.

Simply wrapping the Debug.Log events isn't enough to meet some of the complex demands for projects and likely will break existing analytic solutions used by some teams that rely on the way Unity exposes Debug.Log.

I think we'll need to pull this and go back to the architecture board before proceeding.

davidkline-ms:
> this approach is far to limiting.

Can you describe how it is limiting (aside from enabling developers to write a logging system)? I'd like to understand your concerns.

Per @StephenHodgson, "When registering multiple services they can each have their own logging level, so it should be okay."

We have direct customer feedback that a means of reducing logging volume is needed and they have responded favorably to the current proposal (this PR).

It feels like a replaceable logging service would be a large amount of overhead.

Thoughts?

StephenHodgson:
> They have responded favorably.

I'd like to see some more direct feedback here on GitHub haha. 

Yeah, Simon my previous concerns were addressed. They're just "Routing" the messages. Not changing the logging levels in Unity itself, which I am very much against.

davidkline-ms:
> I'd like to see some more direct feedback here on GitHub haha.

That is something that is actively being addressed. Some (much?) of this feedback will be quoted and posted by me. :)

ryzngard:
> Can you describe how it is limiting (aside from enabling developers to write a logging system)? I'd like to understand your concerns.

Idk about Simon, but I think the current limitation is that we aren't really fixing the strong dependency our services have on Unity output. This is replacing one strong implementation with another. I think if we're already planning on replacing calls everywhere we should abstract to an interface to represent the needed contract. This makes it way easier to iterate in the future.  If we use an interface, then we can have meaningful conversations about how to best provide extensibility for 2019 without having to do a breaking change at all.

StephenHodgson:
I'm def not in favor of replacing the default unity logging output.

ryzngard:
> I'm def not in favor of replacing the default unity logging output. 

I'm not saying that though. Imagine that the current proposal was the same, except minor tweaks like : 

```c# 
public class DebugLogger : ILogger
{
    public void Assert(LoggingLevels logLevel, bool condition, string message)
    {
         if (IsEnabled(logLevel))
        {
            Debug.Assert(condition, message);
        }
    }
}
``` 

Then callers call an `ILogger.Assert` interface instead of `DebugHelpers.Assert` directly. Even if it's a singleton right now, even if we ever only have one, it helps us be explicit about the logging contract things expect.

StephenHodgson:
I do like that, but I worry about the number of services we already have in that extension services profile list, lol. 

I might need to update it to sort them better or display them better at least.
Goes back to having the type of logger to use in the dropdown next to the platform settings for said service.

davidkline-ms:
> I'm def not in favor of replacing the default unity logging output.

In considering @ryzngard 's proposal, I am wondering if customers may wish to have a FILE logger as an option.... We could, if we move to interfaces, provide two implementations: debug output and file output.

davidkline-ms:
Will open a proposal that will allow for developer extendable logging (without a breaking change) that would look similar to .net's debug and trace listener pattern. This new proposal should target the 2019 MRTK product as it is beyond fixing the customer reported issue.

davidkline-ms:
> Nice. Like the last commit cleanup

Thanks!

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -68,6 +69,11 @@ public MixedRealityToolkitConfigurationProfile ActiveProfile
         {
             get
             {
+                if (activeProfile == null)
+                {
+                    // NOTE: To avoid a circular reference, this message must not be controlled by the DebugUtilties class.
+                    Debug.LogError("Critical: The MixedRealityToolkit is not configured with a profile.");
```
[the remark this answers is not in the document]
keveleigh: What are the valid cases?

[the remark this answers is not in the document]
StephenHodgson: We weren't  checking here to begin with. So apparently there were valid cases before this change.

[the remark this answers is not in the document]
keveleigh: I wonder if those aren't valid cases but weren't caught because they weren't checked, and we should find a way to fix those up now that it's properly checking.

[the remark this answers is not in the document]
StephenHodgson: Pretty sure they're valid.

Issue #3147, opened 2018-11-15
WindowsMixedRealitySpatialObserver extension service should not be in Core/Devices/MixedReality


StephenHodgson:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
The `WindowsMixedRealitySpatialObserver` extension service is not an input device and should be co-located with the System that depends on it in the MixedRealityToolkit-SDK\Features\SpatialAwareness folder.

Issue #3115, opened 2018-11-11
vNEXT: Update inconsistent naming of core system implementations


StephenHodgson:
All core system services should have consistent naming.
With the new naming conventions, I've noticed we've got some inconsistent names.

Ideal: `<FeatureName>System`

I think extension services are fine, and don't need to update the names right away, but I'd at least like to address the core system services first.

List of good:
- `MixedRealitySpatialAwarenessSystem`

List of bad:
- `MixedRealityBoundaryManager`
- `MixedRealityDiagnosticsManager`
- `MixedRealityInputManager`
- `MixedRealityTeleportManager`

SimonDarksideJ:
I thought we went through and cleared this up for the Beta, gosh darn it.
Needs fixing ASAP

Yoyozilla:
trivial fix. David will update.

davidkline-ms:
Boundary, Diagnostics and Teleport systems have previously been updated. The remaining piece is input

davidkline-ms:
Fixed with #3342 (and other system move PRs)

Issue #3053, opened 2018-11-01
vNext Task: Move default service implementations from SDK to core


davidkline-ms:
# Overview
After discussion on GitHub and shiproom, we have decided to revert the change from #2455 and move the default implementations of the core systems and services from the SDK to the Core.

This will enable customers to download a single, slim package and be able to create a working experience.

The SDK will become a "presentation layer" that contains standard assets, prefabs, utility scripts, etc that can be optionally used in an application.

In addition, a subset of device support will ship in the core (currently Windows Mixed Reality and OpenVR) will be included in the core, with add-on device support shipping in separate packages. 

# Proposed updated folder structure
- MixedRealityToolkit
    - Core
        - Services
            - `<Feature>`
                - `<Platform>`

Services folder to include:
- `MixedRealityToolkit`
- `BaseService`
- `BaseEventSystem`
- `BaseExtensionService`

Features folder to include:
- `<Feature>System`
- `Base<Feature>ExtensionService`

Platform folder to include:
- `<Platform specific><Feature>ExtensionServices` derived from the base feature extension service
- An assembly definition for the feature's platform specific services so they can be added and removed via package manager with ease.

# Requirements
* Default core system implementations moved from SDK to Core layer

# Acceptance Criteria
- [ ] I can install just the core, press play and see results.

Issue #2905, opened 2018-10-10
vNext task: separate registered components (ex: device managers) into assemblies


davidkline-ms:
# Overview
Today, the MRTK core assembly contains the code for all of the supported devices (windows mixed reality, open vr, etc). To improve modularization and to allow developers to only include the code that they intend to use (ex: HoloLens only app should not be required to ship OpenVR support code, and the other way around), we should take the Devices code out of the core dll and create individual assemblies.

This will enable faster expansion of device support as well as better customization for developers.

# Requirements
- Core MRTK assembly contains interface and type definitions
- Core MRTK assembly contains the Mixed Reality Manager code
- Core MRTK does NOT contain device implementation code
- Default registered components profile needs to be revisited to properly support the new division of binaries.

# Acceptance Criteria
- [ ] Installing MRTK Core package can configure the scene with the Mixed Reality Manager
- [ ] User is prompted (?) to add the appropriate device support to the Registered Components profile
- [ ] All devices are considered "extensions" to the MRTK

ryzngard:
We need to make this available in NuGet.

StephenHodgson:
This already happens?

I'm not sure the this makes much sense as a task, since Unity already splits these things apart into their own assemblies after each compile.

StephenHodgson:
Should also be availible in Unity's package manager as well.

davidkline-ms:
complete

PR #3148, opened 2018-11-15
Spatial Awareness system refactor & move

StephenHodgson:
Overview
---
Updated the spatial observer location to be in the SDK, with the spatial awareness default implementation.

Changes
---
- Fixes: #3147 
- Fixes: #3226
- addresses #2905 for Spatial Awareness System and components by adding assembly definitions for platform specific code. Future updates when package manager arrives will install/uninstall platform specific stuff as needed.
- addresses #3053 & #3115 for Spatial Awareness System and components.

TODO
---
- [x] #3170
- [x]  #3173
- [x] #3190

Breaking Changes
---
1. If users have a custom configuration profile, they will need to re-assign the spatial awareness system type.
2. If have a custom Extension Services provider profile, and using the default spatial observer implementation, they'll need to update their service entry with the updated type.

- All of the configuration settings from the spatial awareness system have moved into the spatial observer data providers. The spatial awareness system now only routes the event data from the spatial observers to the handlers per arch guidelines.


Assets/MixedRealityToolkit-SDK/Features/SpatialAwareness/System/WindowsMixedRealitySpatialObserver.cs
```diff
@@ -121,20 +106,19 @@ public override void Destroy()
...
         private float lastUpdated = 0;
-#endif // UNITY_WSA
 
         /// <inheritdoc />
         public override IDictionary<int, GameObject> Meshes
         {
             get
             {
-                Dictionary<int, GameObject> meshes = new Dictionary<int, GameObject>();
+                var meshes = new Dictionary<int, GameObject>();
```
davidkline-ms: var makes me sad :)

StephenHodgson: Lol, but you already know the type you're creating on the right :p

davidkline-ms: Yeah, I know... old habits die hard :)

StephenHodgson:
@davidkline-ms Let's also hold off on merging this until we figure out what we wanna do with the default implementations. Going to open a proposal task soon.

StephenHodgson:
@davidkline-ms okay this PR is ready again

StephenHodgson:
Seems like I broke something. Not ready for review yet.

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealityRegisteredServiceProvidersProfile.asset
```diff
@@ -56,7 +56,7 @@ MonoBehaviour:
     runtimePlatform: -1
     configurationProfile: {fileID: 0}
   - componentType:
-      reference: Microsoft.MixedReality.Toolkit.Core.Devices.SpatialAwareness.WindowsMixedRealitySpatialObserver,
+      reference: Microsoft.MixedReality.Toolkit.Core.Services.SpatialAwarenessSystem.BaseSpatialObserver,
```
StephenHodgson: Ah it's only the base spatial observer, need to update this with the actual windows spatial observer.

StephenHodgson:
Fixed

davidkline-ms:
Also receiving:

Assets\MixedRealityToolkit\_Core\Definitions\SpatialAwarenessSystem\MixedRealitySpatialObserverDataProvidersProfile.cs(15,60): warning CS0649: Field 'MixedRealitySpatialObserverDataProvidersProfile.registeredSpatialObserverDataProviders' is never assigned to, and will always have its default value nu

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -454,7 +433,7 @@ public SpatialAwarenessMeshLevelOfDetail MeshLevelOfDetail
 
         /// <inheritdoc />
         /// <remarks>The observer manages the mesh collection.</remarks>
-        public IDictionary<int, GameObject> Meshes => SpatialAwarenessObserver.Meshes;
+        public IReadOnlyDictionary<int, SpatialMeshObject> Meshes => SpatialAwarenessObserver.Meshes;
```
davidkline-ms: Please see earlier comment. Thx

Assets/MixedRealityToolkit/_Core/Interfaces/SpatialAwareness/IMixedRealitySpatialAwarenessSystem.cs
```diff
@@ -128,9 +133,9 @@ public interface IMixedRealitySpatialAwarenessSystem : IMixedRealityEventSystem,
         Material MeshOcclusionMaterial { get; set; }
 
         /// <summary>
-        /// Gets the collection of <see cref="GameObject"/>s being managed by the spatial awareness mesh subsystem.
+        /// Gets the collection of <see cref="SpatialMeshObject"/>s being managed by the spatial awareness mesh subsystem.
         /// </summary>
-        IDictionary<int, GameObject> Meshes { get; }
+        IReadOnlyDictionary<int, SpatialMeshObject> Meshes { get; }
```
davidkline-ms: Please see earlier comment. Thx

Assets/MixedRealityToolkit-SDK/MixedRealityToolkit.SDK.asmdef
```diff
@@ -3,7 +3,8 @@
     "references": [
         "Microsoft.MixedReality.Toolkit",
         "Microsoft.MixedReality.Toolkit.InputSystem",
-        "Microsoft.MixedReality.Toolkit.Core.Utilities.Async"
+        "Microsoft.MixedReality.Toolkit.Core.Utilities.Async",
+        "Microsoft.MixedReality.Toolkit.Core.Devices.SpatialObservers.WindowsMixedReality"
```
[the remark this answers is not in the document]
StephenHodgson: If it's in the list, then obviously yeah. We do.

StephenHodgson:
I don't have a HL to test on, so I'll likely need to circle back to this after I get back next week.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/WindowsMixedReality/WindowsMixedRealitySpatialObserver.cs
```diff
@@ -121,33 +104,19 @@ public override void Destroy()
...
-                Dictionary<int, GameObject> meshes = new Dictionary<int, GameObject>();
-                // NOTE: We use foreach here since Dictionary<key, value>.Values is an IEnumerable.
-                foreach (int id in meshObjects.Keys)
-                {
-                    meshes.Add(id, meshObjects[id].GameObject);
-                }
-                return meshes;
-            }
-        }
+        public override IReadOnlyDictionary<int, SpatialMeshObject> Meshes => meshObjects;
```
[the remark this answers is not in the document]
StephenHodgson: Okay, let me take another look at this.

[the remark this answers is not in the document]
StephenHodgson: For 1: Are we sure that applications don't need access to the `SpatialMeshObject`?

What's an example of how an application could request them on demand?

For 2: This makes sense on why we want a copy.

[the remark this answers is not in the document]
StephenHodgson: I was under the impression that the `IReadOnlyCollection` would provide an immutable copy, but after some research, sadly [it doesn't](https://visualstudiomagazine.com/articles/2012/08/07/new-read-only-collection-interfaces-for-net.aspx).

>One caveat to note is that the interfaces do not provide an immutable copy of the collection but rather a read-only view of the source mutable collection.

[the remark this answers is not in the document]
davidkline-ms: >For 1: Are we sure that applications don't need access to the SpatialMeshObject?
What's an example of how an application could request them on demand?

Abbreviated example, if n app wants the renderer for a specific mesh:

List<GameObject> meshes = SpatialAwarenessSystem.Meshes;
…
meshes[i].GetComponent<Renderer>();

While this may not be the most performant mechanism, it protects the internal SpatialMeshObject (SMO) data from being modified by an app. (While today, SMO is defined as a struct, a new PR will make it a class).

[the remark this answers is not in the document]
StephenHodgson: Yeah let's not use `getComponent` as it's expensive, esp for a large collection of objects.

Part of the problem is that the `SpatialMeshObject` is using unprotected fields instead of properties counter to the coding guidelines.

Assets/MixedRealityToolkit-SDK/MixedRealityToolkit.SDK.asmdef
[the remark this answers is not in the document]
StephenHodgson: Oh, this was left over from when the spatial awareness system was in the SDK. 

It can be removed now that it's been moved into the core.

Assets/MixedRealityToolkit/_Core/Interfaces/SpatialAwareness/Handlers/IMixedRealitySpatialAwarenessMeshHandler.cs
```diff
@@ -6,24 +6,27 @@
 
 namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.SpatialAwarenessSystem.Handlers
 {
-    public interface IMixedRealitySpatialAwarenessMeshHandler : IEventSystemHandler
+    /// <summary>
+    /// 
+    /// </summary>
+    public interface IMixedRealitySpatialAwarenessMeshHandler<T> : IEventSystemHandler
```
davidkline-ms: why change this interface to be a generic?

Assets/MixedRealityToolkit/_Core/Interfaces/SpatialAwareness/Handlers/IMixedRealitySpatialAwarenessSurfaceFindingHandler.cs
```diff
@@ -6,24 +6,28 @@
 
 namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.SpatialAwarenessSystem.Handlers
 {
-    public interface IMixedRealitySpatialAwarenessSurfaceFindingHandler : IEventSystemHandler
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public interface IMixedRealitySpatialAwarenessSurfaceFindingHandler<T> : IEventSystemHandler
```
davidkline-ms: same question re: changing to be a generic

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/WindowsMixedReality/WindowsMixedRealitySpatialObserver.cs
```diff
@@ -121,33 +104,19 @@ public override void Destroy()
...
-                Dictionary<int, GameObject> meshes = new Dictionary<int, GameObject>();
-                // NOTE: We use foreach here since Dictionary<key, value>.Values is an IEnumerable.
-                foreach (int id in meshObjects.Keys)
-                {
-                    meshes.Add(id, meshObjects[id].GameObject);
-                }
-                return meshes;
-            }
-        }
+        public override IReadOnlyDictionary<int, SpatialMeshObject> Meshes => SpatialMeshObjects;
```
davidkline-ms: unfortunately, this pattern will not protect the caller from changes to the underlying collection.  the following snippet demonstrates the exception that occurs (InvalidOperation, collection was modified).

```
        Dictionary<int, float> dataSet = new Dictionary<int, float>();
        IReadOnlyDictionary<int, float> ReadOnlyDataSet => dataSet;

            InitializeDataSet();

            Random random = new Random(DateTime.Now.Second);

            while (true)
            {
                foreach(int i in ReadOnlyDataSet.Keys)
                {
                    // Print the key-value pair
                    Console.WriteLine($"{i} : {ReadOnlyDataSet[i]}");

                    // Remove the item
                    Console.WriteLine($"Remove {i}");
                    dataSet.Remove(i);

                    // Add a new item to the dictionary
                    Console.WriteLine("Add new item");
                    dataSet.Add(random.Next(), (float)random.NextDouble());
                }
            }
```

davidkline-ms: I have a fix for the readonly / copy requirement that I will submit as a PR on @StephenHodgson's fork.

Assets/MixedRealityToolkit/_Core/Interfaces/SpatialAwareness/Handlers/IMixedRealitySpatialAwarenessMeshHandler.cs
StephenHodgson: Because it's more powerful and flexible as a generic. Just look how effective it is for the input system.

Then we can have different kinds of data types that can be returned using the handler.

Assets/MixedRealityToolkit/_Core/Definitions/SpatialAwarenessSystem/MixedRealitySpatialAwarenessProfile.cs
```diff
@@ -143,12 +163,12 @@ public class MixedRealitySpatialAwarenessProfile : BaseMixedRealityProfile
 
         [SerializeField]
         [Tooltip("Physics layer on which to set the planar surfaces")]
-        private int surfaceFindingPhysicsLayer = 31;
+        private LayerMask surfaceFindingPhysicsLayerMask;
 
         /// <summary>
         /// The desired Unity Physics Layer on which to set the planar surfaces.
```
StephenHodgson:
```suggestion
        /// The desired Unity Physics Layer(s) on which to set the planar surfaces.
```

Assets/MixedRealityToolkit/_Core/Interfaces/SpatialAwareness/Handlers/IMixedRealitySpatialAwarenessMeshHandler.cs
```diff
@@ -6,24 +6,27 @@
 
 namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.SpatialAwarenessSystem.Handlers
 {
-    public interface IMixedRealitySpatialAwarenessMeshHandler : IEventSystemHandler
+    /// <summary>
+    ///
```
StephenHodgson:
```suggestion
    /// The event handler for all Spatial Awareness Mesh Events.
```

Assets/MixedRealityToolkit/_Core/Interfaces/SpatialAwareness/Handlers/IMixedRealitySpatialAwarenessSurfaceFindingHandler.cs
```diff
@@ -6,24 +6,28 @@
 
 namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.SpatialAwarenessSystem.Handlers
 {
-    public interface IMixedRealitySpatialAwarenessSurfaceFindingHandler : IEventSystemHandler
+    /// <summary>
+    ///
```
StephenHodgson:
```suggestion
    /// The event handler for all Spatial Awareness Surface Finding Events.
```

Assets/MixedRealityToolkit/_Core/Definitions/SpatialAwarenessSystem/MixedRealitySpatialAwarenessProfile.cs
```diff
@@ -166,7 +166,7 @@ public class MixedRealitySpatialAwarenessProfile : BaseMixedRealityProfile
         private LayerMask surfaceFindingPhysicsLayerMask;
 
         /// <summary>
-        /// The desired Unity Physics Layer on which to set the planar surfaces.
+        /// The desired Unity Physics Layer(s) on which to set the planar surfaces.
```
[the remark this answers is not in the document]
StephenHodgson: Sure it does. That's why I made this change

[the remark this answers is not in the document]
davidkline-ms: From the unity documentation, GameObjects live on a singular layer

layer | The layer the game object is in.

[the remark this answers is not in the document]
StephenHodgson: Oh I see. Thanks for pointing this out.

Well I'd still like to see a drop down of layers by the tag manager. I'll see about adding an attribute to do this for us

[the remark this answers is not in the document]
davidkline-ms: That might be handy. It implies that either the toolkit (or the customer) has added a label name via the tag manager (there is code that does this in the HTK branch).

Sounds like a good proposal to file.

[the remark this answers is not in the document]
StephenHodgson: Done

image

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/EnumFlagsAttributeDrawer.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Inspectors.PropertyDrawers
     /// From https://answers.unity.com/questions/486694/default-editor-enum-as-flags-.html
     /// </summary>
     [CustomPropertyDrawer(typeof(EnumFlagsAttribute))]
-    public class EnumFlagsAttributeDrawer : PropertyDrawer
+    public sealed class EnumFlagsAttributeDrawer : PropertyDrawer
```
Jarodshow: Why should this be sealed?

StephenHodgson:
Just a heads up this PR is not ready for review yet.

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealitySpatialAwarenessProfile.asset
```diff
@@ -26,7 +28,7 @@ MonoBehaviour:
   meshOcclusionMaterial: {fileID: 2100000, guid: a809c7ea182b699409443d46d5b42435,
     type: 2}
   useSurfaceFindingSystem: 0
-  surfaceFindingPhysicsLayer: 31
+  surfaceFindingPhysicsLayer: 30
```
davidkline-ms: Not fond of having a separate layer defined for surfaces vs meshes as there are only 24 user layers. That said, we should get customer feedback on the usability rather than block on this.

davidkline-ms:
> Just a heads up this PR is not ready for review yet.

What needs to be changed?

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealitySpatialAwarenessProfile.asset
StephenHodgson: We can make them the same layer. We didn't prevent them from being the same, but if users want to make them different they can.

StephenHodgson:
I still need to finish building out the data provider profiles

StephenHodgson:
Rather do all the breaking changes together. In either case it's not ready for review. Likely will break apart the PRs. Just depends on what relies on what.

davidkline-ms:
Just a note, other pending Spatial Awareness work is blocked by this change in order to reduce conflicts and need of reimplementing fixes.  Thx!

Assets/MixedRealityToolkit-SDK/Profiles/DefaultMixedRealitySpatialAwarenessProfile.asset
StephenHodgson: The nice thing about this too is that if the layer is already set then it moves it to the next availible layer.

StephenHodgson: I'm not sure if you noticed that.

StephenHodgson:
@davidkline-ms this PR is now ready to review.

StephenHodgson:
The spatial awareness system now has a registry of all the spatial observer data providers that can be ran with the system. These data providers will usually be platform specific.

I removed all of the spatial observer configuration options from the spatial awareness system to the observer itself. 

image

image

This opens up the possibility to have platform specific observer configurations with our platform specific implementations. 

The spatial observer data provider registry also requires that a valid profile be set for the data provider to be registered.

Assets/MixedRealityToolkit/_Core/Definitions/Utilities/ProfileMenuItemIndices.cs
```diff
@@ -19,6 +19,7 @@ public enum CreateProfileMenuItemIndices
         BoundaryVisualization,
         ControllerVisualization,
         SpatialAwareness,
+        SpatialAwarenessDataProviders,
```
StephenHodgson: need to remove this, I don't think we're using it anymore.

Assets/MixedRealityToolkit-SDK/Profiles/SpatialAwareness/DefaultWindowsMixedRealitySpatialMeshObserverProfile.asset
```diff
@@ -0,0 +1,26 @@
...
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: 90a4bdbfdea331745b57570c107eeaae, type: 3}
+  m_Name: DefaultWindowsMixedRealitySpatialMeshObserverProfile
+  m_EditorClassIdentifier: 
+  isCustomProfile: 1
+  startupBehavior: 0
+  observationExtents: {x: 3, y: 3, z: 3}
+  isStationaryObserver: 1
```
keveleigh: To [match the original value](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3148/files#diff-5b6ab8fcf25beaea64478b8dd17d1768L17) and the expectations of the demo scene, which describes the ability to move around and watch the mesh update around you.

```suggestion
  isStationaryObserver: 0
```

Assets/MixedRealityToolkit-SDK/Profiles/SpatialAwareness/DefaultWindowsMixedRealitySpatialMeshObserverProfile.asset
```diff
@@ -0,0 +1,26 @@
...
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInternal: {fileID: 0}
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: 90a4bdbfdea331745b57570c107eeaae, type: 3}
+  m_Name: DefaultWindowsMixedRealitySpatialMeshObserverProfile
+  m_EditorClassIdentifier: 
+  isCustomProfile: 1
```
StephenHodgson:
```suggestion
  isCustomProfile: 0
```

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/EnumFlagsAttributeDrawer.cs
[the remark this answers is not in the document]
davidkline-ms: We should track / discuss this change as a separate issue.

Assets/MixedRealityToolkit/_Core/Interfaces/DataProviders/SpatialObservers/IMixedRealitySpatialMeshObserver.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System.Collections.Generic;
+using Microsoft.MixedReality.Toolkit.Core.Definitions.SpatialAwarenessSystem;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.DataProviders.SpatialObservers
+{
+    /// <summary>
+    /// The interface contract for Mixed Reality spatial observers.
+    /// </summary>
+    public interface IMixedRealitySpatialMeshObserver : IMixedRealitySpatialObserverDataProvider
```
davidkline-ms: Please re-add the MeshPhysicsLayer property and the MeshPhysicsLayerMask helper.

Assets/MixedRealityToolkit/_Core/Interfaces/DataProviders/SpatialObservers/IMixedRealitySpatialObserverDataProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Gets or sets the frequency, in seconds, at which the spatial observer updates.
+        /// </summary>
+        float UpdateInterval { get; set; }
+
+        /// <summary>
+        /// The physics layer to use when creating spatial objects for this observer.
+        /// </summary>
+        int PhysicsLayer { get; set; }
```
davidkline-ms: Please remove per earlier comments.

Assets/MixedRealityToolkit/_Core/Interfaces/DataProviders/SpatialObservers/IMixedRealitySurfaceObserver.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.DataProviders.SpatialObservers
+{
+    public interface IMixedRealitySurfaceObserver : IMixedRealitySpatialObserverDataProvider
```
davidkline-ms: Please re-add SurfacePhysicsLayer property and the SurfacePhysicsLayerMask helper

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -226,23 +227,41 @@ private void InitializeServiceLocator()
             // If the Spatial Awareness system has been selected for initialization in the Active profile, enable it in the project
             if (ActiveProfile.IsSpatialAwarenessSystemEnabled)
             {
+#if UNITY_EDITOR
+                // Setup the default spatial awareness layers in the project settings.
+                LayerExtensions.SetupLayer(31, MixedRealitySpatialAwarenessSystemProfile.SpatialAwarenessMeshesLayerName);
+                LayerExtensions.SetupLayer(30, MixedRealitySpatialAwarenessSystemProfile.SpatialAwarenessSurfacesLayerName);
```
davidkline-ms: Per earlier comment, user definable layers are a limited quantity in Unity. The Spatial Awareness System should pre-define only one layer and allow customers to define additional layers as needed/desired.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -0,0 +1,289 @@
...
+
+        #region IMixedRealitySpatialAwarenessSystem Implementation
+
+        /// <inheritdoc />
+        public HashSet<IMixedRealitySpatialObserverDataProvider> DetectedSpatialObservers { get; } = new HashSet<IMixedRealitySpatialObserverDataProvider>();
+
+        /// <inheritdoc />
+        public bool IsObserverRunning(IMixedRealitySpatialObserverDataProvider observer)
+        {
+            return false;
```
davidkline-ms: Never returns anything other than false. This method needs to be completed.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        /// <inheritdoc />
+        public uint GenerateNewObserverId()
```
davidkline-ms: Should this be public? If it needs to be so that it can be called by observers, it should be documented as not being for application use.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        /// <inheritdoc />
+        public void RaiseSpatialAwarenessObserverDetected(IMixedRealitySpatialObserverDataProvider observer)
+        {
+            DetectedSpatialObservers.Add(observer);
+        }
+
+        /// <inheritdoc />
+        public void RaiseSpatialAwarenessObserverLost(IMixedRealitySpatialObserverDataProvider observer)
```
davidkline-ms: When would an observer be lost?

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -0,0 +1,289 @@
...
+                if (spatialObserver.SourceId == observer.SourceId)
+                {
+                    spatialObserver.StopObserving();
+                    break;
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public void RaiseSpatialAwarenessObserverDetected(IMixedRealitySpatialObserverDataProvider observer)
```
davidkline-ms: When would an observer be detected?

ProjectSettings/TagManager.asset
```diff
@@ -35,8 +35,8 @@ TagManager:
   - 
   - 
   - 
-  - 
-  - 
+  - Spatial Awareness Surfaces
```
davidkline-ms: Please see earlier feedback re: limited user defined layers

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
StephenHodgson: We do the same thing in the input system to request a new unique source id for controllers. Same idea applies here.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySpatialObserverDataProvider.cs
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            if (MixedRealityToolkit.SpatialAwarenessSystem != null)
+            {
+                SourceId = MixedRealityToolkit.SpatialAwarenessSystem.GenerateNewObserverId();
+            }
+
+            StartupBehavior = profile.StartupBehavior;
+            UpdateInterval = profile.UpdateInterval;
+            PhysicsLayer = profile.PhysicsLayer;
```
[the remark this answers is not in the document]
StephenHodgson: This _does_ support multiple forms of spatial data. 

This is the base class to support those multiple forms. Each type derives from this base class and has its own configuration settings.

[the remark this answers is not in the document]
StephenHodgson: > Please move this property to the specific observer types as it breaks previously established behavior.

No, it belongs here.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySpatialMeshObserver.cs
```diff
@@ -0,0 +1,225 @@
...
+                Debug.LogError($"Missing profile for {name}");
+                return;
+            }
+
+            MeshLevelOfDetail = profile.MeshLevelOfDetail;
+            MeshTrianglesPerCubicMeter = profile.MeshTrianglesPerCubicMeter;
+            MeshRecalculateNormals = profile.MeshRecalculateNormals;
+            MeshDisplayOption = profile.MeshDisplayOption;
+            MeshVisibleMaterial = profile.MeshVisibleMaterial;
+            MeshOcclusionMaterial = profile.MeshOcclusionMaterial;
```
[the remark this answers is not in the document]
StephenHodgson: It does, in the [base profile](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3148#discussion_r240732385).

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySpatialMeshObserver.cs
```diff
@@ -0,0 +1,225 @@
...
+        public uint MeshTrianglesPerCubicMeter { get; private set; }
+
+        /// <inheritdoc />
+        public bool MeshRecalculateNormals { get; }
+
+        /// <inheritdoc />
+        public SpatialMeshDisplayOptions MeshDisplayOption { get; }
+
+        /// <inheritdoc />
+        public Material MeshVisibleMaterial { get; }
```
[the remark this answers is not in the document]
StephenHodgson: It's still there, in the base spatial observer as `PhysicsLayer`

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySpatialObserverDataProvider.cs
```diff
@@ -0,0 +1,153 @@
...
+        #region IMixedRealitySpatialObserverDataProvider Implementation
+
+        /// <inheritdoc />
+        public AutoStartBehavior StartupBehavior { get; }
+
+        /// <inheritdoc />
+        public float UpdateInterval { get; set; }
+
+        /// <inheritdoc />
+        public int PhysicsLayer { get; set; }
```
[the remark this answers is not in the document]
StephenHodgson: This belongs in the base class. As likely all observers will have specific physics layers we need to set for them.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
StephenHodgson: Good catch, I'll be sure to implement this asap.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
StephenHodgson: If we unplug a hardware sensor like a depth camera, for example.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
StephenHodgson: When the application starts, or when hardware that supports an observer is plugged in.

ProjectSettings/TagManager.asset
StephenHodgson: link?

davidkline-ms:
Question: If an observer supports multiple types of data (ex: points and planes) how would it [de]serialize the data from two profiles?

StephenHodgson:
> Question: If an observer supports multiple types of data (ex: points and planes) how would it [de]serialize the data from two profiles?

Depends on the specific implementation, which is why we have this arch pattern where you can composite interfaces into your concrete classes.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySpatialObserverDataProvider.cs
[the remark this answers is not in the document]
davidkline-ms: As defined, a single observer that reports both meshes and planes, for example, can only support a single physics layer. The desire is to enable a single observer that can manage multiple forms.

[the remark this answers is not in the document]
StephenHodgson: Mesh and surface observers should be different data providers, (and was treated as such in the "current" definition).

The fact that it only reports a single physics layer is intentional as the observer keeps track of its own generated GameObjects.

Surface observers will create and have their own meshes on physics layers defined by the `IMixedRealitySpatialObserverDataModel`

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="profile"></param>
+        protected BaseMixedRealitySurfaceObserver(string name, uint priority, BaseMixedRealitySurfaceObserverProfile profile) : base(name, priority, profile)
+        {
+            if (profile == null)
+            {
+                Debug.LogError($"Missing profile for {name}");
+                return;
+            }
+
+            SurfaceFindingMinimumArea = profile.SurfaceFindingMinimumArea;
```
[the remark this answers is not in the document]
StephenHodgson: I think you're missing the point here. Each surface observer is its own entity with its own configuration options.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/Profiles/BaseMixedRealitySpatialMeshObserverProfile.cs
```diff
@@ -0,0 +1,67 @@
...
+        public Material MeshVisibleMaterial => meshVisibleMaterial;
+
+        [SerializeField]
+        [Tooltip("Material to use when spatial meshes should occlude other objects")]
+        private Material meshOcclusionMaterial = null;
+
+        /// <summary>
+        /// The material to be used when spatial meshes should occlude other objects.
+        /// </summary>
+        public Material MeshOcclusionMaterial => meshOcclusionMaterial;
```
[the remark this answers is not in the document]
StephenHodgson: No.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/Profiles/BaseMixedRealitySpatialObserverProfile.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        /// <summary>
+        /// The frequency, in seconds, at which the spatial observer updates.
+        /// </summary>
+        public float UpdateInterval => updateInterval;
+
+        [PhysicsLayer]
+        [SerializeField]
+        [Tooltip("Which physics layer should this spatial observer use to create it's objects?")]
+        private int physicsLayer = -1;
```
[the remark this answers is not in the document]
StephenHodgson: The default layer is whichever layer is availible for us to set the spatial observer layers on.

If the application is already using this layer we'll have problems.

The system is smart enough to setup the observer physics layers to the first availible layer and then update the default profile to it.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/Profiles/BaseMixedRealitySurfaceObserverProfile.cs
```diff
@@ -0,0 +1,96 @@
...
+        public bool DisplayPlatformSurfaces => displayPlatformSurfaces;
+
+        [SerializeField]
+        [Tooltip("Material to use when displaying platform surfaces")]
+        private Material platformSurfaceMaterial = null;
+
+        /// <summary>
+        /// The material to be used when automatically displaying platform surfaces.
+        /// </summary>
+        public Material PlatformSurfaceMaterial => platformSurfaceMaterial;
```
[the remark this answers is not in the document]
StephenHodgson: No.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// A queue of <see cref="SurfaceId"/> that need their meshes created (or updated).
+        /// </summary>
+        private readonly Queue<SurfaceId> meshWorkQueue = new Queue<SurfaceId>();
+
+        /// <summary> 
+        /// To prevent too many meshes from being generated at the same time, we will 
+        /// only request one mesh to be created at a time.  This variable will track 
+        /// if a mesh creation request is in flight. 
+        /// </summary> 
+        private readonly Dictionary<int, SpatialMeshObject> outstandingMeshObjects = new Dictionary<int, SpatialMeshObject>();
```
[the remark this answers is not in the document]
StephenHodgson: A 0 or 1 object?

Didn't seem like that to me. There was at most 5-6 objects in this dictionary when I was testing it earlier.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
```diff
@@ -0,0 +1,313 @@
...
+
```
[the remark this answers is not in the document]
StephenHodgson: Because it added unnecessary complexity to the system.

Assets/MixedRealityToolkit/_Core/Definitions/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystemProfile.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <summary>
+    /// Configuration profile settings for setting up the spatial awareness system.
+    /// </summary>
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Mixed Reality Spatial Awareness System Profile", fileName = "MixedRealitySpatialAwarenessSystemProfile", order = (int)CreateProfileMenuItemIndices.SpatialAwareness)]
+    public class MixedRealitySpatialAwarenessSystemProfile : BaseMixedRealityProfile
+    {
+        /// <summary>
+        /// The name of the Spatial Awareness Mesh Physics Layer.
+        /// </summary>
+        public const string SpatialAwarenessMeshesLayerName = "Spatial Awareness Meshes";
```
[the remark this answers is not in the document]
StephenHodgson: We only add them if there are layers availible.

If none are availible, users get a message and from there they can choose their own as their own discretion.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/SpatialAwareness/BaseMixedRealitySpatialMeshObserverProfileInspector.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            if (foldout)
+            {
+                EditorGUI.indentLevel++;
+                EditorGUILayout.PropertyField(meshLevelOfDetail);
+                EditorGUILayout.PropertyField(meshTrianglesPerCubicMeter);
+                EditorGUILayout.PropertyField(meshRecalculateNormals);
+                EditorGUILayout.PropertyField(meshDisplayOption);
+                EditorGUILayout.PropertyField(meshVisibleMaterial);
+                EditorGUILayout.PropertyField(meshOcclusionMaterial);
```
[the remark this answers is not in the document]
StephenHodgson: No.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/SpatialAwareness/BaseMixedRealitySpatialObserverProfileInspector.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace Microsoft.MixedReality.Toolkit.Core.Inspectors.Profiles.SpatialAwareness
+{
+    [CustomEditor(typeof(BaseMixedRealitySpatialObserverProfile))]
+    public abstract class BaseMixedRealitySpatialObserverProfileInspector : MixedRealityBaseConfigurationProfileInspector
+    {
+        private SerializedProperty startupBehavior;
+        private SerializedProperty observationExtents;
+        private SerializedProperty isStationaryObserver;
+        private SerializedProperty updateInterval;
+        private SerializedProperty physicsLayer;
```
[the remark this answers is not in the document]
StephenHodgson: No.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/SpatialAwareness/BaseMixedRealitySurfaceObserverProfileInspector.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            surfaceFindingMinimumArea = serializedObject.FindProperty("surfaceFindingMinimumArea");
+            displayFloorSurfaces = serializedObject.FindProperty("displayFloorSurfaces");
+            floorSurfaceMaterial = serializedObject.FindProperty("floorSurfaceMaterial");
+            displayCeilingSurfaces = serializedObject.FindProperty("displayCeilingSurfaces");
+            ceilingSurfaceMaterial = serializedObject.FindProperty("ceilingSurfaceMaterial");
+            displayWallSurfaces = serializedObject.FindProperty("displayWallSurfaces");
+            wallSurfaceMaterial = serializedObject.FindProperty("wallSurfaceMaterial");
+            displayPlatformSurfaces = serializedObject.FindProperty("displayPlatformSurfaces");
+            platformSurfaceMaterial = serializedObject.FindProperty("platformSurfaceMaterial");
```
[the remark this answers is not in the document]
StephenHodgson: No.

Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/SpatialAwareness/BaseMixedRealitySurfaceObserverProfileInspector.cs
```diff
@@ -0,0 +1,84 @@
...
+                EditorGUI.indentLevel++;
+                EditorGUILayout.PropertyField(surfaceFindingMinimumArea, minimumAreaContent);
+                EditorGUILayout.PropertyField(displayFloorSurfaces);
+                EditorGUILayout.PropertyField(floorSurfaceMaterial, floorMaterialContent);
+                EditorGUILayout.PropertyField(displayCeilingSurfaces);
+                EditorGUILayout.PropertyField(ceilingSurfaceMaterial, ceilingMaterialContent);
+                EditorGUILayout.PropertyField(displayWallSurfaces);
+                EditorGUILayout.PropertyField(wallSurfaceMaterial, wallMaterialContent);
+                EditorGUILayout.PropertyField(displayPlatformSurfaces);
+                EditorGUILayout.PropertyField(platformSurfaceMaterial, platformMaterialContent);
```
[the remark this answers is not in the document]
StephenHodgson: No.

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/EnumFlagsAttributeDrawer.cs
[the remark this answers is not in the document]
StephenHodgson: What's your aversion to making things sealed?

People shouldn't be deriving classes from this anyway.

Assets/MixedRealityToolkit/_Core/Interfaces/DataProviders/SpatialObservers/IMixedRealitySpatialMeshObserver.cs
StephenHodgson: No.

Assets/MixedRealityToolkit/_Core/Interfaces/DataProviders/SpatialObservers/IMixedRealitySpatialObserverDataProvider.cs
StephenHodgson: No. It's staying.

Assets/MixedRealityToolkit/_Core/Interfaces/DataProviders/SpatialObservers/IMixedRealitySurfaceObserver.cs
StephenHodgson: No.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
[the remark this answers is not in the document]
StephenHodgson: I see why that could be a problem looking back at the summary.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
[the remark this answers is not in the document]
davidkline-ms: The issue is that, as defined here, a single observer cannot support meshes and planes on different layers.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySpatialObserverDataProvider.cs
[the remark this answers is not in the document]
davidkline-ms: > Mesh and surface observers should be different data providers, (and was treated as such in the "current" definition).

Not necessarilly. There are scenarios where a single data provider needs to be responsible for both. Requiring the data to be separated into multiple observers would lead to duplicated code and potentially out of sync data.,

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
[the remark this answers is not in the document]
davidkline-ms: Or points and planes.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/Profiles/BaseMixedRealitySpatialObserverProfile.cs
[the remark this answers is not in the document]
davidkline-ms: > The system is smart enough to setup the observer physics layers to the first availible layer and then update the default profile to it.

Wouldnt that also work if the layer were defaulted to 31 (consistency with HTK)?

Assets/MixedRealityToolkit/_Core/Definitions/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystemProfile.cs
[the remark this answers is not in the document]
davidkline-ms: The concern is that MRTK should not be reserving multiple layers by default. For green field applications, they are likely to not have pre-defined layers and customers may assume that reserved layers are required by MRTK.

ProjectSettings/TagManager.asset
davidkline-ms: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3148/files#r240744467

Assets/MixedRealityToolkit/_Core/Inspectors/PropertyDrawers/EnumFlagsAttributeDrawer.cs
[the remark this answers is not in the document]
davidkline-ms: > What's your aversion to making things sealed?

The concern is that this change is not related to the goal of the PR. Also, the code has been tested as-is for a significant amount of time. There is not actual aversion of sealing clases, just a desire to consider each case individually and to not add unrelated changes in PRs.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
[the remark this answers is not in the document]
StephenHodgson: That's because they should be different observers.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
StephenHodgson: I don't understand. It should be used for custom implementations, which is why it's in the interface.

Assets/MixedRealityToolkit/_Core/Definitions/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystemProfile.cs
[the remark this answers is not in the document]
StephenHodgson: We're not reserving anything. Just setting layers if they are free. users can edit their layers at any time. 

I don't really see any cause for concern here, as typically green field applications won't really care what layers are set.

Now established applications with layers that are already defined? Yes, I can see cause for concern there, but it's not overwriting anything that's already established, and notifies the developer if a layer is unavailable.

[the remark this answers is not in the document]
davidkline-ms: While it is true that customers can edit their layers, by adding a description, it implies that MRTK requires the layer.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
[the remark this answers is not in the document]
keveleigh: I don't think it make sense for them to be different observers for the same reason Windows gestures were combined into the device manager.

If the observers are using the same shared source of data, they should be combined into one that can do both so we don't have to keep track of the same data in multiple places and risk getting out of sync.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
davidkline-ms: Here's an example comment:

```
        /// <remarks>
        /// This method is to be called by implementations of the <see cref="IMixedRealitySpatialObserver"/> interface, and not by application code.
        /// </remarks>
```

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/Profiles/BaseMixedRealitySpatialObserverProfile.cs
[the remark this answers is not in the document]
davidkline-ms: There are some very good components and behaviors to be found in the HTK that customers have been successful with. Building upon these is important.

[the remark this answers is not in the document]
davidkline-ms: And a benefit for both contributors and customers as time tested code and practices are reused.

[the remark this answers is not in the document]
StephenHodgson: Yes, this is true, but I think something trivial as a layer assignment should be considered, esp when that can be changed. It needs to be configured by the user, and defaults set if none exist.

There's lots of pain points from HTK that we're trying to avoid as well. This is one of them.

Assets/MixedRealityToolkit/_Core/Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
StephenHodgson: Sure.

wassx:
What's the best way to set the mesh in/visible? Just dis/enable?

StephenHodgson:
> What's the best way to set the mesh in/visible? Just dis/enable?

@wassax, that's a very good question. I think Ideally I'd like to see it be the following:

```
MixedRealityToolkit.SpatialAwarenessSystem.RequestMeshVisibility(bool isVisible, enum spatialObjectType);
```

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/Profiles/BaseMixedRealitySpatialMeshObserverProfile.cs
```diff
@@ -1,13 +1,24 @@
...
 using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Core.DataProviders.SpatialObservers
 {
     public abstract class BaseMixedRealitySpatialMeshObserverProfile : BaseMixedRealitySpatialObserverProfile
     {
+        [PhysicsLayer]
+        [SerializeField]
+        [Tooltip("The optional mesh Physics layer override of the base PhysicsLayer")]
+        private int meshPhysicsLayerOverride = -1;
```
davidkline-ms: Please add a similar property to the SurfaceObserver profile. Thanks,

davidkline-ms:
> MixedRealityToolkit.SpatialAwarenessSystem.RequestMeshVisibility(bool isVisible, enum spatialObjectType);

Meshes have multiple forms of visibility (please see the MeshDisplayOption enum() that need to be supported. #3171 will support changing between them (visible mesh, occlusion mesh and disabling)

davidkline-ms:
accidentally closed when adding a comment. sorry

StephenHodgson:
I could see this being simpler still like:

```
MixedRealityToolkit.SpatialAwarenessSystem.EnableSpatialObjectVisibilityOfType(T spatialObjectType);

MixedRealityToolkit.SpatialAwarenessSystem.DisableSpatialObjectVisibilityOfType(T spatialObjectType);
```

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/Profiles/BaseMixedRealitySpatialMeshObserverProfile.cs
StephenHodgson: Done :)

davidkline-ms:
> I could see this being simpler still like:
MixedRealityToolkit.SpatialAwarenessSystem.EnableSpatialObjectVisibilityOfType(T spatialObjectType);

Except it isn't a simple Boolean choice, there are scenarios where an application needs to visualize the mesh and then apply the occlusion material. The MeshDisplayOption enum is a replacement for the HTK's draw visible meshes property and requiring the developer customer to change the material.

The proposed two method approach is more limited.

davidkline-ms:
Let's please table the visibility discussion to the pending PR for #3171 so that we can resolve the remaining issues with this one (the change to the WMR observer implementation no longer limiting to one in-flight mesh bake process).

Thanks.

StephenHodgson:
np, I didn't add the visibility stuff to this PR. I also thought it was a bit outside the scope.

StephenHodgson:
> the change to the WMR observer implementation no longer limiting to one in-flight mesh bake process

Is this the last issue to resolve?

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
```diff
@@ -37,6 +38,23 @@ protected BaseMixedRealitySurfaceObserver(string name, uint priority, BaseMixedR
             PlatformSurfaceMaterial = profile.PlatformSurfaceMaterial;
         }
 
+        /// <inheritdoc />
+        public override int PhysicsLayer
```
davidkline-ms: It seems that, since we are providing an override property, the override should report the base layer if not configured and should not replace the value of the base property (it should remain easy to access)

The implementation is also different from the mesh layer override

```
        public override int PhysicsLayer => MeshPhysicsLayerOverride >= 0 ? MeshPhysicsLayerOverride : base.PhysicsLayer;
```

davidkline-ms: Suggest something similar to the following (for both current overrides)

```
public int MeshPhysicsLayerOverride => (meshPhysicsLayerOverride > -1) ? meshPhysicsLayerOverride : base.PhysicsLayer;
```

Rather than overriding the base property.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
```diff
@@ -141,9 +138,9 @@ public override void Destroy()
         /// <summary> 
         /// To prevent too many meshes from being generated at the same time, we will 
         /// only request one mesh to be created at a time.  This variable will track 
-        /// if a mesh creation request is in flight. 
+        /// if a mesh creation request is in flight.
         /// </summary> 
-        private readonly Dictionary<int, SpatialMeshObject> outstandingMeshObjects = new Dictionary<int, SpatialMeshObject>();
+        private SpatialMeshObject? pendingSpatialObject = null;
```
davidkline-ms: Can you please also restore the spareSpatialObject field and associated logic? It helps preserve performance via object reuse vs recreation.

```
        /// <summary>
        /// When surfaces are replaced or removed, rather than destroying them, we'll keep
        /// one as a spare for use in outstanding mesh requests. That way, we'll have fewer
        /// game object create/destroy cycles, which should help performance.
        /// </summary>
        protected SpatialMeshObject? spareMeshObject = null;```

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
StephenHodgson: I don't see the difference.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
StephenHodgson: I don't understand. Can you open a pr with the suggested edits?

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
keveleigh: I think this pattern makes sense to me. This way, everything can keep referring to the generic `PhysicsLayer`, which will return the override in cases where it's set.

The logic does appear different to the mesh implementation though.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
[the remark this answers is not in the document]
StephenHodgson: > I don't think it makes sense for them to be different observers for the same reason Windows gestures were combined into the device manager.

That def makes sense.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
davidkline-ms: Will do.

Assets/MixedRealityToolkit/_Core/DataProviders/SpatialObservers/BaseMixedRealitySurfaceObserver.cs
davidkline-ms: I'll update the surfaces observer to match the mesh observer behavor when i open my PR to restore the spare object logic. Hopefully that will help clarify

davidkline-ms:
There appears to be an issue with adding / updating meshes. With a non-stationary observer, I am not receiving adds or updates reliably. Initial mesh data is received and displayed, and if I move far enough outside of the startup volume, the meshes will get removed, eventually resulting in no mesh display (only the description panel).

StephenHodgson:
> With a non-stationary observer, I am not receiving adds or updates reliably. 

No worries. Yeah I'll try and check that out too.

wassx:
Tested and it seems to fix our issue with #3226. Really looking forward seeing this in the master. Maybe with a visibility switch :)

StephenHodgson:
Agreed, really looking forward to having the services setup properly so the confusion doesn't spread.

davidkline-ms:
>  with a visibility switch :)

Visibility switch will follow this change fairly quickly. I have a prototype that will build upon this change.

wassx:
Hi guys! Sorry for the impatience, but any news on this PR?

davidkline-ms:
> Had to convert it back to c#6 equivalent

What was the reason for needing to change the syntax?

StephenHodgson:
Had to back port it because I'm ~500 commits ahead of master using 2018.3+ which takes advantage of the c# 7 features.

image

davidkline-ms:
Why close the PR and delete the branch?

Issue #3131, opened 2018-11-12
vNEXT Proposal: Demos should use the default config profile


StephenHodgson:
# Overview
Currently we have some custom configuration profiles for the demo examples. Ideally we'd like new users who are starting out to not have to choose profiles when they press configure the first time.

Currently if you download the sdk and the examples, you'll need to assign the correct profile if more than the default exists because Diagnostics has it's own custom configuration profile. ( a confused new comer had to ping me to ask which the right one was ).

![image](https://user-images.githubusercontent.com/13334553/48381603-87ee7c00-e691-11e8-8739-6b078a8916e7.png)

Ideally most/all of our demo scripts should use the default configuration profiles and should be created with that in mind.

StephenHodgson:
I do like the idea of dynamically adding speech commands.

Likely the input actions won't have to be updated to make demos work correctly.

Issue #2965, opened 2018-10-24
diagnostics scene item added twice if visible is configured to true


davidkline-ms:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
If the diagnostics profile has Visible set to true, two Diagnostics game objects will appear in the scene; 1 at edit time, the second added at runtime.

I have confirmed that existing scenes receive only the runtime game object and diagnostic data is displayed as expected.

**To Reproduce**
_(Links to sample GitHub project preferred)_
1. Enable diagnostics in the configuration profile
2. Set Visible to True in the diagnostics profile.
3. Create a new Scene
4. Click Mixed Reality Toolkit > Configure
5. Enter play mode.

**Expected behavior**
Only one Diagnostics game object to be in the scene.

**Actual behavior**
A second Diagnostics game object is added to the scene in play mode.

**Unity Editor Version**
2018.2.13f1

**Mixed Reality Toolkit Release Version**
mrtk_development as of 2pm 10/24/2018

PR #3150, opened 2018-11-16
Moved Diagnostics system from sdk -> core and updated

StephenHodgson:
Overview
---
- Moved diagnostics system from SDK to core.
- Removed demo configuration profiles
    - Updated speech commands with diagnostic keywords
- Updated demo scene to turn on diagnostics on start so we didn't have to have a custom profile for the demo.
- Diagnostics system is now patterned after the other systems and the internal data types have been turned into public structs for ease of use and better perf.

Changes
---
- Fixes: #2965
- Fixes: #3131
- addresses #3053 and #3115 for Diagnostics System and components.

Breaking Changes
---
- Users who have created their own configuration profile will need to reassign the Diagnostics type, and re-configure their diagnostics profile.


davidkline-ms:
@StephenHodgson, this is approved pending resolving the merge conflict.

StephenHodgson:
Fixed, but seemed to have reset the reviews.

Issue #3037, opened 2018-10-30
vNext: If loading a system fails with an exception, no systems are active


davidkline-ms:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
This bug should NOT stop the beta release as I have not found a 100% reliable repro,

If an exception is encountered when starting up a core or extension system, other systems are not activated (not attempted or ignored).

On HoloLens, when this occurs, a skybox is rendered and the spatial awareness system is not loaded.

**To Reproduce**
I have hit this randomly running on HoloLens. The easiest way to validate a fix is to force an exception in MixedRealityToolkit.Initialize().

**Expected behavior**
The failing system to not be available and other, unrelated systems to be active.

**Actual behavior**
Unrelated systems are not loaded.

**Unity Editor Version**
2018.2.13f1 and 2018.2.14f1

**Mixed Reality Toolkit Release Version**
mrtk_2018beta_stabilization

**Additional context**
I have a fix for continuing the Initialize method. Still need to uncover all related exceptions when later code assumes a service has successfully started.

StephenHodgson:
What's causing the system to fail loading?

davidkline-ms:
Activation failure. The times I have seen it are a nullreference exception trying to start the input system.

davidkline-ms:
It is important to both try to root cause the exception as well as be resilient to such failures.

davidkline-ms:
Chances are that the complete fix will be more involved that what we should take for beta.

StephenHodgson:
Wondering if a try catch around any activation would help.

davidkline-ms:
> Wondering if a try catch around any activation would help.

That's in my working branch right now... :)

PR #3159, opened 2018-11-16
add exception handling around Activator.CreateInstance calls

davidkline-ms:
Overview
---
Adds exception handling around usage of Activator.CreateInstance to prevent the premature termination of MixedRealityToolkit.Initialize().

Changes
---
- Fixes: #3037 .


Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -174,8 +174,16 @@ private void InitializeServiceLocator()
...
-                if (!RegisterService(typeof(IMixedRealityInputSystem), Activator.CreateInstance(ActiveProfile.InputSystemType) as IMixedRealityInputSystem) ||
-                    InputSystem == null)
+                bool registrationSucceeded = false;
+                try
+                {
+                    registrationSucceeded = RegisterService(typeof(IMixedRealityInputSystem), Activator.CreateInstance(ActiveProfile.InputSystemType) as IMixedRealityInputSystem);
+                }
+                catch (Exception e)
+                {
+                    DebugUtilities.LogCriticalError($"Registering the Input System threw: {e.GetType().ToString()} - {e.Message}.");
```
ryzngard: Should we add `DebugUtilities.LogCriticalError(string message, Exception e)` to have common formatting?

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -174,8 +174,16 @@ private void InitializeServiceLocator()
                 Utilities.Editor.InputMappingAxisUtility.CheckUnityInputManagerMappings(Definitions.Devices.ControllerMappingLibrary.UnityInputManagerAxes);
 #endif
 
-                if (!RegisterService(typeof(IMixedRealityInputSystem), Activator.CreateInstance(ActiveProfile.InputSystemType) as IMixedRealityInputSystem) ||
-                    InputSystem == null)
+                bool registrationSucceeded = false;
+                try
+                {
+                    registrationSucceeded = RegisterService(typeof(IMixedRealityInputSystem), Activator.CreateInstance(ActiveProfile.InputSystemType) as IMixedRealityInputSystem);
```
ryzngard: What if we had `RegisterService<T>(Func<T> factory)` and simplified this across all of them?

ryzngard: Callers would do `RegisterService<IMixedRealityInputSystem>(() => Activator.CreateInstance(…))`

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: maybe. Not a bad thought

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: I had thought of that as a possible optimization.

StephenHodgson:
Agreed

SimonDarksideJ:
This will cause conflicts with your new fix PR @davidkline-ms #3164

davidkline-ms:
Merged revert of debug log change. Output from test runner shown below:

capture

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -557,30 +552,78 @@ public bool RegisterService(Type type, IMixedRealityService service)
...
+
+            if (concreteType == null)
+            {
+                Debug.LogError("Unable to register a service with a null concrete type.");
+                return false;
+            }
+
+            if (!typeof(IMixedRealityService).IsAssignableFrom(concreteType))
+            {
+                Debug.LogError($"Unable to register the {concreteType.GetType()} service. It does not implement the IMixedRealityToolkitService interface.");
```
keveleigh: The interface checked above is `IMixedRealityService`, but this comment references `IMixedRealityToolkitService`. I think the comment needs to be updated.

StephenHodgson:
Looks good, just [a few CRs](https://github.com/davidkline-ms/MixedRealityToolkit-Unity/pull/5)

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
keveleigh:
```suggestion
                Debug.LogError($"Unable to register the {concreteType.GetType()} service. It does not implement the IMixedRealityService interface.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -229,12 +224,10 @@ private void InitializeServiceLocator()
                     if (Application.platform.IsPlatformSupported(configuration.RuntimePlatform))
 #endif
                     {
-                        if (configuration.ComponentType.Type != null)
+                        if (configuration.ComponentType.Type != null &&
```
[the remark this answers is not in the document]
davidkline-ms: that feels outside of the scope of this change as the original code had the != null (do not want to fundamentally change behavior here).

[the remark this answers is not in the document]
StephenHodgson: Seems reasonable as it checked if the service is registered correctly.

If the type is null then we should at least be notified.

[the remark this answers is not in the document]
StephenHodgson: In the internal registration we're already checking if the type is null.

I'm wondering if we even need this check anyway.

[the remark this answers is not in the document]
StephenHodgson: Yup. Confirmed. Redundant check can go away completely.

Issue #3162, opened 2018-11-16
vNEXT: Log utility does not navigate to the user to the log source.


StephenHodgson:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
With the new log utilities, when a user double clicks on the log message it no longer takes them to the source of the message, but instead navigates them to the debug utility class.

**To Reproduce**
1. Create a log using the log utility
2. Click on log in console
3. Takes user to the log utilities class, not the log source.

**Expected behavior**
Double clicking the log in the unity editor console should take the user to the source of the log, not the utility class that actually raises it.

Issue #3161, opened 2018-11-16
vNEXT: Configuring an empty scene for the MRTK raises errors


StephenHodgson:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
 Mixed Reality Toolkit

**Describe the bug**
When configuring a new scene, errors occur

**To Reproduce**
1. Create new scene
2. Click configure.
3. See errors

![image](https://user-images.githubusercontent.com/13334553/48650452-fd688e00-e9aa-11e8-8ef1-989c14bb6ea3.png)

SimonDarksideJ:
Again, like with #3160 we need to tighten up our testing of the project if a single PR can reduce the dev repo to this state @davidkline-ms

Issue #3160, opened 2018-11-16
vNEXT: Tests fail


StephenHodgson:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
 Mixed Reality Toolkit

Overview
---
MixedRealityToolkitTests are failing.

![image](https://user-images.githubusercontent.com/13334553/48649785-9f3aab80-e9a8-11e8-80e8-4affb5ca1a22.png)

davidkline-ms:
I believe these errors are a result of some methods previously not logging if the active profile was null. Now, activeprofile itself logs.

davidkline-ms:
Here's an example of this: EnableAllServices used to silently fail. Perhaps the test should be update to expect an assert

            //If the Mixed Reality Toolkit is not configured, stop. 
            if (activeProfile == null) { return; }

SimonDarksideJ:
Question is @davidkline-ms how did we merge the Logging PR if the tests hadn't been run?
Should not have gotten this far.  Guess we need to tighten up our testing and got to work on the automated CI testing to ensure PR's cannot be committed if existing tests are failing.

PR #3164, opened 2018-11-17
Address 3160, 3161 and 3162

davidkline-ms:
- Fixes: #3160
- Fixes: #3161 
- Fixes: #3162 (also investigating a better, longer term fix)


davidkline-ms:
These are literal string compares. It feels as if the tests (as modified) accomplish the same goal (check the return value / object state) without relying upon specific text strings being logged. It feels like one of them would have failed even before the change (was expecting a word not in the actual log message).

I would be open for an issue to track this.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -549,13 +584,19 @@ public bool RegisterService(Type type, IMixedRealityService service)
 
             if (type == null)
             {
-                DebugUtilities.LogError("Unable to add a manager of type null.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to add a manager of type null.");
```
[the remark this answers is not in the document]
davidkline-ms: Can this be a separate PR? These were the strings that previously existed.

Assets/MixedRealityToolkit-Tests/MixedRealityToolkitTests.cs
```diff
@@ -178,7 +176,6 @@ public void Test11_TestMixedRealityExtensionService2DoesNotReturn()
 
             // Validate non-existent component
             MixedRealityToolkit.Instance.GetService(typeof(ITestComponent2), "Test2");
-            LogAssert.Expect(LogType.Error, "Unable to find Test2 Manager.");
```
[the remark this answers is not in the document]
davidkline-ms: The test should probably not be checking the log string then (requires an exact match), and checking the return from GetService.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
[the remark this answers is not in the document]
StephenHodgson: No, let's get it taken care of. We're already modifying the lines anyway.

Assets/MixedRealityToolkit-Tests/MixedRealityToolkitTests.cs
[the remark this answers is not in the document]
StephenHodgson: That's how Unity's test suite works. 

Please fix the test.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -376,8 +395,12 @@ private void InitializeInstance()
                 {
                     if (instance != null)
                     {
-                        DebugUtilities.Assert(instance.transform.parent == null, "The MixedRealityToolkit should not be parented under any other GameObject!");
-                        DebugUtilities.Assert(instance.transform.childCount == 0, "The MixedRealityToolkit should not have GameObject children!");
+                        if ((ActiveProfile.LoggingLevel & LoggingLevels.Assert) != 0)
+                        {
+
```
StephenHodgson: Please remove extra line

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -549,13 +584,19 @@ public bool RegisterService(Type type, IMixedRealityService service)
...
+                }
                 return false;
             }
 
             if (service == null)
             {
-                DebugUtilities.LogError("Unable to add a manager with a null instance.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to add a manager with a null instance.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to add a service with a null instance.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -616,13 +663,19 @@ public IMixedRealityService GetService(Type type, bool showLogs = true)
 
             if (!IsInitialized)
             {
-                DebugUtilities.LogCriticalError($"Unable to get {nameof(type)} Manager as the Mixed Reality Manager has not been initialized!");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.CriticalError) != 0)
+                {
+                    Debug.LogError($"Critical: Unable to get {nameof(type)} Manager as the Mixed Reality Manager has not been initialized!");
```
StephenHodgson:
```suggestion
                    Debug.LogError($"Critical: Unable to get {nameof(type)} service as the Mixed Reality Toolkit has not been initialized!");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -616,13 +663,19 @@ public IMixedRealityService GetService(Type type, bool showLogs = true)
...
+                }
                 return null;
             }
 
             if (type == null)
             {
-                DebugUtilities.LogError("Unable to get null manager type.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to get null manager type.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get null service type.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -660,13 +716,19 @@ public IMixedRealityService GetService(Type type, string serviceName, bool showL
...
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get null service type.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -660,13 +716,19 @@ public IMixedRealityService GetService(Type type, string serviceName, bool showL
...
+                }
                 return null;
             }
 
             if (string.IsNullOrEmpty(serviceName))
             {
-                DebugUtilities.LogError("Unable to get manager by name without the name being specified.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to get manager by name without the name being specified.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get service by name without the name being specified.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -699,7 +764,10 @@ public void UnregisterService(Type type)
 
             if (type == null)
             {
-                DebugUtilities.LogCriticalError("Unable to remove null manager type.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to remove null manager type.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to remove null service type.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -683,7 +745,10 @@ public IMixedRealityService GetService(Type type, string serviceName, bool showL
 
             if (service == null && showLogs)
             {
-                DebugUtilities.LogWarning($"Unable to find {serviceName} Manager.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Warning) != 0)
+                {
+                    Debug.LogWarning($"Unable to find {serviceName} service.");
```
StephenHodgson: > Let's do in a separate PR

But we already made a [partial change](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3164/files#diff-f21536a45f2f15aa366accfb4e1e7d9bL686). Let's go all the way.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -731,13 +799,19 @@ public void UnregisterService(Type type, string serviceName)
 
             if (type == null)
             {
-                DebugUtilities.LogCriticalError("Unable to remove null manager type.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to remove null manager type.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to remove null service type.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -731,13 +799,19 @@ public void UnregisterService(Type type, string serviceName)
...
+                }
                 return;
             }
 
             if (string.IsNullOrEmpty(serviceName))
             {
-                DebugUtilities.LogCriticalError("Unable to remove manager by name without the name being specified.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to remove manager by name without the name being specified.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to remove service by name without the name being specified.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -764,7 +838,10 @@ public void DisableService(Type type)
         {
             if (type == null)
             {
-                DebugUtilities.LogError("Unable to disable null manager type.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to disable null manager type.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to disable null service type.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -792,13 +869,19 @@ public void DisableService(Type type, string serviceName)
...
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to disable null service type.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -792,13 +869,19 @@ public void DisableService(Type type, string serviceName)
...
+                }
                 return;
             }
 
             if (string.IsNullOrEmpty(serviceName))
             {
-                DebugUtilities.LogError("Unable to disable manager by name without the name being specified.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to disable manager by name without the name being specified.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to disable service by name without the name being specified.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -825,7 +908,10 @@ public void EnableService(Type type)
         {
             if (type == null)
             {
-                DebugUtilities.LogError("Unable to enable null manager type.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to enable null manager type.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to enable null service type.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -853,13 +939,19 @@ public void EnableService(Type type, string serviceName)
...
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to enable null service type.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -853,13 +939,19 @@ public void EnableService(Type type, string serviceName)
...
+                }
                 return;
             }
 
             if (string.IsNullOrEmpty(serviceName))
             {
-                DebugUtilities.LogError("Unable to enable manager by name without the name being specified.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to enable manager by name without the name being specified.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to enable service by name without the name being specified.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -891,7 +983,10 @@ public List<IMixedRealityService> GetActiveServices(Type type)
         {
             if (type == null)
             {
-                DebugUtilities.LogWarning("Unable to get managers with a type of null.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to get managers with a type of null.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get service with a type of null.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -910,7 +1005,10 @@ public List<IMixedRealityService> GetActiveServices(Type type, string serviceNam
 
             if (type == null)
             {
-                DebugUtilities.LogWarning("Unable to get managers with a type of null.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to get managers with a type of null.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get service with a type of null.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -1075,7 +1173,10 @@ private static bool IsCoreSystem(Type type)
         {
             if (type == null)
             {
-                DebugUtilities.LogWarning("Null cannot be a core manager.");
+                if ((Instance?.ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Null cannot be a core manager.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Null cannot be a core system!");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -1104,7 +1205,10 @@ private void GetService(Type type, out IMixedRealityService service)
         {
             if (type == null)
             {
-                DebugUtilities.LogWarning("Unable to get a component with a type of null.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to get a component with a type of null.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get a service with a type of null.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -1122,7 +1226,10 @@ private bool GetService(Type type, string serviceName, out IMixedRealityService
...
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get a service with a type of null.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -1156,7 +1266,10 @@ private void GetService(Type type, ref List<IMixedRealityService> services)
         {
             if (type == null)
             {
-                DebugUtilities.LogWarning("Unable to get components with a type of null.");
+                if ((ActiveProfile.LoggingLevel & LoggingLevels.Error) != 0)
+                {
+                    Debug.LogError("Unable to get components with a type of null.");
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get services with a type of null.");
```

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -1167,7 +1280,10 @@ private void GetServices(Type type, string serviceName, ref List<IMixedRealitySe
...
```
StephenHodgson:
```suggestion
                    Debug.LogError("Unable to get services with a type of null.");
```

Assets/MixedRealityToolkit-Tests/MixedRealityToolkitTests.cs
```diff
@@ -33,8 +33,6 @@ public void Test01_InitializeMixedRealityToolkit()
         [Test]
         public void Test02_TestNoMixedRealityConfigurationFound()
         {
-            LogAssert.Expect(LogType.Error, "No Mixed Reality Configuration Profile found, cannot initialize the Mixed Reality Toolkit");
```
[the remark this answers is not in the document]
davidkline-ms: This doesn't break the test case.. The Assert.IsFalse(MixedRealityToolkit.HasActiveProfile);  and Assert.IsNull(MixedRealityToolkit.Instance.ActiveProfile);  are the true tests and are less fragile (a one character change in the code breaks string based comparisons like this).

Assets/MixedRealityToolkit-Tests/MixedRealityToolkitTests.cs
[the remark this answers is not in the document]
davidkline-ms: I will implement a less fragile version of the test by checking that the value returned by GetService is null, using Assert.IsTrue()

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -177,7 +178,10 @@ private void InitializeServiceLocator()
                 if (!RegisterService(typeof(IMixedRealityInputSystem), Activator.CreateInstance(ActiveProfile.InputSystemType) as IMixedRealityInputSystem) ||
                     InputSystem == null)
                 {
-                    DebugUtilities.LogCriticalError("Failed to start the Input System!");
+                    if ((ActiveProfile.LoggingLevel & LoggingLevels.CriticalError) != 0)
```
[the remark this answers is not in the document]
StephenHodgson: Yeah it really seems like we're just duplicating the existing functionality of the editor...

https://docs.unity3d.com/ScriptReference/Logger.html

Assets/MixedRealityToolkit-Tests/MixedRealityToolkitTests.cs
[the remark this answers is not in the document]
davidkline-ms: > We also test the errors we "Expect" when something is not correct. 

While I agree that this is a good endeavor, it can make the tests fragile. If a method changes the log string (ex: a spelling error is fixed, the strings get localized, or if we add a feature like this one, that allows developers to disable some log messages) the test will fail.

One possible solution is to return error _codes_ and validate those values, rather than a specific log string.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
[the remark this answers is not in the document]
davidkline-ms: Fair enough (and Stephen already changed the values).

Assets/MixedRealityToolkit-Tests/MixedRealityToolkitTests.cs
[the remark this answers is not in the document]
davidkline-ms: Part of the feedback on this PR was to consolidate the message that occurs when ActiveProfile is null (have the getter log the messsage). That resulted in the system displaying an error when the MixedRealityToolkit component is first added to the scene (by default, there is no profile configured).

Is the preference to go back to logging a message in each caller if they detect a null ActiveProfile? I am fine with that.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -69,17 +69,18 @@ public MixedRealityToolkitConfigurationProfile ActiveProfile
...
-                }
 #if UNITY_EDITOR
                 if (!Application.isPlaying && activeProfile == null)
                 {
                     UnityEditor.Selection.activeObject = Instance;
                     UnityEditor.EditorGUIUtility.PingObject(Instance);
                 }
+#else
+                // If there is no active profile, the system cannot function. _Always_ display this message in the debug output.
+                if (activeProfile == null)
```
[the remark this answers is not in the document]
davidkline-ms: > this call should NOT have been moved from it's original location,

That change was made based on feedback... I can easily re-add the log messages at the points of access.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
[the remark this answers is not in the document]
davidkline-ms: We received customer feedback to enable a path to configure the volume of the logging from the toolkit. This PR's pattern has been successfully leveraged in past projects.

> why would Anyone ever want to "hide" critical errors?

Adding the ability to filter critical errors is about customer choice. Some developers do not wish to see the error logs and simply handle return values to check for errors.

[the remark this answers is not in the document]
StephenHodgson: I completely disagree. The fact that there's a critical error needs to be addressed by the developer in question or the whole system will not work. If they don't wish the critical error in question, about the status of the _entire_ system, then they can prevent this message by removing the toolkit all-together.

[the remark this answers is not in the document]
StephenHodgson: TL;DR Errors like these shouldn't be simply ignored, but fixed and addressed.

davidkline-ms:
Looking at this:

109158-unity-logger-update

It appears that unity provides a similar solution (with a few more clicks). I'll approve the rollback PR (#3167) and check back with the customer.

PR #3166, opened 2018-11-18
Utility to swap source code with binary assembly compiled by editor

StephenHodgson:
Overview
---
This utility adds the ability to swap out the source code for the compiled assembly found in the project's `Library/ScriptAssemblies` folder. The source code is then hidden from the editor by pre-pending the name with a period.

Using this utility reduces the number of projects found in the solution when the assembly is used in place of the source code and enables platform specific code to run in the editor (For example, future usage with WMR controllers).
image

To replace source code with assembly:
---
- Select any Assembly Definition
- Click on the settings (gear) icon in the upper right of the inspector window
- Click replace source with assembly

To replace assembly with source code:
---
- Select either the plugin, or the Assembly Definition
- Click on the settings (gear) icon in the upper right of the inspector window
- Click Replace assembly with source

Known issues:
---
- Can **_ONLY_** be used on assembly definitions that do not contain objects that derive from `UnityEngine.Object`
- Sometimes when replacing the source with the assembly there is a reflection invocation exception about a missing `UnityEngine.CoreModule`. Seems to finish importing correctly and works as expected. Might be a unity bug? More research is needed. This error can be safely dismissed.
    - I thought about just catching the exception, but I'm waiting to follow up from Unity QA about possible causes and even solutions.
- Sometimes when replacing the source with the assembly and the user tabs out or the editor loses focus and then regains focus, it may raise warnings about missing meta files. This warning can be safely dismissed.


Jarodshow:
> Using this utility reduces the number of projects found in the solution when the assembly is used in place of the source code and enables platform specific code to run in the editor (For example, future usage with WMR controllers).

How does this work?

StephenHodgson:
Just copies the dll into the main project and puts a period in front of all the source code file names to hide them from the editor.

Jarodshow:
Also, the [.asmdef.json file for MixedRealityToolkit.Core.Utilities.Editor.Setup](https://gist.github.com/Jarodshow/f540452d2b9f50d9b4b02b07cffbdcb1) should be included.

StephenHodgson:
Lol that's actually the first dll I created with this utility.

I just didn't add it to the MRTK until now

SimonDarksideJ:
@davidkline-ms can we get this PR moving, there should be no reason it's been hanging around for this long.
This ultimately has long term benefits as we can reduce the number of projects appearing in a users runtime project

StephenHodgson:
Yup, it'll also make re-loading the project much faster as well.

davidkline-ms:
> there should be no reason it's been hanging around for this long.

Breaking change PRs have taken priority over this one, @keveleigh is testing and reviewing now.

StephenHodgson:
@keveleigh what are your repro steps?

StephenHodgson:
Oh, nvm I see what happened. The script references needed to be updated too. Well, looks like we can only do this with non-monobehavior scripts. Back to the drawing board.

keveleigh:
@StephenHodgson 

1. Pull this PR's branch
2. Open in Unity 2018.2.18f1 (both Standalone and UWP fail the same way)
3. Click on MixedRealityToolkit.asmdef
4. Right Click -> Replace Source with Assembly
5. Wait for source to be replaced
6. View the Console
7. Wait ~5 seconds for things to reload
8. Observe the above behavior

StephenHodgson:
Oh this can also be used on editor assemblies as well

StephenHodgson:
Going to close this until I figure out how to update the `UnityEngine.Object references`

Issue #3170, opened 2018-11-20
vNEXT Task: Refactor services to have their own registry of data providers


StephenHodgson:
# Overview
Currently all of our Devices are registered as extension services in the main configuration profile, which makes them easy to find, but hard to understand which systems they go to and what their purpose is. Ideally we shouldn't be registering devices as extension services, as that extension service profile should only be for custom services.

There are a few things we can do, such as giving each system its own registry, with more specific interface types to adhere to, so we don't get bombarded with a huge list of services to register. 

I'd also like to rename the "Devices" to to "Data Providers". It made sense for the input system at the time, but as we add more systems (like spatial understanding or networking) we want to categorize these modules to better explain what their function is. This will also give devices a clearer distinction than just a "service" but rather something a service relies on and consumes data from.

All folders and namespaces with "Device" will be renamed to "DataProvider".

for example: 

- `Microsoft.MixedReality.Toolkit.Core.DataProvider.<Type>.<API/Platform>`
- `Microsoft.MixedReality.Toolkit.Core.DataProvider.Controller.WindowsMixedReality`
- `Microsoft.MixedReality.Toolkit.Core.DataProvider.SpatialObservers.WindowsMixedReality`
- `Microsoft.MixedReality.Toolkit.Core.DataProvider.Networking.WebRTC`

One of the biggest changes here is that the device managers currently registered for the input system will move from the extension service registry into the controller mapping profile. This will look very similar to the extension service registry, but with a few key differences:

- Users will be able to only register data provider types

For spatial awareness, the same kind of rule would apply where users would only be able to register spatial observer data providers in a similar registry.

As services need be registered by the `MixedRealityToolkit` service locator as it's initialized, we'll need to add an additional loop in each of the core system registration areas where we check the system's profile for registered data providers and register their services.

# TL;DR

- Extension service profile should only be for custom services that users create for their application.
- Devices (DataProviders) will be registered with each system using a more explicit type.

# Requirements

- A data provider is an abstract hardware/software representation of a resource located outside of the application
- Systems should have a registry of data provider implementations and manage those data providers accordingly
- Data providers should never be accessed by other systems, or services directly, and absolutely never by the presentation & application layer
- The systems should have a means of forwarding events, and providing the application/presentation layer with the expected data in a standardized format.
- Data providers will likely have a base interface/implementation that will provide the common data that the system expects from all resources that it will manage
[...]

PR #3173, opened 2018-11-20
Base PR for IMixedRealityDataProviders

StephenHodgson:
Overview
---
- Added `IMixedRealityDataProvider` interface and base abstract type.
- Updated `SystemType` to filter out abstract types from being assigned in the dropdown inspector.
- Added assert to the base profile inspector to ensure that the `Base.OnEnable()` is being called correctly from derived types.
- Updated public `MixedRealityComponents` to be a readonly list and renamed it to `RegisteredMixedRealityServices`

Changes
---
- Fixes: #3170


Assets/MixedRealityToolkit/_Core/Services/BaseDataProvider.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Core.Services
+{
+    /// <summary>
+    /// The base data provider implements <see cref="IMixedRealityDataProvider"/> and provides default properties for all data providers.
+    /// </summary>
+    /// <remarks>
+    /// Empty, but reserved for future use, in case additional <see cref="IMixedRealityDataProvider"/> properties or methods are assigned.
+    /// </remarks>
+    public abstract class BaseDataProvider : BaseServiceWithConstructor, Interfaces.IMixedRealityDataProvider
```
davidkline-ms: Curious if we actually need a contstructor-less base implementation. Wouldn't having a base and a basewithconstructor create potential traps (read: easy to make mistakes) for customers creating services and providers?

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -2,9 +2,11 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.MixedReality.Toolkit.Core.Definitions;
+using Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities;
 using Microsoft.MixedReality.Toolkit.Core.Extensions;
 using Microsoft.MixedReality.Toolkit.Core.Interfaces;
 using Microsoft.MixedReality.Toolkit.Core.Interfaces.BoundarySystem;
+//using Microsoft.MixedReality.Toolkit.Core.Interfaces.DataProviders.SpatialObservers;
```
davidkline-ms: Presumably, this wont be needed Systems which require data providers would be responsible for initiating load correct? That would also potentially allow them to set additional requirements (ex; implemented interfaces).

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -190,7 +200,19 @@ private void InitializeServiceLocator()
             // If the Spatial Awareness system has been selected for initialization in the Active profile, enable it in the project
             if (ActiveProfile.IsSpatialAwarenessSystemEnabled)
             {
-                if (!RegisterService<IMixedRealitySpatialAwarenessSystem>(ActiveProfile.SpatialAwarenessSystemSystemType) || SpatialAwarenessSystem == null)
+                if (RegisterService<IMixedRealitySpatialAwarenessSystem>(ActiveProfile.SpatialAwarenessSystemSystemType) && SpatialAwarenessSystem != null)
+                {
+                    //if (ActiveProfile.SpatialAwarenessProfile.SpatialObserverDataProviders != null &&
```
davidkline-ms: Along the lines of the earlier comment. If the toolkit component provided a method that systems which use data providers to call at initialization time, we can save a good deal of code here and allow systems / extension services to add/remove data provider requirements without requiring the toolkit component to change.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -557,12 +569,22 @@ public bool RegisterService(Type type, IMixedRealityService service)
         /// </summary>
         /// <typeparam name="T">The interface type for the system to be registered.</typeparam>
         /// <param name="concreteType">The concrete type to instantiate.</param>
+        /// <param name="supportedPlatforms">The runtime platform to check against when registering.</param>
         /// <param name="args">Optional arguments used when instantiating the concrete type.</param>
         /// <returns>True, if the service was successfully registered.</returns>
-        public bool RegisterService<T>(Type concreteType, params object[] args)
+        public bool RegisterService<T>(Type concreteType, SupportedPlatforms supportedPlatforms = (SupportedPlatforms)(-1), params object[] args)
```
davidkline-ms: How were services being limited to platforms previously?  Does this duplicate work?

davidkline-ms: Mostly asking because I do not see (in this PR) a removal of the old filtering methology

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
```diff
@@ -557,12 +569,22 @@ public bool RegisterService(Type type, IMixedRealityService service)
...
         {
             if (ActiveProfile == null) { return false; }
 
+#if !UNITY_EDITOR
+            if (!Application.platform.IsPlatformSupported(supportedPlatforms))
+#else
+            if (!UnityEditor.EditorUserBuildSettings.activeBuildTarget.IsPlatformSupported(supportedPlatforms))
+#endif
+            {
+                return false;
```
davidkline-ms: add a Debug.LogWarning here?

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
StephenHodgson: All systems should be platform agnostic.

Extension services may or may not be platform specific.

Data providers are platform specific.

Previously we were iterating over the list of extension services and filtering in the for loop.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
StephenHodgson: No, because it could be a valid state to return here.

StephenHodgson: We don't want warnings that a service wasn't registered even though it's the correct behavior.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
keveleigh: @davidkline-ms removal of the old filtering methodology [is here](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3173/files/633623d5b39bacd549ebc8f72bdf94a93e41b483#diff-f21536a45f2f15aa366accfb4e1e7d9bL222)

StephenHodgson: Thanks, That's the section I was referring to

davidkline-ms: Thanks for pointing that out @keveleigh. I knew I was scanning over something without noticing it. :)

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: > All services should be registered with the service locator only.

True, though it might be useful / easier to read if the services triggered loading their data providers during initialization.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: > This namespace is here so we can iterate through the spatial awareness service's data providers.

It may likely that data providers do not require any additional methods. The way that #3180 is reworking the observers, that will be true, We can leave the comment here until that is completed.

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
StephenHodgson: By the time initialization happens all services should already be registered

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: Fair enough. we have lots of log messages as it is :)

Assets/MixedRealityToolkit/_Core/Services/MixedRealityToolkit.cs
davidkline-ms: Not a blocker at this time, more of something to consider when data providers are added.

Assets/MixedRealityToolkit-Tests/MixedRealityToolkitTests.cs
```diff
@@ -156,7 +156,7 @@ public void Test10_TestMixedRealityExtensionServices()
             MixedRealityToolkit.Instance.RegisterService(typeof(ITestComponent1), new TestComponent1("Test Component 1", 10));
             MixedRealityToolkit.Instance.RegisterService(typeof(ITestComponent2), new TestComponent2("Test Component 2", 10));
             MixedRealityToolkit.Instance.RegisterService(typeof(IFailComponent), new TestFailComponent("Fail Component", 10));
-            LogAssert.Expect(LogType.Error, $"Unable to register {typeof(IFailComponent)}. Concrete type does not implement the IMixedRealityExtensionService implementation.");
+            LogAssert.Expect(LogType.Error, $"Unable to register {typeof(IFailComponent)}. Concrete type does not implement IMixedRealityExtensionService nor IMixedRealityDataProvider.");
```
davidkline-ms:
```suggestion
            LogAssert.Expect(LogType.Error, $"Unable to register {typeof(IFailComponent)}. Concrete type does not implement IMixedRealityExtensionService or IMixedRealityDataProvider.");
```

StephenHodgson:
I might actually do the update to the tests in the next PR. 

I'd like to clean that up and organize it a bit as well.

StephenHodgson:
Okay this PR is ready.

My next PR will address some other changes and add better unit tests

Issue #2861, opened 2018-09-30
Motion controller has a high chance of being not recognized in Unity 2018.2.10


marklinmao:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
HoloToolkit

**Describe the bug**
Motion controller not recognized in Unity **2018.2.10f0**. The model fails back to a white cube and no response to trigger events

**To Reproduce**
_(Links to sample GitHub project preferred)_
1. Create a new Unity project, import MRTK, Apply the project settings (for 'occluded headset') and scene settings.
2. Connect headset and 'play'.
3. Click the trigger on the controller
Please refer to the attached image for exception info.

**Expected behavior**
The 

**Actual behavior**
You will see the there is high change (80%) that the controller is not recognized(it falls back to be visualized as a white cube), while 20% chance it is. This problem was not existing in earlier version(e.g. **2018.1.1f1**) of Unity.

**Screenshots**
![lostcontorllerproblem](https://user-images.githubusercontent.com/4719869/46259324-4e462500-c4d8-11e8-95ea-743071f76e28.png)

**Unity Editor Version**
2018.2.10f0

**Mixed Reality Toolkit Release Version**
HoloToolkit-Unity-2017.4.1.0

Cheesecakes:
I'm also seeing this issue in 2018.2.9f1.
Very annoying.

marklinmao:
Any news about this issue, please?

jerrygg:
@keveleigh Please fix it, it has blocked the normal development!

keveleigh:
@jerrygg Please take a look at #3181. I believe this is fixed in that PR.

marklinmao:
As I verified, it is fixed! Thank you @keveleigh !

keveleigh:
This should be fixed in the 2017.4.3 release. Please reopen if not.

marklinmao:
Nice to hear that, Thank!

Best Regards,
Mao LIN


Kurtis <notifications@github.com> 于 2018年11月26日周一 20:51写道：

Issue #2805, opened 2018-09-18
Motion controller models not always loaded when playing in editor


ivan2007:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
HoloToolkit

**Describe the bug**
Sometimes, playing in editor, the motion controller models are not loaded.
This happens systematically after the debug with visual studio.

**To Reproduce**
1) Import MRTK and  InputManagerTestScene
2) Press Play in Unity (the motion controller models are loaded) 
3) Attach the VS debug
4) Detach debug
5) Stop the player
6) Press play in Unity (the motion controller models are not loaded) 

**Expected behavior**
Motion controller models always loaded

**Actual behavior**
I made some test and these are my notes:
1) The models are not loaded because there aren't detected spatial sources ( no sources from spatialInteractionManager->GetDetectedSourcesAtTimestamp(perceptionTimestamp) of EditorMotionController).
2) StartTrackingController is called from Awake (if InteractionManager retrives sources) and when an interaction source is detected
3) When the models are not loaded StartTrackingController is called from Awake. In this case if StartTrackingController is called when an interaction source is detected the models are loaded.
To test this I made a loop inside LoadSourceControllerModel to retry when the model is not loaded and this is the result:
![image](https://user-images.githubusercontent.com/17752475/45698637-c15eab80-bb68-11e8-9a43-6a997bd40449.png)

Only after the sources are detected the models are loaded.
It seems that, sometimes, also if InteractionManager retrives source states, these are not ready until the InteractionSourceDetected event is raised.

**Unity Editor Version**
2017.4.11f1

**Mixed Reality Toolkit Release Version**
2017.4.0.0

StephenHodgson:
Thanks for the write up!

ivan2007:
Many thanks for all your hard work on this project!

StephenHodgson:
I wonder if we're forgetting to get a list of already active controllers, and only subscribing to the events

stufflesimpsons:
Can someone please show me how to get all this stuff off my phone.😃

keveleigh:
This should be fixed in the 2017.4.3 release. Please reopen if not.

Issue #2682, opened 2018-08-31
When a controller model is not loaded switching off and on again the controller does not change anything


ivan2007:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
HoloToolkit

**Is your feature request related to a problem? Please describe.**
Sometimes the controllers model are not loaded. 
I saw this both in the editor and in the final version of my application but I was not able to debug it (I saw the message 'Model task is null; loading alternate').
When this happens by switching off and on again the controller does not change anything because the controller key is not removed when the controller is unloaded.
 
**Describe the solution you'd like**
If a controller model is not loaded when I switch off it I suggest to remove the key and destroy the alternate model.

**Additional context**
MRTK 2017.4.0
Unity 2017.4.9f1

keveleigh:
This should be fixed in the 2017.4.3 release. Please reopen if not.

Issue #1645, opened 2018-01-17
Motion controller model not always loaded when playing in Editor


ivan2007:
## Overview
Motion controller model not always loaded when playing in Editor
## Expected Behavior
Motion controller model loaded when playing in Editor
## Actual Behavior
Sometimes the motion controller model is not loaded when playing in Editor
## Steps to reproduce
Play in editor
Immediately activate the window of another program

I made a little debug taking a look at the source code of MotionControllerModel.dll.
The dll make a request of SpatialInteractionManager for the active window, so if you change the window when the toolkit is trying to get the model, the model is not loaded.
I modified the code of the dll to trace some messages.

When the model is not loaded because the active window is not Unity I get:
![image](https://user-images.githubusercontent.com/17752475/35069941-f4aae1d6-fbdb-11e7-86d6-7eebb1b1cbdd.png)

When the model is loaded the active window is with class name 'UnityHoloInEditorWndClass' and I get:

![image](https://user-images.githubusercontent.com/17752475/35070025-4f145904-fbdc-11e7-8119-05b86b784087.png)

Sometimes also if the focus is on Unity, the model is not loaded. In this case the active window is with class name 'UnityContainerWndClass' (the main window of Unity):
![image](https://user-images.githubusercontent.com/17752475/35070222-e03e7a54-fbdc-11e7-9660-a4bd1a5af755.png)

When the model is not loaded it seems that Unity doesn't have a top level window with class name 'UnityHoloInEditorWndClass':
![image](https://user-images.githubusercontent.com/17752475/35070419-86940d6a-fbdd-11e7-8369-72fbd94272eb.png)

When the model is loaded it seems that Unity always has a window with class name 'UnityHoloInEditorWndClass' but it is not a top level window:
![image](https://user-images.githubusercontent.com/17752475/35070482-bd615f64-fbdd-11e7-904e-6ab814866145.png)

When the model is not loaded this two workarounds can be usefull:
1) Stop the app, close the portal and restart the app letting Unity to start the portal.
or
2) Stop the app and switch off the controllers. Start the app and then switch on the controllers.

## Unity Editor Version
2017.3.0f3
## Mixed Reality Toolkit Release Version
Current release

DDReaper:
The gLTF loader is being updated with the latest version from the provider, so this test case will need re-testing then. 
Nut it sounds like a Unity importer issue with them disposing but not properly regenerating internal code when the Unity editor goes out of focus.

[...]

PR #3181, opened 2018-11-21
Delay trying to load controller models to enable Editor loading

keveleigh:
Overview
---
When launching in the editor, there's a race condition between sources being present in the `DetectedSources` list and the editor's holographic window being available, which is needed to get the platform sources to load the model.

To resolve this, I moved the initial attempt to load out of both `Awake` and `onBeforeRender`, so it now only tries to load in `Update` or `SourceDetected`. I've found this to fix controller model loading in the editor in all my tests.

Changes
---
- Fixes: #1645
- Fixes: #2682
- Fixes: #2805
- Fixes: #2861


keveleigh:
@marklinmao Sounds like you might be using an older release. The most recent release (2017.4.2) updated the model's loaded coordinate system, so this version of the script accounts for the difference.

Issue #3185, opened 2018-11-24
vNEXT Proposal: Refactor MixedRealityToolkit.ActiveProfile.ActiveServices into core service locator


StephenHodgson:
# Overview
This dictionary of `ActiveServices` should probably live in the `MixedRealityToolkit`, and renamed to `ActiveSystems`.

PR #3186, opened 2018-11-26
Updated service locator and unit tests

StephenHodgson:
Overview
---
- Updated the service locator with better names, cleaner more compact code, and unit tests.
- Organized unit tests a bit more.
- Renamed `EnableService` to `EnableAllServicesByType`
- Renamed `DisableService` to `DisableAllServicesByType`
- When un-registering a service it will also call disable and destroy on that service, much how like the service registration will call initialize and enable on the service if the service locator has already went through its initialization.

Changes
---
- Fixes: #3185 

Blocked by
---
- [x] #3173


Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -0,0 +1,543 @@
...
+            GameObject gameObject = GameObject.Find(nameof(MixedRealityToolkit));
+            Assert.AreEqual(nameof(MixedRealityToolkit), gameObject.name);
+        }
+
+        [Test]
+        public void Test_02_TestNoMixedRealityConfigurationFound()
+        {
+            LogAssert.Expect(LogType.Error, "No Mixed Reality Configuration Profile found, cannot initialize the Mixed Reality Toolkit");
+
+            MixedRealityToolkit.Instance.ActiveProfile = null;
```
davidkline-ms: This test appears to set the value it is expecting and only truly checking the log string... Is this intended?

Since the log message precedes setting ActiveProfile to null, it appears that

MixedRealityToolkit.Instance.ActiveProfile = null;

is not necessary for the test.

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -0,0 +1,543 @@
...
+
+            TestUtilities.CleanupScene();
+        }
+
+        [Test]
+        public void Test_03_CreateMixedRealityToolkit()
+        {
+            TestUtilities.InitializeMixedRealityToolkitScene();
+
+            // Create Test Configuration
```
davidkline-ms: This comment does not appear accurate, the line above appears to do the configuration creation and the asserts validate the results.

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -0,0 +1,543 @@
...
+
+            // Tests
+            Assert.IsNotNull(extensionService1);
+        }
+
+        [Test]
+        public void Test_04_02_01_UnregisterMixedRealityDataProviderByType()
+        {
+            // Redo register test service for a valid instance.
+            Test_04_01_RegisterMixedRealityDataProvider();
```
davidkline-ms: Rather than re-running a test (which may fail) please consider a helper function that sets the desired initial conditions?

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -0,0 +1,543 @@
...
+            Assert.IsEmpty(MixedRealityToolkit.ActiveSystems);
+            Assert.AreEqual(2, MixedRealityToolkit.RegisteredMixedRealityServices.Count);
+            Assert.AreEqual(extensionServices.Count, MixedRealityToolkit.RegisteredMixedRealityServices.Count);
+        }
+
+        [Test]
+        public void Test_04_04_UnregisterMixedRealityDataProvidersByType()
+        {
+            // Redo register test services for a valid instances.
+            Test_04_03_RegisterMixedRealityDataProviders();
```
davidkline-ms: Rather than re-running a test (which may fail) please consider a helper function that sets the desired initial conditions?

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -0,0 +1,543 @@
...
+
+            // Tests
+            Assert.IsNotNull(extensionService1);
+        }
+
+        [Test]
+        public void Test_05_02_01_UnregisterMixedRealityExtensionServiceByType()
+        {
+            // Redo register test service for a valid instance.
+            Test_05_01_RegisterMixedRealityExtensionService();
```
davidkline-ms: Rather than re-running a test (which may fail) please consider a helper function that sets the desired initial conditions?

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -0,0 +1,543 @@
...
+            Assert.IsEmpty(MixedRealityToolkit.ActiveSystems);
+            Assert.IsEmpty(MixedRealityToolkit.RegisteredMixedRealityServices);
+            LogAssert.Expect(LogType.Error, $"Unable to find {typeof(ITestExtensionService1).Name} service.");
+        }
+
+        [Test]
+        public void Test_05_02_02_UnregisterMixedRealityExtensionServiceByTypeAndName()
+        {
+            // Redo register test service for a valid instance.
+            Test_05_01_RegisterMixedRealityExtensionService();
```
davidkline-ms: Rather than re-running a test (which may fail) please consider a helper function that sets the desired initial conditions?

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -0,0 +1,543 @@
...
+            Assert.IsEmpty(MixedRealityToolkit.ActiveSystems);
+            Assert.AreEqual(2, MixedRealityToolkit.RegisteredMixedRealityServices.Count);
+            Assert.AreEqual(extensionServices.Count, MixedRealityToolkit.RegisteredMixedRealityServices.Count);
+        }
+
+        [Test]
+        public void Test_05_04_UnregisterMixedRealityExtensionServicesByType()
+        {
+            // Redo register test services for a valid instances.
+            Test_05_03_RegisterMixedRealityExtensionServices();
```
davidkline-ms: Rather than re-running a test (which may fail) please consider a helper function that sets the desired initial conditions?

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
StephenHodgson: The test is that it fails, so yes.

That's what was there previously. This test was not modified.

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
StephenHodgson: Why when the state is satisfactory from the previous test? 

If the previous test would fail then this one should too.

davidkline-ms: In my previous test dev life, I found tests to be best (and most reliable) when they could be run completely in isolation. It is possible to train the system if one test depends upon another.

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
davidkline-ms: > That's what was there previously. This test was not modified.

Understood. I happened to notice it on this review. Sorry.

> The test is that it fails, so yes.

Is the test that you explicity set ActiveProfile to null and then expect the values to be updated? Or is it that not initializing has the values set as asserted?

Given that the log assert is above setting the profile to null, i suspect the latter.

StephenHodgson: Yes the latter, but the first sounds like a nice addition to our tests.

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
StephenHodgson: Moving the code to a different location doesn't add any benefit, and adds more overhead to maintainability.

While I do agree that tests should run in isolation, I think that applies to the specific modules being tested.

Assets/MixedRealityToolkit-Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
SimonDarksideJ: Agreed, Tests should be self contained.  If needed, use shared initialisers to "setup" the test, but tests should never be dependant on each other.

StephenHodgson:
Will do.

StephenHodgson:
Any update on this guys?

StephenHodgson:
Thanks @keveleigh!

PR #3188, opened 2018-11-26
Audio Manager 2.0

joekellyms:
Overview
---
Added node-based editor for audio events and runtime audio manager.

### DIFF
--- a/Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/AudioManagerTest.cs
+++ b/Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/AudioManagerTest.cs
@@ -1,4 +1,7 @@
-﻿using UnityEngine;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
 using Microsoft.MixedReality.Toolkit.Audio;
 
 public class AudioManagerTest : MonoBehaviour
--- a/Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/AudioManagerTest.cs
+++ b/Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/AudioManagerTest.cs
@@ -1,0 +1,13 @@
+﻿using UnityEngine;
+using Microsoft.MixedReality.Toolkit.Audio;
+
+public class AudioManagerTest : MonoBehaviour
+{
+    [SerializeField]
+    private AudioEvent testEvent;
+
+    private void Start()
+    {
+        AudioManager.PlayEvent(this.testEvent, this.gameObject);
+    }
+}
--- a/Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
+++ b/Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
@@ -1,0 +1,118 @@
+﻿using System;
+
+namespace Microsoft.MixedReality.Toolkit.Audio
+{
+    /// <summary>
+    /// The runtime instance of an AudioParameter on an ActiveEvent
+    /// </summary>
+    public class ActiveParameter : IDisposable
+    {
+        /// <summary>
+        /// The value of the parameter
+        /// </summary>
+        private float currentValue = 0;
+        /// <summary>
+        /// The result of the parameter's curve on the currentValue
+        /// </summary>
+        private float currentResult = 0;
+        /// <summary>
+        /// Has the ActiveParameter been set to a value independent from the main AudioParameter
+        /// </summary>
+        private bool isDirty = false;
+        /// <summary>
+        /// Whether the object is being disposed
+        /// </summary>
+        private bool disposed = false;
+
+        /// <summary>
+        /// Constructor: Create a new ActiveParameter from the EventParameter
+        /// </summary>
+        /// <param name="root">The EventParameter to apply to this event</param>
+        public ActiveParameter(EventParameter root)
+        {
+            this.rootParameter = root;
+        }
+
+        /// <summary>
+        /// The EventParameter being used
+        /// </summary>
+        public EventParameter rootParameter { get; private set; }
+        /// <summary>
+        /// The value of the root parameter, unless the ActiveParameter has been independently set
+        /// </summary>
+        public float CurrentValue
+        {
+            get
+            {
+                if (this.isDirty)
+                {
+                    return this.currentValue;
+                }
+                else
+                {
+                    return this.rootParameter.parameter.CurrentValue;
+                }
+            }
+            set
+            {
+                this.currentValue = value;
+                this.currentResult = this.rootParameter.ProcessParameter(this.currentValue);
+                this.isDirty = true;
+            }
+        }
+
+        /// <summary>
+        /// The result of the current value applied to the response curve
+        /// </summary>
+        public float CurrentResult
+        {
+            get {
+                if (this.isDirty)
+                {
+                    return this.currentResult;
+                }
+                else
+                {
+                    return this.rootParameter.CurrentResult;
+                }
+            }
+        }
+
+        /// <summary>
+        /// Clear the modified value and use the global parameter's value
+        /// </summary>
+        public void Reset()
+        {
+            this.currentValue = this.rootParameter.parameter.CurrentValue;
+            this.currentResult = this.rootParameter.CurrentResult;
+            this.isDirty = false;
+        }
+
+        /// <summary>
+        /// Clear the object from memory
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+
+            GC.SuppressFinalize(this);
+        }
+
+        /// <summary>
+        /// Implement the IDisposable interface
+        /// </summary>
+        /// <param name="disposing"></param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!this.disposed)
+            {
+                if (disposing)
+                {
+                    //No managed resources to dispose of here?
+                }
+
+                this.disposed = true;
+            }
+        }
+    }
+}
--- a/Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
+++ b/Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
 
 namespace Microsoft.MixedReality.Toolkit.Audio
 {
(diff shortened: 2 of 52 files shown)


Assets/MixedRealityToolkit-Examples/Demos/Audio/Scripts/AudioManagerTest.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using Microsoft.MixedReality.Toolkit.Audio;
+
```
StephenHodgson: missing namespace

davidkline-ms: recommend using Microsoft.MixedReality.Toolkit.Examples.Demos

Assets/MixedRealityToolkit/AudioManager/AudioManager.cs
```diff
@@ -0,0 +1,310 @@
...
+
+using UnityEngine;
+using System.Collections.Generic;
+
+namespace Microsoft.MixedReality.Toolkit.Audio
+{
+    /// <summary>
+    /// The manager that handles the playback of AudioEvents
+    /// </summary>
+    public class AudioManager : MonoBehaviour
```
davidkline-ms: Do you anticipate a per-scene audio manager instance or per-app?

Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
```diff
@@ -0,0 +1,121 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.MixedReality.Toolkit.Audio
+{
+    /// <summary>
+    /// The runtime instance of an AudioParameter on an ActiveEvent
+    /// </summary>
+    public class ActiveParameter : IDisposable
```
StephenHodgson: what's the purpose of the disposable here?

Doesn't seem like we're managing any memory or anything.

Assets/MixedRealityToolkit/AudioManager/AudioBank.cs
```diff
@@ -0,0 +1,120 @@
...
+using UnityEditor;
+#endif
+
+namespace Microsoft.MixedReality.Toolkit.Audio
+{
+    /// <summary>
+    /// A collection of AudioEvents and AudioParameters
+    /// </summary>
+    [CreateAssetMenu]
+    public class AudioBank : ScriptableObject
```
StephenHodgson: Nice!

Assets/MixedRealityToolkit/AudioManager/DebugMessage.cs
```diff
@@ -0,0 +1,59 @@
...
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+
+namespace Microsoft.MixedReality.Toolkit.Audio
+{
+    /// <summary>
+    /// Prints a message to Unity's console log
+    /// </summary>
+    public class DebugMessage : AudioNode
```
davidkline-ms:
```suggestion
    public class AudioDebugMessage : AudioNode
```

Assets/MixedRealityToolkit/AudioManager/AudioBank.cs
```diff
@@ -0,0 +1,120 @@
...
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+
+namespace Microsoft.MixedReality.Toolkit.Audio
+{
+    /// <summary>
+    /// A collection of AudioEvents and AudioParameters
+    /// </summary>
+    [CreateAssetMenu]
```
StephenHodgson: Let's add this under the pre-existing MixedRealityToolkit menu under asset/create

Assets/MixedRealityToolkit/AudioManager/AudioBank.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        private void OnEnable()
+        {
+            if (this.audioEvents == null)
+            {
+                this.audioEvents = new List<AudioEvent>();
+            }
+        }
+
+#if UNITY_EDITOR
```
StephenHodgson: Seems like this needs an inspector

Assets/MixedRealityToolkit/AudioManager/AudioEvent.cs
```diff
@@ -0,0 +1,287 @@
...
+
+            for (int i = 0; i < this.nodes.Count; i++)
+            {
+                this.nodes[i].Reset();
+            }
+        }
+
+        #region EDITOR
+
+#if UNITY_EDITOR
```
StephenHodgson: Should also be moved into an editor script

Assets/MixedRealityToolkit/AudioManager/AudioManager.cs
```diff
@@ -0,0 +1,310 @@
...
+            for (int i = 0; i < ActiveEvents.Count; i++)
+            {
+                ActiveEvents[i].ClearSolo();
+            }
+        }
+
+        #endregion
+    }
+
+    public enum AudioLanguage
```
StephenHodgson: please move enums into their own files.

StephenHodgson: This enum could likely be more generic than just for audio?

davidkline-ms: perhaps use the .net culture codes instead (ex: en-US, fr-CA, etc). that would allow for easier globalization (we will need this type of support for text soon as well).

Assets/MixedRealityToolkit/AudioManager/AudioManager.cs
joekellyms: Definitely per app. You don't need it in the scene because it will add itself the first time an audio event is played.

StephenHodgson: If it's not something we need to have in the scene, we may want to make this a system with a configuration profile with audio banks etc

Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
joekellyms: This was an internal suggestion, and relic from the previous audio manager to avoid garbage collection on the ActiveEvents that are constantly being created and destroyed as sounds are played.

Assets/MixedRealityToolkit/AudioManager/AudioBank.cs
joekellyms: AudioBanks are edited in a separate window, so the inspector just provides a button for a shortcut to open the editor window using that bank. Having two interfaces for editing banks worries me, but I'm happy to discuss.

Assets/MixedRealityToolkit/AudioManager/AudioManager.cs
joekellyms: I agree it would be more useful as a generic selection. Would that change where it lives?

Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
StephenHodgson: Interesting.

Assets/MixedRealityToolkit/AudioManager/AudioManager.cs
davidkline-ms: Not sure a separate enum for AudioLanguage would be needed with this suggestion. Instead, code that currently references AudioLanguage would use the .net CultureInfo class.

https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo?view=netframework-4.7.2

While it's a bit more work for the developer to create a CultureInfo, it enables the toolkit to leverage the .net globalization features more seamlessly.

Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
davidkline-ms: >  to avoid garbage collection

Did it make a noticeable performance difference?

Assets/MixedRealityToolkit/AudioManager/AudioManager.cs
joekellyms: I would love for it to not have to be in the scene, but something needs to run the update function for the active events. If there's a better way than adding an object with the AudioManager script to the scene, I'm all for it!

Assets/MixedRealityToolkit/AudioManager/ActiveParameter.cs
joekellyms: Honestly, I never noticed a performance difference, it seemed like a premature optimization, so I'm totally fine with it either way.

davidkline-ms: Let's go without IDisposable. If we find a perf/garbage issue, we can address then :)

Assets/MixedRealityToolkit/AudioManager/AudioManager.cs
davidkline-ms: If you have time, we should sit and discuss. It might make sense to make the audio manager a service and it might make sense to keep it as-is.

StephenHodgson:
@joekellyms I bet you're not able to push changes to that branch since it's locked.

You might want to fork this repository, if you haven't already, and reopen this PR off of your forks branch so you can push change requests.

Assets/MixedRealityToolkit/AudioManager/Editor/AudioGraph.cs
```diff
@@ -0,0 +1,742 @@
...
+
+        /// <summary>
+        /// Display the graph window and automatically open an existing AudioBank
+        /// </summary>
+        /// <param name="bankToLoad"></param>
+        public static void OpenAudioGraph(AudioBank bankToLoad)
+        {
+            AudioGraph graph = GetWindow<AudioGraph>();
+            graph.audioBank = bankToLoad;
+            graph.Show();
```
Jarodshow: `titleContent` is never set here

Assets/MixedRealityToolkit/AudioManager/AudioManager.cs
Railboy: Based on how we've used this manager in our projects a service would make the most sense IMO.

StephenHodgson: Yeah, I think we'll def need to servicily it and likely needs to be part of the core systems.

SimonDarksideJ:
Any update on this PR, generally it all looks good.  just needs servicifying and a few renames generally.
I can help with the servicification (my new word for 2019) if need be

davidkline-ms:
>Any update on this PR, generally it all looks good.

Joe and I are going to sync early the week of 1/14 to finalize the PR. We are planning on making it the first MRTK "extension" so servicification (I like your new word) is likely not 100% needed, though it is definitely on the table

SimonDarksideJ:
I would argue @davidkline-ms that even extensions / addons (especially if they are backed by MS) should follow the architectural patterns to be classified as an official addon.  So we should progress to make this the best it can be and in alignment.

davidkline-ms:
This feature will be provided as an MRTK extension. @joekellyms and @davidkline-ms own moving and finalizing the code.

davidkline-ms:
I will work directly with Joe to create an extension in a separate repository. Closing.

PR #3203, opened 2018-12-02
Some fixes

Jarodshow:
## Changes

- Updated two profile inspector icons

- The `MixedRealityInputActionRulesInspector` had an error where you could see the "Back to Configuration Profile" message even if there was no MixedRealityToolkit in your scene.


Assets/MixedRealityToolkit/_Core/Inspectors/Profiles/MixedRealityInputActionRulesInspector.cs
```diff
@@ -97,11 +97,6 @@ public override void OnInspectorGUI()
                 !MixedRealityToolkit.Instance.ActiveProfile.IsInputSystemEnabled ||
                  MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.InputActionsProfile == null)
             {
-                if (GUILayout.Button("Back to Configuration Profile"))
```
StephenHodgson: This is the correct behavior.

SimonDarksideJ: Agreed, this is how it is used on al the profiles.

Jarodshow:
The initial problem was the first conditional checked that if the MixedRealityToolkit was configured, and if it wasn't, still tried to access the configuration profile.

backto

PR #3208, opened 2018-12-03
Add standard Microsoft Code of Conduct

chbecker-ms:
Add standard Microsoft Code of Conduct.


StephenHodgson:
@davidkline-ms should we put this on dev so that master doesn't get ahead?

davidkline-ms:
> @davidkline-ms should we put this on dev so that master doesn't get ahead? 

I'm fine going right into master and integrating to dev

PR #3210, opened 2018-12-03
Advanced State Sharing System (and related sub-systems)

Railboy:
Advanced state Sharing System (and related sub-systems)
---

Descriptions of SessionManager, UserManager / DevicePool and AnchorSynchronizer / UserAlignment will be expanded as discussion continues.

You can find demos for most major components and systems in MixedRealityToolkit-Examples/Demos/Sharing/ along with a readme explaining what they do.

These components are currently implemented using Photon's PUN2 package . That package is included in this branch along with an app ID for testing.

StateArray
---
The backbone of app state sharing. Essentially a dictionary of structs. Changes made to this dictionary on the server are stored locally until `Flush` is called, at which point the new states propagate to all clients. Changes made on the client are stored locally until they are replaced by a value from the server. 
```
public interface IStateArray<T> : IStateArrayBase, IEnumerable<T> where T : struct, IItemState<T>
{
    T this[sbyte key] { get; set; }
    void Flush();
}
```

State structs must implement the IItemState interface:
```
public interface IItemState<T>
{
    short Key { get; }
    bool IsDifferent(T from);
    T Merge(T clientValue, T serverValue);
}
```

`Key` is used to assign and retrieve the state. (We found that keeping the key attached to states vastly simplifies state sharing in general.)

`IsDifferent `is used to determine when a state has changed. This is preferable to a straight value equality comparison for a couple of reasons:
- Value-type comparisons can generate garbage.
- States often have fields or properties which would result in redundant comparisons (eg a Vector3 property which draws from a set of shorts for its value).
- States often have fields or properties that require interpretive comparison (eg time-related values).

`Merge `is used to resolve conflicts between client and server states. If an incoming state from the server is found to be different than a locally stored state on the client, the two states are merged before the client value is over-written.

**Issues**
- **Server** concept should be changed to **Authority** to make system more peer-to-peer friendly.
- We need a better system for crunching / serializing item states, especially position / rotation states.

AppState
---
Manager class for user-defined StateArrays. Makes the business of getting / setting / adding states to StateArrays straightforward.

Main interfaces / methods:
```
public interface IAppStateReadOnly
{
    T GetState<T>(short stateKey) where T : struct, IItemState<T>;
    void Flush<T>() where T : struct, IItemState<T>;
    void Flush();
    ...
}

public interface IAppStateReadWrite : IAppStateReadOnly
{
    void AddState<T>(T state) where T : struct, IItemState<T>;
    void SetState<T>(T state) where T : struct, IItemState<T>;
    ...
}
```

`Flush` serves a different purpose on client / server. On client, flush sends all locally stored values to the server via the state pipe. On server, Flush just calls Flush on the targeted StateArray(s).

**Initialization**
By default AppState has no StateArrays. On startup AppState automatically seeks out IAppStateSource components, which provide a list of IItemState types for which StateArrays are needed. (Eg `AppStateUsersSource` which requests StateArrays for UserState and UserSlot.)
```
public interface IAppStateSource : INetworkBehaviour
{
    IEnumerable<Type> StateTypes { get; }

    void GenerateRequiredStates(IAppStateReadWrite appState);
}
```

**Issues**
- In IL2CPP builds state array types must be defined ahead of time. (In .NET builds types can be created dynamically.) I'm working on an inspector that detects whether a state array type has been defined to accompany required state types.

StateObjects
---
This system handles a ton of boilerplate for spawning / updating objects associated with a particular item state. Typical use case would be generating and updating Holograms.

Developers create a set of prefabs for each state type requiring an associated in-game object. Each prefab must have a component that inherits from the StateObject base class, which implements the IStateObject interface:
```
public interface IStateObject<T> : IGameObject where t : struct, IItemState<T>
{
    short ItemNum { get; }
    bool IsUserType { get; }
    bool IsLocalUser { get; }

    IAppStateReadOnly AppState { get; }
    IStateView StateView { get; }

    T PreviousState { get; }
    T CurrentState { get; }
}
```
Additional components on this prefab (including the class inheriting from StateObject) can implement the IStateListener interface to receive updates about state changes:
```
public interface IStateListener<T> where T : struct, IItemState<T>
{
    void OnStateInitialize(IStateObject<T> stateObj, T initState);
    void OnStateChange(IStateObject<T> stateObj, T oldState, T newState);
}
```
Lastly, the state object can be used as a conduit for state changes via the ChangeState method:
```
public class HologramStateObject : StateObject<HologramState>, IStateListener<HologramState>

    void OnStateChange(IStateObject<HologramState> stateObj, HologramState oldState, HologramState newState) 
    {
        if (timeToChangePosition)
        {
            newState.Position = transform.position;
            // automatically writes to app state
            ChangeState(newState);
        }
    }
    ...
}
```

SessionManager
---
Provides a skeletal framework for a structured mixed-reality sharing experience using AppState and StateObjects. Runs set of developer-defined 'stages' in a sequence. Example of a session stage progression might be:
1. JoinGame
2. ChooseRole
3. PlayGame
4. ShowHighScores
(Repeat)

Session stages are stored as prefabs and instantiated at runtime. These prefabs must have a component that implements the abstract class SessionStageBase, which has the following methods:
```
public abstract void OnStageEnter (ITimeSource timeSource, IAppStateReadWrite appState,  IUserView users, IStateView stateView);
public abstract void OnStageExit (ITimeSource timeSource, IAppStateReadWrite appState,  IUserView users, IStateView stateView);      
public abstract IEnumerator TransitionIn (ITimeSource timeSource, IAppStateReadWrite appState,  IUserView users, IStateView stateView);
public abstract IEnumerator RunStage (ITimeSource timeSource, IAppStateReadWrite appState,  IUserView users, IStateView stateView);
public abstract IEnumerator TransitionOut (ITimeSource timeSource, IAppStateReadWrite appState,  IUserView users, IStateView stateView);
```
TransitionIn, RunStage and TransitionOut are enumerators to make starting / stopping / pausing trivial.

The main purpose of a session stage is to monitor / modify the app state in response to user actions. This keeps flow of control for the app as a whole out of the hands of individual StateObjects making it easier to track / control.

Since stages are instantiated as prefabs they can be accompanied by objects which are controlled by the stage script (eg a scoreboard). This provides an alternative to the one-state, one-object approach of StateObjects.

**Issues**
- Should probably be renamed to differentiate from network session.
- No support for scene loading / unloading, which is a must IMO.
- No support for branching experiences.

UserManager / DeviceSource
---
System for managing users and devices associated with a user. Uses AppState to manage user slots and user states.

**Issues**
- Lobby system is bare-bones.
- Requires manual redefinition of enum values for UserTypeEnum / UserTeamEnum - would be better if this were templatized.
- Assumes the existence of an AnchorSynchronizer.

AnchorSynchronizer / UserAlignment
---
System for synchronizing locally acquired anchor positions, then using that shared information to align users to the same shared physical space.


Assets/JsonDotNet.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: 9f042f1852f0c5548b3b681cd6086653
+guid: fedd074218b31794b950eee60e347590
```
StephenHodgson: please revert

Assets/JsonDotNet/Assemblies/AOT/Newtonsoft.Json.XML
```diff
@@ -0,0 +1,8015 @@
+<?xml version="1.0"?>
+<doc>
+    <assembly>
+        <name>Newtonsoft.Json</name>
```
StephenHodgson: Is there a way to implement this with the built in Unity JsonUtilities?

Assets/JsonDotNet.meta
StephenHodgson: oh nvm 

Seems like we need to update this branch with the latest dev stuff before we can review.

Assets/JsonDotNet/Assemblies/AOT/Newtonsoft.Json.XML
Railboy: I'm 95% certain there is. I just need to do more testing.

Assets/MixedRealityToolkit-Examples/Demos/Sharing/BasicAppState/AppStateBasic.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Pixie.StateControl;
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Pixie.Demos
```
StephenHodgson: Don't forget to update the namespaces

Railboy: We're going to delay namespace changes until just prior to merge. These scripts live in a separate repo that gets changed frequently in response to internal feedback. In the earlier PR the MRTK namespaces made it tough to maintain parity. (That's partly why the PR was inert for so long.)

Assets/MixedRealityToolkit/_Core/Sharing/AnchorControl/AlignmentState.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        #region raw values
+        public short PositionX;
+        public short PositionY;
+        public short PositionZ;
+
+        public short RotationX;
+        public short RotationY;
+        public short RotationZ;
```
chrisfromwork: These shorts feel like they should be private

Assets/MixedRealityToolkit/_Core/Sharing/AnchorControl/AlignmentState.cs
```diff
@@ -0,0 +1,65 @@
...
+            UserNum = userID;
+            StateUtils.ShortPosIn(Vector3.zero, out PositionX, out PositionY, out PositionZ, MaxPositionRange);
+            StateUtils.ShortRotIn(Vector3.zero, out RotationX, out RotationY, out RotationZ);
+        }
+
+        short IItemState.Key { get { return UserNum; } }
+
+        public short UserNum;
+
+        public Vector3 Position
```
chrisfromwork: Comments on what these positions and angles are relative to would likely prove helpful.

Assets/MixedRealityToolkit/_Core/Sharing/AnchorControl/AnchorSynchronizer.cs
```diff
@@ -0,0 +1,236 @@
...
+
+                default:
+                    throw new Exception("Can't do this when state is " + state);
+            }
+
+            state = AnchorSyncStateEnum.CreatingSharedAnchors;
+            StartCoroutine(CreateAnchorStatesInternal(definitions, appState));
+        }
+
+        private IEnumerator CreateAnchorStatesInternal(IAnchorDefinitions definitions, IAppStateReadWrite appState)
```
chrisfromwork: Do we need to dynamically monitor changes for AppState UserSlots? This function looks like it is only intended to be called during initialization, but i feel like we may want to create/destroy user anchors as users leave and enter the stage/session?

Assets/MixedRealityToolkit/_Core/Sharing/AnchorControl/AnchorSynchronizer.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        public AnchorSyncStateEnum State { get { return state; } }
+        
+        [SerializeField]
+        private AnchorSyncStateEnum state = AnchorSyncStateEnum.Stopped;
+        private Dictionary<short, Transform> sharedAnchorTransformLookup = new Dictionary<short, Transform>();
+        private Dictionary<short, Transform> userAnchorTransformLookup = new Dictionary<short, Transform>();
+        private List<Transform> sharedAnchorTransformList = new List<Transform>();
+        private List<Transform> userAnchorTransformList = new List<Transform>();
+        private Transform sceneRootTransform;
```
chrisfromwork: Is there an observable field in this anchor synchronizer that is intended to be a transform to the shared experience's origin?

Assets/MixedRealityToolkit/_Core/Sharing/AppSystems/Sessions/SessionStageBase.cs
```diff
@@ -0,0 +1,179 @@
...
+        private IAppStateReadWrite appState;
+
+        private bool forceMoveNext = false;
+
+        public void ForceComplete()
+        {
+            forceMoveNext = true;
+        }
+
+        public abstract void OnStageEnter(ITimeSource timeSource, IAppStateReadWrite appState);
```
chrisfromwork: What's the expected behavior for these OnStageEnter and OnStageExit calls? Above, they aren't implemented. Is the intent of these methods fulfilled with TransitionIn & TransitionOut?

Assets/MixedRealityToolkit/_Core/Sharing/StateControl/Logging/AppStateLogger.cs
```diff
@@ -0,0 +1,127 @@
+﻿using Pixie.AppSystems.TimeSync;
+using Pixie.Core;
+using Pixie.Initialization;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using UnityEngine;
+using Newtonsoft.Json;
```
chrisfromwork: Is netwonsoft a dependency already existing for the mrtk? If not, are there instructions for how to pick up this dependency/should we start taking this on as a dependency?

Assets/MixedRealityToolkit/_Core/Sharing/AppSystems/Sessions/SessionStageBase.cs
Railboy: OnStageEnter / Exit are intended for instantaneous transitions. Eg in an arcade shooter this might switch the behavior mode of all the enemies in the level.

TransitionIn and TransitionOut are intended for cases where you either want a smooth transition from one state to another, or when you don't want the stage to actually begin until some condition is met. Eg, waiting for all joined users to set their state to 'Ready' before a meeting starts.

Assets/MixedRealityToolkit/_Core/Sharing/AnchorControl/AlignmentState.cs
Railboy: These fields are being serialized so they need to stay public. This is one reason I don't like this method of crunching down position / rotation data. If you have suggestions for alternatives let me know.

Assets/MixedRealityToolkit/_Core/Sharing/Core/IItemState.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace Pixie.Core
+{
+    public interface IItemState
+    {
+        /// <summary>
+        /// The key for the item
+        /// </summary>
+        short Key { get; }
```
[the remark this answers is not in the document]
Railboy: I figure if we're syncing more than 32,767 synchronized states for a single type then we probably shouldn't be using this system anyway.

A unsigned short couldn't hurt though. There are 2-3 cases where I use a negative value to detect invalid states but we could refactor those out.

Assets/MixedRealityToolkit/_Core/Sharing/AnchorControl/AnchorSynchronizer.cs
Railboy: No, the transforms I'm using here are strictly helpers. The system exposes alignment info via state data exclusively. You could apply that data to a scene origin transform pretty easily.

The UserAlignment script applies the data to a transform holding the user's camera. That setup assumes you have a static scene (eg lightmapped) and want to align the users to the scene, not the other way around.

chrisfromwork: Are there any known scenarios where we would have a non-static scene? Is there any thought on how content anchored to other users could be updated with this system?

Assets/MixedRealityToolkit/_Core/Sharing/StateControl/Logging/AppStateLogger.cs
chrisfromwork: sounds good

chrisfromwork:
It's probably worth calling out to the working group that this review introduces a first approach to anchor sharing. I'm not sure how developed the thinking is around anchor sharing, but as it matures the logic here will likely need to be refactored/rewritten.

Assets/MixedRealityToolkit/_Core/Sharing/Core/StateArray.cs
```diff
@@ -0,0 +1,395 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Pixie.Core
+{
+    // We use this class to get / set items in our sync list state in a standard way
+    public class StateArray<T> : IStateArray<T> where T : struct, IItemState, IItemStateComparer<T>
```
chrisfromwork: Requiring T to be a struct seems limiting. Whats the thinking around requiring it to be a struct?

Assets/MixedRealityToolkit/_Core/Sharing/StateControl/Logging/AppStateLogger.cs
SimonDarksideJ:
```
using Newtonsoft.Json;
```
Noooooooo. Should be using the Unity JSON NOT Newtonsoft.  This should be avoided at ALL costs

Assets/MixedRealityToolkit/_Core/Sharing/Core/StateArray.cs
Railboy: Yes this is a deliberate limitation. The goal is to establish a central authority for state data. If you get a state directly from AppState then it's guaranteed to be an up-to-date copy. If you get it from anywhere else it may be stale / modified. And if you want to change a state you must do it through AppState for anyone else to be aware of those changes.

This all cuts down on the state 'spaghetti' you can get when objects are responsible for maintaining their own state data. It makes everything much easier to track and debug.

As a bonus it makes compression easier by encouraging the complete decoupling of state data from classes that use it.

Worth noting that these concerns don't come up in smaller apps. They can handle a bit of state spaghetti without an issue.

chrisfromwork: Is there a good example in this review of a preexisting class that populates a state array?

Assets/MixedRealityToolkit/_Core/Sharing/AppSystems/Sessions/SessionStageBase.cs
chrisfromwork: can we get an overview/timeline for all of these events added to the description?

Assets/JsonDotNet/Assemblies/Standalone/Newtonsoft.Json.dll.meta
```diff
@@ -0,0 +1,75 @@
+fileFormatVersion: 2
```
SimonDarksideJ: We should NOT be using Newtonsoft due to it's compatibility issues on platforms.
This should be replaces / updated to use the Unity JSON implementation
https://docs.unity3d.com/Manual/JSONSerialization.html

Railboy: @StephenHodgson and @chrisfromwork have brought this up and I've agreed that swapping it out is best. Once we have time to stress test it with state data it'll be removed.

Railboy:
I'm going to shut this down and re-open with bare-bones sharing. It's clear there's too much to chew on for one PR.

PR #3227, opened 2018-12-05
Fixing incorrect alpha channel values for channel maps and alpha tested (cutout) materials.

Cameron-Micka:
Materials using channel maps or alpha tested (cutout) materials could report incorrect alpha values. This is not something apparent in usual usage, but could present problems during Mixed Reality or BEV capture.


davidkline-ms:
Thank you, @Cameron-Micka! Love your shader work.

Cameron-Micka:
Thank you @davidkline-ms! It's great to be part of the MRTK.

Issue #3218, opened 2018-12-04
[Beta] vNext Gaze does not work with Unity UI on device in Hololens


phosphoer:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit vNext

**Describe the bug**
While gaze input seems to be working in the editor, I cannot get it working on device (Hololens). I have a CanvasUtility on my UI canvas, and when I build for Hololens and run on device I do not get any highlight effect when gazing at a button, nor am I able to press it with air tap. 

**To Reproduce**
[Repro project](https://github.com/phosphoer/repro-mrtk-ui-device)
1. Download the repro project and open in Unity
2. Run the Scenes/mrtk-unity-ui scene in editor
3. Drag the camera to the left in the scene view and verify that the button is highlighted by the gaze cursor
4. Build for UWP and deploy to a Hololens
5. Run the app on device and observe that the button is not highlighted by gaze 

**Expected behavior**
The UI button is highlighted by gaze on device

**Actual behavior**
The UI button is not affected by gaze on device

**Unity Editor Version**
2018.2.14f1

**Mixed Reality Toolkit Release Version**
vNext Beta

StephenHodgson:
Can you also verify that the canvas utility was on the UI you were trying to focus on?

StephenHodgson:
Oh this isn't really a bug as it's not using the beta focus provider.

Seems this project is using the refactored one in the PR that hasn't been merged yet.

I'll move over the test scenes into the PR and make sure these are addressed.

phosphoer:
> Can you also verify that the canvas utility was on the UI you were trying to focus on?

I mentioned this was the case in the bug description. 

> Oh this isn't really a bug as it's not using the beta focus provider.
> Seems this project is using the refactored one in the PR that hasn't been merged yet.

Yes I was using the one from the pull request you mentioned contained fixes, in hopes this was one of them. The bug is also present in the latest beta release. 

Is there perhaps something easy I could be missing? Is there some platform-specific settings in one of the configuration profiles that might be excluding something I need in the build?

StephenHodgson:
> Is there perhaps something easy I could be missing? Is there some platform-specific settings in one of the configuration profiles that might be excluding something I need in the build?

I don't think so, but I'm def interested in getting it fixed asap

StephenHodgson:
[...]

PR #3230, opened 2018-12-05
Fix UI raycasting on Hololens

phosphoer:
Overview
---
This change addresses the issue described in #3218. Unity UI is raycasted into using a special disabled camera whose projection matrix informs the raycast. The camera was being created with a target display index of 1, or the second display, which does not exist on Hololens. This seems to be ignored in editor when there is no other display, but on Hololens it causes all UI raycasts to return no results, likely due to an invalid projection matrix for a non-existent display.  

There were a bunch of other properties on the camera being set unnecessarily, which I also removed in the interest of cleanliness. 

Changes
---
- Fixes: #3218

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
@@ -333,24 +333,14 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
             cameraObject.transform.parent = transform;
             cameraObject.transform.localPosition = Vector3.zero;
             cameraObject.transform.localRotation = Quaternion.identity;
+
+            // The raycast camera is used to raycast into the UI scene, it doesn't need to render 
+            // anything so is disabled, however its projection matrix will be used to calculate raycast info 
+            // so it is helpful to have a very close nearclip plane
             uiRaycastCamera = cameraObject.AddComponent<Camera>();
             uiRaycastCamera.enabled = false;
-            uiRaycastCamera.clearFlags = CameraClearFlags.Depth;
-            uiRaycastCamera.cullingMask = CameraCache.Main.cullingMask;
-            uiRaycastCamera.orthographic = true;
-            uiRaycastCamera.orthographicSize = 0.5f;
-            uiRaycastCamera.nearClipPlane = 0.1f;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -333,24 +333,14 @@ private void CreateUiRaycastCamera()
...
-            uiRaycastCamera.renderingPath = RenderingPath.UsePlayerSettings;
-            uiRaycastCamera.targetTexture = null;
-            uiRaycastCamera.useOcclusionCulling = false;
-            uiRaycastCamera.allowHDR = false;
-            uiRaycastCamera.allowMSAA = false;
-            uiRaycastCamera.allowDynamicResolution = false;
-            uiRaycastCamera.targetDisplay = 1;
-            uiRaycastCamera.stereoTargetEye = StereoTargetEyeMask.Both;
+            uiRaycastCamera.nearClipPlane = 0.01f;
+            uiRaycastCamera.farClipPlane = 100f;
```
StephenHodgson: Wondering if this should match the global pointer extents

phosphoer: Other than ensuring the raycast begins at a close enough distance to the camera that it doesn’t miss nearby UI, I don’t believe it much matters what these values are in terms of ray casting, particularly farclip

phosphoer: Will do some more testing tomorrow to find out whether this is the case

StephenHodgson: Not worried about the near clip, just the far clip

phosphoer: Did some more testing, the near and far clip actually have no effect on the raycasting that I can tell. I'll just remove the code that sets them to avoid confusing future readers.

phosphoer: Oh, I just saw that it was merged already haha, I'll make a very small PR

StephenHodgson: Sounds good

Issue #3233, opened 2018-12-06
vNEXT Proposal: Add profile to System constructors


StephenHodgson:
# Overview
Since each profile is required by most systems (The exception being the teleport system, but they may change in the future). Systems should setup their initial values using the profiles provided in the constructor.

This better implements the Dependency injection patterns we've been utilizing and adding robustness to the systems. All initial variable values will be set in the constructor.

# Requirements

* BaseSystem class with the profile constructor parameter.
* Update to the MixedRealityToolkit to register a system with the profile constructor arg.
* Updates to all the systems to set up their initial conditions based off of the provided profile.

# Acceptance Criteria

- [x] I can register a system with its profile
- [ ] All systems have been updated to initialize their initial values using the provided profile.

davidkline-ms:
Moving to v3. This will be a breaking change for custom system implementations. App code is not expected to be impacted.

davidkline-ms:
complete

PR #3234, opened 2018-12-06
Implementation of injecting profile into system constructors

StephenHodgson:
Overview
---
Implements the proposed change of adding the profile to the system constructors.

Changes
---
- Fixes: #3233 

Breaking Changes
---
- Added surface properties to `IMixedRealityBoundarySystem` to facilitate conformity and initialization of properties in the constructor.
- Systems shouldn't read from the profile after initializing their properties in their constructor.


StephenHodgson:
Fixed

StephenHodgson:
Had to make the boundary system changes to make it all work.

Not sure a second PR is really necessary but I suppose I can pull it out. I'd still have to change the constructor assignments.

Issue #3245, opened 2018-12-08
Diagnostics System crashes HoloLens application on start


FejZa:
## Overview
When using the latest mrtk_development state ov vNext having the Diagnostics System enabled and visible will crash the application on start when deployed to HoloLens.

## Expected Behavior
Diagnostics System is enabled and probably should be visible in the top left corner of the HoloLens FOV

## Actual Behavior
Application crashes on start right after Made With Untiy splashscreen.

## Steps to reproduce
- Import the latest vNext state
- Target UWP
- Configure Diagnostics System enabled and visible
- Configure the scene for MRTK
- Build and deploy to HoloLens

## Unity Editor Version
2018.3.0b12

## Mixed Reality Toolkit Release Version
mrtk_development branch

StephenHodgson:
Do you have a stack trace of the crash?

FejZa:
>	GameAssembly.dll!il2cpp::icalls::System::Microsoft::Win32::NativeMethods::GetProcessTimes(int handle, __int64 * creation, __int64 * exit, __int64 * kernel, __int64 * user) Line 30	C++

 	GameAssembly.dll!NativeMethods_GetProcessTimes_mFF1FA0FF468CE0698346FB025408FE4661F5F779(int ___handle0, __int64 * ___creation1, __int64 * ___exit2, __int64 * ___kernel3, __int64 * ___user4, const MethodInfo * method) Line 15064	C++

 	GameAssembly.dll!NativeMethods_GetProcessTimes_mCAB6B28520672C332F9250F565B1178788F8BF64(SafeProcessHandle_tEF75BF77F5F4E121334E2A97EE8E1F6685F38CF7 * ___handle0, __int64 * ___creation1, __int64 * ___exit2, __int64 * ___kernel3, __int64 * ___user4, const MethodInfo * method) Line 15094	C++

 	GameAssembly.dll!Process_GetProcessTimes_mAEE0C477ECD21E85F86EA92ED23C46A1007C5705(Process_t7F28AE318E6CCF89716D05E78E2CBD669767D6A1 * __this, const MethodInfo * method) Line 29835	C++

 	GameAssembly.dll!Process_get_TotalProcessorTime_mE9B83B34CBD1CA8217CE045CA1457D327BC5833A(Process_t7F28AE318E6CCF89716D05E78E2CBD669767D6A1 * __this, const MethodInfo * method) Line 29949	C++

[...]

PR #3247, opened 2018-12-09
Diagnostics Fixes

StephenHodgson:
Overview
---
Fixes exception on UWP platform when trying to get cpu use

Changes
---
- Fixes: #3245


StephenHodgson:
@FejZa does this fix the issue for you?

FejZa:
@StephenHodgson I can confirm this fixed the issue. The app launches fine now and doesn't crash. Can't see any diagnostics though. But I saw that is already being discussed in another issue.

StephenHodgson:
The current implementation doesn't show anything in the scene yet.

Issue #3158, opened 2018-11-16
Improve Workflow for Stabilization Plane


jbienzms:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Is your feature request related to a problem? Please describe.**
As stated in the documentation [here](https://github.com/MicrosoftDocs/mixed-reality/blob/master/mixed-reality-docs/focus-point-in-unity.md):

> On a HoloLens running the Windows 10 April 2018 Update or later, (turning on "Enable Depth Buffer Sharing") will analyze the depth buffer to pick an optimal stabilization plane automatically.

That article incorrectly states that it's turned on by default now, but I see #1939 proposed to address this issue. (And I support that change.) 

Regardless of whether or not #1939 gets implemented, we need to provide better guidance about what to do when **Depth Buffer Sharing** is turned on. In fact, what _should_ be done to the MRTK when this is enabled? Should `StabilizationPlaneModifier` be disabled entirely? Should it's `SetStabilizationPlane` field be set to false? And what changes, if any, should be made to the `InputManager` prefab?

At the very least, I feel something should be added to the **Apply Mixed Reality Project Settings** screen that describes this option. Ideally there would be something that prompts the user to see if all of their devices will be running RS4 or later and changes the settings accordingly.

StephenHodgson:
@jbienzms We don't have the same settings configuration screen in vNEXT, but we can add it to the prompt when importing the first time.

It's a bit strange to grab the SDK and set this property on it, but it is possible to set.

Troy-Ferrell:
Also FYI #5319

Issue #1939, opened 2018-04-11
Enable Shared Depth Buffer in ProjectSettingsWindow (V2)


Jarodshow:
## Overview
Maybe we should enable depth buffer sharing by default with `Enable XR`? Obviously only supported in 2017.3.1f1 and later.

StephenHodgson:
Possible duplicate.

The shared depth buffer only works with the immersive headsets and the RS 4 OS on the HoloLens, so this a bit tricky to implement without breaking compatibility for those that are still using the older os on the HoloLens.

Jarodshow:
Perhaps it could be a prompt afterwards noting that it only works on RS4 and IHMD?

Troy-Ferrell:
Unity, starting around roughly Unity 2018 LTS, has by default for new projects Depth Buffer sharing enabled. The problem with an automatic "Apply scene settings" or simple prompt is that some developers do not like MRTK being that intrusive especially under the covers in their project. The PR #5199 is an attempt to provide a tool to make these configurations easy to set but also informative

Issue #3183, opened 2018-11-22
Linux standalone sharedassets are corrupted


wassx:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
Building a minimal scene for linux standalone with build target Linux64.
In the scene (and project) there is just a mixedrealityplayspace with a camera (tagged) and the mrtk.
Everything compiles and looks fine.
When starting the standalone app (in my case done in a docker container) the Player.log is filled with
```
Initialize engine version: 2018.2.16f1 (39a4ac3d51f6)
Forcing GfxDevice: Null
GfxDevice: creating device client; threaded=0
NullGfxDevice:
    Version:  NULL 1.0 [1.0]
    Renderer: Null Device
    Vendor:   Unity Technologies
FMOD initialized on nosound output
Begin MonoManager ReloadAssembly
- Completed reload, in  0.038 seconds
The file '/files/holox_Data/sharedassets0.assets' is corrupted! Remove it and launch unity again!
[Position out of bounds!]

(Filename:  Line: 221)
```

This error messages spams the log.
Cleaning the build folder, library,... did not work.
**This error occurs every 95% of builds/launches.**

Please see attached project.

**Unity Editor Version**
2018.2.16f1

**Mixed Reality Toolkit Release Version**
mrtk_development 
https://github.com/Microsoft/MixedRealityToolkit-Unity/tree/49f5c4c22b66a7e2c212f15b982056dcb31ff4cd
22.11.2018

StephenHodgson:
Thanks for the report.

Could you try going back a few commits and see when this was introduced?

Have you also been able to reproduce this for other platforms like windows or osx?

wassx:
I tried with a version shortly after vnext_beta, same issue. Can only reproduce with linux64. Will try to step back and find latest working version.

StephenHodgson:
Much appreciated.

wassx:
Oh man that was quite some digging. Dating back to 27.10.2018
PR https://github.com/Microsoft/MixedRealityToolkit-Unity/commit/7eca8c5e668f4abc2162e28bfdfed051fd2a6cb6 still works.
The issue occurs with: https://github.com/Microsoft/MixedRealityToolkit-Unity/commit/5d2f3064448941de761fa5fd3c326e8cdd1cd98f
[...]

PR #3248, opened 2018-12-09
Updated editor setup scripts

StephenHodgson:
Overview
---
- Removed some unused methods and cleaned up the class a bit.
- Enables shared depth buffer for Windows Mixed Reality in the project by default.

Changes
---
- Fixes: #3158
- Fixes: #1939
- [x] Fixes: #3259

Blocked by
- [ ] #3269


StephenHodgson:
@wassx I didn't intend to try and fix #3183 with this change, but can you test it for me?

Assets/MixedRealityToolkit/_Core/Utilities/Editor/Setup/.MixedRealityEditorSettings.cs
```diff
@@ -169,49 +176,18 @@ private static bool FindDirectory(string directoryPathToSearch, string directory
...
-            }
-
-            AssetDatabase.SaveAssets();
-            AssetDatabase.Refresh(ImportAssetOptions.ForceUpdate);
-        }
+        #region IActiveBuildTargetChanged Implementation
 
         /// <inheritdoc />
-        public int callbackOrder { get; private set; }
+        public int callbackOrder => 0;
```
StephenHodgson:
```suggestion
        public int callbackOrder { get { return 0; } }
```

alcooper91:
Hi Stephen, I won’t be able to test this. 
@davidkline-ms could you or Kurtis test this or have someone follow up?

wassx:
I'll test today. Thx!

wassx:
Sry, the issue is still there: _'sharedassets0.assets' is corrupted!_

wassx:
Yep, deleted Library and my build directories, so unity reloaded everything

StephenHodgson:
> StabilizationPlaneModifier 

We don't have that in vNEXT

jbienzms:
Hey @keveleigh do you know anyone who can tell us exactly what the right thing is to do when Shared Depth Buffer is enabled for RS4? Should we disable  `StabilizationPlaneModifier` or set it's `SetStabilizationPlane` field to false? Or maybe something else?

jbienzms:
> > StabilizationPlaneModifier
> 
> We don't have that in vNEXT

How does the stabilization plane get set in vNEXT? Is RS4 or later required for vNEXT?

StephenHodgson:
> RS4 or later required for vNEXT

That is correct. We designed it with the assumption that the stabilization plane will not be modified by the app.

jbienzms:
Okay. I can approve the current fix assuming that only RS4+ will be used.

StephenHodgson:
Although I think you're right Jarrod, we should call out that we're setting the stabilization plane in the dialog.

Let me make that change before you approve.

StephenHodgson:
Done :)

Issue #3253, opened 2018-12-10
Some controllers don't send input events on each frame if input is active.


StephenHodgson:
## Overview
As I was trying to implement an example demo, I realized that input events didn't seem to be sent consistently. 

Looking into a few of the controller classes I realized that we only send input events if the data has changed, but that doesn't mean we always send events for things like continuous button presses or joystick data.

## Expected Behavior
Input events should be raised whenever data is availible and not at the considered "default" state. For example, a Joystick is pushed forward to 1, 0 should raise an event every frame saying that's what the current position of the joystick is from 0, 0 through 1, 0 and continue to raise events while the position is at 1, 0.

## Actual Behavior
Input events are only raised when the data has changed, so we only ever get a single event that the joystick position changed from 0 through 1, but once it's at 1, 0 all events stop.

This would sometimes cause things to "Spring Back" to their initial position.

luis-valverde-ms:
Reopening this as it is still an issue when for example trying to drive a rotation using a thumbstick. Once the thumbstick is pressed all the way its value won't change, events won't be raised and the rotation will stop.

PR #3254, opened 2018-12-11
Fixed input events

StephenHodgson:
Overview
---
Input events weren't being continuously raised for some event types.

Changes
---
- Fixes: #3253

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Definitions/Devices/MixedRealityInteractionMapping.cs
+++ b/Assets/MixedRealityToolkit/_Core/Definitions/Devices/MixedRealityInteractionMapping.cs
@@ -263,6 +263,30 @@ namespace Microsoft.MixedReality.Toolkit.Core.Definitions.Devices
             }
         }
 
+        private bool updated;
+
+        /// <summary>
+        /// Has the value been updated since the last reading?
+        /// </summary>
+        public bool Updated
+        {
+            get
+            {
+                bool returnValue = updated;
+
+                if (updated)
+                {
+                    updated = false;
+                }
+
+                return returnValue;
+            }
+            private set
+            {
+                updated = value;
+            }
+        }
+
         #endregion Interaction Properties
 
         #region Definition Data Items
@@ -304,6 +328,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Definitions.Devices
                 }
 
                 Changed = rawData != value;
+                // use the internal reading for changed so we don't reset it.
+                Updated = changed || value != null;
                 rawData = value;
             }
         }
@@ -327,6 +353,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Definitions.Devices
                 }
 
                 Changed = boolData != value;
+                // use the internal reading for changed so we don't reset it.
+                Updated = changed || value;
                 boolData = value;
             }
         }
@@ -357,6 +385,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Definitions.Devices
                 else
                 {
                     Changed = !floatData.Equals(value);
+                    // use the internal reading for changed so we don't reset it.
+                    Updated = changed || !floatData.Equals(0f);
                     floatData = value;
                 }
             }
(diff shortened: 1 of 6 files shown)


Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
```diff
@@ -367,7 +384,8 @@ private void UpdateThumbStickData(InteractionSourceState interactionSourceState,
                         interactionMapping.Vector2Data = interactionSourceState.thumbstickPosition;
 
                         // If our value changed raise it.
-                        if (interactionMapping.Changed)
+                        if (interactionMapping.Changed ||
+                           !interactionMapping.Vector2Data.Equals(Vector2.zero))
```
StephenHodgson: I'm half tempted to move this into the interactions changed check instead of adding it everywhere in each of our implementations. @SimonDarksideJ thoughts?

StephenHodgson: Nah that doesn't make sense to me. Then "Changed" isn't really what  it is

SimonDarksideJ: I'm uneasy about this line testing vector2.zero,  As it would mean a value that changed from say 0.1 to 0 wouldn't invoke a changed state, which is has.
We shouldn't ignore 0 as a changed state, I can see where you were going (ignoring if the value remains at 0, but the changed check already does this.

StephenHodgson:
I'll write up some unit tests to validate your concerns and make any needed changes @SimonDarksideJ

Assets/MixedRealityToolkit/_Core/Devices/MixedReality/WindowsMixedRealityController.cs
keveleigh: The change from 0.1 to 0 would be caught be the first half of the check, because `interactionMapping.Changed` will be true.

StephenHodgson: That was my thought as well.

It's no worries, that's what unit tests are for! woot!

StephenHodgson:
So the unit tests pass with the concern you brought up, but I don't feel comfortable updating what `Changed` means. I'd like to wrap this up into maybe another bool where we can ask if they value has been updated (would return true, even if the value is the same)

Thoughts?

Assets/MixedRealityToolkit/_Core/Definitions/Devices/MixedRealityInteractionMapping.cs
```diff
@@ -263,6 +263,30 @@ private set
             }
         }
 
+        private bool updated;
+
+        /// <summary>
+        /// Has the value been updated since the last reading?
```
keveleigh: This summary (and the name `Updated`) doesn't seem accurate.

All the places that set this check for "changed or not the default value", which doesn't guarantee that the value has been updated since the last reading. It might be the same reading but non-default.

Assets/MixedRealityToolkit/_Core/Definitions/Devices/MixedRealityInteractionMapping.cs
```diff
@@ -327,6 +353,8 @@ public bool BoolData
                 }
 
                 Changed = boolData != value;
+                // use the internal reading for changed so we don't reset it.
+                Updated = changed || value;
```
keveleigh: Firing `Pressed` events based on this logic might lead to confusing behavior.

For example, if I release the trigger, `changed` will be true but `value` will be false, and the system will fire an `InputUp` followed by an `InputPressed` (since `Updated` will be true).

It seems like we should only be firing `Pressed` events whenever the value isn't the default, instead of also whenever it changes.

StephenHodgson: Makes sense. What changes do you think would need to happen to address this?

keveleigh: I think just removing `changed ||` from each line fixes it? So then `Updated` (or whatever it gets called) will be true when the input is active / value isn't the default, and false otherwise. Then, `Changed` tells us when the new value != the old value, and `Updated` tells us when the action is "pressed" for the purposes of the Pressed events.

StephenHodgson: Could you open a PR on my branch? I'm not sure I fully understand the changes you're requesting.

Thanks :)

PR #3268, opened 2018-12-13
MRTK_DefaultPointerLine material is behind ui.

wassx:
Overview
Is there a reason why the pointer line has different material than the cursor?

shader
shader2

Changes
Changed to same shader as used by the default cursor material (with unity 2018.2.17f1)


keveleigh:
@wassx Unfortunately, it looks like the MRTK Standard Shader isn't compatible with Unity's LineRenderers, since the pointer gradient and colors are no longer showing up after this PR:

| Expected  | Actual|
| ------------- | ------------- |
| image  | image  |

I may have to revert this change, but I'm looking into a solution for you.

Cameron-Micka:
@keveleigh This PR https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3382 should fix the line color issue! All you need to do is enabled "Vertex Colors" on the material.

Issue #3192, opened 2018-11-28
UWP apps built via command line (without opening Unity Editor) after a clean do not launch


alcooper91:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
When building a project which uses MRTK from the command line after a clean of any Unity Library files, a built UWP does not launch.  We seem to be getting out of bounds serialization errors from levels/shared assets.  An empty scene seemed to be able to launch, but once I begin adding MRTK dependencies, I get issues.

I have tracked the issue down, and it appears to be related to the Microsoft.MixedReality.Toolkit.Core.Utilities.Editor.Setup.dll.  When I delete this dll and remove the . from in front of MixedRealityEditorSettings.cs, I am able to clean, and build from the command line and successfully launch the app.

Jarodshow:
#3183 Hmm...

StephenHodgson:
I wonder what about this dll is breaking builds 🤔

alcooper91:
I believe there’s something with serialization if the Editor is never opened

davidkline-ms:
Stack trace:

UnityPlayer.dll!DebugStringToFilePostprocessedStacktrace(struct DebugStringToFileData const &) Unknown
UnityPlayer.dll!DebugStringToFile(struct DebugStringToFileData const &) Unknown
UnityPlayer.dll!DebugStringToFile<class core::basic_string<char,class core::StringStorageDefault > >(class core::basic_string<char,class core::StringStorageDefault > const &,int,char const ,int,enum LogMessageFlags,int,int,void ()(struct CppLogEntry const &)) Unknown
UnityPlayer.dll!CachedReader::OutOfBoundsError(unsigned int,unsigned int) Unknown
UnityPlayer.dll!CachedReader::UpdateReadCache(void *,unsigned int) Unknown
UnityPlayer.dll!CachedReader::Read(unsigned char &) Unknown
UnityPlayer.dll!StreamedBinaryRead<0>::TransferBasicData(unsigned char &) Unknown
UnityPlayer.dll!StreamedBinaryRead<0>::Transfer(unsigned char &,char const *,enum TransferMetaFlags) Unknown
[...]

PR #3269, opened 2018-12-13
Editor update 2018.3.0f2

StephenHodgson:
Overview
---
Updates the [project's editor to 2018.3.02f](https://unity3d.com/unity/whats-new/unity-2018.3.0)

Key changes:
---
- .net 4.x equivalent is the default editor scripting runtime, and we no longer need to have the embedded dll for updating it.
- Updated prefabs will not be backwards compatible with prev versions of the editor.

Changes
---
- Fixes: #3192 
- Fixes: #3183


StephenHodgson:
Ready for review.

All tests pass.

UWP and Standalone builds pass and run.

StephenHodgson:
> Does this change remove the need for #3248?

... no?

davidkline-ms:
I'm hitting a CS0019 in SystemType when trying to build the project (UWP target, .net backend) that I do not recall from 2018.2.  Investigating now.

davidkline-ms:
Assets\MixedRealityToolkit\_Core\Definitions\Utilities\SystemType.cs(46,21): error CS0019: Operator '&&' cannot be applied to operands of type 'bool' and 'method group'

```
                if (Type != null && Type.IsAbstract)
```

davidkline-ms:
Feels like we have seen this before.... trying to find the related issue / pr

StephenHodgson:
> UWP target, .net backend

Why are you even trying that? It's marked deprecated and will be removed very soon.

davidkline-ms:
> Why are you even trying that? It's marked deprecated and will be removed very soon.

Because it makes debugging easier, and unity has committed to supporting the .net backend in 2018's LTS for the next 2 years.

StephenHodgson:
Makes no difference. We shouldn't be supporting it anyway.

davidkline-ms:
To best support the maximum number of customers, we should support the maximum functionality provided by Unity.

StephenHodgson:
> maximum functionality provided by Unity

You mean when I opened a bug report on something for .net a few months back and they told me to use IL2CPP cause they weren't committed to fixing it?

StephenHodgson:
To best support our customers we shouldn't be using deprecated features.

keveleigh:
>To best support our customers we shouldn't be using deprecated features.

We aren't forcing customers to use deprecated features, but are providing them the option to if they choose to.

Our publicly stated support level has been that we will enable building the Toolkit with the .NET backend as long as it ships with a version of Unity we support. If Unity has bugs, we can't control that, but we can give our customers the choice.

StephenHodgson:
> Our publicly stated support level has been that we will enable building the Toolkit with the .NET backend

I didn't agree to that. I guess we'll agree to disagree here.

In either case, I don't understand why a deprecated feature should block a PR.

davidkline-ms:
> I don't understand why a deprecated feature should block a PR.

Trying to understand the source of the issue as very recently I was able to build using the .net backend on 2018.2.

keveleigh:
>In either case, I don't understand why a deprecated feature should block a PR.

It was being discussed here because it was discovered while testing this PR, but I'm pretty sure it's not a blocker here if it isn't 2018.3-specific.

davidkline-ms:
We had a fix for a very similar issue in June (2018). I am more than happy to (re)apply the same logic.

https://github.com/Microsoft/MixedRealityToolkit-Unity/commit/cb4feb50de5af307b30ad253aba37f8657ad2a57#diff-9be24a654529dda945cca11020cecab8

StephenHodgson:
Sounds good. likely do it as a separate PR.

SimonDarksideJ:
Interesting, why would we get a missing package for Oculus when we don't support it yet.  Best to ensure oculus isn't in the supported SDK's rather than include the package ref

davidkline-ms:
#3273 is tracking the net backend compile issue. PR forthcoming.

davidkline-ms:
Just to be super clear, if we take this change, we are breaking support for previous Unity releases (many package manager errors on load, MethodAccess exceptions, etc). 

Are we comfortable requiring everyone to move to 2018.3?

My intuition says it is the right thing to do as we are currently planning on requiring Unity 2018.4 upon release. 

Asking this question here to gather community feedback prior to the change.

wassx:
We are testing it on our side. Due to the fix of #3183 it looks promising.

StephenHodgson:
Always forwards, never backwards.

We're still in beta and we all agreed to update until we made our first release.

StephenHodgson:
The player settings quick link is broken in the build window for 2018.3 as well.

image

Tried looking though the [CS reference](https://github.com/Unity-Technologies/UnityCsReference/search?q=ExecuteMenuItem&unscoped_q=ExecuteMenuItem) to see what they changed it to, but apparently it's completely gone now?

PR #3272, opened 2018-12-14
Simple state sharing

Railboy:
Simple state sharing service
---
Baseline for a simple, flexible state sharing service.

On a continuum between POC apps and robust, enterprise-ready apps this service should fall somewhere in the middle. It doesn't try to guide you down an 'optimal' path like [advanced state sharing](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3210), but it's not an automatic drag-and-drop solution either.

```
public interface ISimpleState : IMixedRealityService
{
    Task<bool> HasKey(string key);
    Task SetState<T>(string key, T state);
    Task<T> GetState<T>(string key);
}
```

This service could be implemented using:
- a remote database
- a local database
- a local file
- an MRTK network connection service
- Photon
- etc. 

Tasks make it easy to deal with varying amounts of latency in these approaches.

SimpleStatePlayerPrefs
----
Wraps PlayerPrefs in the SimpleState interface. Used for testing.

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
+++ b/Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
@@ -8,6 +8,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.Sharing
     /// </summary>
     public interface ISimpleState : IMixedRealityService
     {
+        bool IsConnected { get; }
+
         Task<bool> HasKey(string key);
         Task SetState<T>(string key, T state);
         Task<T> GetState<T>(string key);
--- a/Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
+++ b/Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
@@ -1,0 +1,15 @@
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.Sharing
+{
+    /// <summary>
+    /// Simplest possible non-automatic state sharing.
+    /// Will typically be networked but could also plug into a file system or local database.
+    /// </summary>
+    public interface ISimpleState : IMixedRealityService
+    {
+        Task<bool> HasKey(string key);
+        Task SetState<T>(string key, T state);
+        Task<T> GetState<T>(string key);
+    }
+}
--- a/Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
+++ b/Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Threading.Tasks;
 using Microsoft.MixedReality.Toolkit.Core.Interfaces.Sharing;
+using Microsoft.MixedReality.Toolkit.Core.Utilities.Async.AwaitYieldInstructions;
+using Microsoft.MixedReality.Toolkit.Core.Utilities.Async;
 using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Core.Services.Sharing
@@ -12,10 +14,10 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services.Sharing
     /// </summary>
     public class SimpleStatePlayerPrefs : BaseService, ISimpleState
     {
-        public override void Initialize()
-        {
-            Name = "SimpleStatePlayerPrefs";
-        }
+        /// <inheritdoc />
+        public override string Name { get { return "SimpleStatePlayerPrefs"; } }
+
+        public bool IsConnected { get; set; }
 
         public int SimulatedLatency { get; set; }
 
@@ -26,6 +28,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services.Sharing
 
             await Task.Delay(SimulatedLatency);
 
+            await new WaitForUpdate();
+
             return PlayerPrefs.HasKey(key);
         }
         
@@ -35,6 +39,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services.Sharing
                 throw new Exception("Key cannot be empty.");
 
             await Task.Delay(SimulatedLatency);
+
+            await new WaitForUpdate();
 
             if (PlayerPrefs.HasKey(key))
             {
@@ -64,7 +70,19 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services.Sharing
 
             await Task.Delay(SimulatedLatency);
 
-            string serializedObject = JsonUtility.ToJson(state);
+            await new WaitForUpdate();
+
+            string serializedObject = string.Empty;
+            try
+            {
+                serializedObject = JsonUtility.ToJson(state);
+            }
+            catch (Exception e)
+            {
+                Debug.LogError("Failed to serialize object of type " + typeof(T) + ": " + e.ToString());
+                return;
+            }
+
             PlayerPrefs.SetString(key, serializedObject);
         }
     }
--- a/Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
+++ b/Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
@@ -10,23 +10,12 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services.Sharing
     /// Uses JSON to store and retrieve serialized objects from PlayerPrefs
     /// Useful for testing state sharing without a connection
     /// </summary>
-    public class SimpleStatePlayerPrefs : ISimpleState
+    public class SimpleStatePlayerPrefs : BaseService, ISimpleState
     {
-        public string Name { get { return "SimpleStatePlayerPrefs"; } }
-
-        public uint Priority { get { return 0; } }
-
-        public void Destroy() { }
-
-        public void Disable() { }
-
-        public void Enable() { }
-
-        public void Initialize() { }
-
-        public void Reset() { }
-
-        public void Update() { }
+        public override void Initialize()
+        {
+            Name = "SimpleStatePlayerPrefs";
+        }
 
         public int SimulatedLatency { get; set; }
 
--- a/Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
+++ b/Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
@@ -1,0 +1,82 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.MixedReality.Toolkit.Core.Interfaces.Sharing;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Services.Sharing
+{
+    /// <summary>
+    /// A local simulation of state sharing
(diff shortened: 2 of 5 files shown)


Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public async Task SetState<T>(string key, T state)
+        {
+            if (string.IsNullOrEmpty(key))
+                throw new Exception("Key cannot be empty.");
+
+            await Task.Delay(SimulatedLatency);
+
+            string serializedObject = JsonUtility.ToJson(state);
+            PlayerPrefs.SetString(key, serializedObject);
```
StephenHodgson: I can see issues with this, if we're not back on the main thread.

Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            return default(T);
+        }
+
+        public async Task SetState<T>(string key, T state)
+        {
+            if (string.IsNullOrEmpty(key))
+                throw new Exception("Key cannot be empty.");
+
+            await Task.Delay(SimulatedLatency);
```
StephenHodgson: Simulated Latency?

Railboy: The PlayerPrefs implementation is just for testing offline. Most implementations will be pulling state from a remote source.

Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
```diff
@@ -0,0 +1,82 @@
...
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Services.Sharing
+{
+    /// <summary>
+    /// A local simulation of state sharing
+    /// Uses JSON to store and retrieve serialized objects from PlayerPrefs
+    /// Useful for testing state sharing without a connection
+    /// </summary>
+    public class SimpleStatePlayerPrefs : ISimpleState
```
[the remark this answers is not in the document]
Railboy: I like this. Will change.

Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    /// <summary>
+    /// Simplest possible non-automatic state sharing.
+    /// Will typically be networked but could also plug into a file system or local database.
+    /// </summary>
+    public interface ISimpleState : IMixedRealityService
+    {
+        Task<bool> HasKey(string key);
+        Task SetState<T>(string key, T state);
+        Task<T> GetState<T>(string key);
```
chrisfromwork: Do we expect that most users will just want to obtain state values on each frame update? Would it be more convenient to use subscriptions/callbacks on value updates here compared to only having a getter?

Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.Sharing
+{
+    /// <summary>
+    /// Simplest possible non-automatic state sharing.
+    /// Will typically be networked but could also plug into a file system or local database.
+    /// </summary>
+    public interface ISimpleState : IMixedRealityService
+    {
+        Task<bool> HasKey(string key);
+        Task SetState<T>(string key, T state);
```
chrisfromwork: Do we have a perspective on how multiple asynchronous sets should behave? It feels like there is the possibility for a set to fail, which leads me to think this should return a Task<bool> compared to just a Task?

Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
chrisfromwork: It seems like this GetState may also fail/we may want to think about defining this as some sort of Task<bool, T> compared to only a Task<T>

Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// <summary>
+    /// A local simulation of state sharing
+    /// Uses JSON to store and retrieve serialized objects from PlayerPrefs
+    /// Useful for testing state sharing without a connection
+    /// </summary>
+    public class SimpleStatePlayerPrefs : BaseService, ISimpleState
+    {
+        public override void Initialize()
+        {
+            Name = "SimpleStatePlayerPrefs";
```
StephenHodgson: Please remove this. You'll likely want to override the Name property instead.

Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public int SimulatedLatency { get; set; }
```
StephenHodgson: inheritdoc?

Same with all the other interface implementations so that the xml docs propagate down to the rest of the things that use this service.

Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
Railboy: Maybe we could let devs pass an optional default that gets returned in the event of a failure?
`Task<T> GetState<T>(string key, T default);`
Or it could just throw an exception. For this system simplicity should take priority over robustness.

Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
Railboy: Do we have a generic dispatcher we could use?

StephenHodgson: yea

```
await new WaitForUpdate();
```

Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public async Task SetState<T>(string key, T state)
+        {
+            if (string.IsNullOrEmpty(key))
+                throw new Exception("Key cannot be empty.");
+
+            await Task.Delay(SimulatedLatency);
+
+            string serializedObject = JsonUtility.ToJson(state);
```
chrisfromwork: Is json capable of serializing any type? It seems like we may need to move away from a generic T towards a type that has guaranteed seriailization

Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
chrisfromwork: Asynchronous/multi-threaded exception handling seems a bit more difficult than assessing a return value. I'd suggest the default value or a valueOnFailure argument compared to throwing.

Railboy: Most likely so. Any implementation should cache stuff locally so we're not reaching across the network for stale data. But this process should be invisible. Callbacks are probably a notch or two more sophisticated than our needs in this case.

Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
Railboy: In the long term I'm counting on our [networking service](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3313) to handle serialization. So I'm hoping that system will be able to tell us whether we can serialize a type at compile time.

Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Interfaces.Sharing
+{
+    /// <summary>
+    /// Simplest possible non-automatic state sharing.
+    /// Will typically be networked but could also plug into a file system or local database.
+    /// </summary>
+    public interface ISimpleState : IMixedRealityService
```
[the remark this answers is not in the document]
Railboy: Roughly what form do you imagine that taking?

Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
Railboy: Fair point but I have a feeling it won't be used in practice. Here's a straightforward case:
```
await simpleState.SetKey("Object1.Position", transform.position);
await simpleState.SetKey("Object1.Rotation", transform.eulerAngles);
await simpleState.SetKey("Object1.Active", gameObject.activeSelf);
...
```
This is awkward by contrast. I have a feeling devs would opt to plow ahead and ignore failed set operations:
```
if (!simpleState.SetKey("Object1.Position", transform.position).Result)
    throw new Exception("Failed to set key Object1.Position"); // or handle some other way

if (!simpleState.SetKey("Object1.Rotation", transform.eulerAngles).Result)
    throw new Exception("Failed to set key Object1.Rotation");

if (!simpleState.SetKey("Object1.Active", gameObject.activeSelf).Result)
    throw new Exception("Failed to set key Object1.Active");
```
If we assume serialization issues will be caught at compile time (big if) then connection errors will be where 99% of problems arise. Maybe an IsConnected property would give them enough to bail early?

```
if (!simpleState.IsConnected)
    return;

await simpleState.SetKey("Object1.Position", transform.position);
...
```

Assets/MixedRealityToolkit/_Core/Sharing/SimpleStatePlayerPrefs.cs
Railboy: Nit: slightly awkward to use because it lives in the `Microsoft.MixedReality.Toolkit.Core.Utilities.Async.AwaitYieldInstructions` namespace, and without an explicit `using Microsoft.MixedReality.Toolkit.Core.Utilities.Async;` it raises a `'WaitForUpdate' does not contain a definition for 'GetAwaiter'` error. And that namespace isn't suggested in the Quick Actions menu.

chrisfromwork:
I also feel like given our new packaging story this could go into the experimental/extensions directory to allow people to use it. I would avoid adding new interface definitions to the core until the sharing working group has confirmed that they will be widely adopted.

Assets/MixedRealityToolkit/_Core/Interfaces/Sharing/ISimpleState.cs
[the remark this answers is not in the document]
chrisfromwork: good question, this seems like something we could bring up with the working group. That going said, given the time this review has taken, i wouldn't hold off on getting the code into the branch before having a persistence story

Railboy:
Closing this to pave the way for the stripped down app state extension.

Issue #3289, opened 2018-12-19
Have services implement IDisposable pattern


StephenHodgson:
## Overview
I'd like to propose that services and data providers implement `IDisposable` pattern.

Currently we take advantage of the unity events for disable and destroy which can be called multiple times, and in any order because they're publically availible methods.

Implementing this pattern gives the developer of each service more control over when resources and the service is deconstructed.

For the most part many services may not need to be changed or updated, and this pattern will be optional to implement in general.

PR #3290, opened 2018-12-19
added Disposable pattern to service interface.

StephenHodgson:
Overview
---
Implements IDisposable pattern for services.

Changes
---
- Fixes: #3289


keveleigh:
It might have been cleaner to just make `BaseService` an `IDisposable`, so we don't force devs into implementing `IDisposable` in their own service implementations. Might be worth discussing.

Issue #3195, opened 2018-11-29
Multiple Input detected in a loop


achaperon:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit VNext

**Describe the bug**

**To Reproduce**
https://github.com/achaperon/MRTK_Test
1. Open project and compile for HoloLens on IL2CPP
2. Launch the App
3. Make Airtap
4. Show the marker in front of the HoloLens and make Airtap at the same time

**Expected behavior**
Only one airtap should be detected per gesture

**Actual behavior**
Sometimes, just showing the marker trigger OnInputDown with the "select" Input Action
When the marker is in front of the HoloLens and you make the "ready" gesture. Multiple Airtap are detected (1 per update).
Actually, at each Update we have OnInputDown and OnInputUp with the "select" Input Action. 

**Unity Editor Version**
2018.2.10

**Mixed Reality Toolkit Release Version**
Beta Release 2018.9 MRTK VNext

**Additional context**

StephenHodgson:
Thanks for the report

keveleigh:
Should be fixed with #3293. Please report back and reopen if not.

PR #3291, opened 2018-12-20
Fix issue where HoloLens hands were using the same MixedRealityInteractionMapping array

keveleigh:
Overview
---
Since the controller mapping profile only contains one array for HoloLens hands (since there's no differentiation between handedness), both hands are reusing the same array. This causes their values to overwrite each other, and causes events to fire endlessly when one hand is selecting and the other isn't.

This bug fix instead creates a copy of the profile's array, so we aren't reusing the same structure. This will also fix cases where a source might be lost while selecting, which would cause the state to change when a second, unselecting source is detected.

Changes
---
- Fixes: #3195 
- Fixes: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3253#issuecomment-446037832

### DIFF
--- a/Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
@@ -117,7 +117,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Devices
                         MixedRealityInteractionMapping[] profileInteractions = controllerMappings[i].Interactions;
                         MixedRealityInteractionMapping[] newInteractions = new MixedRealityInteractionMapping[profileInteractions.Length];
 
-                        for (int j = 0; j < profileInteractions.Length; i++)
+                        for (int j = 0; j < profileInteractions.Length; j++)
                         {
                             newInteractions[j] = new MixedRealityInteractionMapping(profileInteractions[j]);
                         }
--- a/Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
+++ b/Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
@@ -114,7 +114,15 @@ namespace Microsoft.MixedReality.Toolkit.Core.Devices
                         controllerMappings[i].Handedness == ControllerHandedness &&
                         controllerMappings[i].Interactions.Length > 0)
                     {
-                        AssignControllerMappings(controllerMappings[i].Interactions);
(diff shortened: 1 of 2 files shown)


keveleigh:
Closing to investigate compatibility with Yoon's upcoming PR.

Assets/MixedRealityToolkit/_Core/Devices/BaseController.cs
```diff
@@ -114,7 +114,15 @@ public bool SetupConfiguration(Type controllerType)
                         controllerMappings[i].Handedness == ControllerHandedness &&
                         controllerMappings[i].Interactions.Length > 0)
                     {
-                        AssignControllerMappings(controllerMappings[i].Interactions);
+                        MixedRealityInteractionMapping[] profileInteractions = controllerMappings[i].Interactions;
+                        MixedRealityInteractionMapping[] newInteractions = new MixedRealityInteractionMapping[profileInteractions.Length];
+
+                        for (int j = 0; j < profileInteractions.Length; j++)
+                        {
+                            newInteractions[j] = new MixedRealityInteractionMapping(profileInteractions[j]);
```
StephenHodgson: Wouldn't this duplicate mappings for every controller?

@SimonDarksideJ thoughts?

keveleigh: Yep, it would.

As mentioned in the overview above:

>This will also fix cases where a source might be lost while selecting, which would cause the state to change when a second, unselecting source is detected.

keveleigh: A generic solution like this will fix the problem for all mappings that might be reused, like the touch mappings which appear to be reused for each individual touch.

StephenHodgson: Oh, I see, it's only duplicating it for the specific controller being configured, correct?

StephenHodgson: So if someone was to have two touch controllers be spawned then each would make it's own copy to use, do I have that right?

keveleigh: You have that exactly right. This happens in the setup of individual controllers.

What was happening before was that both HoloLens hand controllers (or touch controllers, etc) were referencing the same single array from the profile, so their values were fighting and overwriting each other.

StephenHodgson: I can see why that'd be a problem, lol

Issue #2960, opened 2018-10-23
[vNext] DragAndDropHandler moves with head


cre8ivepark:
**Describe the bug**
The object is moving with the head movement Not hand movement.
Symptom shows on both HoloLens & immersive headset.

**To Reproduce**
1. Create an object. 
2. Assign 'DragAndDropHandler.cs' script onto it.
3. Build & Deploy.
4. Grab with Airtap, hold and move. 

**Expected behavior**
The object should move with hand movement.

**Actual behavior**
The object does not move with hand movement.
The object is moving with HEAD(gaze cursor) movement.

**Unity Editor Version**
2018.1.9f1

**Mixed Reality Toolkit Release Version**
mrtk_development

cre8ivepark:
Found the symptom on immersive headset too. Objects move with headset movement.
Multiple objects are moving together instead of focused one. 'Focus Enabled' check box does not solve the issue.

StephenHodgson:
Likely this needs to be updated to utilize the gesture handler.

cre8ivepark:
@StephenHodgson In [DragAndDropHandler.cs](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit-SDK/Features/Input/Handlers/DragAndDropHandler.cs) currently, currentPointer.**TryGetPointerPosition**(out inputPosition) in HoloLens gives us headset position, not the hand postion. 

FYI, in HTK, we were able to access hand position using **TryGetGripPosition**. What is the corresponding method for this? We have same issue on TwoHandManipulation script too.


```
                case InteractionSourceInfo.Hand:
                    currentInputSource.**TryGetGripPosition**(currentInputSourceId, out inputPosition);
                    break;
                case InteractionSourceInfo.Controller:
                    currentInputSource.TryGetPointerPosition(currentInputSourceId, out inputPosition);
                    break;
```

StephenHodgson:
Yes, it's because the script needs to be updated to use the gesture handler.

cre8ivepark:
@StephenHodgson With IMixedRealityGestureHandler, how can we get hand grab position?

cre8ivepark:
@StephenHodgson That would be really helpful for the HoloLens developers, understanding how to migrate to vNext! Thank you!

cre8ivepark:
[...]

PR #3294, opened 2018-12-20
Fix DragAndDropHandler to properly support hands over the gaze pointer

keveleigh:
Overview
---
The original issue with DragAndDropHandler was that the gaze pointer was sending its own input source across the Pointer events, instead of the input source that originally sent the event.

This updates the pointer event methods to allow for an input source that's different from the pointer's input source. This allows matching position events to pointer events, and vice versa.

This change also adds support for moving objects with the tip of a controller's pointer ray, instead of with the source's position. Over the gaze pointer, the position is used.

Changes
---
- Fixes: #2960

Blocked By
---
- [ ] #3301

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
@@ -29,6 +29,10 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
         private const float MovementThreshold = 0.01f;
 
         [SerializeField]
+        [Tooltip("If true, the gaze cursor will disappear when the pointer's focus is locked, to prevent the cursor from floating idly in the world.")]
+        private bool setCursorInvisibleWhenFocusLocked = true;
+
+        [SerializeField]
         [Tooltip("Maximum distance at which the gaze can hit a GameObject.")]
         private float maxGazeCollisionDistance = 10.0f;
 
@@ -296,6 +300,11 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
             if (MixedRealityRaycaster.DebugEnabled && gazeTransform != null)
             {
                 Debug.DrawRay(GazeOrigin, (HitPosition - GazeOrigin), Color.white);
+            }
+
+            if (setCursorInvisibleWhenFocusLocked && GazePointer?.IsFocusLocked == GazeCursor?.IsVisible)
+            {
+                GazeCursor.SetVisibility(!GazePointer.IsFocusLocked);
             }
         }
 
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
@@ -232,8 +232,8 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
 
             public override bool TryGetPointerRotation(out Quaternion rotation)
             {
-                rotation = Quaternion.identity;
-                return false;
+                rotation = gazeTransform.rotation;
+                return true;
             }
 
             #endregion IMixedRealityPointer Implementation
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
@@ -239,9 +239,9 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
             /// </summary>
             /// <param name="mixedRealityInputAction">The input action that corresponds to the pressed button or axis.</param>
             /// <param name="handedness">Optional handedness of the source that pressed the pointer.</param>
-            public void RaisePointerDown(MixedRealityInputAction mixedRealityInputAction, Handedness handedness = Handedness.None)
+            public void RaisePointerDown(MixedRealityInputAction mixedRealityInputAction, Handedness handedness = Handedness.None, IMixedRealityInputSource inputSource = null)
             {
-                MixedRealityToolkit.InputSystem.RaisePointerDown(this, handedness, mixedRealityInputAction);
+                MixedRealityToolkit.InputSystem.RaisePointerDown(this, mixedRealityInputAction, handedness, inputSource);
             }
 
             /// <summary>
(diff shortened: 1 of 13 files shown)


Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
```diff
@@ -28,6 +28,10 @@ public class GazeProvider : InputSystemGlobalListener, IMixedRealityGazeProvider
 
         private const float MovementThreshold = 0.01f;
 
+        [SerializeField]
+        [Tooltip("If true, the gaze cursor will disappear when the pointer's focus is locked, to prevent the cursor from floating idly in the world.")]
+        private bool setCursorInvisibleWhenFocusLocked = true;
```
StephenHodgson: why not just update the cursor position to the "locked" object's position?

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -783,19 +783,19 @@ public void RaiseFocusExit(IMixedRealityPointer pointer, GameObject unfocusedObj
             };
 
         /// <inheritdoc />
-        public void RaisePointerDown(IMixedRealityPointer pointer, MixedRealityInputAction inputAction)
+        public void RaisePointerDown(IMixedRealityPointer pointer, Handedness handedness, MixedRealityInputAction inputAction)
```
StephenHodgson: pointers likely shouldn't know about handedness, as they're not controllers, but objects controlled and driven by controllers.

Likely the InputDown/up should be used instead.

Assets/MixedRealityToolkit/_Core/Definitions/Physics/RayStep.cs
```diff
@@ -43,6 +47,11 @@ public void CopyRay(Ray ray, float rayLength)
             Terminus = Origin + (Direction * Length);
         }
 
+        public bool Contains(Vector3 point)
+        {
+            return Vector3.Distance(Origin, point) + Vector3.Distance(point, Terminus) - Length < epsilon;
```
StephenHodgson: pretty sure there's a built in value for this we can use instead?

Assets/MixedRealityToolkit/_Core/EventDatum/Input/MixedRealityPointerEventData.cs
```diff
@@ -31,10 +31,20 @@ public class MixedRealityPointerEventData : InputEventData
         /// </summary>
         /// <param name="pointer"></param>
         /// <param name="inputAction"></param>
+        /// <param name="handedness"></param>
+        /// <param name="inputSource"></param>
         /// <param name="count"></param>
-        public void Initialize(IMixedRealityPointer pointer, MixedRealityInputAction inputAction, int count = 0)
+        public void Initialize(IMixedRealityPointer pointer, MixedRealityInputAction inputAction, Handedness handedness = Handedness.None, IMixedRealityInputSource inputSource = null, int count = 0)
```
StephenHodgson: again pointers shouldn't know about handedness, although I do see the benefit of the input source reference.

Assets/MixedRealityToolkit-SDK/Features/Input/GazeProvider.cs
keveleigh: If this is set to false, that's what will happen. It looks a little strange, in my opinion, with the gaze cursor floating off and following the object, so I added the configuration to hide it if desired.

Assets/MixedRealityToolkit/_Core/Definitions/Physics/RayStep.cs
keveleigh: Yeah, I tried to use Unity's built in epsilon, but I was hitting slightly larger values than that supports.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
keveleigh: Pointer events [already know about handedness](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3294/files/725d665e1f3a8440b737bea2ad77cf4b68b751f4#diff-83ff2790c22f67c0d7d4a7a413be9c29L795). This isn't anything new, but git is confusing the two lines. My link above links to this exact line already existing.

Assets/MixedRealityToolkit/_Core/EventDatum/Input/MixedRealityPointerEventData.cs
keveleigh: Pointers [already know about handedness](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/_Core/EventDatum/Input/MixedRealityPointerEventData.cs#L49).

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
StephenHodgson: Maybe that's the problem. They shouldn't. The input source is what should know about handedness.

keveleigh: The problem with what?

StephenHodgson: The pointer doesn't have a handedness property. So the property shouldn't be passed through the event data.

The pointer should only have a reference to the input source, who's controller implementation has the handedness property.

Again. Pointers shouldn't know about handedness.

Assets/MixedRealityToolkit/_Core/EventDatum/Input/MixedRealityPointerEventData.cs
StephenHodgson: No. They don't have a property with handedness. Nor should they. Those kinds of properties only exist in the IMixedRealityController

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
keveleigh: I'm not adding anything new here. The existing methods and events all have handedness information. If you'd like to propose removing them, please file a proposal for discussion. This PR has nothing to do with handedness.

StephenHodgson: Well then let's come back to this PR once that change is in.

Give me 10 min.

keveleigh: This PR doesn't depend on a change to handedness. That proposal can be addressed following this one.

StephenHodgson: Yes it does, because I don't want to make a change to the IMixedRealityInputSystem contract more than once.

StephenHodgson:
Is there a reason why we merged a PR that had change requests?

StephenHodgson:
Since when did we merge pull requests that are blocked or have change requests?

This is absolutely unacceptable.

SimonDarksideJ:
I would have to agree, we cannot be in a position where changes proposed by the community are refused to be accepted until SM staff are finally happy, and then merge a change that the community has severe objections too.  This change introduces more risk as highlighted by @StephenHodgson above, these comments just seem to be ignored by consensus internally.

davidkline-ms:
This change solved a customer usability issue on HoloLens. We have heard feedback from our partners that previous behavior was confusing and unintuitive.

Issue #3299, opened 2018-12-21
Proposal: Add a convenience method for easier access to detected controllers


StephenHodgson:
## Overview
I'd like to add a convenience method to the input system interface contract to make it easier to get a reference to a detected controller using the input source.

This will make getting controller properties easier.

```
void IMixedRealityPointerHandler.OnPointerUp(MixedRealityPointerEventData eventData)
{
    var controller = MixedRealityToolkit.InputSystem.TryGetController(eventData.InputSource);
    Debug.Assert(controller != null);
    Handedness handedness = controller.ControllerHandedness;
}
```

Yoyozilla:
seems useful, not adoption blocking. will discuss on teams and slack and close before 1/16

Yoyozilla:
Agreed this is good to have, but not critical for 2018 release. We will opportunistically pull this into 2018.

PR #3300, opened 2018-12-21
added TryGetController method

StephenHodgson:
Overview
---
Added `TryGetController` utility method to Input System

Changes
---
- Fixes: #3299

### DIFF
--- a/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
+++ b/Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
@@ -473,17 +473,19 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input
         #region IMixedRealityController Utilities
 
         /// <inheritdoc />
-        public IMixedRealityController TryGetController(IMixedRealityInputSource inputSource)
+        public bool TryGetController(IMixedRealityInputSource inputSource, out IMixedRealityController controller)
         {
             foreach (IMixedRealityController mixedRealityController in DetectedControllers)
             {
                 if (inputSource.SourceId == mixedRealityController.InputSource.SourceId)
                 {
-                    return mixedRealityController;
+                    controller = mixedRealityController;
+                    return true;
                 }
             }
 
-            return null;
+            controller = null;
+            return false;
         }
 
         #endregion IMixedRealityController Utilities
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -470,6 +470,24 @@ public void ClearFallbackInputStack()
 
         #endregion Fallback Input Handler Options
 
+        #region IMixedRealityController Utilities
+
+        /// <inheritdoc />
+        public IMixedRealityController TryGetController(IMixedRealityInputSource inputSource)
```
keveleigh: Should this be `public bool TryGetController(IMixedRealityInputSource inputSource, out IMixedRealityController controller)` to match some of our other `TryGet` methods? I'm not sure if we've been consistent with the naming/behavior everywhere, but all the ones I just found on my search behave this way.

StephenHodgson: Sure I can update it to be that instead :)

Assets/MixedRealityToolkit/_Core/EventDatum/Input/InputEventData.cs
```diff
@@ -16,7 +17,14 @@ public class InputEventData : BaseInputEventData
         /// <summary>
         /// Handedness of the <see cref="IMixedRealityInputSource"/>.
         /// </summary>
-        public Handedness Handedness { get; private set; } = Handedness.None;
+        public Handedness Handedness
+        {
+            get
+            {
+                var controller = MixedRealityToolkit.InputSystem.TryGetController(InputSource);
```
StephenHodgson: I suppose with this change we should also remove the handedness property in `Initialize`

StephenHodgson: Or we can remove `eventData.Handedness` all together

keveleigh: Strongly against removing `eventData.Handedness`. It provides easy, discoverable access to data that is valuable when performing handed interactions.

StephenHodgson: Pointers shouldn't know about hands so it wouldn't matter.

PR #3301, opened 2018-12-21
Add ability to update the PointerData for a pointer with locked focus

keveleigh:
Overview
---
Previously, a pointer with locked focus wasn't being updated. Its data was stale from the moment focus became locked. This leads to strange visualizations when the locked object or the pointer is moved, leaving the cursor behind or rendering the pointer line to the wrong point.

Changes
---
- Related to #2960, #3294


davidkline-ms:
@keveleigh, please resolve merge conflict. thx!

PR #3303, opened 2018-12-29
MRTK/Standard Shader Clipping Primitives Improvements & Additions

Cameron-Micka:
Overview
---
Improved per-pixel primitive clipping feature by adding:

primitiveclipping

-	Performant sphere, and box shape clipping (plane clipping already existed)
-	Ability to specify which side of the primitive to clip against (inside or outside)
-	User friendly shader GUI and gizmo support
-	Examples in the material gallery

primitiveclippinggizmos

Other changes:
-	Improved refractive glass material
-	Moved the HoverLight component to the core for nameof support in the shader GUI


Assets/MixedRealityToolkit/_Core/Utilities/ClippingPrimitive.cs
```diff
@@ -0,0 +1,256 @@
...
+        protected virtual void UpdateRenderers()
+        {
+            if (renderers == null)
+            {
+                return;
+            }
+
+            for (int i = 0; i < renderers.Count; ++i)
+            {
+                Renderer renderer = renderers[i];
```
StephenHodgson: There's a couple of places in each of the scripts where there are variable names that hide each other.

A simple fix is an underscore prefix for these local vars `_renderer`

Assets/MixedRealityToolkit/_Core/Utilities/ClippingPrimitive.cs
```diff
@@ -0,0 +1,256 @@
...
+                return;
+            }
+
+            Initialize();
+            UpdateRenderers();
+        }
+
+        protected void LateUpdate()
+        {
+            UpdateRenderers();
```
StephenHodgson: Do we also want to check is playing or is not the editor in this update loop as well?

StephenHodgson: Do we want both the update loops running at the same time?

Assets/MixedRealityToolkit/_Core/Utilities/ClippingPrimitive.cs
Cameron-Micka: Ah, I wish Unity would remove that obsolete member. It's odd I don't get that warning locally anymore?
renderer


I'll change them all to _renderer just in case.

Assets/MixedRealityToolkit/_Core/Utilities/ClippingPrimitive.cs
Cameron-Micka: I agree this has got somewhat confusing since the preprocessor removals when moving to vNext. (This code was pushed down from ClipPlane.cs) I'll add a comment to make it a bit more clear.

davidkline-ms:
> Such a good feature. Thanks @Cameron-Micka!

Agreed!

Cameron-Micka:
@radicalad @davidkline-ms thank you!

Issue #3262, opened 2018-12-12
Unity UI Buttons do not work with WMR Motion Controllers


PhilippRaab:
## Overview

The click event of the Unity UI buttons is not triggered with the WMR Motion Controllers. The button events with spatial mouse on the other hand are working correctly.

## Expected Behavior
Pressing the Trigger (Select) on the WMR Motion Controllers while pointing at a button should trigger the OnClick event.

## Actual Behavior
OnClick event is not triggered.

## Steps to reproduce

- Use scene and script in [MinimalUIButtonTest.zip](https://github.com/Microsoft/MixedRealityToolkit-Unity/files/2671976/MinimalUIButtonTest.zip) with latest MRTK and standard profile. 
- Start project in Unity Editor.
- Click on the button with the spatial mouse. The Unity console logs "Click"
- Click on the button with a WMR Motion Controller. There is no console log.

## Unity Editor Version
2018.2.17f1

## Mixed Reality Toolkit Release Version
mrtk_development branch

StephenHodgson:
Thanks for the report, I'll check it out as soon as I have time.

StephenHodgson:
I'm not sure that we have it to set up to fire the "OnClick" event out of the box, you'll need to write an input handler wrapper that will forward the event to the ui button.

We're investigating ways to handle input without the wrapper (We have an interactions PR with a lot of the basic functionality in the works.)

PhilippRaab:
Thank you for investigating. The spatial mouse works out of the box and what I understand raises the same events as the controllers. It would be nice when all input devices behave identical.

Could you provide a example of the input handler wrapper, to get a temporary workaround?

wassx:
Tested with latest update form vnext (5f945d25514a87e4d072d894a6a9f90f1c909f31)
and now I get 1 click per pointer click. The important part was to add the pointer click handler and let the event handle by a buttonhandler class. The button itself has no events linked.
**ADDITIONALLY** I delete the camera, added a blank new camera and let vnext configure the mixedrealityplayspace. 
Now the pointer of the WMR works on UI Elements without a collider and triggers a single event on 'select'
Only thing is, that the ray of the controller disappears and the pointer seems to be a little bit off.
![image](https://user-images.githubusercontent.com/3955210/50182390-84dd5000-030f-11e9-9d7a-df4bd0e2aa9d.png)

[...]

PR #3305, opened 2019-01-02
Fix unity ui button click and focus with pointer

wassx:
Overview
---
I was not able to focus and click a unity ui button without colliders.
This was discussed in https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3262
Now I found a solution which works for windows mixed reality controller pointer inputs and not GazeProvider only.

After digging into the code, I found out the difference between gaze and pointer in the focusprovider.
For the gazeprovider, the unity ui needs no collider to get hovered, but for the pointer it seemed necessary.
Basically it turns out, that for the gazeinput the (FocusProvider:707)
`if (pointer.Details.LastRaycastHit.distance > uiRaycastResult.distance)
                        {
                            overridePhysicsRaycast = true;
                        }`
is true, but for the pointer input, the LastRaycastHit.distance is always smaller than the uiRaycastResult.distance. Hence it does not override the physics raycast and the pointer cursor does not stop at the ui.
For me it is suspicious that the ray starting point (for linepointer) is the starting point of the RayStep and not the controller position. But the distances compared should be from head->hitTarget vs. controller->hitTarget.

Thanks to @andriworld for the fix with the focusedObject!

Changes
---
- Fixes: # https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3262


Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -904,6 +904,10 @@ private GraphicInputEventData HandlePointerUp(IMixedRealityPointer pointer)
 
             GraphicInputEventData graphicEventData;
             FocusProvider.TryGetSpecificPointerGraphicEventData(pointer, out graphicEventData);
+            
+            GameObject focusedObject = FocusProvider.GetFocusedObject(pointer);
```
StephenHodgson: We should already have a valid focused object in the `HandleEvent` call, otherwise we're just covering up the problem, not fixing it.

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -361,12 +361,6 @@ public GameObject GetFocusedObject(IMixedRealityPointer pointingSource)
             FocusDetails focusDetails;
             if (!TryGetFocusDetails(pointingSource, out focusDetails)) { return null; }
 
-            GraphicInputEventData graphicInputEventData;
```
wassx: I think this code is some leftover.

Assets/MixedRealityToolkit-SDK/Features/Input/MixedRealityInputManager.cs
```diff
@@ -904,10 +904,6 @@ private GraphicInputEventData HandlePointerUp(IMixedRealityPointer pointer)
 
             GraphicInputEventData graphicEventData;
             FocusProvider.TryGetSpecificPointerGraphicEventData(pointer, out graphicEventData);
-            
-            GameObject focusedObject = FocusProvider.GetFocusedObject(pointer);
-            graphicEventData.selectedObject = focusedObject;
```
wassx: graphicEventData had the selected object already in it's .pointerCurrentRaycast.gameObject

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -391,6 +385,7 @@ public bool TryGetSpecificPointerGraphicEventData(IMixedRealityPointer pointer,
             if (TryGetPointerData(pointer, out pointerData))
             {
                 graphicInputEventData = pointerData.GraphicEventData;
+                graphicInputEventData.selectedObject = pointerData.GraphicEventData.pointerCurrentRaycast.gameObject;
```
wassx: Don't know if this is the correct way, but here I assign the needed object

wassx:
This fix works currently for WMR only. Tried on the hololens and now the gazepointer goes through the ui. Investigating,...

wassx:
Retested and it works on Hololens and WMR. The mentioned issue above resulted from a wrong display in my uiraycastcamera. Just had to reset.

davidkline-ms:
@wassx, I have pulled your change (and the zip file from the bug) and am not getting logs for clicking with the motion controller or mouse. I have the button object setup as shown here: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3262#issuecomment-448374935

thoughts?

wassx:
Hi, the setup is now easier. Please have a look on the added sample scene. There's no additional point click handler needed as this is done via the eventsystem. Tested with wmr controller spatial mouse and hololens.
[buttonTest.zip](https://github.com/Microsoft/MixedRealityToolkit-Unity/files/2746941/buttonTest.zip)
image

Assets/MixedRealityToolkit-SDK/Features/Input/FocusProvider.cs
```diff
@@ -801,6 +796,11 @@ private bool RaycastGraphicsStep(PointerData pointerData, RayStep step, LayerMas
             {
                 if (pointerData.CurrentPointerTarget != null)
                 {
+                    float distance = 0f;
+                    for (int i = 0; i <= pointer.RayStepIndex; i++) {
+                        distance += pointer.Pointer.Rays[i].Length;
```
keveleigh: I think this will overshoot the true distance of the raycast, since the hit might be somewhere in the middle of the final ray step. For small raysteps, it'll probably be negligible, but it might cause a problem with larger ray steps. Not fully sure if that'll cause an issue with this calculation though.

davidkline-ms:
> Please have a look on the added sample scene. 

I'll definitely do that ASAP.  Also, this PR will need to resolve a conflict before it can be merged.

Thanks!

davidkline-ms:
when I pull the change, I am not finding an example scene. there also appears to be some issues wrt the recent namespace and folder name changes (ex: mrtk-examples -> mrtk.examples) …

I'll take some more time on this later in the week once all the major moving pieces (ex: platform providers moving into separate assemblies) stop moving

thanks!

davidkline-ms:
Thanks for this fix @wassx! Sorry for the approval delay.

PR #3318, opened 2019-01-09
Add ConfigurationProfil connection for Extension Service

achaperon:
Overview
---
Add ScriptableObject member as configurationProfil in the BaseExtensionService class to store the profil.
Add the configuration.ConfigurationProfil as additionnal parameter when registering the IMixedRealityExtensionService.

Change all IMixedRealityExtensionService implementation constructor to add ScriptableObject has third parameter to accept the configuration profil.

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit-Tests/Services/TestExtensionService1.cs
+++ b/Assets/MixedRealityToolkit-Tests/Services/TestExtensionService1.cs
@@ -2,12 +2,13 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.MixedReality.Toolkit.Core.Services;
+using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Tests.Services
 {
     internal class TestExtensionService1 : BaseExtensionService, ITestExtensionService1
     {
-        public TestExtensionService1(string name, uint priority) : base(name, priority) { }
+        public TestExtensionService1(string name, uint priority, ScriptableObject profil) : base(name, priority, profil) { }
 
         public bool IsEnabled { get; private set; }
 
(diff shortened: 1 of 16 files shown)


Assets/MixedRealityToolkit-Tests/Services/TestExtensionService1.cs
```diff
@@ -2,12 +2,13 @@
...
 
 using Microsoft.MixedReality.Toolkit.Core.Services;
+using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Tests.Services
 {
     internal class TestExtensionService1 : BaseExtensionService, ITestExtensionService1
     {
-        public TestExtensionService1(string name, uint priority) : base(name, priority) { }
+        public TestExtensionService1(string name, uint priority, ScriptableObject profil) : base(name, priority, profil) { }
```
SimonDarksideJ: typo throughout the commit
"ScriptableObject profil" -> "ScriptableObject profile"

SimonDarksideJ: mostly "profil" to "profile"

achaperon: But I didn't get the point of the other comment.
The configuration profile is now a part of the BaseExtensionService but not a part of the BaseService class.
Actually on the MRTK, the profile is the only parameter that is on the configuration but not used at all on the services.
Do you have any concern about connecting this parameter. For me it make sens to have it on the BaseExtensionService as it's also a parameter that can define the service itself.
You didn't use it on your extension services but for me it's very useful to configure mine.
Can you detail more your concern about it? we should find a solution to that.
Best regards,

PR #3350, opened 2019-01-18
[vNext][UX]Lines closest point / length

Railboy:
Overview
---
Adds two functions to BaseMixedRealityLineDataProvider:

```
/// Iterates along line until it finds the point closest to worldPosition
public Vector3 GetClosestPoint(Vector3 worldPosition, int resolution = 5, int maxIterations = 5)

 /// Iterates along line until it finds the length closest to worldposition.
public float GetNormalizedLengthFromWorldPos(Vector3 worldPosition, int resolution = 5, int maxIterations = 5)
```

These make it easier to create line-based effects that react to objects in the world (eg controllers).

lines


Assets/MixedRealityToolkit/_Core/Utilities/Lines/DataProviders/BaseMixedRealityLineDataProvider.cs
```diff
@@ -464,6 +464,74 @@ public void SetPoint(int pointIndex, Vector3 point)
...
+            if (start > 0)
+            {
+                searchLength = (end - start) / resolution;
+            }
+            else
+            {
+                searchLength = end / resolution;
+            }
+
+            Vector3 currentPoint = Vector3.zero;
```
StephenHodgson: I'm not sure you need this. You'll be overwriting it each time you get the unclamped point.

Railboy: If you mean the initialization is unnecessary, you're right. It's something I do out of habit. If you mean I could do away with `currentPoint` altogether that's also true, though the result would be less readable.

davidkline-ms: I'm a huge fan of readability. Since you do not read currentPoint before you write to it, you can safely do away with the initialization, but I _do_ like the habit of always initializing variables.

StephenHodgson: I disagree. Harder to immediately understand what is going on.

PR #3348, opened 2019-01-17
Tree changes to align with packaging and component separation

davidkline-ms:
Overview
---
This change updates the layout of the source tree to better align with the separation of components and makes it easier to identify which code files are part of a given component.

As part of this change, namespaces have been modified to reflect the componentization.

Please note that this change has the potential to break client code in the following ways:
- Custom profiles will require modification to point to the new assembly/namespace/class names
- Code that directly calls concrete classes will need to be modified.

Also, please note that the following device/providers have not been moved, due to changes required as outlined in #3074 : 
- UnityInput
- WindowsMixedReality
- OpenVR

These will be addressed in a future change.

Changes
---
- Fixes #3341

PR #3354, opened 2019-01-18
complete top level folder rename, remove _Core folder

davidkline-ms:
Overview
---
This change is the second part of the tree alignment update, started with #3348. The top level folders that previously contained a '-' separator have been updated to use a '.' and the contents of the MixedRealityTookit/_Core folder have been moved to MixedRealityToolkit.

There are _zero_ code changes in this PR.

This change was discussed and approved in shiproom on 2019-01-16.


radicalad:
@Railboy Christmas came early (or late I guess since its January).

@davidkline-ms - just to confirm, This isn't a breaking change correct?

davidkline-ms:
> just to confirm, This isn't a breaking change correct?

as no code changed (and as the asmdef files were untouched), this should _not_ break any code

PR #3355, opened 2019-01-19
changing accessability on ClippingPrimitive.ClippingSide

radicalad:
Overview
Changing ClippingPrimitive.ClippingSide to an explicit property. For some reason the lambda prevents other classes from accessing it.

Changes
---
Added an explicit get/set.


Assets/MixedRealityToolkit/_Core/Utilities/ClippingPrimitive.cs
```diff
@@ -30,7 +30,7 @@ public enum Side
         /// <summary>
         /// The renderer(s) that should be affected by the primitive.
         /// </summary>
-        public Side ClippingSide => clippingSide;
+        public Side ClippingSide { get { return clippingSide; } set { clippingSide = value; } }
```
davidkline-ms: for ease of reading, please reformat similar to 

```
public Side ClippingSide
{
    get { … }
    set { … }
}
```

PR #3372, opened 2019-01-25
[vNext][UX] Line Usability Tweaks

Railboy:
Overview
---
Various updates to make line tools more usable for artists and designers. These changes are based on feedback gathered while using these tools over the past few weeks.

Changes
---
- ~~Lines.DataProviders and Lines.Renderers namespaces have been consolidated into Lines. This was a big discoverability problem. Even after working with these tools for weeks it was frequently necessary to hunt and peck for class names.~~
- ~~BaseMixedRealityLineDataProvider has been renamed to BaseLineDataProvider. Again, class is now more discoverable via intellisense.~~
- Line renderers now update in-editor when base data provider is being manipulated.
-  Line data providers now draw a simple gizmos approximation. Working with more than one line at a time could be confusing as they would disappear when not selected.
- Line renderers no longer use RequireComponent for their line data provider. There were many situations where we wanted renderer components to exist on a separate object from its data provider.
- When RotationMode is set to None, lines now return transform's rotation. This leads to more predictable results.
- When a MixedRealityLineRenderer is disabled it now automatically disables its accompanying LineRenderer, which is the expected behavior.
- StripMeshRenderer now uses GetVelocity instead of GetRotation to avoid 'nocks' in mesh.
- StripMeshLineRenderer serializes its mesh renderer so the renderer doesn't get created again at runtime.


davidkline-ms:
> StripMeshLineRenderer serializes its mesh renderer so the renderer doesn't get created again at runtime.

Any risk inherent in this? Is renderer creation a one time or many time cost per line instance?

Assets/MixedRealityToolkit/Utilities/Lines/Renderers/StripMeshLineRenderer.cs
```diff
@@ -5,26 +5,30 @@
...
         private Material lineMaterial = null;
-
         [SerializeField]
         private float uvOffset = 0f;
 
-        private Mesh stripMesh;
-        private Material lineMatInstance;
+        [SerializeField]
+        [HideInInspector]
         private MeshRenderer stripMeshRenderer;
```
keveleigh: Should we (do we, and I missed it?) try to grab this [in `OnValidate`](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/b170096ebfb6ed9a2c33191df0f500e8f11c5876/Assets/MixedRealityToolkit-SDK/Features/Audio/Influencers/AudioLoFiEffect.cs#L96) so it's serialized at edit time?

keveleigh:
>Given the point in the schedule, I would like to shelve this until v3 and take the changes / provide a migration path there.

I'm good with taking this now (but am happy to go with the consensus). It feels fairly self-contained and is stated to improve our support for artists and designers, which is definitely an area I feel has room for improvement.

Assets/MixedRealityToolkit/Utilities/Lines/Renderers/StripMeshLineRenderer.cs
Railboy: We create / destroy this object in OnEnable / Disable, and the component runs in edit mode, so it will be created and serialized at edit time by default. (Does that answer your question?)

keveleigh: Didn't realize the component runs in edit mode! Great!

Railboy:
>Per our discussion earlier this week, I'd recommend leaving the DataProviders namespace and just mark each class in it as Obsolete. [Obsolete("This class has been moved to blah blah blah", false)]. That will prevent the breaking change and include your much more intuitive namespace updates.

@radicalad I'd be fine with this if not for the confusing editor behavior. Eg clicking AddComponent and typing in Line would reveal 2 versions of every line-related component - one from the old namespaces, one from the new. This could also lead to confusing mismatches between incompatible data providers and renderers. Probably best to rip the band-aid off.

>Given the point in the schedule, I would like to shelve this until v3 and take the changes / provide a migration path there.

@davidkline-ms Sure, I can live with that. In the meantime I can open separate PRs for some of the non-breaking changes.

davidkline-ms:
> I'd be fine with this if not for the confusing editor behavior.

You can set the isError property to true in the ObsoleteAttribute. this will force a compile failure but give the correct information to the developer

davidkline-ms:
> Sure, I can live with that.

Per @keveleigh's comment, if it makes artist / designer lives easier, I can support the change. It _does_ look nicely self contained

Railboy:
@davidkline-ms @radicalad I'm not sure I'm making the problem clear. I think we've bumped into a special case where our planned usage of Obsolete breaks down.

This is what developers will see if we mark the components from the old namespace as [Obsolete] instead of just removing them:

obsolete

Which of the listed component do you use? Both function. If you choose the obsolete component by mistake, the script will be removed in a future update, leaving your game object with a missing script reference.

We could keep devs from attaching obsolete components by using Obsolete with error on the abstract base line class  - each obsolete concrete class would throw an error like `namespaces 'Lines.DataProviders' and 'Lines.Renderers' have been consolidated into 'Lines' - you must delete the old components before the project will compile.`

But asking folks to delete files every time they start a fresh project is pretty obnoxious.

Thoughts?

radicalad:
Bummer.

Railboy:
I think I'll revert the namespace and class name changes until we've settled on an 'Obsolete' convention that doesn't suffer from this double-component problem.

I'll keep the minor (non-breaking) tweaks in place.

PR #3382, opened 2019-01-29
Channel Packer Editor + Other Shader Features

Cameron-Micka:
Overview
---
Adding a few improvements to the MRTK/Standard shading system:

- A GPU accelerated channel packer editor for easy channel map generation
- Spherical harmonics support for better lighting
- Vertex color support (can now be used with the Unity LineRenderer)
- Improved lightmapping support with a meta pass
- HDR emission support
- Minor bug fixes

Screenshot of the channel packer:
channelpacker

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/ChannelPackerWindow.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/ChannelPackerWindow.cs
@@ -1,0 +1,238 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using UnityEditor;
+using UnityEngine;
+using System.IO;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Inspectors
+{
+    public class ChannelPackerWindow : EditorWindow
+    {
+        private enum Channel
+        {
+            Red = 0,
+            Green = 1,
+            Blue = 2,
+            Alpha = 3,
+            RGBAverage = 4
+        }
+
+        private const float defaultUniformValue = -0.01f;
+
+        private Texture2D metallicMap;
+        private Channel metallicMapChannel = Channel.Red;
+        private float metallicUniform = defaultUniformValue;
+        private Texture2D occlusionMap;
+        private Channel occlusionMapChannel = Channel.Green;
+        private float occlusionUniform = defaultUniformValue;
+        private Texture2D emissionMap;
+        private Channel emissionMapChannel = Channel.RGBAverage;
+        private float emissionUniform = defaultUniformValue;
+        private Texture2D smoothnessMap;
+        private Channel smoothnessMapChannel = Channel.Alpha;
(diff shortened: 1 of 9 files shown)


StephenHodgson:
What's the channel packer for?

Any examples of its usage?

Cameron-Micka:
@StephenHodgson That's a great question! There is a small example of it's usage already in the MaterialGallery:

channelexample

Let me see with some folks to see if I can share some better example images. But, it's main purpose is to make it easier for content creators to pack all their PBR textures into a single texture (channel map). This results in less texture samples in the shader, and better performance. Before this PR it was a very manual process to pack all your textures together in Photoshop, etc. But, with this tool you can do it "automagically" within Unity. It will also try and auto populate most of the settings for you from a Unity Standard material if you already have one for reference you want to convert to a MRTK/Standard material.

StephenHodgson:
Awesome, I think this will also help a lot with the glTF asset importer as well. I already have to swap the metal channels because Unity has to be different from every other engine when implementing PBR, haha.

Cameron-Micka:
Cool! Yeah, Unity just had to be "different."

Here is a better example of the channel maps (and spherical harmonics) in action. This has per pixel metallicness, smoothness, emission, and occlusion all coming from one channel map:

mushroom

StephenHodgson:
Last question, does this also have hooks in to call directly so I can pack textures through an automated import process?

Assets/MixedRealityToolkit/Inspectors/ChannelPackerWindow.cs
```diff
@@ -0,0 +1,238 @@
...
+            string path = EditorUtility.SaveFilePanel("Save Channel Map", "", filename, "png");
+
+            if (path.Length != 0)
+            {
+                byte[] pngData = channelMap.EncodeToPNG();
+
+                if (pngData != null)
+                {
+                    File.WriteAllBytes(path, pngData);
+                    Debug.LogFormat("Saved channel map to: {0}", path);
```
Jarodshow: Should this notification maybe be a dialog box with an option to open the folder that the PNG is located in?

StephenHodgson: Likely better to specify a save path in the window, not in a dialog.

Cameron-Micka: This is a great usability idea! I'm on the fence if this would be useful to the user or an annoyance to users who are processing tons of textures to have to hit "ok" after every save?

StephenHodgson: I'm also worried about the automated import process being interrupted with a dialog as well.

Jarodshow: A dialog would be kind of intrusive, I agree with you guys.

Issue #3313, opened 2019-01-04
vNext Task: Define core networking interfaces


davidkline-ms:
## Overview
Goals:
- Componentized (ie task focused interfaces)
- Flexible for virtually any transport
- Will serve as the basis for higher level networking and sharing components

Current thinking can be found in #3243 and the attached image.

![20190104_105504](https://user-images.githubusercontent.com/13281406/50709006-2efced80-101b-11e9-9d90-509e67574ea1.jpg)

maluoi:
When sending and receiving an `Object`, does that imply it will serialize/deserialize that `Object` for network transfer? Since serialization is a big topic, it might be nice to know the serialization capabilities of the network API underneath. I'd love to be able to query if my particular `Object` can or cannot be serialized before trying to send it!

davidkline-ms:
> When sending and receiving an Object, does that imply it will serialize/deserialize that Object for network transfer?

Good question! Object was a placeholder in the above. It may be that the API supports simpler times like strings and arrays of intrinsic types.

davidkline-ms:
Looks like we should be able to check if a type is serializable on most platforms...

https://docs.microsoft.com/en-us/dotnet/api/system.type.isserializable?view=netframework-4.7.2

maluoi:
Good serialization code will also only allow specific `Object` types, due to the potential for malicious users injecting hostile data packets that de-serialize into system takeovers :)

davidkline-ms:
> Good serialization code will also only allow specific Object types, due to the potential for malicious users injecting hostile data packets that de-serialize into system takeovers :)

Excellent point, @maluoi! Thanks for keeping us honest! :)

Railboy:
Let's include an args object in the Send method. There will be cases where devs will need to specify things like delivery targets (all vs specific user), delivery method (direct vs via server), etc etc.

Dumping optional stuff into an extendable args class will future proof the interface without cluttering it up.

`Send([TBD] obj, ICancelationToken token, SendArgs args = null)`

>Since serialization is a big topic, it might be nice to know the serialization capabilities of the network API underneath. 

Agreed, and since writing a truly universal serializer is an impossible task devs should have the option to provide a their own. There will be cases where our default can't meet their needs regarding speed / security / bandwidth / platform / whatever:

```
public interface INetworkSession 
{
     void SetFormatter (IRemotingFormatter formatter);
     ...
}
```

Yoyozilla:
[...]

PR #3383, opened 2019-01-30
Refactor of Spectator View (Preview)

chrisfromwork:
Overview
---
This review refactors some previously existing code, spectator view (preview), and adds it to the newer version of the MRTK.

Changes
---
1) A new MixedRealityToolkit.Extensions directory is added for components that exist outside the main MRTK functionality
2) A few basic interfaces are defined to abstract out shared experience components. These will likely change long term but there are hopes that having cleaner definitions than the previous spectator view (preview) will make for a better refactoring experience
    3.1) IMatchMakingService
    3.2) IPlayerService
    3.3) INetworkingService
    3.4) ISpatialCoordinateService
3) Interface definitions are defined for needed spectator view functionality
    4.1) IMarkerDetector
    4.2) IMarkerVisual
    4.3) IPlaybackService
    4.4) IRecordingService
4) An aruco marker detector and aruco marker visual have been defined. Marker detection is conducted using a plugin built out of HoloLensForCV. TBD whether or not the licensing for this repo will suffice. (https://github.com/Microsoft/HoloLensForCV/pull/78 )
5) A platform switcher is added to the editor for toggling between HoloLens, android and iOS builds.
6) A new component, UDPBroadcastNetworkingService, acts as a simple implementation of IMatchMakingService, IPlayerService and INetworkingService. It enables devices to discover one another and send byte payloads back in forth containing information around shared spatial coordinate states
7) A new component, MarkerSpatialCoordinateService, allows devices to coordinate a shared application origin based on showing and detecting ArUco markers

NOTE: This change adds dependencies on Unity's AR Foundation, AR Core and AR Kit.


.gitignore
```diff
@@ -63,3 +63,8 @@ Assets/ThirdParty.meta
 Assets/TextMesh Pro.meta
 Assets/TextMesh Pro/
 --Version/
+
+# ================= #
+# Android generated #
+# ================= #
+*.apk
```
StephenHodgson: Shouldn't this have been left out if one were to build to a build(s) folder?

Assets/MixedRealityToolkit.Extensions/SpectatorView/.gitignore
```diff
@@ -0,0 +1,67 @@
+[Ll]ibrary/
+[Tt]emp/
+[Oo]bj/
+[Bb]uild/
+[Bb]uilds/
+[Ll]ogs/
+[Aa]pp/
+
+# Uncomment this line if you wish to ignore the asset store tools plugin
+# [Aa]ssets/AssetStoreTools*
```
StephenHodgson: Why are we adding a second ignore file?

.gitignore
chrisfromwork: I am somewhat new to building Android applications with unity. I will say that file explorers default path when launched from build settings directs to the repo root, which can cause apks to accidentally get placed out of what may be the best practice directory. Are there any expected scenarios for checking in apps?

chrisfromwork:
This is mainly a product of moving files between projects and can be changed. Most of the ignored items here can actually be removed given the mrtks .gitignore. 

Is there a suggested best practice for excluding dlls that will be built outside of this repo and have their own non-mit licenses? My main interest is not having dlls for opencv aruco detection tracked in this repo

.gitignore
StephenHodgson: It's really the devs choice. In the build window it is pretty smart about putting it in the build/builds folder so it's automatically excluded from the repository.

This change is not needed

chrisfromwork: I've gone ahead and removed this change. We will have to be diligent about not having accidental apk's get checked in by contributors. But I don't have a strong preference and can see how this change was more for my own convenience.

Assets/MixedRealityToolkit.Extensions/SpectatorView/.gitignore
chrisfromwork: It looks like my comments below did not get added to this conversation correctly. I've gone ahead and updated this .gitignore to only check that the various plugin files don't accidentally get checked in.

Railboy:
@chrisfromwork I'm looking at your first pass at defining a networking interface. I'll raise one of the points brought up in #3313

Based on our experiences the networking interface must provide a way to specify:
- who will receive a message - all targets, subset of targets, specific target, etc.
- how the message will be sent - directly to target, via central authority (if order received matters), etc.
- the importance of the message - critical messages will be re-sent until target confirmation is received while other messages may be dropped, etc.

Because this config will vary message-to-message it can't be left to the implementation. But we recognize that not every app will care about how / to whom messages are sent, and that spectator view in particular doesn't and shouldn't care about these future capabilities.

We can kick this can down the road by including an optional config object in the SendData method. It can be ignored by SpectatorView's implementation, but as implementations become more sophisticated we can start packing it with config data:

```
    public interface INetworkingService
    {
        bool SendData(byte[] data, SendArgs args = null);
        event DataHandler DataReceived;
    }
```

That's just one future-proofing approach. I'm sure there are others. Let's discuss further in #3313

chrisfromwork:
@Railboy this seems like valuable feedback that we should try and capture in our sharing architecture onenote.

Given this is a review for a feature branch, do you think not declaring this SendArgs type/interface function argument is enough to warrant waiting on completing this pull request? In one sense, my main motivation for this SpectatorView refactor is to call out the exchanges between networking and other components, and I think the interface stub accomplishes that even if it is subjected to change/causes breaking changes before getting the payload into the main development branch.

I also feel like this UDPBroadcast networking component may be too simple to ever be able to distinguish between different payload priorities. Are there any concerns with networking components ignoring parameters in a SendData argument? Would we want networking components to declare what arguments they actually support/use?

chrisfromwork:
@Railboy I will likely check this in tomorrow to get content into this feature branch if there is no update. That going said, I am on board with refactoring any of this content before moving it into the main development to better align with our long term plans.

Railboy:
@chrisfromwork Nope not at all we can hammer out these details later on. Just want to make sure they're getting brought up and discussed.

>Are there any concerns with networking components ignoring parameters in a SendData argument?

No, I would expect some simpler implementations to ignore them. An exception or warning would be sufficient IMO. Again, we can work out those details later.

provencher:
I know this is a merged PR, but I just noticed some discussions on the INetworkingService and thought it might be useful to add some notion of connect / disconnect, along with a connection object.

In my experience, it makes more sense to think in terms of a IConnection rather than an INetworkService.

This makes it easier to differentiate clients, and add extensibility in terms of host/client specific features.

chrisfromwork:
@provencher this is valuable feedback. Could you provide more context on what you would consider an IConnection? Would there be n IConnections for n clients? Would we expect there to be different networking configurations that then are surfaced as different connections?

provencher:
@chrisfromwork I've been modeling my design philosphy after my experience with Hazel networking.

Here is their connection class. They don't have an interface, and simply use an abstract class, though that isn't a hard req by any means.
https://github.com/DarkRiftNetworking/Hazel-Networking/blob/master/Hazel/Connection.cs

When a user connects, a new connection object is made, and it can be stored in a list. 
There is an event dispatcher for Connect, Disconnect, etc, which allow easy management of each connection state.

With the hazel model you can use udp, reliably and unreliably, as well as tcp.
For my own uses, I've found it practical to use combinations of transport protocols.

For instance, you can use multicast to broadcast your initial message to many locally connected clients, and then safely send missing packets over individual tcp connections. (Note multicasting is not a hazel feature, but easy enough to implement with dot net).

Another practical part of SendBytes, is a flag determining the send mode, especially in the case of UDP.
Hazel has Unreliable, Reliable, and Reliable Fragmented, where large messages are broken into smaller ones and reliably transmitted.

Issue #3295, opened 2018-12-20
vNext Proposal: Remove requirement for IMixedRealityInputHandler<T> to be an IMixedRealityInputHandler


keveleigh:
# Overview
At the moment, in order to get the Pose events from an input source, I also need to register for four other events (the bottom two are already obsolete and will be removed in a future release): 
![image](https://user-images.githubusercontent.com/3580640/50312422-2625ec80-045d-11e9-9bbd-4f3fb3fb4065.png)

Previously, the (now obsolete) `IMixedRealitySourcePoseHandler` was independent of `IMixedRealityInputHandler`, but now when porting, several other, possibly unwanted, events are required to be added.

This is related to #3168.

# Requirements

* Remove dependency that `IMixedRealityInputHandler<T>` has on `IMixedRealityInputHandler`
* Update consumers to either: specify both interfaces or remove unused OnInput* methods

# Acceptance Criteria

- [ ] Interface can be implemented using a single event for things like source pose
- [ ] Application can add support for IInput* by inheriting from `IMixedRealityInputHandler` (or other `IMixedRealityInputHandler<T>`)
- [ ] Breaking change has been documented

PR #3389, opened 2019-01-30
remove IMixedRealityInputHandler from IMixedRealityInputHandler<T>

davidkline-ms:
Overview
---
Not every class that implements `IMixedRealityInputHandler < T >` needs the methods from `IMixedRealityInputHandler`. This change removes the dependency and allows code that so desires to implement one or both without need for empty method implementations.

Changes
---
Fixes: #3295


davidkline-ms:
> is a generic supposed to be behind one of these?

yes. fixed it in the description. thanks!

PR #3390, opened 2019-01-31
Create extension service and data provider attributes

### LINKED ISSUE #3387: Add Attribute to make sure we can migrate to data provider in V3
Implement an attribute that will enable both easy migration from v2 to v3 configuration profiles as well as to surface useful default data when adding the provider (ex: configuration profile, supported platforms, etc).

davidkline-ms:
Overview
---
Created MixedRealityExtensionServiceAttribute and MixedRealityDataProviderAttribute to enable easy migration between v2 (registered services profile) and v3 (service specific profile data provider collections) as well as to enable automatic population of supported runtime platforms as well as the default profile file and path.

The appropriate attributes have been applied to all current extension services.

The runtime platform and default profile registration is a future feature that is not implemented in this PR.

Fixes #3387


Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
```diff
@@ -144,7 +144,38 @@ private static bool IsNewSession
...
+        /// <summary>
+        /// Finds the path of a directory relative to the project folder.
+        /// </summary>
+        /// <param name="directoryPathToSearch">
+        /// The subtree's root path to search in.
+        /// </param>
+        /// <param name="directoryName">
+        /// The name of the directory to search for.
+        /// </param>
+        public static bool FindRelativeDirectory(string directoryPathToSearch, string directoryName, out string path)
```
StephenHodgson: This wasn't intended to be public. It likely belongs in a utility class.

Assets/MixedRealityToolkit/Attributes/MixedRealityExensionServiceAttribute.cs
```diff
@@ -0,0 +1,74 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="runtimePlatforms">The platforms on which the extension service is supported.</param>
+        /// <param name="profilePath">The relative path to the profile asset.</param>
+        /// <param name="packageFolder">The folder to which the path is relative.</param>
+        public MixedRealityExensionServiceAttribute(
```
danhofm: typo: should read `Extension`

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
danhofm: I agree. Would `EditorProjectUtilities` be a good place for this?

davidkline-ms: that seems like a reasonable spot

Assets/MixedRealityToolkit/Attributes/MixedRealityExtensionServiceAttribute.cs
```diff
@@ -0,0 +1,74 @@
...
+using UnityEditor;
+#endif
+
+namespace Microsoft.MixedReality.Toolkit.Core.Attributes
+{
+    /// <summary>
+    /// Attribute that defines the properties of a Mixed Reality Toolkit extension service.
+    /// </summary>
+    [AttributeUsage(System.AttributeTargets.Class, AllowMultiple = false)]
+    public class MixedRealityExtensionServiceAttributeAttribute : Attribute
```
keveleigh: `MixedRealityExtensionServiceAttributeAttribute`? Two `Attributes` seems strange

davidkline-ms: that looks like a typo.... let me check

Assets/MixedRealityToolkit/Utilities/Editor/EditorProjectUtilities.cs
```diff
@@ -1,12 +1,13 @@
...
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using Microsoft.MixedReality.Toolkit.Core.Utilities.Editor.Setup;
 using UnityEditor;
 
 namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Editor
 {
     [InitializeOnLoad]
-    internal class EditorProjectUtilities
+    public class EditorProjectUtilities
```
keveleigh: Does this need to be made public? I feel like most of the scripts that need to use these calls can stay in the core assembly, but I might be overlooking some use cases.

davidkline-ms: the compiler balked when I left it as internal, presumably because other assemblies are using a public property that calls into this class.

davidkline-ms: i can also envision instances where assemblies may wish to know relative folder paths

PR #3410, opened 2019-02-05
[vNext][UX] Tooltip Examples Scene

Railboy:
Overview
---
Our basic ToolTip example scene updated with latest prefabs and manipulation scripts. Also fixes some spawning logic in ToolTipSpawner class.

tooltips


Railboy:
@cre8ivepark Can confirm that focus / press works in VR with a controller but not on HoloLens with gaze / tap.

@keveleigh I'm stumped as to why it's not working, do you have any insight?

SimonDarksideJ:
Can we test that issue outside of this scene as it sounds more like a generic click vs tap issue.
If the issue is replicatable outside of this scene, I suggest we merge this and raise a separate issue.

Railboy:
We tested using a HL V1 device with a more recent OS version. Input and focus worked fine. Agree with @SimonDarksideJ, let's merge this and open a separate issue.

PR #3455, opened 2019-02-14
Search all controller mappings before trying to assign a default interaction

luis-valverde-ms:
Search all controller mappings before trying to assign a default interaction mapping.


SimonDarksideJ:
I'm not sure this makes sense, controller mappings are not transferable between controllers.   The only transference allowed is where it's the same controller can be used on different platforms.

Just allowing ANY mapping to be used on any controller, just because it's the same hand does not make sense to me.
Please correct me if I'm misunderstood the purpose of this PR

luis-valverde-ms:
This is just addressing a bug in SetupConfiguration() where we only check the first entry of controllerMappings when looking for a mapping for the controller.

Issue #3056, opened 2018-11-02
Diagnostics placement is sporadic and displays no data


SimonDarksideJ:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
Diagnostics view is meant to display diagnostic information in the scene.  However, when enabled, the "box" is not in the view (placement of the mesh is odd) and there is no content displayed.

**To Reproduce**

* Enable diagnostics in the configuration profile
* Set Visible to True in the diagnostics profile.
* Create a new Scene
* Click Mixed Reality Toolkit > Configure
* Enter play mode.

**Expected behavior**
Diagnostics view should be in the FoV of the headset/camera with data displayed

**Actual behavior**
Placement of window is sporadic and no data is displayed

**Screenshots**
![image](https://user-images.githubusercontent.com/1793042/47917400-4c89cb80-dea1-11e8-9c8f-ef41cdc3e678.png)

This "window" was placed below and to the left of the player, single sided texture so it's only visible if the player steps back.

**Unity Editor Version**
2018.2.13f1

**Mixed Reality Toolkit Release Version**
2018.9 beta

**Additional Notes**
Tested using Seated / standing mode with no boundary

davidkline-ms:
i'll take this one. I am investigating adding additional data

davidkline-ms:
As part of this, the diagnostic window should be parented to the camera (or parent).

The intent weas to be headlocked so the data was always visible.

StephenHodgson:
> The intent weas to be headlocked so the data was always visible.

Headlocked sounds terrible lol. Can we make that configurable at least?

davidkline-ms:
> Headlocked sounds terrible lol. Can we make that configurable at least?

We *can*, though the greatest utility of this feature is to give you information on the portion of the scene that is in view (ex: as you move through the scene, the data is relevant to what you are seeing)

StephenHodgson:
I noticed that this also doesn't really put anything in the headset, but only prints the values to the editor window

PR #3457, opened 2019-02-14
Replace beta 1 diagnostics visualization with a modified VisualProfiler-Unity fork

davidkline-ms:
Overview
---
NOTE: This is a breaking change from Beta 1 as the diagnostic visualization has been completely replaced.

This change removes much of the beta 1 diagnostics system implementation and replaces it with a modified fork of the Visual Profiler from https://github.com/Microsoft/VisualProfiler-Unity.

The diagnostics profile currently contains only "Is Profiler Visible".

Changes to the forked VisualProfiler script include:

- Renaming the class to MixedRealityToolkitVisualProfiler (to avoid collisions if a customer already has the original in the project)
- Renaming the namespace to Microsoft.MixedReality.Toolkit.Services.DiagnosticsSystem (consistency and collision avoidance)
- Rename initiallyVisible to isVisible and add a read/write property to enable toggling
- Remove keyword support (now handled by demo scripts in conjunction with the Speech Commands Profile)
- Add public property to set the profiler window's parent (hierarchy organization)
- Modify LateUpdate() to show/hide the profiler window as appropriate

Fixes: #3245, #3056


davidkline-ms:
found a bug in the toggle logic... will re-open when fixed

PR #3460, opened 2019-02-15
Declare mouse controller as 'any' handedness

luis-valverde-ms:
That's what the mouse manager expects, otherwise input action mappings are not applied to mouse events.

Overview
---

Changes
---
- Fixes: # .


SimonDarksideJ:
Whereas we don't depict a "Mouse" as a device having a hand, hence it being initially set to none.  (Simply because the mouse itself as an input device cannot denote hand), I'm not opposed to this change as it's merely semantic )

PR #3464, opened 2019-02-15
Controller mapping profile and controller attribute fixes

davidkline-ms:
Overview
---
- Wrap editor specific code in the controller mapping profile in #if / #endif.
- Add "using Microsoft.MixedReality.Toolkit.Core.Extensions" when building UWP on .net backend

This does not 100% resolve .net backend build breaks. There are some changes needed in interactables to work around AppDomain usage

### DIFF
--- a/Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
+++ b/Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
@@ -6,6 +6,10 @@ using Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities;
 using System;
 using System.Linq;
 
+#if WINDOWS_UWP && !ENABLE_IL2CPP
+using Microsoft.MixedReality.Toolkit.Core.Extensions;
+#endif // WINDOWS_UWP && !ENABLE_IL2CPP
+
 namespace Microsoft.MixedReality.Toolkit.Core.Attributes
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityControllerMappingProfile.cs
```diff
@@ -19,24 +19,25 @@ namespace Microsoft.MixedReality.Toolkit.Core.Definitions.Devices
...
-
-        public static Type[] CustomControllerMappingTypes { get => (from type in ControllerMappingTypes where UsesCustomInteractionMapping(type) select type).ToArray(); }
-
         [SerializeField]
         [Tooltip("The list of controller templates your application can use.")]
         private MixedRealityControllerMapping[] mixedRealityControllerMappingProfiles = new MixedRealityControllerMapping[0];
 
         public MixedRealityControllerMapping[] MixedRealityControllerMappingProfiles => mixedRealityControllerMappingProfiles;
 
+#if UNITY_EDITOR
```
SimonDarksideJ: Have we confirmed a .NET build works and deploys without this section in the build on both Windows and Standalone @davidkline-ms ?

davidkline-ms: I have confirmed an il2pp build deploys and runs. there are other parts (interactables) of the project that are preventing successful .net backend builds.

davidkline-ms:
moving to target mrtk_developent to give a bit more time to verify all build targets (and ideally get the interactable changes in place as well)

Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
```diff
@@ -6,6 +6,10 @@
 using System;
 using System.Linq;
 
+#if WINDOWS_UWP && !ENABLE_IL2CPP
```
chrisfromwork: did we get any comments in ship room whether we would support .net? From what i understand newer versions of unity don't provide .net tooling?

davidkline-ms: The standing policy is to support the backends that Unity supports. 2018.4 will continue to support the .NET backend for 2 years following its release.

As we move to Unity 2019, we will only support il2cpp. For as long as we support Unity 2018, we will also support .NET.

keveleigh: Could this be simplified to just `#if ENABLE_DOTNET`?

davidkline-ms: I wasn't aware of that define! We should create an issue to replace all of the  #if WINDOWS_UWP && !ENABLE_IL2CPP instances with it!

PR #3471, opened 2019-02-18
Add pipeline status widget to .md

MenelvagorMilsom:
Overview
---
- Adding status widget for Azure Pipelines CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ The vNext branch is taking all the best lessons learned from the original Mixed 
 
 > Find this readme, other documentation articles and the MRTK api reference on our [developer documentation on github.io](https://microsoft.github.io/MixedRealityToolkit-Unity/). 
 
+# Build Status
+
+| Branch | Status |
+|---|---|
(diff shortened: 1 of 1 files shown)


MenelvagorMilsom:
Presently, continuous integration is disabled for this pipeline, but I can enable it as soon as required.

docfx.json
```diff
@@ -0,0 +1,39 @@
+{
```
paco-ms: You probably didn't mean to include this file as part of the PR ?
```suggestion
{remove this file}
```

MenelvagorMilsom: Yeah, this file was given to me by Bernadette. She said that it should be added to the MRTK root dir.

paco-ms: Ok. If we (reasonably) can we should try to separate the commits for the azure pipeline and those for generating documentation.

paco-ms:
Let's hold on this commit until we move the build pipeline into a public project. We want users to be able to inspect build results.

davidkline-ms:
Marking as do not merge until the pipeline changes are complete

MenelvagorMilsom:
@keveleigh @davidkline-ms I've updated the status badge to point to the newer build pipeline. This build pipeline just had it's visibility set to public so we should be ready to merge this PR.

PR #3476, opened 2019-02-19
Remove unneeded/dated artifact build instructions

chrisfromwork:
I've never needed to build artifacts for the MRTK. We should remove these steps to remove any confusion

Overview
---

Remove artifact build steps that aren't needed

Changes
---
- Fixes: # .


chrisfromwork:
@jwittner do these artifact build scripts work? I have hit some issues in the past where these don't succeed with my local setup.

jwittner:
Open to other recommendations for the docs if they’re confusing, but this process is going to be expanded on in the near future and the artifacts will be key to verifying that your changes are compliant with some of the ways the MRTK will be distributed.

chrisfromwork:
I think it's a question of whether it's needed for using the mrtk vs needed for contributing to the mrtk. I hit issues when trying to run through these steps locally, and after talking to different contributors it didn't seem like these steps were often used. I'd like to detract from hesitation in using the mrtk/remove these instructions from the landing page if these scripts are susceptible to failure and aren't needed. I'm onboard for gaining more context on what they are for. But it might be more appropriate to but the instructions in the wiki compared to on the landing page?

chrisfromwork:
Regardless of what we decide here, I greatly appreciate this effort around scripted building/package generation/additional context for validating changes. If I can be of any help for creating better documentation, let me know.

PR #3479, opened 2019-02-19
Small Performance Tweaks Round 1

Railboy:
Overview
---
First round of small performance tweaks. These were all problems we spotted in the profiler while testing. Tried to avoid breaking changes or heavy refactoring.

Changes
---
- Set new default for line steps in pointer prefabs and updated pointer inspector to give a perf warning.
- Disabled auto-sync transforms in physics settings. This helped with physics spikes while using manipulation scripts. [See docs.](https://docs.unity3d.com/ScriptReference/Physics-autoSyncTransforms.html)
- Added a new performance-oriented setting in line data providers.
- Replaced a lot of Vector3 math in very hot code, especially RayStep.
- Removed some unnecessary allocations.


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/LinePointer.cs
```diff
@@ -15,6 +15,11 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
     [RequireComponent(typeof(DistorterGravity))]
     public class LinePointer : BaseControllerPointer
     {
+        [Range(2, 50)]
```
keveleigh: Should we go ahead and start this at `1`, or is a minimum of 2 required? (thinking of cases like a straight line)

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Pointers/LinePointerInspector.cs
```diff
@@ -18,6 +18,7 @@ public class LinePointerInspector : BaseControllerPointerInspector
         private SerializedProperty lineRenderers;
 
         private bool linePointerFoldout = true;
+        private const int maxRecommendedLinecastResolution = 25;
```
keveleigh: The [Tooltip on LinePointer](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3479/files#diff-42582bc1e2308bec1f932a22d56c0488R20) recommends 20 as the max for perf reasons. Should we standardize on 20 or 25 between here and there?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/LinePointer.cs
davidkline-ms: also, is 50 still too many?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Pointers/LinePointerInspector.cs
```diff
@@ -42,12 +43,19 @@ public override void OnInspectorGUI()
             if (linePointerFoldout)
             {
                 EditorGUI.indentLevel++;
+
+                int lineCastResolutionValue = lineCastResolution.intValue;
+                if (lineCastResolutionValue > maxRecommendedLinecastResolution)
+                {
+                    EditorGUILayout.LabelField("Note: values above " + maxRecommendedLinecastResolution + " should only be used when your line is expected to be highly non-uniform.", EditorStyles.miniLabel);
```
davidkline-ms: can you define "highly non-uniform"?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/LinePointer.cs
Railboy: In 95% of cases probably yes. There are cases where 50 could be useful - eg a really weird or inventive spline pointer. The inspector does give you a warning if you set it above 25 (soon to be 20).

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Pointers/LinePointerInspector.cs
Railboy: A controller styled after a rope, whip or tentacle. Something with lots of bends that needs to 'thread the needle' when detecting collisions.

davidkline-ms:
@Railboy, what was the reason for reverting the cached collider change?

Railboy:
@davidkline-ms Caching the collider bounds was causing some unexpected behavior.

PR #3486, opened 2019-02-20
Update controller pointer line to support color again

keveleigh:
Overview
---
Thanks to #3382, we can re-add color to the controller lines!

From 
image

to

image

Changes
---
- Fixes: #3268


keveleigh:
Yeah, the pointers were originally designed to have a slight transparency on either end:

image

but I've reverted it to opaque:

image

Cameron-Micka:
Ah gotcha! If people want that back we could also switch them to be "Additive," you still pay the overdraw cost of transparency but don't have to worry about transparency sorting (but you might get overly bright pointers if they stack up on bright backgrounds).

keveleigh:
@Cameron-Micka ooh, thanks for the heads up! I'll look into that

PR #3490, opened 2019-02-21
Change WMR.RemoveController to take in an InteractionSource

keveleigh:
Overview
---
The method didn't need anything out of `InteractionSourceState`, so this removes the need to call `.Source` multiple times.

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
@@ -373,7 +373,7 @@ namespace Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality
             InteractionSourceState[] states = InteractionManager.GetCurrentReading();
             for (var i = 0; i < states.Length; i++)
             {
-                RemoveController(states[i]);
+                RemoveController(states[i].source);
             }
         }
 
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -373,7 +373,7 @@ public override void Disable()
             InteractionSourceState[] states = InteractionManager.GetCurrentReading();
             for (var i = 0; i < states.Length; i++)
             {
-                RemoveController(states[i]);
+                RemoveController(states[i].source);
```
chrisfromwork: nit: may be worth adding null check

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -478,7 +478,7 @@ private void InteractionManager_InteractionSourceDetected(InteractionSourceDetec
         /// <param name="args">SDK source updated event arguments</param>
         private void InteractionManager_InteractionSourceLost(InteractionSourceLostEventArgs args)
         {
-            RemoveController(args.state);
+            RemoveController(args.state.source);
```
chrisfromwork: nit: may be worth adding null check

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -438,16 +438,16 @@ private WindowsMixedRealityController GetController(InteractionSource interactio
         /// Remove the selected controller from the Active Store
         /// </summary>
         /// <param name="interactionSourceState">Source State provided by the SDK to remove</param>
-        private void RemoveController(InteractionSourceState interactionSourceState)
+        private void RemoveController(InteractionSource interactionSource)
```
chrisfromwork: nit: may be worth adding return if null

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
keveleigh: Since this is coming from the platform, I think it's guaranteed to not be null.
@davidkline-ms, do you happen to know if the platform might ever provide null here?

SimonDarksideJ:
@davidkline-ms just spotted this PR, I hope you tested that the controller is removed from the scene when the event is fired.
The interaction source was used to get the ID of the controller that is used to track it in the framework and hence remove it.  In the device managers, we use the Physical characteristics of the controller to track it.

keveleigh:
>The interaction source was used to get the ID of the controller that is used to track it in the framework and hence remove it.

Right, that's exactly what this PR is doing. Previously, the remove method took in a state, asked it for its source, and did nothing else with the state. Now it gets passed the source directly.

PR #3496, opened 2019-02-21
Fix air-tap in Holographic Remoting (workaround for Unity issue #1033526)

NorbertNemec:
Workaround for a blocker bug in our development workflow

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
@@ -414,14 +414,35 @@ namespace Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality
                     break;
                 case DeviceInputType.Select:
                     {
+                        bool selectPressed = interactionSourceState.selectPressed;
+
+                        // BEGIN WORKAROUND: Unity issue #1033526
+                        // See https://issuetracker.unity3d.com/issues/hololens-interactionsourcestate-dot-selectpressed-is-false-when-air-tap-and-hold
+                        // Bug was discovered May 2018 and still exists as of today Feb 2019 in version 2018.3.4f1, timeline for fix unknown
+                        // The bug only and the workaround affect only the development workflow via Holographic Remoting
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
```diff
@@ -414,14 +414,35 @@ private void UpdateTriggerData(InteractionSourceState interactionSourceState, Mi
                     break;
                 case DeviceInputType.Select:
                     {
+                        bool selectPressed = interactionSourceState.selectPressed;
+
+                        // BEGIN WORKAROUND: Unity issue #1033526
+                        // See https://issuetracker.unity3d.com/issues/hololens-interactionsourcestate-dot-selectpressed-is-false-when-air-tap-and-hold
+                        // Bug was discovered May 2018 and still exists as of today Feb 2019 in version 2018.3.4f1, timeline for fix unknown
+                        // The bug only and the workaround affect only the development workflow via Holographic Remoting
+                        if (UnityEngine.XR.WSA.HolographicRemoting.ConnectionState == UnityEngine.XR.WSA.HolographicStreamerConnectionState.Connected
```
keveleigh: Looks like this is also an issue when using Simulate in Editor.
Do you happen to know if it's possible to detect when we're in that state? It's still reporting as a `Hand`, but the `ConnectionState` isn't connected.
I wonder if we can change this line to just `Application.isEditor`, since if we're getting a hand over this API in the editor, I'm pretty sure we're either remoting or simulating.

NorbertNemec: The question inspires me to an alternative logic: instead of checking for HolographicRemoting we could simply check of Hand - assuming that the only supported "pressed" gesture of a hand is the air-tap, we can then safely assume that it should be anyPressed==selectPressed and correct accordingly. I'll change the patch, keeping a similar set of asserts for all the assumptions.

NorbertNemec: Just pushed a generalization that should cover Simulation, only assuming that Hand can never emulate any buttons except select.

PR #3499, opened 2019-02-21
Fix GettingStarted.md List indices

zevisert:
Overview
---
Had a few repeated list levels: 2 and 5 appeared more than once

Changes
---
- Documentation only


keveleigh:
Thanks for the contribution!

chrisfromwork:
Thanks!

PR #3502, opened 2019-02-21
MixedRealityToolkit Profile Inspector

Railboy:
Overview
---
Draws a sub-inspector for the active MRTK configuration profile in the MixedRealityToolkit singleton inspector.

I've clicked the MixedRealityToolkit object in the scene hundreds of times and every time I'm still surprised when I don't immediately see the configuration profile. Having to click that first field and exit the scene heirarchy to view the profile is an uncomfortable mental switch.

This will give first-time devs a better first impression of the MRTK and help make the profile system more discoverable.

inspectorprofile


keveleigh:
@davidkline-ms and I have been discussing for a while seeing if there's a good way to make all profiles display / fold out like this. Helps prevent constant context switching between profiles and makes things more visible, but it could also become cluttered. Fold outs would help with that (though it probably makes sense for the base layer to just display, like this)

davidkline-ms:
This is beautiful and an elegant change!  Taking for beta 2 :)

Railboy:
Foldouts for sub-profiles are next on my list. I just needed to scratch this itch today.

davidkline-ms:
> Foldouts for sub-profiles are next on my list. I just needed to scratch this itch today.

Let's do this one together since we need to be able to persist the foldout setting between sessions.

PR #3506, opened 2019-02-22
Read the default profile and default platforms from the ExtensionServiceAttribute

keveleigh:
Overview
---
This PR adds the ability to have the registered service providers profile to auto-set its values based on the already set values in the MixedRealityExtensionServiceAttribute.

To start, it tries to load the default profile from an asset path, and it sets the supported platforms automatically.

(video didn't pick up the service type dropdown)

service_provider


StephenHodgson:
Very cool idea guys!

Issue #3514, opened 2019-02-23
[MRTK v2 Beta 2] Focus provider error


cre8ivepark:
## Overview
Error message related to FocusProvider. 
'Microsoft.MixedReality.Toolkit.Services.InputSystem.FocusProvider' is missing the class attribute 'ExtensionOfNativeClass'!

![2019-02-22 16_52_32-window](https://user-images.githubusercontent.com/13754172/53279155-eaa8d800-36c2-11e9-818a-2aa1805fca33.png)
![2019-02-22 16_52_25-window](https://user-images.githubusercontent.com/13754172/53279156-ec729b80-36c2-11e9-8f45-201d712467b7.png)

## Expected Behavior
No errors.

## Steps to reproduce
1. Clone MRTK repo. Switch to mrtk_release branch.
2. Open in Unity
3. Open BoundingBoxExample scene

## Unity Editor Version
2018.3.2f1

## Mixed Reality Toolkit Release Version
mrtk_release branch

cre8ivepark:
It seems like this scene has old '**MixedRealityPlayspace**' object. Other scenes do not have 'FocusProvider' in the Camera.

keveleigh:
Removing the FocusProvider from the scene will resolve this. Looks like one of the example scenes wasn't updated to the new pattern.

FocusProvider is now set in the input system profile:
![image](https://user-images.githubusercontent.com/3580640/53279212-53905000-36c3-11e9-98d6-f55360b67306.png)

cre8ivepark:
Yes, it would be great to have a pass on all example scenes.

cre8ivepark:
I did a quick review on the scene and found the same issues in these scenes:
InspectorFieldsGallery
TooltipExamples (This scene has old 'MixedRealityManager' with 'Body' object.)

keveleigh:
Opened PR: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3515

davidkline-ms:
fixed in mrtk_development

sundarmuru:
The 'BoundingBox' example scene also has the same issue.

PR #3515, opened 2019-02-23
Fix example scenes setup around FocusProvider

keveleigh:
Overview
---
Some of the example scenes had a stale setup pre-#3113

Changes
---
- Fixes: #3514


sundarmuru:
The 'BoundingBox' example scene has also the same issue.

PR #3533, opened 2019-02-27
Mrtk Inspector Upgrade (Part 2)

Railboy:
Overview
---
Adds sub-profile inspector rendering. Whenever a profile field is drawn in a default MRTK configuration profile a dropdown is created which lets the user display the contents of the field's profile. This allows users to quickly access and modify all active profiles in a project when they select the MixedRealityToolkit object in the scene.

Configuration profiles are made aware of when they are being drawn as sub-profiles and omit logos and 'Back' buttons in this state.

Dropdown states are stored in editor prefs and will persist across sessions.

Problems
----
- If every nested sub-profile from the main configuration profile is open at once the inspector can become a bit sluggish.
- There's no check for recursive profile inspectors being drawn. I can't imagine what scenario would result in this occurring but in theory it's a problem.

subprofile


davidkline-ms:
That image looks WONDERFUL!

radicalad:
THANK YOU

StephenHodgson:
Seems cluttered

StephenHodgson:
I think ideally we would rather want to show this in its own window that isn't bound to an inspector.

Kinda like how Unity's new preferences window works.

That was my initial plan/idea anyway.

PR #3537, opened 2019-02-27
Bezier Line Data Provider

Railboy:
Overview
---
We've often found ourselves in need of a simpler / more lightweight alternative to the Spline data provider. This PR adds a Bezier line data provider and inspector.

bezier

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Lines/DataProviders/BezierDataProvider.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Lines/DataProviders/BezierDataProvider.cs
@@ -31,6 +31,10 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Lines.DataProviders
         [Header("Bezier Settings")]
         [SerializeField]
         private BezierPointSet controlPoints = new BezierPointSet(0.5f);
+        [Tooltip("If true, control points 2 and 3 will be transformed relative to points 1 and 4 respectively")]
+        [SerializeField]
+        private bool useLocalTangentPoints = false;
+        private Vector3 localOffset;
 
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit/Utilities/Lines/DataProviders/BezierDataProvider.cs
```diff
@@ -0,0 +1,121 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities;
+using System;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Lines.DataProviders
+{
+    public class BezierDataProvider : BaseMixedRealityLineDataProvider
```
davidkline-ms: Please decorate the class with a DataProvider attribute. Let me know if you need help deciding what values to use.  Thanks!

Railboy: The lines aren't associated with a service so I'm not sure what to use for serviceInterfaceType.
(Also would all the other line data providers need this attribute as well? Currently none of them do.)

davidkline-ms: Good point. Let's chat offline and figure out a good solution.

davidkline-ms: per our conversation, you shouldn't need the attribute if the class does not implement IMixedRealityDataProvider.

PR #3548, opened 2019-02-28
glTF Utilities based off of built in Unity Serialization

StephenHodgson:
Overview
---
Adds support for glTF model importing as standard model prefabs/objects.

image

# glTF
Unity [glTF 2.0](https://github.com/KhronosGroup/glTF) serialization using `UnityEditor.JsonUtility`.

- [x] Import Asset
    - [x] Import asset at runtime
    - [x] Custom asset importer for editor
    - [x] .gltf file extension
    - [x] .glb file extension
    - [x] Json string
    - [x] Construct GameObject from GltfObject Data
    - [x] Apply texture to GameObject
    - [x] Apply PBR Settings to material
    - [x] Use the MRTK Standard Shader
    - [ ] Asynchronous texture loading
    - [ ] Import skins
    - [ ] Import Animations
    - [ ] Import Cameras
    - [ ] Apply extension properties to imported asset
- [ ] Export Asset
    - [ ] Export asset at runtime
    - [ ] Export asset from editor
    - [ ] .gltf file extension
    - [ ] .glb file extension

# FAQ
To load gltf or glb files at runtime without a custom shader, you'll need to make sure that the standard shader is included in your final build with each of the required variants compiled and located either in your resources folder, or used in your scene. For more information check out the [Unity Docs explaining how to access shader properties at runtime](https://docs.unity3d.com/Manual/MaterialsAccessingViaScript.html).

## To build a glTF-Object from json:
```
GltfObject gltfObject = GltfUtility.GetGltfObjectFromJson(json);
```

## To build a glTF-Object from uri:
```
GltfObject gltfObject = gltfObject = await GltfUtility.ImportGltfObjectFromPathAsync(uri);
```

## To Construct a unity GameObject from glTF-Object:
```
await gltfObject.ConstructAsync();
```

## To access built Gltf GameObject scene reference
```
gltfObject.GameObjectReference
```


StephenHodgson:
The demo scene needs a bit of love, but should load after 5 seconds.

chrisfromwork:
I will spend some more time to thoroughly look through this. One quick question: Do we know where these glb and gltf files came from? Are they something that we can checkin while maintaining the MIT license?

StephenHodgson:
Pretty sure they came from Microsoft, but I included the same license from that was found in the UnityGLTF library over at Khronos.

StephenHodgson:
If you guys would like to provide better model samples for testing, please open a pull request with the appropriate file changes.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfAccessorSparseIndices.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.MixedReality.Utilities.Gltf.Schema
```
keveleigh: Should this be namespace Microsoft.MixedReality **.Toolkit.** Utilities.Gltf.Schema?

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/README.md
```diff
@@ -0,0 +1,9 @@
+# Boom Box
+## Screenshot
+
+![screenshot](screenshot/screenshot.jpg)
```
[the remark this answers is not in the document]
StephenHodgson: This file was copied over from the Khronos model folder.

Should I remove it completely, or add the screenshot?

Kinda goes back to @chrisfromwork's question of licensing.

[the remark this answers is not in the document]
keveleigh: Looks like there's an updated license for these Microsoft-contributed models, putting them into the public domain: https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/BoomBox/README.md

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+            {
+                Debug.LogError("Uri is not valid.");
+                return null;
+            }
+
+            GltfObject gltfObject;
+            bool isGlb = false;
+            bool loadAsynchronously = Application.isPlaying;
+
+            if (Application.isPlaying) { await BackgroundThread; }
```
StephenHodgson:
```suggestion
            if (loadAsynchronously) { await BackgroundThread; }
```

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/README.md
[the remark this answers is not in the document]
StephenHodgson: That repo looks pretty sweet, wish I knew about it sooner!

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferView.cs
```diff
@@ -0,0 +1,48 @@
...
+        public int byteStride = -1;
+
+        /// <summary>
+        /// The target that the WebGL buffer should be bound to.
+        /// All valid values correspond to WebGL enums.
+        /// When this is not provided, the bufferView contains animation or skin data.
+        /// </summary>
+        public GltfBufferViewTarget target = GltfBufferViewTarget.None;
+
+        public GltfBuffer Buffer { get; internal set; }
```
chrisfromwork: There are a lot of public properties with internal setters like this Buffer throughout the schema namespace. Do we want to provide comment summaries for them since they are public?

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ExportGltf.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Serialization
+{
+    public class ExportGltf
+    {
+        // TODO Implement
```
chrisfromwork: nit: it may make more sense to add this class later so folks don't accidentally see this class name and assume the functionality exists

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfChunkType.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Serialization
+{
+    public enum GltfChunkType : uint
+    {
+        Json = 0x4e4f534a,
```
chrisfromwork: Similar to schema enums, this seems like another location where a summary call out for the value could help troubleshoot in any future regressions

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfConversions.cs
```diff
@@ -0,0 +1,444 @@
...
+
+            if (convert)
+            {
+                axes = Vector3.Scale(axes, CoordinateSpaceConversionScale) * -1.0f;
+            }
+
+            return new Quaternion(axes.x, axes.y, axes.z, quaternionArray[3]);
+        }
+
+        public static float[] SetQuaternionValue(this Quaternion quaternion, bool convert = true)
```
chrisfromwork: nit: it may be worth adding summaries to these get/sets calling out that a get converts from gltf to unity coordinate space wheras a set converts from unity to gltf

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfAssetImporter.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEditor.Experimental.AssetImporters;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Serialization.Editor
+{
+    [ScriptedImporter(1, "gltf")]
+    public class GltfAssetImporter : ScriptedImporter
```
chrisfromwork: big fan of these

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
```diff
@@ -0,0 +1,92 @@
...
+                        gltfTexture.Texture.name = textureName;
+                    }
+
+                    context.AddObjectToAsset(textureName, gltfTexture.Texture);
+                }
+                else
+                {
+                    if (!gltfTexture.Texture.isReadable)
+                    {
+                        var textureImporter = AssetImporter.GetAtPath(path) as TextureImporter;
```
chrisfromwork: probably should check for null here

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            if (reImport)
+            {
+                var importer = AssetImporter.GetAtPath(context.assetPath);
+                importer.SaveAndReimport();
+                return;
+            }
+
+            for (var i = 0; i < gltfAsset.GltfObject.meshes.Length; i++)
```
chrisfromwork: It looks like meshes aren't always explicitly created for the class. Do you know if its possible for this public field to end up null when deserializing? if so, we should probably have a null check for meshes

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                GltfMesh gltfMesh = gltfAsset.GltfObject.meshes[i];
+
+                string meshName = string.IsNullOrWhiteSpace(gltfMesh.name) ? $"Mesh_{i}" : gltfMesh.name;
+
+                gltfMesh.Mesh.name = meshName;
+                context.AddObjectToAsset($"{meshName}", gltfMesh.Mesh);
+            }
+
+            foreach (GltfMaterial gltfMaterial in gltfAsset.GltfObject.materials)
```
chrisfromwork: may also need a check for materials being null based on the above comment

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
```diff
@@ -0,0 +1,92 @@
...
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Serialization.Editor
+{
+    public static class GltfEditorImporter
+    {
+        public static async void OnImportGltfAsset(AssetImportContext context)
+        {
+            var gltfAsset = (GltfAsset)ScriptableObject.CreateInstance(typeof(GltfAsset));
+            var importedObject = await GltfUtility.ImportGltfObjectFromPathAsync(context.assetPath);
+
+            gltfAsset.GltfObject = importedObject;
```
chrisfromwork: ImportGltfObjectFromPathAsync can return null, in which case we should probably return prior to any of this logic

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+                return null;
+            }
+
+            GltfObject gltfObject;
+            bool isGlb = false;
+            bool loadAsynchronously = Application.isPlaying;
+
+            if (loadAsynchronously) { await BackgroundThread; }
+
+            if (uri.Contains(".gltf"))
```
chrisfromwork: may want to make sure that this .gltf extension and the below .glb extension are at the end of the file. I'm not sure if youd ever get handed a .gltf.meta file, but you could have a more robust check here

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+                return null;
+            }
+
+            gltfObject.Uri = uri;
+            int nameStart = uri.Replace("\\", "/").LastIndexOf("/", StringComparison.Ordinal) + 1;
+            int nameLength = uri.Length - nameStart;
+            gltfObject.Name = uri.Substring(nameStart, nameLength).Replace(isGlb ? ".glb" : ".gltf", string.Empty);
+
+            gltfObject.LoadAsynchronously = loadAsynchronously;
+            await gltfObject.ConstructAsync();
```
chrisfromwork: nit: it feels a bit misleading to await a function call if you are attempting to setup true synchronous behavior. I guess in most use cases a developer would want the created gltf object so they will have to await this function call regardless. But it may be worth brainstorming whether theres a friendly synchronous vs asynchronous function definition that we could construct

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+        {
+            var gltfObject = JsonUtility.FromJson<GltfObject>(jsonString);
+
+            for (int i = 0; i < gltfObject.extensionsRequired?.Length; i++)
+            {
+                var extensionsRequired = GetGltfExtensionObjects(jsonString, gltfObject.extensionsRequired[i]);
+
+                foreach (var extensionRequired in extensionsRequired)
+                {
+                    // TODO Update this after KHR_materials_pbrSpecularGlossiness extension is supported
```
chrisfromwork: is this a product of the gltf codebase this was based off of? I would prefer not to check in commented out code/we should probably Log warnings here if they aren't supported.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+            }
+
+            var meshPrimitiveAttributes = GetGltfMeshPrimitiveAttributes(jsonString);
+
+            for (int i = 0; i < gltfObject.meshes.Length; i++)
+            {
+                for (int j = 0; j < gltfObject.meshes[i].primitives.Length; j++)
+                {
+                    gltfObject.meshes[i].primitives[j].Attributes = JsonUtility.FromJson<GltfMeshPrimitiveAttributes>(meshPrimitiveAttributes[0]);
+                    meshPrimitiveAttributes.Remove(meshPrimitiveAttributes[0]);
```
chrisfromwork: nit: I'd prefer if meshPrimitiveAttributes was cleared after looping compared to making these continual remove calls. It's not super intuitive whats going on by just always referencing the 0 index item

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+                        Debug.LogWarning($"Unsupported Extension: {gltfObject.extensionsUsed[i]}");
+                    }
+                }
+            }
+
+            var meshPrimitiveAttributes = GetGltfMeshPrimitiveAttributes(jsonString);
+
+            for (int i = 0; i < gltfObject.meshes.Length; i++)
+            {
+                for (int j = 0; j < gltfObject.meshes[i].primitives.Length; j++)
```
chrisfromwork: we should have some sort of check that we never try and access an item in the meshPrimitiveAttributes that doesn't exist. Even though these items are based on the same logic, some sort of regression could occur where the meshPrimitiveAttributes size isn't equal to the sum of the mesh primitive lengths.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+        /// <param name="glbData">Raw glb byte data.</param>
+        /// <returns><see cref="GltfObject"/></returns>
+        /// <remarks>Returned <see cref="GltfObject"/> still needs to be initialized using <see cref="ConstructGltf.ConstructAsync"/>.</remarks>
+        public static GltfObject GetGltfObjectFromGlb(byte[] glbData)
+        {
+            const int stride = sizeof(uint);
+
+            var magicNumber = BitConverter.ToUInt32(glbData, 0);
+            var version = BitConverter.ToUInt32(glbData, stride);
+            //var length = BitConverter.ToUInt32(glbData, stride * 2);
```
chrisfromwork: if this code isn't used/needed we should remove it

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+                return null;
+            }
+
+            if (version != 2)
+            {
+                Debug.LogError("Glb file version mismatch! Glb must use version 2");
+                return null;
+            }
+
+            var chunk0Length = BitConverter.ToUInt32(glbData, stride * 3);
```
chrisfromwork: its probably worth calling out what chunk0 and chunk1 contain here as well as what influences their sizes in a comment.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,370 @@
...
+
+        /// <summary>
+        /// Get a collection of glTF Extensions using the handle provided.
+        /// </summary>
+        /// <param name="jsonString">The json string to search.</param>
+        /// <param name="handle">The handle to look for.</param>
+        /// <returns>A collection of snippets with the json string that defines the object.</returns>
+        private static Dictionary<string, string> GetGltfExtensionObjects(string jsonString, string handle)
+        {
+            // Bug: sometimes name isn't always before extension declaration
```
chrisfromwork: what frequency have you seen this as an issue?

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ConstructGltf.cs
```diff
@@ -0,0 +1,576 @@
...
+                joint0Accessor = gltfObject.accessors[meshPrimitive.Attributes.JOINTS_0];
+                joint0Accessor.BufferView = gltfObject.bufferViews[joint0Accessor.bufferView];
+                joint0Accessor.BufferView.Buffer = gltfObject.buffers[joint0Accessor.BufferView.buffer];
+            }
+
+            if (gltfObject.LoadAsynchronously) await Update;
+
+            var mesh = new Mesh
+            {
+                indexFormat = vertexCount > 65535 ? IndexFormat.UInt32 : IndexFormat.UInt16,
```
chrisfromwork: nit: add constant calling out what 65535 is. If its a max there should be a type max definition that should be used instead.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ConstructGltf.cs
```diff
@@ -0,0 +1,576 @@
...
+            GltfAccessor textCoord2Accessor = null;
+            GltfAccessor textCoord3Accessor = null;
+            GltfAccessor colorAccessor = null;
+            GltfAccessor indicesAccessor = null;
+            GltfAccessor tangentAccessor = null;
+            GltfAccessor weight0Accessor = null;
+            GltfAccessor joint0Accessor = null;
+            int vertexCount = 0;
+
+            if (meshPrimitive.Attributes.POSITION >= 0)
```
chrisfromwork: it feels like we could iterate over meshPrimitives.Attributes values and stub out whatever custom items are needed (like vertex count case by case). right now whats going on below is legible but makes this function super long

chrisfromwork: may need some sort of dictionary accessor for the later assignment

chrisfromwork:
I'm not sure how to monitor/locate this pull request without assigning myself. if theres a better mechanism for monitoring changes, let me know.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferView.cs
davidkline-ms: as much as I prefer doc comments, if this is khronos code, adding the summaries may make it harder to integrate updates, thoug

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ConstructGltf.cs
davidkline-ms: UInt16.MaxValue may be a good choice

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferView.cs
chrisfromwork: I don't have a strong preference on this one, but would want to aid developers if the worth outweighs the maintenance costs. @StephenHodgson what are your thoughts?

StephenHodgson: No Khronos code, this is a buffer to ease access to the data.

If anything the fields represent the data structure of the glTF standard.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ExportGltf.cs
StephenHodgson: Would you rather me throw an unimplemented exception?

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
StephenHodgson: safe cast needs no null check, right?

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferView.cs
chrisfromwork: Should we add a summary since these items have public getters?

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
StephenHodgson: .EndsWith might be better

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
davidkline-ms: if the cast cannot be completed as requested, the result will be null. there should be a check to avoid potential NullReferenceException

davidkline-ms: an alternate check would be

if (!(AssetImporter.GetAtPath(path) is TextureImporter)) { /* handle error case */ }

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ExportGltf.cs
chrisfromwork: I would suggest opening an issue/feature request after we get in the bulk gltf importer code. For now, I'd prefer if this class wasn't added so we don't generate any confusion around what we are intending to support. I would appreciate getting a gltf exporter created in the MRTK, but it may be a lower priority than other features/issues.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferView.cs
keveleigh: The existing summary tagged properties appear to match what's in the Khronos repo, while this undocumented part is new: https://github.com/KhronosGroup/UnityGLTF/blob/master/GLTFSerialization/GLTFSerialization/Schema/BufferView.cs#L44

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
StephenHodgson: `LoadAsynchronously` is internal. External consumers will not get this option. It's only there to aid in the scripted asset import

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: yeah that sounds great

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
StephenHodgson: warnings are logged below

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
chrisfromwork: Yeah, i don't think a failed cast throws here. It may be equivalent to dynamic cast in c++

davidkline-ms: it wont throw (except in very specific cases), but it will return null.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/as

StephenHodgson: Ah pattern matching. C# 8 features. Love them.

StephenHodgson: It's safe enough that I know exactly what the path for the file is and I know it'll be a texture importer? It's like casting for the inspector window. We know that Unity is smart enough that it knows what the inspector type it's defined for is the type that it expects.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
StephenHodgson: It's part of the glTF standard. It's how they defined it in the glb data.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
StephenHodgson: It's more or less a problem with some exporters that don't adhere to the glTF schema.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
chrisfromwork: It's probably a question of how defensive of code we want in the MRTK. It's likely safe to assume that we would never hit a null texture importer here. Considering its used in the editor, a failed import probably doesn't break app scenarios requiring extremely high reliability. Is the main preference to not add a check here around code cleanliness compared to reliability?

StephenHodgson: I'll put it this way. If the path was invalid, then we'd never get to the asset importer here anyway.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: We may actually benefit from validating the glbData payload/size given this length value.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: After looking at this longer, it seems like a developer can infer what's contained in the glbData by looking further down in the parsing steps. If this header, chunk, chunk scheme was more complicated it may be worth calling out what all the different components contain, but it seems reasonable that this amount of detail isn't needed.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: Do we generate a valid game object if extensions are declared before the name? It's probably worthwhile calling something out here that the name isn't always before the extension/developers could encounter issues, but we may as well provide more information/avoid making things look like a bug in the MRTK if they're a bug in the exporter:

// Assumption: This code assumes that a name is declared before extensions in the glTF schema. This may not work for all exporters. Some may fail to adhere to the glTF schema.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
davidkline-ms: If the path can be invalid, can it also be pointing to a non-TextureImporter?

StephenHodgson: If the path was invalid, then the whole glTF structure would be invalid. (Or the glTF object would fail to load bc it's missing a resource). This can only ever be a Texture Importer. If you can prove that it could be otherwise, then I'll be willing to update it.

davidkline-ms: The security dev/tester in me is always expecting a custom crafted payload that triggers an exploit. I always fall on the side of validation over assuming the data is good.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
StephenHodgson: Correct, we should be, but I didn't have time to implement it.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ExportGltf.cs
StephenHodgson: Remaining todo items are listed in the description of the pull request.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
StephenHodgson: lmao, well then that'd be a Unity security bug and above my paygrade.

The Unity developer in me knows we'll always have a texture importer.

davidkline-ms: Since we are providing a platform, we should protect the application from corrupt data streams, intentional or otherwise. A null check feels prudent here.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ConstructGltf.cs
chrisfromwork: picking up this fix

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: picking up this fix

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: picking up this fix

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
StephenHodgson: prove to me it can be null, and I'll add the check.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ConstructGltf.cs
chrisfromwork: Considering that Attributes seems to be a direct deserialized class from the json payload, i can see why this code looks this way/why its a bit harder to change. We may as well keep this this way for now/disregard my suggestions.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferViewTarget.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    public enum GltfBufferViewTarget
+    {
+        None = 0,
+        ArrayBuffer = 34962,
```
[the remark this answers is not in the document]
StephenHodgson: Likely the schema info found at https://github.com/KhronosGroup/glTF/tree/master/specification/2.0/schema

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfComponentType.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    public enum GltfComponentType
+    {
+        Byte = 5120,
```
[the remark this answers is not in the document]
StephenHodgson: Likely found here:
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0/schema

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferView.cs
StephenHodgson: Likely should be the description found in the schema
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0/schema

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfChunkType.cs
StephenHodgson: Defined in the schema https://github.com/KhronosGroup/glTF/tree/master/specification/2.0/schema

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ExportGltf.cs
chrisfromwork: Are there any specific todo items that you think need to be completed before we check in this review? Do unchecked todos get translated into issues?

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
chrisfromwork: picking up this fix

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
chrisfromwork: picking up this fix

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
chrisfromwork: picking up this fix

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: picking up this fix

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: picking up this fix

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: picking up this fix, looks like meshPrimitiveAttributes is declared on the function stack so there's no reason to clear

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: We should probably call out which specific todo items we want before checking in. Items required before checking in we should fix/uncomment. Items not required before checking in I would prefer to have stubbed out with warnings if at all calling out specific issues filed for the repo.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ExportGltf.cs
StephenHodgson: Good question. 

Top list items: 
- MRTK Standard Shader Support
- Async Texture loading

The rest could be in a single task or broken up into tasks.

StephenHodgson:
Still need to update the demo scenes to work on a final built app on device. I should have that finished by this evening.

StephenHodgson:
@chrisfromwork using the latest changes seems to be producing strange results with the new mrtk shader.
image

StephenHodgson:
I't also now crashing when I import the gltf models (Tried to import the boom box)

StephenHodgson:
Okay so interesting thing, I was trying to use the new models from the https://github.com/KhronosGroup/glTF-Sample-Models but it seems those don't work?

Checked in on the diff from the assets. Seems like they've updated the formatting? Likely need to add the default values to the classes. I had them at -1 to show as "Uninitialized" but it seems they expect some of these to always be the default.
image

StephenHodgson:
Found a fix for the import crash. I think it was something I introduced when I was attempting to ensure that the import from path method was being called from the main thread.

StephenHodgson:
Okay last thing to do is get the glTF demo loading scene to work on device. The glb scene works already by downloading the model from the Khronos model GitHub repository.

The glb scene is a good demo for showing how to download and construct a glTF model from a remote resource location. Both demo scenes could use some TLC

chrisfromwork:
We have some follow up/polish tasks around the mrtk standard shader. I hope that any changes would be additive compared to breaking here. One main thing was the emissive texture that exists in the standard shader does not exist in the mrtk standard shader. I'll take a specific look at the boom box and will screen shot the inspectors for both the standard and mrtk standard shader. Hopefully then we can get guidance from @Cameron-Micka on what improvements can be made.

Do you think this is in a stable state/ready for review?

StephenHodgson:
Should be ready for review now

StephenHodgson:
@chrisfromwork what's left on this?

chrisfromwork:
@StephenHodgson https://github.com/StephenHodgson/MixedRealityToolkit-Unity/pull/46
After we get in this review we should be ready to go.

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## License Information
```
davidkline-ms: Can we create models of our own, Perhaps using Paint3D, to avoid additional licensing?

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/Lantern/README.md
```diff
@@ -0,0 +1,10 @@
+# Lantern
+
+## Screenshot
+
+![screenshot](screenshot/screenshot.jpg)
+
+## License Information
```
davidkline-ms: Same question re: the boombox model.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfAnimationChannelPath.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// The name of the node's TRS property to modify, or the weights of the Morph Target it istantiates.
```
davidkline-ms: Can we get an expansion of "TRS" to help customers better understand what it is?

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfAnimationChannelTarget.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// The index of the node and TRS property that an animation channel targets.
```
davidkline-ms: Same TRS request here.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferViewTarget.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// The target that the GPU buffer should be bound to.
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/bufferView.schema.json
+    /// </summary>
+    public enum GltfBufferViewTarget
+    {
+        None = 0,
```
davidkline-ms: Can we get summary tags for these enum values? Would help customers know what they should use,

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfCameraType.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// Specifies if the camera uses a perspective or orthographic projection.
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/camera.schema.json
+    /// </summary>
+    public enum GltfCameraType
+    {
+        perspective,
```
davidkline-ms: These enum values should be capitalized. Also, can we get summary tags for them?

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfChildOfRootProperty.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    [Serializable]
+    public class GltfChildOfRootProperty : GltfProperty
+    {
+        /// <summary>
+        /// The user-defined name of this object.
+        /// This is not necessarily unique, e.g., an accessor and a buffer could have the same name,
+        /// or two accessors could even have the same name.
+        /// </summary>
+        public string name;
```
davidkline-ms: Can we capitalize the field?

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfComponentType.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/accessor.schema.json:componentType <para/>
+    /// The datatype of components in the attribute.  All valid values correspond to WebGL enums.
+    /// The corresponding typed arrays are 'Int8Array', 'Uint8Array', 'Int16Array', 'Uint16Array', 'Uint32Array', and 'Float32Array', respectively.
+    /// 5125 (UNSIGNED_INT) is only allowed when the accessor contains indices, i.e., the accessor is only referenced by 'primitive.indices'.
+    /// </summary>
+    public enum GltfComponentType
+    {
+        Byte = 5120,
```
davidkline-ms: Can we please get summary tags for these enum values?

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfDrawMode.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// The type of primitives to render.
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/mesh.primitive.schema.json
+    /// </summary>
+    public enum GltfDrawMode
+    {
+        Points = 0,
```
davidkline-ms: Please add summary tags for these enum values

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfImage.cs
```diff
@@ -0,0 +1,45 @@
...
+    public class GltfImage : GltfChildOfRootProperty
+    {
+        #region Serialized Fields
+
+        /// <summary>
+        /// The uri of the image.  Relative paths are relative to the .gltf file.
+        /// Instead of referencing an external file, the uri can also be a data-uri.
+        /// The image format must be jpg, png, bmp, or gif.
+        /// </summary>
+        public string uri;
```
davidkline-ms: The MRTK coding convention captializes public fields.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfInterpolationType.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// Interpolation algorithm.
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/animation.sampler.schema.json
+    /// </summary>
+    public enum GltfInterpolationType
+    {
+        LINEAR,
```
davidkline-ms: Can we add summary tags? Also, our coding convention does not use all caps names.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMagnificationFilterMode.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// Magnification filter mode.
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/sampler.schema.json
+    /// </summary>
+    public enum GltfMagnificationFilterMode
+    {
+        None = 0,
```
davidkline-ms: Can we get summary tags here?

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMaterial.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// https://github.com/KhronosGroup/glTF/tree/master/specification/2.0/schema
+    /// </summary>
+    [Serializable]
+    public class GltfMaterial : GltfChildOfRootProperty
+    {
+        /// <summary>
+        /// A set of parameter values that are used to define the metallic-roughness
+        /// material model from Physically-Based Rendering (PBR) methodology.
+        /// </summary>
+        public GltfPbrMetallicRoughness pbrMetallicRoughness;
```
davidkline-ms: Per coding conventions, public fields should be capitalized.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMeshPrimitiveAttributes.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Common mesh primitive attributes.
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/mesh.primitive.schema.json
+    /// </summary>
+    /// <remarks>
+    /// Application specific semantics are not supported
+    /// </remarks>
+    [Serializable]
+    public class GltfMeshPrimitiveAttributes
+    {
+        public int POSITION = -1;
```
davidkline-ms: Should update capitalization (ex: POSITION -> Position).

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMinFilterMode.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// Minification filter mode.
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/sampler.schema.json
+    /// </summary>
+    public enum GltfMinFilterMode
+    {
+        None = 0,
```
davidkline-ms: Please ass summary tags.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfObject.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        /// <summary>
+        /// Names of glTF extensions required to properly load this asset.
+        /// </summary>
+        public string[] extensionsRequired;
+
+        /// <summary>
+        /// An array of accessors. An accessor is a typed view into a bufferView.
+        /// </summary>
+        public GltfAccessor[] accessors;
```
davidkline-ms: please captialize public fields (throughout)

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfWrapMode.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Schema
+{
+    /// <summary>
+    /// Texture wrap mode.
+    /// https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/sampler.schema.json
+    /// </summary>
+    public enum GltfWrapMode
+    {
+        None = 0,
```
davidkline-ms: please add summary tags

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ColliderType.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Gltf.Serialization
+{
+    public enum ColliderType
+    {
+        None,
```
davidkline-ms: please add summary tags

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
davidkline-ms: It still feels prudent to check for null to prevent possible app crash / failure case with no developer information.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfAnimationChannelPath.cs
chrisfromwork: This is a direct copy from khronos gltf schema. I'd rather stay consistent with their documentation

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfAnimationChannelTarget.cs
chrisfromwork: This is a direct copy from khronos gltf schema. I'd rather stay consistent with their documentation

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/README.md
chrisfromwork: What motivations are there for not reusing this?

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/Lantern/README.md
chrisfromwork: What motivations are there for not using this?

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMinFilterMode.cs
chrisfromwork: Summary already contains enum definition location. It will be more up to date than comments here.

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/README.md
davidkline-ms: it's a dependency that requires license registration for us to ship. if we create our own test models, there is no extra license validation step before MRTK ships

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfCameraType.cs
chrisfromwork: We already call out where these enums are declared. individual enum items don't look to have descriptions from khronos

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfImage.cs
StephenHodgson: I'm very well aware of the coding conventions. I wrote them.

We do not when they're supposed to be serialized by the json serializer.

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/README.md
keveleigh: I'm not sure we have to register things that are public domain.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMeshPrimitiveAttributes.cs
StephenHodgson: Again, this matches the schema per the json serializer constraints. They need to match the casing or it will fail.

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/README.md
keveleigh: And being provided by Microsoft likely helps. I can follow up on this.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/Importers/GltfEditorImporter.cs
StephenHodgson: We've [already established that it's literally impossible for this to be null](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3548/files/e03899bf6c018096a55c1c76c72b5c2fe194db73#diff-3e8ceaf1e38fdaa605a993db92b04f9aR31).

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfChildOfRootProperty.cs
chrisfromwork: Synced offline: onboard with this naming based on khronos schema declarations

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfImage.cs
chrisfromwork: Synced offline: onboard with this naming based on khronos schema declarations

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfInterpolationType.cs
chrisfromwork: Synced offline: url should be enough, most enums don't have summary stubs from khronos gltf standard schema declaration.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMagnificationFilterMode.cs
chrisfromwork: Synced offline: url should be enough, most enums don't have summary stubs from khronos gltf standard schema declaration.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMaterial.cs
chrisfromwork: Synced offline: onboard with this naming based on khronos schema declaration

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/ColliderType.cs
chrisfromwork: Wasn't able to find this on khronos declaration.

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfWrapMode.cs
chrisfromwork: Individual enum summaries didn't exist

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfBufferViewTarget.cs
chrisfromwork: Individual enum declarations didn't exist

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfComponentType.cs
chrisfromwork: Individual enum summaries didn't exist

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfDrawMode.cs
chrisfromwork: individual enum declarations didn't exist

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfMeshPrimitiveAttributes.cs
chrisfromwork: synced offline, everyones onboard with this naming.

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/README.md
chrisfromwork: Thanks @keveleigh let us know.

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/Lantern/README.md
chrisfromwork: Will follow up with other comment thread

Assets/MixedRealityToolkit/Utilities/Gltf/Schema/GltfObject.cs
chrisfromwork: keeping consistent with khronos schema declarations

chrisfromwork:
Confirmed gltf scene now works. Once we have legal sign off we should be able to get this in.

chrisfromwork:
New pull request out to fix namespace changes. closing this for now https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3673

PR #3552, opened 2019-03-01
Misc doc fixes and Unity Object lifetime check bypass fixes

StephenHodgson:
Overview
---
Fixed documentation errors.
Fixed [Unity.Object lifetime check bypass errors](https://github.com/JetBrains/resharper-unity/wiki/Possible-unintended-bypass-of-lifetime-check-of-underlying-Unity-engine-object)

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
@@ -126,8 +126,7 @@ namespace Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality
         /// <summary>
         /// The <see cref="GameObject"/> to which observed objects are parented.
         /// </summary>
-        private GameObject ObservedObjectParent => observedObjectParent ?? (observedObjectParent = SpatialAwarenessSystem?.CreateSpatialAwarenessObjectParent("WindowsMixedRealitySpatialMeshObserver"));
-
+        private GameObject ObservedObjectParent => observedObjectParent != null ? observedObjectParent : (observedObjectParent = SpatialAwarenessSystem?.CreateSpatialAwarenessObjectParent("WindowsMixedRealitySpatialMeshObserver"));
 
         private IMixedRealitySpatialAwarenessSystem spatialAwarenessSystem = null;
 
@@ -248,7 +247,7 @@ namespace Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality
 
         /// <inheritdoc />
         public Material VisibleMaterial { get; set; } = null;
-        
+
         /// <inheritdoc/>
         public override void Resume()
         {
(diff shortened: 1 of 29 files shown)


Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
```diff
@@ -126,8 +126,7 @@ public override void Destroy()
         /// <summary>
         /// The <see cref="GameObject"/> to which observed objects are parented.
         /// </summary>
-        private GameObject ObservedObjectParent => observedObjectParent ?? (observedObjectParent = SpatialAwarenessSystem?.CreateSpatialAwarenessObjectParent("WindowsMixedRealitySpatialMeshObserver"));
-
+        private GameObject ObservedObjectParent => observedObjectParent != null ? observedObjectParent : (observedObjectParent = SpatialAwarenessSystem?.CreateSpatialAwarenessObjectParent("WindowsMixedRealitySpatialMeshObserver"));
```
davidkline-ms: why this change? isn't the original version the pattern established for MRTK?

Assets/MixedRealityToolkit/Providers/BaseSpatialAwarenessObserver.cs
```diff
@@ -96,7 +95,7 @@ public override int GetHashCode()
 
         private GameObject observedObjectParent = null;
 
-        public GameObject ObservedObjectParent => observedObjectParent ?? (observedObjectParent = SpatialAwarenessSystem?.CreateSpatialAwarenessObjectParent(Name));
+        public GameObject ObservedObjectParent => observedObjectParent != null ? ObservedObjectParent : (observedObjectParent = SpatialAwarenessSystem?.CreateSpatialAwarenessObjectParent(Name));
```
davidkline-ms: are you suggesting this as the new pattern globally within MRTK?

keveleigh: The difference here is that most of those patterns are referring to interfaces, while this one is a `GameObject`

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
keveleigh: The difference here is that most of those patterns are referring to interfaces, while this one is a `GameObject`, and Unity uses a custom equality check for null which doesn't work in this case.

Assets/MixedRealityToolkit/Providers/BaseSpatialAwarenessObserver.cs
keveleigh: This looks like it'll overflow the stack by calling `ObservedObjectParent` repeatedly.

```suggestion
        public GameObject ObservedObjectParent => observedObjectParent != null ? observedObjectParent : (observedObjectParent = SpatialAwarenessSystem?.CreateSpatialAwarenessObjectParent(Name));
```

Issue #3562, opened 2019-03-04
Performance: Interactable Awake() extremely slow


kircher1:
## Overview
Creating an Interactable is extremely slow. There are two sub-issues here:

* Interactable.OnAwake calls a helper function, InteractableEvent.GetEventTypes(), which walks through *all types* in all loaded assemblies, looking for types which implement RecieverBase.
* Furthermore, SetupThemes() is also an extremely slow function. I haven't dug into this to see exactly where time is being spent though.

In my project and on my machine, GetEventTypes() takes around 40ms and SetupThemes() takes slightly longer. That is unacceptable for performance and results in a noticeable frame glitch. 

Also note, that time spent is just for instantiating a _single_ interactable. In my project, many interactables could be loaded throughout the lifetime of the app or even in a single frame.

## Expected Behavior
Interactable should not cause frame glitches.

## Actual Behavior
Interactable causes frame glitches when loaded.

## Steps to reproduce
* Add Profiler.BeginSample()/EndSample() to the InteractableEvent.GetEventTypes method.
* Create some interactables in a scene
* Observe spikes in the frame profiler

![interactableperf](https://user-images.githubusercontent.com/20366429/53761820-4f122700-3e7b-11e9-9c6c-77eea70b3d3c.png)

## Unity Editor Version
2018.3.6

## Mixed Reality Toolkit Release Version
Beta 2

kircher1:
This issue is only reproducable when an Interactable is instantiated at runtime, i.e. I don't see any sort of frame spike on frame 0 from Interactables that were already added to the scene...

Attached is a script that can be used to generate an Interactable every second. I used this in the InteractablesGallery scene with the Balloon prefab. You can see a consistent frame spike every second if this script is running. 

[InteractableGenerator.zip](https://github.com/Microsoft/MixedRealityToolkit-Unity/files/2928316/InteractableGenerator.zip)

davidkline-ms:
@kircher1, thanks for the report! we will investigate and advise / fix

cre8ivepark:
@killerantz for visibility

kircher1:
This appear to be fixed in RC1 Refresh.

PR #3563, opened 2019-03-04
Interactable startup performance fix

Railboy:
Overview
---
InteractableEvent and InteractableProfileItem now store the results of the static GetEventTypes() and GetThemeTypes() functions in static vars. (These results are cleared on recompile.)

Changes
---
- Fixes: #3562

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableEvent.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableEvent.cs
@@ -92,11 +92,28 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Interactable.Events
         }
 
         /// <summary>
+        /// True when event types have been compiled once by GetEventTypes
+        /// </summary>
+        private static bool eventListsCompiled = false;
+
+        /// <summary>
+        /// The shared list of compiled event types
+        /// </summary>
+        private static EventLists eventLists;
+
+        /// <summary>
         /// Get the recieverBase types that contain event logic
         /// </summary>
         /// <returns></returns>
         public static EventLists GetEventTypes()
         {
+            // This only needs to be done once.
+            // Recompilation will reset this value.
+            if (eventListsCompiled)
+            {
+                return eventLists;
+            }
+
             List<Type> eventTypes = new List<Type>();
             List<string> names = new List<string>();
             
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableEvent.cs
```diff
@@ -91,12 +91,29 @@ public ReceiverData AddReceiver(Type type)
             return data;
         }
 
+        /// <summary>
+        /// True when event types have been compiled once by GetEventTypes
+        /// </summary>
+        private static bool eventListsCompiled = false;
```
SimonDarksideJ: Why static?  This would mean one interactable could have it's events compiled and others might not yet be

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableEvent.cs
```diff
@@ -114,10 +131,13 @@ public static EventLists GetEventTypes()
                 }
             }
 
-            EventLists lists = new EventLists();
-            lists.EventTypes = eventTypes;
-            lists.EventNames = names;
-            return lists;
+            eventLists = new EventLists();
```
SimonDarksideJ: With eventlists being static, won't subsequent Interactables reset the previous controls events?

Railboy: This method just compiles a list of available ReceiverBase types. That only needs to be done once. I cache the results in a static field so subsequent requests return the cached types instead.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableEvent.cs
Railboy: The function searches loaded assemblies for a list of objects of type ReceiverBase. The list isn't tailored to each Interactable - it's global. So once it's compiled once, we can just return the cached list regardless of who's asking.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableEvent.cs
killerantz: The list is created ad-hoc while in the editor which allows for new scripts to be added to the list, but once the app starts running, caching it is the right call. New scripts cannot be added during runtime. This was an oversight on my part.

Railboy: The list gets purged when the project recompiles so editor behavior should remain the same.

Railboy:
Closing this. It looks like InteractableEvent and InteractableProfileItem have been changed recently in ways that make this PR unnecessary.

radicalad:
> Closing this. It looks like InteractableEvent and InteractableProfileItem have been changed recently in ways that make this PR unnecessary.

What changes were made to make this PR unnecessary? I'm still seeing perf issues concerning Interactable.Awake()

PR #3579, opened 2019-03-07
Changed the order of checks

Alexees:
Changed the order of checks since the platform can filter classes that might be null

Overview
While implementing the Oculus Go I received 5 errors every time I started the app. Since Service Providers specific to certain platforms are stripped by their containing ASMDEFs, checking for null first and the supported platform second does not seem very idea.

Changes
Changed the check order

### DIFF
--- a/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
@@ -618,6 +618,15 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
                 return false;
             }
 
+#if !UNITY_EDITOR
+            if (!Application.platform.IsPlatformSupported(supportedPlatforms))
+#else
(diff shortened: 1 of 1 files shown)


StephenHodgson:
@keveleigh why was this merged without review?

SimonDarksideJ:
I was a little surprised by this too, not even a second review by one of the MRTK team recorded.

keveleigh:
Huh...I definitely merged the two other PRs, but did not merge this one. Unsure what happened here, but I would have at least approved before merging.

keveleigh:
I figured out what happened! This commit (along with corresponding revert commits) was included in #3581 and #3584. I guess GitHub marked this as merged when it saw the commit contained in this PR as merged.

@Alexees, can you please open a new PR recreating this change? As-is in mrtk_development, this change isn't present as it was reverted in the other two merged PRs.

Alexees:
@keveleigh yes, sorry for the inconvenience. I thought I branched correctly, but the branches were all based on each other and this was an attempt to base them all on the same commit... did fail it seems.

Issue #3580, opened 2019-03-07
LinePointer always lags behind


Alexees:
## Overview
Moving a controller around one can clearly see that the LinePointer is detached from the controller, basically lerping behind.

## Expected Behavior
The LinePointer follows the controller on the spot

## Actual Behavior
The LinePointer has some sort of lerping behind behaviour

## Steps to reproduce
Use a simple scene and watch the LinePointer

## Unity Editor Version
2018.3.7f1

## Mixed Reality Toolkit Release Version
mrtk_development

Alexees:
You would think controller, raycast and linerenderer are related, but they are not. Controller orientation and raycasting happen during the same Update step, which is fine because they are initiated from the MixedRealityToolkit's Update method. However the LineRenderer runs it's own Update loop, which of course is undeterministically happening before or after the update of the controller.
In my case, it's before, that's why it lags behind.

keveleigh:
Closed via #3581

PR #3581, opened 2019-03-07
MRTK line renderer fix

Alexees:
Fixes: #3580


Alexees:
@keveleigh I'm confused about this one. According to the mrtk_development branch, the merge is empty. But this commit shows one file change.

Alexees:
@keveleigh should I create another pullrequest or can this be fixed? This change is definitely missing in mrtk_development, even though it's avaiable here.

keveleigh:
> should I create another pullrequest or can this be fixed? This change is definitely missing in mrtk_development, even though it's avaiable here.

I think you'll need to create another PR. This looks to be caused by the same thing that happened in #3579, where the commit (and a revert commit) was included in a different PR (in this case, #3584).

Issue #3582, opened 2019-03-07
TeleportCursor visible for one frame when not targeting a floor


Alexees:
## Overview
When not hitting the floor, the TeleportCursor is still activated for a frame. the first time it's within the camera or at 0,0,0, I'm not sure.

## Expected Behavior
Shows when hitting the floor only

## Actual Behavior
Shows for one frame no matter what

## Steps to reproduce
Use the teleport function in a scene without anything

## Unity Editor Version
2018.3.7f1

## Mixed Reality Toolkit Release Version
mrtk_development

keveleigh:
Fixed via #3584

PR #3584, opened 2019-03-07
MRTK TeleportPointer

Alexees:
Overview
The issue was that the TeleportPointer was set visible during the teleport request. If we actually hit something is evaluated during the raycast phase, which happened next frame and set the visibility itself, so there is no need to pre-visualize the pointer, which gets hidden in case of not hitting anything anyways.

Changes
---
- Fixes: #3582

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
@@ -375,21 +375,12 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
             if (eventData.Pointer.PointerId == PointerId)
             {
                 IsTeleportRequestActive = false;
-                BaseCursor?.SetVisibility(true);
             }
             else
             {
                 IsTeleportRequestActive = true;
                 BaseCursor?.SetVisibility(false);
             }
-        }
-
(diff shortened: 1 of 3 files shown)


Railboy:
Good catch.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
```diff
@@ -384,14 +383,6 @@ public override void OnTeleportRequest(TeleportEventData eventData)
             }
         }
 
-        /// <inheritdoc />
-        public override void OnTeleportStarted(TeleportEventData eventData)
-        {
-            // Turn off all pointers while we teleport.
-            IsTeleportRequestActive = true;
```
keveleigh: Seems like it might be good to keep this line, to prevent additional input while the teleport movement / animation is happening. Were you seeing it cause issues?

 https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/007187adfabe7c548055986c0f5bee1ba9a8e035/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs#L254

keveleigh: Nevermind! I see now you removed it because `BaseControllerPointer` has an identical block of code.

PR #3586, opened 2019-03-07
Mrtk Inspector Upgrade (Part 3)

Railboy:
Overview
---
This PR adds the MixedRealityServiceProfileAttribute class and adds inspector warnings for incorrectly configured registered services.

The attribute should be used on any profile class that is consumed by an IMixedRealityService.

```
[AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
public class MixedRealityServiceProfileAttribute : Attribute
{
    public MixedRealityServiceProfileAttribute(Type serviceType) { ServiceType = serviceType; }
    public Type ServiceType { get; private set; }
}
```
```
[MixedRealityServiceProfile(typeof(IMixedRealitySpatialAwarenessMeshObserver))]
public class MixedRealitySpatialAwarenessMeshObserverProfile : BaseMixedRealityProfile 
{
   ...
}
```
This is intended mostly for registered service providers. But core services profiles have been given the attribute for consistency.

In registered service configurations, if you attempt to set a profile that doesn't match the service type, an error _may_ be thrown by the service. But this is not guaranteed and is easily missed. Now a warning is displayed and remains in place until the problem is fixed:

profilevalidation

Issues
---
This is just an inspector warning. The attribute should really be used to enforce correct profile usage.

Next steps
---
Profile attributes are a first step towards automating some tedious profile-related tasks.

### DIFF
--- a/Assets/MixedRealityToolkit/Attributes/MixedRealityServiceProfileAttribute.cs
+++ b/Assets/MixedRealityToolkit/Attributes/MixedRealityServiceProfileAttribute.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Core.Interfaces;
+using System;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Attributes
+{
+    /// <summary>
+    /// Attribute that defines which service a profile is meant to be consumed by.
+    /// Only applies to profiles that are consumed by types implementing IMixedRealityService.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
+    public class MixedRealityServiceProfileAttribute : Attribute
+    {
+        public MixedRealityServiceProfileAttribute(Type serviceType) { ServiceType = serviceType; }
+
+        public Type ServiceType { get; private set; }
+    }
+}
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
@@ -151,7 +151,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Inspectors.Profiles
                 {
                     string showFoldoutKey = GetSubProfileDropdownKey(property);
                     bool showFoldout = SessionState.GetBool(showFoldoutKey, false);
-                    showFoldout = EditorGUILayout.Foldout(showFoldout, showFoldout ? "Hide " + property.displayName + " contents" : "Show " + property.displayName + " contents");
+                    showFoldout = EditorGUILayout.Foldout(showFoldout, showFoldout ? "Hide " + property.displayName + " contents" : "Show " + property.displayName + " contents", true);
 
                     if (showFoldout)
                     {
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
@@ -84,8 +84,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Inspectors.Profiles
 
                 if (!profileTypeIsValid)
                 {
-                    EditorGUILayout.HelpBox("The profile is not configured to be used with a service of type " + serviceType.Name
-                        + "\nIs your profile missing the " + typeof(MixedRealityServiceProfileAttribute).Name + " attribute?", MessageType.Warning);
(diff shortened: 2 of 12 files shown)


Railboy:
@StephenHodgson What would that involve? I'm keen on establishing more direct / followable links between profiles and services that consume them. That approach sounds like it could be more direct.

StephenHodgson:
I think the attributes definitely help solve that as well. Great idea!

StephenHodgson:
> What would that involve?

Likely adding the registry for the data provider and their profiles directly in the services profile. You'd be able to specify a specific kind of profile (be it via base type or interface).

davidkline-ms:
@Railboy, i like where you are going with the new attribute. @keveleigh and I were talking about adding a MixedRealityServiceAttribute and adding this type of information there. We should chat.

davidkline-ms:
Actually, the more I think about it, the more I like having a separate attribute to define profile information, as there will definitely be services, providers, etc that do not use a profile.

Assets/MixedRealityToolkit/Attributes/MixedRealityServiceProfileAttribute.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.MixedReality.Toolkit.Core.Interfaces;
+using System;
+
+namespace Microsoft.MixedReality.Toolkit.Core.Attributes
+{
+    /// <summary>
+    /// Attribute that defines which service a profile is meant to be consumed by.
+    /// Only applies to profiles that are consumed by types implementing IMixedRealityService.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
```
davidkline-ms: AllowMultiple and Inherited should likely be false to avoid needing collision resolution.

```suggestion
    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
```

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -55,18 +56,41 @@ protected static bool RenderProfile(SerializedProperty property, GUIContent guiC
...
+                    {
+                        profileTypeIsValid = true;
+                        break;
+                    }
+                }
+
+                if (!profileTypeIsValid)
+                {
+                    EditorGUILayout.HelpBox("The profile is not configured to be used with a service of type " + serviceType.Name
+                        + "\nIs your profile missing the " + typeof(MixedRealityServiceProfileAttribute).Name + " attribute?", MessageType.Warning);
```
davidkline-ms: Most consumers of MRTK will not be creating their own profile classes. Suggest the message read something like:

"This profile is not supported for <service>. Using an unsupported service may result in unexpected behavior."

Railboy:
> I like having a separate attribute to define profile information, as there will definitely be services, providers, etc that do not use a profile.

@davidkline-ms Exactly my thinking, and there may also be profile types that are consumed by more than one type of service. In which case you can just add multiple attributes to the profile.

Railboy:
Now fixes #3591.

SimonDarksideJ:
True @davidkline-ms / @Railboy , Profiles are optional for services and not all of them could possibly need one.
But we should probably also have an attribute the inspector can pick up to state when a profile is mandatory, and warn the user if they haven't configured one too.

davidkline-ms:
> we should probably also have an attribute the inspector can pick up to state when a profile is mandatory

Great suggestion! There has been some talk within the team about adding a service attribute (like there is for extension services and data providers). That would be a good place to specify profileRequired = <true|false>

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -101,7 +124,7 @@ private static bool RenderProfileInternal(SerializedProperty property, GUIConten
 
                 if (!renderedProfile.IsCustomProfile && profile.IsCustomProfile)
                 {
-                    if (GUILayout.Button(new GUIContent("</>", "Replace with a copy of the default profile."), EditorStyles.miniButton, GUILayout.Width(32f)))
+                    if (GUILayout.Button(new GUIContent("Clone", "Replace with a copy of the default profile."), EditorStyles.miniButton, GUILayout.Width(42f)))
```
StephenHodgson: Did you pick 42 cause it was the answer to life, the universe, and everything? 😉

Railboy: Of course.

StephenHodgson: @Ecnassianer brought up a [good comment](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3605#issuecomment-472139263). 

I'm thinking "clone" or something along those lines makes more sense.

Railboy: I went with that suggestion.

Issue #3577, opened 2019-03-06
MixedRealityInputSystem does not handle Focus events correctly


Alexees:
## Overview
Any Interactable component will suffer from incorrect handling of Focus events. If more than one pointer is involved (gaze and controller), a successful exit depends on the order the pointers enter and exit.

- The MixedRealityInputSystem will only ever call FocusEnter once on a gameobject, regardless of how many pointers entered it (pendingOverallFocusEnterSet makes sure of that). This fact makes the whole Pointer counting in the Interactable class useless.
- AddPointer and RemovePointer in the Interactable class add and remove first comes first serve. That means, if you exit focus with the pointer that entered second or later, OnFocusExit won't be called.

Either MixedRealityInputSystem calls Enter and Exit for EACH cursor, or Interactable cannot rely on a specific cursor that entered and exited.

## Expected Behavior
OnFocusExit always works

## Actual Behavior
It does not always work

## Steps to reproduce
- Gaze at an Interactable object
- Point at it with the controller
- Move the controller away
- Look away

The interactable stays in the selected state

## Unity Editor Version
2018.3.7f1

## Mixed Reality Toolkit Release Version
mrtk_development

Alexees:
I'm currently debugging this and the root cause for this is that gazing is calling OnFocusEnter and Exit, but the pointing cursor is not.  The moment the gaze exits, it checks for residing cursors and since the motion controller is still focusing, SetFocus will be called with true. When when then removing the cursor, the same call is not made twice.

Alexees:
Hm, ok:
- The MixedRealityInputSystem will only ever call FocusEnter once on a gameobject, regardless of how many gazes entered it (pendingOverallFocusEnterSet makes sure of that)
- AddPointer and RemovePointer in the Interactable class add and remove first comes first serve. That means, if it's not the same cursor that entered which exits first, the single exit call is passed, but the cursor count is still 1, because the other cursor is in the list.

Either MixedRealityInputSystem calls Enter and Exit for EACH cursor, or Interactable cannot rely on a specific cursor that entered and exited.

Alexees:
@davidkline-ms I think this is a problem and worth a discussion:
[...]

PR #3607, opened 2019-03-12
Mrtk focus handling

Alexees:
Overview
This all started with the 3D button in the Interactables gallery not getting reset when more than one cursor was involved and the one entering first was not the one exiting first.

Problem1:
the interactable class was collecting pointers that entered and exited. Unfortunately, the MixedRealityInputSystem was tracking focused objects on its own by just putting them in a HashMap for single enter and exit calls.

Solution:
I changed MixedRealityInputSystem to using a Dictionary to collect all pointers that entered a gameobject. I also introduced the Interface IMixedRealityFocusAmountHandler based on which a receiver can choose if it wants to receive every single Focus event or just the first enter or exit (which is the default).

Problem2:
After this I wanted to test how focussing went by listening to it from an InteractionReceiver. Unfortunately, all the Focus events just bypassed the regular call stack (HandleEvent<T>) and just called it on the focused object

Solution:
Uniformly routing all calls through the baseinputsystem

Problem3:
The MixedRealityInputSystem makes a fairly long attempt to get the focusedObject, which in case of the FocusExit event, is null.

Solution:
I split calls to IMixedRealityFocusHandler off so I could pass the focusedObject directly from the FocusProvider for these calls. In all other situations it's still the old way.
I actually did not know how to do this better, but I think the FocusedObject could very well be part of the inputeventdata since it's needed for all calls and the back and forth between MixedRealityInputSystem and FocusProvider is a little too much

Changes
---
- Fixes: #3577


Alexees:
Maybe this approach is incorrect. The following circumstance does not make total sense to me:
There are 4 calls to focus. OnBeforeFocusChange, OnFocusEnter, OnFocusExit and OnFocusChanged.
First of all I would rename them to OnPreFocusChange and OnPostFocusChange. OnBeforeFocusChange and OnFocusChanged were always called while OnFocusEnter and OnFocusExit were only called once. My initial problem with Interactable could have been resolved by not counting pointers at all.
But to come up with this I had to walk my way through this.

So the real question from my side would be:
Keep it like this? (OnBeforeFocusChange and OnFocusChanged being called every time while OnFocusEnter and OnFocusExit are called just once for the first and last)
Or go with removing OnBeforeFocusChange and OnFocusChanged and use my interface to decide if you want to receive OnFocusEnter or OnFocusExit once or multiple times?

With my pull request that's both at once

Alexees:
Just had an idea I want to try tomorrow, euch reduces code once more and interfaces for enter and exit to just one.
I don't think we need prechange and changed at all

Assets/MixedRealityToolkit/EventDatum/Input/FocusEventData.cs
```diff
@@ -10,13 +10,18 @@ namespace Microsoft.MixedReality.Toolkit.Core.EventDatum.Input
     /// <summary>
     /// Describes an Input Event associated with a specific pointer's focus state change.
     /// </summary>
-    public class FocusEventData : BaseEventData
+    public class FocusEventData : BaseInputEventData
```
StephenHodgson: What's the benefit of this change?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -266,7 +276,7 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
             // Get the focused object for each pointer of the event source
             for (int i = 0; i < baseInputEventData.InputSource.Pointers.Length; i++)
             {
-                GameObject focusedObject = FocusProvider?.GetFocusedObject(baseInputEventData.InputSource.Pointers[i]);
+                GameObject focusedObject = getFocusedObject(baseInputEventData.InputSource.Pointers[i]);
```
StephenHodgson: `getFocusedObject` is this a method?

Assets/MixedRealityToolkit/EventDatum/Input/FocusEventData.cs
Alexees: The MixedRealityInputSystem's HandleEvent method casts to BaseInputEventData. I routed the focus events through it too, therefore the change.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
Alexees: Yes. I head to split OnFocus events off of the regular HandleEvent call, because for the exit FocusProvider?.GetFocusedObject will return null. For this case I passed the focused object along with it through a lamda to this.

Alexees:
Ok, I think I'm done.
FocusProvider:
- The UpdateFocusedObjects method is much cleaner now
- The focus calls in UnregisterPointer were wrong, one was missing, fixed that
- Focus events can now also be observed from InteractionReceivers since they're now also using the MixedRealityInputSystem's HandleEvent method for propagation
- I extended IMixedRealityFocusHandler with ReceiveAllFocusChanges, which lets you choose if you want to receive the first and last or all focus events

Still the question remains if that should be removed, or IMixedRealityFocusChangedHandler should

davidkline-ms:
Thanks for submitting this @Alexees! The team will review and provide feedback as soon as possible.

Assets/MixedRealityToolkit/EventDatum/Input/FocusEventData.cs
keveleigh: What's the benefit of sending focus events through the `HandleEvent` path instead of directly to the focused / unfocused object?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -745,15 +719,23 @@ public void RaiseFocusEnter(IMixedRealityPointer pointer, GameObject focusedObje
...
                     var casted = ExecuteEvents.ValidateEventData<FocusEventData>(eventData);
-                    handler.OnFocusEnter(casted);
+                    if (handler.ReceiveAllFocusChanges)
+                    {
+                        handler.OnFocusEnter(casted);
+                    }
+                    else
+                    {
+                        if (casted.FocusProvider.IsOnlyFocusingPointer(casted.NewFocusedObject, casted.Pointer))
+                        handler.OnFocusEnter(casted);
```
keveleigh: It's a little unclear without the braces, but I think the indentation is off here. Preferably, braces would be added here and on line 758 for readability.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
davidkline-ms: Minor issue: MRTK code style capitalizes the first letter of method names.

davidkline-ms:
> Problem1:
the interactable class was collecting pointers that entered and exited. Unfortunately, the MixedRealityInputSystem was tracking focused objects on its own by just putting them in a HashMap for single enter and exit calls.

Thank you for pointing this out! 

@Alexees, I do have a concen i would love your feedback on with regards to this PR... Changing to send FocusEnter and FocusExit for each pointe, while giving controls that wish to track each pointer individually more information, it would _require_ all controls to have their own tracking logic to avoid potential odd UX and state behaviors.

Given the above concern, i _DO_ like the idea of being able to remove the HashSet(s) in the focus management code.

What are your thoughts?

Assets/MixedRealityToolkit/EventDatum/Input/FocusEventData.cs
Alexees: the InteractionReceiver could listen to everything that happens to another gameobject, except for these.
this path uses global listeners while the previous bypassed them.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
Alexees: it's a System.Action parameter. does it supply here too?

Alexees:
how does it require all controls to have their own tracking logic? As I see it, those which don't just return false for ReceiveAllFocusChanges and will only receive first enter and last exit. those which want can receive them all by returning true.
different tracking logic wasn't even possible yet and IF for some reason an object is capable of handling more than one pointer individually, I think that needs custom logic anyways. like ripping a button in half :D

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
Alexees: done

StephenHodgson:
@keveleigh I noticed that the base focus handler no longer properly raises `OnBeforeFocusChange` maybe a regression from when we broke up the interfaces?

Alexees:
@StephenHodgson is you mean this #3608, those have indeed but been called anymore.

StephenHodgson:
@Alexees nope, that's different. It was moved into a different interface and that interface wasn't added back to the base focus handler.

keveleigh:
@StephenHodgson, yep, looks like a regression that's addressed in this PR: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3607/files#diff-4a9aa84a9b6c99967553bf081bc15b74R16

luis-valverde-ms:
+1 to @Alexees question: 
> Are OnPreFocusChange and OnFocusChanged intended to be fired before and after any actual focus change? (Like Unity's Awakes being called before any Starts)
Do we have any use cases in mind for this? Otherwise, my preference would be to have just OnFocusEnter/Exit, raise them for all pointers regardless of the state of the target and let UX components decide if they care just about the first/last. The code on the UX side should be as simple as keeping a counter.

StephenHodgson:
> Do we have any use cases in mind for this? 

Yes we were using it in the BaseFocusHandler to update our list of focused objects.

There's other things using this interface as well like `ICursorModifier`, `InteractionReceiver`, and the `CursorModifier`

davidkline-ms:
I believe this change is very close to being approved. The team was reviewing change proposals and liked where this was going.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -22,9 +22,8 @@ namespace Microsoft.MixedReality.Toolkit.Services.InputSystem
     public class FocusProvider : BaseService, IMixedRealityFocusProvider
     {
         private readonly HashSet<PointerData> pointers = new HashSet<PointerData>();
-        private readonly HashSet<GameObject> pendingOverallFocusEnterSet = new HashSet<GameObject>();
-        private readonly HashSet<GameObject> pendingOverallFocusExitSet = new HashSet<GameObject>();
-        private readonly List<PointerData> pendingPointerSpecificFocusChange = new List<PointerData>();
+        private readonly Dictionary<GameObject, HashSet<IMixedRealityPointer>> overallFocusSet = new Dictionary<GameObject, HashSet<IMixedRealityPointer>>();
```
luis-valverde-ms: Could we just keep the count of pointers over each object instead?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -745,15 +719,25 @@ public void RaiseFocusEnter(IMixedRealityPointer pointer, GameObject focusedObje
                 delegate (IMixedRealityFocusHandler handler, BaseEventData eventData)
                 {
                     var casted = ExecuteEvents.ValidateEventData<FocusEventData>(eventData);
-                    handler.OnFocusEnter(casted);
+                    if (handler.ReceiveAllFocusChanges)
```
luis-valverde-ms: I feel like this logic doesn't belong here. We have three classes (FocusManager, InputSystem and FocusHander) cooperating to decide if the event should be raised. We should instead raise the event always and leave it to the FocusHandler to decide what to do with it.

Alexees: that's what @davidkline-ms wasn't sure with, he mentions that further up the discussion.

luis-valverde-ms: > Changing to send FocusEnter and FocusExit for each pointe, while giving controls that wish to track each pointer individually more information, it would require all controls to have their own tracking logic to avoid potential odd UX and state behaviors.

I think that's the relevant bit. Correct me if I'm wrong but I think the logic needed to implement the current behavior is something like this:

```
int count = 0;
Pointer pointer;

OnFocusEnter(event)
{
	count++;
	if (count == 1)
	{
		pointer = event.pointer;
	}
}

OnFocusExit(event)
{
	count--;
	if (event.pointer == pointer)
	{
		pointer = null;
	}		
} 
```
It doesn't seem too bad but we could have that in BaseFocusHandler, something like:
```
class BaseFocusHandler
{
	bool receiveAllFocusEvents = false;
	
	abstract HandleFocusEnter();
	
	sealed OnFocusEnter(event)
	{
		if (!receiveAllFocusEvents)
		{
			count++;
			if (count == 1)
			{
				HandleFocusEnter(event);
			}			
		}
	}
	
	// ...	
}
```
What do you think?

Alexees: this is about what this does already, with the tiny convenience extra that you don't have to check for yourself, that's what the InputManager does.
If handler.ReceiveAllFocusChanges is false you just won't receive but the single ones. if it's true you will.
My feeling is that's cleaner.

luis-valverde-ms: I have to disagree with you there. Making this a pure UI concern, i.e. private to the component handling focus, provides better encapsulation which makes the system as a whole easier to understand, use and change. 

With the current solution in this PR, we rely on implementations of the input system interface checking the property in the focus handler before raising focus enter/left events. This is not enforced in any way by the interface definitions, so you will only realize you got it wrong when it fails at runtime instead of at compile time. If UI components take responsability for it instead, it is more likely they will work fine with a different input system implementation.

StephenHodgson: I agree with @luis-valverde-ms this should be something the UI should be concerned about, not the focus provider.

Alexees: @luis-valverde-ms it is enforced by the interface IMixedRealityFocusHandler. you make that Choice when implementing it.

Alexees: @StephenHodgson just to be clear about this. I would say the majority of cases requires just a single enter and exit. And the receivers do require an interface to receive them.
This means most of the time you write a counter every time you implement this interface if you want just the first and the last call. Or you start with a base class that does it, introducing virtual calls for receiving just single calls.
With this PR it's merely letting ReceiveAllFocusChanges return false and you're good to go. And it makes it very clear during the implementation what calls will be received, all or one.
My hand up for the PR.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
Alexees: The idea of an AR Indiana Jones reaching out for the golden Trophy just came to be. I think you want to be able to make sure that both hands and not just a hand and the gaze are reaching out to it before you throw in the glow shader :)

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
luis-valverde-ms: @Alexees, totally agree with you in that we should make the most common use case as easy as possible to implement. Not at the cost of introducing that dependency in the Input System though. 

I feel that we already have the mechanism to allow UI implementers to choose between first/all focus events in the shape of the Focus and FocusChanged handler interfaces. If there are no strong objections, I'm going to go ahead and change Interactable to keep track of the pointers using FocusChanged so we can close #3577.

luis-valverde-ms: PR ready: #3670

Alexees:
From my side this is good to go now:

- Sorted IMixedRealityFocusHandler and IMixedRealityFocusChangedHandler out
- Fixed unregistered pointers not receiving all 3 necessary calls (PreFocusChanged, Exit, FocusChanged)
- Removed Focus counting from the FocusProvider, all enter and exit calls now reach gameobjects
- Allowed all 4 Focus events to be listened to by modalHandlers

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -854,70 +840,30 @@ private bool RaycastGraphicsStep(PointerData pointerData, RayStep step, LayerMas
         /// </summary>
         private void UpdateFocusedObjects()
         {
-            Debug.Assert(pendingPointerSpecificFocusChange.Count == 0);
-            Debug.Assert(pendingOverallFocusExitSet.Count == 0);
-            Debug.Assert(pendingOverallFocusEnterSet.Count == 0);
-
-            // NOTE: We compute the set of events to send before sending the first event
-            //       just in case someone responds to the event by adding/removing a
-            //       pointer which would change the structures we're iterating over.
```
luis-valverde-ms: Is this no longer a valid concern?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -854,70 +840,30 @@ private bool RaycastGraphicsStep(PointerData pointerData, RayStep step, LayerMas
...
                     {
-                        pendingOverallFocusExitSet.Add(pointer.PreviousPointerTarget);
+                        MixedRealityToolkit.InputSystem.RaiseFocusExit(currentPointer, pendingUnfocusObject);
                     }
 
-                    if (pointer.CurrentPointerTarget != null)
+                    if (pendingFocusObject != null)
                     {
-                        pendingOverallFocusEnterSet.Add(pointer.CurrentPointerTarget);
+                        MixedRealityToolkit.InputSystem.RaiseFocusEnter(currentPointer, pendingFocusObject);
```
luis-valverde-ms: >MixedRealityToolkit.InputSystem.RaiseFocusEnter(currentPointer, pendingFocusObject); [](start = 24, length = 84)

Raising focus enter/left for all pointers will break all focus handlers relying on the existing behavior. We already have focus changed events to detect multiple pointers,

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/BaseFocusHandler.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.Input.Handlers
     /// Base Component for handling Focus on <see href="https://docs.unity3d.com/ScriptReference/GameObject.html">GameObject</see>s.
     /// </summary>
     [RequireComponent(typeof(Collider))]
-    public abstract class BaseFocusHandler : MonoBehaviour, IMixedRealityFocusHandler
+    public abstract class BaseFocusHandler : MonoBehaviour, IMixedRealityFocusHandler, IMixedRealityFocusChangedHandler
```
luis-valverde-ms: >IMixedRealityFocusChangedHandler [](start = 87, length = 32)

Good catch. I would gladly take the changes to this file as a separate change.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/BaseFocusHandler.cs
```diff
@@ -38,12 +38,18 @@ public virtual bool FocusEnabled
...
 
         /// <inheritdoc />
         public virtual void OnFocusExit(FocusEventData eventData) { }
 
+        #endregion IMixedRealityFocusHandler implementation
+
+        #region IMixedRealityFocusChangedHandler Implementation
+
         /// <inheritdoc />
         public virtual void OnBeforeFocusChange(FocusEventData eventData)
```
luis-valverde-ms: >virtual [](start = 15, length = 7)

This is the reason we got in trouble to begin with. Can we have this and all other 'virtual' in focus handler methods replaced by override, so we get a build error if the interface we're implementing changes?

Alexees:
@luis-valverde-ms As I understood it everybody agreed on letting UI elements do the counting, like you you did in your recent pull request. If Enter and Exit should be fired once by default, why did you implement pointer counting then ?

luis-valverde-ms:
@Alexees , I implemented the counting based on FocusChanged events which are already meant to be fired always and allows us to keep the existing behavior for Enter and Exit.

Alexees:
I think I don't get it. why counting if we are supposed to receive each enter and exit only once anyways?

luis-valverde-ms:
(Before)FocusChanged and Enter/Exit are different events. (Before)FocusChanged is raised always, Enter/Exit just on first enter/last exit.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -681,30 +699,19 @@ public void RaisePreFocusChanged(IMixedRealityPointer pointer, GameObject oldFoc
...
-
             if (newFocusedObject != null)
             {
-                ExecuteEvents.ExecuteHierarchy(newFocusedObject, focusEventData, OnPreFocusChangedHandler);
+                HandleEvent(focusEventData, newFocusedObject, OnPreFocusChangedHandler);
             }
 
             // Raise Focus Events on the pointers cursor if it has one.
-            if (pointer.BaseCursor != null)
+            if (pointer?.BaseCursor?.GameObjectReference != null)
```
luis-valverde-ms: This changes the current behavior with consequences I find difficult to predict.

You were saying this change is motivated by InteractionRecevier not receiving focus events for its interactables. Were those interactables under the receiver in the hierarchy? I'm not sure it is meant to receive them otherwise. I they were under the interactable or you think the behavior should be different, could you open an issue for it? 

In general we should try to narrow down the scope of PRs as much as possible and have an issue backing them, to make it easier for reviewers to understand and approve them quickly.

Alexees:
I unterstand that, what i don't unterstand is why you additionally count pointers?

luis-valverde-ms:
Where am I counting those pointers?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
Alexees: Am I correct that the FocusProvider is a service and should therefore not be concerned with limiting focus enter and exit calls, which it currently does?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
Alexees: InteractionRecevier, as a derivate of InputSystemGlobalListener, registers as a modal listener and therefore receives everything regardless where in the hierarchy it is. All events could be listened to that way, except for the 4 Focus events.
You are correct, I am going to split this up into their own concerns tomorrow. This has gotten consfusing enough already.

luis-valverde-ms: My point about InteractionReceiver was that if it is meant to have its interactables below it in the hierarchy it would receive those focus events when they bubble up. There is nothing in code enforcing interactables to be below the InteractionReceiver though. @radicalad , do you know if InteractionReceiver is meant to have its interactables as siblings/descendants?

radicalad: InteractionReceiver and the Interactables are entirely different. I agree InteractionReceiver should handle focus, and it did in the [HoloToolkit version](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/htk_release/Assets/HoloToolkit/UX/Scripts/Receivers/InteractionReceiver.cs).

InteractionReceiver is a easy way to run shared logic on a group of  (I'm not saying _interactable_ since its overloaded) objects in the scene. Simply inherit InteractionReceiver, and override whichever events you desire. Objects in InteractionReceiver's list will bubble up.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
Alexees: > If there are no strong objections, I'm going to go ahead and change Interactable to keep track of the pointers using FocusChanged so we can close

@luis-valverde-ms if we leave the FocusProvider as is, there is no need to count pointers in Interactables, because FocusProvider already does and fires enter and exit just once.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
luis-valverde-ms: Thanks for the info @radicalad . So, just to be clear, does InteractionReceiver require the objects it is keeping track of to be descendants of its game object?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
luis-valverde-ms: I'm no Interactable expert but it seems that it is meant to keep a list of all the pointers focusing it and expose it via the `Focusers` property.

Alexees: Ok, so to sum it up:
- The Focus Provider counts pointers and makes sure enter and exit are called just once
- PreFocusChanged and FocusChanged are called for all changes
- Interactables count pointers too, so there's information about which pointers are focusing

To be honest, I don't like having 4 different events for focusing and two or more places where pointers are counted, but that's probably just me.
I'd rather like to see a solution with only 2 events and one place for counting, referenced when needed. I mean all other events do not have pre and post events. PrePointerClick anyone?

MenelvagorMilsom:
@Alexees don't consider the PR validation failures here as blocking. They're caused by an issue that should now be fixed. 
Unfortunately, the fixed version no longer uses the same pipeline, so we can't just rerun the checks. If you want to test against the new PR validation, you could maybe try forcing a change to your branch.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
Alexees: Actually not. Proceeding with your recommendation this now just counts pointers for each focusable gameobject.

luis-valverde-ms:
@Alexees , thanks for looking into that and sorry for the delayed response, we're quite busy with RC1 refresh.

I think there is an issue with UpdateFocusedObjects() in the PR: if an object loses focus from a pointer and gains it from another in the same frame, it will or won't receive focus enter/exit events depending on the order in which the two pointers are processed in the loop. Incidentally, we also seem to be raising focus changed instead of focus enter. 

Is this PR something you're still actively working on? Otherwise, can we mark it as closed seeing how the code currently checked in in mrtk_development works alright?

Alexees:
> I think there is an issue with UpdateFocusedObjects() in the PR: if an object loses focus from a pointer and gains it from another in the same frame, it will or won't receive focus enter/exit events depending on the order in which the two pointers are processed in the loop. Incidentally, we also seem to be raising focus changed instead of focus enter.

You're right, cross changing pointers is a problem.

> Is this PR something you're still actively working on? Otherwise, can we mark it as closed seeing how the code currently checked in in mrtk_development works alright?

I'll check the current implementation and will see if this here is still necessary

Alexees:
This issue is no more, but one issue still remains. I'll open another pull request for it and start from the current development implementation.

luis-valverde-ms:
Thanks! Can you also open a bug for the outstanding issue if there's not one already? Also, if you're going to work off this PR, I'd rather simplify the code and do the bookeeping for pointer enter/left directly in UpdateFocusedObjects() instead of delegating it to the FocusCounter class. Performance shouldn't be an issue here and the less we state we cache the better.

Alexees:
Ok, I will. Just realized that this PR is valuable for the new bug as it is related closely.

PR #3608, opened 2019-03-12
IMixedRealityFocusHandler cleanup

Alexees:
Overview
There were still traces of the two obsolete methods in IMixedRealityFocusHandler. I removed them along with the methods.


davidkline-ms:
Thanks for the cleanup, @Alexees!

PR #3609, opened 2019-03-12
Add missing meta files

keveleigh:
Overview
---
Some of the docfx files had meta files generated for them when opening in Unity.


thalbern:
I suppose Unity thinks they're part of the project as they're yaml files. Is there a way to add them to some sort of ignore list so Unity won't create those meta data files?

keveleigh:
>Is there a way to add them to some sort of ignore list so Unity won't create those meta data files?

Not that I'm aware of. We can stop them from being tracked by git, but I'm not sure we can stop them from being created.

PR #3617, opened 2019-03-13
Editor Null Reference Fixes

andreiborodin:
Overview
---
This change fixes a few null reference bugs in the editor when MRTK is unavailable, and a prefab is opened. The simplest repro is:

1. Create a new scene
2. Open any prefab
3. Observe null reference being thrown.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
+++ b/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
@@ -37,10 +37,10 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Interactable
         protected Type[] themeTypes;
         protected string[] shaderOptions;
 
-        protected string[] actionOptions;
+        protected string[] actionOptions = null;
 
         protected static bool ProfilesSetup = false;
-        
+
         // indent tracker
         protected static int indentOnSectionStart = 0;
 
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -55,26 +55,29 @@ protected virtual void OnEnable()
...
         }
-        
+
         protected virtual void RenderBaseInspector()
         {
             base.OnInspectorGUI();
         }
 
-        public override void OnInspectorGUI()
+        public sealed override void OnInspectorGUI()
```
davidkline-ms: why add sealed here? why would we want no further sub-classes to be able to override?

andreiborodin: Yep, good question, we have the RenderCustomInspector available for overriding, so that path can be used. I needed to move actionOptions to this location and wanted to prevent overriding of this behavior by a sub class. My thinking is that if there is a compelling reason to undo it that RenderCustomInspector doesn't satisfy, it can be undone then. Today no class overrides it.

davidkline-ms: @keveleigh, @wiwei, thoughts?

wiwei: I think it's fine here, given that the intent is to make sure that the code below:

 if (actionOptions == null && !Interactable.TryGetInputActions(out actionOptions))
            {
                EditorGUILayout.HelpBox("Mixed Reality Toolkit is missing, configure it by invoking the 'Mixed Reality Toolkit > Configure...' menu", MessageType.Error);
            }

Actually gets run. Being able to further override makes it easy to forget to add this (I assume that this check is actually critical).

It might be worth having a minor comment about that (i.e. "please override through RenderCustomInspector to ensure that correct prerequisite state checks are performed on your behalf")

andreiborodin: Good idea on the comment, will add it.

Issue #3614, opened 2019-03-13
Proposal: Reworking internal execution paths


Alexees:
## Overview
There are execution orders that are not correct. There might be more, but so far I found the following issues, trying to bundle them together here:
#3612, #3585, #3580

It all boils down to the following two problems:

- MRTK has two main service paths, activeServices and registeredMixedRealityServices. The former executes the FocusProvider, the latter executes the MixedRealityInputSystem

Issue 1:
- FocusProvider: Updates all LinePointers. In PreRaycast positions get set on BaseMixedRealityLineDataProvider, but those get run through their localMatrix which has not been set yet. This only happens in PostRaycast when the component gets enabled

Issue 2:
- FocusProvider: Updates all LinePointers which perform all their Raycasts and Update hit info
- MRTK: Updates MixedRealityInputSystem in registeredMixedRealityServices. During this phase the TeleportPointer can perform movement and rotation. This makes all previously calculated Raycasts obsolete for all further calculations.

Bottom line: Raycasted PointerData as well as the darn BaseMixedRealityLineDataProvider matrix )who needs that anyways?) can be old.

## Unity Editor Version
2018.3.8f1

## Mixed Reality Toolkit Release Version
mrtk_development

![Execution_Graph](https://user-images.githubusercontent.com/32508825/54424664-b5b50180-4713-11e9-8839-5a81de22f323.png)

StephenHodgson:
The orders of the updates aren't random, actually. They're ordered by the service's priority.

But yes, it's true that all the systems get their updates called first, then the list of services.

StephenHodgson:
@Alexees also great job with the graph! It really helps visualize the problem :)

Alexees:
I was referreing to Update from Unity. The three groups I created under "random ordered Updates" all contain Unity Updates, not Updates based on MRKT.

StephenHodgson:
Gotcha. One last thing I also wanted to point out is that the events processed from the input system are not bound to an update loop. Once the button is registered as pressed in the data provider it immediately gets bubbled up to the input system, which then routes it to the appropriate GameObject.

Alexees:
@StephenHodgson I double checked. All InputEvents actually ARE bound to the MRTK Update loop. Now that I saw that I remember that some time ago there was a major decision made if you wanted to let the inputs bubble up on their own or if you wanted to pull them. Guess the decision felt on the latter. I think @keveleigh was involved in this.

I updated the graph

[...]

PR #3624, opened 2019-03-14
Mrtk play space update

Alexees:
Overview
This fix encapsulates the MixedRealityPlayspace Transform so the only way to alter it is by its PerformTransformation method. This method queues the transformations and executes them at the beginning of the next frame to before the services are updated in MixedRealityToolkit. This ensures, that the playspace trransform is modified before raycasting logic happens.

Fixes part of #3614


Assets/MixedRealityToolkit/Services/MixedRealityPlayspace.cs
```diff
@@ -0,0 +1,90 @@
+﻿using UnityEngine;
+using System;
+using System.Collections.Generic;
+using Microsoft.MixedReality.Toolkit.Core.Utilities;
+
+public class MixedRealityPlayspace
```
StephenHodgson: could this be static?

Alexees: yes, sure

Alexees:
Not at all. staticed it completely

Railboy:
>This method queues the transformations and executes them at the beginning of the next frame

I really, really like this approach.

Assets/MixedRealityToolkit/Services/MixedRealityPlayspace.cs
```diff
@@ -0,0 +1,90 @@
+﻿using UnityEngine;
+using System;
+using System.Collections.Generic;
+using Microsoft.MixedReality.Toolkit.Core.Utilities;
+
+public static class MixedRealityPlayspace
```
StephenHodgson: Why put this in services if it doesn't have a service implementation?
Would this actually be just a utility class?

Alexees: I moved it over to Utilities. Left is the name. None of the classes there have MixedReality in their names. But the Playspace, as an instance in the scene is best identified with it.
What's your suggestion?

Alexees:
what's the framerate without this change?

Railboy:
@davidkline-ms Interesting. I'll gather some profiler data and see if I can spot why.

Alexees:
The result of the discussion in #3614 led to a solution, that makes this approach obsolete.

davidkline-ms:
@Alexees, the service registration work (related to #3545) will greatly benefit from this PR. I hope you do not mind if i revive it.

Alexees:
@davidkline-ms no I don't ;)

davidkline-ms:
Thanks! It may not be the full implementation you have, but it'll make things much nicer with the new updates.

PR #3625, opened 2019-03-14
BaseCursor LateUpdate instead of Update

Alexees:
Overview
The Base Cursor was in a race condition with the MixedRealityToolkit updating Raycast data

Changes
---
- Fixes: part of #3614 .

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
@@ -233,7 +233,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Cursors
 
         #region MonoBehaviour Implementation
 
-        private void Update()
+        private void LateUpdate()
         {
             UpdateCursorState();
             UpdateCursorTransform();


Ecnassianer:
It just occurred to me that moving these changes to LateUpdate means that anything that naively uses the cursor state during Update will be using last frame's data. This will add a frame of latency to anything that's trying to follow or collide with the cursor.

Evenso, I'm a big fan of LateUpdating the cursor, since it reduces latency on the cursor.

What do you all think about at least partially updating the cursor during Update, and then again in LateUpdate? We would remove a frame of delay from anything that's checking the cursor during Update.

Alexees:
@Ecnassianer I just realized that moving to LateUpdate was not even necessary, the problem I tried to solve  layed deeper down the Toolkit. But what even reverting won't fix is people listening for changes in Update, because THEN they get a race condition with the controllers since those and the listener both run on regular Unity Update's. If you want this fixed the right way you Update the controllers from the toolkit.

Alexees:
@Ecnassianer I just realized that moving to LateUpdate was not even necessary, the problem I tried to solve  layed deeper down the Toolkit. But what even reverting won't fix is people listening for changes in Update, because THEN they get a race condition with the controllers since those and the listener both run on regular Unity Update's. If you want this fixed the right way you Update the controllers from the toolkit.

Ecnassianer:
Hey no worries, it took me 6 hours from the time I read this until it dawned on me! :D This ordering stuff is super complicated! You're doing great, keep at it!

PR #3626, opened 2019-03-14
Mrtk data provider matrix update

Alexees:
Overview
The BaseMixedRealityLineDataProvider was updating its matrices in Update, which did not match the timing of the LinePointer needing the matrix in its PreRaycast phase. Since it's the only class activating this component in OnPostRaycast and is raycasting all the time, I replaced its internal Update with a MatrixUpdate in the LinePointer's PreRaycast phase.

Changes
---
- Fixes: part of #3614.


Railboy:
Adding an UpdateMatrix call to PreRaycast is a good idea, but we shouldn't remove it from the Update (or LateUpdate) call. Lines are used outside of pointer classes and this update is necessary for them to function in UseMatrix mode.

Alexees:
@Railboy I reverted that

PR #3629, opened 2019-03-14
Fixing Teleportation to be visually disabled when the system is disab…

andreiborodin:
Overview
---
The bug is that when you disable the Teleport System, the visuals still activate when you try to teleport. The fix here is to check that teleport system is in fact enabled.

This change also contains 2 minor fixes in Profile Editors that hit an ArgumentException when either compiling or the window is behind another window.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
@@ -101,7 +101,8 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
         #region IMixedRealityPointer Implementation
 
         /// <inheritdoc />
-        public override bool IsInteractionEnabled => !IsTeleportRequestActive && teleportEnabled;
+        public override bool IsInteractionEnabled => !IsTeleportRequestActive && teleportEnabled
+            && (MixedRealityToolkit.IsInitialized && MixedRealityToolkit.HasActiveProfile && MixedRealityToolkit.Instance.ActiveProfile.IsTeleportSystemEnabled);
 
         /// <inheritdoc />
         public override float PointerOrientation
@@ -251,7 +252,11 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Pointers
         public override void OnInputChanged(InputEventData<Vector2> eventData)
         {
             // Don't process input if we've got an active teleport request in progress.
-            if (IsTeleportRequestActive) { return; }
+            if (IsTeleportRequestActive
+                || !(MixedRealityToolkit.IsInitialized && MixedRealityToolkit.HasActiveProfile && MixedRealityToolkit.Instance.ActiveProfile.IsTeleportSystemEnabled))
+            {
+                return;
+            }
 
             if (eventData.SourceId == InputSourceParent.SourceId &&
                 eventData.Handedness == Handedness &&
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
```diff
@@ -56,6 +56,12 @@ protected override void OnEnable()
         {
             base.OnEnable();
 
+            if (target == null)
```
wiwei: I think the class hierarchy goes:

MixedRealityToolkitConfigurationProfileInspector -> BaseMixedRealityToolkitConfigurationProfileInspector -> BaseMixedRealityProfileInspector

Since BaseMixedRealityToolkitConfigurationProfileInspector doesn't override OnEnable, in theory this behavior should already exist from the code in BaseMixedRealityProfileInspector.

Unless this is being done because you're afraid that future people who override BaseMixedRealityToolkitConfigurationProfileInspector  will not call base.OnEnable?

wiwei: What I mean to say is, it seems like this instance of the code isn't necessary because it's handled in the base class. If it's being done again in a subclass (and only one of them!) it would be great to understanding the motivation behind it.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
```diff
@@ -101,7 +101,8 @@ protected Gradient GetLineGradient(TeleportSurfaceResult targetResult)
         #region IMixedRealityPointer Implementation
 
         /// <inheritdoc />
-        public override bool IsInteractionEnabled => !IsTeleportRequestActive && teleportEnabled;
+        public override bool IsInteractionEnabled => !IsTeleportRequestActive && teleportEnabled
+            && (MixedRealityToolkit.IsInitialized && MixedRealityToolkit.HasActiveProfile && MixedRealityToolkit.Instance.ActiveProfile.IsTeleportSystemEnabled);
```
wiwei: This line basically means "if teleportation system is online" -> suggestion is to make some other helper bool (i.e. private bool IsTeleportationSystemPresent) because there are two instances of these three specific things you have to reference at the same time (and it looks like in the same way, since you need to call inited first, then active profile, then teleportsystemenabled)

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
andreiborodin: Had to short circuit the sub class.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
andreiborodin: Yeah, was thinking of doing that, will do that

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
wiwei: Ah doh, yes that's obvious now

PR #3630, opened 2019-03-14
Enable .NET backend scripting support

wiwei:
Given that we have customers who are migrating and need to continue to work on the .NET backend, we need to keep supporting this until Unity itself gets rid of .NET backend support.

This change is broken up into some less complex things (which we will describe first) and some more complex things (serializing actual assembly types instead of doing runtime based reflection).

Note that this is actually a breaking change for people who have defined their own assets that use the interactable based theme/event/state systems. The way to migrate is to do the similar .asset/.unity filechanges included in this PR (i.e. adding a proto field that is the AssemblyQualifiedName). My hope is that I might be able to just write a migrator script to make this easier.

Less complex things:
[Certain Types APIs](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.getcustomattributes?view=netframework-4.7.2#System_Reflection_MemberInfo_GetCustomAttributes_System_Type_System_Boolean_) aren't supported in UWP .NET, but thankfully have a workaround (i.e. getting the type info and going from there).

More complex things:
The rest of the changes have to deal with the fundamental issues around using APIs that don't exist in UWP .NET (i.e. AppDomain). Unfortunately, there isn't even a replacement for these (yes, it's possible to enumerate through the installed app's DLLs, but on master unity releases i.e. when the customer actually pushes the app for real, those assemblies are all removed). Fundamentally, we shouldn't be using reflection at runtime to power dependency hookup behavior (there are other patterns we can follow i.e. DI, Service locators, lotsa other stuff).

I ultimately chose to implement things this way as a workaround, not because I believe this is fundamentally the right long term approach (i.e. I want us to be able to figure out the best way of authoring easy to use UI helpers which properly hookup dependencies), but because we have an immediate need to have .NET backend support working, and doing this this way would be the least "destructive" of the common MRTK use case.

I could have, for example, done this using SciprtableObjects (as singletons) but would need to attach these objects to SOME object in the scene (which would have messied the scene graph).

I could have added some other profile based system to the MRTK, but I want to get wide buy-in first that such a thing makes sense.

The drawback of this approach comes from the fact that now future refactors of the namespace of these objects will require additional asset changes (see the asset things that I've done).


Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
```diff
@@ -60,9 +61,18 @@ public class MixedRealityControllerAttribute : Attribute
...
         /// </summary>
+        /// <remarks>
+        /// This function is only available in a UnityEditor context.
+        /// </remarks>
         public static MixedRealityControllerAttribute Find(Type type)
         {
+#if WINDOWS_UWP && !ENABLE_IL2CPP
+            // Type.GetCustomAttributes() doesn't exist on UWP .NET, so we have to indirectly use
+            // the TypeInfo instead.
+            return type.GetTypeInfo().GetCustomAttributes(typeof(MixedRealityControllerAttribute), true).FirstOrDefault() as MixedRealityControllerAttribute;
```
andreiborodin: What we generally did was implement a simple extension to Type that would look like this:

#if WINDOWS_UWP
public static TypeInfo AsTypeInfo(this Type type) => return type.GetTypeInfo();
#else
public static Type AsTypeInfo(this Type type) => return type;
#endif

This way, you could just us AsTypeInfo() through out and not worry

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/TypeResolution/InteractableTypeFinder.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// </summary>
+        private static List<InteractableType> GetTypesFromAssemblies(Type type, TypeRestriction typeRestriction, Assembly[] assemblies)
+        {
+            List<InteractableType> interactableTypes = new List<InteractableType>();
+            foreach (Assembly assembly in assemblies)
+            {
+                foreach (Type assemblyType in assembly.GetTypes())
+                {
+                    TypeInfo info = assemblyType.GetTypeInfo();
+                    bool exactBaseMatch = typeRestriction == TypeRestriction.AllowBase && assemblyType.Equals(type);
```
andreiborodin: I may be misreading this, but isn't this just adding the Type you are searching for to the list; and since Types are unique there will be just this one added, so this check could be omitted.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/TypeResolution/InteractableTypeFinder.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// A helper that uses reflection to find objects that implement base types of the
+    /// Interactable types that populate the various state, theme, and event inspectors.
+    /// </summary>
+    public class InteractableTypeFinder
+    {
+        /// <summary>
+        /// Controls the behavior of the InteractableTypeFinder.FindTypes function. See individual
+        /// enum values for more details.
+        /// </summary>
+        public enum TypeRestriction
```
davidkline-ms: is this enum intended to be referenced by client code? if so, our guidelines ask for a separate code file. If not, please mark as private.

thanks!

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/TypeResolution/InteractableTypeFinder.cs
```diff
@@ -0,0 +1,140 @@
...
+        private static List<InteractableType> GetTypesFromAssemblies(Type type, TypeRestriction typeRestriction, Assembly[] assemblies)
+        {
+            List<InteractableType> interactableTypes = new List<InteractableType>();
+            foreach (Assembly assembly in assemblies)
+            {
+                foreach (Type assemblyType in assembly.GetTypes())
+                {
+                    TypeInfo info = assemblyType.GetTypeInfo();
+                    bool exactBaseMatch = typeRestriction == TypeRestriction.AllowBase && assemblyType.Equals(type);
+                    bool derivedMatch = info.BaseType != null && info.BaseType.Equals(type);
```
andreiborodin: There is info.IsSubclassOf(type) that you can use here, as it may not be immediate base class.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/TypeResolution/InteractableTypeFinder.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// <summary>
+        /// Ensures a cache entry is setup for all types in the InteractableType enum.
+        /// </summary>
+        /// <remarks>
+        /// Note that this is not invoked at runtime and is assumed to be invoked from a single
+        /// threaded UI context, and is thus not locked.
+        /// </remarks>
+        private static void EnsureCacheForTypes(List<Type> types, TypeRestriction typeRestriction)
+        {
+            List<Type> cacheMisses = new List<Type>();
```
andreiborodin: Use HashSet<Type> here

Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
```diff
@@ -60,9 +61,18 @@ public class MixedRealityControllerAttribute : Attribute
         /// <summary>
         /// Convenience function for retrieving the attribute given a certain class type.
         /// </summary>
+        /// <remarks>
+        /// This function is only available in a UnityEditor context.
```
davidkline-ms: I don't see a check for editor in this file. Did I miss something?

Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
```diff
@@ -5,6 +5,7 @@
 using Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities;
 using System;
 using System.Linq;
+using System.Reflection;
```
davidkline-ms: Is this only used for editor code? If so, please consider wrapping in an editor preprocessor check

Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
davidkline-ms: Good eye, @andreiborodin! 

We do have some extensions (ReflectionExtensions.cs) and I believe this may be already provided.

Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
wiwei: Added to the .NET backend ifdef down below.

Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
wiwei: Nah, I think I had merged my change with some other commit I had (i.e. I had a cherry pick which unity editor-ed this entire thing out). I opted to go with this approach instead since there actually was an alternative here.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/TypeResolution/InteractableTypeFinder.cs
wiwei: While it's not going to be referenced by customer client code (i.e. this is an implementation detail of how the MRTK guts looks up these values), it's still located in a location where labelling it private will mean the other internal code won't be able to access it. I've gone ahead with its own file.

killerantz:
"My hope is that I might be able to just write a migrator script to make this easier." If this is possible, please?

killerantz:
I really like the approach, with once source and format for types. Definitely and improvement.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/TypeResolution/InteractableTypeFinder.cs
wiwei: sg, will do that instead. This was taken from the original code.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/TypeResolution/InteractableTypeFinder.cs
wiwei: Yeah, you're reading this correctly. This was mostly a translation from the original code, which handled both the base case and the derived case in the same if statement. I just moved this out of this block.

Assets/MixedRealityToolkit/Attributes/MixedRealityControllerAttribute.cs
wiwei: Done, thanks for the pointer!

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/TypeResolution/InteractableTypeFinder.cs
wiwei: Done, thanks

wiwei:
> "My hope is that I might be able to just write a migrator script to make this easier." If this is possible, please?

I probably need to find/write a unity asset parser to be able to automate this, and it would only work in cases where things are in the unity binary format.

blgrossMS:
@davidkline-ms and @wiwei, we are getting compilation issues still even with this change:
image

Is there something else we have to change in our build settings?

Issue #1769, opened 2018-02-27
Enable Single Pass Stereo Rendering as Default


Cameron-Micka:
## Overview
Apologies if this has been suggested before, but I noticed that the stereo rendering method in the UWP player settings is set to "Multi Pass." I would recommend the default behavior be set to Single Pass (and Single Pass Instanced in the future).

Note, this won't be a performance improvement for most of the example scenes within the MRTK, but complex "production scenes" will likely see a performance improvement. We will also have to update a handful of shaders that are lacking stereo single pass support.

## Unity Editor Version
2017.2.1f1

## Mixed Reality Toolkit Release Version
Tested on [Dev_Working_Branch](https://github.com/Microsoft/MixedRealityToolkit-Unity/commit/a66dba1bfcaf941f604d75025ad1df94d232a908)

Cameron-Micka:
I can do that! I know the MixedReality/Standard shader works with single pass, but I'll want to check the existing shaders as well (I should have some time to test them out tomorrow).

wiwei:
Closing issues older than 180 days. If this is still an issue, please reactivate with recent information.

PR #3632, opened 2019-03-14
Updates the default MRTK project settings to use Single Pass Instance Rendering

wiwei:
As part of working with a number of customers, we've found that many people are using the default Multi Pass rendering mechanism, which is a slow way of rendering.  We've seen in more complex scenarios that this can lead to slowdowns of 5ms per frame.

We should also author a perf doc to recommend setting this value in customers' projects (along with other things like disabling the included visual profiler when doing perf measurements)

### DIFF
--- a/ProjectSettings/ProjectSettings.asset
+++ b/ProjectSettings/ProjectSettings.asset
@@ -46,7 +46,7 @@ PlayerSettings:
   defaultScreenHeight: 768
   defaultScreenWidthWeb: 960
   defaultScreenHeightWeb: 600
-  m_StereoRenderingPath: 0
+  m_StereoRenderingPath: 2
   m_ActiveColorSpace: 0
   m_MTRendering: 1
   m_StackTraceTypes: 010000000100000001000000010000000100000001000000


wiwei:
#1769

wiwei:
@Cameron-Micka 

Looking at issue #1769, I noticed that this was suggested long ago - did you ever end up checking those shaders against the single pass instanced setting?

StephenHodgson:
Does TextMeshPro support this yet?

Cameron-Micka:
Hi wiwei, the MRTK/Standard shader works will all of the stereo rendering paths (as do all the “built in” Unity shaders. But, Stephen is right Text Mesh Pro still lacks official support. I have local changes to Text Mesh Pro which we should figure out the best way to get into the MRTK so we can enable single pass instanced.

wiwei:
Closing this out with some more information - Cameron ended up adding TMP support with some changes in:

https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development_hl2/Assets/MixedRealityToolkit/StandardAssets/Shaders/MixedRealityTextMeshPro.shader

I'll have to defer to Cameron for the details.

PR #3635, opened 2019-03-15
Introduced a Pre Services Update step for the registered services

Alexees:
Overview
InputDevices need to move prior to the FocusProvider performing all the Raycasts. This introduces this necessary step.

Changes
---
- Fixes part of #3614


Assets/MixedRealityToolkit/Providers/BaseDeviceManager.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Providers
     /// <summary>
     /// Base Device manager to inherit from.
     /// </summary>
-    public class BaseDeviceManager : BaseExtensionService, IMixedRealityDeviceManager
+    public abstract class BaseDeviceManager : BaseExtensionService, IMixedRealityDeviceManager
```
SimonDarksideJ: Why change the base implementation to an Abstract. Shouldn't do this unless there is a concrete reason as Abstraction adds a small but noticeable overhead.

Alexees:
@SimonDarksideJ If there's still something wrong with it you'd need to explain. I went over the scripts and reverted what was unnecessary for this.

Assets/MixedRealityToolkit/Providers/BaseDeviceManager.cs
Alexees: @SimonDarksideJ would you be so kind as to post a link or something confirming using abstract inheritance is slower than normal inheritance? I couldn't find anything about it.

Alexees: > Overall this seems sensible but needs a full round of performance validation and testing.

@SimonDarksideJ I agree here clearly. From what I can tell there are quite a couple of calls that are meant to alter the position of things. This is getting tricky when talking about let's say mouse delta, or scroll delta.
I guess someone would need to go through there and filter by what is relevant for the raycasting phase to be correct for all input devices. If that's even all there is to it.

Assets/MixedRealityToolkit/Providers/BaseDeviceManager.cs
```diff
@@ -67,5 +67,7 @@ protected virtual IMixedRealityPointer[] RequestPointers(SystemType controllerTy
 
             return pointers.Count == 0 ? null : pointers.ToArray();
         }
+
+        public virtual void PreServiceUpdate() { }
```
StephenHodgson: I think something like `BeforeUpdate` or something similar?

The update is driven by Unity's game loop events anyway. They're not really specific to services.

Anyone have thoughts on a name?

Railboy: `EarlyUpdate` exists in Unity's experimental API.

Railboy:
Based on discussion in #3614 it looks like this PR should be closed. Is that correct or am I misinterpreting everyone's conclusions?

Alexees:
@Railboy I think so. But how it's going to be implemented remains. Will @luis-valverde-ms be doing this? or is it just switching the order of the two service lists?

luis-valverde-ms:
@Alexees , we will eventually (within a few weeks, I hope) get to it but if it is blocking you, you're welcome to give it a try.

PR #3638, opened 2019-03-15
Mrtk avoid unity editor bug

Alexees:
Overview
The Unity Editor still detects a motion controller during the first frame when it was active during the last run regardless of it being now turned off or not. You'd need a second run to start with no controllers detected at all. Actual detection always happens a couple of frames later.
It's save to ignore the detection during the first frame altogether

Unity case 1137312

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
@@ -269,6 +269,8 @@ namespace Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality
 
             interactionmanagerStates = InteractionManager.GetCurrentReading();
 
+            // Avoids a Unity Editor bug detecting a controller from the previous run during the first frame
+#if !UNITY_EDITOR
             // NOTE: We update the source state data, in case an app wants to query it on source detected.
             for (var i = 0; i < interactionmanagerStates?.Length; i++)
             {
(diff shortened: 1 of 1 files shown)


chrisfromwork:
This seems pretty straight forward. I'm not the most informed on toolkit code standards around macro usage. Hopefully additional reviewers can provide more insights

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -269,6 +269,8 @@ public override void Enable()
 
             interactionmanagerStates = InteractionManager.GetCurrentReading();
 
+            // Avoids a Unity Editor bug detecting a controller from the previous run during the first frame
+#if !UNITY_EDITOR
```
StephenHodgson: Why not just delay the initialization of this service instead of changing the code?

Railboy: What would that look like?

thalbern:
@alexees this is causing the WMR controllers to not being displayed until you trigger the windows button. I was trying to repro your problem without your fixes but I never get into a state where the controllers are still displayed.
Maybe you can help me with my repro steps.
- turned on WMR motion controllers, made sure they're visible in the cliffhouse
- opened and hit play on HandInteractionExample scene of MRTK
- observed controller representations are rendered 
- stopped unity play mode
- turned off motion controllers
- hit play button again in HandInteractionExample scene
- observed controller representations are gone

In my setup I only get callbacks for InteractionSourceDetected in frame 1 - so there's no callback later on like you described, only if I hit the Windows button on the controllers. (testing with unity 2018.4, MRTK RC2 stabilization branch)

thalbern:
just got a repro - i didn't see the models bc they were in my start position

PR #3640, opened 2019-03-15
Service registrar interface (#3545 part 1)

davidkline-ms:
Overview
---
* Add IMixedRealiyServiceRegistrar interface
* Update MixedRealityToolkit component to implement the new interface
* Improve consistency with regards to static vs instance methods/properties (ex: now need .Instance for both Register and Unregister calls)
* Clean up some unused method args 

Verified using Solver demo scene and unit tests

NOTE: You may need to delete the Library folder after taking this change.

[Added for clarity]
This is the first in a series of changes related to providing flexibility for customers to implement their own service locator object / service containers.


Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -39,7 +39,7 @@ float IMixedRealityFocusProvider.GlobalPointingExtent
         {
             get
             {
-                if (MixedRealityToolkit.HasActiveProfile &&
+                if (MixedRealityToolkit.Instance.HasActiveProfile &&
```
chrisfromwork: Can instance be null?

davidkline-ms: If memory serves, if Instance is null and the component is in the scene, it will be instantiated. The core pattern has not changed (see other examples of calling through .Instance).

Thanks!

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -25,20 +25,20 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
     /// The Profile can be swapped out at any time to meet the needs of your project.
     /// </summary>
     [DisallowMultipleComponent]
-    public class MixedRealityToolkit : MonoBehaviour
+    public class MixedRealityToolkit : MonoBehaviour, IMixedRealityServiceRegistrar
```
chrisfromwork: This change makes it possible to have a different class implement IMixedRealityServiceRegistrar. However, it doesn't seem to fix the fundamental problem that a MixedRealityToolkit item needs to exist. Is the point of this review to better abstract service registration? Or are we actually wanting to enable an ability to setup a custom registrar. I would say this review addresses the former. But imo, without doing the deep scrub to remove MixedRealityToolkit static calls for instance, we're still setting up a weird platform architecture.

chrisfromwork: Until Instance is of type IMixedReailtyServiceRegistrar compared to MixedRealityToolkit, I don't think we've actually achieved a componentization story.

davidkline-ms: This is the first of several PRs. The goal of this one is to be as seamless as possible for current code. The next one will pass an instance of IMixedRealityServiceRegistrar to the service instances to eliminate the need to call MixedRealityToolkit.Instance explicitly

davidkline-ms: The complete set of changes for #3545 exceeds the guidelines for small, targeted PRs :)

chrisfromwork: If thats a second review, I can sign off on this first wave of changes. But if this isn't planned work, Im not sure what we are intending to enable through this interface definition.

davidkline-ms: Yes, this is definitely planned work.

The intent here is to get the interface and the first implementation reviewed. I should have the second one (pass the interface to services and some additional static->instance changes) 
prepared today.

chrisfromwork: Long term i think we should remove this instance call.

davidkline-ms:
> Long term i think we should remove this instance call.

Which instance call? MixedRealityToolkit.Instance? If so i agree and am in the process of making that change

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
```diff
@@ -0,0 +1,143 @@
...
+    public interface IMixedRealityServiceRegistrar
+    {
+        #region IMixedRealityService registration
+
+        /// <summary>
+        /// Registers a service of the specified type.
+        /// </summary>
+        /// <typeparam name="T">The interface type of the service to be registered (ex: IMixedRealityBoundarySystem).</typeparam>
+        /// <param name="serviceInstance">An instance of the service to be registered.</param>
+        bool RegisterService<T>(T serviceInstance) where T : IMixedRealityService;
```
wiwei: Would it be worth leaving a comment in remarks that calling RegisterService twice with the same service type will fail? 

Generally when I see Register/Unregister patterns, they are usually:

1) Register and you get back a handler, which you can use to unregister later.
2) Register and it must be one-per-thing you register, so unregister will either find that or not find that.

I think this design is option 2.

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// <summary>
+        /// Registers a service of the specified type.
+        /// </summary>
+        /// <typeparam name="T">The interface type of the service to be registered (ex: IMixedRealityBoundarySystem).</typeparam>
+        /// <param name="concreteType">The concrete type to instantiate.</param>
+        /// <param name="supportedPlatforms">The platform(s) on which the service is supported.</param>
```
wiwei: Is it safe to assume that for the function above, SupportedPlatforms will pretty much be ALL (i.e. even if I add a new enum type in the future to SupportedPlatforms, does that mean that people who call RegisterService with the function above will implicitly get pulled into supporting those new enum types?)

wiwei: Woops sent too soon - by function above, I meant the first overload that just takes a type :)

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
```diff
@@ -0,0 +1,143 @@
...
+            SupportedPlatforms supportedPlatforms = (SupportedPlatforms)(-1),
+            params object[] args) where T : IMixedRealityService;
+
+        /// <summary>
+        /// Unregisters a service of the specified type.
+        /// </summary>
+        /// <typeparam name="T">The interface type of the service to be unregistered (ex: IMixedRealityBoundarySystem).
+        /// <param name="name">The name of the service to unregister.</param>
+        /// <returns>True if the service was successfully unregistered, false otherwise.</returns>
+        /// <remarks>If the name argument is not especified, the first instance will be unregistered</remarks>
```
wiwei: Out of curiousity, what's the use case for not specifying the name? I read this comment as saying that things will be unregistered in FIFO pattern. I can see the coordination here being tricky (i.e. this allows for scenarios where the order in which different components call unregister can get tricky).

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
chrisfromwork: Probably should call out what our return behavior is in this summary. I don't have a strong preference on whether or not a duplicate registration returns true or false. But we should be clear to users what the expected behavior is.

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
chrisfromwork: its probably safer to require explicit platform declarations compared to having a global all. But its kind of tedious making these sorts of updates.

I would imagine that either choice could cause issues:
1) Without global all - Someone starts using a new platform, a change is required to all services to start supporting said platform 
2) With global all - Someone starts using a new platform, first use of specific services results in needing to remove platform support for said service. This removes global all/all additional platforms again start requiring an update to this specific service

It kind of seems like 1 is safer even if its more leg work. 2 seems like it would generate more frustration for developers/may generate stranger behaviors when out in the wild.

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
```diff
@@ -0,0 +1,143 @@
...
+        /// <returns>True if the service was successfully unregistered, false otherwise.</returns>
+        bool UnregisterService<T>(T serviceInstance) where T : IMixedRealityService;
+
+        /// <summary>
+        /// Checks to see if a service of the specified type has been registered.
+        /// </summary>
+        /// <typeparam name="T">The interface type of the service (ex: IMixedRealityBoundarySystem).
+        /// <param name="name">The name of the service.</param>
+        /// <returns>True if the service is registered, false otherwise.</returns>
+        bool IsServiceRegistered<T>(string name = null) where T : IMixedRealityService;
```
wiwei: Similar question to the one above, what's the use case for having a null name (i.e. is this "is any service registered?")

wiwei: Woooooops sent too soon again, followup here is trying to understand the use case for asking that question (i.e. who would use "is any service registered?")

davidkline-ms:
Re: @wiwei comments (for some reason, i cannot reply). The goal of this PR was to leave the core behavior of the current service locator as unchanged as possible. 

These suggetions should _definitely_ be incorportated into future PRs in this series.

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
davidkline-ms: It;s a good way to check to see if, for example, the boundary system is running.

The call would look like

```c#
if (registrar.IsServiceRegistered<IMixedRealityBoundarySystem>()) { ... }
```

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
davidkline-ms: core systems are un-named.

one of the goals of this pr is to limit behavioral change in the service locator. 

the team is definitely willing to entertain thoughts of how to make improvements to the existing locator, feel free to file issues and they can be addressed in the not too distant future

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
davidkline-ms: this is the current service locator behavior. it was a primary goal to minimize functional change when applying this PR.

the team is definitely willing to entertain thoughts of how to make improvements to the existing locator, feel free to file issues and they can be addressed in the not too distant future

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
davidkline-ms: agreed. one of the goals of this change was to limit behavioral change.

the team is definitely willing to entertain thoughts of how to make improvements to the existing locator, feel free to file issues and they can be addressed in the not too distant future

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
wiwei: Got it - I think what I was trying to wrap my head around was, I didn't see the other register calls having a name option, so I wasn't sure how this name would be populated to the registrar.

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        /// <summary>
+        /// Gets the instance of the registered service.
+        /// </summary>
+        /// <typeparam name="T">The interface type of the service (ex: IMixedRealityBoundarySystem).
+        /// <param name="name">The name of the service.</param>
+        /// <param name="showLogs">Indicates whether or not diagnostic logging should be performed in case of an error</param>
+        /// <returns>The registered service instance as the requested type.</returns>
+        T GetService<T>(string name = null, bool showLogs = true) where T : IMixedRealityService;
```
wiwei: for the showLogs parameter, could you provide a little more background in how this would be used as well? i.e. usually I would imagine that logging is a general system thing (i.e. Logging.enabled) rather than having logging enabled on a per function parameter basis.

i.e. if we wanted to enable logs on other functions, would this mean we'd have to update them to also have a showLogs parameter?

Full transparency, I don't have super deep background in the way things work already, so I'm just more wondering out loud how this works + what patterns we've adopted.

Assets/MixedRealityToolkit.Tests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -402,30 +395,28 @@ public void Test_05_04_UnregisterMixedRealityExtensionServicesByType()
...
+            var extensionService1 = MixedRealityToolkit.Instance.GetService<ITestExtensionService1>();
+            var extensionService2 = MixedRealityToolkit.Instance.GetService<ITestExtensionService2>();
 
             // Validate
             Assert.IsNotNull(extensionService1);
             Assert.IsNotNull(extensionService2);
 
-            var success1 = MixedRealityToolkit.Instance.UnregisterService(typeof(ITestExtensionService1));
-            var success2 = MixedRealityToolkit.Instance.UnregisterService(typeof(ITestExtensionService2));
+            var success1 = MixedRealityToolkit.Instance.UnregisterService<ITestExtensionService1>();
```
wiwei: Looks much nicer than having to pass typeof :)

wiwei: Errr,  "use typeof"

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
davidkline-ms: There was previously an issue with customers seeing _lots_ of debug output if a service could not be acquired (ex: was not loaded). The showLogs param was an attempt to reduce that output to one message when there was a legitimate issue finding the serivce.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -121,23 +121,191 @@ public void ResetConfiguration(MixedRealityToolkitConfigurationProfile profile)
...
+            Type concreteType, 
+            SupportedPlatforms supportedPlatforms = (SupportedPlatforms)(-1), 
+            params object[] args) where T : IMixedRealityService
+        {
+            if (isApplicationQuitting)
+            {
+                return false;
+            }
+
+#if !UNITY_EDITOR
```
wiwei: Future comment for us, for these it might be helpful to add to our Extensions or have a helper class abstract this out. I find that having interspersed #if defs can be tricky visually, when really what the code should read is like:

if (IsPlatformSupported(supportedPlatforms))

And the helper function could abstract away the platform specific issues (i.e. as a caller in RegisterService I really don't care about how to see this check is done, only that it is handled on my behalf.

wiwei: So yeah I don't think we need to make a change in this persay, I just want to see if others agree that the approach described above is agreeable to them.

davidkline-ms: great thing to add in the future. would be additive not a breaking change :)

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -121,23 +121,191 @@ public void ResetConfiguration(MixedRealityToolkitConfigurationProfile profile)
         /// <remarks>
         /// Systems can only be registered once by <see cref="Type"/>
         /// </remarks>
-        public static IReadOnlyDictionary<Type, IMixedRealityService> ActiveSystems => activeSystems;
+        public IReadOnlyDictionary<Type, IMixedRealityService> ActiveSystems => new Dictionary<Type, IMixedRealityService>(activeSystems) as IReadOnlyDictionary<Type, IMixedRealityService>;
 
         private static readonly List<Tuple<Type, IMixedRealityService>> registeredMixedRealityServices = new List<Tuple<Type, IMixedRealityService>>();
```
wiwei: Whenever I see global statics being modified, I tend to worry about concurrency - it might be worth leaving a comment stating explicitly that we don't need to lock/guard against concurrency access because all callers will access in a single-threaded fashion.

Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs
wiwei: Got it - I see the existing code which had this.

I think for future I would probably try to push us to augment our logging to have different criticalities (i.e. info, warn, error) and do filtering based on that so that we can avoid littering the log with more noisy stuff but still switch it on in cases where we want to debug further.

But yeah, this is a separate thing from this PR :)

wiwei:
Overall lgtm, not having to do typeof so often looks much nicer :)

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: interesting point. the primary goal for removing the statics is consistency and reducing developer confusion (ex: dont i need an instance to be able to query for data)

davidkline-ms:
first part of #3545 completed.

StephenHodgson:
Send kinda fast getting these PRs merged before the community can review them 😏

PR #3643, opened 2019-03-17
Reduces redundant code in MixedRealityToolkit

Alexees:
Overview
A lot of code is copy and pasted. This reduces the code paths to one

### DIFF
--- a/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
@@ -939,8 +939,6 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
                 execute(system.Value);
             }
 
-            activeSystems.Clear();
-
             foreach (var service in registeredMixedRealityServices)
             {
                 execute(service.Item2);
--- a/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
@@ -56,6 +56,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
             }
         }
 
+        private bool HasProfileAndIsInitialized => activeProfile != null && IsInitialized;
+
         /// <summary>
         /// The active profile of the Mixed Reality Toolkit which controls which services are active and their initial configuration.
         /// *Note configuration is used on project initialization or replacement, changes to properties while it is running has no effect.
@@ -137,7 +139,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
         /// <inheritdoc />
         public bool RegisterService<T>(T serviceInstance) where T : IMixedRealityService
         {
-            return RegisterServiceInternal<T>(serviceInstance);
+            return RegisterServiceInternal(serviceInstance);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -891,129 +893,60 @@ public void DisableAllServicesByTypeAndName(Type interfaceType, string serviceNa
...
+            if (!HasProfileAndIsInitialized) { return false; }
 
-            // Destroy all systems
             foreach (var system in activeSystems)
             {
-                system.Value.Destroy();
+                execute(system.Value);
             }
 
             activeSystems.Clear();
```
wiwei: With this change, activeSystems.clear() is now invoked for all paths (i.e. Enable, Update) - if the intent is to keep the original flow I think that this line needs to be removed.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -891,129 +893,60 @@ public void DisableAllServicesByTypeAndName(Type interfaceType, string serviceNa
...
-            {
-                service.Item2.Disable();
-            }
+            ExecuteOnAllServices(service => service.Disable());
         }
 
         private void DestroyAllServices()
         {
-            // If the Mixed Reality Toolkit is not configured, stop.
-            if (activeProfile == null) { return; }
```
wiwei: In the new code, I think that this activeProfile check is missing.

wiwei:
This is a great change :) Thanks for doing this!

Just a couple of minor comments around some of the code that was shuffled around

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
Alexees: yes, that was a mistake

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
Alexees: It's replaced by HasProfileAndIsInitialized within ExecuteOnAllServices, which first checks for the activeProfile

Alexees:
> This is a great change :) Thanks for doing this!
> 
> Just a couple of minor comments around some of the code that was shuffled around

@wiwei thanks :)
Right now I really love to do this, really satisfies me

davidkline-ms:
> Right now I really love to do this, really satisfies me

We feel the same! Glad you are part of the community!

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -891,129 +893,58 @@ public void DisableAllServicesByTypeAndName(Type interfaceType, string serviceNa
...
+            if (!ExecuteOnAllServices(service => service.Destroy())) { return; }
 
-            // If the Mixed Reality Toolkit is not initialized, stop.
-            if (!IsInitialized) { return; }
+            activeSystems.Clear();
+            registeredMixedRealityServices.Clear();
+        }
+
+        private bool ExecuteOnAllServices(Action<IMixedRealityService> execute)
```
StephenHodgson: What was the decision to use an action instead of just calling the methods directly?

This would cut down on the runtime overhead and allocations. Good to remember that much of this code will run on mobile devices.

Have we done a perf comparison?

SimonDarksideJ: Agreed, due to the tight performance marks on the operations of the services, Actions should be avoided unless absolutely necessary

Alexees: Are you sure you're not over-optimizing this at the expense of readability and maintainability? Those lambdas are executed not even 20 times per frame.
Just found this Stackoverflow thread which tested performance differences between various possible ways of calling methods and the numbers are so tiny, I don't think they really matter.

https://stackoverflow.com/questions/35805609/performance-of-expression-compile-vs-lambda-direct-vs-virtual-calls

If it does matter, I'd like to know how you guys test code performance as this would be critical to know for everyone who wants to participate in improving MRTK.

Railboy: Agree perf is critical in services but I would need to see a stopwatch test showing that this approach actually has a measurable impact before ruling it out.

Alexees: Was wondering if this receives a yes or nor while I was cleaning up my branches

davidkline-ms: @Alexees, the team was very busy preparing the RC1 release and this PR was postponed.

I will personally pull and test your changes this week.

Thanks!

PR #3647, opened 2019-03-18
Enable using networking component for multiple components + support multiple scenes

chrisfromwork:
Overview
---
Spectator view has had some limitations exposed by community members:
1) Its not easy to use the current networking component for sending state updates other than for the marker spatial coordinate system payloads
2) Its not clear how to get spectator view working with multiple scenes

This review adds some functionality to begin improving this experience:
1) The networking interface now has a priority flag for sending payloads. Non spatial coordinate system payloads can be marked as critical so that they are forced to send.
2) A SceneRoot spectator view object was created so that when switching scenes. This makes it possible for the SceneRoot to be dynamically set when switching between scenes.
3) SpectatorView now calls DontDestroyOnLoad for its game object so that SpectatorView components remain across scenes.

Changes
---
- Fixes: # .


chrisfromwork:
@andreiborodin and @Railboy these are some quick changes to try and provide guidance to a community member. We may want to build out this example scene to have multiple scenes/prove out that the critical message path can be used for scene transitions. What are your twos thoughts on these interface changes?

andreiborodin:
> using UnityEngine;

nit: missing copyright


---
Refers to: Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/SceneRoot.cs:1 in 5c5c30a. [](commit_id = 5c5c30ab3f5b515a8778664ae3d1f0efcf479d72, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/SceneRoot.cs
```diff
@@ -0,0 +1,20 @@
+﻿using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.SpectatorView
+{
+    public class SceneRoot : MonoBehaviour
+    {
+        void Start()
```
andreiborodin: nit: default modifier

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/SpectatorView.cs
```diff
@@ -23,6 +20,8 @@ public class SpectatorView : MonoBehaviour
         [SerializeField] MonoBehaviour RecordingServiceVisual;
         [SerializeField] List<MonoBehaviour> PlayerStateObservers;
 
+        [HideInInspector] public Matrix4x4 LocalOriginToSharedOrigin = Matrix4x4.identity;
```
andreiborodin: I suppose these aren't expected to be serialized as well?

I would make these properties

chrisfromwork: Are properties not serialized in unity scenes?

andreiborodin: Nope, only fields, public or marked SerializeField

---
In reply to: [266633797](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3647#discussion_r266633797) [](ancestors = 266633797)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Interfaces/INetworkingService.cs
```diff
@@ -3,11 +3,17 @@
 
 namespace Microsoft.MixedReality.Toolkit.Extensions.SpectatorView.Interfaces
 {
+    public enum Priority
```
Railboy: Suggestion `public enum NetworkPriority`

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Interfaces/INetworkingService.cs
```diff
@@ -3,11 +3,17 @@
...
+        Default,
+        Critical
+    }
+
     public delegate void DataHandler(string playerId, byte[] payload);
 
     public interface INetworkingService
     {
-        bool SendData(byte[] data);
+        bool SendData(byte[] data, Priority priority);
```
Railboy: Suggestion `bool SendData(byte[] data, Priority priority = Priority.Default);`

Railboy:
>What are your twos thoughts on these interface changes?

I think this is an incremental step in the right direction.

We're ill-equipped to deal with multiple scenes across the board, not just in spectator view. We need to accumulate a few service-specific solutions like this before we can figure out a general strategy.

PR #3661, opened 2019-03-20
Fix default port values for udp broadcast ui

chrisfromwork:
Overview
---
small fix

Right now if you type in invalid numbers for this ui dialogue, the default port values get set to a number that isn't the udp broadcasting components default value. This should make the process less confusing.

Changes
---
- Fixes: # .


chrisfromwork:
@provencher i'm having issues adding you as a reviewer, have you gone through the msft license/cla process?

provencher:
@chrisfromwork I have not. Could you point me to a link so that I could go through this process?

chrisfromwork:
I don't understand why this license/cla failed. Regardless, this change should now be in.

Issue #3590, opened 2019-03-09
Proposal: Webrtc Extension


bengreenier:
## Overview

This is a proposal (and intent to implement) basic webrtc facilities as an [MRTK extension](https://github.com/Microsoft/MixedRealityToolkit-Unity/wiki/Creating-a-new-Extension). The goal is to enable audio, video, and data channels for use from managed code, without the need to write custom native plugins (however, do note that native plugins will be implemented under the hood for the caller, as they are necessary to bind to webrtc on a given target platform). 

The goal at this time is to begin with support the following targets:
+ Win32 Standalone
+ Android
+ WinUWP (non-ARM at this time)

Support for other targets (with ARM being the likely next candidate) can be added in the future. The planned approach will mirror [the webrtc unityplugin example](https://webrtc.googlesource.com/src/+/refs/heads/master/examples/unityplugin/), and will be largely a refactor from [bengreenier/webrtc-unity-plugin](https://github.com/bengreenier/webrtc-unity-plugin) which supports webrtc (`m71`) facilities and includes some examples.

The native plugin work (a slight modification from google source, to support WINUWP and il2cpp) is based on [webrtc-uwp](https://github.com/webrtc-uwp), and can be seen in [this PR](https://github.com/webrtc-uwp/webrtc/pull/11).

In my opinion it's worth including a disclaimer here that __this is a basic, best-effort solution__ and is not intended for high-scale production use at this time, but is designed to get us *closer* to a production story. I'm thinking of it as one of the core pieces in a larger puzzle. Hopefully that makes sense to folks.

### Managed code

MIT licensed by Microsoft CSE team.

### Native Plugins

Webrtc is licensed pretty [permissively](https://webrtc.org/license/), and the example codebase on which this work is built is [explicitly licensed](https://webrtc.googlesource.com/src/+/refs/heads/master/examples/unityplugin/simple_peer_connection.h#5) the [same way](https://webrtc.googlesource.com/src/+/refs/heads/master/LICENSE). Hopefully this makes it relatively straightforward for us to depend on the native code that makes this possible.

anders-lundgren:
[...]

PR #3664, opened 2019-03-20
Add a basic webrtc extension (#1)

bengreenier:
* Add a basic webrtc extension ☎️ 
* Source for this extension is derived from https://github.com/bengreenier/webrtc-unity-plugin
  - Which is, in turn, dependent on the https://github.com/webrtc-uwp/webrtc-uwp-sdk project, specifically the https://github.com/webrtc-uwp/webrtc/tree/releases/m71/examples/unityplugin example codebase. We have also made some modifications to that example to support uwp.

Overview
---

This addresses #3590 by providing a basic extension to encapsulate webrtc a/v/d capabilities on `winuwp` and `win32` platforms. It includes a basic signaling example based on https://github.com/bengreenier/node-dss, an example scene (and prefab containing the actual example object logical layout).

We had some concerns about making sure we do the right thing on the licensing side as well (see original proposal) so we'll need to ensure we've done that work before we take this.

@vladkol this includes the signaling interface as discussed - please let me know if that meets your needs or needs further adjustment.

Changes
---
+ Rebased from https://github.com/bengreenier/MixedRealityToolkit-Unity/commits/feature/WebRTC - that individual commit context may be useful for reviewers, unsure.


Assets/MixedRealityToolkit.Extensions/Webrtc/Plugins/Win32/x64/webrtc_unity_plugin.dll.meta
```diff
@@ -0,0 +1,128 @@
...
+  serializedVersion: 2
+  iconMap: {}
+  executionOrder: {}
+  defineConstraints: []
+  isPreloaded: 0
+  isOverridable: 0
+  isExplicitlyReferenced: 0
+  platformData:
+  - first:
+      '': Any
```
chrisfromwork: we should probably avoid having *.dll.meta files as well. Nuget for unity won't drop dlls into this path/you'll want to include meta files in the nuget packages that you distribute with your binaries if you ever distribute nuget packages.

Assets/MixedRealityToolkit.Extensions/Webrtc/Plugins/.gitignore
```diff
@@ -0,0 +1 @@
+*.dll
```
chrisfromwork: do you think that you will want to track anything in the plugin directory? I know that long term there could be some android components. But it may be worth removing all meta files and the plugins directory itself for this check in.

chrisfromwork:
> 

nit: these are currently tabs, you should convert to spaces in vs by selecting the content, Edit -> Advanced -> Untabify selected lines

---
Refers to: Assets/MixedRealityToolkit.Extensions/Webrtc/Materials/YUVFeedShader.shader:31 in 0322617. [](commit_id = 032261778f41733fb0bef4c1c1e7c86902c9d228, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/IPeer.cs
```diff
@@ -0,0 +1,117 @@
+﻿using System;
+using Microsoft.MixedReality.Toolkit.Extensions.Webrtc.Delegates;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Webrtc
+{
+    public interface IPeerImpl
```
chrisfromwork: >IPeerImpl [](start = 21, length = 9)

Is there a reason that IPeeImpl doesn't match the file name?

chrisfromwork: >IPeerImpl [](start = 21, length = 9)

nit: i think we should have different files for different interfaces.  Also is there a reason for the Impl ending compared to another term that may be more telling to why there are two interfaces?

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Webrtc.cs
```diff
@@ -0,0 +1,369 @@
...
+#if UNITY_WSA && !UNITY_EDITOR
+using Windows.UI.Core;
+using Windows.Foundation;
+using Windows.Media.Core;
+using Windows.Media.Capture;
+using System.Threading.Tasks;
+using Windows.ApplicationModel.Core;
+#endif
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Webrtc
```
chrisfromwork: >Webrtc [](start = 52, length = 6)

nit: given RTC is an acronym it may be worth capitalizing. After this initial review is checked in it will be difficult to change.

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Webrtc.cs
```diff
@@ -0,0 +1,369 @@
...
+                return;
+            }
+
+#if UNITY_ANDROID
+            AndroidJavaClass systemClass = new AndroidJavaClass("java.lang.System");
+            string libname = "jingle_peerconnection_so";
+            systemClass.CallStatic("loadLibrary", new object[1] { libname });
+            Debug.Log("loadLibrary loaded : " + libname);
+
+            /*
```
chrisfromwork: >/* [](start = 12, length = 2)

nit: fix spacing.

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerEvents.cs
```diff
@@ -0,0 +1,389 @@
...
+        /// <summary>
+        /// Internal queue used to marshal work back to the main unity thread
+        /// </summary>
+        private Queue<Action> MainThreadWorkloads = new Queue<Action>();
+
+        /// <summary>
+        /// Initialization handler
+        /// 
+        /// TODO(bengreenier): This could be auto-bound with editor script
+        /// </summary>
```
chrisfromwork: nit: don't check in todos. Open a separate issue for this todo or remove the code/call out what is not implemented with no intention of fixing.

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerVideo.cs
```diff
@@ -0,0 +1,158 @@
...
+{
+    /// <summary>
+    /// Provides video frame hooks for Webrtc
+    /// </summary>
+    public class WebrtcPeerVideo : MonoBehaviour
+    {
+        /// <summary>
+        /// Instance of a local video player which will have frame data written to it
+        /// </summary>
+		[Tooltip("Video player instance that local stream data will be written to")]
```
chrisfromwork: >		[T [](start = 0, length = 4)

nit: convert to spaces

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerVideo.cs
```diff
@@ -0,0 +1,158 @@
...
+        public WebRtcVideoPlayer LocalPlayer;
+
+        /// <summary>
+        /// Instance of a remote video player which will have frame data written to it
+        /// </summary>
+		[Tooltip("Video player instance that remote stream data will be written to")]
```
chrisfromwork: >		[To [](start = 0, length = 5)

nit: convert to spaces

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerVideo.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// <summary>
+        /// Internal queue used for storing and processing remote video frames
+        /// </summary>
+        private FrameQueue remoteFrameQueue = new FrameQueue(5);
+
+        /// <summary>
+        /// Initialization handler
+        /// 
+        /// TODO(bengreenier): This could be auto-bound with editor script
+        /// </summary>
```
chrisfromwork: nit: maybe do a code search for your directory to make sure all todos are removed

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerVideo.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        /// <summary>
+        /// Internal yuv data copy routine
+        /// </summary>
+        private void CopyYuvToBuffer(IntPtr dataY, IntPtr dataU, IntPtr dataV,
+        int strideY, int strideU, int strideV,
+        uint width, uint height, byte[] buffer)
+        {
+            unsafe
```
chrisfromwork: >unsafe [](start = 12, length = 6)

unsafe code can't always be adopted by community members. Do you know if adding this script to the mrtk and opening one of the mrtk demos requires allowing unsafe code?

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerVideo.cs
```diff
@@ -0,0 +1,158 @@
...
+            CopyYuvToBuffer(dataY, dataU, dataV, strideY, strideU, strideV, width, height, packet.Buffer);
+            packet.width = (int)width;
+            packet.height = (int)height;
+            remoteFrameQueue.Push(packet);
+        }
+
+        /// <summary>
+        /// Internal yuv data copy routine
+        /// </summary>
+        private void CopyYuvToBuffer(IntPtr dataY, IntPtr dataU, IntPtr dataV,
```
chrisfromwork: >CopyYuvToBuffer [](start = 21, length = 15)

can we gets some comments calling out the specific yuv codec that you are iterating over? or at least the ratio of Y:U:V and how pixel values are laced?

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcUIControls.cs
```diff
@@ -0,0 +1,163 @@
+﻿using System;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Webrtc
+{
+    /// <summary>
+    /// This SHOULD NOT BE USED FOR PRODUCTION.
+    ///
```
chrisfromwork: what does this mean?

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Webrtc.cs
```diff
@@ -0,0 +1,369 @@
...
+        /// Flag to log all errors to the Unity console automatically
+        /// </summary>
+        [Tooltip("This will log all errors to the Unity console")]
+        public bool AutoLogErrors = true;
+
+        /// <summary>
+        /// Set of ice servers
+        /// </summary>
+        [Tooltip("(Optional) Set of Ice servers")]
+        public List<ConfigurableIceServer> IceServers = new List<ConfigurableIceServer>()
```
vladkol: We should support Ice-credentials along with the server list coming from the signaling service as a part of the negotiation/authentication.

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Marshalling/Deque.cs
```diff
@@ -0,0 +1,843 @@
+﻿using DequeUtility;
+
+namespace System.Collections.Generic
+{
+    /// <summary>
+    /// A genetic Deque class. It can be thought of as
```
chrisfromwork: >genetic [](start = 10, length = 7)

nit: generic

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Marshalling/NativePeer.cs
```diff
@@ -0,0 +1,470 @@
...
+    /// </summary>
+    public class NativePeer : IPeer
+    {
+        /// <summary>
+        /// Internal marshalling definitions to the native-library
+        /// </summary>
+        private static class NativeMethods
+        {
+#if UNITY_ANDROID
+            private const string dllPath = "libjingle_peerconnection_so";
```
chrisfromwork: >libjingle_peerconnection_so [](start = 44, length = 27)

is this the standard webrtc binary for android? Or is this built out of some other repo?

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Signaling/ISignaler.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Webrtc.Signaling
+{
+    public interface ISignaler
+    {
+        /// <summary>
+        /// Event that occurs when the signaler receives a new message
+        /// </summary>
+        event Action<SignalerMessage> OnMessage;
```
vladkol: We need a connected/disconnected event here, so we can track the signaling connection state. As part of 'connected' event Ice server list and Ice credentials may come too.

Assets/MixedRealityToolkit.Extensions/Webrtc/Plugins/Win32/x64/webrtc_unity_plugin.dll.meta
bengreenier: ah ok. my thinking was that this makes it clearer that some dlls need to get placed, but you're right that it comes at a cost of "clutter" that likely won't be viable for folks. I'll cleanup.

Assets/MixedRealityToolkit.Extensions/Webrtc/Plugins/.gitignore
bengreenier: A small dev experience thing that i think __is__ nice about having some of this is that the folder structure is already there - makes placing the dlls just a little less painful. 

otherwise, no probably not for the time being.

chrisfromwork: i'd probably opt for not adding the plugins directory here. its probably worth adding to the desription that these files need to go into a plugins directory, but its not super clear where the plugins directory would end up being for most applications. If folks copy your content into their own projects they might already have their own plugins directory that they'd prefer to use.

jwittner:
As this isn't yet targeted for mrtk_development, this is merely a related tangent, please do point me in the right direction if I should pick this up elsewhere.

Now that I'm seeing it happen, I'm really not a fan of putting any Extensions right into the main MRTK repo. Can we not make a MRTK.WebRTC repo and manage this as a first class extension there? Or just bite the bullet and own this thing without the Extension moniker. Even with it, it'll still be confusing what is and isn't maintained or at what levels. If the code/assets are in the repo, they should be revisioned alongside any breaking MRTK changes or the internal consistency of the repo as a whole is broken.

The Extension namespace is kind of lame too and belittles the value of what might be very powerfull tooling that's managed by large engaged communities. Better I think to just publish it as MRTK.WebRTC and just push your artifacts as prerelease versions, or use 0.x.x versions until it's ready to be treated with the same rigor or gets a lot of community adoption.

Making the extensions stand alone will help them move faster, keep the lines of responsibility clear, and empower the community to do more. With the forthcoming NuGet packages, we can show them how to organize their extensions and take a dependency on the MRTK. I think this better helps the ecosystem to thrive and draws much nicer lines by building around and adjacent to the MRTK instead of in it.

@davidkline-ms, thoughts?

jwittner:
As an example, I pulled your PR into the MRTK and then deleted everything that was under Assets/MixedRealityToolkit.Extensions and everything still worked and my automated broken reference testing tool didn't find anything missing. Why couple together this technology with a dependency it doesn't even take?

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/IPeer.cs
bengreenier: not a good one - for clearer code-reuse i stil the (reused) method defs so that both the underlying peer and the unity events wrapper could consume it. will rename and split out to something more logical.

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerVideo.cs
bengreenier: well you've gone and stumped me - because normally i'd agree and expect to see "Allow unsafe code" in player settings would be needed for this to work at all, in the included Example, as well as in other ones (basically anywhere this script is built..) but i'm not seeing that behavior locally...

I've got it unchecked and am able to run `mono` and `il2cpp` backend-based projects, in editor and in build (standalone). perhaps somehow related to this method being private?

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcUIControls.cs
bengreenier: this actually should be removed (deprecated that sample) in favor of clearer WebrtcSignalControls (more production applicable, extendable).

bengreenier:
@chrisfromwork 

> 
> <s>https://webrtc.github.io/samples/src/content/peerconnection/munge-sdp/</s>
> https://github.com/bengreenier/webrtc-unity-plugin
> https://unitylist.com/p/cxl/Web-Rtc-Unity-Plugin-Sample

munge-sdp is now removed (ended up not being a valuable sample for this case, at least at the moment). Need to add:

https://webrtc.googlesource.com/src/+/refs/heads/master/examples/unityplugin

> Do you know if any files included in this change are direct ports from somewhere else? Are most of the scripts in this change things you yourself wrote?

Most are mine, the direct ports are the ones that do not have the Microsoft copyright text.

>You should keep the original namespaces for items that you didn't write. 

I wasn't sure this was the right call for MRTK, as they were in the global namespace...can you advise?

>For all items that are your own original contribution, you should add the Microsoft licensing to the top of each file

Done.

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Webrtc.cs
bengreenier: this is unfortunately not trivial (as the native library also takes information in this odd format at this time). I will book a feature request for it but i will not be able to refactor to this myself at this point in time.

bengreenier:
@chrisfromwork @jwittner Barring the outcome of [this discussion](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3664#issuecomment-475702848) i'm going to hold off filing issues related to this work at this time.

As such, i'll use this comment to capture the items that were recently in-code TODOs, but should become issues/feature-requests should we land this work here.

- [ ] Feature Request: Support Ice credentials on a per IceServer basis (@vladkol)
- [ ] Refactor: NativePeer.GetUniqueId => C# property
- [ ] Editor script DevUX: WebrtcPeerEvents.OnInitialized => autobind to Webrtc.OnInitialized event when the WebrtcPeerEvents component is added in the inspector 
- [ ] Editor script DevUX: WebrtcPeerVideo.OnInitialized => autobind to Webrtc.OnInitialized event when the WebrtcPeerVideo component is added in the inspector 
- [ ] Refactor: WebrtcSignalControls.HeaderAttribute should be replaced with Editor script verification that sibling component is attached
- [ ] Refactor: WebRtcVideoPlayer.FrameQueue has a non-standard lifetime - being allocated and de-alloced in the WebrtcPeerVideo component...this should likely be revised

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Webrtc.cs
bengreenier: captured (and tagged) in comment below, as it is unclear at this time if GH issues here is the right place to capture this work.

vladkol: The format is fine. All we need is a way to accept [TURN Auth message](https://www.ietf.org/proceedings/87/slides/slides-87-behave-10.pdf)

```json
{
  username: "1375043478:abcd1234",
  password: <HMAC("1375043487:abcd1234", SS)>
  ttl: 86400,
  uris: [
    "turn:turn.bar.com:3478?proto=udp",
    "turn:turn.bar.com:3478?proto=tcp",
    "turns:turn.bar.com:443?proto=tcp"
] }
```
I can make necessary changes. Although, it’s important to keep existing properties too, so developer can provide values event if signaling server gives nothing.

bengreenier: > TLDR: sure, that sounds great. I'll look to you for the changset to allow a signaling server to provide this info.

ah - thought you were proposing supporting giving native library something like:

```
NativePeer(List<IceServer> servers)
```

where

```
struct IceServer
{
    string Uri;
    int Port;
    string Username;
    string Credential;
}
```

which imo is a more common way to describe a set of ice servers. Current native lib only supports once `Username` and `Credential` even if N ice servers urls is given.

vladkol:
One set of credentials for a list of servers is how this REST authentication thing is done.

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcSignalControls.cs
```diff
@@ -0,0 +1,203 @@
...
+                        break;
+                    default:
+                        Debug.Log("Unknown message: " + msg.MessageType + ": " + msg.Data);
+                        break;
+                }
+            };
+
+            signaler.OnFailure += (Exception ex) =>
+            {
+                Debug.Log(ex.Message);
```
markgrossnickle: This should update the SignalerStatusLabel instead of (or in addition to) logging.

markgrossnickle:
@bengreenier  Great work! 

I'm currently playing around with this by using it to connect a Hololens to a PC. Perhaps I am doing something wrong or there are still issues being solved at the webrtc level but here are the issues I am seeing:

1. No audio from PC to Device. Audio does work from Device to PC.  (I need to test on more machines to rule out a device issue).
2. Very low performance. 3-4 fps using the default settings of the app.  I was able to get 7 fps if I didn't display the local rendering on the Hololens  and 15 fps if I didn't display the local or the remote rendering on the Hololens (which is our target use case).    I plan on tweaking settings more to see if I can slow down the recording to keep the fps high but don't want to re-invent the wheel if there are already plans in place for performance.  
3. The video stream from the PC to the device gets further and further behind the longer you play.  Seems like it is not dropping frames but instead just playing back all frames at a slow speed?

I made the following changes to get this to work:
- I changed the UI to use world space in order to work with the Hololens
- I added properties to define targetId and deviceId rather than using the unique identifier since it was cumbersome. 

I'm going to keep tinkering and seeing if I can get the audio to work and performance improved.  Thanks again for contributing this extension. It is extremely helpful for an upcoming feature I am writing.

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerVideo.cs
StephenHodgson: It's generally located as a checkbox in the assembly definition

michael-house:
There appear to be two efforts to get WebRTC up and running on Hololens. The repo this PR is derived from:
https://github.com/bengreenier/webrtc-unity-plugin

And another:
https://github.com/webrtc-uwp/webrtc-uwp-sdk

Is the webrtc_unity_plugin.dll included in this PR compiled from the work being done with the webrtc-uwp-sdk? It appeared instead to be based on Google's example: https://webrtc.googlesource.com/src/+/refs/heads/master/examples/unityplugin/

Is there a relationship between these two repos? Will this extension be updated to use webrtc-uwp-sdk eventually?

markgrossnickle:
@michael-house The webrtc-unity-plugin pulls in a dll from webrtc-uwp-sdk.  I'm not sure about the google link you are referencing but I believe all of the webrtc-uwp-sdk is based on a google webrtc library. 

Check this branch out of webrtc-uwp-sdk if you are trying to rebuild the plugin dll:
https://github.com/webrtc-uwp/webrtc/tree/cse/m71-ks

It has a project within it called Examples.UnityPlugin which creates the dll referenced in this pull request.

michael-house:
@markgrossnickle Thank you! I saw mentions of m71, which I knew was a branch in the webrtc-uwp-sdk, but the interfaces looked to more closely match the google webrtc library (which is linked in the "Why" section of the readme for webrtc-unity-plugin). 

I had been working directly with the webrtc-uwp-sdk library to get webrtc into a project. It was struggle given I couldn't test it in Unity, I had to two IL2CPP development builds, which take forever! The webrtc-unity-plugin was easy to get integrated into a project, because I was able to debug in Unity.

Since you've apparently also set this up, maybe you can tell me: The video I'm getting on the desktop from the HoloLens is not a MRC video stream, it's just straight HoloLens camera capture, is that your experience as well? When I was working with the web-rtc-uwp library, I was able to get streaming from the HoloLens to the desktop, and it was a full MRC stream.

markgrossnickle:
I think that is this issue: https://github.com/webrtc-uwp/webrtc-uwp-sdk/issues/130   (So it is being worked on)

markgrossnickle:
Update on my notes above. It was a computer issue on the no Audio I listed above. I am getting audio when using another microphone. 

I tried to change the Interval/fps to be 1 fps to see if I could get better performance but I didn't have much luck. I was in simple_peer_connection.cc and added the code below around line 582.  I confirmed in impl_webrtc_VideoCapture that the FPS was set to 1 during StartCapture.  However, the playback on the PC was faster than 1fps (showed 5+fps).  Probably a user error as I am still getting my head wrapped around this library and c++ in general. 

`std::unique_ptr<cricket::VideoCapturer> capture = OpenVideoCaptureDevice();
    if (capture) {
      webrtc::MediaConstraintsInterface::Constraints* mandatory =
          new webrtc::MediaConstraintsInterface::Constraints();
      webrtc::MediaConstraintsInterface::Constraints* optional =
          new webrtc::MediaConstraintsInterface::Constraints();

      mandatory->push_back(webrtc::MediaConstraintsInterface::Constraint(
          webrtc::MediaConstraintsInterface::kMaxFrameRate, rtc::ToString(1)));
      mandatory->push_back(webrtc::MediaConstraintsInterface::Constraint(
          webrtc::MediaConstraintsInterface::kMinFrameRate, rtc::ToString(1)));

      webrtc::MediaConstraintsInterface* constraints = new wrapper::impl::org::
          webRtc::MediaConstraints::MediaConstraintsImpl(*mandatory, *optional); 

      rtc::scoped_refptr<webrtc::VideoTrackInterface> video_track(
          g_peer_connection_factory->CreateVideoTrack(
              kVideoLabel, g_peer_connection_factory->CreateVideoSource(
                               std::move(capture), constraints)));`

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Marshalling/FrameQueue.cs
```diff
@@ -0,0 +1,115 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.WebRTC.Marshalling
+{
+    // FramePacket のキュー。
+    // 指定サイズ以上Pushすると、指定サイズ以下になるよう末尾のデータから削除される。
+    // スレッドセーフ。
```
andreiborodin: Do we want to update the comment?

andreiborodin:
> using System.Collections.Generic;

Do we want to add the copyrights to these files?

---
Refers to: Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Marshalling/FrameQueue.cs:1 in 701d81c. [](commit_id = 701d81c22a0eece3151d7525d9ab14e7e2e1328a, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/WebrtcPeerVideo.cs
bengreenier: 🤦‍♂️ @StephenHodgson, oh wow, silly me - thanks for calling that out!

@chrisfromwork i'm not sure how to best validate this...here's what i tried:

il2cpp builds of `BoundryVisualization` and `WebrtcBasicSignalExample` seem to both include this code (ctrl+f in generated vs solution) - so i would guess that unsafe flag is always needed when building scripts from my asmdef, which seem to be built in all scenarios.

bengreenier:
@michael-house 

> 
> 
> There appear to be two efforts to get WebRTC up and running. The repo this PR is derived from:
> https://github.com/bengreenier/webrtc-unity-plugin
> 
> And another:
> https://github.com/webrtc-uwp/webrtc-uwp-sdk
> 
> Is the webrtc_unity_plugin.dll included in this PR compiled from the work being done with the webrtc-uwp-sdk? It appeared instead to be based on Google's example: https://webrtc.googlesource.com/src/+/refs/heads/master/examples/unityplugin/
> 
> Is there a relationship between these two repos? Will this extension be updated to use webrtc-uwp-sdk eventually?

https://github.com/bengreenier/webrtc-unity-plugin is managed glue that depends on native code. That native code is a version of https://webrtc.googlesource.com/src/+/refs/heads/master/examples/unityplugin/ that is built from the https://github.com/webrtc-uwp/webrtc-uwp-sdk source tree (which is effectively a webrtc fork with lots of additions for uwp to work). we use that fork so that we can modify the sample slightly to add support for the uwp target. So they are all closely related :)


<br/>

@markgrossnickle 


>   1. No audio from PC to Device. Audio does work from Device to PC. (I need to test on more machines to rule out a device issue).
>  2. Very low performance. 3-4 fps using the default settings of the app. I was able to get 7 fps if I didn't display the local rendering on the Hololens and 15 fps if I didn't display the local or the remote rendering on the Hololens (which is our target use case). I plan on tweaking settings more to see if I can slow down the recording to keep the fps high but don't want to re-invent the wheel if there are already plans in place for performance.
>    3. The video stream from the PC to the device gets further and further behind the longer you play. Seems like it is not dropping frames but instead just playing back all frames at a slow speed?

1. glad to see you got around that. we should take a docs work item to clarify that this path will always use the default communication device (as there is no longer a webrtc api to specify audio device).
2. this is known, @jamescadd do we have an issue tracking holo perf in webrtc-uwp? tagging @phongcao as i know he was investigating some perf fixes for holo v1 on the releases/m71 codepaths (is there any update you can share publically Phong?)
3. hadn't noticed that myself. definitely something we should investigate a fix for or at-minimum track a bug for.

Thanks for taking the time to validate some of this work!

Assets/MixedRealityToolkit.Extensions/Webrtc/Scripts/Marshalling/FrameQueue.cs
bengreenier: I'll look to you/other MRTK folks for this, as [captured above](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3664#issuecomment-475829055) much of this logic is carried over from another project - not sure if it's best practice to modify this or leave as is.

markgrossnickle:
@bengreenier Thanks for the response. I think I may have another issue for your list as I am unable to create the data channel.  

I am attempting to call PeerEventsInstance.AddDataChannel() from within the OnPeerReady.AddListener event (right below where the stream is added). 

I get the following log when trying to connect the data channel:

> (impl_webrtc_videocapturer.cpp:464): CaptureDevice::StartCapture: INTERVAL     1000000000 FPS 1
> 'WebrtcUnityPlugin.exe' (Win32): Loaded 'C:\Program Files\Intel\Media SDK\libmfxhw32.dll'. Cannot find or open the PDB file.
> The thread 0x1420 has exited with code 13430019 (0xcced03).
> The thread 0x146c has exited with code 13430018 (0xcced02).
> The thread 0x1294 has exited with code 13430017 (0xcced01).
> The thread 0x1298 has exited with code 13430016 (0xcced00).
> The thread 0x102c has exited with code 13430016 (0xcced00).
> The thread 0x1460 has exited with code 13430017 (0xcced01).
> The thread 0x1328 has exited with code 13430018 (0xcced02).
> The thread 0x12b0 has exited with code 13430019 (0xcced03).
> 'WebrtcUnityPlugin.exe' (Win32): Unloaded 'C:\Program Files\Intel\Media SDK\libmfxhw32.dll'
> Exception thrown at 0x774B2CE2 (KernelBase.dll) in WebrtcUnityPlugin.exe: WinRT originate error - 0xC00D36B3 : 'The stream number provided was invalid.'.
> (impl_webrtc_videocapturer.cpp:529): CaptureDevice::StartCapture: returning
> (messagequeue.cc:513): Message took 449ms to dispatch. Posted from: webrtc::VideoCapturerTrackSource::Initialize@../../pc/videocapturertracksource.cc:357
> (messagequeue.cc:513): Message took 453ms to dispatch. Posted from: webrtc::PeerConnectionFactoryProxyWithInternal<class webrtc::PeerConnectionFactoryInterface>::CreateVideoSource@C:\projects\Taqtile\webrtc-71-ks2\webrtc-uwp-sdk\webrtc\xplatform\webrtc\api/peerconnectionfactoryproxy.h:59
> The thread 0x1470 has exited with code 0 (0x0).
**> (datachannel.cc:145): Failed to initialize the RTP data channel due to invalid DataChannelInit.
> (simple_peer_connection.cc:631): Fails to create data channel**

I am also not sure what the ''The stream number provided was invalid.'.' is being thrown from but audio/video continues to work even after that is thrown.

jamescadd:
@markgrossnickle Could you confirm that you're running in release mode?  @bengreenier re: performance, I wanted to double check that your work include the H.264 decoder from Kevin (pretty sure it does). @phongcao reports good performance while both encoding and decoding video on HoloLens so I'm not sure why the plugin has issues.

One additional issue to list above - if you're using a dual GPU machine on the PC side you need to force PeerCC to use NVIDIA so that there are no issues.

markgrossnickle:
@jamescadd I've been switching back and forth between release and debug. I'm not sure which test you are referring to but I believe the FPS I recorded was in release.  I am not sure about the delayed playback.  I am out today but could test again tomorrow.  Currently trying to figure out why I can't get the data channel to connect.

chrisfromwork:
@Yoyozilla Is it possible to get an explicit owner for getting this into mrtk_development? Ben has done some great work here and many customers could benefit from this code.

jwittner:
@chrisfromwork I think we're waiting on @davidkline-ms  getting an answer on keeping extensions out of the MRTK repo? Original reasoning for doing so [is here](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3664#issuecomment-475702848).

I'm still faced with these questions:

> Why couple together this technology with a dependency it doesn't even take?
> Going further, why even label this with the MixedReality bucket? Why not Microsoft.WebRTC.Unity?

Maybe my perception is askew, but to me the MRTK should be focused on features fundamental to Mixed Reality, and not broadly valuable features that are useful in Mixed Reality applications. When we do the latter we cut off the rest of the Unity community who might use the features by burdening them with Mixed Reality baggage. WebRTC features would be super valuable for many different types of Unity developers, and I'd be sad to see this capability buried in a place they'll never find.

michael-house:
I agree with @jwittner, I think it would be more appropriate to have WebRTC as a standalone project somewhere. However, we could have an MRTK service "wrapper" as an extension. This would allow people to easily integrate the standalone WebRTC project as a MRTK extension service, with MRTK targeted features like MRC or a WebRTC signalling via some kind of shared session (like HTK sharing).

bengreenier:
Hey folks - I haven't gone completely dark on this work, but have moved to mostly Microsoft internal communications about it to help make sure that whatever we land is the right fit for the community. Apologies for that delay, I'm spread pretty thin these days.


> 
> This fails for me with the following error. I tried both in my current branch which has some local changes and I tried with a clean git checkout using the steps in this doc:

@markgrossnickle sounds like you're working the build issue out over in [this issue](https://github.com/webrtc-uwp/webrtc-uwp-sdk/issues/135) - glad those folks have been able to chime in.

> 
> @bengreenier re: performance, I wanted to double check that your work include the H.264 decoder from Kevin (pretty sure it does). @phongcao reports good performance while both encoding and decoding video on HoloLens so I'm not sure why the plugin has issues.
> 
> One additional issue to list above - if you're using a dual GPU machine on the PC side you need to force PeerCC to use NVIDIA so that there are no issues.

Thanks @jamescadd, it does include the h264 decoder work. I'm not clear at this point what might be leading to these perf issues, but I suspect there may be some additional perf fixes that were expecting to land, but hadn't yet? I wish I had more context, but I don't at this time.

> 
> I'm still faced with these questions:
> 
> > Why couple together this technology with a dependency it doesn't even take?
> > Going further, why even label this with the MixedReality bucket? Why not Microsoft.WebRTC.Unity?

@jwittner, @michael-house I agree, this makes sense. I worry that we may have a hard time finding a set of owners if we go down that path, but probably that's the right thing to go solve. Happy to help, but not sure how to do so at this time.

> 
> @Yoyozilla Is it possible to get an explicit owner for getting this into mrtk_development? Ben has done some great work here and many customers could benefit from this code.

@chrisfromwork thanks! It sounds like @djee-ms may be playing some role in this regard, but if the right thing to do is to rethink the path toward this deliverable, then i'm happy having this work live on my personal repo for now, and end up (in some incarnation) somewhere more concrete in the future. This thread serves as a great way for the community to find it in the meantime. 😄

bengreenier:
Hi folks, as discussed with the MRTK team, we'll close this out for now, and @djee-ms will work to land similar work (but with a more solid and flexible design 🚀) in the near future.

Issue #3359, opened 2019-01-21
FocusManager does not compute distance to Canvas correctly


mgrman:
## Overview
When raycasting UI with EventSystem, Unity returns distance value not from camera center, but from near clip plane. This can be fixed in FocusManager, when converting from camera space to worldSpace add nearClipPlane to the distance. This fixes the issue. See
https://issuetracker.unity3d.com/issues/incorrect-distance-is-returned-when-using-eventsystem-dot-raycastall

## Expected Behavior
Cursor focusing on WorldSpace Canvas is floating above it exactly at Cursor's Surface Distance.

## Actual Behavior
Cursor focusing on WorldSpace Canvas is floating above it, more that Cursor's Surface Distance.

## Steps to reproduce
Create a scene with WorldSpace canvas and a Raycast camera that has non-zero nearClip plane (bigger makes it more obvious). The cursor has an offset as well (set Cursor Surface Distance to be 0). The cursor will be floating above the world space canvas.

## Unity Editor Version
2018.3.1

## Mixed Reality Toolkit Release Version
2017.4.2.0

cellarmation:
I agree with mgrman. This issue also causes the behavior I reported [here](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3322). I had worked out that the nearClipPlane wasn't being taken into account but I couldn't get the values to add up quite right with what I was suspecting. I didn't know about the Cursor Surface Distance, so that probably explains the discrepancy.

I believe the MRTK code did work as expected pre Unity 2017.4.3f1, and has been broken since probably due to the fix [934908](https://issuetracker.unity3d.com/issues/using-canvas-on-world-space-or-camera-buttons-are-not-clickable-if-pos-z-is-set-less-than-zero).

cellarmation:
I suspect mgrman is referring to [Line 642 in FocusManager](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/049b6c2d5edb5656215183c05d93c2adf97a3d32/Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs#L642).
Changing this to:
`newUiRaycastPosition.z = uiRaycastResult.distance + UIRaycastCamera.nearClipPlane;`
[...]

Issue #3322, opened 2019-01-09
Unity UI buttons have priority over MRTK buttons


cellarmation:
## Overview

A Unity UI button being behind a MRTK/HTK button causes user interactions (gaze/highlight and clicks) to be received by the Unity UI button preventing the MRTK/HTK button from being usable. It seems like the Unity UI controls have a higher priority and will steal interaction events unless the MRTK/HTK button is significantly in front. 

This worked as I would expect in Unity 2017.2.0f3 & HoloToolkit-Unity-2017.4.2.0, but it does not in Unity 2017.4.15f1 & HoloToolkit-Unity-2017.4.3.0, so I believe something has changed in Unity or MRTK.

## Expected Behavior
The first button to be hit with a raycast from the user's gaze/controller should have priority and get the event independent of if it is a Unity UI button or a MRTK/HTK button, like it does in Unity 2017.2.0f3 & HoloToolkit-Unity-2017.4.2.0.

## Actual Behavior
Unity UI button's normally get the event unless the MRTK/HTK button is significantly in front of the Unity UI button The attached image shows the camera gazing at the timer button that has been moved in front of the Unity UI button in the example InputManagerTest scene. As you can see the Unity UI button is highlighted (yellow).

![unityui](https://user-images.githubusercontent.com/3090483/50929417-0cbd0280-1455-11e9-86c8-a06df1e2a367.png)

## Steps to reproduce
1) Create a new project
2) Import HoloToolkit-Unity-2017.4.3.0.unitypackage
3) Import HoloToolkit-Unity-Examples-2017.4.3.0.unitypackage
4) Change the scene to HoloToolkit-Examples\Input\Scenes\InputManagerTest.unity
5) Move the "Timer Button" in front of the "Unity UI Button"
6) Run the project and gaze at the timer button

## Unity Editor Version
Unity 2017.4.15f1

## Mixed Reality Toolkit Release Version
HoloToolkit-Unity-2017.4.3.0

cellarmation:
After doing a bit of further testing I have found that the issue isn't present in Unity 2017.2.0f3 with HoloToolkit-Unity-2017.4.3.0, but is in Unity 2017.4.15f1 HoloToolkit-Unity-2017.4.2.0. So, the change in behavior has been caused by Unity doing something different between Unity 2017.2.0f3 and Unity 2017.4.15f1.

cellarmation:
[...]

PR #3666, opened 2019-03-20
UI Ray-cast distances should take into account the near clip plane

cellarmation:
Ray-cast distance calculations should take into account that ray-casts from the UI camera originate from the near clipping plane of the camera since Unity 2017.4.3 and newer versions.

Overview
---

MRTK/HTK does not currently calculate the distance to a UI canvas correctly (#3359) or correctly work out the depth of UI objects when trying to determine if a UI or physics object should get focus (#3322). This is only an issue in Unity 2017.4.3 and later versions due to the introduction of the Unity fix [934908](https://issuetracker.unity3d.com/issues/using-canvas-on-world-space-or-camera-buttons-are-not-clickable-if-pos-z-is-set-less-than-zero)

Changes
---
Currently the original MRTK claims to support Unity 2017.2+ so I have assumed we need to variant the code based on the version of Unity to only fix affected versions.

I added the definition of UNITY_2017_4_3_OR_NEWER so that the existing behavior can be preserved for unaffected Unity versions. Unity only includes _OR_NEWER macros for year.x versions but this change appeared in Unity 2017.4.3 so I have added a definition for UNITY_2017_4_3_OR_NEWER combining existing definitions.

Lines 646 - 650 resolves issue #3359 
Lines 697 - 701 and 709 - 713 are also needed to resolve issue #3322

- Fixes: #3359 & #3322

### DIFF
--- a/Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs
+++ b/Assets/HoloToolkit/Input/Scripts/Focus/FocusManager.cs
@@ -1,6 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+#if UNITY_2017_4_OR_NEWER && !UNITY_2017_4_0 && !UNITY_2017_4_1 && !UNITY_2017_4_2
+#define UNITY_2017_4_3_OR_NEWER
+#endif
+
 using System;
 using System.Collections.Generic;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


cellarmation:
A similar fix is needed for MRTK v2, but I will wait for the outcome of this review before putting together that as a separate pull request.

keveleigh:
Closing and reopening to try to trigger the license check to resolve.

cellarmation:
Raised equivalent fix for MRTKv2, [#3754](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3754).

PR #3667, opened 2019-03-20
Service Registration (#3545) - Part 2 [Breaking Change for Service Implementors]

davidkline-ms:
This part of the service registration (#3545) work includes:

- Removing the need for BaseServiceWithConstructor by updating the BaseExtensionService and BaseDataProvider classes
- Add BaseCoreSystem class and inherit from it for all core system services
- Update core system services to use the data provided (profile and service specific data) to initialize
- Focus provider now inherits from BaseDataProvider
- Gaze provider gets data that previously came from the MixedRealityToolkit.Instance directly from the input system when instantiated

This change removes most of the hard-coded references to MixedRealityToolkit.Instance.

Note: Platform providers and the InputSystemGlobalListener were _not_ updated as part of this change. They will be handled in part 3.


Assets/MixedRealityToolkit.Services/BoundarySystem/MixedRealityBoundarySystem.cs
```diff
@@ -570,7 +590,8 @@ public bool Contains(Vector3 location, Boundary.Type boundaryType = Boundary.Typ
             }
 
             // Handle the user teleporting (boundary moves with them).
-            location = MixedRealityToolkit.Instance.MixedRealityPlayspace.InverseTransformPoint(location);
+            // todo: update...
```
wiwei: Can this TODO be clarified a little (i.e. what to update, how to update, what's the impact of not doing this?)

wiwei: And I think same for the todo down below.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityGazeProvider.cs
```diff
@@ -15,6 +15,16 @@ public interface IMixedRealityGazeProvider
...
+        /// <summary>
+        /// The <see cref="IMixedRealtyInputSystem"/> instance for which this object is providing gaze data.
+        /// </summary>
+        IMixedRealityInputSystem InputSystem { set; }
+
+        /// <summary>
+        /// The transform of the Mixed Reality Playspace
+        /// </summary>
+        Transform Playspace { set; }
+
```
chrisfromwork: why no get?

Assets/MixedRealityToolkit/Services/BaseCoreSystem.cs
```diff
@@ -0,0 +1,30 @@
...
+{ 
+    public abstract class BaseCoreSystem : BaseEventSystem
+    {
+        public BaseCoreSystem(
+            IMixedRealityServiceRegistrar registrar,
+            BaseMixedRealityProfile profile = null) : base()
+        {
+            Registrar = registrar;
+            ConfigurationProfile = profile;
+            Priority = 5; // Core systems have a higher default priority than other services
```
wiwei: >5 [](start = 23, length = 1)

What's the significance of the number 5? i.e. would it be helpful to have a file with priority constants define what the core system priorities should be, and where the starting point of other service priorities should be?

for example:

public static readonly int CoreServicePriority = 5;
public static readonly int NonCoreServicePriorityStart = 100;

So to manage priorities among non-core you would always do YourPriority + NonCoreServicePriorityStart

Assets/MixedRealityToolkit/Interfaces/SpatialAwareness/Observers/IMixedRealitySpatialAwarenessMeshObserver.cs
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.MixedReality.Toolkit.Core.Definitions.SpatialAwarenessSystem;
-using Microsoft.MixedReality.Toolkit.Core.Devices;
+
 using System.Collections.Generic;
 using UnityEngine;
```
chrisfromwork: nit: remove unused line

Assets/MixedRealityToolkit/Providers/BaseDeviceManager.cs
```diff
@@ -14,15 +15,21 @@ namespace Microsoft.MixedReality.Toolkit.Core.Providers
     /// <summary>
     /// Base Device manager to inherit from.
     /// </summary>
+    // todo: inherit BaseDataProvider...
     public class BaseDeviceManager : BaseExtensionService, IMixedRealityDeviceManager
```
chrisfromwork: we need to stop adding arbitrary todos to the repo. If this has another associated issue it needs to be logged here. Otherwise, if this is never done we really generate confusing content in the code.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -709,10 +729,11 @@ public Transform MixedRealityPlayspace
...
         {
-            if (!newInstanceBeingInitialized && !IsInitialized && !UnityEditor.EditorApplication.isPlayingOrWillChangePlaymode)
-            {
-                ConfirmInitialized();
-            }
+            // todo: trying something
+            //if (!newInstanceBeingInitialized && !IsInitialized && !UnityEditor.EditorApplication.isPlayingOrWillChangePlaymode)
+            //{
+            //    ConfirmInitialized();
+            //}
```
wiwei: Should this be uncommented out now?

Assets/MixedRealityToolkit/Providers/UnityInput/MouseDeviceManager.cs
```diff
@@ -21,10 +21,15 @@ public class MouseDeviceManager : BaseDeviceManager, IMixedRealityExtensionServi
         /// <summary>
         /// Constructor.
         /// </summary>
+        /// <param name="registrar">The <see cref="IMixedRealityServiceRegistrar"/> instance that loaded the service.</param>
         /// <param name="name">Friendly name of the service.</param>
         /// <param name="priority">Service priority. Used to determine order of instantiation.</param>
         /// <param name="profile">The service's configuration profile.</param>
-        public MouseDeviceManager(string name, uint priority, BaseMixedRealityProfile profile) : base(name, priority, profile) { }
+        public MouseDeviceManager(
+            IMixedRealityServiceRegistrar registrar,
```
chrisfromwork: >  [](start = 52, length = 1)

nit: a lot of these new constructors have extra spaces at the end of their lines. Please remove.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -396,7 +399,8 @@ private void InitializeServiceLocator()
 #if UNITY_EDITOR                
                 LayerExtensions.SetupLayer(31, "Spatial Awareness");
 #endif
-                if (!RegisterService<IMixedRealitySpatialAwarenessSystem>(ActiveProfile.SpatialAwarenessSystemSystemType) && SpatialAwarenessSystem != null)
+                object[] args = { this }; // todo: add profile
```
chrisfromwork: >// todo: add profile [](start = 42, length = 20)

another todo - add the issue or remove it.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -424,7 +430,21 @@ private void InitializeServiceLocator()
                 for (int i = 0; i < ActiveProfile.RegisteredServiceProvidersProfile.Configurations?.Length; i++)
                 {
                     var configuration = ActiveProfile.RegisteredServiceProvidersProfile.Configurations[i];
-                    RegisterService<IMixedRealityExtensionService>(configuration.ComponentType, configuration.RuntimePlatform, configuration.ComponentName, configuration.Priority, configuration.ConfigurationProfile);
+
+                    if (typeof(IMixedRealityDataProvider).IsAssignableFrom(configuration.ComponentType.Type))
+                    {
+                        // todo: pass the service instance to the data provider
+                        RegisterService<IMixedRealityDataProvider>(configuration.ComponentType, configuration.RuntimePlatform, this, null, configuration.ComponentName, configuration.Priority, configuration.ConfigurationProfile);
```
chrisfromwork: remove or add issue

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -424,7 +430,21 @@ private void InitializeServiceLocator()
...
+
+                    }
+                    else if (typeof(IMixedRealityExtensionService).IsAssignableFrom(configuration.ComponentType.Type))
+                    {
+                        RegisterService<IMixedRealityExtensionService>(configuration.ComponentType, configuration.RuntimePlatform, this, configuration.ComponentName, configuration.Priority, configuration.ConfigurationProfile);
+                    }
+                    else
+                    {
+                        // todo: unsupported type
+                    }
```
chrisfromwork: remove or add issue

Assets/MixedRealityToolkit.Services/BoundarySystem/MixedRealityBoundarySystem.cs
```diff
@@ -570,7 +590,8 @@ public bool Contains(Vector3 location, Boundary.Type boundaryType = Boundary.Typ
...
+            location = Playspace.InverseTransformPoint(location);
```
chrisfromwork: >Playspace [](start = 23, length = 9)

moving to a world where less is needed, i feel like we need to start building under the assumption that someone may want to make this "Playspace" transform null. We should have appropriate null checks throughout or a default Playspace value other than null if one isn't provided.

Assets/MixedRealityToolkit.Services/TeleportSystem/MixedRealityTeleportSystem.cs
```diff
@@ -14,8 +15,16 @@ namespace Microsoft.MixedReality.Toolkit.Services.Teleportation
...
     /// </summary>
-    public class MixedRealityTeleportSystem : BaseEventSystem, IMixedRealityTeleportSystem
+    public class MixedRealityTeleportSystem : BaseCoreSystem, IMixedRealityTeleportSystem
     {
+        public MixedRealityTeleportSystem(
+            IMixedRealityServiceRegistrar registrar,
+            Transform playspace) : base(registrar, null) // Teleport system does not use a profile
+        {
+            Playspace = playspace;
+            IsInputSystemEnabled = (registrar.GetService<IMixedRealityInputSystem>(showLogs: false) != null);
```
wiwei: >(registrar.GetService<IMixedRealityInputSystem>(showLogs: false) != null) [](start = 35, length = 73)

Just making sure I understand this, this means that in order to create a teleport system, there's an ordering required (i.e. I have to both new up an input system, and also register it, before I can actually new this up). i.e. it's not possible to create this first, then enable an input system, and have things light up after.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -125,27 +142,30 @@ public override void Initialize()
                 CameraCache.Main.gameObject.EnsureComponent<StandaloneInputModule>();
             }
 
-            if (MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile == null)
+            if (profile == null)
             {
```
chrisfromwork: this check is already conducted above.

Assets/MixedRealityToolkit/Services/BaseCoreSystem.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.MixedReality.Toolkit.Core.Services
+{ 
+    public abstract class BaseCoreSystem : BaseEventSystem
+    {
+        public BaseCoreSystem(
+            IMixedRealityServiceRegistrar registrar,
+            BaseMixedRealityProfile profile = null) : base()
+        {
+            Registrar = registrar;
+            ConfigurationProfile = profile;
```
chrisfromwork: I think we need to support this being null. I think that in many instances there aren't null checks using this Registrar.

Assets/MixedRealityToolkit/Services/BaseExtensionService.cs
```diff
@@ -13,22 +13,35 @@ namespace Microsoft.MixedReality.Toolkit.Core.Services
...
+        /// <param name="priority">The registration priority of the service.</param>
+        /// <param name="profile">The configuration profile for the service.</param>
+        public BaseExtensionService(
+            IMixedRealityServiceRegistrar registrar, 
+            string name = null, 
+            uint priority = DefaultPriority, 
+            BaseMixedRealityProfile profile = null) : base()
         {
+            Registrar = registrar;
+            Name = name;
```
chrisfromwork: Similar to the system cache of this, there is no guarantee this won't be null, but it is accessed in many locations under the assumption its never null.

Assets/MixedRealityToolkit/Services/BaseCoreSystem.cs
Railboy: @wiwei Agreed, I'm not as allergic to magic numbers as some but these priority numbers wig me out. I could see devs arbitrarily giving their service a priority 99 because they don't know the expected range. Probably an issue for a separate PR since it would be a fairly big change.

Assets/MixedRealityToolkit/Providers/BaseDeviceManager.cs
Railboy: In principle I agree but am less anxious about todos included in part [x] of an incomplete series of pull requests.

Assets/MixedRealityToolkit.Services/BoundarySystem/MixedRealityBoundarySystem.cs
davidkline-ms: this was left over from an earlier round of edits. i'll remove :)

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityGazeProvider.cs
davidkline-ms: i didnt see a need for one. i am happy to add it for balance.

Assets/MixedRealityToolkit/Services/BaseCoreSystem.cs
davidkline-ms: 5 was me leaving existing behavior in place. i am definitely open for adding a defined value

Assets/MixedRealityToolkit/Providers/BaseDeviceManager.cs
chrisfromwork: if this goes into mrtk_development it shouldn't be considered an incomplete series. Every time someone forks mrtk_development they'll have to pick through these.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: most likely. thanks for catching it!

Assets/MixedRealityToolkit/Providers/BaseDeviceManager.cs
chrisfromwork: granted i agree that helpers will eventually be removed. I'd just prefer if you called out this service registration issue in the todos (some formal todo tag thats searchable) so we know what they're for/so a deep search can be done on the last pull request to make sure we've addressed everything.

Assets/MixedRealityToolkit/Services/BaseCoreSystem.cs
davidkline-ms: good call

Assets/MixedRealityToolkit.Services/TeleportSystem/MixedRealityTeleportSystem.cs
davidkline-ms: at this time, the teleport system required an input system. that dependency is not intended to be changed in this PR.

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
```diff
@@ -246,7 +255,7 @@ public override bool TryGetPointerRotation(out Quaternion rotation)
             /// <param name="inputSource"></param>
             public void RaisePointerDown(MixedRealityInputAction mixedRealityInputAction, Handedness handedness = Handedness.None, IMixedRealityInputSource inputSource = null)
             {
-                MixedRealityToolkit.InputSystem.RaisePointerDown(this, mixedRealityInputAction, handedness, inputSource);
+                gazeProvider.InputSystem?.RaisePointerDown(this, mixedRealityInputAction, handedness, inputSource);
```
andreiborodin: Can we just pass in InputSystem to the InternalGazePointer?

davidkline-ms: we may be able to. that would be outside of the scope of this PR. the goal is to limit the change as much as possible.

davidkline-ms:
addressed current set of feedback. ready for re-review. thanks!

PR #3668, opened 2019-03-21
Add scenes list to command line build

MenelvagorMilsom:
Adding scene list parsing for command line build

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
@@ -214,15 +214,13 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Build
                         buildInfo.AutoIncrement = true;
                         break;
                     case "-sceneList":
-                        buildInfo.Scenes = buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
-                                                                  select scene.Trim());
+                        buildInfo.Scenes = buildInfo.Scenes.Union(SplitSceneList(arguments[++i]));
                         break;
                     case "-sceneListFile":
                         string path = arguments[++i];
                         if (File.Exists(path))
                         {
-                            buildInfo.Scenes = buildInfo.Scenes.Union(from scene in File.ReadAllText(path).Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
-                                                                      select scene.TrimStart().TrimEnd());
+                            buildInfo.Scenes = buildInfo.Scenes.Union(SplitSceneList(File.ReadAllText(path)));
                         }
                         else
                         {
@@ -247,6 +245,12 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Build
                         break;
                 }
             }
+        }
+
+        private static IEnumerable<string> SplitSceneList(string sceneList)
+        {
+            return from scene in sceneList.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                   select scene.Trim();
         }
 
         /// <summary>
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
@@ -4,6 +4,7 @@
 using Microsoft.MixedReality.Toolkit.Core.Extensions;
 using Microsoft.MixedReality.Toolkit.Core.Utilities.Editor;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
@@ -212,9 +213,16 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Build
                     case "-autoIncrement":
                         buildInfo.AutoIncrement = true;
                         break;
-                    case "-scenes":
-                        buildInfo.Scenes = from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
-                                           select scene.Trim();
+                    case "-sceneList":
+                        buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                                               select scene.Trim());
+                        break;
+                    case "-sceneListJson":
+                        string path = arguments[++i];
+                        if (File.Exists(path))
+                        {
+                            buildInfo.Scenes.Union(JsonUtility.FromJson<List<string>>(File.ReadAllText(path)));
+                        }
                         break;
                     case "-buildOutput":
                         buildInfo.OutputDirectory = arguments[++i];
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
@@ -213,7 +213,8 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Build
                         buildInfo.AutoIncrement = true;
                         break;
                     case "-scenes":
-                        // TODO parse json scene list and set them.
+                        buildInfo.Scenes = from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                                           select scene.TrimStart().TrimEnd();
                         break;
                     case "-buildOutput":
                         buildInfo.OutputDirectory = arguments[++i];
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
@@ -214,7 +214,7 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Build
                         break;
                     case "-scenes":
                         buildInfo.Scenes = from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
-                                           select scene.TrimStart().TrimEnd();
+                                           select scene.Trim();
                         break;
                     case "-buildOutput":
                         buildInfo.OutputDirectory = arguments[++i];
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
@@ -214,14 +214,19 @@ namespace Microsoft.MixedReality.Toolkit.Core.Utilities.Build
                         buildInfo.AutoIncrement = true;
                         break;
                     case "-sceneList":
-                        buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
-                                               select scene.Trim());
+                        buildInfo.Scenes = buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
```diff
@@ -213,7 +213,8 @@ public static void ParseBuildCommandLine(ref IBuildInfo buildInfo)
                         buildInfo.AutoIncrement = true;
                         break;
                     case "-scenes":
-                        // TODO parse json scene list and set them.
+                        buildInfo.Scenes = from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
```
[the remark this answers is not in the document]
jwittner: I see this note on the [official cmd line docs](https://docs.unity3d.com/Manual/CommandLineArguments.html), which don't adhere to here. That said, it still seems to execute if it's not in a namespace.

> To use -executeMethod, you need to place the enclosing script in an Editor folder.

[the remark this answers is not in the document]
MenelvagorMilsom: @jwittner when you tested this locally, did your repo contain the namespace changes? We had issues with this on CI. We're now using the following on CI and it works fine:
`Microsoft.MixedReality.Toolkit.Build.Editor.UnityPlayerBuildTools.StartCommandLineBuild`

I have tested this locally and it works fine within the namespace.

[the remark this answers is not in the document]
MenelvagorMilsom: I tested this again with absolute and relative paths. Both gave me valid build results and didn't corrupt my VS solution. I also tested with an invalid path to be sure. The build failed gracefully and again, no corruption.

Looking at the stack trace for your exception it looks like this error has more to do with `SyncSolution` than the scene command line args.

@StephenHodgson do you have anything to add for these `SyncSolution` errors?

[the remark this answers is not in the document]
StephenHodgson: This also works for me as well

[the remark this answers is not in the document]
StephenHodgson: The only thing `SyncSolution` does is trigger the editor to generate the .sln and .csproj files for the docfx generation.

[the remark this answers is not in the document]
StephenHodgson: @jwitter I suggest adding the inner stack trace and message to the where the log error is thrown for better information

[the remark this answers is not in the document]
StephenHodgson: At first glance I'd be surprised if this line is parsing the scene information correctly.

I'm not a linq fan but it looks good, but likely there's something missing here to get the correct scene path information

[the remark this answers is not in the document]
jwittner: Ok, I got this work, I was using the wrong, recently updated, namespace.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
```diff
@@ -213,7 +213,8 @@ public static void ParseBuildCommandLine(ref IBuildInfo buildInfo)
...
+                                           select scene.TrimStart().TrimEnd();
```
jwittner: nit: just call `scene.Trim()`.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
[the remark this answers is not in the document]
jwittner: Ok, I'm seeing really weird issues with the solution being corrupted, but its happening even if I disable the SyncSolution in this code. As such I'll try to repro on another PC - could be a VS problem, and resolve this issue here.

MenelvagorMilsom: @StephenHodgson you mentioned concerns about this linq statement. You're not convinced it covers all cases? Perhaps you could elaborate on some of you concerns.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
jwittner: Yay! Thanks for the change. =)

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
StephenHodgson: Where was that comment?

I think my main concern was if the paths defined were relative or absolute.

jwittner: In my testing of this I had to define my paths relative to the project root or Unity would complain at me.

StephenHodgson: My last question. If I had a json file that defined my scene list, could I use that instead?

Seems a bit easier than writing it all in the command line arg.

MenelvagorMilsom: Both absolute paths and paths relative to the project root will work here.

The command line list may be a touch easier on CI as we'd either have to create a file on the fly or check one into the repo. Maybe we could have both? 
-sceneList: comma separated list string provided directly to the command line
-sceneListJson: path to json file that contains a scene list

If both are provided we take the union.

StephenHodgson: I think providing an scene list file is probably the best way to go here, but does this change add that functionality or is it just a csv string?

StephenHodgson: Because I also had plans of creating a helper menu in the build window to create a scene list that would then spit out a json file at a known location (likely the root of the build folder)

MenelvagorMilsom: I do think that we should allow for a csv string as well. I'm happy to have a scene list file option too, but for automated builds, the csv string is one step less. 
Is my proposal for two command line switches above not acceptable?

MenelvagorMilsom:
```
case "-sceneList":
    buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
                           select scene.Trim());
    break;
case "-sceneListJson":
    string path = arguments[i++];
    if (File.Exists(path))
    {
        buildInfo.Scenes.Union(JsonUtility.FromJson<List<string>>(File.ReadAllText(path)));
    }
    break;
```
Something like this?

StephenHodgson: Looks great!

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
```diff
@@ -212,8 +213,16 @@ public static void ParseBuildCommandLine(ref IBuildInfo buildInfo)
...
                         buildInfo.AutoIncrement = true;
                         break;
-                    case "-scenes":
-                        // TODO parse json scene list and set them.
+                    case "-sceneList":
+                        buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                                               select scene.Trim());
+                        break;
+                    case "-sceneListJson":
+                        string path = arguments[i++];
```
StephenHodgson:
```suggestion
                        string path = arguments[++i];
```

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
```diff
@@ -212,8 +213,16 @@ public static void ParseBuildCommandLine(ref IBuildInfo buildInfo)
...
                         break;
-                    case "-scenes":
-                        // TODO parse json scene list and set them.
+                    case "-sceneList":
+                        buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                                               select scene.Trim());
+                        break;
+                    case "-sceneListJson":
+                        string path = arguments[++i];
+                        if (File.Exists(path))
```
jwittner: Should probably error if the file doesn't exist? Or at least warn.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
```diff
@@ -212,8 +213,16 @@ public static void ParseBuildCommandLine(ref IBuildInfo buildInfo)
                     case "-autoIncrement":
                         buildInfo.AutoIncrement = true;
                         break;
-                    case "-scenes":
-                        // TODO parse json scene list and set them.
+                    case "-sceneList":
+                        buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                                               select scene.Trim());
+                        break;
+                    case "-sceneListJson":
```
jwittner: `sceneListJson` makes me think I can provide json as text. Maybe `sceneListFile` or `sceneListJsonFile`?

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
MenelvagorMilsom: So I just tested this on Device, and unfortunately it hadn't built the scenes in the json file. I ended up finding some warnings in the build output. Turns out Unity JsonUtility does support deserializing top level json lists ([link](https://forum.unity.com/threads/how-to-load-an-array-with-jsonutility.375735/#post-2437941)). We'd need to wrap this in an object, and then deserialize to an c# object that contains a string list.

So rather than complicate and confuse things, I'm just going to allow for reading csv from file as well as from command line.

jwittner:
I actually prefer the CSV file instead - almost recommended that yesterday but didn't want to add more work for you.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
```diff
@@ -212,8 +213,21 @@ public static void ParseBuildCommandLine(ref IBuildInfo buildInfo)
...
-                        // TODO parse json scene list and set them.
+                    case "-sceneList":
+                        buildInfo.Scenes = buildInfo.Scenes.Union(from scene in arguments[++i].Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
+                                                                  select scene.Trim());
+                        break;
+                    case "-sceneListFile":
+                        string path = arguments[++i];
+                        if (File.Exists(path))
+                        {
+                            buildInfo.Scenes = buildInfo.Scenes.Union(from scene in File.ReadAllText(path).Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
```
jwittner: Nit: scene.Trim()

jwittner: Nit: Also would be nice to pull out the text parsing to a function so these two long lines aren't mostly duplicated code.

MenelvagorMilsom: Strange, how'd I reintroduce that? I really should do a better pass before pushing.

PR #3670, opened 2019-03-21
Modified Interactable to keep track of pointers using focus changed events

luis-valverde-ms:
Use the `OnBeforeFocusChange` handler to keep track of all the pointers entering and leaving the interactable instead of using `OnFocusEnter/Exit` as those are only triggered for the first pointer entering/last exiting.

- Fixes: #3577 .

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
@@ -35,7 +35,7 @@ namespace Microsoft.MixedReality.Toolkit.SDK.UX.Interactable
 
     [System.Serializable]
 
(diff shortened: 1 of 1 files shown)


Alexees:
@davidkline-ms This pull request does not solve the problem I was solving. I don't know what went wrong, but I feel overlooked here. My pull request merely needed a conclusion for the discussion how enter and exit should be handled. If someone would have told me I would have implemented it in my pull request, but instead, after 15 days of my post being open, this request came along and was approved within 4 hours?
Maybe my pull request was too long and lost focus due to that discussion, but I don't think it's the right approach to bypass a pullrequest that's mostly well thought out, unless nobody cared about actually looking at the code changes, since this change here is quite sparse compared to mine.

Railboy:
@Alexees This was approved quickly because it made a straightforward change that was easy to understand and test. I don't see these changes as a substitute for your work in #3607. Your PR may ultimately merge with or supersede what's been done here.

Issue #3595, opened 2019-03-11
WMR Controller Model and Shader no longer included?


kaedenjames:
## Overview
The newer versions of MRTK-vNext (v2.0.0 Beta 2), don't seem to contain the ability to load the native Controller meshes & shaders. Is there anyway to get these implemented like older versions, or is this something that will be added in the -near- future? The MixedRealityConfigurationGuide states:

> The MRTK does not currently support native SDK's controller models as Unity does not yet provide the capability to load / render gLTF models, which is the default type of models provided by most SDKs. This will be enhanced when this is available.

However, previous versions of the HoloToolKit seemed to have this implemented? I can understand why controllers for Oculus and OpenVR haven't been implemented yet, but why not the WMR ones. Is there anyway to add the older methods into a project running V2.0.0 Beta 2?

## Unity Editor Version
2018.3.7f1
## Mixed Reality Toolkit Release Version
2.0.0 Beta 2

chrisfromwork:
Hello,

We are planning on adding back this functionality. It was originally removed because the underlying gltf loader had a dependency on a newtonsoft json.net library. This dependency often caused version conflicts with different contributors' projects. The first wave of changes is currently in review. It adds a new gltf loader based on Unity's json serializer. A second review should appear shortly to hook back up the wmr controller model loading.

GLTF loader pull request:
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3548

keveleigh:
Was added for WMR in #3673. Please let us know if you see any issues!

PR #3673, opened 2019-03-22
Namespace fixes for - glTF Utilities based off of built in Unity Serialization

chrisfromwork:
Overview
---
This is a new pull request to address some namespace breaks related to https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3548

Changes
---
- Fixes: #3595 .


Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityControllerVisualizationProfile.cs
```diff
@@ -110,5 +110,47 @@ public GameObject GetControllerModelOverride(Type controllerType, Handedness han
...
+        /// </summary>
+        /// <param name="controllerType">The type of controller to query for</param>
+        /// <param name="hand">The specific hand assigned to the controller</param>
+        public SystemType GetControllerVisualizationTypeOverride(Type controllerType, Handedness hand)
+        {
+            for (int i = 0; i < controllerVisualizationSettings.Length; i++)
+            {
+                if (controllerVisualizationSettings[i].ControllerType != null &&
+                    controllerVisualizationSettings[i].ControllerType.Type == controllerType &&
+                    (controllerVisualizationSettings[i].Handedness == hand || controllerVisualizationSettings[i].Handedness == Handedness.Both))
```
wiwei: These conditions are shared between GetControllerVisualizationTypeOverride and GetControllerModelOverride and GetUseDefaultModelsOverride, and probably could be a little helper.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfConversions.cs
```diff
@@ -0,0 +1,440 @@
...
+            bool mirrored = Vector3.Dot(calculatedZ, z) < 0.0f;
+
+            scale.x = x.magnitude * (mirrored ? -1.0f : 1.0f);
+            scale.y = y.magnitude;
+            scale.z = z.magnitude;
+
+            rotation = Quaternion.LookRotation(matrix.GetColumn(2), matrix.GetColumn(1));
+        }
+
+        internal static int[] GetIntArray(this GltfAccessor accessor, bool flipFaces = true)
```
wiwei: >flipFaces [](start = 75, length = 9)

What is flipFaces?

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
```diff
@@ -0,0 +1,384 @@
...
+            {
+                Debug.LogError("Unsupported file name extension.");
+                return null;
+            }
+
+            gltfObject.Uri = uri;
+            int nameStart = uri.Replace("\\", "/").LastIndexOf("/", StringComparison.Ordinal) + 1;
+            int nameLength = uri.Length - nameStart;
+            gltfObject.Name = uri.Substring(nameStart, nameLength).Replace(isGlb ? ".glb" : ".gltf", string.Empty);
+
```
wiwei: Might be possible to use a path API that does this:

[https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfilenamewithoutextension?view=netframework-4.7.2](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfilenamewithoutextension?view=netframework-4.7.2)

It looks like it exists in the .NET versions we care about.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
```diff
@@ -511,8 +525,131 @@ private void UpdateMenuData(InteractionSourceState interactionSourceState, Mixed
             }
         }
 
+        /// <summary>
+        /// Update the Controller Model
+        /// </summary>
+        /// <param name="interactionSourceState"></param>
+        private void UpdateControllerModel(InteractionSourceState interactionSourceState)
+        {
+            if (controllerModelInitialized ||
```
wiwei: >controllerModelInitialized [](start = 16, length = 26)

Once this value is set to true, it looks like UpdateControllerModel will never actually update anyting - is this just maybe misnamed (i.e. it's more like "ensure controller model"). 

Given the current naming I would expect the controller model to update in response to changes, but once this value is true, this entire statement will short circuit and nothing will happen.

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfConversions.cs
chrisfromwork: flip faces changes the triangle definitions/indices order. This changes the outward/normal direction which then results in changing which side of the mesh surface is drawn.

I think that the MRTK standard shader has some different flags that can override/disregard this behavior and draw both sides. But i would also guess this is a direct port from khronos's logic and should be kept

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
chrisfromwork: agreed that this is confusing. will update to EnsureControllerModel

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityControllerVisualizationProfile.cs
chrisfromwork: changed to SettingContainsParameters

Assets/MixedRealityToolkit/Utilities/Gltf/Serialization/GltfUtility.cs
chrisfromwork: seems to work. Removes some flags/seems good

PR #3674, opened 2019-03-22
App State Sharing Extension

Railboy:
Overview
---
A stripped-down, MRTK-service-ified version of the AppState system proposed in #3210

AppState
---
The AppState service provides a means to synchronize a set of structs among connected users. That's it - it doesn't know or care about scene objects, and it's up to the dev to decide what info to synchronize and when.

If you want examples of other systems that can build on AppState to generate / manipulate scene objects, there are several in #3210.

AppStateData / AppStateDataSubscriptions
---
This is a Photon-specific implementation of the IAppStateData interface. IAppStateData is the conduit through which the AppState synchronizes data, and IAppStateDataSubscriptions is used to configure which users receive which data. I would like to remove the Photon dependency and use an MRTK networking service in place of these components.

Profile Setup
---
The profile inspector helps you search for and configure app state structs and (in the case of IL2CPP builds) state array types.

AppState

Other Issues
---
- This PR has a dependency on Newtonsoft. That's a byproduct of the Photon dependency. I would like to eliminate both.
- The connection process handled by PhotonServiceConnector is opaque and flakey. This is another issue that could potentially be solved by using an MRTK networking service.


SimonDarksideJ:
Have you tried to simply replace NewtonSoft within the MRTK side with the UnityJSON utility?  and have you run in to some issues which prevent this working?

Railboy:
I haven't. I'm hesitant to put effort into modifying Photon when we're working towards ditching it altogether.

SimonDarksideJ:
Not necessarily saying alter Photon, just the MRTK specific code, since this branch shouldn't contain any actual Photon code from their asset, the user has to download that themselves.  Just ensuring our interface isn't using Newtonsoft.

Railboy:
AppStateData now uses JsonUtility. I've removed the shared Newtonsoft folder and restored PUN 2's dll. Removing PUN 2 now removes the Newtonsoft  dependency as well.

chrisfromwork:
I think there's a lot of goodness in enabling photon scenarios for the community, but this does raise a few questions:

1) Where was this photon code obtained from? Is there a specific repo/commit that can be referenced for this port? Is it possible to get instructions on how to obtain this code without having to directly check everything into the MRTK? Does anyone know what licenses are associated with these scripts?

2) If we can get instructions on how to obtain this code, it may be more practical to try and understand how your logic can take dependencies on this photon code base without requiring it to be included in the repo. One quick thing may be to stub out specific photon logic behind a macro that a developer has to manually enable.

Assets/MixedRealityToolkit.Extensions/StateControl/AppRoleEnum.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.MixedReality.Toolkit.Extensions.StateControl
+{
+    public enum AppRoleEnum : byte
+    {
+        Client = 1,
+        Server = 2,
+        //Host = Client | Server
+    }
```
chrisfromwork: its unclear what this comment is for

Assets/MixedRealityToolkit.Extensions/StateControl/AppRoleEnum.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Microsoft.MixedReality.Toolkit.Extensions.StateControl
+{
+    public enum AppRoleEnum : byte
+    {
```
chrisfromwork: nit: public enums should have summaries.

Assets/MixedRealityToolkit.Extensions/StateControl/AppState.cs
```diff
@@ -0,0 +1,221 @@
...
+                if (!Initialized)
+                    yield break;
+
+                foreach (Type stateType in appStateProfile.SynchronizedTypes)
+                    yield return stateType;
+            }
+        }
+
+        // Source of our sync lists (implementation will be network back end specific)
+        IAppStateData dataSource;
```
chrisfromwork: docfx won't pick up single // comments but will pick up /// summaries. Its probably worth having any comment you've added use the three slashes vs two in order to reach teh widest audience.

Assets/MixedRealityToolkit.Extensions/StateControl/AppStateProfile.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public Type DataSourceType { get { return dataSourceType.Type; } }
+
+        public SubscriptionModeEnum SubscriptionMode { get { return subscriptionMode; } }
+
+        public string ExperienceName { get { return experienceName; } }
+
+        [SerializeField]
+        private AppRoleEnum appRole = AppRoleEnum.Client;
```
chrisfromwork: nit: serialized fields should have tooltips

Railboy:
>Where was this photon code obtained from? Is there a specific repo/commit that can be referenced for this port?

This is using the PUN 2 package from the Unity asset store. Questions like these are part of why I want to move away from Photon - downloading a fresh version of PUN 2 could break the app state, for instance, and we would have no way of knowing. This situation may improve as things move to the package manager but we're not there yet.

Railboy:
>I think that @andreiborodin and you should sync to see what refactoring could be planned

Agreed, and @andreiborodin I think this use case can help shape your network interface, particularly when it comes to targeting individual connections for sending data.

Railboy:
Closing this and re-opening as draft for clarity.

PR #3677, opened 2019-03-25
fixed docfx errors

thalbern:
fixed malformed tags, typos, missing namespace identifiers, wrong unity crefs / hrefs, obsolete toc entries


Assets/MixedRealityToolkit/Attributes/obj.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
+guid: fc844539ae188ed479aac171497a0077
```
fast-slow-still: Is this meta file intentional? It looks like an artifact from a generated "obj" folder which was then deleted?

Assets/MixedRealityToolkit.Examples/Demos/Audio/Scripts/obj.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
+guid: 1e2e3b9f4400da24da455b576ba53c6c
+folderAsset: yes
+DefaultImporter:
```
fast-slow-still: Is this meta file intentional? It looks like an artifact from a generated "obj" folder which was then deleted?

PR #3680, opened 2019-03-25
Fix documentation changes to work with docfx generation

chrisfromwork:
Overview
---
Fix documentation changes to work with docfx generation:
1) Update crefs for classes outside of immediate namespace
2) Move images declared in readme.md's to appropriate directory and fix references.

Changes
---
- Fixes: # .


MenelvagorMilsom:
@chrisfromwork don't consider the PR validation failures here as blocking. They're caused by an issue that should now be fixed. 
Unfortunately, the fixed version no longer uses the same pipeline, so we can't just rerun the checks. If you want to test against the new PR validation, you could maybe try forcing a change to your branch.

davidkline-ms:
Maybe a close / reopen will trigger the new pipeline checks?

MenelvagorMilsom:
I'd say that would do it. It may be worth making sure we get it checked before we merge this PR, so we can be sure we'll get 100% green docs gen.

chrisfromwork:
@davidkline-ms @MenelvagorMilsom the new pipeline seems to have kicked off. Can both of you review this regardless of the pipeline success?

PR #3691, opened 2019-03-26
Component extensions use == null instead unity's Object.bool

wassx:
Overview
--- After switching to unity 2018.3.9f1 I experienced missing canvas utility scripts on our prefabs. Trying to set them manually didn't work as it seems that EnsureComponent in ComponentExtensions.cs finds already a component but does not check if it is already destroyed.

To be honest, the unity docs about "exists" checks are confusing me.

Changes
---
- Fixes: # .


wassx:
PR has wrong assumptions. Closing

Issue #3689, opened 2019-03-25
Latest MRTK-Development doesn't compile in Unity


dapt1989:
## Overview
Unity doesn't compile with the current implementation on MRTK-Development
## Expected Behavior
Unity should compile and build for both il2cpp and .NET
## Actual Behavior
The commit below does not work, but there is a commit that will compile and build
mrtk_development: 580c7a8a2d29ce417fc0838e4e78c61ccd502bd7 

## Steps to reproduce
Open Unity
Compile project and look at the console

![MRTK-Development](https://user-images.githubusercontent.com/48930515/54958657-b881d480-4f13-11e9-80a3-3b09752b12b1.PNG)

## Unity Editor Version
Unity 2018.3.1f1
## Mixed Reality Toolkit Release Version
mrtk_development commit: 90941f812eb07d828db06a9f2fdd0bf306716182

Alexees:
@dapt1989 what's the build configuration this did not build with? (Platform, backend)

dapt1989:
The error is happening before the build process starts. The unity project as a whole doesn't compile trying to access the MRTK addon, which blocks the project from being able to build. The project is still using a .NET backend and below is the Unity Build Settings window. 

![BuildSettingsForError](https://user-images.githubusercontent.com/48930515/55018399-077d4780-4fb0-11e9-9e68-ef6a42816bae.PNG)

The errors that Unity flags for this bug and #3688 are different so it looks like some issues were solved and new ones appeared thus the reason I created separate bugs, but I apologize if that duplicate complicates things.

davidkline-ms:
@dapt1989, can you try a newer release of Unity 2018.3? I believe there were some API changes in recent patches (.3, if my memory serves me) that prevent using 2018.3.1.

Thanks!

Railboy:
Just verified that this and one other script - InspectorGenericFields.cs - are blocking .NET compilation on in Unity 2018.3.3. Working on a PR now.

dapt1989:
I am able to run with MRTK in editor now, but I am getting errors around glb and gltf importing. 

Scripted importers UnityGLTF.GLTFImporter and Microsoft.MixedReality.Toolkit.Utilities.Gltf.Serialization.Editor.GlbAssetImporter are targeting the glb extension, rejecting both.
UnityEditor.Experimental.AssetImporters.ScriptedImporter:RegisterScriptedImporters()

and

[...]

Issue #3656, opened 2019-03-19
Add .NET Backend validation build to CI


paco-ms:
## Overview
Most developers use IL2CPP on their day to day work. This means .NET backend is easy to break.
Add an additional build configuration to CI that uses the .NET backend (instead of IL2CPP). CI should fail if that configuration fails.

wiwei:
Got this working, will be sending out PR for this soon...

keveleigh:
Pretty sure this is complete, as of #4521

PR #3696, opened 2019-03-26
Minor changes to ensure MRTK compiles with .NET

Railboy:
Overview
---
Minor changes to ensure MRTK compiles with .NET in Unity 2018.3.3+. Logic is identical.

Changes
---
- Fixes: #3689
- Reinforces: #3656


Alexees:
@Railboy just curious, why ref and not out?

Railboy:
I tend to use ref when I can't guarantee the value will make sense.

wiwei:
About the ref vs out distinction - it's probably a good example of a coding guideline we should have (i.e. either could be fine). It's interesting seeing MSDN guidelines saying not to use either:

https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1021-avoid-out-parameters?view=vs-2017

Though this is more for API design rather than internal implementation detail, which this falls under.

I'll take a note of this and see if we can get this codified better.

Issue #3707, opened 2019-03-27
Proposal: Getters should handle null checking serialized fields


Alexees:
## Overview
Unity has the problem of not allowing ?. and ?? to be used on SerializedFields, since they're not overloadable like == and != are.
Additionally, by design, all such fields are accessible by Properties only, or at least that's what I assume from looking through a lot of scripts.
If properties are part of the coding guideline and those properties were to do the null checks themselves and actually return null, one can use the Null coalescing and null conditional operators throughout any other scripts, which would inprove readability.

affected is everything deriving from UnityEngine.Object

Example:

```
if (MixedRealityToolkit.Instance != null &&
    MixedRealityToolkit.Instance.ActiveProfile != null &&
    MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile != null &&
    MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.ControllerMappingProfile != null)
```

vs.

`if (MixedRealityToolkit.Instance?.ActiveProfile?.InputSystemProfile?.ControllerMappingProfile)`

davidkline-ms:
great suggestion! thanks!

StephenHodgson:
Can you confirm that the [Resharper Unity plugin still raises warnings/errors](https://github.com/JetBrains/resharper-unity/wiki/Possible-unintended-bypass-of-lifetime-check-of-underlying-Unity-engine-object) about this even after moving the null checks to the properties?

Alexees:
I don't have Resharper, sorry

StephenHodgson:
I'm not entirely convinced that moving the null check will actually fix the lifetime bypass as it's an overload on the type itself.

Alexees:
it won't fix it if references to these objects are saved and those are checked with null coalescing and conditional operators, but as long as you only ever directly use the properties, it will work.  
throughout the current code both mentioned ways of accessing fields are currently used. since you'd need to distinguish between unity objects and regular ones anyway if you ever want to use those operators, there's not much you can do to prevent the use of the wrong accessing way except for disallowing the use of the two operators altogether.

luis-valverde-ms:
I do like the idea of being able to use ?. and ?? but I feel that we're better off disallowing it completely than selectively supporting it. I think that allowing it in some properties will lead to:

1. Client code using it in properties that don't support it because they have seen it used in properties that did.
2. Property implementers forgetting to add the null check.

[...]

PR #3708, opened 2019-03-27
Mrtk base controller changes

Alexees:
I changed a couple of the already existing properties to forward either their backing instance or null. This makes way for using Null coalescing and conditional operators, in this particular example the BaseController class.

- Starts fixing: #3707  .


Alexees:
@MenelvagorMilsom This is a freshly created pull request and seem to suffer from the same problem the one you contacted me about had.

MenelvagorMilsom:
@Alexees that's strange. Again this current problem is unrelated to your PR. For some reason Unity is still holding a lock from previous builds. I'll investigate and get back to you.

MenelvagorMilsom:
Oh wait, looks like things are fixed in [build 345](https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=345). Apologies, but the CI system is new and quite brittle.

MenelvagorMilsom:
So it looks like the newest failure may actually be related to your changes. Can you try running build.ps1 found in the MRTK root and verify that it doesn't fail.

StephenHodgson:
Can you confirm that the Resharper Unity plugin still raises warnings/errors about this even after moving the null checks to the properties?

Alexees:
From the looks of #3707 it's more like not making use of the two operators at all.

PR #3711, opened 2019-03-27
added new doc structure

thalbern:
-added new doc structure
-removed obsolte toc entries and files / simplified
-restructured and documented docfx.json build file 
-added placeholder for coming soon content


api_doc/toc.yml
```diff
@@ -1,13 +0,0 @@
-- name: Conceptual
-  items: 
-  - name: Spatial Awareness
```
fast-slow-still: Do I understand this right, that this toc is going away and instead of pointing to this, toc.yml/toc.yml points directly at obj/api/toc.yml?

thalbern: yep - yoyo restructured the docs and in the new layout we won't have conceptual and generated api docs in the same top level category which makes this file obsolete

fast-slow-still: Great, thanks!

---
In reply to: [269657290](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3711#discussion_r269657290) [](ancestors = 269657290)

fast-slow-still: I'm not sure why your reply showed up as a new unrelated comment. \_(~)_/

---
In reply to: [269657754](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3711#discussion_r269657754) [](ancestors = 269657754,269657290)

thalbern: ¯\_(ツ)_/¯

PR #3714, opened 2019-03-27
Event system fix where the event listeners array was being modified while an event was getting dispatched

keveleigh:
Fix a bug in input system where the event listeners array was being modified while an event was getting dispatched, because of nested dispatch.

### DIFF
--- a/Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
+++ b/Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
@@ -15,8 +15,8 @@ namespace Microsoft.MixedReality.Toolkit
     {
         #region IMixedRealityEventSystem Implementation
 
-        private static bool isExecutingEvents = false;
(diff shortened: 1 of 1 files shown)


StephenHodgson:
Could you explain this one a bit more? I don't understand how there was a nested dispatch?

StephenHodgson:
The change looks good though. I just want to better understand the problem.

StephenHodgson:
Also, this wasn't an input system specific bug since it's the base event system that was modified, just to be clear the Teleport, Boundary, and Input systems are all affected by this.

julenka:
Hi Stephen, here is the example that caused a bug:

A Controller would raise OnInputDown when select button was pressed.
  in the OnInputDown handler for the BaseControllerPointer, the BaseControllerPointer would raise OnPointerDown by calling MixedRealityToolkit.InputSystem.RaisePointerDown

In the first level, something unregistered itself, which caused the list to be modified.

StephenHodgson:
Thanks Julia :)

PR #3724, opened 2019-03-28
Tooltips Fixes / Text Mesh Pro Assets

Railboy:
Overview
---
Updates Tooltip prefabs to use more performant settings by default and replaces their TextMesh components with TextMeshPro components.

Also brings in some related Text Mesh Pro assets and fixes:
- Single Pass Instanced compatible TMP shader from @Cameron-Micka 
- Standard Text Mesh Pro font assets from @cre8ivepark 

Other improvements / fixes:
- Tooltips now validate their hierarchy and auto-generate any missing transforms
- Removed circular dependency that caused ToolTipConnector to add a second ToolTip component
- Added handles for manipulating pivot and anchor
- Added Bezier ToolTip prefab

ToolTips


mr0ng:
Amazing, thank you!

PR #3725, opened 2019-03-28
Pointer focus dispatch rework

keveleigh:
Overview
---
From @julenka:

This PR fixes a bug in the input system where pointers are dispatching pointer events to all other pointers attached to an input source, which causes a large number of null pointer exceptions and also duplicate events to be sent to the system.

## Explanation of the Bug
Currently MixedRealityInputSystem dispatches any event `E` to all pointers attached to `E`.InputSource. Therefore, when a Pointer wants to send a PointerDown event to the object it is focusing, instead the following happens when trying to dispatch pointer event `E` from pointer `P`:

1. Initialize `E` with information about pointer `P` so that `eventData.Pointer` returns `P`
1. Get the InputSource `S` attached to pointer `P`
2. For all pointers `P2` on `S`...
  a. If `P2` is focused on some object `O`
  b. Dispatch pointer event `E` to `O`, **even if P2 ~= P.**

This causes two problems when you have multiple pointers on an input source, for example the multiple pointers we have on fully articulated hands:

1. Each PointerDown event will get sent N times, where N = # pointers attached to a source.
2. If one pointer is focused on something, but another is not, then in step 1, `E.Pointer` is set to something that doesn't actually have object focus (`E.Pointer.Result.Object == null`). Then, if something tried to access `E.Pointer.Result.Object` it will get null pointer. This violates an assumption that for any PointerEvent `E2` being dispatched, `E2` is focusing some object.

## Explanation of the Fix
When dispatching a pointer event, this event should only be sent to global event listeners and the object being focused by the pointer, NOT all other pointers attached pointer.InputSource.

Create a new method `HandlePointerEvent` that should be used to dispatch all events for pointers (which are the only things that can provide object focus). This method only dispatched events to global handlers and the object being focused, not to other pointers on the input source.

Also refactor the code a bit to avoid duplicate code.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
@@ -244,6 +244,47 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// <inheritdoc />
         public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.EventFunction<T> eventHandler)
         {
+            Debug.Assert(eventData != null);
+            Debug.Assert(!(eventData is  MixedRealityPointerEventData), "HandleEvent called with a pointer event. All events raised by pointer should call HandlePointerEvent");
+
+            if (disabledRefCount > 0)
+            {
+                return;
+            }
+
+            var baseInputEventData = ExecuteEvents.ValidateEventData<BaseInputEventData>(eventData);
+            DispatchEventToGlobalListeners(baseInputEventData, eventHandler);
+            
+            if (baseInputEventData.used)
+            {
+                // All global listeners get a chance to see the event,
+                // but if any of them marked it used,
+                // we stop the event from going any further.
+                return;
+            }
+
+            Debug.Assert(baseInputEventData.InputSource.Pointers != null, $"InputSource {baseInputEventData.InputSource.SourceName} doesn't have any registered pointers! Input Sources without pointers should use the GazeProvider's pointer as a default fallback.");
+
+            var modalEventHandled = false;
+            // Get the focused object for each pointer of the event source
+            for (int i = 0; i < baseInputEventData.InputSource.Pointers.Length && !baseInputEventData.used; i++)
+            {
+                modalEventHandled = DispatchEventToObjectFocusedByPointer(baseInputEventData.InputSource.Pointers[i], baseInputEventData, modalEventHandled, eventHandler);
+            }
+
+            if (!baseInputEventData.used)
+            {
+                DispatchEventToFallbackHandlers(baseInputEventData, eventHandler);
+            }
+        }
+        
+        /// <summary>
+        /// Handles a pointer event
+        /// Assumption: We only send pointer events to the objects that pointers are focusing, except for global event listeners (which listen to everything)
+        /// In contract, all other events get sent to all other pointers attached to a given input source
+        /// </summary>
+        private void HandlePointerEvent<T>(BaseEventData eventData, ExecuteEvents.EventFunction<T> eventHandler) where T : IMixedRealityPointerHandler
+        {
             if (disabledRefCount > 0)
             {
                 return;
@@ -251,17 +292,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
             Debug.Assert(eventData != null);
             var baseInputEventData = ExecuteEvents.ValidateEventData<BaseInputEventData>(eventData);
-            Debug.Assert(baseInputEventData != null);
-            Debug.Assert(!baseInputEventData.used);
-
-            if (baseInputEventData.InputSource == null)
-            {
-                Debug.LogError($"Failed to find an input source for {baseInputEventData}");
-                return;
-            }
-
-            // Send the event to global listeners
-            base.HandleEvent(eventData, eventHandler);
+            DispatchEventToGlobalListeners(baseInputEventData, eventHandler);
 
             if (baseInputEventData.used)
             {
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -271,70 +302,86 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
...
+                DispatchEventToFallbackHandlers(baseInputEventData, eventHandler);
             }
+        }
 
-            var modalEventHandled = false;
+        /// <summary>
+        /// Dispatch an input event to all global event listeners
+        /// Return true if the event has been handled by a global event listener
+        /// </summary>
+        private void DispatchEventToGlobalListeners<T>(BaseInputEventData baseInputEventData, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
```
StephenHodgson: Could we update the language to match Unity's?

`HandleGlobalListenerEvent<T>` would be a better name.

(same for all the other new methods)

StephenHodgson:
Isn't this a direct effect of making the GazeProvider re-throw the input event as a pointer event? (Which I heavily disagreed to doing) Shouldn't we address alternative solutions that explore ways to raise the pointer event without having to route the event through the gaze provider?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -244,24 +244,55 @@ public override void Disable()
         /// <inheritdoc />
         public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.EventFunction<T> eventHandler)
         {
+            Debug.Assert(eventData != null);
```
StephenHodgson: Why move these?

keveleigh:
>Isn't this a direct effect of making the GazeProvider re-throw the input event as a pointer event?

Not at all. The GazeProvider only sends a pointer event if the [input source actually has it registered as a pointer](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs#L379), which only happens in the case of [no other pointers being registered to the input source](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Providers/BaseGenericInputSource.cs#L25).

This is result of sources needing to have multiple pointers for different interactions. For example, a controller that has both a far line pointer and a near grab pointer (somewhat like the HTK's Grabbables example). The near pointer's event doesn't need to be routed to the object being focused by the far pointer (which is also fired by the [base pointer rethrowing the input event](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs#L447) as a pointer event).

StephenHodgson:
I really didn't understand the response. You confirmed that it wasn't a direct effect of being re-thrown, but then showed an example that was directly affected by re-throwing the input down event in the base controller.

keveleigh:
I confirmed that events are being re-thrown, yes, but that's not the issue. Regardless of where pointer events are sent from, if they're sent via the existing `HandleEvent` path, we'll run into issues when multiple pointers are involved on the same input source.

This change makes it so that pointer events are handled more like [focus events are currently handled](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs#L748). The only object that needs to know about it is the object the pointer is actually focusing. In the pointer event case, we also dispatch to the global listeners, which we don't do with focus.

StephenHodgson:
Okay that makes more sense to me.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -244,24 +244,55 @@ public override void Disable()
...
 
-            // Send the event to global listeners
-            base.HandleEvent(eventData, eventHandler);
+            Debug.Assert(baseInputEventData.InputSource.Pointers != null, $"InputSource {baseInputEventData.InputSource.SourceName} doesn't have any registered pointers! Input Sources without pointers should use the GazeProvider's pointer as a default fallback.");
+
+            var modalEventHandled = false;
+            // Get the focused object for each pointer of the event source
+            for (int i = 0; i < baseInputEventData.InputSource.Pointers.Length && !baseInputEventData.used; i++)
+            {
+                modalEventHandled = DispatchEventToObjectFocusedByPointer(baseInputEventData.InputSource.Pointers[i], baseInputEventData, modalEventHandled, eventHandler);
```
StephenHodgson: is this used anywhere after being assigned?

keveleigh: Yes, on that line it's passed into the `DispatchEventToObjectFocusedByPointer` method.

StephenHodgson: Oh I see, it's a loop. Thanks :)

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
keveleigh: Moved them back below the `if`

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -271,70 +302,86 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
...
+        private bool DispatchEventToObjectFocusedByPointer<T>(IMixedRealityPointer mixedRealityPointer, BaseInputEventData baseInputEventData,
+            bool modalEventHandled, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
+        {
+            GameObject focusedObject = FocusProvider?.GetFocusedObject(mixedRealityPointer);
 
-                    if (modalInput != null)
-                    {
-                        modalEventHandled = true;
+            // Handle modal input if one exists
+            if (modalInputStack.Count > 0 && !modalEventHandled)
```
StephenHodgson: Since we created methods for all the other dispatch types, doesn't it make sense to also have a `DispatchEventToModalHandlers` too?

keveleigh: Since focused objects and the modal stack [work together to ensure the right objects get the event](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3725/files#diff-4cee86eaf4620444450371e114592641R357), it makes sense to me to keep them intertwined.

StephenHodgson:
So we're still merging things in with changes requested?

PR #3730, opened 2019-03-29
Fixing Un-Linked events in AppBar Buttons

cre8ivepark:
Overview
---
Fixing Un-Linked events in AppBar Buttons

Changes
---
Button events got un-linked after namespace change.


Railboy:
Very strange that this build would fail given that we're only modifying a prefab.

wiwei:
The build is failing because the build machines don't have the required SDKs to actually build the new articulated hands support - has nothing to do with this change.

I'm working with the folks who own the build resourcing to ensure that we get those SDKs added.

chrisfromwork:
@MenelvagorMilsom do these build breaks seem related to this change? It kind of looks like a path is missing for the build compared to a code related break.

PR #3731, opened 2019-03-29
Adding new README files

cre8ivepark:
Overview
---
Adding new README files

Changes
---
New README files added to /Documentation


Documentation/README_Interactable.md
```diff
@@ -0,0 +1,285 @@
...
+
+![Interactable](/External/ReadMeImages/Interactable/InteractableInspector_basicSteps.png)
+
+1. A Collider must exist on the GameObject with the Interactable or the child of the Interactable for it to receive input.
+2. Use the OnClick event to make something happen.
+3. Add visual feedback by linking a Target to a Profile and assigning a Theme.
+
+
+## Input Settings
+The basic features allow for button style interactions, such as pointer focus and clicks, that maps to interaction states to drive themes which are setup through the Interactable profile. Controller or hand focus, down, up and click (both near and far) are handled. Functionality can be extended using external scripts that can set state manually.
```
wiwei: Do we have any examples to point to for how to do this sentence? It seems really cool that it suggests a capability exists, and I think it would be helpful to have an example as well.

If this is going to be part of a next drop of docs that's okay (or if this actually refers to the stuff later in this file having that explicit would be nice)

Documentation/README_Interactable.md
```diff
@@ -0,0 +1,285 @@
...
+
+<img src="/External/ReadMeImages/Interactable/InputFeatures_short.png" width="450">
+
+### Input Actions
+Select the action, from the input configuration or controller mapping profile, that the Interactable should react to.
+See the InputProfile or the DefaultMixedRealityInputActionsProfile for more on how Input Actions are setup and intended to be used in the application.
+ 
+### Enabled
+Sets the Interactable's enabled state, which will disable some input handling and update the themes to reflect the current state which is disabled.
```
wiwei: I think it actually may be helpful here to have the example first (i.e. visible submit button that looks greyed out) - the first part of this paragraph was slightly confusing to me to read until I got the specific example (i.e. there is an overloading of the term "enabled/disabled" which comes from maybe flat UI interaction land vs. the concept of enabled/disabled in Unity land)

I otherwise would have changed this myself but I wanted to see your thoughts on this.

Documentation/README_Interactable.md
```diff
@@ -0,0 +1,285 @@
...
+This is different from disabling input all together (using Enable Input) This means we have a button that would normally be interactive, but at this moment it is disabled (could be like a submit button waiting for all the required fields to be completed) and has a visual look and feel to denote it's disabled state.
+ 
+### IsGlobal
+Focus is not required to detect input actions, default behavior is false.
```
wiwei: Is there an example of what IsGlobal could accomplish (i.e. this allows for voice to work globally?)

Documentation/README_Interactable.md
```diff
@@ -0,0 +1,285 @@
...
+ 
+### Voice Commands
+A voice command to trigger an OnClick event. This will also trigger a quick state change to drive any themes visuals.
+ 
+Note: Make sure there is a unique voice command on each button. The Voice Recognizer is global (even if the Interactable isn't) and will not register the same voice command twice; in this case an error will be thrown.
```
wiwei: Do you know if this is a runtime error, or in-editor error? i.e. will I only find out about this problem when things crash on the device, or will I get a chance to know in the editor.

Documentation/README_ObjectCollection.md
```diff
@@ -0,0 +1,33 @@
+# Object Collection
```
wiwei: Neat.

Documentation/README_SpatialAwareness.md
```diff
@@ -0,0 +1,6 @@
+# Spatial Mapping
+![Spatial Mapping](/External/ReadMeImages/SpatialAwareness/MRTK_SpatialAwareness_Main.png)
+
+## How to use Spatial Mapping
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Quis ipsum suspendisse ultrices gravida. Risus commodo viverra maecenas accumsan lacus vel facilisis.
```
wiwei: I think we should probably delete this file, or add a 'coming' soon instead of the filler text. (Probably delete? Coming soon isn't usually very helpful)

README.md
```diff
@@ -4,19 +4,15 @@
 
 MRTK is a Microsoft Driven opensource project. 
 
-MRTK-Unity provides a set of foundational components and features to accelerate MR app development in Unity. Latest Release of MRTK (V2) supports HoloLens/HoloLens 2, WMR, and OpenVR platform. Start your project with MRTK, and 
+MRTK-Unity provides a set of foundational components and features to accelerate MR app development in Unity. Latest Release of MRTK (V2) supports HoloLens/HoloLens 2, WMR, and OpenVR platform.
```
wiwei: I think it would be good to revert these changes and take the changes that Julia just made.

wiwei: And by these I just mean the first two changed sections (feature areas / overview)

wiwei:
I also ended up making some commits to fix typos and random issues (like github markdown not supporting footnotes out of the box). Please double check my changes too :)

Documentation/README_Interactable.md
cre8ivepark: We do have an exmaple scene ‘InteractableExamples’

Documentation/README_Interactable.md
cre8ivepark: @killerantz Could you please take a look?

Documentation/README_SpatialAwareness.md
cre8ivepark: Let’s delete for now. Or if someone can add simple description about the example scene, it would be great.

Documentation/README_Interactable.md
killerantz: Is it worth adding a diagram?

Documentation/README_Interactable.md
killerantz: I can see the confusion, maybe we do have an example of a button in four different states to show the visual differences and try to bring the reference back to UI states instead of Unity functionality.

Documentation/README_Interactable.md
killerantz: IsGlobal only refers to input like down, up and click. Does the object require focus for it to handle those events. Voice Commands being "global" is handled with the Requires Focus value under Voice Commands. Requires Focus only shows up when a valid string is entered into the Voice Commands field. (the two things are similar, but handle different use cases)

Documentation/README_Interactable.md
killerantz: I have seen this error in the Editor.

Documentation/README_Interactable.md
killerantz: I'll push up a couple more images tomorrow.

PR #3732, opened 2019-03-29
Spectator View Spatial Localization Abstraction

andreiborodin:
Added Spatial Localization abstraction, and created two classes against it showcasing the abstraction and that will be used as part of later refactor of SpecatorView.


Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerVisualizeCoordinateService.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
+{
+    /// <summary>
+    /// A variant of marker based <see cref="ISpatialCoordinateService"/> implementation. This one tracks coordinates displayed on the screen of current mobile device.
+    /// The logic is that every time you start tracking a new coordinate is created and shown on the screen, after you stop tracking that coordinates location is no longer updated with the device.
+    /// </summary>
+    public class MarkerVisualizeCoordinateService : SpatialCoordinateServiceBase<int>
```
chrisfromwork: >MarkerVisualizeCoordinateService [](start = 17, length = 32)

nit: may be worth naming MarkerVisualizer* compared to MarkerVisualize*

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerVisualizeCoordinateService.cs
```diff
@@ -0,0 +1,100 @@
...
+            public override Vector3 WorldToCoordinateSpace(Vector3 vector) => worldToCoordinate.MultiplyPoint(vector.AsUnityVector()).AsNumericsVector();
+
+            public override Quaternion WorldToCoordinateSpace(Quaternion quaternion) => (worldToCoordinateRotation * quaternion.AsUnityQuaternion()).AsNumericsQuaternion();
+        }
+
+        private readonly IMarkerVisual markerVisual;
+        private readonly UnityEngine.Transform markerInWorldSpace;
+        private readonly Func<int> generateMarkerId;
+
+        private SpatialCoordinate markerBeingVisualized;
```
chrisfromwork: >markerBeingVisualized [](start = 34, length = 21)

nit: maybe just markerCoordinate?

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/ISpatialCoordinate.cs
```diff
@@ -0,0 +1,43 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Numerics;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.Sharing
+{
+    /// <summary>
+    /// This represents a spatial coordinate that can then be used to convert position and rotation to and from this coordinate space.
+    /// </summary>
+    public interface ISpatialCoordinate
```
chrisfromwork: >ISpatialCoordinate [](start = 21, length = 18)

General question based on this marker scenario: Do we want a ISpatialCoordinate to remain static in location? Is that a fundamental truth we want to build against? This marker location seems like a first instance where technically the spatial coordinate moves dynamically with device tracking. If we decide that all spatial coordinates should be considered static, we may need to dynamically create spatial coordinates for markers as devices move around a scene. If we decide that a spatial coordinate isn't static, we may want to add some functionality that allows coordinates to be static/non static. We may also want some functionality to be able to update the coordinate location by its creator?

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/SpatialCoordinateServiceBase.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Numerics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.Sharing
+{
+    /// <summary>
+    /// Helper base class for <see cref="ISpatialCoordinateService"/> implementations.
```
chrisfromwork: >ISpatialCoordinateService [](start = 41, length = 25)

nit: you may already be doing this, but with these new crefs we should run docfx generation locally and fix any warnings. I think this resolves correctly, but its worth confirming.

When we move to pr's against mrtk_development, docfx generation should be a build step. But right now there is no build infrastructure for our feature/spectatorView.

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/SpatialLocalizer.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.Sharing
+{
+    /// <summary>
+    /// Very simple consumer of <see cref="ISpatialCoordinate"/> to demonstrate usage.
+    /// </summary>
+    public class SpatialLocalizer : MonoBehaviour
+    {
+        [SerializeField]
+        private GameObject targetRoot = null;
```
chrisfromwork: nit: Let's go ahead and add tooltips to serialized fields on their creation to avoid large painful documentation pushes in the future. :)

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/SpatialLocalizer.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        [SerializeField]
+        private bool autoToggleActive = false;
+
+        public Vector3 CoordinateRelativePosition;
+        public Quaternion CoordinateRelativeRotation;
```
chrisfromwork: nit: add summary

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/SpatialLocalizer.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public ISpatialCoordinate Coordinate { get; set; }
+
```
chrisfromwork: nit: add summary

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/SpatialLocalizer.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <summary>
+    /// Very simple consumer of <see cref="ISpatialCoordinate"/> to demonstrate usage.
+    /// </summary>
+    public class SpatialLocalizer : MonoBehaviour
+    {
+        [SerializeField]
+        private GameObject targetRoot = null;
+
+        [SerializeField]
+        private bool autoToggleActive = false;
```
chrisfromwork: nit: tooltip

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/SpatialLocalizer.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.Sharing
+{
+    /// <summary>
+    /// Very simple consumer of <see cref="ISpatialCoordinate"/> to demonstrate usage.
+    /// </summary>
+    public class SpatialLocalizer : MonoBehaviour
```
chrisfromwork: >SpatialLocalizer [](start = 17, length = 16)

nit: may be more appropriate to call this out as a coordinate localizer. There may one day be more elaborate helper classes that uses multiple coordinates for positioning items.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerVisualizeCoordinateService.cs
```diff
@@ -0,0 +1,100 @@
...
+            // TODO we can probably get rid of UpdateTick and use a Unity synchronization context here to wait frames
+            await Task.Delay(-1, cancellationToken).IgnoreCancellation();
+
+            markerVisual.HideMarker();
+            markerBeingVisualized.IsLocated = false;
+            markerBeingVisualized = null;
+        }
+
+        public void UpdateTick()
+        {
```
chrisfromwork: nit: add summary or inheritdoc

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/Extensions.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+            catch (Exception ex)
+            {
+                UnityEngine.Debug.LogError("Encountered an exception with a FireAndForget task.");
+                UnityEngine.Debug.LogException(ex);
+            }
+        }
+
+        /// <summary>
+        /// Prevents <see cref="TaskCanceledException"/> or <see cref="OperationCanceledException"/> from trickling up.
```
chrisfromwork: >OperationCanceledException [](start = 71, length = 26)

nit: I'm personally curious whether these exceptions resolve correctly with docfx.

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/Extensions.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// <summary>
+    /// Helper Unity side extensions to enable the core abstraction to be Unity agnostic.
+    /// </summary>
+    public static class Extensions
+    {
+        /// <summary>
+        /// Converts a <see cref="Vector3"/> to a <see cref="UVector3"/>.
+        /// </summary>
+        public static UVector3 AsUnityVector(this Vector3 input) => new UVector3(input.X, input.Y, input.Z);
+
```
chrisfromwork: How hard do we feel a migration to System.Numerics types will be in the future if we only used Unity types for the time being?

Non-Unity middleware may be unable to play well with .Net types, which I think is the main intended worth in this change. It looks like this namespace exists in .NET 4.0+ frameworks. I don't know if anyone uses .NET frameworks older than 4.0, but I don't think this would resolve correctly in this current form for those that do. I think I have seen some issues related to this based on folks using the HoloToolkit with older Unity/.NET versions. Does the consumer confusion that could be generated with this change outweigh the worth in standardizing on a non-Unity type?

From what I understand the intention here is to build a more universal experience by using .Net types. Although these extension methods are helpful, I think that most Unity developers will still prefer Unity types. I think for the time being I would argue that we should continue to use Unity types. Once we have a stronger perspective on how to build components compatible with Unreal and Unity, it may be a better time to try and introduce these sorts of type conversions.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerVisualizeCoordinateService.cs
andreiborodin: Yep, will update.

---
In reply to: [270508873](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3732#discussion_r270508873) [](ancestors = 270508873)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerVisualizeCoordinateService.cs
andreiborodin: Yeah, will update as well.

---
In reply to: [270509711](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3732#discussion_r270509711) [](ancestors = 270509711)

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/ISpatialCoordinate.cs
andreiborodin: Yeah, I was thinking of having a IsTracked {get;set;} value for each ISpatialCoordinate as well. I feel because we are experimental, we can do this when the need arises.

---
In reply to: [270511687](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3732#discussion_r270511687) [](ancestors = 270511687)

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/SpatialCoordinateServiceBase.cs
andreiborodin: I use VS's intellisense filling these in, so should resolve correctly.

---
In reply to: [270512318](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3732#discussion_r270512318) [](ancestors = 270512318)

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/SpatialLocalizer.cs
andreiborodin: Good call

---
In reply to: [270513793](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3732#discussion_r270513793) [](ancestors = 270513793)

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/Extensions.cs
andreiborodin: I am choosing to go this route for several reasons:
- Feel the pain of going this route, so that when we do make a full choice we understand. (I will be documenting this)
- It is simple to flip back to only Unity types, and we are in the Experimental branch anyways.
- Experiment with having this code used in Unreal.

---
In reply to: [270518326](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3732#discussion_r270518326) [](ancestors = 270518326)

Assets/MixedRealityToolkit.Extensions/Sharing/Localization/Extensions.cs
andreiborodin: They should be able to, as VS intellisense resolved them.

---
In reply to: [270515114](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3732#discussion_r270515114) [](ancestors = 270515114)

chrisfromwork: I know that Bernadette stated that there may be issues with resolving namespaces with docfx. I know i have hit issues where editor items couldn't resolve classes across asmdefs correctly. It's not a big deal for now but may break the build when opening a pr against mrtk_development

PR #3737, opened 2019-03-29
Bugfix: Pan and scroll interactions in the Hands Demo are not working with hand rays 

johnppella:
Overview
---
Bugfix: Pan and scroll interactions are not working with hand rays.


Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/HandInteractionPan.cs
```diff
@@ -489,51 +491,37 @@ public void OnTouchCompleted(HandTrackingInputEventData eventData)
...
+         if (touchType == TouchType.HandRay)
         {
             if (touchingSource == null)
             {
                 currentPointer = eventData.InputSource.Pointers[0];
                 currentController = currentPointer.Controller;
                 touchingSource = eventData.InputSource;
-                // touchingPoint = currentPointer.Rays[touchingSource.Pointers[0].Result.RayStepIndex].Terminus;
+
                 TryGetHandRayPoint(currentController, out touchingPoint);
```
julenka: >TryGetHandRayPoint(currentController, out touchingPoint); [](start = 16, length = 57)

Is TryGetHandRayPoint needed any more? It looks like this code is doing something like point = event.pointer.controller.pointer. Seems like you could just do `touchingPoint = eventData.Pointer.Position`.

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/HandInteractionPan.cs
```diff
@@ -469,16 +468,19 @@ public void OnTouchStarted(HandTrackingInputEventData eventData)
             if (touchingSource == null)
             {
                 currentPointer = eventData.InputSource.Pointers[0];
-                currentController = eventData.Controller;
+                currentController = currentPointer.Controller;
```
julenka: Is there still a need to track currentController?

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/HandInteractionPan.cs
johnppella: yes- change made

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/HandInteractionPan.cs
johnppella: agree about redundancy. I removed currentPointer though- because controller seems more useful

SimonDarksideJ:
Would this change better be renamed to 
> Bugfix: Pan and scroll interactions in the Hands Demo are not working with hand rays

As this change does not touch the framework or SDK, only updates the demo.

P.S. 
I hope this isn't actually a Core issue with hand interaction being "Fixed" in a demo.  As was done many times in the HTK.  Please confirm this is only a DEMO issue.

keveleigh:
>As was done many times in the HTK.

For context, do you have specific examples?

SimonDarksideJ:
> For context, do you have specific examples?

Hopefully that question was aimed at the audience, but the biggest example of "Implementation" code being bundles inside a demo/example would be the entire Button / Interaction experience in the HTK.
Many have tried to then copy / implement those features in their projects for the HTK, failed and then called the toolkit not nice things.

That aside, this is a discussion about a fix going in to the MRTK, not a discussion about the past grievances of the HTK.

Is this code a fix for the demo, or has implementation code been implemented in the example/demo?

keveleigh:
Thanks for the additional context.

@cre8ivepark and @johnppella, we should do a pass on the scripts in the Examples folders and pull out any that make sense as an SDK component. This script does seem like a good candidate. Not blocking this PR, but we should do it before RC1.

davidkline-ms:
retargeting to the dev branch reset previous approvals. merging

PR #3738, opened 2019-03-29
Service registration (#3545) part 3 - systems load data providers

davidkline-ms:
This PR is a continuation of the work on #3545.

Changes include:
- Add IMixedRealityServiceConfiguration
- Add MixedRealityInputDataProviderConfiguration and MixedRealitySpatialObserverConfiguration
- Rename BaseDeviceManager to BaseInputDeviceManager and update constructor args
- Update the input system profile to contain the list of input device managers
- Update the input system implementation to initiate load/unload of data providers
- Add a spatial awareness profile to contain the collection of observers
- Update the spatial awareness system to initiate load/unload of observers
- Fix bugs in Enable for Speech (not checking for existing recognizer) and Dictation (possible re-registering of event handlers)
- Add hand joint service and input simulator to input data providers
- IMixedRealityDataProvider no longer inherits from IMixedRealityExtension service as they require a service (and are now registered by their parent service)
- Limit registered services profile to only listing extension services
- Rename "Additional Service Provider" to "Extension Services" in the Toolkit config inspector
- Remove all data providers from the registered services profile


Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
```diff
@@ -109,7 +120,9 @@ public override void Enable()
         /// <inheritdoc />
         public override void Update()
         {
-            if (!Application.isPlaying || MixedRealityToolkit.InputSystem == null) { return; }
+            IMixedRealityInputSystem inputSystem = Service as IMixedRealityInputSystem;
```
StephenHodgson: This line occurs an awful lot

davidkline-ms: Service is define as an IMixedRealityService to enable it to store any derived type. This line casts to the expected type, in this case IMixedRealityInputSystem in such a way as to get a null, rather than a typecastexception in the event that the wrong type (or null) is stored.

Assets/MixedRealityToolkit/Interfaces/Services/IMixedRealityDataProvider.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.MixedReality.Toolkit
     /// <summary>
     /// Generic interface for all Mixed Reality Data providers
```
luis-valverde-ms: >Generic interface for all Mixed Reality Data providers [](start = 8, length = 54)

A brief explanation here on when a type should be a data provider instead of just a service could be useful.

davidkline-ms: Good suggestion. Will add that in the near future.

Assets/MixedRealityToolkit/Definitions/SpatialAwareness/BaseSpatialAwarenessObserverProfile.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace Microsoft.MixedReality.Toolkit.SpatialAwareness
+{
+    public abstract class BaseSpatialAwarenessObserverProfile : BaseMixedRealityProfile
+    {
+        [SerializeField]
+        [Tooltip("How should the observer behave at startup?")]
+        private AutoStartBehavior startupBehavior = AutoStartBehavior.AutoStart;
+
+        /// <summary>
+        /// Indicates if the observer is to start immedi ately or wait for manual startup.
```
wiwei: >immedi ately [](start = 50, length = 12)

Nit: immediately

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -607,7 +625,8 @@ private void GestureRecognizer_HoldCompleted(HoldCompletedEventArgs args)
             var controller = GetController(args.source, false);
             if (controller != null)
             {
-                MixedRealityToolkit.InputSystem.RaiseGestureCompleted(controller, holdAction);
+                IMixedRealityInputSystem inputSystem = Service as IMixedRealityInputSystem;
+                inputSystem.RaiseGestureCompleted(controller, holdAction);
```
wiwei: >inputSystem [](start = 16, length = 11)

Unrelated to this change, but do you know why there's only a ? check on the first function (i.e. GestureRecognizer_HoldStarted but none here or below - i.e. I would assume that GestureRecognizer_ManipulationStarted would need something similar if the issue is only on starts)

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
```diff
@@ -26,14 +29,18 @@ public class WindowsSpeechInputProvider : BaseDeviceManager, IMixedRealitySpeech
...
+            Transform playspace,
             string name = null, 
             uint priority = DefaultPriority, 
-            BaseMixedRealityProfile profile = null) : base(registrar, name, priority, profile) { }
+            BaseMixedRealityProfile profile = null) : base(registrar, inputSystem, inputSystemProfile, playspace, name, priority, profile) { }
 
         /// <summary>
         /// The keywords to be recognized and optional keyboard shortcuts.
         /// </summary>
+        // todo: fix...
```
wiwei: > // todo: fix... [](start = 7, length = 16)

Delete this TODO (or fix it before checking in)

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
davidkline-ms: good question. i'll be sure to do a thorough scrub in the near future to ensure we are doing the right thing

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
davidkline-ms: will do. thought I removed all of them :)

StephenHodgson:
Another rubber stamped PR without review from the community?

You guys aren't giving us enough time to review these.

chbecker-ms:
This is a change we had baking for a while which was dependent on the mrtk_development_hl2 code which only moved up last week.

We are driving for transparency by doing as much work as we can in GitHub directly. This is one that didn't make the initial train that populated _hl2.

PR #3741, opened 2019-04-01
Update #if wrappings around speech providers

keveleigh:
Overview
---
The speech and dictation providers had their constructors wrapped inside `#if` statements, which seemed to be preventing their docs from being properly generated. I moved the statements around to reduce the scope they were wrapping.

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
@@ -13,8 +13,6 @@ using UnityEngine.Windows.Speech;
 
 namespace Microsoft.MixedReality.Toolkit.Windows.Input
 {
-#if UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
-
     [MixedRealityExtensionService(SupportedPlatforms.WindowsStandalone | SupportedPlatforms.WindowsUniversal | SupportedPlatforms.WindowsEditor)]
     public class WindowsDictationInputProvider : BaseDeviceManager, IMixedRealityDictationSystem
     {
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
```diff
@@ -31,11 +29,10 @@ public class WindowsDictationInputProvider : BaseDeviceManager, IMixedRealityDic
             uint priority = DefaultPriority,
             BaseMixedRealityProfile profile = null) : base(registrar, name, priority, profile) { }
 
-        /// <summary>
-        /// Is the Dictation Manager currently running?
-        /// </summary>
+        /// <inheritdoc />
         public bool IsListening { get; private set; } = false;
 
+#if UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
```
thalbern: won't this be a compile error if those defines are not set because the braces are not matching anymore?

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
```diff
@@ -41,6 +39,12 @@ public class WindowsSpeechInputProvider : BaseDeviceManager, IMixedRealitySpeech
         /// </summary>
         public IMixedRealityInputSource InputSource = null;
 
+        /// <summary>
+        /// The minimum confidence level for the recognizer to fire an event.
+        /// </summary>
+        public RecognitionConfidenceLevel RecognitionConfidenceLevel { get; set; }
+
+#if UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
```
thalbern: same here

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
keveleigh: Oof yeah, good catch. It causes a few other problems too...working through them now.

keveleigh: I moved them around. The "other problems" I mentioned turned out to be unimportant.

PR #3743, opened 2019-04-01
Input simulation docs

lukastoenneMS:
Overview
---
User documentation for Input Simulation features.

Changes
---
- Fixes: # .


Documentation/InputSimulation/InputSimulationService.md
```diff
@@ -0,0 +1,153 @@
...
+## Enabling the Input Simulation Service
+
+Input simulation is enabled by default in MRTK.
+
+Input simulation is an optional [Mixed Reality service](../../External/Documentation/MixedRealityServices.md). It can be added or removed in to the [Registered Service Providers profile](../TODO.md). Select _Microsoft.MixedReality.Toolkit.Input > InputSimulationService_ as the __Component Type__. The __Runtime Platform__ should always be _Windows Editor_ since the service depends on keyboard and mouse input. The Input Simulation Service has its own profile linked under __Configuration Profile__.
+
+| __Warning__: Any type of profile can be assigned to services at the time of this writing. If you assign a different profile to the service, make sure to use a profile of type _Input Simulation_ or it will not work! |
+| --- |
+
+![Default Input Simulation service settings](../Images/MRTK_InputSimulation_RegisteredServicesEnable.png)
```
wiwei: Can you confirm that this is still the case after Dave submits his changes here:

https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3738/files#diff-72713bda47fb5e09eb55e5066ef13b29

i.e. I see a huge swath of configurations getting axed, one of them being Input Simulation

wiwei: @davidkline-ms

wiwei:
Regarding @cre8ivepark's changes, it would probably be good to not have large files checked in

https://robinwinslow.uk/2013/06/11/dont-ever-commit-binary-files-to-git/

But since then it looks like maybe we can use LFS to manage these instead

Maybe in the interim can we just upload these videos to youtube or something until we have our large file hosting story figured out?

Documentation/InputSimulation/InputSimulationService.md
davidkline-ms: That will indeed need to change. The input simulation service will be registered in the input system profile going forward (and in fact is there by default)

lukastoenneMS:
> 
> 
> Regarding @cre8ivepark's changes, it would probably be good to not have large files checked in
> 
> https://robinwinslow.uk/2013/06/11/dont-ever-commit-binary-files-to-git/
> 
> But since then it looks like maybe we can use LFS to manage these instead
> 
> Maybe in the interim can we just upload these videos to youtube or something until we have our large file hosting story figured out?

I'm going to delete the videos and replace links once we decide where we can host them instead (commits will be squashed in the end, so no binaries in the history).

Documentation/InputSimulation/InputSimulationService.md
lukastoenneMS: The InputSim service was moved to the Input System profile, but

1. The profile for Input Simulation Service is still not type-specific, so the warning is still valid IMO. You can select any of the 20-30 existing default profiles to link here. The service has an internal check, so it will simply not activate, but the asset selection is not filtering profiles by type.
I will have to update the images and paths for accessing the profile though.

2. I don't see the service (aka. data provider - what's the difference again?) being registered by default. I had to add one manually to get hand sim back.

lukastoenneMS:
>     * Could you please move the images to /External/Documentation/Images folder for consistency? We place images for the README files there too. (/External/ReadMeImages/)

Done.

lukastoenneMS:
Videos have been replaced with YouTube links now.

Documentation/InputSimulation/InputSimulationService.md
```diff
@@ -0,0 +1,170 @@
+# Input Simulation Service
+
+The Input Simulation Service emulates the behaviour of devices and platforms that may not be available in the Unity editor. Examples include:
+* Hololens or VR device head tracking
+* Hololens version 1 hand gestures
```
chrisfromwork: nit: may not need to say "version" and could just use HoloLens and HoloLens 2

Documentation/InputSimulation/InputSimulationService.md
```diff
@@ -0,0 +1,170 @@
...
+  <img src="../../External/Documentation/Images/MRTK_InputSimulation_HandSimulationMode.png" title="Full Hand Mesh" width="80%" class="center" />
+</a>
+
+The __Hand Simulation Mode__ switches between two distinct input models.
+
+* _Articulated Hands_: Simulates a fully articulated hand device with joint position data.
+
+   Emulates Hololens 2 interaction model.
+
+   Interactions that are based precise positioning of the hand or use touching can be simulated in this mode.
```
luis-valverde-ms: >based  [](start = 25, length = 6)

based on

PR #3744, opened 2019-04-01
Moved some extensions from Longbow into MRTK. Merged those that alrea…

witian:
Overview
---
Moved some extensions from Longbow into MRTK. Merged those that already exist in MRTK.

Changes
---
Added CanvasExtensions and Texture2DExtensions.
Added new functionality to ArrayExtensions, CameraExtensions, and Color32Extensions.

- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit/Extensions/CameraExtensions.cs
+++ b/Assets/MixedRealityToolkit/Extensions/CameraExtensions.cs
@@ -39,5 +39,33 @@ namespace Microsoft.MixedReality.Toolkit
 
             return (Mathf.Abs(yaw) < horizontalFovHalf && Mathf.Abs(pitch) < verticalFovHalf);
         }
+
+        /// <summary>
+        /// Gets the frustrum size at a given distance from the camera.
+        /// </summary>
+        /// <param name="camera">The camera to get the frustrum size for</param>
+        /// <param name="distanceFromCamera">The distance from the camera to get the frustrum size at</param>
+        /// <returns></returns>
+        public static Vector2 GetFrustrumSizeForDistance(this Camera camera, float distanceFromCamera)
+        {
+            Vector2 frustrumSize = new Vector2
+            {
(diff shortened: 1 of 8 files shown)


Assets/MixedRealityToolkit/Extensions/CameraExtensions.cs
```diff
@@ -39,5 +39,33 @@ public static bool IsInFOV(this Camera camera, Vector3 position)
...
             return (Mathf.Abs(yaw) < horizontalFovHalf && Mathf.Abs(pitch) < verticalFovHalf);
         }
+
+        /// <summary>
+        /// Gets the frustrum size at a given distance from the camera.
+        /// </summary>
+        /// <param name="camera">The camera to get the frustrum size for</param>
+        /// <param name="distanceFromCamera">The distance from the camera to get the frustrum size at</param>
+        /// <returns></returns>
+        public static Vector2 GetFrustrumSizeForDistance(this Camera camera, float distanceFromCamera)
```
keveleigh: I think it's [Frustum](https://en.wikipedia.org/wiki/Frustum), not Frustrum.

witian: You are right! I'll fix that.

Assets/MixedRealityToolkit/Extensions/Texture2DExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <summary>
+        /// Captures a region of the screen and returns it as a texture.
+        /// </summary>
+        /// <param name="x">x position of the screen to capture from (bottom-left)</param>
+        /// <param name="y">y position of the screen to capture from (bottom-left)</param>
+        /// <param name="width">width of the screen area to capture</param>
+        /// <param name="height">height of the screen area to capture</param>
+        /// <returns>A Texture2D containing pixels from the region of the screen specified</returns>
+        /// <remarks>You should call this in OnPostRender.</remarks>
+        public static Texture2D CaptureScreenRegion(int x, int y, int width, int height)
```
StephenHodgson: Why not just use a render texture? It'll probably be more efficient as well.

PR #3745, opened 2019-04-01
Added the Game Object Pool classes from Longbow into MRTK.

witian:
Overview
---
Added the Game Object Pool classes from Longbow into MRTK.

Changes
---
Added GameObjectPool, GameObjectCreator, IGameObjectCreatorHandler, GenericPrefabInstanceCreator
- Fixes: # .


Assets/MixedRealityToolkit/Utilities/GameObjectManagement/GameObjectCreator.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.GameObjectManagement
```
StephenHodgson: Wouldn't this make more sense as a service?

witian: Maybe. This was a direct port from Longbow. I can revisit this to see how it would look as a service and whether or not that would make sense.

witian: After looking into the service architecture a little more deeply for MRTK, I don't think that the Game Object Pool is ideal as a service.
The GameObjectPool is meant to be instanced. While you can have as many creators in a GameObjectPool as you want, it doesn't make since to have a GameObjectPool as a single entity to persist across the experience. The logic for this is that different scenes may have their own GameObjectPools that are meant to go away when the scene goes away, as the objects that are being recycled may not make sense to hang around in other scenes.
Additionally you create your own creators in code. While there is a generic one readily available, it will not satisfy object creation in all cases.

witian: Note that GameObjectPools are not necessarily scene bound either. They can go away whenever you want them to at any point.

Railboy: It would probably make more sense to write a pooling service from scratch than to rip apart / re-assemble this one.

witian: Again, it doesn't necessarily need to be a service. It certainly _could_ be, but the benefit of leaving it as is provides the ability to drop pools as needed. The GameObjectPool is not a singleton or meant to be a single pool. It certainly is _like_ a service, but I'm not convinced it should be unless you can articulate why it is better as a service.

Assets/MixedRealityToolkit/Utilities/GameObjectManagement/GenericPrefabInstanceCreator.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.GameObjectManagement
+{
+    /// <summary>
+    /// Creator for making prefab instances
+    /// </summary>
+    /// <seealso cref="Longbow.Helpers.IGameObjectCreator" />
```
keveleigh: This reference needs updating.

Assets/MixedRealityToolkit/Utilities/GameObjectManagement/GameObjectCreator.cs
Railboy: Agreed, I'm saying it doesn't seem like the benefits of this tool would survive adaptation to a service.

StephenHodgson: Makes sense to me :)

Assets/MixedRealityToolkit/Utilities/GameObjectManagement/IGameObjectCreatorHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.GameObjectManagement
+{
+    /// <summary>
+    /// Optional interface that GameObjects (instantiated and recycled by a
+    /// GameObjectPool) can implement in order to handle preparation for
+    /// recycling and reuse.
+    /// </summary>
+    public interface IGameObjectCreatorHandler
```
StephenHodgson: Do we need `Handler` suffix?

StephenHodgson: What about `IGameObjectPoolEntity` or `Item` or something along those lines?

I think the handler suffix might confuse people with its usage as a handler to a service.

witian: Yes, or at least, something similar. The interface here is not an interface for a GameObject creator, but rather handles actions taken by the creator on GameObjects, such as when it is preparing to recycle a GameObject or is preparing to use it.

witian: In this way a GameObject that is in the pool can have a script that implements that interface (if needed) to carry out actions when recycling or use preparations are carried out on it. Perhaps "**IGameObjectCreatorListener**" or similar would be better as a name?

Assets/MixedRealityToolkit/Utilities/GameObjectManagement/GenericPrefabInstanceCreator.cs
witian: Fixed

Assets/MixedRealityToolkit/Utilities/GameObjectManagement/IGameObjectCreatorHandler.cs
witian: Updated the name to IGameObjectCreatorListener

PR #3748, opened 2019-04-01
Extension Service Creation Wizard

Railboy:
Overview
---
This PR adds a wizard to help newcomers create a basic extension service.

Making the transition from singletons to services can be difficult. This wizard can supplement our other documentation and sample code by enabling devs to create new services with (roughly) the same ease as creating a new Monobehaviour script.

Testing Goals
---
Looking for general feedback on the creation process. Does the tool make clear what you're doing and why?

Also looking for people to break the process any way they can think of - closing the window at odd moments, launching new windows, deleting created components halfway through the process - anything you can do to trip it up and cause errors. The wizard doesn't necessarily need to recover from odd behavior but it should be able to fail gracefully.

Process
---
The wizard is launched from the MixedRealityToolkit menu:

CreateExtensionServiceMenu

Wizard introduces the concept of services and links to documentation. Then it asks devs to choose a name. Class name collisions are automatically detected:

CreateExtensionServiceStart

The purpose of each asset is briefly introduced and output locations are chosen:

CreateExtensionServiceOutput

At this point assets are generated and the wizard waits for recompilation.

CreateExtensionServiceAssets

Generated service scripts include some prompts similar to new Monbehavior scripts:
```
namespace Microsoft.MixedReality.Toolkit.Extensions
{
    [MixedRealityExtensionService(SupportedPlatforms.WindowsStandalone|SupportedPlatforms.MacStandalone|SupportedPlatforms.LinuxStandalone|SupportedPlatforms.WindowsUniversal)]
    public class NewService : BaseExtensionService, INewService, IMixedRealityExtensionService
    {
        private NewServiceProfile newServiceProfile;

        public NewService(IMixedRealityServiceRegistrar registrar,  string name,  uint priority,  BaseMixedRealityProfile profile) : base(registrar, name, priority, profile) 
        {
            newServiceProfile = (NewServiceProfile)profile;
        }

        {
            // Do service initialization here.
        }

        public override void Update()
        {
            // Do service updates here.
        }
    }
}
```

Finally, devs can optionally choose to register their service in the current profile:

CreateExtensionServiceRegister1

If they choose to do so, the service is registered and they are given sample code showing how to access their service:

CreateExtensionServiceRegister3

CreateExtensionServiceRegister2

Issues
---
- You can't return to the wizard to register the service later. Since registration is the most potentially confusing step I feel the wizard should be able to handle this somehow.
- I'm sure there are ways to generate a service that fails to compile. In the event that it doesn't compile the wizard will display an error but it won't be able to recover.
- Doesn't support automatically common extension service interfaces like IDataProvider. This should be a goal.
- Doesn't support implementing base service interfaces like IInputService. I'm not sure this should be a goal, though.


davidkline-ms:
Would also be nice to have a wizard for data providers.

Assets/MixedRealityToolkit/Inspectors/ExtensionTemplates/ExtensionInterfaceTemplate.txt
```diff
@@ -0,0 +1,10 @@
+using System;
+using Microsoft.MixedReality.Toolkit.Extensions;
```
StephenHodgson: Should these files be in the project structure?
Might be better in a child folder of the assets folder or something similar.

Assets/MixedRealityToolkit/Inspectors/ExtensionTemplates/ExtensionServiceWizard.cs
```diff
@@ -0,0 +1,370 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities;
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Editor
+{
+    public partial class ExtensionServiceWizard : EditorWindow
```
StephenHodgson: Why is this a partial class?

Railboy: Relic from earlier iteration, will remove.

Assets/MixedRealityToolkit/Inspectors/ExtensionTemplates/ExtensionInterfaceTemplate.txt
Railboy: It does feel messy. Open to suggestions.

I tried storing templates as read-only literal strings since they're so short but that was also messy.

StephenHodgson: I think think it'll be better to keep it outside the editor so that it doesn't take as long to import. 

Text files always seem to take the longest for whatever reason.

StephenHodgson:
In case I didn't say this already, this is absolutely awesome @Railboy!

This was what I had in mind way back in the day, but we weren't at a place where I thought it was beneficial to have yet.

Railboy:
@Yoyozilla I've created a doc under Utilities. I'm not sure if this is detailed enough. I'm keeping it light since the wizard is sort of supposed to function as a document itself.

Railboy:
Removed _DO NOT MERGE_ tag per discussion with @wiwei

davidkline-ms:
Removing the experimental tag as this feels like a great feature to add.

julenka:
Hi @Railboy we are hoping to complete PRs in time for the RC1 refresh -- could you maybe add documentation like Yoyo requested? Is this something you feel is ready for the RC1 refresh?

Railboy:
Definitely. I've added some documentation under [Utilities.](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3748/commits/c8f478ec3a4021fc5fcce4ea5529d316b4baa9fd#diff-5c9d135302833ae1c8de63e908e4f445)

PR #3749, opened 2019-04-01
Fix MRTK Pointer Unity UI interaction 

danielhofmann-ms:
Overview
---
- Add MixedRealityInputModule inheriting from StandaloneInputModule to add MRTK pointer events for Unity UI.

- Introduce reliable mechanism for a pointer to detect when its current pointer target is about to change.
    - Remove raycast code-duplication (and thus knowledge about FocusProvider implementation) in PokePointer. We need to make sure to also take into account Unity UI graphics raycasts.

- Make all objects of a NearInteractionTouchable subtree touchable (needed for Unity UI).

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
@@ -22,13 +22,17 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
         [SerializeField]
         protected GameObject visuals;
-
-        private Vector3 closestNormal = Vector3.forward;
+        
         private float closestDistance = 0.0f;
 
-        private NearInteractionTouchable currentTouchableDown = null;
+        // The closest touchable component limits the set of objects which are currently touchable.
+        // These are all the game objects in the subtree of the clostest touchable component's owner object.
         private NearInteractionTouchable closestProximityTouchable = null;
-        
+        // The current object that is being touched. We need to make sure to consistently fire 
+        // poke-down / poke-up events for this object. This is also the case when the object within
+        // the same current closest touchable component's changes (e.g. Unity UI control elements).
+        private GameObject currentTouchableObjectDown = null;
+
         protected void OnValidate()
         {
             Debug.Assert(distBack > 0, this);
@@ -51,32 +55,29 @@ namespace Microsoft.MixedReality.Toolkit.Input
             }
 
             // Check proximity
-            closestProximityTouchable = null;
+            NearInteractionTouchable newClosestTouchable = null;
             {
-                float closestDist = distFront; // NOTE: Start at distFront for cutoff
+                closestDistance = distFront; // NOTE: Start at distFront for cutoff
                 foreach (var prox in NearInteractionTouchable.Instances)
                 {
-                    if (!prox.ColliderEnabled)
-                        continue;
-
-                    float dist = prox.DistanceToSurface(Position);
-                    if (dist < closestDist)
+                    if (prox.ColliderEnabled)
                     {
-                        closestNormal = prox.Forward;
-                        closestDist = dist;
-                        closestProximityTouchable = prox;
+                       float dist = prox.DistanceToSurface(Position);
+                       if (dist < closestDistance)
+                       {
+   
+                           closestDistance = dist;
+                           newClosestTouchable = prox;
+                       }
                     }
                 }
-                closestDistance = closestDist;
             }
-            IsActive = IsNearObject;
-            visuals.SetActive(IsNearObject);
 
             // Determine ray direction
             Vector3 rayDirection = Rotation * Vector3.forward;
-            if (closestProximityTouchable != null)
+            if (newClosestTouchable != null)
             {
-                rayDirection = -closestProximityTouchable.Forward;
+                rayDirection = -newClosestTouchable.Forward;
             }
 
             // Build ray (poke from in front to the back of the pointer position)
@@ -84,24 +85,24 @@ namespace Microsoft.MixedReality.Toolkit.Input
             Vector3 end = Position + distFront * rayDirection;
             Rays[0].UpdateRayStep(ref start, ref end);
 
-            // Check if we are about to leave the currently focused object.
-            if (currentTouchableDown != null)
+            // Check if the currently touched object is still part of the new touchable.
+            if (currentTouchableObjectDown != null)
             {
-                RaycastHit hitInfo = default(RaycastHit);
-                PerformRayCastInternal(Rays[0], out hitInfo);
-
-                if (hitInfo.transform?.gameObject != currentTouchableDown.gameObject)
+                if (!IsObjectPartOfTouchable(currentTouchableObjectDown, newClosestTouchable))
                 {
-                    Debug.Assert(Result?.CurrentPointerTarget == currentTouchableDown.gameObject);
-
-                    // We need to raise the event now, since the pointer's focused object will change after we leave this function.
-                    // This will make sure the correct object receives the Leave event.
-                    TryRaisePokeUp(Position);
+                    TryRaisePokeUp(Result.CurrentPointerTarget, Position);
                 }
             }
 
             line.SetPosition(0, Position);
             line.SetPosition(1, end);
+
+            // Set new touchable only now: If we have to raise a poke-up event for the previous touchable object,
+            // we need to to so using the previous touchable in TryRaisePokeUp().
+            closestProximityTouchable = newClosestTouchable;
+
+            IsActive = IsNearObject;
+            visuals.SetActive(IsNearObject);
         }
 
         public override void OnPostSceneQuery()
@@ -111,14 +112,13 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 float dist = Vector3.Distance(Result.StartPoint, Result.Details.Point) - distBack;
                 bool newIsDown = (dist < debounceThreshold);
 
-                // Send events
                 if (newIsDown)
                 {
                     TryRaisePokeDown(Result.CurrentPointerTarget, Position);
                 }
                 else
                 {
-                    TryRaisePokeUp(Position);
+                    TryRaisePokeUp(Result.CurrentPointerTarget, Position);
                 }
             }
 
@@ -126,7 +126,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
             {
                 line.endColor = line.startColor = new Color(1, 1, 1, 0.25f);
             }
-            else if (currentTouchableDown == null)
+            else if (currentTouchableObjectDown == null)
             {
                 line.endColor = line.startColor = new Color(1, 1, 1, 0.75f);
             }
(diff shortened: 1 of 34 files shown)


Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityPointer.cs
```diff
@@ -116,5 +116,10 @@ public interface IMixedRealityPointer : IEqualityComparer
         /// Called after performing the scene query.
         /// </summary>
         void OnPostSceneQuery();
+
+        /// <summary>
+        /// Called during the scene query just before the current pointer target changes.
+        /// </summary>
+        void OnCurrentPointerTargetAboutToChange();
```
wiwei: >OnCurrentPointerTargetAboutToChange [](start = 13, length = 35)

Nit: I haven't really scene "aboutToChange" phrasing in APIs as a common thing (I basic search finds eclipse does this).

i.e. in keeping with other naming this could be OnPreCurrentPointerTargetChange()

This also allows us to add another function later called:

OnPostCurrentPointerTargetChange()

Instead of:

OnCurrentPointerTargetJustChanged()

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -22,13 +22,17 @@ public class PokePointer : BaseControllerPointer, IMixedRealityNearPointer
...
         [SerializeField]
         protected GameObject visuals;
-
-        private Vector3 closestNormal = Vector3.forward;
+        
         private float closestDistance = 0.0f;
 
-        private NearInteractionTouchable currentTouchableDown = null;
+        // The closest touchable component limits the set of objects which are currently touchable.
+        // These are all the game objects in the subtree of the clostest touchable component's owner object.
```
wiwei:
```suggestion
        // These are all the game objects in the subtree of the closest touchable component's owner object.
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -177,52 +172,66 @@ private void TryRaisePokeDown(GameObject targetObject, Vector3 touchPosition)
...
 
-        public bool TryGetNearGraspPoint(out Vector3 position)
+        private static bool IsObjectPartOfTouchable(GameObject targetObject, NearInteractionTouchable touchable)
+        {
+            return targetObject != null && touchable != null &&
+                (targetObject == touchable.gameObject ||
+                // Descendant game objects are touchable as well. In particular, this is needed to be able to send
+                // touch events to Unity UI control elements.
+                (targetObject.transform != null && touchable.gameObject.transform != null &&
+                targetObject.transform.IsChildOf(touchable.gameObject.transform)));
```
wiwei: Is this something that could possibly run with every frame? If so, do you have a sense of how painful the perf impact is of doing this descendent enumeration on each frame?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -704,19 +715,23 @@ private void UpdatePointer(PointerData pointer)
                     // Otherwise, continue
                     LayerMask[] prioritizedLayerMasks = (pointer.Pointer.PrioritizedLayerMasksOverride ?? FocusLayerMasks);
 
-                    // Clear the hit
-                    pointer.ClearHits();
-
                     // Perform raycast to determine focused object
-                    QueryScene(pointer, prioritizedLayerMasks);
+                    PointerHitResult hit = new PointerHitResult(); // TODO: cache and clear
```
wiwei: >// TODO: cache and clear [](start = 67, length = 24)

Nit: Please clean up the TODO prior to checking this in (i.e. shouldn't be per-update allocations)

wiwei: (Same with the thing down below) - i.e. hitUi

---
In reply to: [270985314](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3749#discussion_r270985314) [](ancestors = 270985314)

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -772,60 +811,61 @@ private void ReconcilePointers()
...
                             GameObject closest = null;
                             float closestDistance = Mathf.Infinity;
                             Vector3 objectHitPoint = testPoint;
 
                             foreach (Collider collider in colliders)
                             {
                                 // Policy: in order for an collider to be near interactable it must have
-                                // a NearIneractionGrabbable component on it
+                                // a NearInteractionGrabbable component on it.
+                                // FIXME: This is assuming only the grab pointer is using SceneQueryType.SphereOverlap.
```
wiwei: > // FIXME: This is assuming only the grab pointer is using SceneQueryType.SphereOverlap.  [](start = 31, length = 89)

Can you add a little more context about this FIXME, or file an issue with tracking information on this? It's difficult to understand without additional context how I would approach fixing this, how I would verify it, etc.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -0,0 +1,317 @@
...
+    [RequireComponent(typeof(Camera))]
+    public class MixedRealityInputModule : StandaloneInputModule, IMixedRealityPointerHandler, IMixedRealitySourceStateHandler
+    {
+#if UNITY_EDITOR
+        [UnityEditor.CustomEditor(typeof(MixedRealityInputModule))]
+        public class Editor : UnityEditor.Editor
+        {
+            public override void OnInspectorGUI()
+            {
+                MixedRealityInputModule t = (MixedRealityInputModule)target;
```
wiwei: >t [](start = 40, length = 1)

inputModule (or inputTarget) (I think anything that isn't a single-letter variable name)

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
danielhofmann-ms: Thanks for pointing this out. I actually forgot to address this

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -0,0 +1,317 @@
...
+            public override bool RequiresConstantRepaint()
+            {
+                return true;
+            }
+        }
+#endif
+
+        // TODO: Make sure Unity UI input capturing participates in MRTK focus locking correctly.
+        // TODO: Consider simulating touch events rather than mouse events for PokePointer. What are the tradeoffs?
+
```
wiwei: Nit: It would probably be better to have these tracked as issues rather than TODOs (or having TODOs that point to filed issues). Unless we plan on resolving these in the next week, these will likely get forgotten and then not be helpful in the future. It's unfortunately the fate of all TODOs for people who have too much todo.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -0,0 +1,317 @@
...
+                    if (pointer.IsInteractionEnabled
+                        && pointer.Rays != null
+                        && pointer.Rays.Length > 0
+                        && pointer.SceneQueryType == Physics.SceneQueryType.SimpleRaycast)
+                    {
+                        ProcessMouseEvent((int)pointer.PointerId);
+                    }
+                    else
+                    {
+                        // Remove Unity pointer if it exists, but leave the MRTK pointer in pointersData unless/until the pointer's source device is lost/disconnected.
```
wiwei: >// Remove Unity pointer if it exists, but leave the MRTK pointer in pointersData unless/until the pointer's source device is lost/disconnected. [](start = 24, length = 143)

Nit: This is a great comment, my only comment is it's probably more appropriate to have on the function itself, rather than here in the caller. The reason being, we probably don't wanna copy paste this for each time we call this, and also because this is an implementation detail of the function itself.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -0,0 +1,317 @@
...
+                ProcessMouseEvent((int)pointer.PointerId);
+
+                pointerData.lastMousePoint3d = null; // Mark last mouse position invalid.
+            }
+        }
+
+        /// <summary>
+        /// Adds MRTK pointer support as mouse input for Unity UI.
+        /// </summary>
+        protected override MouseState GetMousePointerEventData(int id)
```
wiwei: >id [](start = 67, length = 2)

Nit: pointerId

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -0,0 +1,317 @@
...
+        protected MouseState GetMousePointerEventDataForMrtkPointer(PointerData pointerData)
+        {
+            IMixedRealityPointer pointer = pointerData.pointer;
+
+            // Reset the RaycastCamera for projecting (used in calculating deltas)
+            Debug.Assert(pointer.Rays != null && pointer.Rays.Length > 0);
+            RaycastCamera.transform.position = pointer.Rays[0].Origin;
+            RaycastCamera.transform.rotation = Quaternion.LookRotation(pointer.Rays[0].Direction);
+
+            // Populate data
```
wiwei: >// Populate data [](start = 12, length = 16)

Nit: This comment doesn't seem to jive with what's done (i.e. "populate" vs "reset")

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -0,0 +1,317 @@
...
+            pointerData.eventDataLeft.Reset();
+
+            Vector3 viewportPos = new Vector3(0.5f, 0.5f, 1.0f);
```
wiwei: >0.5f, 0.5f, 1.0f [](start = 46, length = 16)

Can you add a comment about the significant of these values?

StephenHodgson:
Also, does this PR replace all the input modules with the HoloLens one? Wouldn't that break other platform support?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
danielhofmann-ms: Extended the comment and created this ticket: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3758

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
danielhofmann-ms: Actually the behavior has changed a bit from the original version, so I added a new comment within the function. We're just processing a final mouse event in case the pointer is currently down and   invalidate the last mouse point.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
danielhofmann-ms: This was part of the original MixedRealityInputModule. Will create a ticket.

danielhofmann-ms: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3759

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
danielhofmann-ms: Actually that comment refers to the whole section of setting pointerData,eventDataLeft. Not sure how to better communicate this?

danielhofmann-ms:
> 
> 
> You shouldn't need to add the `MixedRealityInputModule` to enable canvas interactions with input.
> 
> We had specifically designed the input system to utilize the standard input module

`MixedRealityInputModule` inherits from `StandaloneInputModule`. It extends the standard input module  to also handle MRTK pointer events by translating them into mouse events. This also gives us more sophisticated UI behavior like panning / dragging as implemented by `StandaloneInputModule`. Also, this way the `MixedRealityInputSystem` only needs to care about firing MRTK pointer events and not need to replicate Unity UI event firing in some way.

Of course, as you mentioned, all scenes need to be changed to use `MixedRealityInputSystem` instead of `StandaloneInputModule` now. MRTK prints an error if that is not the case. 
I would be curious to know if there is any other way of extending `StandaloneInputModule` and letting it handle generated MRTK pointer mouse events without inheriting from it?

danielhofmann-ms:
> 
> 
> Also, does this PR replace all the input modules with the HoloLens one? Wouldn't that break other platform support?

`MixedRealityInputModule` is not Hololens specific. It adds general MRTK pointer handling for all platforms.

StephenHodgson:
> It adds general MRTK pointer handling for all platforms.

Then is this a change we should propose to Unity to handle correctly?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -0,0 +1,316 @@
...
+using UnityEngine;
+using UnityEngine.EventSystems;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    [RequireComponent(typeof(Camera))]
+    public class MixedRealityInputModule : StandaloneInputModule, IMixedRealityPointerHandler, IMixedRealitySourceStateHandler
+    {
+#if UNITY_EDITOR
+        [UnityEditor.CustomEditor(typeof(MixedRealityInputModule))]
```
StephenHodgson: Editor classes should be separated out into their own files and moved into the appropriate editor folders

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
danielhofmann-ms: Yes, it's possibly run once every frame. Just profiled a bit in VR and it took < 0.01ms (< 0.1% frame time) on average. Not sure if caching this result makes sense, if we want to allow the game object hierarchy to change at runtime.

danielhofmann-ms:
> 
> 
> > It adds general MRTK pointer handling for all platforms.
> 
> Then is this a change we should propose to Unity to handle correctly?

The additional functionality that `MixedRealityInputModule` introduces is MRTK specific: It takes MRTK pointer events from the `MixedRealityInputSystem` and translates them to Unity UI events.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
wiwei: Ah got it - yeah it's tough without having some larger banner comment. This is fairly nit-y but in this case it might just be okay to delete that small comment - it doesn't really add that much to the code - it's clear from the code below that you're setting pointerData.eventDataLeft (i.e. you're populating data), and the rest of the code in this function already has good comments (explaining in detail what each population)

In the end, this is a super minor thing and up to you, I've just gotten feedback on comments before and have left comments similar to this in the past and reflecting on it have found that some were more helpful some were less, etc.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
wiwei: 👍  sounds good, just wanted to get a gut check here to make sure we considered it.

I suppose in our examples we probably have more simple geometry/hierarchy, so just something for us to keep in mind in the future.

PR #3754, opened 2019-04-01
UI Ray-cast distances should take into account the near clip plane - MRTKv2

cellarmation:
This is an equivalent change to pull request [#3666](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3666) which was for HTK, but this is for MRTKv2.

Ray-cast distance calculations should take into account that ray-casts from the UI camera originate from the near clipping plane of the camera since Unity 2017.4.3 and newer versions.

Overview
---

MRTK/HTK does not currently calculate the distance to a UI canvas correctly ([#3359](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3359)) or correctly work out the depth of UI objects when trying to determine if a UI or physics object should get focus ([#3322](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3322)). This is only an issue in Unity 2017.4.3 and later versions due to the introduction of the Unity fix [934908](https://issuetracker.unity3d.com/issues/using-canvas-on-world-space-or-camera-buttons-are-not-clickable-if-pos-z-is-set-less-than-zero)

Changes
---

MRTKv2 claims to support Unity 2018+ so I have assumed that there is no need to variant the code to support versions prior to Unity 2017.4.3.

Lines 783 resolves issue [#3359](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3359)
Lines 847 and 855 are also needed to resolve issue [#3322](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3322)


StephenHodgson:
Are there any before/after tests/screenshots we can have to ensure this change adequately addresses the problem?

cellarmation:
Steps to reproduce are in the linked issues. I explain in both how to reproduce using the example input scenes InputManagerTest.unity &    InteractablesGallery.unity so no test code is needed. Issue #3359 also includes a before and after screenshot from investigation. Can provide more evidence on #3322 if needed.

luis-valverde-ms:
@cellarmation , thanks for spotting and fixing this!

davidkline-ms:
@cellarmation, can you please merge the latest mrtk_development branch into yours and resolve the merge conflict?

thanks!

cellarmation:
I will look into this and commit a proposed fix for comment. The FocusProvider.cs code has changed quite a bit so the change might need to be a bit different.

The overall strategy/fix is to just make sure whenever we use the .distance of a RaycastResult from a UI raycast we add on the nearClipPlane or it won't be comparable to a physics raycast or extend out the expected distance from the original camera location.

I can see there is now code that creates the uiRaycastCamera and trys to set the [nearClipPlane to zero](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/0143b99f7add526f9405ca54013303d92fd9e2ca/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs#L514). This might minimize the impact of the bug. It won't properly resolve it though as you can't have a nearClipPlane of zero its not allowed, Unity will round it up to a small value. This might just make the bug harder to spot, or push it into cases where the user customizes it (if they can?).

davidkline-ms:
@cellarmation,

Thank you for doing this work! it is greatly appreciated

cellarmation:
I have done some testing on the updated code. I was wrong in believing that the nearClipPlane could not be set to zero, that restriction only applies to perspective cameras and the UI raycast camera is orthographic.

The same underlying distance comparison issue exists, so I have updated my push request to resolve the merge conflict and apply the equivalent fix to the new file. However, testing shows mrtk_development in its current state does not now have issue [https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3359](#3359) or issue [https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3322](#3322) because the nearClipPlane is zero. The UI raycast camera is also now made from a private function at run-time so I guess its unlikely that the value would ever be anything other than zero.

I am not sure how worthwhile this push is now. This would make the code more technically correct but wont really have any functional impact with the current code. It might be worth still including as this is a latent bug that could resurface if things are changed in the future.

luis-valverde-ms:
@cellarmation , I'd say that for the sake of simplicity it may be better to leave the code as it is and add an assert documenting the assumption that the near plane is zero:
`Debug.Assert(UIRaycastCamera.nearClipPlane == 0, "Near plane must be zero for raycast distances to be correct");`
If that's fine by you we can do it in your PR or I can do it separately. Thanks again for looking into this!

cellarmation:
Good suggestion, I am happy with that. I have updated the feature branch as you suggested.

PR #3761, opened 2019-04-02
fixed hand mesh loading on hl2

thalbern:
changed flag for gltf model load fail - default case is now that we load the model by calling the base class TryRenderControllerModel. 
fixes model loading for devices that use the articulated hands controller

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
@@ -94,7 +94,7 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input
         private MixedRealityPose currentGripPose = MixedRealityPose.ZeroIdentity;
 
         private bool controllerModelInitialized = false;
-        private bool failedToObtainControllerModel = false;
+        private bool isControllerModelLoaded = false;
 
         #region Update data functions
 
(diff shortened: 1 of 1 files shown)


SimonDarksideJ:
All this does is rename the property used to validate if the model was loaded or not.  What is the fix precisely?

Or is it to ensure the flag is set when the model is actually loaded? (if I'm reading this right)

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
```diff
@@ -607,10 +607,13 @@ private async void CreateControllerModelFromPlatformSDK(uint interactionSourceId
                 }
             }
 
-            if (gltfGameObject == null)
+            if (gltfGameObject != null)
+            {
+                isControllerModelLoaded = true;
```
danielhofmann-ms: Should we assign `isControllerModelLoaded` in any case? 
i.e. 
`isControllerModelLoaded = (gltfGameObject != null);`

Otherwise, if `CreateControllerModelFromPlatformSDK` get called a second time the flag may not be set to the correct value if loading fails.

thalbern:
> All this does is rename the property used to validate if the model was loaded or not. What is the fix precisely?
> 
> Or is it to ensure the flag is set when the model is actually loaded? (if I'm reading this right)

we have a controller for articulated hands on top of mixedRealityController that doesn't do gltf model loading and was relying on the default model loading that is done in TryRenderControllerModel before the gltf change got in. now with those gltf changes the default behavior is not to load anything if we don't go the glft path, so we won't have any hands or joints visible on hololens 2

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
thalbern: good point! I'll change that :)

PR #3763, opened 2019-04-02
Small fixes for PressableButton

danielhofmann-ms:
Overview
---
Mark eventData as being used in PressableButton and remove creation of stale finalMarker object

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
@@ -250,15 +250,19 @@ namespace Microsoft.MixedReality.Toolkit.UI
                     }
                 }
             }
+
+            eventData.Use();
         }
 
         void IMixedRealityTouchHandler.OnTouchUpdated(HandTrackingInputEventData eventData)
         {
             touchPoints.Add(eventData.InputData);
+            eventData.Use();
         }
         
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -250,15 +250,19 @@ void IMixedRealityTouchHandler.OnTouchStarted(HandTrackingInputEventData eventDa
                     }
                 }
             }
+
+            eventData.Use();
```
radicalad: I'm really not sure what the point of eventData.Use() is since any handled event stops bubbling.

danielhofmann-ms: At least for global event handlers, this leads to not propagating the event to the pointers' focused objects. This logic is in `MixedRealityInputSystem.HandleEvent`.

danielhofmann-ms: But I would really be interested what Unity's use of this property is.

Another example: On components on the same game object all events are executed, no matter whether `used` is set or not.

radicalad: Ah okay, so there is a global event handler implication here.

> Another example: On components on the same game object all events are executed, no matter whether used is set or not.

Exactly. Not helpful.

PR #3764, opened 2019-04-02
Create DictationHandler script for easy access to dictation service

luis-valverde-ms:
Created a script to provide basic access to the dictation service without having to resort to scripting, like SpeechInputHandler does for the speech service.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/DictationHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/DictationHandler.cs
@@ -51,7 +51,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// </summary>
         public StringUnityEvent OnDictationError;
 
-        private bool isRecording = false;
         private IMixedRealityDictationSystem dictationSystem;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


wiwei:
Note that this should probably be going into mrtk_development instead of mrtk_development_hl2, now that the HL2 payload is in mrtk_development.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/DictationHandler.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// Event raised when the recognizer stops. Contains the final recognized string.
+        /// </summary>
+        public StringUnityEvent OnDictationComplete;
+
+        /// <summary>
+        /// Event raised when an error occurs. Contains the string representation of the error reason.
+        /// </summary>
+        public StringUnityEvent OnDictationError;
+
+        private bool isRecording = false;
```
wiwei: > private bool isRecording = false; [](start = 7, length = 34)

Is this intended to be used for logging/debugging purposes only? It looks like this is getting set correctly but not inspected.

wiwei: Or was this kinda to make sure that StartRecording doesn't invoke dictationSystem.StartRecording in the case it gets double-called?

---
In reply to: [271491908](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3764#discussion_r271491908) [](ancestors = 271491908)

luis-valverde-ms: The second. It must have been lost when refactoring. Thanks for the catch!

PR #3768, opened 2019-04-02
Spatial Awareness architecture documentation revision

davidkline-ms:
Brings the spatial awareness architecture documentation up to the RC1 code base.

Old images removed / refreshed. Old interface/class/type pages replaced with cross-references to official API docs.


davidkline-ms:
this can be re-targeted at the mrtk_development branch once the _hl2 branch is merged.

Documentation/Architecture/SpatialAwareness/SpatialAwarenessSystemArchitecture.md
```diff
@@ -2,42 +2,68 @@
...
+#### Class Documentation
+
+- [MixedRealitySpatialAwarenessSystem](xref:Microsoft.MixedReality.Toolkit.SpatialAwareness.MixedRealitySpatialAwarenessSystem)
+
+
+### Mesh Observer
+
+The Windows Mixed Reality Spatial Mesh Observer is a great example of a spatial  observer and is included as part of the Mixed Reality Toolkit Windows Mixed Reality provider package. 
+
+The latest released code is available on [GitHub](https://github.com/Microsoft/MixedRealityToolkit-Unity/tree/mrtk_release/Assets/MixedRealityToolkit.Providers/WindowsMixedReality).
```
wiwei: This doesn't link to the direct file (the one above does however - however the one above also actually points to mrtk_development_hl2, which it probably shouldn't).

Maybe just point to mrtk_development for both?

I did some digging and it looks like:

1) It's not possible to reference a true github root.
2) It's not possible for markdown to render differently based on the branch it's in (i.e. having some macro that is replaced with the current branch).

Documentation/Architecture/SpatialAwareness/SpatialAwarenessSystemArchitecture.md
```diff
@@ -2,42 +2,68 @@
...
- 
 ## Example Implementation
 
-The Mixed Reality Toolkit's default implementation of the spatial awareness system can be found in the Assets/MixedRealityToolkit/SDK/Features/SpatialAwarenessSystem folder. This implementation can be used as an example of how to build your own spatial awareness system. It is well documented and demonstrates each of the features described in this architecture specification.
+### Service
+
+The default spatial awareness system is a good example for developers looking to create their own implementation or to learn more about how the system works. It is provided as part of the Mixed Reality Toolkit Spatial Awareness service package. 
+
+The latest released code is available on [GitHub](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development_hl2/Assets/MixedRealityToolkit.Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs).
+
```
wiwei: Unless the plan is to flesh these out in the next change, this section (and the next class documentation section) don't really add that much to this page - it leaves me wondering more than it tells me.

Documentation/Architecture/SpatialAwareness/SpatialAwarenessSystemArchitecture.md
davidkline-ms: hmmm. good catch. i'll take a look

Documentation/Architecture/SpatialAwareness/SpatialAwarenessSystemArchitecture.md
davidkline-ms: do you recommend removing the example implementation section?

Documentation/Architecture/SpatialAwareness/SpatialAwarenessSystemArchitecture.md
davidkline-ms: now pointing to the correct branch

Documentation/Architecture/SpatialAwareness/SpatialAwarenessSystemArchitecture.md
davidkline-ms: xref references (suggested by @thalbern ) are resolved by docfx, I believe

davidkline-ms:
I'll fix :)

PR #3773, opened 2019-04-02
Service Facades / Service Custom Inspectors

Railboy:
Overview
---
Adds editor-only 'service facade' objects to MRTK singleton. Facades are lightweight scene objects representing active services. They resemble classic MonoBehaviour singletons.

They give developers a better sense of what services are active and offer a quick way to check up on the state of the service.

Facades

Testing Goals
---
On the whole we're looking for feedback on ease of use and inspector clarity. In addition:
- Deleting / unparenting / duplicating facade objects should have no effect. Look for situations where these actions cause errors or where facade objects persist in an odd state.
- Look for cases where inspector errors are caused by the state of a service.
- Look for app or service states where toggling 'Use Service Inspectors' in the mixed reality toolkit config profile causes errors.

Documentation Links
---
To help new devs get their bearings, facades link to online docs by way of the DocLink attribute. This functions similarly to the HelpUrl attribute in Unity.

Documentation

Service Custom Inspectors
---
Facades may implement custom inspector / scene GUI / gizmo functions. This is handled similarly to Unity inspectors. A class implementing the IMixedRealityServiceInspector interface is given the MixedRealityServiceInspector attribute, and the facade uses that class to draw the contents of its inspector.

If no inspector is defined, the facade will draw the contents of the service's configuration profile. If the service uses no profile, the facade simply says no inspector is defined for the service.

I've created the following custom inspectors:
- FocusProvider

PointerInspector

- HandJointService

HandJointInspector

- SpatialAwarenessSystem

SpatialObservers

Performance
---
Facades are only used in editor so they won't affect runtime performance. Impact should be minimal but needs to be formally tested. Devs can opt out of using them by un-checking a 'Use Service Inspectors' field in the MRTK configuration profile.

Issues
---
- After starting / stopping a scene, facades vanish until the scene is started again or until recompilation occurs. (This may be an issue with services rather than facades.)

#3545


Railboy:
@davidkline-ms Just to recap - the plan is to merge this into mrtk_development with facades set to OFF by default. This will make it easier for folks to test and provide feedback with minimum disruption.

Once we feel confident that the feature is in good shape we'll set facades to ON in the default config profile.

davidkline-ms:
> Just to recap - the plan is to merge this into mrtk_development with facades set to OFF by default.

Sounds perfect! Will review and test ASAP. Will also help with the known issues :)

Assets/MixedRealityToolkit/Attributes/DocLinkAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// Used primarily by service inspector facades.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
+    public class DocLinkAttribute : Attribute
+    {
+        public DocLinkAttribute(string url) { URL = url; }
+
+        public string URL { get; private set; }
+    }
+}
```
julenka: Missing a newline at end.

Assets/MixedRealityToolkit/Attributes/MixedRealityServiceInspectorAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// Attach to a class implementing IMixedRealityServiceInspector to generate a facade inspector.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    public class MixedRealityServiceInspectorAttribute : Attribute
+    {
+        public MixedRealityServiceInspectorAttribute(Type serviceType)
+        {
+            if (!typeof(IMixedRealityService).IsAssignableFrom(serviceType))
```
julenka: I believe we should be using braces in if statements like

```
if (condition)
{
then do stuff
}
```

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitFacadeHandler.cs
```diff
@@ -0,0 +1,130 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Facades
```
julenka: Do we need a new namespace for this?

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitFacadeHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        static MixedRealityToolkitFacadeHandler()
+        {
+            SceneView.onSceneGUIDelegate += UpdateServiceFacades;
+            EditorApplication.playModeStateChanged += OnPlayModeStateChanged;
+        }
+
+        private static void OnPlayModeStateChanged(PlayModeStateChange obj)
+        {
+            if (MixedRealityToolkit.Instance == null)
```
julenka: Braces

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitFacadeHandler.cs
```diff
@@ -0,0 +1,130 @@
...
+
+                    if (s1Name == s2Name)
+                    {
+                        return s1.Priority.CompareTo(s2.Priority);
+                    }
+
+                    return s1Name.CompareTo(s2Name);
+                });
+
+            for (int i = 0; i < servicesToSort.Count; i++)
```
julenka: Braces...there's more below, would be nice to fix please for consistency.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
```diff
@@ -0,0 +1,256 @@
...
+using UnityEngine;
+using Microsoft.MixedReality.Toolkit.Input;
+using Microsoft.MixedReality.Toolkit.Utilities;
+
+namespace Microsoft.MixedReality.Toolkit.Editor
+{
+    [MixedRealityServiceInspector(typeof(HandJointService))]
+    public class HandJointServiceInspector : BaseMixedRealityServiceInspector
+    {
+        private const string ShowHandPreviewInSceneViewKey = "MRTK_HandJointServiceInspector_ShowHandPreviewInSceneViewKey";
```
julenka: Private fields should start with lower case. See [coding guidelines](https://microsoft.github.io/MixedRealityToolkit-Unity/CodingGuidelines.html#naming-conventions)

Railboy: This is following a convention set by other inspectors. No issue with making them lower case, but we should do it in another PR where we change the other inspectors to be consistent.

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitFacadeHandler.cs
Railboy: I think so. I want to keep facades sort of cordoned off since they only live in the editor.

Railboy:
>I also have not tested the facades in a branch, and I now David Kline is working on something similar. Could you please make sure to get his sign off?

@julenka this is the implementation of the POR @davidkline-ms laid out in #3545 - he has seen it and is on board.

>Finally, for the hand visualization custom editor, is there a way to check / uncheck all of the joints?

There are 'All / None' buttons you can click. You can also just show fingers / fingertips.

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitFacadeHandler.cs
Railboy: @davidkline-ms I know we're trying to keep namespaces to a minimum, what's your take here?

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitFacadeHandler.cs
```diff
@@ -0,0 +1,163 @@
...
+            // Update all self-registered facades
+            foreach (ServiceFacade facade in ServiceFacade.ActiveFacadeObjects)
+            {
+                if (facade == null)
+                {
+                    continue;
+                }
+
+                facade.CheckIfStillValid();
+            }
```
wiwei: There's a bug here where an exception is thrown if you:

1) Enable facades
2) Try to drag any of the facades into like MixedRealityInputSystem into MixedRealityPlayspace.
Hit an exception:

System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
  at System.ThrowHelper.ThrowInvalidOperationException (System.ExceptionResource resource) [0x0000b] in <d7ac571ca2d04b2f981d0d886fa067cf>:0 
  at System.Collections.Generic.List`1+Enumerator[T].MoveNextRare () [0x00013] in <d7ac571ca2d04b2f981d0d886fa067cf>:0 
  at System.Collections.Generic.List`1+Enumerator[T].MoveNext () [0x0004a] in <d7ac571ca2d04b2f981d0d886fa067cf>:0 
  at Microsoft.MixedReality.Toolkit.Utilities.Facades.MixedRealityToolkitFacadeHandler.UpdateServiceFacades (UnityEditor.SceneView sceneView) [0x0021e] in D:\github\MixedRealityToolkit-Unity\Assets\MixedRealityToolkit\Inspectors\MixedRealityToolkitFacadeHandler.cs:97 
  at (wrapper delegate-invoke) <Module>.invoke_void_SceneView(UnityEditor.SceneView)
  at UnityEditor.SceneView.CallOnSceneGUI () [0x000da] in C:\buildslave\unity\build\Editor\Mono\SceneView\SceneView.cs:2848 
  at UnityEditor.SceneView.HandleSelectionAndOnSceneGUI () [0x0000d] in C:\buildslave\unity\build\Editor\Mono\SceneView\SceneView.cs:2125 
  at UnityEditor.SceneView.OnGUI () [0x00376] in C:\buildslave\unity\build\Editor\Mono\SceneView\SceneView.cs:1986 
  at (wrapper managed-to-native) System.Reflection.MonoMethod.InternalInvoke(System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00032] in <d7ac571ca2d04b2f981d0d886fa067cf>:0  occurred

The issue is that face.CheckIfStillValid will destroy the facade (i.e. it will destroy the input system one and then recreate), but this changes the ActiveFacadeObjects collection (because in the ServiceFacade's own  OnDestroy method, it removes itself from the ActiveFacadeObjects list).

The work is probably then to snap a copy of the this array and iterate over the copy of the array instead.

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
```diff
@@ -13,9 +13,10 @@ namespace Microsoft.MixedReality.Toolkit.Input
         typeof(IMixedRealityInputSystem),
         SupportedPlatforms.WindowsEditor,
         "Input Simulation Service",
-        "Profiles/DefaultMixedRealityInputSimulationProfile.asset", 
+        "Profiles/DefaultMixedRealityInputSimulationProfile.asset",
         "MixedRealityToolkit.SDK")]
-    public class InputSimulationService : BaseInputDeviceManager, IInputSimulationService
```
wiwei: >IInputSimulationService [](start = 66, length = 23)

Just for my own understanding, what was the reason for removing the IInputSimulationService from the InputSimulationService implementation?

From a casual reader it seems like this implementation should implement the interface, so I'm curious why it's being removed.

wiwei:
There appears to also be a bug here where choosing another profile under one of the facades doesn't have any effect:

image

Basically if you:
1) Go to the InputSystem facade
2) Click the "targety" button to choose a different input system profile.
3) Click on any other profile to choose them (or use your keyboard to choose them)
4) Notice that it doesn't update the profile.

Assets/MixedRealityToolkit/Utilities/Facades/ServiceFacade.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (service == null || transform.parent != facadeParent)
+            {
+                if (Application.isPlaying)
+                {
+                    GameObject.Destroy(gameObject);
+                }
+                else
+                {
+                    GameObject.DestroyImmediate(gameObject);
+                }
```
wiwei: For my own reference, are we calling DestroyImmediate because we need gameObject == null after this call? And is this not necessary when in play mode?

If this is common knowledge it's probably just a matter of me catching up.

wiwei:
There are some new warnings that I can see in this branch (actually, just one I'm sure about, another one I'm not certain)

I'm not sure about:
A meta data file (.meta) exists but its asset 'Assets/MixedRealityToolkit/Inspectors/InputTemplates' can't be found. When moving or deleting files outside of Unity, please ensure that the corresponding .meta file is moved or deleted along with it.

The one I'm sure about i'm adding right to the line that is causing it

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -0,0 +1,405 @@
...
+            if (string.IsNullOrEmpty(header))
+                header = target.ToString();
+
+            EditorGUI.LabelField(labelRect, header, EditorStyles.boldLabel);
+
+            /*GUILayout.BeginHorizontal();
+            GUILayout.FlexibleSpace();
+            GUILayout.Label(EditorGUIUtility.isProSkin ? logoDarkTheme : logoLightTheme, GUILayout.MaxHeight(128f));
+            GUILayout.FlexibleSpace();
+            GUILayout.EndHorizontal();*/
```
wiwei: Nit: Please remove commented out code prior to checking in.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -0,0 +1,405 @@
...
+
+        [SerializeField]
+        private Texture2D logoLightTheme = null;
+
+        [SerializeField]
+        private Texture2D logoDarkTheme = null;
+
+        protected virtual void Awake()
+        {
+            string assetPath = $"{MixedRealityEditorSettings.MixedRealityToolkit_RelativeFolderPath}/StandardAssets/Textures";
```
wiwei: >  string assetPath = $"{MixedRealityEditorSettings.MixedRealityToolkit_RelativeFolderPath}/StandardAssets/Textures"; [](start = 10, length = 116)

This line is throwing a warning because it's using a deprecated API. Please use the MixedRealityToolkitFiles helper.

I THINK you wanna do something like

MixedRealityToolkitFiles.MRTKDirectories.First(), though Andrei would be more authoritative than I am.

The reason this was done was to make it so that you could import our code/binaries into not the root assets location.

wiwei:
It looks like there may also be a bug in the way we handle array types under inspectors for example:

image

In this picture the user should be able to add Observers (but there is no + button to do so) whereas they can in the main MRTK thing.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/BaseMixedRealityServiceInspector.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.﻿
+
+using System;
+using UnityEditor;
+
+namespace Microsoft.MixedReality.Toolkit.Editor
+{
+    public class BaseMixedRealityServiceInspector : IMixedRealityServiceInspector
```
wiwei: >  public class BaseMixedRealityServiceInspector : IMixedRealityServiceInspector [](start = 2, length = 79)

Nit: Can you add some doccomments for this base class?

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
```diff
@@ -0,0 +1,271 @@
...
+        private const string ShowHandJointFoldoutKey = "MRTK_HandJointServiceInspector_ShowHandJointFoldoutKey";
+        private const string ShowHandJointKeyPrefix = "MRTK_HandJointServiceInspector_ShowHandJointKeyPrefixKey_";
+        private static bool ShowHandPreviewInSceneView = false;
+        private static bool ShowHandJointFoldout = false;
+
+        private const float previewJointSize = 0.02f;
+        private static readonly Color previewJointColor = new Color(0.5f, 0.1f, 0.6f, 0.75f);
+        private static readonly Color enabledColor = GUI.backgroundColor;
+        private static readonly Color disabledColor = Color.Lerp(enabledColor, Color.clear, 0.5f);
```
wiwei: > private static readonly Color disabledColor = Color.Lerp(enabledColor, Color.clear, 0.5f); [](start = 7, length = 91)

Thoughts on having a Colors.cs where these types of constants can go?

I see disabledColor copy pasted across a few different service inspector implementations.

wiwei:
It's a bit odd that you can Add Component to these facade objects, and I'm not really sure how those play out - I don't think it's worth doing magic here to remove user-added components, so maybe in the docs that we write for this we should call it out?

I don't see any way to disable the Add Component button, but also my Unity-fu isn't as strong as yours.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/FocusProviderInspector.cs
```diff
@@ -0,0 +1,55 @@
...
+
+                IPointerResult pointerResult = pointer.Result;
+                if (pointerResult == null)
+                {
+                    EditorGUILayout.ObjectField("Focus Result", null, typeof(GameObject), true);
+                }
+                else
+                {
+                    EditorGUILayout.ObjectField("Focus Result", pointerResult.CurrentPointerTarget, typeof(GameObject), true);
+                }
```
wiwei: I think it might be possible to simplify these lines to 

EditorGUILayout.ObjectField("Focus Result", pointer.Result?.CurrentPointerTarget, typeof(GameObject), true);

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/FocusProviderInspector.cs
```diff
@@ -0,0 +1,55 @@
...
+
+                EditorGUILayout.EndVertical();
+
+                numPointersFound++;
```
wiwei: >numPointersFound [](start = 16, length = 16)

This is a super nit but, since we only care about the presence of pointers, we can just use a bool type (i.e. pointersExist)

wiwei:
I'm also hitting a bug now where:

1) Enable service facades.
2) Notice that facades are correctly added to the scene graph.
3) Hit play
4) After coming back from play, the facades aren't re-added to the scene graph. I have to toggle the "service facades" checkbox.

This seems like maybe one of the more critical bugs to fix (if you can repro this)

wiwei:
The focus provider inspector looks to be empty by default, so I'm not sure what scenarios there are where this should be populated:

image

wiwei:
The back navigation-ability of modifying profiles is a bit weird:

For example:

1) Go the boundary scene facade
image

2) Double click the visualization profile.

3) Click "back to configuration profile"
4) It doesn't take you back.

This feels somewhat busted, though as far as the other issues called out here is probably less troublesome.

wiwei:
The teleport system inspector doesn't appear to have any way of adding events:

image

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
```diff
@@ -0,0 +1,271 @@
...
+                GUI.color = disabledColor;
+                EditorGUILayout.Toggle("Left Hand Tracked", false);
+                EditorGUILayout.Toggle("Right Hand Tracked", false);
+            }
+            else
+            {
+                GUI.color = enabledColor;
+                EditorGUILayout.Toggle("Left Hand Tracked", handJointService.IsHandTracked(Handedness.Left));
+                EditorGUILayout.Toggle("Right Hand Tracked", handJointService.IsHandTracked(Handedness.Right));
+            }
```
wiwei: I'm having a little trouble finding where to use this inspector - can you share some instructions on how to make it visible?

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
```diff
@@ -0,0 +1,271 @@
...
+
+                            case TrackedHandJoint.ThumbTip:
+                            case TrackedHandJoint.ThumbDistalJoint:
+                            case TrackedHandJoint.ThumbMetacarpalJoint:
+                            case TrackedHandJoint.ThumbProximalJoint:
+                                setting = true;
+                                break;
+
+                            default:
+                                break;
```
wiwei: Thoughts on adding an assert/log if we get any other hand joint? This would be a long way in the future but I guess if the platform ever added any more high fidelity joint locations, it might be interesting to get an explicit assert/warning to know that this section needs to be fixed up.

It's also possible that I don't have a fundamentally sound understanding of human anatomy and that this is the exhaustive list of joints, so let me know which it is.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
```diff
@@ -0,0 +1,271 @@
...
+
```
wiwei: Nit: It's a bit weird reading the default case show up before the non-default case, if only because default always executes last regardless of the order here (so seeing out of code visually laid out out of the order in which it executes is a bit odd to see)

That being said, I can't find any coding guidelines that state this explicitly - it seems like it's implicit in guidelines:

[https://google.github.io/styleguide/javaguide.html#s4.8.4-switch](https://google.github.io/styleguide/javaguide.html#s4.8.4-switch)
[https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements](https://google.github.io/styleguide/cppguide.html#Loops_and_Switch_Statements)
[https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es70-prefer-a-switch-statement-to-an-if-statement-when-there-is-a-choice](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es70-prefer-a-switch-statement-to-an-if-statement-when-there-is-a-choice)

This comment is for here and below as well

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
```diff
@@ -0,0 +1,271 @@
...
+                #endregion
+
+                EditorGUILayout.BeginVertical(EditorStyles.helpBox);
+                foreach (TrackedHandJoint joint in Enum.GetValues(typeof(TrackedHandJoint)))
+                {
+                    if (joint == TrackedHandJoint.None)
+                    {
+                        continue;
+                    }
```
wiwei: For my own understanding, what would be the ramifications of setting the value on this hand joint? we special case it a bunch but if there's no harm in setting it (or just enumerating over the relevant set of hand joint enums) it might simplify some of this code.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/BaseMixedRealityServiceInspector.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public virtual bool DrawProfileField { get { return true; } }
+
+        public virtual bool AlwaysDrawSceneGUI { get { return false; } }
+
+        public virtual void DrawGizmos(object target) { }
+
+        public virtual void DrawInspectorGUI(object target) { }
+
+        public virtual void DrawSceneGUI(object target, SceneView sceneView) { }
```
wiwei: >public virtual void DrawSceneGUI(object target, SceneView sceneView) { } [](start = 8, length = 72)

Would it also be possible to have some comments regarding the type that "target" will be - I see that it gets cast to some other interesting things in subclasses, so maybe setting some general expectations (i.e. it will be of type [CustomEditor(typeof(ServiceFacade))] whatever is in CustomEditor) or MixedRealityServiceInspector

In general these types of comments are valuable especially with generic object types, to help scope down the set of things consumers/extenders would have to deal with with mental overhead.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/BaseMixedRealityServiceInspector.cs
wiwei: Ah, nevermind, I guess <inheritdoc> would probably be good enough here because you have it already on IMixedRealityServiceInspector

---
In reply to: [278680928](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3773#discussion_r278680928) [](ancestors = 278680928)

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/BaseMixedRealityServiceInspector.cs
wiwei: Actually this is probably just a comment for IMixedRealityServiceInspector's parameters. 

---
In reply to: [278734920](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3773#discussion_r278734920) [](ancestors = 278734920)

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -0,0 +1,405 @@
...
+        public override void OnInspectorGUI()
+        {
+            OnHeaderGUI();
+
+            ServiceFacade facade = (ServiceFacade)target;
+
+            if (facade.Service == null)
+            {   // Facade has likely been destroyed
+                return;
+            }
```
wiwei: It's interesting to see this check happen here but not in OnHeaderGUI - i.e. if the facade has been destroyed, is it still interesting to do the OnHeaderGUI stuff as well?

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -0,0 +1,405 @@
...
+
+            if (!MixedRealityToolkit.IsInitialized || !MixedRealityToolkit.Instance.HasActiveProfile)
+            {   
+                return;
+            }
+
+            InitializeServiceInspectorLookup();
+
+            bool drawDocLink = DrawDocLink(facade.ServiceType);
```
wiwei: >drawDocLink [](start = 17, length = 11)

Should drawDocLink also be included in drewSomething down below?

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
Railboy: This looks like a confused merge due to the doc link attribute. I'll put it back.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -0,0 +1,405 @@
...
+        /// </summary>
+        /// <param name="serviceType"></param>
+        /// <returns></returns>
+        private bool DrawDataProviders(Type serviceType)
+        {  
+            // If this is a data provider being used by other services, mention that now
+            dataProviderList.Clear();
+            foreach (MixedRealityDataProviderAttribute dataProviderAttribute in serviceType.GetCustomAttributes(typeof(MixedRealityDataProviderAttribute), true))
+            {
+                dataProviderList.Add(" • " + dataProviderAttribute.ServiceInterfaceType.Name);
```
wiwei: >• [](start = 39, length = 1)

O_O

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
Railboy: I agree this would be nice. @chrisfromwork has a related proposal open here: #3908

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -0,0 +1,405 @@
...
+        {
+
+            IMixedRealityServiceInspector inspectorInstance;
+            if (GetServiceInspectorInstance(serviceType, out inspectorInstance))
+            {
+                if (!inspectorInstance.DrawProfileField)
+                {   // We've been instructed to skip drawing a profile by the inspector
+                    return false;
+                }
+            }
```
wiwei: Super Nit: 
if (GetServiceInspectorInstance(serviceType, out inspectorInstance) && !inspectorInstance.DrawProfileField) 
{
  return false;
}

It's just adding an unnecessary extra scope.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
Railboy: I considered it but decided a doc link doesn't really 'count.' Could go either way though.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
Railboy: Good question. I've never seen OnHeaderGUI get called for destroyed targets so Unity must do some under-the-hood check prior.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -0,0 +1,405 @@
...
+
+            bool foundAndDrewProfile = false;
+
+            // Draw the base profile stuff
+            if (typeof(BaseCoreSystem).IsAssignableFrom(serviceType))
+            {
+                SerializedObject activeProfileObject = new SerializedObject(MixedRealityToolkit.Instance.ActiveProfile);
+                // Would be nice to handle this using some other method
+                // Would be nice to handle this with a lookup instead
```
wiwei: >// Would be nice to handle this with a lookup instead [](start = 16, length = 53)

We could create a const/final static dictionary for this class that has

Dictionary<Type, string> = {
  {IMixedRealityInputSystem, "inputSystemProfile"},
  {IMixedRealityBoundarySystem, "boundaryVisualizationProfile"},
 ....
}

And then use that to lookup the property name.

Since it looks like the only delta here is the property name.

Alternatively we could just move:

BaseMixedRealityProfileInspector.RenderProfile(serviceProfileProp, null, false, serviceType);
                    EditorGUILayout.Space();
                    foundAndDrewProfile = true;

Out of each if/elseif statement and into another block like this:

SerializedObject activeProfileObject = new SerializedObject(MixedRealityToolkit.Instance.ActiveProfile);
                // Would be nice to handle this using some other method
                // Would be nice to handle this with a lookup instead
SerializedProperty serviceProfileProp;
                if (typeof(IMixedRealityInputSystem).IsAssignableFrom(serviceType))
                {
                    serviceProfileProp = activeProfileObject.FindProperty("inputSystemProfile");
                    foundAndDrewProfile = true;
                }
                else if (typeof(IMixedRealityBoundarySystem).IsAssignableFrom(serviceType))
                {
                    serviceProfileProp = activeProfileObject.FindProperty("boundaryVisualizationProfile");
                    foundAndDrewProfile = true;
                }
                else if (typeof(IMixedRealityDiagnosticsSystem).IsAssignableFrom(serviceType))
                {
                    serviceProfileProp = activeProfileObject.FindProperty("diagnosticsSystemProfile");
                    foundAndDrewProfile = true;
                }
                else if (typeof(IMixedRealitySpatialAwarenessSystem).IsAssignableFrom(serviceType))
                {
                    serviceProfileProp = activeProfileObject.FindProperty("spatialAwarenessSystemProfile");
                    foundAndDrewProfile = true;
                }

if (foundAndDrewProfile)
{
  BaseMixedRealityProfileInspector.RenderProfile(serviceProfileProp, null, false, serviceType);
                    EditorGUILayout.Space();
                    foundAndDrewProfile = true;
}

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
Railboy: It just asks for a joint that doesn't exist. We can always cache the enums we want on startup to avoid this.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -0,0 +1,405 @@
...
+                 let attribute = classType.GetCustomAttribute<MixedRealityServiceInspectorAttribute>(true)
+                 where attribute != null
+                 select new { ClassType = classType, Attribute = attribute };
+
+            foreach (var result in typesWithMyAttribute)
+            {
+                inspectorTypeLookup.Add(result.Attribute.ServiceType, result.ClassType);
+            }
+
+            initializedServiceInspectorLookup = true;
```
wiwei: >initializedServiceInspectorLookup = true; [](start = 12, length = 41)

Do you know if this implies that if I ever add new service inspectors, I need to reboot unity in order to pick them up (or close my project and reopen my project) or close my scene and reopen my scene?

It looks like this is doing a static one-time initialization, which is fine, and if this is gonna be the behavior we should just document this (i.e. in docs)

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/SpatialAwarenessSystemInspector.cs
```diff
@@ -0,0 +1,136 @@
...
+
+                observerIndex++;
+            }
+        }
+
+        private Color GetObserverColor(int observerIndex)
+        {
+            if (observerIndex >= observerColors.Length)
+            {
+                observerIndex -= observerColors.Length;
```
wiwei: >observerIndex -= observerColors.Length; [](start = 16, length = 39)

This doesn't work out if the observerIndex is > 2x the observerColors length.

For example, you have 13 observers, and the observerColors length is 6.

This will still try to map it to position 7, which doesn't exist.

The canonical way to do wraparound here is just modulo i.e.:

observerIndex = observerIndex % observerColors.Length;

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
Railboy: It will be refreshed every time recompilation occurs.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/HandJointServiceInspector.cs
Railboy: If you're asking where can you see it in action, it exists on the hand joint service object created under the MRTK object. It's just like the pointer inspector.

Railboy:
Other comments:

>The teleport system inspector doesn't appear to have any way of adding events:

As a general rule I'm avoiding inspectors that let you alter services. I think it will be much cleaner if they just let you view its state.

>It's a bit odd that you can Add Component to these facade objects

This is harmless enough but we should try to prevent it. Let's think of ways to catch it in the validation check.

>This seems like maybe one of the more critical bugs to fix (if you can repro this)

This is one of the known issues called out in the **Issues** section.

>The focus provider inspector looks to be empty by default, so I'm not sure what scenarios there are where this should be populated:

If you hit Play, and pointers are instantiated, the list will be populated. We can mention this in a help box, maybe.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/SpatialAwarenessSystemInspector.cs
Railboy: Not sure what I was smoking here - the line should be `observerIndex  = 0;`

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -648,7 +648,6 @@ private void InitializeInstance()
                     if (instance != null)
                     {
                         Debug.Assert(instance.transform.parent == null, "The MixedRealityToolkit should not be parented under any other GameObject!");
-                        Debug.Assert(instance.transform.childCount == 0, "The MixedRealityToolkit should not have GameObject children!");
```
wiwei: > Debug.Assert(instance.transform.childCount == 0, "The MixedRealityToolkit should not have GameObject children!"); [](start = 23, length = 114)

Is it possible to assert that there isn't any non-facade child? I guess this is pretty minor because the user will see it get nuked anyway right.

Railboy: Yeah it should be pretty self-evident what's going on. I'll be eager to hear customer feedback on this one, though.

Assets/MixedRealityToolkit/Utilities/Facades/ServiceFacade.cs
Railboy: DestroyImmediate is how you destroy objects when the application isn't in playmode.

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
Railboy: I'm marking this as resolved but I'd like to revisit it in the future.

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -43,25 +49,80 @@ protected virtual void OnEnable()
...
+        /// <param name="serviceType"></param>Optional service type to limit available profile types.
         /// <returns>True, if the profile changed.</returns>
-        protected static bool RenderProfile(SerializedProperty property, GUIContent guiContent, bool showAddButton = true, Type serviceType = null)
+        public static bool RenderProfile(SerializedProperty property, GUIContent guiContent, bool showAddButton = true, Type serviceType = null)
         {
             return RenderProfileInternal(property, guiContent, showAddButton, serviceType);
         }
 
+        /// <summary>
+        /// Renders a non-ediable object field and an editable dropdown of a profile.
```
wiwei: >ediable [](start = 26, length = 7)

editable

Railboy: For little fixes like this will you do a code suggestion? Kurtis has been using them and they're amazing. You can merge fixes directly from the comment.

wiwei: Normally yes, for larger changes like this I've been using codeflow which doesn't have that integration.

---
In reply to: [279445503](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3773#discussion_r279445503) [](ancestors = 279445503)

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -43,25 +49,80 @@ protected virtual void OnEnable()
...
+        }
+
         /// <summary>
         /// Renders a <see cref="Microsoft.MixedReality.Toolkit.BaseMixedRealityProfile"/>.
         /// </summary>
         /// <param name="property">the <see cref="Microsoft.MixedReality.Toolkit.BaseMixedRealityProfile"/> property.</param>
         /// <param name="showAddButton">Optional flag to hide the create button.</param>
+        /// <param name="serviceType"></param>Optional service type to limit available profile types.
         /// <returns>True, if the profile changed.</returns>
+        /// </summary>
```
wiwei: > /// </summary> [](start = 7, length = 15)

Nit: Extra summary tag (this line can be removed.

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -78,8 +139,45 @@ private static bool RenderProfileInternal(SerializedProperty property, GUIConten
...
+                SessionState.SetBool(showFoldoutKey, showFoldout);
+            }
+
+            // Find the profile type so we can limit the available object field options
+            Type profileType = null;
+            if (serviceType != null)
+            {
+                // If profileTypes has a count greater than one, then it won't be possible to use
+                // EditorGUILayout.ObjectField to restrict the set of profiles to a single type - in this
+                // case all profiles of BaseMixedRealityProfile will be visible in the picker.
```
wiwei: This comment I think was copy pasted from the remarks but it's a bit stale (i.e. it references a types count which doesn't exist within this scope)

wiwei: And also I think that this still leads to an issue where if there are 2+ types it will always only return the first type (is this just something that will never happen?). Basically the comments no longer match the code (i.e. it used to be that 0 == show everything, 1== show only that type, 2 == show everything) but now it's (0 == show everything, 1 == show only that type, 2 == show a random one based on however the underlying collection enumeration works)

---
In reply to: [279475209](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3773#discussion_r279475209) [](ancestors = 279475209)

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitFacadeHandler.cs
```diff
@@ -0,0 +1,171 @@
...
+                    {
+                        continue;
+                    }
+
+                    facade.CheckIfStillValid();
+                }
+            }
+            catch(Exception e)
+            {
+                e = null;
```
wiwei: These lines generate a compiler warning in unity (i.e. e being unused):

Assets\MixedRealityToolkit\Inspectors\MixedRealityToolkitFacadeHandler.cs(109,29): warning CS0168: The variable 'e' is declared but never used

wiwei:
I still don't fully understand how the active pointers thing is supposed to show up - i.e. even in play mode my scene looks like this:

image

And this is with "Use Service Inspectors" on.

wiwei:
Overall this looks this looks great - I think there are still some bugs/issues (i.e. back navigation, edit vs read only) that would prevent us from enabling this on by default.

If I were to summarize the known issues here, would it be accurate to say:
1) Profile navigation (going back up) isn't working because it goes to the 'root' profile.
2) Disjoint set of edit vs read functionality (i.e. you still have to use the full profile experience to configure things until we can do the work to make it super

wiwei:
For any other other bugs that weren't addressed above, can you open an issue to track them just so that we don't lose them, or go through the rediscovery process again in the future.

wiwei:
I don't know what's going on with the builds right now, but I *think* that there's a docs break:


[19-04-29 08:54:15.042]Warning:[BuildCommand.BuildCore.Build Document.LinkPhaseHandlerWithIncremental.ConceptualDocumentProcessor.Save](Documentation/MixedRealityConfigurationGuide.md#L343)Invalid file link:(~/Documentation/Images/MixedRealityToolkitConfigurationProfileScreens/MRTK_EnableServiceInspectors).
	2 Warning(s)

Documentation/MixedRealityConfigurationGuide.md
```diff
@@ -327,4 +328,16 @@ If your controller representation in the scene needs to be offset from the physi
...
+
+## Service Inspectors
+
+Service Inspectors are an editor-only feature that generates in-scene objects representing active services. Selecting these objects displays inspectors which offer documentation links, control over editor visualizations and insight into the state of the service.
+
+![](../Documentation/Images/MixedRealityToolkitConfigurationProfileScreens/MRTK_ServiceInspectors.png)
+
+You can enable service inspectors by checking 'Use Service Inspectors' under Editor Settings in your Configuration Profile.
+
+![](../Documentation/Images/MixedRealityToolkitConfigurationProfileScreens/MRTK_EnableServiceInspectors)
```
wiwei:
```suggestion
![](../Documentation/Images/MixedRealityToolkitConfigurationProfileScreens/MRTK_EnableServiceInspectors.png)
```

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -43,25 +49,80 @@ protected virtual void OnEnable()
         /// <param name="property">the <see cref="Microsoft.MixedReality.Toolkit.BaseMixedRealityProfile"/> property.</param>
         /// <param name="guiContent">The GUIContent for the field.</param>
         /// <param name="showAddButton">Optional flag to hide the create button.</param>
+        /// <param name="serviceType"></param>Optional service type to limit available profile types.
```
wiwei:
```suggestion
        /// <param name="serviceType">Optional service type to limit available profile types.</param>
```

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -43,25 +49,80 @@ protected virtual void OnEnable()
...
+                EditorGUILayout.EndHorizontal();
+            }
+        }
+
         /// <summary>
         /// Renders a <see cref="Microsoft.MixedReality.Toolkit.BaseMixedRealityProfile"/>.
         /// </summary>
         /// <param name="property">the <see cref="Microsoft.MixedReality.Toolkit.BaseMixedRealityProfile"/> property.</param>
         /// <param name="showAddButton">Optional flag to hide the create button.</param>
+        /// <param name="serviceType"></param>Optional service type to limit available profile types.
```
wiwei:
```suggestion
        /// <param name="serviceType">Optional service type to limit available profile types.</param>
```

Issue #3776, opened 2019-04-02
White rectangle visible when drawing the dotted pointer dotted line.


wiwei:
## Overview
@Cameron-Micka looked into this and found that there was something off with the way that  Graphics.DrawMeshInstanced in MeshLineRenderer.cs was drawing the DefaultPointerDottedLine. @Railboy I think you recently added support for single pass instanced meshes

## Steps to reproduce
Get a fresh repo (mrtk_development)
Make sure single pass instanced rendering is enabled (as of the writing of this, it's enabled by default)
Open the Solvers scene in OpenVR.
Notice that when you have a controller pointer near objects, an odd white rectangle shows up.

wiwei:
Given that @Cameron-Micka fixed this so quickly (that is, introducing the workaround to avoid using this dotted pointer stuffs), I can remove the Release Candidate label.

PR #3778, opened 2019-04-02
Temporary Fix for Graphics.DrawMesh/DrawMeshInstanced  + Single Pass Instanced Rendering

Cameron-Micka:
Temporarily making the default pointer use a solid line when not focused on anything. Will fix Graphics.DrawMesh/DrawMeshInstanced  + Single Pass Instanced rendering at a later date.

Changes
---
- Fixes: #3776


wiwei:
A little more context, without this fix what happens is, you'll see a large white rectangular box when in OpenVR, now that we have single pass instanced on by default. Single pass instanced is such a huge win that we really don't want to disable it by default (we don't want to give other developers guidance that multi-pass is acceptable, because in reality, it isn't from what we've seen when people actually are developing real apps).

This makes it so that we don't exercise the codepath that has the issue (the instanced dotted line rendering), while still giving us the perf wins of single pass instanced.

wiwei:
The build CI issue is related to a misconfiguration of the CI system itself post merge - I'll reschedule and handle merging this in, thanks for digging into this so quickly!

wiwei:
Merging this one manually - the build is unrelated (a stale configuration) which I'm working on separately.

PR #3780, opened 2019-04-03
Diagnostics Visual Profiler Improvements

Cameron-Micka:
Porting over latest changes from https://github.com/Microsoft/VisualProfiler-Unity including:

- Adding support for Unity's FrameTimingManager to get more accurate frame times on supported platforms (this also unlocks GPU frame timing for supported platforms).
- Optimizations around how strings are formatted and missed frame indicators are rendered (instanced when available).
- UX improvements around how the window can be anchored.
- Minor tooltip/documentation improvements.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
+++ b/Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
@@ -47,7 +47,7 @@ namespace Microsoft.MixedReality.Toolkit.Diagnostics
             set { isVisible = value; }
         }
 
-        [SerializeField, Range(0.0f, 1.0f), Tooltip("The amount of time, in seconds, to collect frames for frame rate calculation.")]
+        [SerializeField, Tooltip("The amount of time, in seconds, to collect frames for frame rate calculation.")]
         private float frameSampleRate = 0.1f;
 
         public float FrameSampleRate
@@ -122,7 +122,7 @@ namespace Microsoft.MixedReality.Toolkit.Diagnostics
         private Quaternion windowVerticalRotation;
         private Quaternion windowVerticalRotationInverse;
 
-        private Matrix4x4[] frameInfoMatricies;
+        private Matrix4x4[] frameInfoMatrices;
         private Vector4[] frameInfoColors;
         private MaterialPropertyBlock frameInfoPropertyBlock;
         private int colorID;
@@ -172,7 +172,7 @@ namespace Microsoft.MixedReality.Toolkit.Diagnostics
                 }
                 else
                 {
-                    Debug.LogWarning("A shader supporting instancing could not be found for the VisualProfiler, falling back to traditional rendering.");
+                    Debug.LogWarning("A shader supporting instancing could not be found for the VisualProfiler, falling back to traditional rendering. This may impact performance.");
                 }
             }
 
@@ -297,15 +297,15 @@ namespace Microsoft.MixedReality.Toolkit.Diagnostics
                 if (defaultInstancedMaterial != null)
                 {
                     frameInfoPropertyBlock.SetMatrix(parentMatrixID, parentLocalToWorldMatrix);
-                    Graphics.DrawMeshInstanced(quadMesh, 0, defaultInstancedMaterial, frameInfoMatricies, frameInfoMatricies.Length, frameInfoPropertyBlock, UnityEngine.Rendering.ShadowCastingMode.Off, false);
+                    Graphics.DrawMeshInstanced(quadMesh, 0, defaultInstancedMaterial, frameInfoMatrices, frameInfoMatrices.Length, frameInfoPropertyBlock, UnityEngine.Rendering.ShadowCastingMode.Off, false);
                 }
                 else
                 {
                     // If a instanced material is not available, fall back to non-instanced rendering.
-                    for (int i  = 0; i < frameInfoMatricies.Length; ++i)
+                    for (int i  = 0; i < frameInfoMatrices.Length; ++i)
                     {
                         frameInfoPropertyBlock.SetColor(colorID, frameInfoColors[i]);
-                        Graphics.DrawMesh(quadMesh, parentLocalToWorldMatrix * frameInfoMatricies[i], defaultMaterial, 0, null, 0, frameInfoPropertyBlock, false, false, false);
+                        Graphics.DrawMesh(quadMesh, parentLocalToWorldMatrix * frameInfoMatrices[i], defaultMaterial, 0, null, 0, frameInfoPropertyBlock, false, false, false);
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


Assets/MixedRealityToolkit/Definitions/Diagnostics/MixedRealityDiagnosticsProfile.cs
```diff
@@ -35,12 +35,47 @@ public class MixedRealityDiagnosticsProfile : BaseMixedRealityProfile
 
         [SerializeField]
         [Tooltip("The amount of time, in seconds, to collect frames for frame rate calculation.")]
-        [Range(0.01f, 1.0f)]
-        private float frameRateDuration = 0.1f;
+        private float frameSampleRate = 0.1f;
```
davidkline-ms: why remove the range? what would happen if the value were 0 or negative?

davidkline-ms: also, can you add a previouslyserializedas attribute to support legacy profiles?

djee-ms: You mean `FormerlySerializedAs` ?

Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
```diff
@@ -38,109 +47,135 @@ public bool IsVisible
...
+                {
+                    defaultInstancedMaterial = new Material(defaultInstancedShader);
+                    defaultInstancedMaterial.enableInstancing = true;
+                    defaultInstancedMaterial.SetFloat("_ZWrite", 0.0f);
+                    defaultInstancedMaterial.SetFloat("_ZTest", (float)UnityEngine.Rendering.CompareFunction.Disabled);
+                    defaultInstancedMaterial.renderQueue = 5000;
+                }
+                else
+                {
+                    Debug.LogWarning("A shader supporting instancing could not be found for the VisualProfiler, falling back to traditional rendering.");
```
djee-ms: Maybe add a warning "This may impact performance"? Otherwise it is not clear why falling back to traditional rendering is a warning.

Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
```diff
@@ -152,6 +187,22 @@ private void Reset()
...
+
+                MeshFilter quadMeshFilter = GameObject.CreatePrimitive(PrimitiveType.Quad).GetComponent<MeshFilter>();
+
+                if (defaultInstancedMaterial != null)
+                {
+                    // Create a quad mesh with artificially large bounds to disable culling for instanced rendering.
+                    // TODO: Use shared mesh with normal bounds once Unity allows for more control over instance culling.
+                    quadMesh = quadMeshFilter.mesh;
+                    quadMesh.bounds = new Bounds(Vector3.zero, Vector3.one * float.MaxValue);
+                }
```
djee-ms: I am a bit confused. The [documentation](https://docs.unity3d.com/ScriptReference/Graphics.DrawMeshInstanced.html) for `Graphics.DrawMeshInstanced` says:
> Meshes are **not further culled** by the view frustum or baked occluders, nor sorted for transparency or z efficiency.

But there are [many](https://forum.unity.com/threads/instancing-and-culling-make-it-stop.511374/) [posts](https://forum.unity.com/threads/graphics-drawmeshinstanced-and-culling-the-how-and-why.633508/) about culling being performed in practice. So this sounds like a bug. Is there a Unity but report for this? I couldn't find any.

Note that for `DrawMeshInstancedIndirect` there is also some culling apparently, which you can work around via the `Bounds` parameter as described [here](https://answers.unity.com/questions/1349465/how-to-avoid-culling-when-instancing-object-via-dr.html) without having to change the mesh.

Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
```diff
@@ -177,53 +228,96 @@ private void LateUpdate()
...
+            // Draw frame info.
+            if (window.activeSelf)
+            {
+                Matrix4x4 parentLocalToWorldMatrix = window.transform.localToWorldMatrix;
+
+                if (defaultInstancedMaterial != null)
+                {
+                    frameInfoPropertyBlock.SetMatrix(parentMatrixID, parentLocalToWorldMatrix);
+                    Graphics.DrawMeshInstanced(quadMesh, 0, defaultInstancedMaterial, frameInfoMatricies, frameInfoMatricies.Length, frameInfoPropertyBlock, UnityEngine.Rendering.ShadowCastingMode.Off, false);
+                }
```
djee-ms: `frameInfoMatricies` => "Matrices" without the second `i`?

Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
```diff
@@ -423,27 +554,86 @@ private static void OptimizeRenderer(Renderer renderer)
             renderer.allowOcclusionWhenDynamic = false;
         }
 
-        private static void MemoryUsageToString(StringBuilder stringBuilder, string displayedDecimalFormat, TextMesh textMesh, string memoryString, ulong memoryUsage)
+        private static void MemoryUsageToString(char[] stringBuffer, int displayedDecimalDigits, TextMesh textMesh, string prefixString, ulong memoryUsage)
```
djee-ms: My first reaction without context is "Why don't you use `string.Format()`?". Should we add a comment to explain the rationale of manual formatting?

Assets/MixedRealityToolkit/StandardAssets/Shaders/InstancedColored.shader
```diff
@@ -0,0 +1,75 @@
...
+        [Enum(UnityEngine.Rendering.CompareFunction)] _ZTest("ZTest", Int) = 4.0 // LEqual
+        [Enum(UnityEngine.Rendering.CullMode)] _Cull("Cull", Int) = 0.0 // Off
+    }
+
+    SubShader
+    {
+        Pass
+        {
+            Name "Main"
+            Tags{ "RenderType" = "Opaque" "PerformanceChecks" = "False" }
```
djee-ms: Can we add some rationale on why we're disabling a performance check of this shader instead of fixing any potential warning emitted by the check?

Assets/MixedRealityToolkit/Definitions/Diagnostics/MixedRealityDiagnosticsProfile.cs
Cameron-Micka: Great idea about backwards compat with FormerlySerializedAs. I'll add that. As for the range question. I didn't think it made sense to constrain users to the 0-1 range. Since they may want to sample every 2 seconds and so on. If the sample rate is <= 0 then the frame rate will just be sampled every frame.

Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
Cameron-Micka: Great idea, added!

Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
Cameron-Micka: This got me for a bit too, since I couldn't figure out why instances weren't being rendered. Once I looked near the origin they appeared, so I was able to narrow it down to a culling issue. Then I found the numerous Unity forum posts about the issue you pointed out. I'm not aware of a bug report, but it seems like there certainly is a bug, or their documentation is out of date. I have an open discussion with Unity around a similar issue so I'll reach out to them about this too!

Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
Cameron-Micka: Good idea, added!

Assets/MixedRealityToolkit/StandardAssets/Shaders/InstancedColored.shader
Cameron-Micka: Good idea, I just removed that tag since it doesn't really make sense for this shader.

Cameron-Micka:
Thank you David and Jerome for the review! I just made the suggested changes.

Assets/MixedRealityToolkit.Services/DiagnosticsSystem/MixedRealityToolkitVisualProfiler.cs
djee-ms: Thanks!

PR #3781, opened 2019-04-03
Fix the broken OpenVR build.

wiwei:
Some of the new code that was introduced added dependencies on UWP APIs
even when calling from a non-UWP context. (i.e. UserInput_Playback.cs)

Also added a UNITY_WSA check around OpenKeyboard because the touch keyboard
APIs are only documented to work on UWP/IOS/Android/etc locations.

### DIFF
--- a/Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_Visualizer/Scripts/UserInput_Playback.cs
+++ b/Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_Visualizer/Scripts/UserInput_Playback.cs
@@ -113,7 +113,11 @@ namespace Microsoft.MixedReality.Toolkit.Examples.Demos.EyeTracking.Logging
 
             try
             {
+#if WINDOWS_UWP
                 if (!UnityEngine.Windows.File.Exists(filename))
+#else
+                if (!System.IO.File.Exists(filename))
+#endif
                 {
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_Visualizer/Scripts/UserInput_Playback.cs
```diff
@@ -113,7 +113,11 @@ public void LoadNewFile(string filename)
 
             try
             {
+#if WINDOWS_UWP
```
sostel: LoadNewFile () is only loaded in the Editor mode. So, the way it's written now, the following part should never be called: 
`#if WINDOWS_UWP`
`if (!UnityEngine.Windows.File.Exists(filename))`
`...`

See UserInput_Playback.Load() -> LoadInEditor() -> LoadNewFile(...). 

Is `!UnityEngine.Windows.File.Exists(filename)` causing the problem?

wiwei: Exactly - from this site:

https://docs.unity3d.com/ScriptReference/Windows.File.html

"This is an alternative for System.IO.File class. Note: this class is only available when targeting Universal Windows Platform.'

In reality we should probably have some file wrapper that abstracts away the platform-dependent options, but that's for another day.

PR #3782, opened 2019-04-03
Fix the doc build

wiwei:
Previous PR https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3731/files didn't update the toc.yml, but the build didn't catch the issue because it was broken due to SDK versioning issues (and then scene rename issues).

Basically there were a chain of things that were broken and we only now got to the point where we can see it working.

### DIFF
--- a/Documentation/toc.yml
+++ b/Documentation/toc.yml
@@ -1,7 +1,5 @@
 ﻿- name: About MRTK
   items:
-  - name: MRTK-SDK
-    href: ../MRTK-SDK.md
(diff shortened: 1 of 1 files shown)


wiwei:
Looking at the checks, this is the first time we've gotten to a clean build again after all of the merges and activity that we have had.

https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3782/checks

In the interest of ensuring that the rest of the PRs also get this I'm going to force merge this in.

PR #3785, opened 2019-04-03
No longer build appx with WSAPlayer by default

MenelvagorMilsom:
No longer build appx with WSAPlayer by default. Also removed build platform defaulting to x86 for uwp builds.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
@@ -163,7 +163,7 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
                 switch (EditorUserBuildSettings.activeBuildTarget)
                 {
                     case BuildTarget.WSAPlayer:
-                        success = await UwpPlayerBuildTools.BuildPlayer(new UwpBuildInfo(true) { BuildAppx = true });
+                        success = await UwpPlayerBuildTools.BuildPlayer(new UwpBuildInfo(true));
                         break;
                     default:
                         var buildInfo = new BuildInfo(true) as IBuildInfo;
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildInfo.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildInfo.cs
@@ -9,7 +9,6 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
     {
         public UwpBuildInfo(bool isCommandLine = false) : base(isCommandLine)
         {
-            BuildPlatform = "x86";
         }
 
         /// <inheritdoc />


Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs
```diff
@@ -163,7 +163,7 @@ public static async void StartCommandLineBuild()
                 switch (EditorUserBuildSettings.activeBuildTarget)
                 {
                     case BuildTarget.WSAPlayer:
-                        success = await UwpPlayerBuildTools.BuildPlayer(new UwpBuildInfo(true) { BuildAppx = true });
+                        success = await UwpPlayerBuildTools.BuildPlayer(new UwpBuildInfo(true));
                         break;
                     default:
                         var buildInfo = new BuildInfo(true) as IBuildInfo;
```
ForrestTrepte: Why does this switch statement exist? It seems like this code is saying:

````
if (unity platform is UWP)
     build with default options
else
    parse command line options
    build with options configured by command line
````

Why wouldn't we want to use the command-line options for BuildTarget.WSAPlayer? Perhaps there's something I'm not understanding about this? Or could we remove the switch statement and replace it with always doing the code in the default case?

Comment (MenelvagorMilsom) [reply Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs]: I believe this question leads to the reason for having a switch:
> Why wouldn't we want to use the command-line options for BuildTarget.WSAPlayer?

So we do, but UwpPlayerBuildTools has to parse additional options specific to UWP builds. If you look at `UwpPlayerBuildTools.ParseBuildCommandLine`, you'll see this specific implementation.

The intention behind having a switch statement is that in future we may also have other build targets that have build steps or options that are specific to that target. When we provide build tools for those targets we can simply add an additional case to this switch statement. 

@StephenHodgson can correct me if I'm wrong.

Comment (ForrestTrepte) [reply Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UnityPlayerBuildTools.cs]: Aha! I hadn't noticed that UwpPlayerBuildTools.BuildPlayer calls its own ParseBuildCommandLine method while UnityPlayerBuildTools.BuildUnityPlayer doesn't call ParseBuildCommandLine so it has to be performed separately in the default case, but shouldn't be called in the WSAPlayer case. Thanks for explaining!

**Comment by ForrestTrepte** [comment]
This change fixes #3779.

File: Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildInfo.cs [hunk 2]
```diff
@@ -9,7 +9,6 @@ public class UwpBuildInfo : BuildInfo
     {
         public UwpBuildInfo(bool isCommandLine = false) : base(isCommandLine)
         {
-            BuildPlatform = "x86";
```
StephenHodgson: If no default platform is passed in the parameters, what does it default to?

MenelvagorMilsom: There is no assumed default, the build scripts will just fail to produce any artifacts.

StephenHodgson: Yup, that's a problem.

MenelvagorMilsom: I don't see that it is necessarily. We don't have a default platform for any other build target. Why do we need one for uwp builds?

wiwei: I'm guessing that maybe the problem could be that other folks are depending on the current behavior of the script in their own build setup, so changes to default values can cause inadvertent downstream issues because of these implicit dependencies. If this isn't the case it would be good to call out exactly what the problem is.

If a caller of this script expects that specifying a WSA target would actually build the appx, but then running this script doesn't produce anything (i.e. it succeeds but just doesn't produce an output), then that would be a bad thing. 

However, if there were some log saying that "no BuildPlatform provided, not building an appx output" then it would be clear from the logs that the script was never passed the right info to build.

Issue #3722, opened 2019-03-28
Input System raises OnInputChanged every frame for pressed buttons


luis-valverde-ms:
## Overview

It seems that IMixedRealityInputSystem.RaiseOnInputPressed() is meant to be raised every frame a binary input (e.g. binary button) is down: see MouseController.Update and GenericJoystickController.UpdateButton. The event it is triggering though should only be raised on input changes.

I don't think we should be broadcasting the current state of an input every frame. Users should instead rely on the input down/up events and access directly the input source/controller if they need to know the current state. 

We should substitute the input pressed interface with something like RaiseOnFloatInputChanged() that is called to report changes to float input values. Note that WindowsMixedRealityController is already using RaiseOnInputPressed to report changes to trigger button values.

davidkline-ms:
This feels like event overload. i definitely agree with @luis-valverde-ms that this should be fixed.

chrisfromwork:
@luis-valverde-ms is there a perspective on how long this would take to fix? It would be great to have a fix for this in rc1, but if its going to take a lot of time it may slip to rc2.

luis-valverde-ms:
Opened PR for this: #3787

wiwei:
I believe this is done (by #3787) so closing this out.

PR #3787, opened 2019-04-03
Renamed 'input pressed' to 'float input changed' and raise only on changes to inputs

luis-valverde-ms:
This implements the changes suggested in issue #3722

### DIFF
--- a/Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
+++ b/Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
@@ -306,42 +306,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
         #endregion Input Down
 
-        #region Input Pressed
-
-        /// <summary>
-        /// Raise Input Pressed.
-        /// </summary>
-        /// <param name="source"></param>
-        /// <param name="inputAction"></param>
-        void RaiseOnInputPressed(IMixedRealityInputSource source, MixedRealityInputAction inputAction);
-
-        /// <summary>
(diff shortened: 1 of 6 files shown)


luis-valverde-ms:
This PR should also result in the of closing #3253.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
```diff
@@ -306,60 +306,45 @@ public interface IMixedRealityInputSystem : IMixedRealityEventSystem
...
+        #region Float Input Changed
 
         /// <summary>
-        /// Raise the input up event.
+        /// Raise Float Input Changed.
         /// </summary>
         /// <param name="source"></param>
         /// <param name="inputAction"></param>
-        void RaiseOnInputUp(IMixedRealityInputSource source, MixedRealityInputAction inputAction);
+        /// <param name="inputValue"></param>
```
wiwei: ><param name="inputValue"></param> [](start = 12, length = 33)

Can you leave some doccomments about the what this inputValue could be (i.e. is it between 0 and 1? Can it be negative? What does it represent?)

Or is it that it's so specific to the implementor that there's no reason to have this documented here?

luis-valverde-ms: The architecture is so generic that it is very difficult to make any assertions on what this value may represent. I think all we can say is: _this is a float value reported by an input source_.

wiwei: Sounds good!

PR #3794, opened 2019-04-03
PressableButton Inspector Improvement

Railboy:
Overview
---
This was created in response to feedback that the PressableButton settings were difficult to understand.

- Draggable handles for changing button press distance, release threshold and max push distance.
- Visible guides show you where these values exist in space.
- Labels attempt to explain what each plane represents. (These could be improved).
- PressableButton no longer creates / destroys transforms every time push events begin / end. Initial transform is created automatically by inspector. It will create transform if not found at runtime as a last resort.
- Button reports touch / pressed states in inspector.
- Button prefabs have been updated.

Would appreciate feedback on clarity.

ButtonInspector

SideView

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
@@ -17,6 +17,8 @@ namespace Microsoft.MixedReality.Toolkit.UI
     [RequireComponent(typeof(BoxCollider))]
     public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
     {
+        const string InitialMarkerTransformName = "Initial Marker";
+        
         [SerializeField]
         [Tooltip("The object that is being pushed.")]
         private GameObject movingButtonVisuals = null;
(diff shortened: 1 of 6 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -257,44 +190,32 @@ void IMixedRealityTouchHandler.OnTouchStarted(HandTrackingInputEventData eventDa
...
-        private void ClearPathMarkers()
-        {
-            if (initialTransform != null)
+            // If we don't find them, create them
+            if (initialTransform == null)
             {
-                initialTransform.parent = null;
-                DestroyImmediate(initialTransform.gameObject);
-                initialTransform = null;
+                initialTransform = new GameObject(InitialMarkerTransformName).transform;
```
keveleigh: Do we even need to create a new GameObject and save this as a transform? As far as I can tell, we only use its position. Could we just save off the `Vector3` position?

danielhofmann-ms: For now we need at least a transform in order for correctly pushing the button in local space. This way the button can have a parent node that is moving (like a solver).

danielhofmann-ms: Additionally, I think the idea here is to also make the initial transform configurable in editor.

PR #3807, opened 2019-04-03
Cleanup and update of ET navigation demo and scripts

sostel:
Overview
---
- Removed some deprecated parameters in the scroll scripts that were still shown in the Editor.
- Updated hand zoom parameter that was specific to zoom, but was shown for scroll.

### DIFF
--- a/Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_ScrollPanZoom/Scripts/BaseClasses/PanZoomBase.cs
+++ b/Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_ScrollPanZoom/Scripts/BaseClasses/PanZoomBase.cs
@@ -49,24 +49,24 @@ namespace Microsoft.MixedReality.Toolkit.Examples.Demos.EyeTracking
         /// <summary>
         /// Zoom acceleration defining the steepness of logistic speed function mapping.
         /// </summary>
-        internal float zoomAcceleration = 10f;
+        internal float ZoomAcceleration = 10f;
 
         /// <summary>
         ///  Maximum zoom speed. 
         /// </summary>
         [Tooltip("Maximum speed when zooming.")]
-        internal float zoomSpeedMax = 0.02f;
+        internal float ZoomSpeedMax = 0.02f;
 
         /// <summary>
         ///  Minimum scale of the texture for zoom in - e.g., 0.5f (half the original size).
         /// </summary>
-        internal float zoomMinScale = 0.1f;
+        internal float ZoomMinScale = 0.1f;
 
         /// <summary>
         ///  Maximum scale of the texture for zoom out - e.g., 1f (the original size) or 2.0f 
         ///  (double the original size).
         /// </summary>
-        internal float zoomMaxScale = 1.0f;
+        internal float ZoomMaxScale = 1.0f;
 
         /// <summary>
         /// Size of the GameObject's collider when being looked at.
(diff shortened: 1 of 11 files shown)


Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_ScrollPanZoom/Scripts/BaseClasses/PanZoomBase.cs
```diff
@@ -121,7 +121,7 @@ public abstract class PanZoomBase : MonoBehaviour,
         private Vector3 navPos = Vector3.zero;
         private bool isFocused = false;
         internal bool isZooming = false;
-        public bool handZoomEnabledOnStartup = false;
+        internal bool ZoomGestureEnabledOnStartup = false;
```
davidkline-ms: curious as to why this changed to internal

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_ScrollPanZoom/Scripts/PanZoom_RectTransf.cs
```diff
@@ -70,6 +70,10 @@ public class PanZoom_RectTransf : PanZoomBase_RectTransf
         [SerializeField]
         private MixedRealityInputAction Zoom_Gesture = MixedRealityInputAction.None;
 
+        [Tooltip("Enable or disable hand gestures for zooming on startup.")]
+        [SerializeField]
+        private bool Zoom_GestureEnabledOnStartup = false;
```
davidkline-ms: the pattern for MRTK has been to use lower case private field names and not using underscores.

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_ScrollPanZoom/Scripts/BaseClasses/PanZoomBase.cs
sostel: The PanZoomBase class is consumed both by the scroll as well as the pan-and-zoom scripts. The scroll scripts don't use zoom though and hence it doesn't make sense to show this parameter in the Unity Editor.

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Demo_ScrollPanZoom/Scripts/PanZoom_RectTransf.cs
sostel: Good to know . I guess then I have to start a larger cleanup.

davidkline-ms: I just noticed this is in a demo script. while the standards should be the same, I do not believe the team is quite a strict :)

sostel: Nah, this is a valid ask. We were thinking of moving these over to core at some point as they seem to be useful to the community. I'll update it.

sostel: Quick follow up question: Entirely lowercase? Or just the beginning of the name lowercase?

davidkline-ms: just the beginning :)

PR #3811, opened 2019-04-03
Getting started and configuring documentation for the diagnostics system

davidkline-ms:
Getting started and configuration docs.


davidkline-ms:
> please feel free to borrow documentation from here

Trust me, that was the plan all along :)

Issue #3801, opened 2019-04-03
Establish better guidance on the build/OS/tooling requirements


wiwei:
With our latest branches, we now we require insider preview SDKs (along with insider OSes) in order to build the code. 

This is troublesome when trying to build against an older device (i.e. HL1) which doesn't use any of the new APIs. There are a few possibilities here:

1) Make it easier to get access to those SDKs without requiring new OSes. (maybe possible, I feel like I've done this in the past but haven't tried this in recent history)
2) Have some code/build level flags that strip out later SDK requiring code
3) Something else.

wiwei:
Now that the SDK is no longer in preview mode, the recommendation is just to install the public SDK which handles this. I think for future cases (i.e. if there's some other preview API we need to depend on, we'll have to have some recommendation/wrapping of the binaries so that people don't need to install preview SDKs, but for now there's no real action here)

PR #3813, opened 2019-04-03
Update the Getting Started page to mention the required windows SDK.

wiwei:
Due to the new HL2 payload, in order to build against new MRTK you need
to have the latest windows SDK (as of right now it's still in the
insider state).

While we have an issue open to see if we can disentangle this (https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3801) in the interim we should at least document the actual requirements


Yoyozilla:
I think we should have a link to the diag tool doc from here.

PR #3814, opened 2019-04-04
Fix the docs build (again)

wiwei:


wiwei:
I'm forcing this in because it unbreaks the build.

Issue #3791, opened 2019-04-03
Proposal: Simplify input system interface by removing method versions without handedness


luis-valverde-ms:
We currently have two versions for most methods for raising events in IMixedRealityEventSystem, one that receives handedness and one that doesn't. Example:

```
void RaiseOnInputDown(IMixedRealityInputSource source, MixedRealityInputAction inputAction);
void RaiseOnInputDown(IMixedRealityInputSource source, Handedness handedness, MixedRealityInputAction inputAction);
```
I propose removing the version that receives no handedness and change current calls to pass Handedness::None instead.

davidkline-ms:
If this is taken post RC1, we _must_ maintain the non-handedness events as pass-through to the handedness events and mark them as obsolete.

we are at the no breaking changes point.

wiwei:
This sounds like a good thing to do, but I do not believe that this would stop the train. I think that the reality is that there will always be a stream of things we want to improve, but at some point we need to snap to a release in order to get some degree of feedback on that snap (and then obviously iterate on that additional feedback).

To cite a crazy alternative, it's possible that we could ship with this as-is and then customers will tell us that they love the simpler version and really don't care about handedness (which sounds kinda odd to me but hey who knows).

wiwei:
This has been done, closing this out.

PR #3819, opened 2019-04-04
Removed variants of methods to raise input events that receive no handedness.

luis-valverde-ms:
Implemented the changes described in proposal #3791 . As a side effect, this PR changes the handedness of some input events raised by mouse and unity touch controllers from `Handedness.None` to the actual handedness of each controller as defined by `IMixedRealityController.ControllerHandedness`.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
@@ -943,18 +943,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
             };
 
         /// <inheritdoc />
-        public void RaiseOnInputDown(IMixedRealityInputSource source, MixedRealityInputAction inputAction)
-        {
-            inputAction = ProcessRules(inputAction, true);
-
-            // Create input event
-            inputEventData.Initialize(source, inputAction);
-
-            // Pass handler through HandleEvent to perform modal/fallback logic
-            HandleEvent(inputEventData, OnInputDownEventHandler);
-        }
-
-        /// <inheritdoc />
         public void RaiseOnInputDown(IMixedRealityInputSource source, Handedness handedness, MixedRealityInputAction inputAction)
         {
             inputAction = ProcessRules(inputAction, true);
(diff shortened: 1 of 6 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -942,18 +942,6 @@ public void RaisePointerUp(IMixedRealityPointer pointer, MixedRealityInputAction
                 handler.OnInputDown(casted);
             };
 
-        /// <inheritdoc />
-        public void RaiseOnInputDown(IMixedRealityInputSource source, MixedRealityInputAction inputAction)
-        {
-            inputAction = ProcessRules(inputAction, true);
-
-            // Create input event
-            inputEventData.Initialize(source, inputAction);
```
keveleigh: Should we also remove the corresponding `Initialize` variants?

Assets/MixedRealityToolkit/EventDatum/Input/HandTrackingInputEventData.cs
```diff
@@ -30,7 +30,7 @@ public class HandTrackingInputEventData : InputEventData<Vector3>
         /// <param name="touchedObject">This is a the global position of the HandTrackingInputSource that created the EventData</param>
         public void Initialize(IMixedRealityInputSource inputSource, IMixedRealityController controller, Handedness sourceHandedness, Vector3 touchPoint)
```
wiwei: >sourceHandedness [](start = 116, length = 16)

If this is intentionally ignored, it might be worth leaving a comment saying so (i.e. it's weird that we're being told a source handedness for hand tracking data and then saying it doesn't matter)

keveleigh: Hmm yeah, does seem weird that it's ignored. Not sure that was intentional. (Also, looks like the method summary has some stale parameters)

keveleigh: Whoops I didn't refresh!

PR #3821, opened 2019-04-04
Allow extension of MixedRealityInputSystem

provencher:
Overview
---
- MixedRealityInputSystem was not extensible, as methods were not virtual.
- Simply converted all public void methods to be virtual, to make it easier to extend the class without duplicating it.

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
@@ -444,7 +444,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// While input is disabled no events will be sent out and the cursor displays
         /// a waiting animation.
         /// </summary>
-        public void PushInputDisable()
+        public virtual void PushInputDisable()
         {
             ++disabledRefCount;
 
@@ -463,7 +463,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// Pop disabled input state. When the last disabled state is 
         /// popped off the stack input will be re-enabled.
         /// </summary>
-        public void PopInputDisable()
+        public virtual void PopInputDisable()
         {
             --disabledRefCount;
             Debug.Assert(disabledRefCount >= 0, "Tried to pop more input disable than the amount pushed.");
@@ -482,7 +482,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// <summary>
         /// Clear the input disable stack, which will immediately re-enable input.
         /// </summary>
-        public void ClearInputDisableStack()
+        public virtual void ClearInputDisableStack()
         {
             bool wasInputDisabled = disabledRefCount > 0;
             disabledRefCount = 0;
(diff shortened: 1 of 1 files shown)


Railboy:
@SimonDarksideJ @StephenHodgson Are there objections due to performance concerns? This may be a good opportunity to establish some performance testing standards.

provencher:
I rebased the change on the latest development branch, as an api change caused a code conflict

StephenHodgson:
@provencher why not just make a copy and modify what you need and register your custom implementation with the configuration profile?

That's essentially why we created it, so you wouldn't have to extend what was already there.

provencher:
Copying the script is messy and also prone to breaking when the MixedRealityInputSystem is changed. I'd rather extend the public functions to simply add the behavior I need.

For my purposes, I simply need to catch the inputs as they're fired from the platform, to forward them through my codebase. I don't need to change what the core script is doing beyond that.

StephenHodgson:
Makes sense. Maybe when the code is more stable then.

provencher:
It seems this change was reverted in PR https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3847

I understand that the performance costs were undesirable for the default user, and respect that perspective.

However, what about maintaining an extensible version of the script, alongside the more performant non-extensible version? Folks making changes to the script would have to apply it to both for a PR to be merged, and it would grant the flexibility offered by this change, without causing default users to suffer performance costs.

Moreover, this would ensure that a privately maintained copy wouldn't be left behind by rapid changes.

I think this principle could apply to many of the important classes used by the profiles.

StephenHodgson:
Instead of having another copy, is there anything we can do to update the class to better achieve your goal?

Is there something missing from these methods we should be adding?

provencher:
The missing functionality is the ability to catch inputs as they're received by the base class, in order to forward them to other systems before they're discarded.

Currently event handlers process the inputs, but there is no way to hook into that pipeline without adding an event handler to the stack.

StephenHodgson:
I think you could inherit from the base class and override the `HandleEvent` method. 

That's really where the routing takes place, and that method is already virtual.

provencher:
Looking at that method, it's hard for me to gather context about what's happening at the moment the function is called. How can you tell what input raised the event from within an inherited generic HandleEvent override?

StephenHodgson:
You can validate the event data with the type you're expecting. (each of the input event delegates actually do this each time before actually calling the interface method)

```
var baseInputEventData = ExecuteEvents.ValidateEventData<BaseInputEventData>(eventData);

// and to find the input source that raised it:

baseInputEventData.InputSource
```

Once you have the input source you can pretty much get anything you want out of it.

provencher:
I'm surprised that validation method is considered performant enough by the standards that reverted this PR to be honest.

image

I had no idea every event was casted every time it passed through to ensure safety. 

So the solution for me is to add a switch statement that casts the event to every time possible until it hits the type I'm looking for?

provencher:
On another note, @wiwei, would it be possible to get some performance metrics showing the performance hit of virtualizing the functions, given that you reverted the change?

StephenHodgson:
You're right, it probably doesn't need to be validated all the time. Mainly left it in there for people who tried to do something they shouldn't.

davidkline-ms:
Some context around the decision to revert:

(from @andreiborodin) "There are pieces of code that must execute regardless if extension forgets to call base or not; and in those cases we could allow for empty virtual method that are called.

So, MRTK.DoSomethingImportant() can include a call to an empty virtual method OnDoingSomethingImportant() to enable extension but prevent overriding/replacing DoSomethingImportant code."

We are doing some investigation and will work with @provencher and others to land on a good solution.

PR #3824, opened 2019-04-04
Remove unused enum

keveleigh:
Overview
---
Removes an old enum that's no longer used.


wiwei:
I'm going to merge this in and run an all-up single build.

PR #3832, opened 2019-04-04
Update AppBar to make touch events work in all cases

### LINKED ISSUE #3809: AppBar's show/done button is not pressable on the device
## Overview
On HL2 device, App Bar's Show/Adjust/Done buttons are not pressable. 
In the Unity editor, everything works fine.

## Expected Behavior
Works with hand interactions

## Actual Behavior
Does not work with hand press interaction

## Steps to reproduce
Build and deploy HandInteracitonExamples scene.

Railboy:
Overview
---
Updates AppBar to make touch events work in all cases. Buttons no longer fade out when disabled but the visual loss is minimal.

Changes
---
AppBarButton gameobjects are now set active / inactive based on visibility. Also fixed an edge case in NearInteractionTouchable where a disabled collider on an inactive gameobject could still be considered 'enabled.'

Fixes #3809 but doesn't address the underlying issue of touchables not receiving touch events when intersecting with touchables with disabled colliders.

- Fixes: #3809

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/AppBar/AppBarButton.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/AppBar/AppBarButton.cs
@@ -18,7 +18,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
         public int DisplayOrder { get { return displayOrder; } }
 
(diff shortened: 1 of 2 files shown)


cre8ivepark:
Tested on the device. Now the App Bar works well!

PR #3834, opened 2019-04-04
Merges the README.md changes from mrtk_release into mrtk_development

wiwei:
It turns out that we've been checking in conflicting changes to README.md in multiple locations.

This takes most of the changes from mrtk_release (it's been the most active area of change)


wiwei:
I did things this way because merging via a straight up PR from release->development through github itself was a bit scary.

PR #3839, opened 2019-04-04
add mrtk packages documentation

davidkline-ms:
Documents the packaging categories, including details on the contents of foundation packages


davidkline-ms:
I agree @Railboy. Once we have examples, the doc will be updated :)

PR #3845, opened 2019-04-04
Update the readme to point to the required SDK.

wiwei:


README.md
```diff
@@ -43,10 +43,7 @@ MRTK-Unity provides a set of foundational components and features to accelerate
 - Diagnostic Tool
 
 # Getting Started with MRTK 
-1. [Download MRTK](Documentation/DownloadingTheMRTK.md)
-2. Follow this [Getting Started Guide](Documentation/GettingStartedWithTheMRTK.md)
-3. Check out [Mixed Reality Toolkit configuration guide](Documentation/MixedRealityConfigurationGuide.md)
-4. Check out building blocks and example scenes(see the table below)
+Please check out the [Getting Started Guide](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/GettingStartedWithTheMRTK.html)
```
andreiborodin: Could we drop the quick steps here, and say for more details see the guide?

Something like, but a bit more perhaps:
1. Download the MRTK <reference to latest release>
2. Import unity package into your project
3. Run MRTK > Configure

wiwei: I'm just going to revert this part of the change.

wiwei: Actually

wiwei: This guide is a lot more helpful than the existing docs, and I'm still in favor of keeping that around. I don't know how we can strike a balance here of actually helpful detail vs. going too far (i.e. I think the suggestion would provide too few details even if fleshed out).

PR #3862, opened 2019-04-05
Add TargetPositionLockedOnFocusLock for pointers

danielhofmann-ms:
Add `TargetPositionLockedOnFocusLock ` for pointers to to allow the cursor to not be position-locked on HandInteractionPan surfaces

Overview
---
In order for the pannable texture to be fixed to the pointer target position, the pointer target position needs to be movable when focus is locked. This change ensures that the focused object is maintained, even if the pointer ray cast may not be hitting the focused object anymore. This only happens when `IMixedRealityPointer.IsTargetPositionLockedOnFocusLock` is set to false (usually in an event handler). The default setting is to keep the position locked on focus lock.

Since focus lock is handled per pointer, target position lock needs to be handled on the same level.

The initial idea was to only allow a `FocusSessionConfiguration` object to be modified in the FocusEnter event. But since the FocusEnter event is not pointer specific, as it is only raised for the first pointer entering the focus, this is currently not possible.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPan.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPan.cs
@@ -2,14 +2,13 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.MixedReality.Toolkit.Input;
-using Microsoft.MixedReality.Toolkit.Utilities;
+using Microsoft.MixedReality.Toolkit.Physics;
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.EventSystems;
-using UnityEngine.SceneManagement;
 using UnityEngine.Serialization;
 
-public class HandInteractionPan : BaseFocusHandler, IMixedRealityTouchHandler, IMixedRealityInputHandler
+public class HandInteractionPan : BaseFocusHandler, IMixedRealityTouchHandler, IMixedRealityPointerHandler
 {
     private enum TouchType
     {
@@ -50,7 +49,6 @@ public class HandInteractionPan : BaseFocusHandler, IMixedRealityTouchHandler, I
     private Mesh mesh;
     private Vector2 uvOffset = Vector2.zero;
     private Vector2 uvTotalOffset = Vector2.zero;
-    private IMixedRealityInputSource touchingSource;
     private Vector3 touchingInitialPt;
     private Vector3 touchingPoint;
     private Vector2 touchingUV;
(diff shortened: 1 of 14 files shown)


Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityPointer.cs
```diff
@@ -57,6 +57,11 @@ public interface IMixedRealityPointer : IEqualityComparer
         /// </summary>
         bool IsFocusLocked { get; set; }
 
+        /// <summary>
+        /// Specifies whether the pointer's target position (cursor) is locked as well when focus is locked.
```
luis-valverde-ms: >locked [](start = 72, length = 6)

Maybe add locked 'to the target'? I didn't realize what it meant on first read.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -868,14 +868,11 @@ public void RaiseFocusExit(IMixedRealityPointer pointer, GameObject unfocusedObj
         /// <inheritdoc />
         public void RaisePointerDown(IMixedRealityPointer pointer, MixedRealityInputAction inputAction, Handedness handedness = Handedness.None, IMixedRealityInputSource inputSource = null)
         {
+            pointer.IsFocusLocked = (pointer.Result?.Details.Object != null);
```
luis-valverde-ms: >pointer.IsFocusLocked = (pointer.Result?.Details.Object != null); [](start = 12, length = 65)

This seems like a breaking change. I'm fine with it if it doesn't break our current UI but why is it necessary?

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityPointer.cs
danielhofmann-ms: focus lock already means locked to the target. This is specifically for the target position. So "locked to the position target on the target"?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
danielhofmann-ms: Should not break anything, as previously we did set back the focus lock on pointer up, So should always be false at this point.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPan.cs
```diff
@@ -169,30 +165,15 @@ private void ValidityCheck()
         meshFilter = gameObject.GetComponent<MeshFilter>();
         if (meshFilter == null)
         {
-            Debug.Log("The GameObject: " + this.gameObject.name + " " + "does not have a Mesh component.");
+            mesh = null;
+            Debug.Log($"The GameObject: '{gameObject.name}' does not have a Mesh component.");
```
luis-valverde-ms: >Debug.Log [](start = 12, length = 9)

Remove?

PR #3866, opened 2019-04-05
Properly set up example scenes with input module

keveleigh:
Overview
---
A few example scenes were missing the new input module, and some had stale cameras.


cre8ivepark:
@keveleigh I think this is because they are supposed to be loaded through eye tracking root example scene.

keveleigh:
@cre8ivepark Right, yep! This only changes that root example scene. The other, additively loaded scenes are untouched.

PR #3881, opened 2019-04-08
Hand simulation cleanup to better define the purpose of SimulatedHand Data vs. SimulatedHandState.

lukastoenneMS:
Overview
---
SimulatedHandData contains all information necessary for SimulatedHand. It can be used to reproduce hand movement (this will become important for tests!)

SimulatedHandState is used only for keeping track of changes through keyboard and mouse input. It gets baked into SimulatedHandData when hand simulation is used, but other methods of creating hand movement are possible (test animation).

The InputSimulationService now owns the actual SimulatedHandData instances (for both hands) and exposes them publicly. This will later allow tests to modify the hand data without using the SimulatedHandDataProvider, thus separating it from keyboard and mouse input.

Testing for update hand device data happens in the service now instead of the SimulatedHandDataProvider. This avoids the indirection through OnHandDataChanged callbacks and makes the service less dependent on the SimulatedHandDataProvider.

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs
```diff
@@ -15,216 +15,210 @@
...
+        /// </summary>
+        public bool IsAlwaysVisibleRight = false;
 
-        private MixedRealityInputSimulationProfile profile;
+        private SimulatedHandState HandStateLeft;
+        private SimulatedHandState HandStateRight;
 
-        private SimulatedHandState stateLeft;
-        private SimulatedHandState stateRight;
+        // If true then hands are control by user input
```
luis-valverde-ms: >control [](start = 34, length = 7)

controlled

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs
```diff
@@ -255,41 +249,64 @@ public void Update()
...
+                }
             }
-        }
 
-        private void AnimateGesture(SimulatedHandState state, float gestureAnimDelta)
-        {
-            if (state.IsAlwaysTracked)
+            // Update tracked state of a hand.
+            // If hideTimeout value is null, hands will stay visible after tracking stops.
+            DateTime currentTime = DateTime.UtcNow;
```
luis-valverde-ms: >DateTime currentTime = DateTime.UtcNow [](start = 12, length = 38)

Apparently this can be quite unprecise (~50ms intervals): [https://stackoverflow.com/questions/2143140/c-sharp-datetime-now-precision](https://stackoverflow.com/questions/2143140/c-sharp-datetime-now-precision). I'd say that for timestamps in general it's better to use Stopwatch.GetTimestamp: [https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.gettimestamp?view=netframework-4.7.2](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.gettimestamp?view=netframework-4.7.2)

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHand.cs
```diff
@@ -8,6 +8,71 @@
 
 namespace Microsoft.MixedReality.Toolkit.Input
 {
+    /// <summary>
+    /// Snapshot of simulated hand data.
+    /// </summary>
+    [System.Serializable]
+    public class SimulatedHandData
+    {
+        private static readonly int jointCount = Enum.GetNames(typeof(TrackedHandJoint)).Length;
```
luis-valverde-ms: >Enum.GetNames(typeof(TrackedHandJoint)).Length [](start = 49, length = 46)

This will include the 'None' enum value, in case that's an issue.

lukastoenneMS: The purpose of this is to get the necessary array size, so we can use `(int)TrackedHandJoint` as an index. The `None` item is pretty much unused, so not a problem (apart from minor size overhead, not worth optimizing at this point).

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs
lukastoenneMS: Legacy code, but good point anyway. Right now the timestamp is only used to detect if data has been changed in the same frame. I believe it's supposed to prevent multiple hand data updates in the same frame.

I would leave it as is for now, since it works alright. Will have to revisit timestamps anyway when looking into support for simulated update frequency, i.e. simulate 45fps hand updates of the Hololens instead of 60fps Unity updates. I'll add a comment to change this in future.

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHand.cs
```diff
@@ -45,6 +45,8 @@ public void Copy(SimulatedHandData other)
 
         public bool Update(bool isTrackedNew, bool isPinchingNew, HandJointDataGenerator generator)
         {
+            // TODO: DateTime can be quite imprecise, better use Stopwatch.GetTimestamp
```
luis-valverde-ms: Nit: I believe DateTime has quite high internal resolution, what can be imprecise is the value returned by UtcNow

lukastoenneMS: I see, will clarify the comment.

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs
```diff
@@ -15,216 +15,212 @@
...
     public class SimulatedHandDataProvider
     {
         private static readonly int jointCount = Enum.GetNames(typeof(TrackedHandJoint)).Length;
 
+        protected MixedRealityInputSimulationProfile profile;
+
         /// <summary>
-        /// This event is raised whenever the hand data changes.
-        /// Hand data changes at 45 fps.
+        /// If true then the hand is always visible, regardless if simulating.
```
thalbern:
```suggestion
        /// If true then the hand is always visible, regardless of simulating.
```

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs
```diff
@@ -15,216 +15,212 @@
...
         /// </summary>
-        public event Action OnHandDataChanged = delegate { };
-
-        public SimulatedHandData CurrentFrameLeft = new SimulatedHandData();
-        public SimulatedHandData CurrentFrameRight = new SimulatedHandData();
+        public bool IsAlwaysVisibleLeft = false;
+        /// <summary>
+        /// If true then the hand is always visible, regardless if simulating.
```
thalbern:
```suggestion
        /// If true then the hand is always visible, regardless of simulating.
```

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityToolkit.Services.InputSimulation.asmdef
```diff
@@ -1,7 +1,8 @@
 {
     "name": "Microsoft.MixedReality.Toolkit.Services.InputSimulation",
     "references": [
-        "Microsoft.MixedReality.Toolkit"
+        "Microsoft.MixedReality.Toolkit",
+        "Microsoft.MixedReality.Toolkit.SDK"
```
julenka: I notice we are now adding dependency on .SDK which I think is okay since foundation now has toolkit, services, and SDK. But calling it out in case it is not okay.

PR #3886, opened 2019-04-08
Adding EventRouting for Camera events, added option to ClippingPrimitive

radicalad:
**Overview**
Adding a helper script to Utilities that enables components to subscribe to `OnPreRender()` without living on a Camera component. We can add more as needed.

**Context**
I ran into an issue with TextMeshPro objects dynamically creating a sub TMP object at `OnWillRenderObject()`. This means any check for new objects that may exist in `Update()` or `LateUpdate()` will have to wait a minimum of a frame to catch the newly created TMP object.

**Changes**
--- Updated `ClippingPrimitive`. Now has an option to subscribe to  `OnPreRender()`.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/ClippingPrimitive.cs
+++ b/Assets/MixedRealityToolkit/Utilities/ClippingPrimitive.cs
@@ -1,4 +1,4 @@
-﻿  // Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
 using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit/Utilities/ClippingPrimitive.cs
```diff
@@ -93,6 +130,20 @@ protected void OnDisable()
         {
             UpdateRenderers();
             ToggleClippingFeature(false);
+
+            if(cameraMethods != null)
```
Cameron-Micka: Super nit: Missing space after "if."

Assets/MixedRealityToolkit/Utilities/ClippingPrimitive.cs
```diff
@@ -117,6 +168,11 @@ protected void LateUpdate()
             UpdateRenderers();
```
Cameron-Micka: Should be wrap this UpdateRenderers call in a "if (useOnPreRender)" too?

Railboy:
It's frustrating that this workaround is necessary. Probably worth opening a bug against TMP.

Assets/MixedRealityToolkit/Utilities/ClippingPrimitive.cs
radicalad: Its not necessary as `OnCameraPreRender(CameraEventRouter router)` will never get called if `UseOnPreRender` wasn't set.

radicalad: Per Desk conversation, I agree. Its deferring LateUpdate's call to OnPreRender.

radicalad:
I agree. There is literally no way to catch the object unless you have a script on the parent TMP object, or this solution. Absurd.

Assets/MixedRealityToolkit/Utilities/ClippingPrimitive.cs
```diff
@@ -93,6 +130,20 @@ protected void OnDisable()
         {
             UpdateRenderers();
             ToggleClippingFeature(false);
+
+            if (cameraMethods != null)
+            {
+                cameraMethods.OnCameraPreRender -= OnCameraPreRender;
```
StephenHodgson: I noted we unsubscribed properly here in disable, but what about OnEnable?
Does this fail after being disabled once?

radicalad: Okay, had a conversation with @Cameron-Micka about this topic. I didn't know `Start` isn't called when an object is re-enabled. I'm shocked I haven't run into problems yet. :) Good catch - I'll update.

Issue #3788, opened 2019-04-03
Proposal: Decouple Raw Input from Pointer/Cursor/Focus handling


provencher:
## Overview
The service locator system groups everything affected by input into the same service.

This results in one profile dictating focus handler, cursor, pointer and input sources.
There is no separation of Action(Raw Input) and Reaction(Cursor movement, Focus handling, Pointer casting).

## Expected Behavior
A better organization would be to have an Input profile (Action), separated from an interaction profile (Reaction).

This would allow someone to customize how the input is processed, without needing to modify all the waterfall consequences of UX interactions.

## Actual Behavior
In order to change how input is processed, it is necessary to modify how Cursor, Pointer and Focus are handled.

## Steps to reproduce
Try and modify how gestures or buttons are handled before being based onto the UX interaction systems. The root input profile carries too much responsibility.

## Unity Editor Version
Unity 2018.3.8f1

## Mixed Reality Toolkit Release Version
Beta2 - MRTK development branch - commit f664a417a11756c3d545f7a6c9b31f1608e56a1e

chrisfromwork:
@provencher do you feel that this is something that you would like to implement or is this a proposal that you would like to see picked up by another community member?

provencher:
Given that I'm still familiarizing myself with these systems, I don't think I'm the best person to go ahead and design this refactor, let alone implement in a way that works well for all supported platforms, without breaking support for existing users.

I'd very much prefer to see this done by someone more experienced working with this project.

I'm happy to discuss proposals on how best to go about this however.

The way I went about this in a similar system, was to create scriptable objects for all the different inputs.

On one side, the platforms can trigger events in these scriptable objects.

On the other side, the individual buttons, or interaction receivers can have a serialized field with the inputs they'll bind to.

This grants the flexibility of easily repurposing which inputs trigger which actions, on a granular per object level.

The issue with this system, as far as I can tell, reading through the MRTK codebase, is that the events are passed down from the focus handling system, at least in the cases I read through. This has the benefit of only triggering input events for objects that are in focus, but it makes it difficult to program a catch all event for a non focused entity in the scene. It also makes it difficult to repurpose inputs reactions without fundementally changing a behavior script.

Splitting up this behavior is something I'm not sure how best to approach in a non destructive way.

StephenHodgson:
> The service locator system groups everything affected by input into the same service.

[...]

PR #3887, opened 2019-04-08
Global Input Event System

provencher:
Overview
---
Following up on PR https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3821
Along with Proposal https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3788

This change serves to make it easier for users to implement bindings to all inputs supported by the MRTK, without using the focus system, while also preserving all the input parameters passed through the platform data implementations.

- Global event system profile is a new optional profile which allows users of the MRTK to define a mapping through which all inputs route.

- Setting the global input event system profile field in the input profile to none means no global input routing is done, and a simple null check is performed, to minimal performance impact.

- A default profile is created for educational purposes

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityGlobalInputEventSystemProfile.cs
```diff
@@ -0,0 +1,204 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Configuration profile settings for setting configuring global input event responses
+    /// </summary>
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Mixed Reality Global Input Event System Profile", fileName = "MixedRealityGlobalInputEventSystemProfile", order = (int)CreateProfileMenuItemIndices.InputActions)]
+    public class MixedRealityGlobalInputEventSystemProfile : BaseMixedRealityProfile, IMixedRealityGlobalInputEventSystem
+    {
+        public virtual void RaiseSourceDetected(IMixedRealityInputSource source, IMixedRealityController controller = null)
```
StephenHodgson: Profiles are only used to store data, any logic should be contained in the service/system.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputEventSystem.cs
```diff
@@ -0,0 +1,410 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.MixedReality.Toolkit.Input;
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    public interface IMixedRealityInputEventSystem
```
StephenHodgson: a new interface for this?

what happened to the existing one?

StephenHodgson: Oh I see, it was just moved.

StephenHodgson: Was there any actual changes to this class or was it just moved?

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityGlobalInputEventSystemProfile.asset
```diff
@@ -0,0 +1,15 @@
...
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
+  m_GameObject: {fileID: 0}
+  m_Enabled: 1
+  m_EditorHideFlags: 0
+  m_Script: {fileID: 11500000, guid: 0b7fd900dbf72e340b2bc70ee7cec819, type: 3}
+  m_Name: MixedRealityGlobalInputEventSystemProfile
+  m_EditorClassIdentifier: 
+  isCustomProfile: 1
```
StephenHodgson: Why have a profile if no configurable data is stored in it?

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityGlobalInputEventSystemProfile.cs
provencher: I looked into making it a service - however it doesn't need to have a lifecycle, which is mandated by that system.

MixedRealityGlobalInputEventSystemProfile is meant to be an extensible template. 
In my application, I can expose events to the editor if I choose to do so. 

In practice, it is easier to use and extend as a profile.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputEventSystem.cs
provencher: No this is indeed a new interface.

IMixedRealityInputEventSystem is different from IMixedRealityInputSystem, in that IMixedRealityInputSystem implements IMixedRealityInputEventSystem and  IMixedRealityEventSystem.

This is just a logical split of the events, in such a way that IMixedRealityInputEventSystem contains all the event responses listed in IMixedRealityEventSystem. This allows another class, like the MixedRealityGlobalInputEventSystemProfile, to implement only the input event methods.

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityGlobalInputEventSystemProfile.asset
provencher: As I explained above, the profile here is simply a template. 

This profile is meant to be extensible, and as such, a child implementation can expose configurable data to their app.

provencher:
@luis-valverde-ms 
As I've explained in #3788 - it seems I misunderstood some aspects of the global input listener system.

I've managed to make use of it in a satisfactory manor - however, as I also pointed out in that proposal, I do believe the string system for conducting the mappings between actions leaves a lot to be desired, as is rather prone to error from mistyping a string.

That said, I'm able to get the work I need to done, so thanks for the assistance!

StephenHodgson:
> I do believe the string system for conducting the mappings between actions leaves a lot to be desired, as is rather prone to error from mistyping a string.

Just want to clarify, we're not using a string based lookup

provencher:
@StephenHodgson - right there's not a string based lookup, but when you want to retrieve the action associated with say - Select, the only way to know what MixedRealityInputAction maps to Select, is to check it's Description string.

Unless I'm mistaken about this, in which case, please enlighten me.

luis-valverde-ms:
@provencher , the specific actions to use are usually setup manually in scripts by users in the editor. When would you need to find a "Select" action from code and how would making the action an scriptable object would help you in that case? Would you not just end up with a harcoded asset path instead of a string?

provencher:
@luis-valverde-ms the scriptable object would be something I'd pass as a serialized reference into a monobehavior that's listening to the input action. Even if the path to the scriptable object is hardcoded, it would still reference a common asset which would contain an action name that could change. 

MRTK code would reference that scriptable object directly, and so if it were to change, then it would also change in my project, since the reference would be the same.

The problem with hardcoded strings is that if the name of action changes on a profile, it will no longer work in my listener script.

StephenHodgson:
@provencher why are you using the string? I don't think it was ever intended to be used that way. 

Better to just check against the struct itself which is why we wrote the overload for the equality operator.

StephenHodgson:
>Right there's not a string based lookup, but when you want to retrieve the action associated with say - Select, the only way to know what MixedRealityInputAction maps to Select, is to check it's Description string.
>
>Unless I'm mistaken about this, in which case, please enlighten me.

No we check the id based on the input action profile. This profile was always intended to be static, and wouldn't really change much after setup unless you add more actions.

StephenHodgson:
The other reason why we choose a `struct` instead of a ScriptableObejct was so that we could later convert it to ECS

provencher:
@StephenHodgson 

From my experience, those actions are not globablly defined in a statically accessible way.
This seems to be because different platforms support different input actions, and the actions are defined in a serialized list in the profile.

This means that, in order to know what actions are defined, I have to look at the editor and know which strings exist and what their constraints are.

Because of this, I have to make this list:

`        // Default input mappings
        // Digital
        private const string cSelect = "Select";
        private const string cMenu = "Menu";

        // 6 Dof
        private const string cGripPose = "GripPose";
        private const string cPointerPose = "Pointer Pose";
        private const string cIndexFingerPose = "Index Finger Pose";

        // Dual Axis
        private const string cTeleportDirection = "Teleport Direction";
        private const string cScroll = "Scroll";
        private const string cMouseDelta = "Mouse Delta";

        // Single Axis
        private const string cTrigger = "Trigger";
        private const string cGripPress = "Grip Press";

        // No Axis
        private const string cHoldAction = "Hold Action";
        private const string cManipulateAction = "Manipulate Action";
        private const string cNavigationAction = "Navigation Action";`

Then I have to cache all the locally created input actions

`        protected override void OnEnable()
        {
            base.OnEnable();

            // Reset mapping dictionary
            _actionMappingDictionary.Clear();

            // Locally cache all input mappings
            foreach (var action in ActiveInputActionProfile.InputActions)
            {
                _actionMappingDictionary.Add(action.Description, action);
            }`

Then in order to see what action was passed in through an event, I have to check the descriptor as a key in my dictionary

`      private MixedRealityInputAction GetActionForInput(string input)
        {
            _actionMappingDictionary.TryGetValue(input, out var action);
            return action;
        }


            if (GetActionForInput(cHoldAction) == action)`

So in essence, I don't compare the action to the string, but in code, triggering code according to the action given requires knowing the descriptor ahead of time.

StephenHodgson:
> From my experience, those actions are not globally defined in a statically accessible way.

That's a result of MSFT changing the intended use and architecture of the toolkit and part of why some of the other developers forked.

> This means that, in order to know what actions are defined, I have to look at the editor and know which strings exist and what their constraints are.

Not exactly. It was intended to set the input action in the editor inspector `InputHandler` as it provided an easy dropdown with all of the available actions a user could take.

Now, as far as I understand, you're attempting to mesh the established framework you've been working on with the toolkits, is that correct? imho it is never good to combine two frameworks together, but you will inevitably need to choose either one or the other.

StephenHodgson:
> Not exactly. It was intended to set the input action in the editor inspector InputHandler as it provided an easy dropdown with all of the available actions a user could take.

This def makes setting it in code much harder, yes and that's where the ScriptableObject updates can def help :)

provencher:
Yeah the scriptable objects would definitely help my use case. 

Regarding the meshing of frameworks. I don't that's the issue here. My goal was simply to identify all the actions you can initiate from a global event listener. That should be a standard usecase of the MRTK.

StephenHodgson:
Also, one last thing to note, you could always get the list of input actions form the InputActionProfile via

`ActiveProfile.InputSystemProfile.InputActionsProfile.InputAction[]` Yeah it's a little convoluted, but might be easier to get if the current active profiles of the input system were exposed in a readonly accessor:

`IInputSystem.InputActions` or something like that

provencher:
@StephenHodgson  if you look at the code sample I pasted above, I do access that list, however it does not tell me which input actions they are. That's why I cache that list in a dictionary using the description as the key, and hardcode the names of actions to access them

StephenHodgson:
Well if you're able to access that list at edit time, you could assign them in a ScriptableObject to match/pair them up without having to keep a dictionary.

provencher:
Sure I could, though Luis already opened a PR to that so I won't have to.

The dictionary makes it so I still compare input actions instead of strings at runtime.

I had to write less code as well to get this working right now. 

My point is that the actions themselves should be statically available, and I just wrote that functionality myself.

StephenHodgson:
This kinda goes back the the point I was trying to make about getting both of these frameworks you're working with to properly do what you'd like.

I think the approach you're taking isn't the easiest path, and I'd like to better understand the desire to work with it in this way and what some of your requirements are. I'm thinking there's an easier way to do what you're trying to accomplish. Let's chat about it some more with some examples tomorrow.

PR #3889, opened 2019-04-08
Updates CameraCache to handle Main Camera / Playspace destruction

Railboy:
Overview
---
Updates CameraCache to generate a main camera if no cameras are found in the scene. Previously, this would cause multiple null reference exceptions.

Destroying the playspace (accidentally or intentionally), adding the Toolkit to a completely empty scene or destroying the main camera are all common scenarios that the toolkit should be able to handle.

After creating the camera everything proceeds normally including the creation of a MixedRealityPlayspace transform and camera event system. DontDestroyOnLoad settings are also observed.


Railboy:
Closing this to regarget mrtk_development instead of mrtk_release

PR #3890, opened 2019-04-08
Updates CameraCache to handle Main Camera / Playspace destruction

### LINKED ISSUE #4046: "Add to Scene and Configure" doesn't gracefully fail when no camera is present in the scene.
## Describe the bug

When the "Mixed Reality Toolkit->"Add to Scene and Configure..." menu item is selected in a scene **without** a camera a handful of null reference exceptions occur and the MRTK is not configured correctly.

The project we are seeing this occur in loads a camera dynamically at runtime based on certain project restrictions.

## To reproduce

Select "Mixed Reality Toolkit->"Add to Scene and Configure..." in a scene without a camera.

## Expected behavior

All exceptions are gracefully handled. Optionally, anything relying on the Camera or Camera Cache could automatically setup a main camera for the user.

Railboy:
Overview
---
Updates CameraCache to generate a main camera if no cameras are found in the scene. Previously, this would cause multiple null reference exceptions.

Destroying the playspace (accidentally or intentionally), adding the Toolkit to a completely empty scene or destroying the main camera are all common scenarios that the toolkit should be able to handle.

After creating the camera everything proceeds normally including the creation of a MixedRealityPlayspace transform and camera event system. DontDestroyOnLoad settings are also observed.

Testing Goals
---
- In editor, you should be able to delete the camera and/or the playspace at any time without causing issues. The goal is to find a scenario where deleting one or both causes an error.
- On device we should test scenarios where we load a scene with an MRTK instance but no camera or playspace, or else create a new MRTK instance in an empty scene.

Fixes #4046


Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -483,6 +483,11 @@ private void InitializeServiceLocator()
 
         private void EnsureMixedRealityRequirements()
         {
+            if (MixedRealityPlayspace == null)
```
StephenHodgson: Isn't this dynamically created?

Where is that created again, and what would make this null?

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
```diff
@@ -82,7 +82,7 @@ protected override void OnEnable()
                 }
             }
 
-            if (!MixedRealityToolkit.ConfirmInitialized())
+            if (!MixedRealityToolkit.IsInitialized)
```
StephenHodgson: `ConfirmInitialized()` is intentionally called here to ensure that the setup phase works correctly. What's the benefit of just bailing if it's not initialized?

Assets/MixedRealityToolkit/Inspectors/Utilities/MixedRealityInspectorUtility.cs
```diff
@@ -35,11 +35,10 @@ public static bool CheckMixedRealityConfigured(bool showHelpBox = true)
                     {
                         EditorGUILayout.HelpBox("No Mixed Reality Toolkit found in scene.", MessageType.Error);
                     }
-
-                    return false;
                 }
 
-                MixedRealityToolkit.ConfirmInitialized();
```
StephenHodgson: [Same question here](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3890/files#r273248905).

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
```diff
@@ -30,7 +30,12 @@ public static Camera Main
                     switch (cameras.Length)
                     {
                         case 0:
-                            return null;
+                            Debug.LogWarning("No main camera found. The mixed reality toolkit requires at least one camera in the scene. One will be generated now.");
+                            GameObject mainCameraGo = new GameObject("Main Camera");
+                            Camera newMainCamera = mainCameraGo.AddComponent<Camera>();
+                            mainCamera = Refresh(newMainCamera);
```
StephenHodgson: should we go ahead and add the `MainCamera` tag here as well?

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
Railboy: Yes, that happens when you invoke the property. So the null check creates it. If the property returns null something has gone very, very wrong.

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
Railboy: Tagging the camera causes a _SENDMESSAGE CANNOT BE CALLED DURING AWAKE_ warning. I've removed it until I can figure out what Unity is doing under the hood with tag assignment. Pretty weird.

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
Railboy: These two ConfirmInitialized calls - and only these calls - were causing game objects (camera / play space transform) to be created during OnDestroy in play mode, which Unity really doesn't like.

Bailing in the inspector until the toolkit is initialized elsewhere seems to work. The toolkit still gets initialized in the editor without causing life cycle problems. And at runtime it's obviously not an issue.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
chrisfromwork: Conceptually does the MixedRealityToolkit service container require having a MixedRealityPlayspace? Should we actual update components to handle cases where a MixedRealityPlayspace doesn't exist compared to auto generating it?

Railboy: MRTK assumes the existence of a camera, and the camera requires the playspace to function correctly. I think that's a reasonable assumption as long as we continue to make it explicit by including the camera profile in the main config profile.

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
Railboy: Fixed by passing tag during creation.

StephenHodgson:
`Refresh(Camera.main)` on line 23 in `CameraCache` will still throw a NRE. Check my Fork's PR for proper support.

keveleigh:
>`Refresh(Camera.main)` on line 23 in `CameraCache` will still throw a NRE.

Which line causes the NRE? I'm not sure I see anything in `Refresh` that'd be trying to access something on the `null`.

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
```diff
@@ -20,7 +20,15 @@ public static Camera Main
         {
             get
             {
-                var mainCamera = cachedCamera == null ? Refresh(Camera.main) : cachedCamera;
+                var mainCamera = Camera.main;
+
+                if (mainCamera == null)
```
keveleigh: I'm not sure this most recent change makes sense in the flow of this method. Previously, it was written so it:
1. Checks for a main-tagged camera.
2. If it doesn't find one, search the scene for any non-tagged cameras.
3. If it doesn't find any non-tagged cameras, create one.
4. If it does find a non-tagged camera, use that.
5. If it finds more than one, look for name or use the first one.
6. Return

Now, searching for a non-tagged camera will never be hit (because we always create a new main-tagged camera if one doesn't exist), and we may create a new camera even if one already exists:

1. Checks for a main-tagged camera.
2. If it doesn't find one, create a main-tagged camera.
3. Return

Railboy: I considered this as well but I actually think this is an improvement.

There may be cases where you have camera(s) in your scene(s) that aren't meant to serve as a main camera. In those cases by leaving them untagged you can be confident they won't be usurped by the MRTK.

keveleigh: Ahh, good call, yeah. Should we remove much of the rest of this method then, to help prevent confusion, etc?

Railboy: Yeah we can simplify it. And we can re-add a check for the cached camera since we lose the benefit of avoiding the tag check otherwise.

StephenHodgson:
The only thing I don't like about the most recent changes is that in the default scene that ships with the toolkit, it will create a duplicate second camera

Railboy:
@keveleigh Can we try to get this into RC1 refresh?

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
```diff
@@ -20,57 +20,25 @@ public static Camera Main
...
-                                mainCamera = Refresh(cameras[0]);
-                            }
+                // If the cached camera is null, search for main
+                var mainCamera = Camera.main;
 
-                            break;
-                    }
+                if (mainCamera == null)
+                {   // If no main camera was found, create it now
+                    Debug.LogWarning("No main camera found. The mixed reality toolkit requires at least one camera in the scene. One will be generated now.");
```
keveleigh:
```suggestion
                    Debug.LogWarning("No main camera found. The Mixed Reality Toolkit requires at least one camera in the scene. One will be generated now.");
```

Railboy:
I wasn't able to repro this issue either.

StephenHodgson:
Sorry I meant a the demo scene when starting a new unity project

keveleigh:
>Sorry I meant a the demo scene when starting a new unity project, or whatever Unity calls it.

SampleScene.unity?
Are you referring to the old bug where that scene's camera wasn't properly tagged as `MainCamera`? It was [fixed by Unity in 2018.3](https://issuetracker.unity3d.com/issues/main-camera-object-is-not-tagged-with-main-camera-tag-in-a-default-sample-scene).

image

PR #3901, opened 2019-04-09
Revert "Update pressable button, adds inspector"

danielhofmann-ms:
This reverts commit 2538ff87bd1c431d8fd38456c5f129ed3c605e9c.

Should not have gotten into RC1


julenka:
Should we update the RC1 packages after this goes in, then? @wiwei are you the owner of updating the release, or is it Niall? I'm talking about this page: https://github.com/Microsoft/MixedRealityToolkit-Unity/releases. Should we call the release MRTK 2.0.1 RC1?

Also, since this is going to release, I think we should also merge it into development, and then add back Lars's changes into a feature branch so that he can continue working on the inspector improvements.

julenka:
We are currently discussing whether to revert this or just go forward and fix the bugs, please do not merge.

PR #3916, opened 2019-04-10
Alternative Pinch gesture with a steady wrist.

lukastoenneMS:
Overview
---
Alternative Pinch gesture that keeps the wrist stead and moves the index finger and thumb instead. This was requested by users, but rather than having one or the other i would rather provide both options. The fixed index tip variant is still the default, since it is considered better for testing interactions. The alternative with moving index finger is more anatomically correct though.

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandPose.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandPose.cs
@@ -7,7 +7,6 @@ using System;
 
 namespace Microsoft.MixedReality.Toolkit.Input
 {
-
     public class SimulatedHandPose
     {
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandPose.cs
```diff
@@ -127,6 +126,7 @@ public enum GestureId
             Flat,
             Open,
             Pinch,
+            PinchSteadyWrist,
```
luis-valverde-ms: Document the difference here?

lukastoenneMS: Done.

lukastoenneMS:
\azp run

PR #3919, opened 2019-04-10
Fix GridObjectCollection

Alexees:
Overview
GridObjectCollection was not properly working:

- Horizontal and Vertical layouts were not even implemented
- Rows and Columns were not swapped when switching between which comes first, offsetting the whole layout from the center.


wiwei:
@cre8ivepark, I *think* this is in your area of expertise

radicalad:
Something broke prior to porting the collections from HTK -> MRTK. I found this bug but haven't had time to fix it. Thank you @Alexees !

I'd argue the different layout results are in fact the desired ones.

PR #3920, opened 2019-04-10
Prevent MRTK From repeatedly creating profile-less instance automatically

provencher:
Overview
---
- Previously, if you clicked on any MRTK profile with an open scene lacking an MRTK instance, the MRTK Active profile getter would attempt to request an editor ping to an instance that did not exist, thus creating one in your scene without prompt. 

- This behavior led to the MRTK modifying scenes that shouldn't have been modified.

Changes
---
- Fixes: # .


provencher:
Ok this fix does not fully solve the problem. Im gonna open up a bug report

Issue #3884, opened 2019-04-08
"No active toolkit" warning is shown when loading profiles instead of scenes.


lukastoenneMS:
## Overview
We have a check in the MixedRealityToolkitConfigurationProfileInspector which pops up a "helpful" dialog when no MRTK instance exists in the current scene.
![NoActiveToolkitWarning](https://user-images.githubusercontent.com/47210016/55724586-a750c180-5a03-11e9-92de-73e7b083af41.jpg)

Apart from being, in my personal opinion, a bit annoying, there are several problems with the way this is currently implemented:

1. The check for existing MRTK initialization happens in the inspector for the PROFILE. That means it is performed only when an MRTK profile is actually loaded, e.g. by clicking on it in the asset browser. It does not run when a new scene is created or loaded. We probably want to use the the [SceneManager.sceneLoaded](https://docs.unity3d.com/ScriptReference/SceneManagement.SceneManager-sceneLoaded.html) event instead.
2. The check arbitrarily happens when running playmode tests. The test runner starts out with an empty scene before the tests load the actual test scene. AFAIK there is no good way to disable this feature, other than setting the HideNoActiveToolkitWarningKey in the session state, which is probably not desirable (that's what the "Don't show this again" box is for).

## Expected Behavior
Popup appears when a new scene is loaded or created.

## Actual Behavior
Popup appears the first time a MRTK base profile is loaded.

## Steps to reproduce
- Open example scene.
- Select the MRTK profile to show in the inspector (no warning will appear because the scene has a MRTK instance)
- Create a new scene. No warning either because the profile was already loaded.
- Save the default scene.
- Close and reopen the mrtk project (not just the scene!). Now selecting the profile shows the warning dialog because it gets loaded for the first time.

## Unity Editor Version
2018.3.9f1
## Mixed Reality Toolkit Release Version
MRTK v2 RC1 (13b3fcdfabacc7e0d101b0173c7a667c8934c87a)

Issue #3921, opened 2019-04-10
MRTK instance added automatically to all scenes loaded in editor


provencher:
## Overview
With the MRTK setup in a project, as soon as a scene without an MRTK instance is loaded in the editor, the MRTK creates an instance without an empty profile and forces the user to select a profile.

## Expected Behavior
If no MRTK instance is loaded in a scene, then none should be added without the user's express consent. 

Scenes without MRTK components should not have MRTK components automatically added to them. 

## Actual Behavior
If no MRTK instance exists in the scene, one is automatically added to the scene graph, without warning, with an empty profile, thus raising a popup prompting the user to select a profile.

If no profile is selected when the popup appears, the profile selection menu pops up again as soon as the MRTK instance is relesected, thus making it very difficult to delete the generated MRTK instance, as focus is constantly removed from the user.

If the user accepts any profile, then the scene is further modified, with the addition of many MRTK components, and the camera's hierarchy being changed.

## Steps to reproduce
1. Create empty unity project.
2. Create basic scene with a camera.
3. Import MRTK into project.
4. Open basic scene created before MRTK import
5. Exhibit behavior described above.

## Unity Editor Version
Unity 2018.3.8f1

## Mixed Reality Toolkit Release Version
v2 RC1

lukastoenneMS:
I've just opened a PR for moving the "No MRTK" popup away from profiles and into a dedicated scene loaded/new handler: #3922 

This still keeps the dialog (for the time being), but at least it should only happen on scene switching, and not when showing arbitrary profiles.

I'm curious though: From what i understand the code for creating a default MRTK object should only run when you actually confirm "Yes" in the dialog. Will try to repro with the steps above ...

provencher:
@lukastoenneMS I do not think that PR goes far enough to fix this issue. 

I opened another PR which does resolve the issue completely for me.

The issue is that the scene check does not block rogue profiles or editor scripts for seeking the almighty Mrtk static Instance, which has a side effect of generating a new MRTK game object.


https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3925

lukastoenneMS:
Is there any specific profile you selected to make the mrtk object be created ad-hoc? Do i need to be in play mode? if so, shouldn't the mrtk instance be temporary when in play mode?

provencher:
I believe the only difference from a blank scene my particular example has is that there is non-mrtk UI in it.

[...]

PR #3922, opened 2019-04-10
Move MRTK scene check out of the profile inspector and into a utility class.

lukastoenneMS:
Overview

This check should not be performed when loading profiles, since it only happens when a profile is actually visible in the inspector, which can happen at arbitrary times in the workflow or not at all. Profiles as separate assets should not be responsible for modifying scene state.

---

Changes

The new utility class will be created on project load (`[InitializeOnLoad]` attribute) and check for a MRTK instance when a new scene is created (with default objects) or when a scene is loaded.

This also solves the problem that the popup would come up when running tests, since these first start from an empty scene before a test scene is loaded. The check will now be ignored in that case, using the `NewSceneSetup.EmptyScene` parameter.

---
- Fixes: #3884 .

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
@@ -10,9 +10,6 @@ namespace Microsoft.MixedReality.Toolkit.Editor
     [CustomEditor(typeof(MixedRealityToolkitConfigurationProfile))]
     public class MixedRealityToolkitConfigurationProfileInspector : BaseMixedRealityToolkitConfigurationProfileInspector
     {
-        const string HideNoActiveToolkitWarningKey = "MRTK_HideNoActiveToolkitWarningKey";
-        private static bool HideNoActiveToolkitWarning = true;
-
         private static readonly GUIContent TargetScaleContent = new GUIContent("Target Scale:");
 
         // Experience properties
@@ -60,35 +57,6 @@ namespace Microsoft.MixedReality.Toolkit.Editor
             if (target == null)
             {
                 // Either when we are recompiling, or the inspector window is hidden behind another one, the target can get destroyed (null) and thereby will raise an ArgumentException when accessing serializedObject. For now, just return.
-                return;
-            }
-
-            configurationProfile = target as MixedRealityToolkitConfigurationProfile;
-
-            // Create The MR Manager if none exists.
-            if (!MixedRealityToolkit.IsInitialized)
-            {
-                // Search the scene for one, in case we've just hot reloaded the assembly.
-                var managerSearch = FindObjectsOfType<MixedRealityToolkit>();
-
-                if (managerSearch.Length == 0)
-                {
-                    HideNoActiveToolkitWarning = SessionState.GetBool(HideNoActiveToolkitWarningKey, false);
-                    if (!HideNoActiveToolkitWarning)
-                    {
-                        NoActiveToolkitWarning.OpenWindow(configurationProfile);
-                    }
-                    return; 
-                }
-            }
-
-            if (!MixedRealityToolkit.ConfirmInitialized())
-            {
-                return;
-            }
-
-            if (!MixedRealityToolkit.Instance.HasActiveProfile)
-            {
                 return;
             }
 
@@ -314,53 +282,6 @@ namespace Microsoft.MixedReality.Toolkit.Editor
             if (changed)
             {
                 EditorApplication.delayCall += () => MixedRealityToolkit.Instance.ResetConfiguration(configurationProfile);
-            }
-        }
-
-        private class NoActiveToolkitWarning : EditorWindow
-        {
-            private static NoActiveToolkitWarning activeWindow;
-            private MixedRealityToolkitConfigurationProfile configurationProfile;
-            private bool hideWarning = false;
-
-            public static void OpenWindow(MixedRealityToolkitConfigurationProfile configurationProfile)
-            {
-                // If we already have an active window, bail
-                if (activeWindow != null)
-                    return;
-
-                activeWindow = EditorWindow.GetWindow<NoActiveToolkitWarning>();
-                activeWindow.configurationProfile = configurationProfile;
-                activeWindow.maxSize = new Vector2(400, 80);
-                activeWindow.minSize = new Vector2(400, 80);
-                activeWindow.titleContent = new GUIContent("No Active Toolkit Found");
-
-                activeWindow.Show(true); 
-            }
-
-            private void OnGUI()
-            {
-                EditorGUILayout.HelpBox("There is no active Mixed Reality Toolkit in your scene. Would you like to create one now?", MessageType.Warning);
-
-                hideWarning = EditorGUILayout.Toggle("Don't show this again", hideWarning);
-
-                EditorGUILayout.BeginHorizontal();
-                if (GUILayout.Button("Yes"))
-                {
-                    var playspace = MixedRealityToolkit.Instance.MixedRealityPlayspace;
-                    Debug.Assert(playspace != null);
-                    MixedRealityToolkit.Instance.ActiveProfile = configurationProfile;
-
-                    SessionState.SetBool(HideNoActiveToolkitWarningKey, hideWarning);
-                    Close();
-                }
-
-                if (GUILayout.Button("No"))
-                {
-                    SessionState.SetBool(HideNoActiveToolkitWarningKey, hideWarning);
-                    Close();
-                }
-                EditorGUILayout.EndHorizontal();
             }
         }
     }
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
@@ -48,8 +48,6 @@ namespace Microsoft.MixedReality.Toolkit.Editor
         private static bool showRegisteredServiceProperties = true;
         private SerializedProperty registeredServiceProvidersProfile;
 
-        private MixedRealityToolkitConfigurationProfile configurationProfile;
-
         protected override void OnEnable()
         {
             base.OnEnable();
@@ -91,6 +89,8 @@ namespace Microsoft.MixedReality.Toolkit.Editor
 
         public override void OnInspectorGUI()
         {
+            var configurationProfile = (MixedRealityToolkitConfigurationProfile)target;
+
             serializedObject.Update();
             RenderMixedRealityToolkitLogo();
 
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
```diff
@@ -0,0 +1,128 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#if UNITY_EDITOR
```
lukastoenneMS: Note: It would be nicer to have this utility class in the `Microsoft.MixedReality.Toolkit.Core.Utilities.Editor` assembly, which has an explicit Editor dependency. However, the `Microsoft.MixedReality.Toolkit` assembly in turn has a dependency on `Microsoft.MixedReality.Toolkit.Core.Utilities.Editor`, which would create a cycle. That should not be the case IMO, if possible we should try to remove editor code from the core assembly.

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
```diff
@@ -0,0 +1,128 @@
...
+            [SerializeField]
+            public MixedRealityToolkitConfigurationProfile configurationProfile;
+            private bool hideWarning = false;
+
+            public static void OpenWindow(MixedRealityToolkitConfigurationProfile configurationProfile)
+            {
+                // If we already have an active window, bail
+                if (activeWindow != null)
+                    return;
+
```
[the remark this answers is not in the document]
lukastoenneMS: Moved code, will fix it.

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
```diff
@@ -0,0 +1,128 @@
...
+                hideWarning = EditorGUILayout.Toggle("Don't show this again", hideWarning);
+
+                configurationProfile = (MixedRealityToolkitConfigurationProfile)EditorGUILayout.ObjectField(configurationProfile, typeof(MixedRealityToolkitConfigurationProfile), false);
+                if (configurationProfile == null)
+                {
+                    EditorGUILayout.HelpBox("Select a configuration profile.", MessageType.Info);
+                }
+
+                EditorGUILayout.BeginHorizontal();
+                GUI.enabled = (configurationProfile != null);
```
wiwei: >GUI.enabled = (configurationProfile != null); [](start = 16, length = 45)

Reading the code but without great background knowledge on building unity editor UI, it seems a bit odd that we're first setting the enabled bit here conditionally, and then blowing it away after the if statement by always setting it to true on line 115. This question is more for my own understanding, but is this designed to disable the Yes button conditionally (i.e. it affects anything after it's been changed), but the No button will be left enabled in all cases because GUI.enabled is true after that point?

lukastoenneMS: Correct, this disables the "Yes" button unless a profile has been selected (the `GUI.enabled` state applies to all statements after the change). Could be removed if that seems too restrictive, but since the purpose is to end up with a valid MRTK object i think it's helpful. I've tried selecting the default profile, but we don't actually have any definition for that.

provencher:
Going to give this branch a run and see if it improves some of my issues

davidkline-ms:
I'm also looking at this check as part of my service registrar changes (latest here: https://github.com/davidkline-ms/MixedRealityToolkit-Unity/tree/registrar.4). The intent is to completely remove all hard-coded references to MixedRealityToolkit.Instance (and similar usages).

provencher:
As I mentioned in my issue https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3921

Beyond removing removeing all hard-coded references to MixedRealityToolkit.Instance, there is an issue where the getter permits instantiating an mrtk instance while the app is shutting down. 

If this happens in the editor, something that runs in edit mode can carry the instance out of play mode, which is highly undesirable.

provencher:
Some more feedback - This change seems to have broken the MRTK profile inspector for me.

Railboy:
@provencher Could you provide more details on the profile inspector breaking? I'm not seeing anything on my end.

davidkline-ms:
> there is an issue where the getter permits instantiating an mrtk instance while the app is shutting down.

This is something we DEFINITELY need to address. Thanks!

provencher:
Sorry @Railboy - I don't have any specific details other than the fact that when I had this change locally setup, the profile inspector would through some null pointers, even on example scenes when I had the MRTK instance object selected.

These issues went away when I removed the change.

lukastoenneMS:
> 
> 
> I'm also looking at this check as part of my service registrar changes (latest here: https://github.com/davidkline-ms/MixedRealityToolkit-Unity/tree/registrar.4). The intent is to completely remove all hard-coded references to MixedRealityToolkit.Instance (and similar usages).

Had a look at the branch, looks great. Glad we are on the same page as to how profiles should (or should not) work.

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
```diff
@@ -0,0 +1,130 @@
...
+                if (GUILayout.Button("Yes"))
+                {
+                    var playspace = MixedRealityToolkit.Instance.MixedRealityPlayspace;
+                    Debug.Assert(playspace != null);
+                    MixedRealityToolkit.Instance.ActiveProfile = configurationProfile;
```
lukastoenneMS: @davidkline-ms Do you think this use of `MixedRealityToolkit.Instance` needs to be changed to conform to your own branch?

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
```diff
@@ -10,9 +10,6 @@ namespace Microsoft.MixedReality.Toolkit.Editor
     [CustomEditor(typeof(MixedRealityToolkitConfigurationProfile))]
     public class MixedRealityToolkitConfigurationProfileInspector : BaseMixedRealityToolkitConfigurationProfileInspector
     {
```
Railboy: `configurationProfile = (MixedRealityToolkitConfigurationProfile)target;`

This line needs to return. This is the origin of the null reference exceptions.

provencher: That must be what was breaking it for me

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
```diff
@@ -0,0 +1,130 @@
...
+
+                    SessionState.SetBool(HideNoActiveToolkitWarningKey, hideWarning);
+                    Close();
+                }
+                GUI.enabled = true;
+
+                if (GUILayout.Button("No"))
+                {
+                    SessionState.SetBool(HideNoActiveToolkitWarningKey, hideWarning);
```
provencher: I think this setting should be a settings scriptable object somewhere.

That way the variable is persisted in a config object, rather than in an ephemeral session which is cleared on each editor boot.

Moroever, using EditorPrefs could make it difficult to re-enable if need be.

Railboy: We could use [EditorPrefs](https://docs.unity3d.com/ScriptReference/EditorPrefs.html).

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
davidkline-ms: @lukastoenneMS, i'll handle the updates as I merge into my branch/fork. that way, this PR stays as targeted as possible

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
davidkline-ms: Let's file an issue on that once this is in.

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
davidkline-ms: I'll handle the update when i merge into my working branch. thanks!

davidkline-ms:
@Railboy, do you believe this PR still needs changes before merging?

Thanks!

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
StephenHodgson: There's a specific reason for that dependency, because the Mixed Reality Settings class needs to be compiled and ran before the core xrtk assemblies, or else some editor settings will not be properly initialized or setup as needed.

Now not all the classes in that assembly probably belong there, but things have been done for specific reasons and many people seem to be breaking and the effects are spreading though out much of the code base.

lukastoenneMS: What/where is the Mixed Reality Settings class?

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
lukastoenneMS: Fixed, thanks for spotting this.

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
lukastoenneMS: I've logged this here for future discussion: https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3952

davidkline-ms: The MRTK core assembly should have no dependencies, in a perfect world. Let's identify the specific editor assembly classes and methods and investigate moving them into the core assembly, or at least factoring them into an isolated assembly that is safe for the core to reference (and avoid the cycle that @lukastonneMS mentions).

lukastoenneMS:
> 
> 
> @provencher After a fresh import I'm seeing the same null reference exceptions.
> 
> ```
> NullReferenceException: Object reference not set to an instance of an object
> Microsoft.MixedReality.Toolkit.Editor.MixedRealityToolkitConfigurationProfileInspector.OnInspectorGUI () (at Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs:103)
> ```

This was fixed by adding back the `configurationProfile = (MixedRealityToolkitConfigurationProfile)target;` line

Issue #3903, opened 2019-04-09
.NET build support is broken again.


wiwei:
.NET build support was fixed for a little while and then broken by the latest changes. This issue only highlights the oldest PR that has caused problems (there may be others)

https://github.com/Microsoft/MixedRealityToolkit-Unity/commit/4f8ccb2ea294d79daa573eb50c945dca2854ecea#diff-c329a30d3cdfde0ee135eff217e63646

@danielhofmann-ms , can you take a look at fixing this up?

Here's how to repro:
1) Get the latest mrtk_development
2) Switch to the .NET backend in the project settings -> other settings (or player settings I guess depending on your unity version)
3) Build a UWP app (with C# projects enabled)
4) Notice that there are build errors in the InputSystem project due to references to C# 7.2 in the project like:

Severity	Code	Description	Project	File	Line	Suppression State
Error	CS8107	Feature 'readonly references' is not available in C# 7.0. Please use language version 7.2 or greater.	Microsoft.MixedReality.Toolkit.Services.InputSystem	D:\packages\verifyold\Assets\MixedRealityToolkit.Services\InputSystem\FocusProvider.cs	173	Active

Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0738	'FocusProvider' does not implement interface member 'IMixedRealityFocusProvider.GetPointers<T>()'. 'FocusProvider.GetPointers<T>()' cannot implement 'IMixedRealityFocusProvider.GetPointers<T>()' because it does not have the matching return type of 'IEnumerable<T>'.	Microsoft.MixedReality.Toolkit.Services.InputSystem	D:\packages\verifyold\Assets\MixedRealityToolkit.Services\InputSystem\FocusProvider.cs	19	Active

gilbdev:
i got them with "Feature 'readonly references' is not available in C# 6.0" (and not 7.0)
building for UWP, Target Hololens x86.

Note that even when targeting IL2CPP, we can't have the project compiling under visual, which makes it impossible to attach the debugger to Unity Editor, which is a serious issue.

wiwei:
I think things are still broken, taking this to see what the next issue is

Ecnassianer:
Should we have a .NET build as part of our azure build pipelines? This has got to be at least the third time we've fixed .NET only to have it break within a month.

wiwei:
The short answer is yes. The longer answer is getting resourcing to get it fixed. It's on my list of things to work through.

[...]

PR #3923, opened 2019-04-10
fix .net focusprovider build

danielhofmann-ms:
#3903


wiwei:
Providing a little more background here, the "in" keyword is only available in 7.2, but the projects that we're working on (w/ the .NET scripting backend) is only at 7.0 AFAIK.

PR #3925, opened 2019-04-10
Fixed mrtk missing instance bug creating new instances without consent

provencher:
Overview
---
- Previously the MRTK would load instances into open scenes without consent, even if  the user responded no to the popup asking for one to be created. This caused large scale scene modifications in scenes without any previous MRTK presence.

- This fix prevents new instances from being created if the application is not playing and no existing instances are found.

Resolves issue https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3921

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
@@ -557,6 +557,13 @@ namespace Microsoft.MixedReality.Toolkit
                 switch (objects.Length)
                 {
                     case 0:
+#if UNITY_EDITOR
+                        if (!Application.isPlaying && instance == null)
+                        {
+                            Debug.LogWarning("No MRTK instance found in scene");
+                            return null;
+                        }
+#endif
                         Debug.Assert(!newInstanceBeingInitialized, "We shouldn't be initializing another MixedRealityToolkit unless we errored on the previous.");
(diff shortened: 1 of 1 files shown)


Railboy:
I agree 100% with the goal but consider using MixedRealityToolkit.IsInitialized to prevent calls to MixedRealityToolkit.Instance in the first place, rather than creating different behavior in the editor vs. the player.

Editor scripts and profiles should not assume that an instance will exist in the scene and should be able to fail gracefully if one isn't found. If they're doing so we need to clean them up.

provencher:
@Railboy this issue is happening without any custom scripts of any kind on my end.
I implemented this because something in the MRTK is making those inappropriate checks, which is causing this problem. 

@lukastoenneMS and I were discussing the issue in https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3921

provencher:
Ok so despite this fix, I've entered a state where I cannot press play as I am missing an MRTK instance.

Railboy:
@provencher I hear you, I'm saying MRTK editor scripts and profiles shouldn't be doing this and need to be cleaned up if they are.

provencher:
Right - I think @luis-valverde-ms is in the process of investigating the issue. I can close this PR then.

luis-valverde-ms:
I'm on it!

Issue #3931, opened 2019-04-10
PressableButton inspector displays incorrectly on piano keys


julenka:
This bug regards the gizmos inspector for Pressable buttons.

If you have an off-center origin of some sort for the movable visual (see piano keys), the  inspector gizmos are displayed in the wrong place and the moving visual position may even be modified by the inspector code upon selection (with no other user interaction)

Additionally, the initial marker is being generated at edit time.

PR #3933, opened 2019-04-10
PressableButton Inspector Improvement (Part 2)

Railboy:
Overview
---
- Fixes inspector for buttons with offset content transforms
- Enables accurate preview of button settings during touch / press events
- No longer generates initial transform at edit time
- Show Handles is disabled by default and preference is stored

This removes the ability to drag the button content transform in the inspector as it was causing problems with undo. You can still move the transform manually.

This also removes the InitialTransform objects from pressable button and piano key prefabs, since they will be generated at runtime.

ButtonPreview

- Fixes: #3931


Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -10,15 +10,46 @@ namespace Microsoft.MixedReality.Toolkit.Editor
     [CustomEditor(typeof(PressableButton))]
     public class PressableButtonInspector : UnityEditor.Editor
     {
-        private static bool editingEnabled = true;
+        // Struct use to store state of preview
```
julenka: typo:used.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -10,15 +10,46 @@ namespace Microsoft.MixedReality.Toolkit.Editor
...
+            public float touchCageSize;
+            // The actual values that the button uses
+            public float maxPushDistance;
+            public float pressDistance;
+            public float releaseDistanceDelta;
+        }
+
+        const string EditingEnabledKey = "MRTK_PressableButtonInspector_EditingEnabledKey";
+        const string VisiblePlanesKey = "MRTK_PressableButtonInspector_VisiblePlanesKey";
+        private static bool EditingEnabled = false;
```
julenka: I believe private bools should start with lower case. It would be good to [review the coding guidelines here](https://microsoft.github.io/MixedRealityToolkit-Unity/CodingGuidelines.html#naming-conventions).

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -53,6 +85,10 @@ private void OnEnable()
             boxColliderObject = new SerializedObject(touchCage);
             boxColliderSize = boxColliderObject.FindProperty("m_Size");
             boxColliderCenter = boxColliderObject.FindProperty("m_Center");
+
+            buttonContentTransform = transform;
+            if (movingButtonVisuals.objectReferenceValue != null)
```
julenka: Please put braces after ifs like this here, and below.

```
if (condition)
{
  then stuff
}
```

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -61,120 +97,135 @@ private void OnSceneGUI()
             if (Selection.activeObject != button.gameObject)
                 return;
 
-            // Instruct the button to create its path markers
-            button.FindOrCreatePathMarkers();
+            if (!VisiblePlanes)
```
julenka: Braces.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -10,15 +10,46 @@ namespace Microsoft.MixedReality.Toolkit.Editor
...
+        private static bool EditingEnabled = false;
+        private static bool VisiblePlanes = true;
+
         private const float labelMouseOverDistance = 0.025f;
 
         private static GUIStyle labelStyle;
 
         private PressableButton button;
         private Transform transform;
+        private Transform buttonContentTransform;
```
julenka: What's the difference between transform and buttonContentTransform?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -10,15 +10,46 @@ namespace Microsoft.MixedReality.Toolkit.Editor
...
+        const string VisiblePlanesKey = "MRTK_PressableButtonInspector_VisiblePlanesKey";
+        private static bool EditingEnabled = false;
+        private static bool VisiblePlanes = true;
+
         private const float labelMouseOverDistance = 0.025f;
 
         private static GUIStyle labelStyle;
 
         private PressableButton button;
         private Transform transform;
```
julenka: Doesn't the name 'transform' override the monobehaviour 'transform'?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -61,120 +97,135 @@ private void OnSceneGUI()
...
+            Vector3 touchStartOrigin = info.touchCageLocalBounds.center;
+            touchStartOrigin.z -= info.touchCageLocalBounds.extents.z;
+            info.touchStartOrigin = touchStartOrigin;
+            info.touchCageCenter = info.touchCageLocalBounds.center.z;
+            info.touchCageSize = info.touchCageLocalBounds.size.z;
+            
+            // Get the start and end pos for moving content
+            info.startPos = buttonContentTransform.localPosition.z;
+            info.endPos = buttonContentTransform.localPosition.z;
+            info.endPos += maxPushDistance.floatValue / transform.lossyScale.z;
```
julenka: We should make sure that the info endPos will display correctly even if the scale of the pressablebutton is very small like 0.01. Isn't maxPushDistance given in world units?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -61,120 +97,135 @@ private void OnSceneGUI()
...
                 // Set based on distance from start
                 // Adjust for scaled objects
-                newPressDistance = Vector3.Distance(newPressDistPos, startPos) * transform.lossyScale.z;
+                newInfo.pressDistance = Mathf.Abs(newInfo.pressDistPos - newInfo.startPos) * transform.lossyScale.z;
             }
 
             // RELEASE DISTANCE DELTA
             Handles.color = Color.red;
-            DrawPlaneAndHandle(releasePlane, touchCageLocalBounds.size * 0.3f, handleSize, ref newReleaseDistPos, "Release event");
+            DrawPlaneAndHandle(releasePlane, info.touchCageLocalBounds.size * 0.3f, ref newInfo.releaseDistPos, info.touchStartOrigin, "Release event", editingEnabled);
```
julenka: what is this 0.3f representing? Could we maybe take this magic number and have it be a more descriptive variable?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -183,60 +234,71 @@ private void OnSceneGUI()
             Handles.DrawDottedLine(targetStartPlane[1], targetEndPlane[1], 2.5f);
             Handles.DrawDottedLine(targetStartPlane[2], targetEndPlane[2], 2.5f);
             Handles.DrawDottedLine(targetStartPlane[3], targetEndPlane[3], 2.5f);
-
-            serializedObject.ApplyModifiedProperties();
         }
 
-        private void DrawPlaneAndHandle(Vector3[] plane, Vector3 planeSize, float handleSize, ref Vector3 position, string label)
+        private void DrawPlaneAndHandle(Vector3[] plane, Vector3 planeSize, ref float zPosition, Vector3 cagePosition, string label, bool editingEnabled)
```
julenka: just curious, why are you modifying zPosition as a ref variable instead of just returning the new zPosition as a return value for the method?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -183,60 +234,71 @@ private void OnSceneGUI()
...
+            handleVisiblePos.z = 0;
 
-            if (Vector3.Distance(mousePosition, handlePos) < labelMouseOverDistance)
+            if (Vector3.Distance(mousePosition, handleVisiblePos) < labelMouseOverDistance)
             {
                 DrawLabel(plane[1], transform.up - transform.right, label, labelStyle);
                 SceneView.RepaintAll();
             }
 
+            float handleSize = HandleUtility.GetHandleSize(plane[0]) * 0.15f;
```
julenka: magic number

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
Railboy: This is following a convention set by other inspectors. No issue with making them lower case, but we should do it in another PR where we change the other inspectors to be consistent.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
Railboy: This is an Editor script so no transform component exists.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
Railboy: It just represents the size of the plane.

Using magic numbers to adjust SceneGUI handles is another case where I'm following conventions set in other inspectors. No problem with ripping out magic numbers but if we want to do so then we should open a separate PR, establish a style guide and homogenize all our existing inspectors.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
Railboy: Good question - I thought it was in world space initially but without this fix the visual press distance didn't seem to match up with the press events on oddly scaled buttons. This is the issue we worked out over Teams. Can @danielhofmann-ms weigh in on this?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -95,14 +97,13 @@ private void OnSceneGUI()
             if (Selection.activeObject != button.gameObject)
```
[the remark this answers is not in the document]
Railboy: Does that attribute work with OnSceneGUI too? Or is there an equivalent DrawSceneGUI attribute? That would be super handy.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
Railboy: Cruft, will remove.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -61,120 +97,135 @@ private void OnSceneGUI()
...
-            DrawPlaneAndHandle(targetEndPlane, touchCageSize * 0.5f, handleSize, ref newEndPos, "Max move distance");
+            DrawPlaneAndHandle(targetEndPlane, info.touchCageLocalBounds.size * 0.5f, ref newInfo.endPos, info.touchStartOrigin, "Max move distance", editingEnabled);
 
             if (editingEnabled)
             {
                 // Clamp the z value to start position
-                newEndPos.z = Mathf.Max(startPos.z, newEndPos.z);
-                newMaxPushDistance = Vector3.Distance(startPos, newEndPos);
+                newInfo.endPos = Mathf.Max(newInfo.startPos, newInfo.endPos);
+                newInfo.maxPushDistance = Mathf.Abs(newInfo.startPos - newInfo.endPos) * transform.lossyScale.z;
```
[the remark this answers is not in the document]
Railboy: How tired was I?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -61,120 +97,135 @@ private void OnSceneGUI()
...
             Handles.color = Color.yellow;
-            DrawPlaneAndHandle(pressDistancePlane, touchCageLocalBounds.size * 0.35f, handleSize, ref newPressDistPos, "Press event");
+            DrawPlaneAndHandle(pressDistancePlane, info.touchCageLocalBounds.size * 0.35f, ref newInfo.pressDistPos, info.touchStartOrigin, "Press event", editingEnabled);
 
             if (editingEnabled)
             {
                 // Clamp the z values to target start / end
-                newPressDistPos.z = Mathf.Max(startPos.z, newPressDistPos.z);
-                newPressDistPos.z = Mathf.Min(newEndPos.z, newPressDistPos.z);
+                newInfo.pressDistPos = Mathf.Max(newInfo.startPos, newInfo.pressDistPos);
```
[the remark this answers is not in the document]
Railboy: Could go either way on this one. Do you want to give it a shot and see if it feels more intuitive?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -48,8 +48,10 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
 
         [Header("Position markers")]
         [Tooltip("Used to mark where button movement begins. If null, it will be automatically generated.")]
+
         [SerializeField]
-        private Transform initialTransform;
+        [Tooltip("Ensures that the button can only be pushed from the front. Touching the button from the back or side is prevented.")]
+        private bool enforceFrontPush = true;
```
julenka: Looks like this field is already declared above.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -48,8 +48,10 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
 
         [Header("Position markers")]
```
julenka: I believe having the Tooltip attribute here will cause a compile error - "Duplicate tooltip attribute"

julenka:
Approving to unblock the checkin, however I am still seeing one issue: The just in time text fields are not showing up for me when I hover over a handle:

image

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
danielhofmann-ms: Indeed `maxPushDistance` is currently in world space. But I thought all the distances in the info are in local space? So this makes sense.

On the other hand I just realized that lossyScale.z is the axis scaling in world space direction z, not local space z (press direction). So this only works with uniform scales. 

The scale along the local z direction can be computed like this:
`scale = transform.TransformVector(new Vector3(0, 0, 1)).magnitude`
Use `TransformVector` here to also apply scaling.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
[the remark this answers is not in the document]
danielhofmann-ms: Ok, wanted to hear your opinion on this. Will check myself if the other way feels more intuitive. Leave it like this for now.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
Railboy: Yeah we had a wonky merge conflict resolution.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
Railboy: I'm going to leave scaling as-is for this PR and incorporate TransformVector into the next round of fixes.

wiwei:
The CI build apparently hit one of the machines that has some flakey issues, the x86 and arm builds themeselves look good.

PR #3934, opened 2019-04-10
Make camera frame helper classes serializable for future networking synchronization

chrisfromwork:
Overview
---
Update camera related items to be Serializable

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit.Extensions/PhotoCapture/CameraIntrinsics.cs
```diff
@@ -51,7 +51,7 @@ public class CameraIntrinsics
         ///     for apps that choose to implement GPU-based distortion compensation instead of
         ///     using UndistortPoint, which uses the CPU to compute the distortion compensation.
         /// </summary>
-        public Matrix4x4 UndistortedProjectionTransform { get; private set; }
+        public Matrix4x4 UndistortedProjectionTransform;
```
provencher: If you remove properties, you should set default values to these, as they will show up in the unity editor, and uninitialized variables trigger warnings

PR #3935, opened 2019-04-10
Release frame after its been used for aruco marker detection

chrisfromwork:
Overview
---
Its better to not expect users to release a frame. They should have to opt in to incrementing the reference count for said frame.

Changes
---
- Fixes: # .


provencher:
How do users opt in to increase the reference count of the frame?

If you remove this check that means the system will clear the frame out even if users are reading.

That said, this change does mean you can listen to the frame changes without locking the frame in memory. I don't like the old behavior, but the new one seems a bit unsafe.

chrisfromwork:
@provencher the reference is added for the duration of all event calls. You should only need to add a reference while the image frame is held.

Previously, it hadn't been clear that all event callbacks needed to release the frame. If there's a desire to rearchitect this reference counting choice, we should open a separate issue. For the time being, this change will prevent the spectator view marker detector from holding all captured frames.

PR #3936, opened 2019-04-10
Add socketer component as an extension to enable TCP networking for shared experiences

chrisfromwork:
Overview
---
Add socketer component as an extension to enable TCP networking for shared experiences

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/Socketer.cs
```diff
@@ -0,0 +1,173 @@
...
+using Windows.Foundation;
+using Windows.Networking;
+using Windows.Storage.Streams;
+using Windows.Networking.Connectivity;
+using System.Threading;
+#endif
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.Socketer
+{
+#if UNITY_EDITOR || UNITY_STANDALONE || UNITY_WSA || UNITY_IOS
```
provencher: Are these the only types supported by the Socket plugin? This seems like the list of platforms currently supported by the Spectator view feature, though this is somewhat independent of that. 

Also, with this number of platforms, I feel a custom define is necessary.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/Socketer.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        /// <summary>
+        /// Sends a byte array to the other side.  Works for all Socketers except UDP Listeners.
+        /// If the other Socketer is not listening, the messages are lost.
+        /// </summary>
+        /// <param name="message">Message contents.</param>
+        public void SendNetworkMessage(byte[] message)
+        {
+            socketer.SendNetworkMessage(message);
```
provencher: Since this is a public method, it can be called even if Socketer is disabled, which means that it's prone to throwing null pointer exceptions

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/TCPConnectionManager.cs
```diff
@@ -0,0 +1,228 @@
...
+
+        /// <summary>
+        /// Call to start acting as a client connected to the provided server and port
+        /// </summary>
+        /// <param name="serverAddress">server to connect to</param>
+        /// <param name="port">port to use for communication</param>
+        public void ConnectTo(string serverAddress, int port)
+        {
+            Debug.LogFormat($"Connecting to {serverAddress}:{port}");
+            client = SocketerClient.CreateSender(SocketerClient.Protocol.TCP, serverAddress, port);
```
provencher: It would be nice to support other protocols than TCP, since the system does support it.

Though I understand Socketer may not have a connection handling system for non-TCP communication.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/TCPConnectionManager.cs
```diff
@@ -0,0 +1,228 @@
...
+                OnConnected?.Invoke(connection);
+            }
+
+            while (oldConnections.TryDequeue(out connection))
+            {
+                OnDisconnected?.Invoke(connection);
+            }
+
+            IncomingMessage resultPack = null;
+            while (inputMessageQueue.TryDequeue(out resultPack))
```
provencher: C# now supports inline declarations for out variables.

provencher: I think that message queueing shouldn't happen on the Unity thread.
A socket listener should happen on another thread, that listener should store incoming data in a threadsafe queue, which the unity thread polls on update. Otherwise incoming data can stall Unity's operations.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/TCPConnectionManager.cs
chrisfromwork: socketer supports UDP as well as tcp. We could make this class more generic to support both networking protocols. I'll take another look at this.

provencher: One thing to be careful with if you're adding UDP, is that if you're not fragmenting messages, most will get lost. You'll also need to consider flow control to avoid overloading the hardware network card. This varies according to the router you're using as well.

chrisfromwork: for now i think we will keep this as the tcp connection manager. Enabling fragmented udp messages is probably beyond the scope of this review.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/Socketer.cs
chrisfromwork: This needs to be removed or this won't compile for android. I'll update this now.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/TCPConnectionManager.cs
chrisfromwork: Enqueues should be occurring off the main/UI thread. The Connected/Disconnected callbacks of the socket client aren't driven by an update call. Right now only dequeues should be conducted on the UI thread.

chrisfromwork:
Hazel Networking may be a better solution for different developers compared to socketer. However, socketer does enable prototyping/a better networking stack for calibration work. The hope is that the extension points used by socketer could also be used by Hazel for other developers in the future. But Hazel's limited contributor activity over the last few months does generate some hesitation in take a hard dependency on the project.

provencher:
In regards to making use of Hazel, it could be possible to take it up as a fork maintained externally.
My take is that the TCP portions work quite well with a very nice interface. The UDP portions are not perfect, but they do work. Some work needs to be done to manage congestion, but this is true of any UDP messaging service. It does none on it's own, but if messages are reasonably sized, it handles reliability quite well.

chrisfromwork:
@provencher maintaining a Microsoft branded fork of Hazel is likely beyond the scope of what we can do right now. I am going to go ahead and check this in now to unblock calibration work out of the spectator view branches. We can open issues as needed to address before merging this feature branch into mrtk_development/master

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/SocketEndpoint.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            get { return State == ConnectionState.Connected; }
+        }
+
+        /// <summary>
+        /// Call to set the socket endpoint state to disconnected
+        /// </summary>
+        public void Disconnect()
+        {
+            State = ConnectionState.Disconnected;
```
davidkline-ms: i dont actually see a diconnection occuring here.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/SocketEndpoint.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            socketerClient.Message -= Socket_Message;
+            this.incomingQueue = null;
+        }
+
+        /// <summary>
+        /// Call to send data to this endpoint
+        /// </summary>
+        /// <param name="data">data to send</param>
+        public void Send(byte[] data)
```
davidkline-ms: should this take a MessageEvent?

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/IncomingMessage.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.Socketer
+{
+    /// <summary>
+    /// Helper class that wraps an incoming message
+    /// </summary>
+    public class IncomingMessage
+    {
+        public IncomingMessage(SocketEndpoint endpoint, byte[] data, int size)
```
davidkline-ms: should this take a MessageEvent or am i missing something?

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/SocketEndpoint.cs
provencher: I'd say the public interface shouldn't have to care for stamping the message sent time. That's an internal responsibility.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/IncomingMessage.cs
provencher: The socket can only receive bytes, and wrapping the data in a message event could be expensive due to frivolous copying

chrisfromwork:
@davidkline-ms and @provencher i'll pick up these suggestions here in another review. I have some additional changes that were needed to get TCPConnectionManager working as a monobehaviour

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/OutgoingMessage.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.Socketer
+{
+    /// <summary>
+    /// Wrapper for an outgoing message's payload
+    /// </summary>
+    public struct OutgoingMessage
+    {
+        public float time;
+        public byte[] data;
```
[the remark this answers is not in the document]
chrisfromwork: it looks like this struct isn't used anywhere. I'm going to remove it.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/SocketEndpoint.cs
chrisfromwork: will add disconnect logic and update the example

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/SocketEndpoint.cs
chrisfromwork: The worth in changing this doesn't necessarily seem worth the time required for testing. I won't address this for now.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/IncomingMessage.cs
chrisfromwork: The worth in changing this doesn't necessarily seem worth the time required for testing. I won't address this for now.

PR #3937, opened 2019-04-10
Add a qr code extension to the MRTK to enable public facing development around spectator view calibration

chrisfromwork:
Overview
---
This review introduces a new MRTK dll for qr code tracking. This dll is not available to the public and this component can't be used by external developers. However, stubbing out this functionality will enable public facing development around camera calibration/spectator view improvements. Long term this logic should be updated to reflect the appropriate distribution story.

Note: This change also introduces qr code based IMarkerDetector and IMarkerVisuals for use in spectator view.

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit.Extensions/MarkerDetection/IVariableSizeMarkerDetector.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public interface IVariableSizeMarkerDetector : IMarkerDetector
+    {
+        /// <summary>
+        /// Tries to obtain the physical size of a detected marker.
+        /// </summary>
+        /// <param name="markerId">id of detected marker</param>
+        /// <param name="size">The physical size (in meters) of the detected marker</param>
+        /// <returns>True if the size was located</returns>
+        bool TryGetMarkerSize(int markerId, out float size);
```
andreiborodin: Why not update IMarkerDetector with this?

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,297 @@
...
+using Windows.Perception.Spatial;
+using Windows.Perception.Spatial.Preview;
+#endif
+
+using QRCodesTrackerPlugin;
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.QRCodesTracker
+{
+    public static class QRCodeEventArgs
+    {
+        public static QRCodeEventArgs<TData> Create<TData>(TData data)
```
andreiborodin: Why is this method in a different class?

andreiborodin: Also, why have a Create method that does simple constructor instantiation?

---
In reply to: [275471004](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3937#discussion_r275471004) [](ancestors = 275471004)

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,297 @@
...
+        {
+            return new QRCodeEventArgs<TData>(data);
+        }
+    }
+
+    [Serializable]
+    public class QRCodeEventArgs<TData> : EventArgs
```
andreiborodin: This is a very generic event arg, may name it as such: GenericEventArgs; or if you want it QR specific, then put a constraint on TData or better yet a concrete type.

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,297 @@
...
+
+    public class QRCodesManager : MonoBehaviour
+    {
+        [Tooltip("Determines if the QR codes scanner should be automatically started.")]
+        public bool AutoStartQRTracking = true;
+
+        public bool IsTrackerRunning { get; private set; }
+        public QRCodesTrackerPlugin.QRTrackerStartResult StartResult { get; private set; }
+
+        public event EventHandler<bool> QRCodesTrackingStateChanged;
```
andreiborodin: >EventHandler<bool> [](start = 21, length = 18)

Or why not simply employ this approach and do EventHandler<QRCodesTrackerPlugin.QRCode>?

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
```diff
@@ -0,0 +1,297 @@
...
+            {
+                foreach (var ite in qrCodesList)
+                {
+                    if (ite.Value.Code == qrCodeData)
+                    {
+                        return ite.Key;
+                    }
+                }
+            }
+            return new Guid();
```
andreiborodin: >new Guid(); [](start = 19, length = 11)

Do you wish to return an all 0 guid in case of failure to lookup? If so, perhaps just use default(Guid)? Better pattern would be to do TryGet

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
```diff
@@ -0,0 +1,257 @@
...
+#else
+            Debug.LogError("Current platform does not support qr code marker detector");
+#endif
+        }
+
+        /// <inheritdoc />
+        public bool TryGetMarkerSize(int markerId, out float size)
+        {
+            bool found = false;
+            size = 0.0f;
```
andreiborodin: >size = 0.0f; [](start = 12, length = 12)

nit: I tend to not set the value upfront, and only set as close to the return statement or when I know for sure I have the value I intend to set. This way, I get a compile error if I by accident forget to set the value correctly before returning

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
```diff
@@ -0,0 +1,257 @@
...
+#if WINDOWS_UWP && QRCODESTRACKER_BINARY_AVAILABLE
+            _tracking = false;
+            _processMarkers = false;
+#else
+            Debug.LogError("Current platform does not support qr code marker detector");
+#endif
+        }
+
+        /// <inheritdoc />
+        public bool TryGetMarkerSize(int markerId, out float size)
```
andreiborodin: Is this a texture representation of a Guid? Can we parse it and do and do a dictionary lookup?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
```diff
@@ -0,0 +1,257 @@
...
+
+        private void ProcessMarkerUpdates()
+        {
+            lock (_cachedMarkers)
+            {
+                foreach (var markerPair in _cachedMarkers)
+                {
+                    if (!_markerCoordinateSystems.ContainsKey(markerPair.Key))
+                    {
+                        var coordinateSystem = SpatialGraphInteropPreview.CreateCoordinateSystemForNode(markerPair.Key);
```
andreiborodin: Is there an expectation of acceptable failure of creation? I would imagine it should throw an exception if it fails, and otherwise not return null

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
```diff
@@ -0,0 +1,257 @@
...
+        {
+            lock(_cachedMarkers)
+            {
+                if (_cachedMarkers.ContainsKey(e.Data.Id))
+                {
+                    _cachedMarkers.Remove(e.Data.Id);
+                    _processMarkers = true;
+                    if (_markerCoordinateSystems.ContainsKey(e.Data.Id))
+                    {
+                        _markerCoordinateSystems.Remove(e.Data.Id);
```
andreiborodin: You don't need to guard this call with a ContainsKey, Remove should return true/false upon succesful deletion

Assets/MixedRealityToolkit.Extensions/MarkerDetection/IVariableSizeMarkerDetector.cs
chrisfromwork: Hm, so I think there are two types of marker detectors that will have different usage:

1) Marker detectors that need to know the exact physical size of the marker
2) Marker detectors that can determine the exact physical size of the marker

I think we should have clear abstractions for the two types of detectors. I don't think that overloading the same interface with this functionality makes usage very clear. However, it may also be worth moving the setsize function in IMarkerDetector into an IStaticSizeMarkerDetector.cs interface

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
```diff
@@ -0,0 +1,257 @@
...
+            }
+        }
+
+        private void QRCodeRemoved(object sender, QRCodeEventArgs<QRCodesTrackerPlugin.QRCode> e)
+        {
+            lock(_cachedMarkers)
+            {
+                if (_cachedMarkers.ContainsKey(e.Data.Id))
+                {
+                    _cachedMarkers.Remove(e.Data.Id);
```
andreiborodin: Same here, you should be able to do if (_cachedMarkers.Remove(e.Data.Id))

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
```diff
@@ -0,0 +1,257 @@
...
+            foreach (var coordinatePair in _markerCoordinateSystems)
+            {
+                int markerId = -1;
+                lock (_cachedMarkers)
+                {
+                    if (_cachedMarkers.ContainsKey(coordinatePair.Key))
+                    {
+                        markerId = _cachedMarkers[coordinatePair.Key].MarkerId;
+                    }
+                }
```
andreiborodin: Couple of points here:
- If you are locking, you should lock around both dictionaries/managers if they are meant to be in sync
- I would use TryGetValue, and it it returns false do the Debug.Log, set locatedAllMarkers to false and then "continue;"

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
```diff
@@ -0,0 +1,257 @@
...
+        private bool TryGetMarkerId(string qrCode, out int markerId)
+        {
+            markerId = -1;
+            if (qrCode != null &&
+                qrCode.Trim().StartsWith(_qrCodeNamePrefix))
+            {
+                var qrCodeId = qrCode.Trim().Replace(_qrCodeNamePrefix, "");
+                try
+                {
+                    markerId = Convert.ToInt32(qrCodeId);
```
andreiborodin: >Convert.ToInt32(qrCodeId); [](start = 31, length = 26)

You could do Int32.TryParse

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
chrisfromwork: it currently returns null. This fails fairly frequently and doesn't throw in its current form. 

---
In reply to: [275477203](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3937#discussion_r275477203) [](ancestors = 275477203)

Assets/MixedRealityToolkit.Extensions/MarkerDetection/IVariableSizeMarkerDetector.cs
chrisfromwork: moved to a single interface for usability

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
chrisfromwork: Will address qrcodesmanager changes in a separate review. This code is currently being worked on in a different repo.

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
chrisfromwork: I will address qrcodesmanager changes in a separate review. This code is currently being worked on in a different repo.

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
chrisfromwork: I will address qrcodesmanager changes in a separate review. This code is currently being worked on in a different repo.

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
chrisfromwork: I will address qrcodesmanager changes in a separate review. This code is currently being worked on in a different repo.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
chrisfromwork: We should address this in the future. Changes will still be needed in this detector class when migrating to the final packaging story of the qrcodesmanager logic.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
chrisfromwork: I'm not sure if this comment is at the correct location. Currently this just enables a marker detector consumer to provide a marker id and get back a size. The marker id is currently an integer. It doesn't have a determinable mapping to a guid.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/QRCode/QRCodeMarkerDetector.cs
chrisfromwork: moved to a single content lock. updated code cleanliness to reflect less locking

PR #3938, opened 2019-04-10
Add qr code visual and detector to mrtk project

chrisfromwork:
Overview
---
This change adds a qr code detector to the mrtk project. QR code detection will be preferred over aruco markers with HoloLens 2. This logic isn't currently usable by external developers based on missing plugin dlls. Once a distribution story for these dlls is defined, the detector can be enabled.

Changes
---
- Fixes: # .


chrisfromwork:
targetted wrong branch

PR #3942, opened 2019-04-11
Add option to enable / disable pressableButton back press detection

danielhofmann-ms:
Overview
---
Add option to enable / disable back-press detection.

Handle back-press detection differently by only using a check in OnTouchStarted: Test whether the controller was pushing from the front by taking into account its velocity.

### DIFF
--- a/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Prefabs/PianoKey.prefab
+++ b/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Prefabs/PianoKey.prefab
@@ -67,6 +67,7 @@ MonoBehaviour:
   pressDistance: 0.008
   releaseDistanceDelta: 0.001
   returnRate: 10
+  enforceForwardPush: 0
   TouchBegin:
     m_PersistentCalls:
       m_Calls:
@@ -160,7 +161,7 @@ MonoBehaviour:
   eventsToReceive: 0
   touchableSurface: 0
   bounds: {x: 0.004711149, y: 0.024825739}
-  collider: {fileID: 6113324524881885153}
+  touchableCollider: {fileID: 6113324524881885153}
 --- !u!82 &6132438213147574949
 AudioSource:
   m_ObjectHideFlags: 0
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -42,6 +42,10 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
         [Tooltip("Speed of the object movement on release.")]
         private float returnRate = 25.0f;
 
+        [SerializeField]
+        [Tooltip("Ensures that the button can only be pushed from the front. Touching the button from the back or side is prevented.")]
+        private bool enforceForwardPush = true;
```
danielhofmann-ms: I'm curious to hear if anybody has an idea for a more descriptive name here. What do you think?

julenka:
It would be nice to update the handinteractionexamples scene so that the piano keys can be pressed from the side, so that you can run finger across and have them all play.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
Railboy: Seems 100% clear to me, especially with the tooltip.

danielhofmann-ms:
> 
> 
> It would be nice to update the handinteractionexamples scene so that the piano keys can be pressed from the side, so that you can run finger across and have them all play.

The pianokey.prefab already has the new property set to false

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Prefabs/PianoKey.prefab
```diff
@@ -67,6 +67,7 @@ MonoBehaviour:
   pressDistance: 0.008
   releaseDistanceDelta: 0.001
   returnRate: 10
+  enforceForwardPush: 0
```
danielhofmann-ms: allow piano keys to be pushed by entering touch from the side

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -235,22 +256,23 @@ private float GetFarthestPushDistanceAlongButtonAxis()
...
             return Mathf.Clamp(farthestDistance, 0.0f, maxPushDistance);
         }
 
-        private void UpdatePressedState(float pushDistance, float previousPushDistance)
+        private void UpdatePressedState(float pushDistance)
         {
             // If we aren't in a press and can't start a simple one.
             if (!isPressing)
             {
                 // Compare to our previous push depth. Use previous push distance to handle back-presses.
```
JonathanPalmerGD: Might want to update the comment if we are no longer using the previousPushDistance

PR #3946, opened 2019-04-11
Removes GC Alloc from various services

Railboy:
Overview
---
MixedRealityToolkit.Update with default services used to allocate 476 B every frame. This reduces allocations to 0 B per frame.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs
@@ -159,6 +159,9 @@ namespace Microsoft.MixedReality.Toolkit.Input
         // Last timestamp when hands were tracked
         private long lastSimulatedTimestampLeft = 0;
         private long lastSimulatedTimestampRight = 0;
+        // Cached delegates for hand joint generation
+        private SimulatedHandData.HandJointDataGenerator generatorLeft;
+        private SimulatedHandData.HandJointDataGenerator generatorRight;
 
         public SimulatedHandDataProvider(MixedRealityInputSimulationProfile _profile)
         {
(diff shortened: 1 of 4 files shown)


Railboy:
I set the InteractionSourceStates array length to 1k - is that total overkill?

SimonDarksideJ:
Time will tell

wiwei:
A couple of thoughts:

1k seems like a lot so I guess it shouldn't be an issue going forward. The other possibility for avoiding a hardcoded value is to take something like:

https://docs.unity3d.com/ScriptReference/XR.WSA.Input.InteractionManager-numSourceStates.html

We could re-allocate the buffer if the numSourceStates is greater than our current buffer size (i.e. allocate a buffer that is 2x numSourceStates, so that we don't keep growing it if the source state count is monotonically increasing).

Do I actually think that we should do what I said? Nah. Only if we're concerned that we could miss data (i.e. that 1k is actually not big enough, since with this change we won't know that we're missing info due to the count being returned always being capped by the current size of the array)

wiwei:
i.e. basically adopt an array/container resizing strategy to ensure that we can scale up/down to the num readings automatically.

keveleigh:
I'm not sure there's a current way to get that array to be larger than 2 (2 hands on HoloLens or 2 controllers on immersive), so 1k should be plenty haha.

Railboy:
Let's make it 100k just to be safe.

davidkline-ms:
> I'm not sure there's a current way to get that array to be larger than 2 (2 hands on HoloLens or 2 controllers on immersive)

Is that a cap being set by MRTK or by Unity / underlying platform? Not sure we can rely on all platforms limiting to 2. 

> We could re-allocate the buffer if the numSourceStates is greater than our current buffer size (i.e. allocate a buffer that is 2x numSourceStates, so that we don't keep growing it if the source state count is monotonically increasing).

Will's idea above sounds like a very good approach. definitely future-proofing for platforms for which MRTK may become supported.

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs
```diff
@@ -185,8 +188,16 @@ public bool UpdateHandData(SimulatedHandData handDataLeft, SimulatedHandData han
             // TODO: DateTime.UtcNow can be quite imprecise, better use Stopwatch.GetTimestamp
             // https://stackoverflow.com/questions/2143140/c-sharp-datetime-now-precision
             long timestamp = DateTime.UtcNow.Ticks;
-            handDataChanged |= handDataLeft.UpdateWithTimestamp(timestamp, HandStateLeft.IsTracked, HandStateLeft.IsPinching, HandStateLeft.FillCurrentFrame);
-            handDataChanged |= handDataRight.UpdateWithTimestamp(timestamp, HandStateRight.IsTracked, HandStateRight.IsPinching, HandStateRight.FillCurrentFrame);
+
+            // Cache the generator delegates so we don't cg alloc every frame
+            if (generatorLeft == null)
+                generatorLeft = HandStateLeft.FillCurrentFrame;
```
davidkline-ms: please add { } around these if blocks. too easy to add an indented line that looks to be in the block but isnt.

thanks!

keveleigh:
>Is that a cap being set by MRTK or by Unity / underlying platform? Not sure we can rely on all platforms limiting to 2.

It's a WMR-specific Unity API. Not saying it'll never be larger than 2! But I'm not so sure it'll ever be 1k.

Assets/MixedRealityToolkit/Providers/UnityInput/UnityTouchDeviceManager.cs
```diff
@@ -41,7 +41,8 @@ public class UnityTouchDeviceManager : BaseInputDeviceManager
         /// <inheritdoc />
         public override void Update()
         {
-            for (var i = 0; i < UInput.touches.Length; i++)
```
wiwei: Was this also causing GC alloc? Usually I would see this when creating enumerators (i.e. using foreach) so this would be somewhat surprising.

Railboy: Yeah it was creating an array each time touches was called.

wiwei: Danggggggggggggggggggggggggggggggggggggggg

Issue #3892, opened 2019-04-08
InputSimulationProfile should support pad controller 2nd joystick by default 


gilbdev:
## Overview
Currently, when launching a demo with a pad controller in the editor, we can't control cameras

## Expected Behavior
by default for the **DefaultMixedRealityInputSimulationProfile** LookHorizontal and LookVertical should be mapped in respectively 4th Axis and 5th Axis (Invert) in the Input project settings.
If the MRTK can change Input Settings by adding those 2 mappings, it would be great.

## Actual Behavior
LookHorizontal and LookVertical are defined in the DefaultMixedRealityInputSimulationProfile but not in the Input Project Settings.

## Workaround (for those that want it to work as expected)
- Edit > Project Settings... > Input
- Add 2 Axes 
  * LookHorizontal mapped to Joystick Axis, 4th axis
  * LookVertical mapped to Invert, Joystick Axis, 5th axis

## Steps to reproduce
Launch a demo scene, try moving the camera view with a controller right joystick
camera view does not change, needs to use mouse right click + move mouse
 (left joystick for moves works fine)

## Unity Editor Version
2018.3.11f1

## Mixed Reality Toolkit Release Version
v2.0.0-RC1

chrisfromwork:
Hello,

@gilbdev is it possible to get more information on this pad controller? Is it a specific physical device?

gilbdev:
Oh, sorry, i thought i had it specified ...
It's a standard official xboxone controller.

StephenHodgson:
The `DefaultMixedRealityInputSimulationProfile` should be updated to properly utilize InputActions instead of hard coded axis strings

lukastoenneMS:
> 
> 
> The `DefaultMixedRealityInputSimulationProfile` should be updated to properly utilize InputActions instead of hard coded axis strings

I've been looking at input actions and it using them for controlling camera and hand simulation in the editor does not seem feasible:

- We would need to add **a lot** of new actions to the input actions profile:
  * Camera rotation: 2
  * Camera movement: 2
  * Toggle hands: 2 (left/right)
  * Control hands: 2
  * Rotate hands: 3
- The input sim service is an editor-only data provider. The actions profile is a core profile that would have to include the actions nevertheless.
- With each action added the risk of breaking changes increases due to the way action IDs are just indices in the profile. This should be improved in any case, but it makes me want to stay away from input actions in the meantime.
[...]

PR #3956, opened 2019-04-12
Add default LookHorizontal/LookVertical axes to control camera rotation with gamepad.

lukastoenneMS:
Overview
---
Add default axes in the project input configuration to control head rotation with gamepads.

Changes
---
- Fixes: #3892 .


lukastoenneMS:
> 
> 
> shouldn't LookVertical have Invert property to true ?
> not sure about the standard, may be a personal preference.

Yes, i also prefer inverted, but it seems arbitrary.

lukastoenneMS:
> 
> 
> We've already setup axis 3 & 4, why not just utilize what's already there and fix the profile to utilize the correct axes?

Yes, i've been discussing this with Luis. The thinking was to add new input axes with a specific meaning instead of the the generic "AXIS_<n>", but i see the point of using the profile as the point of configuration rather than changing the project settings. I'll change this.

That being said, using input actions as you suggested on the bug report is probably a better solution altogether. I'd like to make the input simulation profile consistent too, and use the same method for assigning actions for camera control and hand simulation.

Issue #3899, opened 2019-04-09
Missing ProgressIndicator from HTK?


mtaulty:
## Overview
The original toolkit had a ProgressIndicator which was a useful prefab and it does not seem to have made it into MRTK V2?

## Expected Behavior
I opened up a HTK project, removed the HTK, added the MRTK V2 and hoped to be able to build.

## Actual Behavior
Failed to build because ProgressIndicator seemed to be missing? It wasn't clear whether there's a direct replacement or whether it's just an omission?

## Steps to reproduce
As above.

## Unity Editor Version
2018.3.11f1

## Mixed Reality Toolkit Release Version
2.0.0-RC1

Yoyozilla:
looks like this is done? closing. 
lars please reopen if there are remaining work.

PR #3941, opened 2019-04-11
MixedRealityToolkit script support for scene switching

luis-valverde-ms:
# IMPORTANT

This is just a draft,  these changes require @davidkline-ms changes to profiles and we may want to include these in those.

I'm currently hitting an issue in AsyncCoroutineRunner when switching scenes. Investigating it. This is the callstack:

```
Assertion failed on expression: 'go.IsActive()'
UnityEngine.GameObject:Find()
Microsoft.MixedReality.Toolkit.Utilities.AsyncCoroutineRunner:get_Instance() (at Assets\MixedRealityToolkit\Utilities\Async\Internal\AsyncCoroutineRunner.cs:50)
Microsoft.MixedReality.Toolkit.Utilities.<>c__DisplayClass13_0:<GetAwaiterReturnVoid>b__0() (at Assets\MixedRealityToolkit\Utilities\Async\AwaiterExtensions.cs:127)
Microsoft.MixedReality.Toolkit.Utilities.AwaiterExtensions:RunOnUnityScheduler(Action) (at Assets\MixedRealityToolkit\Utilities\Async\AwaiterExtensions.cs:144)
Microsoft.MixedReality.Toolkit.Utilities.AwaiterExtensions:GetAwaiterReturnVoid(Object) (at Assets\MixedRealityToolkit\Utilities\Async\AwaiterExtensions.cs:127)
Microsoft.MixedReality.Toolkit.Utilities.AwaiterExtensions:GetAwaiter(WaitUntil) (at Assets\MixedRealityToolkit\Utilities\Async\AwaiterExtensions.cs:71)
Microsoft.MixedReality.Toolkit.<Unregister>d__7:MoveNext() (at Assets\MixedRealityToolkit\Services\BaseEventSystem.cs:58)
System.Runtime.CompilerServices.AsyncVoidMethodBuilder:Start(<Unregister>d__7&)
Microsoft.MixedReality.Toolkit.BaseEventSystem:Unregister(GameObject)
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:Unregister(GameObject) (at Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:435)
Microsoft.MixedReality.Toolkit.Input.InputSystemGlobalListener:OnDisable() (at Assets\MixedRealityToolkit.Services\InputSystem\InputSystemGlobalListener.cs:48)
[...]

PR #3961, opened 2019-04-12
Feature: Scene Transition Service

Railboy:
Overview
---
Adds a SceneTransitionService extension to MRTK to address #3899.

This extension simplifies the business of fading out a scene, displaying a progress indicator, loading a scene, then fading back in.

Scene operations are driven by the SceneSystem service. But in theory any Task-based operation could be used to drive a transition.

Testing Goals
---
- Use each of the progress indicator prefabs on device to check for rendering / material / animation bugs.
- Try different fade colors and fade lengths - are there bad combinations?
- Use fades on multiple camera targets - main cameras, UX cameras, custom combinations of both.
- Observe transitions between scenes with large amounts of content for progress indicator smoothness and other problems.
- Report any problems with MRTK services during transitions.

Details
---
For fading out / in, the service uses a class implementing ICameraFader. By default the CameraFaderQuad class is used, which instantiates a quad in front of the camera. Other methods (post FX, OnGUI, etc) could be substituted.

A few progress indicator options are included by default. Any prefab with a script implementing IProgressIndicator can be substituted.
- Progress Bar
- Rotating Object
- Rotating Orbs

LoadingBar
RotatingObject
LoadingOrbs

Independent Transitions
---
You can also make FadeOut / FadeIn / ShowProgressIndicator calls independently of scene transitions. However these calls will not do anything if a scene transition is in progress.

Extension Service Discussion
---
Scripts and assets have been placed in MixedRealityToolkit/Extensions/SceneTransitionService/
This would be a good time to discuss:
- Is this the pattern we want to follow?
- Do extensions require their own assembly definition?
- Where to extension inspectors live?

Side note: this service was created using the extension creation wizard #3748


davidkline-ms:
> Scripts and assets have been placed in MixedRealityToolkit/Extensions/SceneTransitionService/
This would be a good time to discuss:
Is this the pattern we want to follow?
Do extensions require their own assembly definition?
Where to extension inspectors live?

All good questions! The team needs to 100% land on a policy for MRTK core team supported extensions (though this feels more like a foundational component).

Since extensions can take any form the developer wishes (source code, precompiled. monobehavior, mrtk service) there are no strict requirements wrt assembly definitions.

Extensions need to be self-contained. The inspector should reside in the extension package.

provencher:
@Railboy how are you processing the scene transitions themselves?

From my experience, running async scene loading causes many stutters and can raise issues with some scripts awakening trying to call other objects that do not exist yet. 

Synchronous scene loading locks up the renderer and the image simply goes blank.

Railboy:
@provencher We have to iron out all the scene load handling issues before this service will be usable. Eg #3941. That's part of why I opened this as a draft.

Yoyozilla:
Could I request we have docs on how to use this along with the feature :)

julenka:
Interesting....I'm curious why a progress indicator would be part of the scene transition service, and not just a component in the SDK, sort of like how buttons are components in the SDK. If I wanted to add a progress bar, I would expect it to just be a common control that would live in the same location as other common controls, not in the scene transition service.

Is there a reason why you put the progress bars in the scene transition service, as opposed to adding progress bars as a common control in MRTK.SDK?

SimonDarksideJ:
I'd agree with @julenka , Having the option to include the progress indicators as part of the scene load is great and we should support that.
But the actual assets should form part of the StandardAssets, likely the SDK assets.

Jarodshow:
Is there any benefit to this being its own service as opposed to being part of the SceneService?

Railboy:
@Jarodshow 

>Is there any benefit to this being its own service as opposed to being part of the SceneService?

Yes:

- Pure scene management is technical so it's relatively easy to design a universal interface for it. But scene transitions are art / design driven, which makes a one-size-fits-all approach just about impossible.
- Keeping them separate keeps the already large and complex scene service from getting bloated.
- We're in dire need of useful extension service examples. This is a good candidate.

@SimonDarksideJ @julenka

>Is there a reason why you put the progress bars in the scene transition service, as opposed to adding progress bars as a common control in MRTK.SDK?

>But the actual assets should form part of the StandardAssets, likely the SDK assets.

Makes sense, we can move them.

Railboy:
Marking this as ready for review. Needs documentation and examples, but I want some feedback on the usage of lambdas before moving forward with that.

julenka:
Thanks Lars! This seems like a big change. Any chance we can keep it in draft state until we are done with GA bug fixes and tests? Trying to prioritize the "must haves" first.

SimonDarksideJ:
@julenka I see this as a purely addative change with minimal impact to the toolkit. Unless there is a docs shortage, it should go in now. No reason really to hold it off after ga

luis-valverde-ms:
@wiwei , should this be delayed until GA? Should we have a branch for post GA changes so we don't have them accumulating in individual feature branches?

Railboy:
@julenka @SimonDarksideJ I'm fine either way. The only issue I could see with keeping it out is that @cre8ivepark wants the loading indicators for HTK parity. But those could be broken out into a separate PR now that I've moved them into the SDK folders.

Something else - this is our first example of an extension service. So we want to make sure this is packaged and organized in a way that serves as a template. We don't want to set a bad example.

wiwei:
Ah, a bunch of comments here but here's what i have:

1) The loading indicators do seem to reasonably separate from the other parts of the change, and it would be nice to get them in separately. Especially because they are part of the actual HTK parity workstream, it would be good to separate and review a smaller chunk of stuff at a time.
2) Please add tests to the indicators to ensure that at least some basic level of coverage exists on them (i.e. that they can be programatically created, opened, closed, that reset works, etc). 
3) Since this feature is not something that we are going to be pushing for GA, please have this turned off by default, so that it doesn't affect GA stability. Once we snap for stabilization (7/22) it's fine to turn this on by default ("this" being the scene transition service).

I don't think there's any problem in having non-enabled-by-default code checked in assuming we have good confidence that it won't affect the stability of GA (consider that this literally won't run unless it's part of the default profiles). So in turns of checking this in, it's fine to go in assuming it's not in by default.

But yes, summarizing:

1) Please split out the indicator changes in their own (sorry to ask this of you, I know this is annoying, but this change is huge and it's hard to review large changes, and with the addition of tests it's going to grow larger so having smaller things to review is good).
2) Please have it turned off by default since we aren't aiming for GA for this feature

Railboy:
@wiwei That's fine, I can out the indicators.

Railboy:
Cleaned up and ready to go.

Documentation/toc.yml
```diff
@@ -1,169 +1,171 @@
-- name: Getting Started with MRTK
```
wiwei: I think you changed the line-ending style here right -was it that it was inconsistent with the rest of the stuff?

Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.md
```diff
@@ -0,0 +1,187 @@
...
+
+# Using the Extension
+
+You use the transition service by passing Tasks that are run while the camera is faded out.
+
+## Using Scene System Tasks
+
+In most cases you will be using Tasks supplied by the SceneSystem service:
+
+```
```
wiwei: In all of the cases in the docs that use c# code, consider doing:

```csharp (or c#) 

Instead of just ```

You'll get some fancier colors!

File: Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.md [hunk 3]
```diff
@@ -0,0 +1,187 @@
...
+        );
+}
+
+private async Task ResetScene() 
+{
+    // Go through all enemies in the current scene and move them back to starting positions   
+}
+```
+
+Or you may want to load a scene without using the SceneSystem service:
```
wiwei: Do you have any guidance on when you would recommend doing one of these other two locations, or is this something that is obvious to people who have done a lot of work in this space?

Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.md
```diff
@@ -0,0 +1,187 @@
...
+private async void TransitionToScene()
+{
+    ISceneTransitionService transition = MixedRealityToolkit.Instance.GetService<ISceneTransitionService>();
+
+    // Fades out
+    // Loads scene using Unity's scene manager
+    // Fades back in
+    await transition.DoSceneTransition(
+            () => LoadScene("TestScene1")
+        );
```
wiwei: Super nit: Indentation (here and above for things that are in await blocks)

Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+## Using the Progress Indicator
+
+A progress indicator is anything that implements the `IProgressIndicator` interface. This can take the form of a splash screen, a 3D tagalong loading indicator, or anything else that provides feedback about transition progress.
+
+If `UseDefaultProgressIndicator` is checked in the SceneTransitionService profile, a progress indicator will be instantiated when a transition begins. For the duration of the transition this indicator's `Progress` and `Message` properties can be accessed via that service's `SetProgressValue` and `SetProgressMessage` methods.
+
+Alternatively, when calling `DoSceneTransition` you can supply your own progress indicator via the optional `progressIndicator` argument. This will override the default progress indicator.
+
```
wiwei: Nit: It might be worth moving this to AFTER the code example, because it isn't highlighted/listed in the code example (I was looking for it, didn't find it, and then realized that the code example is more talking about the earlier stuff)

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/Scripts/CameraFaderQuad.cs
```diff
@@ -0,0 +1,204 @@
...
+        public CameraFaderState State { get; private set; }
+
+        private Dictionary<Camera, Quad> quads = new Dictionary<Camera, Quad>();
+        private Color fadeOutColor;
+        private Color fadeInColor;
+        private Color currentColor;
+        private Material quadMaterial;
+
+        public async Task FadeOutAsync(float fadeOutTime, Color color, IEnumerable<Camera> targets)
+        {
```
wiwei: Nit: Missing doccomment

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/Scripts/CameraFaderQuad.cs
```diff
@@ -0,0 +1,204 @@
...
+            fadeOutColor = color;
+            fadeInColor = fadeOutColor;
+            fadeInColor.a = 0;
+
+            if (fadeOutColor.a < 1)
+            {
+                Debug.LogWarning("Target color is not fully opaque.");
+            }
+
+            await Task.Yield();
```
wiwei: >await Task.Yield(); [](start = 12, length = 19)

What's the significance of this particular Task.Yield?

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/Scripts/CameraFaderQuad.cs
```diff
@@ -0,0 +1,204 @@
...
+
+                // Set the quad's layer to something the camera sees
+                for (int layer = 0; layer < 32; layer++)
+                {
+                    if (camera.cullingMask == (camera.cullingMask | (1 << layer)))
+                    {
+                        quad.Renderer.gameObject.layer = layer;
+                        break;
+                    }
+                }
```
wiwei: Is it just because of how transient this quad is going to be that it doens't matter which layer gets picked here? It seems arbitrary here (i.e. we could assign it to a layer that is reserved for app specific stuff)

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/Scripts/CameraFaderQuad.cs
```diff
@@ -0,0 +1,204 @@
...
+
+            switch (State)
+            {
+                case CameraFaderState.Opaque:
+                    break;
+
+                default:
+                    Debug.LogWarning("Can't fade in in state " + State + " - not proceeding.");
+                    break;
+            }
```
wiwei: This is a nit, but I've seen it a couple of times and was hoping to adopt a simpler thing - if you only have a single state you're checking for (i.e. ! = Opaque) just use an if statement, it's simpler in that case. Switches are handy if you're trying to show how logic branches off when multiple options come into play.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/Scripts/ICameraFader.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// Basic interface for fading in / out a color on a camera.
+    /// </summary>
+    public interface ICameraFader
+    {
+        CameraFaderState State { get; }
+
+        Task FadeOutAsync(float fadeOutTime, Color color, IEnumerable<Camera> targets);
+        Task FadeInAsync(float fadeInTime);
```
wiwei: Ah, scratch my comment on the implementation itself - I would just add the docs here and then do the ol' inheritdoc thing

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/ISceneTransitionService.cs
```diff
@@ -0,0 +1,133 @@
...
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.SceneTransitions
+{
+    public interface ISceneTransitionService : IMixedRealityExtensionService
+    {
+        /// <summary>
+        /// Called when transition starts.
+        /// This is called at the beginning of a transition, not at the beginning of a scene load.
+        /// For scene load events, use the Scene System.
```
wiwei: >Scene System. [](start = 43, length = 13)

Nit: Consider using an xref/cref to get people in the right direction.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/ISceneTransitionService.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        /// <summary>
+        /// Fades out, enables progress indicator, executes scene op 1, executes scene op 2, disables progress indicator, fades back in
+        /// </summary>
+        /// <param name="sceneOp1"></param>
+        /// <param name="sceneOp2"></param>
+        /// <param name="progressIndicator"></param>
+        /// <returns></returns>
+        Task DoSceneTransition(Func<Task> sceneOp1, Func<Task> sceneOp2, IProgressIndicator progressIndicator = null);
+
```
wiwei: This is just a convenience version of the thing above right? It feels like you could just do DoSceneTransition({operation1, operation2} and it would be fairly straightforward as well).

It's just a larger API surface to maintain going forward and I'm curious if this means the next thing is a version that takes three functions (and where we draw the line - maybe it's just at two)

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/ISceneTransitionService.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// If FadeTargets is set to custom, you will need to provide a custom set of cameras for fading using this function PRIOR to calling DoSceneTransition.
+        /// </summary>
+        /// <param name="customFadeTargetCameras"></param>
+        void SetCustomFadeTargetCameras(IEnumerable<Camera> customFadeTargetCameras);
+
+        /// <summary>
+        /// Fades target cameras out to color. Can be used independently of scene transitions provided no transition is taking place.
+        /// </summary>
+        /// <param name="fadeColor"></param>
+        /// <param name="targetCameras"></param>
```
wiwei: I think these params were associated with a past life. Same on FadeIn down below

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/ISceneTransitionService.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <returns></returns>
+        Task FadeIn();
+
+        /// <summary>
+        /// Instantiates the default progress indicator and returns its main transform. Can be used independently of scene transitions provided no transition is taking place.
+        /// </summary>
+        /// <returns></returns>
```
wiwei: >/// <returns></returns> [](start = 8, length = 23)

Super nit: Be consistent with the usage of <return> - other places we omit it, and it's probably better to clean up all instances where it's empty because it doesn't add anything.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/SceneTransitionService.cs
```diff
@@ -0,0 +1,433 @@
...
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UnityEngine;
+using System.Collections.Generic;
+using System;
+using Microsoft.MixedReality.Toolkit.UI;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.SceneTransitions
+{
+    [MixedRealityExtensionService(SupportedPlatforms.WindowsStandalone|SupportedPlatforms.MacStandalone|SupportedPlatforms.LinuxStandalone|SupportedPlatforms.WindowsUniversal)]
+    [DocLink("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.html")]
```
wiwei: >DocLink [](start = 5, length = 7)

This link goes to a 404 [https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.html](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.html)

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/SceneTransitionServiceProfile.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        [SerializeField]
+        private float fadeInTime = 0.5f;
+
+        [SerializeField]
+        private CameraFaderTargets fadeTargets = CameraFaderTargets.Main;
+
+        [SerializeField]
+        [Implements(typeof(ICameraFader), TypeGrouping.ByNamespaceFlat)]
+        private SystemType cameraFaderType = default(SystemType);
```
wiwei: Nit: Would be great to have tooltips on these and matching doccomments on the public versions of them above.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/SceneTransitionServiceProfile.cs
```diff
@@ -0,0 +1,52 @@
...
+        private bool useFadeColor = true;
+
+        [SerializeField]
+        private Color fadeColor = Color.black;
+
+        [SerializeField]
+        private float fadeOutTime = 1f;
+
+        [SerializeField]
+        private float fadeInTime = 0.5f;
```
wiwei: Nit: Could also add those fancy range attributes to indicate how low/hi they can be

Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.md
```diff
@@ -0,0 +1,187 @@
+# Scene Transition Service
```
wiwei: Consider also having an example set of scenes and "uber" scene that ties all of this together.

As a reader/someone who has zero experience using this, it's still hard for me to understand the whole picture here and how it's supposed to be used. The docs below help tremendously once you have a better understanding of the area (i.e. you know what you don't know).

Also, having some discussion about how this relates to additive scene loading (it doesn't) and the cases in which you might want to do this vs. the other would help to give a little more context.

The asks in this comment aren't something that I think need to be addressed in this PR, but should be in a subsequent on to improve the initial approachability of this feature.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/Scripts/CameraFaderQuad.cs
```diff
@@ -0,0 +1,204 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.SceneTransitions
+{
+    public class CameraFaderQuad : ICameraFader
```
wiwei: >public class CameraFaderQuad : ICameraFader [](start = 4, length = 43)

Nit: Missing doccomment

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/Scripts/CameraFaderQuad.cs
Railboy: Correct. As long as it's seen by the camera it'll work.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/ISceneTransitionService.cs
Railboy: Yes it's just convenience. I found I was writing a lot of boilerplate for simple transitions. With more complex transitions it didn't seem as problematic. I think we can draw the line at 2.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/Scripts/CameraFaderQuad.cs
Railboy: I think it was a remnant from an earlier version. Removed.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/SceneTransitionService.cs
Railboy: That's just because the PR isn't merged. Once it's merged it will work.

Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.md
Railboy: I'm not sure what you'd prefer. This looks weird to me:

```
 await transition.DoSceneTransition
        (
            () => LoadScene("TestScene1")
        );
```
And I don't feel it makes the process clear enough:
```
await transition.DoSceneTransition (() => LoadScene("TestScene1"))
```

Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.md
Railboy: Sorry I'm not sure what bits 'this' and 'the code example' and 'earlier stuff' are referring to, exactly.

Documentation/toc.yml
Railboy: I'm not sure - it was part of a mostly-automated merge to resolve a merge conflict.

Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.md
Railboy: It should be pretty obvious, they'll either be using Unity's scene manager or their own custom wrapper.

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/SceneTransitionService.cs
wiwei: lolololol

Issue #3951, opened 2019-04-12
No formal parameter for scope in Unity 2019.2.0a11 with DOTS


quakeulf:
## Overview
When inporting the latest MRTK-release from here:
https://github.com/Microsoft/MixedRealityToolkit-Unity/releases

...the sole console puts out this:

Assets\MixedRealityToolkit\Inspectors\MixedRealityPreferences.cs(70,32): error CS7036: There is no argument given that corresponds to the required formal parameter 'scopes' of 'SettingsProvider.SettingsProvider(string, SettingsScope, IEnumerable<string>)'

## Expected Behavior
This should just work. No issues at all.

## Actual Behavior
Can't do anything because this is a compiler error.

## Steps to reproduce
0. Install Unity 2019.2.0a11.

1. Import:
https://github.com/Microsoft/MixedRealityToolkit-Unity/releases

2. Import:
https://github.com/Unity-Technologies/EntityComponentSystemSamples

## Unity Editor Version
Unity 2019.2.0a11

## Mixed Reality Toolkit Release Version
MRTK v2 RC1

quakeulf:
This can be fixed by making:
            **var provider = new SettingsProvider("Project/MRTK")**

Into:
            **var provider = new SettingsProvider("Project/MRTK", SettingsScope.Project)**

StephenHodgson:
Actually I think using the `SettingsScope.User` makes more sense for the mrtk settings to show up in preferences window.

quakeulf:
Could this be added to the code in the repository?

wiwei:
This is an action item on us where we're working through the best model of supporting 2019.1. Even though our docs say that we support 2018.3.x+, in reality we've been focused entirely on the current stable version of unity (i.e. 2018.3.x, not 2019.x).

There's also figuring out ARM64 support and other ramifications of 2019 that we're looking at holistically.

In this particular case going with SettingsScope.Project I believe keeps it in the old place it was at due to how defaults worked with 2018:

https://github.com/Unity-Technologies/UnityCsReference/blob/2018.3/Editor/Mono/Settings/SettingsProvider.cs

Going with .User would change it up for current users (i.e. disrupt their flow based on their current experiences).

In general this change seems safe to me.

StephenHodgson:
> Going with .User would change it up for current users (i.e. disrupt their flow based on their current experiences).

Considering when I first wrote it it was in preferences, I thing that it's already been disrupted.

PR #3962, opened 2019-04-12
Update the MixedRealityPreferences to explicitly provide a SettingScope

wiwei:
Interestingly enough this appears to be the only compile-time incompatibility
with Unity 2019.1. 2019.1 support in the MRTK is still highly experiemental
but with this change the UWP should build (and be deployable to HL2)

Also note that this change doesn't affect 2018.3 behavior, see:
https://github.com/Unity-Technologies/UnityCsReference/blob/2018.3/Editor/Mono/Settings/SettingsProvider.cs

vs

https://github.com/Unity-Technologies/UnityCsReference/blob/2019.1/Editor/Mono/Settings/SettingsProvider.cs

The constructor for this object used to default to SettingsScope.Project
2019.1 removed the default (for the second parameter only).

https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3951

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/MixedRealityPreferences.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/MixedRealityPreferences.cs
@@ -67,7 +67,7 @@ namespace Microsoft.MixedReality.Toolkit.Editor
         [SettingsProvider]
         private static SettingsProvider Preferences()
         {
-            var provider = new SettingsProvider("Project/MRTK")
+            var provider = new SettingsProvider("Project/MRTK", SettingsScope.Project)
             {
                 label = "MRTK",
 


Assets/MixedRealityToolkit/Inspectors/MixedRealityPreferences.cs
```diff
@@ -67,7 +67,7 @@ public static bool IgnoreSettingsPrompt
         [SettingsProvider]
         private static SettingsProvider Preferences()
         {
-            var provider = new SettingsProvider("Project/MRTK")
+            var provider = new SettingsProvider("Project/MRTK", SettingsScope.Project)
```
StephenHodgson: The settings should stay in the preferences window, not be moved to the project settings

```suggestion
            var provider = new SettingsProvider("Project/MRTK", SettingsScope.User)
```

wiwei: This doesn't move anything - the description for this PR shows how the old code defaulted to SettingScope.Project. I pointed out the same issue on:

https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3951

StephenHodgson: Yeah I accidentally moved it when I updated the preferences window a few weeks back.

Believe me, it used to be in preferences when I first wrote it, and these settings have nothing to do with the project build settings.

davidkline-ms: Having things like profile locks at the project level feels right. Developers may wish to unlock profiles for only a specific project to avoid accidentally modifying them in others.

StephenHodgson: Then it might be better to actually split up this window as there are both project and non project specific settings.

PR #3963, opened 2019-04-12
Add GetActiveControllers implementation to InputSimulationService

djohnsomsft:
Currently the list of simulated hands can't be queried from the editor. This implements the method used to query controllers by caching the list each time it is changed and exposing it through the existing interface.


julenka:
Out of curiosity, what were you needing this for?

julenka:
@djohnsomsft Could you please resolve the conflicts? Then we can merge it in.

djohnsomsft:
@julenka Conflicts should be resolved now (did not appear to be any conflicts in the same file).

Regarding what this was needed for, we have objects that are enabled/disabled that individually query controller state at times to respond to hand presence/proximity. When using the simulated hands in the editor, these scripts could not be tested because it shows 0 controllers present at all times.

davidkline-ms:
> When using the simulated hands in the editor, these scripts could not be tested because it shows 0 controllers present at all times.

That sounds like a good issue to file to @lukastoenneMS

PR #3964, opened 2019-04-12
Update some of the MRTK-specific .NET scripting backend callouts.

wiwei:
A couple of things here:

BuildDeployWindow.cs - The user already gets .NET backend deprecation warnings in the regular Unity build window. The language of this warning is a bit strong telling people to switch to IL2CPP that may have good reason to stick to .NET (i.e. they actually want a debuggable experience and a faster inner loop). This changes the message to be closer to what Unity says - i.e. it's deprecated, try to use the new thing, but you don't have to. Originally I was going to remove this but then opted to keep it around in case people are only using this build window (in which case they still have some notification that this isn't going to be supported going forward.)

MixedRealityEditorSettings.cs - This was actually a fairly problematic suggestion, because really going to IL2CPP is not actually required in Unity 2018.3.x (whose LTS version we are going to support.) What's worse is pressing "ignore" actually set a UNITY APPLICATION WIDE setting which would disable this prompt in other projects as well, and would also stop you from seeing suggestions to enable XR settings (i.e. if you switched from Standalone -> UWP, you would never see suggestion to enable XR settings). People who are enabling the .NET backend are doing so explicitly (and likely because they want a more debuggable experience) and we shouldn't force them away from that.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
@@ -455,7 +455,7 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
 
             if (curScriptingBackend == ScriptingImplementation.WinRTDotNET)
             {
-                EditorGUILayout.HelpBox(".NET Scripting backend is depreciated, please use IL2CPP.", MessageType.Warning);
+                EditorGUILayout.HelpBox(".NET Scripting backend is deprecated in Unity 2018 and is removed in Unity 2019.", MessageType.Warning);
             }
 
             var newScriptingBackend = (ScriptingImplementation)EditorGUILayout.IntPopup("Scripting Backend", (int)curScriptingBackend, scriptingBackendNames, scriptingBackendEnum, GUILayout.Width(HALF_WIDTH));
--- a/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
@@ -68,13 +68,6 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
                     message += "- Force Text Serialization\n";
                 }
 
-                var il2Cpp = PlayerSettings.GetScriptingBackend(EditorUserBuildSettings.selectedBuildTargetGroup) == ScriptingImplementation.IL2CPP;
-
-                if (!il2Cpp)
-                {
-                    message += "- Change the Scripting Backend to use IL2CPP\n";
-                }
-
                 var visibleMetaFiles = EditorSettings.externalVersionControl.Equals("Visible Meta Files");
 
                 if (!visibleMetaFiles)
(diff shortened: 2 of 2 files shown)


wiwei:
> There's valid and good reasons for this. I think updating the language is flat out misleading and wrong.

Can you elaborate on what the valid and good reasons are? The updated language is factually true.

jbienz:
I just want to weigh in on this because this change came out of an internal discussion and a HoloLens 2 hack that my team participated in last week. I will do my best to explain the thoughts behind this change, hopefully without "breaking myself into jail".

1. .Net backend is still available in Unity 2018. It may be removed in 2019 but I haven't tested it yet. IMO, this should be a warning at best and not a blocker until it becomes one. If 2019 removes it, please only force the change if 2019 is detected.
2. There was an internal disussion (April 3 through 9th) about debugging on Holo 2. Early in that discussion there was a suggestion to use .Net backend and check the "Unity C# Projects" checkbox. It was mentioned as a temporary workaround until we had improved debugging for 2019.
3. Later I asked if this was still the recommendation because every time we opened our project the MRTK configurator was trying to change it to IL2CPP. I wanted to know if MRTK still supported .Net backend or if it had been removed.
imgo
4. I was informed that .Net backend support had not been removed but that there was an [open issue](https://github.com/Microsoft/MixedRealityToolkit-Unity/commit/4f8ccb2ea294d79daa573eb50c945dca2854ecea#diff-c329a30d3cdfde0ee135eff217e63646). That issue was supposed to have been resolved but it appears someone accidentally stepped on those changes causing a regression.

So, in theory .Net backend is still supported for MRTK with 2018. However (and more importantly), I'm still really unclear on the guidance and best practices for debugging MRTK on HoloLens 2. Especially when targeting Il2CPP.

Someone mentioned [this article](https://dotnetbyexample.blogspot.com/2019/03/debugging-c-code-with-unity-il2cpp.html?m=1) as a guide for debugging IL2CPP at runtime, but someone else mentioned that approach was not currently working. I did not understand why it wasn't currently working.

I hope that helps explain this change and helps you guys decide what to do with the PR. As an aside, I'd really like to know the best practices for debugging soon. IL2CPP or otherwise. Especially with the upcoming hacks we have planned soon.

StephenHodgson:
2019.1.0f2 release notes...

>Universal Windows Platform: Removed the .NET scripting backend.

Well would you look at that

keveleigh:
>Universal Windows Platform: Removed the .NET scripting backend.

Right, which is exactly what this PR says as well, and is the guidance we've been giving:

>and is removed in Unity 2019

wiwei:
> 2019.1.0f2 release notes...
> 
> > Universal Windows Platform: Removed the .NET scripting backend.
> 
> Well would you look at that

Thanks for pointing that out! Note that the code in this PR says:

".NET Scripting backend is deprecated in Unity 2018 and is removed in Unity 2019."

I want to call out that feedback that provides real data is extremely valuable - for example, "we shouldn't do .NET stuff here because we think that there's a massive ongoing support for build cost, because it prevents us using new features." We can use that to figure out if the costs are worth it in light of the benefits (i.e. other customers who are dependent on the .NET backend for their development pipeline, for debuggability on HoloLens 2, etc).

I also want to call out that feedback that only includes generalizations without data or snide comments without any of the former are not helpful and that people should refrain from engaging in that behavior.

Issue #1289, opened 2017-11-02
Add logic to know which SDK version we're targeting to better handle backwards comparability 


StephenHodgson:
Right now, Motion Controller visualizer depends on 16299 SDK calls ~~that will prevent us from building for SDK targets less than 16299.~~

HoloLens OS is still ~~15063~~ 14393.

Update
We will update the min SDK target to 16299 and update the scripts in the toolkit with `Windows.Foundation.Metadata.ApiInformation` API contract checks.

Affected scripts:
- [x] `InteractionExtensions.cs`
- [x] Build And Deploy Window

PR #3965, opened 2019-04-13
Build and deploy window fixes

wiwei:
While looking into other build related issues (i.e. IL2CPP vs .NET) I noticed that there were some issues with the build window (with respect to SDK versioning and chosing the right SDK).

There were a couple of issues:
1) There was a concept of a "min" SDK, but the code was actually doing an equality comparison to look for it. That is, if you had GREATER than the min SDK it would fall over.
2) It looks like even if you set a different target SDK than you current (i.e. if you had a higher target SDK than the current min one specified by the MRTK), it would lower your target.

This does change things a bit in that this will now prefer the latest SDK (over always the min SDK). It's a marginal improvement over the old in that now future progress (where new SDK version drops occur) don't cause you to fall over because you didn't detect the "minimum" (because you actually have higher than the minimum).

Also switches to using the Version type instead of doing string comparisons (as this allows us to properly sort and reason over versions instead of strings)

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildDeployPreferences.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildDeployPreferences.cs
@@ -3,13 +3,14 @@
 
 using Microsoft.MixedReality.Toolkit.Utilities.Editor;
 using Microsoft.MixedReality.Toolkit.WindowsDevicePortal;
+using System;
 using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Build.Editor
 {
     public static class UwpBuildDeployPreferences
     {
-        public const string MIN_SDK_VERSION = "10.0.17134.0";
+        public static Version MIN_SDK_VERSION = new Version("10.0.17134.0");
         private const string EDITOR_PREF_BUILD_CONFIG = "BuildDeployWindow_BuildConfig";
         private const string EDITOR_PREF_FORCE_REBUILD = "BuildDeployWindow_ForceRebuild";
         private const string EDITOR_PREF_CONNECT_INFOS = "BuildDeployWindow_DeviceConnections";
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
@@ -162,7 +162,7 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
         private static float timeLastUpdatedBuilds;
 
         private string[] targetIps;
-        private string[] windowsSdkPaths;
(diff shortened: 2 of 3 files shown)


Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildDeployPreferences.cs
```diff
@@ -3,13 +3,14 @@
...
 using Microsoft.MixedReality.Toolkit.WindowsDevicePortal;
+using System;
 using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Build.Editor
 {
     public static class UwpBuildDeployPreferences
     {
-        public const string MIN_SDK_VERSION = "10.0.17134.0";
+        public static Version MIN_SDK_VERSION = new Version("10.0.17134.0");
```
StephenHodgson: Shouldn't this be whatever version of the insider preview you guys are forcing people to use for mrtk?

StephenHodgson:
I can't exactly remember why anymore, but during the whole transition from HTK to support the WMR headsets there was a good reason to enforce the minimum sdk target to a specific version. 

I'm trying to think if this requirement is really still justified/needed.

There's a utility class that checks if a specific uwp api is availible and should be enough to handle that case (so long as people are actually checking before attempting to call them)

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildDeployPreferences.cs
wiwei: Yes, though this intentionally wasn't changed in this PR to keep the change scoped to changing the version-enumeration code.

wiwei:
> I can't exactly remember why anymore, but during the whole transition from HTK to support the WMR headsets there was a good reason to enforce the minimum sdk target to a specific version.
> 
> I'm trying to think if this requirement is really still justified/needed.
> 
> There's a utility class that checks if a specific uwp api is availible and should be enough to handle that case (so long as people are actually checking before attempting to call them).
> 
> I think this API checker expects the minimum sdk version to be the highest it can check against.
> 
> Historical reference: #1289

https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/9ab0e9b69afe45f7856b5731436ae427c3fe7cc4/Assets/MixedRealityToolkit/Utilities/WindowsApiChecker.cs

The API version checker is a bit divorced from the SDK version - it's more designed to check to see if API support when running downlevel (i.e. running on a lower level version, or running on a platform doesn't support those particular APIs - i.e. running on a desktop PC without hand joint information vs running on the same version but with actual hand joint data)

StephenHodgson:
I know what it's supposed to do, I'm the original author lol.

The point was that there used to be a particular reason to set the minimum sdk level in the build window to be the same as the sdk version checker, but I couldn't remember if it was still valid.

Just a hold over from the crazy wmr days?

PR #3975, opened 2019-04-15
Update the Getting Started doc to clarify what the other MRTK profiles are in the initial picker.

wiwei:
We've gotten some feedback that the fact that the user sees 5 different profiles when they load in for the first time is really confusing - no where do we talk about what they do or why they're there. We're hoping to improve the flow here a little bit improving the documentation (i.e. explaining what to pick - the default one - and what the other ones are for - the example scenes) and by choosing a reasonable default (i.e. highlighting the default one when opening the picker)


Ecnassianer:
Thanks so much for this!

PR #3976, opened 2019-04-15
Clean up some data provider profile selection confusion.

wiwei:
In the Input Data Provider section, each data provider has an optional (or required) profile that configures that particular profile. There are a couple of issues:

1) The Add button doesn't actually work on these, because these data providers have abstract (BasedMixedRealityProfile) for their targets. This means pressing the + button ends up just throwing an exception. Unfortunately fixing this up requires some more invasive changes to the way services/data providers are configured, so I'm opting here to just kill off the + button in these cases (and then add documentation for each data provider about what configuration profile it needs).
2) If there already exist profiles that support that particular service (for example, input simulation), this change restricts the set of options in the picker to only assets/profiles that support the service. This prevents the user from seeing every possible profile and having to know which one to pick. Now for things like the input simulation service, we would only show one thing (i.e. the default one) or two (including the custom one if you've cloned it).

What this means is you see:
image

Instead of:
image

In the case that there is only one type of profile that already exists that handle that particular service.

This is really designed to help with the initial use case (i.e. I'm configuring stuff, I'm new to MRTK, why do these things show all of the profiles and what do I choose)

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
@@ -3,7 +3,9 @@
 
 using Microsoft.MixedReality.Toolkit.Utilities;
 using System;
+using System.Collections.Generic;
 using System.Text;
+using System.Linq;
 using UnityEditor;
 using UnityEngine;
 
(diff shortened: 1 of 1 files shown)


Jarodshow:
This looks great! Aren't there some other profile fields that could use this too?

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -69,41 +72,21 @@ private static bool RenderProfileInternal(SerializedProperty property, GUIConten
...
-            else
-            {
-                EditorGUILayout.PropertyField(property, guiContent);
-            }
-
+            RenderProfileField(property, guiContent, GetProfileTypesForService(serviceType));
             if (property.objectReferenceValue == null)
             {
-                if (showAddButton)
+                var profileTypeName = property.type.Replace("PPtr<$", string.Empty).Replace(">", string.Empty);
```
davidkline-ms: This can be likely be simplified to property.type.Name.

just a note, _not_ needed for this PR as this is pre-existing code.

wiwei: Will take a look at this for a future change. The type is a bit... funky.

PR #3977, opened 2019-04-16
Update the default location that profiles are cloned to.

wiwei:
Currently there place where we clone profiles is somewhat inconsistent - if you clone default root MRTK profile it will go into the CustomProfiles by default. If you use the clone button for the other profiles, they will get dumped in the root Assets folder, which isn't optimal. This makes it so that we at least default to a reasonable thing (the CustomProfiles) while providing a way to override that if so desired.

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityProfileCloneWindow.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityProfileCloneWindow.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Collections.Generic;
-using System.IO;
 using System.Reflection;
 using UnityEditor;
 using UnityEngine;
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityProfileCloneWindow.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityProfileCloneWindow.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Collections.Generic;
+using System.IO;
 using System.Reflection;
 using UnityEditor;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityProfileCloneWindow.cs
```diff
@@ -37,8 +37,9 @@ public SubProfileAction(ProfileCloneBehavior behavior, SerializedProperty proper
             public System.Type ProfileType;
         }
 
+        private const string DefaultCustomProfileFolder = "Assets/MixedRealityToolkit.Generated/CustomProfiles";
         private const string IsCustomProfileProperty = "isCustomProfile";
-        private static readonly Vector2 MinWindowSizeBasic = new Vector2(500, 140);
+        private static readonly Vector2 MinWindowSizeBasic = new Vector2(500, 170);
```
davidkline-ms: Why the change from 140 -> 170? 

By the way, these both feel like "magic numbers" that might warrent a named constant for clarity.

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityProfileCloneWindow.cs
```diff
@@ -188,14 +189,15 @@ private void OnGUI()
             GUILayout.FlexibleSpace();
 
             // Get the selected folder in the project window
-            GetSelectedPathOrFallback(ref targetFolder);
-            targetFolder = EditorGUILayout.ObjectField("Target Folder", targetFolder, typeof(Object), false);
+            targetFolder = EditorGUILayout.ObjectField("Target Folder", targetFolder, typeof(DefaultAsset), false);
+            EditorGUILayout.HelpBox("If no folder is provided, the profile will be cloned to the Assets/MixedRealityToolkit.Generated/CustomProfiles folder.", MessageType.Info);
```
davidkline-ms: nice addition!

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityProfileCloneWindow.cs
jbienz: Maybe it's to make room for the new HelpBox on line 193 (shown below)?

wiwei: Yep, it's a height thing because the new info message increases the required height of the window. I thought about commenting this (i.e. 'this determines the size of the window') but then didn't do it because the naming of the window is relatively obvious and missing changing this leads to a very obvious side effect.

PR #3980, opened 2019-04-16
Exposed manipulation handler parameters to be publicly set

provencher:
Overview
---
Previously the only way to modify settings like rotation constraints was by modifying the serialized property fields. This made it difficult to configure handlers when adding them via another script, as in a factory pattern.

Changes
---
- Fixes: # .


provencher:
@keveleigh could you look into why the CI script is failing? It seems stuck on building a unity package, but that has nothing to do with this PR

provencher:
@wiwei @julenka - Can one of you please look at this?

Cheers

PR #3981, opened 2019-04-16
Prevent MRTK instance from being created OnAppQuitting

provencher:
Overview
---
Generating a new instance during app shutdown in the editor allows instances to persist into edit mode, which cause unwanted behavior.

This fix returns a null Instance of the MRTK if none are found in the scene and the app is quitting while running in the editor.

Previously, the editor could attempt to ping the MRTK instance while the app was shutting down, which caused it to generate an instance during shutdown, which it held onto during the transition back to edit mode.

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
@@ -557,6 +557,14 @@ namespace Microsoft.MixedReality.Toolkit
                 switch (objects.Length)
                 {
                     case 0:
+#if UNITY_EDITOR
+                        // Generating a new instance during app shutdown in the editor allows instances to persist into edit mode, which cause unwanted behavior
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -557,6 +557,14 @@ public static MixedRealityToolkit Instance
                 switch (objects.Length)
                 {
                     case 0:
+#if UNITY_EDITOR
+                        // Generating a new instance during app shutdown in the editor allows instances to persist into edit mode, which may cause unwanted behavior
+                        if (isApplicationQuitting)
+                        {
+                            Debug.LogWarning("Trying to find instance during application shutdown.");
```
davidkline-ms: Looks good!

PR #3982, opened 2019-04-16
Fixed Mouse pointer / controller

thalbern:
- mouse cursor not stuck anymore in center of screen
- only one mouse cursor gameobject is created / controller won't be enabled more than once
- screenspace cursor will be hidden when spatial cursor is used
- fixed other controllers interferring with mouse cursor visibility (hands / camera)
- removed spatial mouse cursor (retrieved the mouse cursor position and sent it back to the mouse cursor to set the position of the cursor to that same position again)
- removed mouse wheel scrolling that modified mouse cursor position (not needed as spatial cursor is snapping to interactable objects)
- added ray to world space position calc to enable object manipulation (bounding box)
- added spatial grab to controller mapping to support dragging of objects
- added mouse cursor speed adjustment setting


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
```diff
@@ -75,15 +84,33 @@ public override IMixedRealityController Controller
         /// <inheritdoc />
         public override void OnPreSceneQuery()
         {
+            // screenspace to ray conversion
             transform.position = CameraCache.Main.transform.position;
 
-            Ray ray = new Ray(Position, Rotation * Vector3.forward);
+            Ray ray = new Ray(transform.position, transform.rotation * Vector3.forward);
```
lukastoenneMS: You can use transform.forward here as a shortcut.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
```diff
@@ -75,15 +84,33 @@ public override IMixedRealityController Controller
...
             }
+
+            // ray to worldspace conversion
+            Camera RaycastCamera = MixedRealityToolkit.InputSystem.FocusProvider.UIRaycastCamera;
+            RaycastCamera.transform.position = Rays[0].Origin;
+            RaycastCamera.transform.rotation = Quaternion.LookRotation(Rays[0].Direction);
+
+            // The RayCastCamera is placed so that the current cursor position is in the center of the camera's view space.
+            Vector3 viewportPos = new Vector3(0.5f, 0.5f, 1.0f);
+            cachedWorldSpacePosition = RaycastCamera.ViewportToWorldPoint(viewportPos);
```
luis-valverde-ms: >cachedWorldSpacePosition [](start = 12, length = 24)

The mouse pointer is a controller pose synchronizer, meaning that it should synchronize the transform its game object to its own position.

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
```diff
@@ -28,16 +28,17 @@ public MouseController(TrackingState trackingState, Handedness controllerHandedn
...
-            new MixedRealityInteractionMapping(4, "Right Mouse Button", AxisType.Digital, DeviceInputType.ButtonPress, MixedRealityInputAction.None, KeyCode.Mouse1),
-            new MixedRealityInteractionMapping(5, "Mouse Button 2", AxisType.Digital, DeviceInputType.ButtonPress, MixedRealityInputAction.None, KeyCode.Mouse2),
-            new MixedRealityInteractionMapping(6, "Mouse Button 3", AxisType.Digital, DeviceInputType.ButtonPress, MixedRealityInputAction.None, KeyCode.Mouse3),
-            new MixedRealityInteractionMapping(7, "Mouse Button 4", AxisType.Digital, DeviceInputType.ButtonPress, MixedRealityInputAction.None, KeyCode.Mouse4),
-            new MixedRealityInteractionMapping(8, "Mouse Button 5", AxisType.Digital, DeviceInputType.ButtonPress, MixedRealityInputAction.None, KeyCode.Mouse5),
-            new MixedRealityInteractionMapping(9, "Mouse Button 6", AxisType.Digital, DeviceInputType.ButtonPress, MixedRealityInputAction.None, KeyCode.Mouse6),
+            new MixedRealityInteractionMapping(0, "Mouse Delta Position", AxisType.DualAxis, DeviceInputType.PointerPosition, MixedRealityInputAction.None),
```
luis-valverde-ms: >0 [](start = 47, length = 1)

Changing the ID of interactions will break mouse controller mappings serialized with the old ones.

luis-valverde-ms: Maybe we could instead reuse entry 0 for "Spatial Pointer"

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
```diff
@@ -65,32 +65,14 @@ public void Update()
...
-                controllerPose.Rotation = InputSource.Pointers[0].BaseCursor.Rotation;
-            }
-
-            mouseDelta.x = -UInput.GetAxis("Mouse Y");
-            mouseDelta.y = UInput.GetAxis("Mouse X");
-            MixedRealityToolkit.InputSystem?.RaiseSourcePositionChanged(InputSource, this, mouseDelta);
-            MixedRealityToolkit.InputSystem?.RaiseSourcePoseChanged(InputSource, this, controllerPose);
-            MixedRealityToolkit.InputSystem?.RaiseSourcePositionChanged(InputSource, this, UInput.mouseScrollDelta);
-
             for (int i = 0; i < Interactions.Length; i++)
```
luis-valverde-ms: We don't seem to be raising events for the Mouse Scroll, Spatial Pointer and Spatial Grip inputs.

Assets/MixedRealityToolkit/Providers/UnityInput/MouseDeviceManager.cs
```diff
@@ -47,6 +47,12 @@ public override void Enable()
                 return;
             }
 
+            if (Controller != null)
```
luis-valverde-ms: >Controller != null [](start = 16, length = 18)

What causes Enable() being called with a non-null controller? It seems that if Disable() is being called correctly this shouldn't happen.

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
```diff
@@ -99,16 +81,6 @@ public void Update()
                     }
                 }
 
-                if (Interactions[i].InputType == DeviceInputType.Scroll)
```
StephenHodgson: >removed mouse wheel scrolling that modified mouse cursor position (not needed as spatial cursor is snapping to interactable objects)

Why was this completely removed?

Why not just update the input action that was driving it? Now the scroll cannot be used by anyone looking to configure it.

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
thalbern: yeah, spatial pointer can be reused , i could've removed it altogether bc only the grip is important to make drag work - i'll add the grip to the bottom of the list and leave the old ids

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
thalbern: spatial grip is only there because we make assumptions that only controllers having spatial grip available in the mapping can drag objects. We don't need to actually use that input type to interact with objects. In case of the mouse you have up and down and pointer position events for that. The other two events were already there and i didn't want to remove them from the mapping to not break anything, but both of them were basically not working / doing anything. I'm afraid I'm not familiar enough with the input system or what the guideline is on now used mappings. What would I have to  implement for those? Spatial cursor doesn't make any sense with mouse, scroll neither with our current implementation as the mouse cursor is snapping to interactable objects and one could argue you could move the camera with mouse wheel, but that code shouldn't be in the mouse pointer / controller I suppose?!

Assets/MixedRealityToolkit/Providers/UnityInput/MouseDeviceManager.cs
thalbern: Disable isn't called necessarily between enable calls. Right now when our services and mrtk get set up we end up in a situation where we're enabling our services twice. I'm not sure what the intend behind this is. 

First time calling enable is when we're registering all our services (we're calling initialize and enable in that step) and after that the MRTK main class calls another EnableAllServices in it's OnEnable() method.

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
thalbern: pointing to my comment to luis above who was asking a similar question. The code how it used to be would just modify the x/y position values which was wrong and didn't work. I couldn't figure out what the intent behind this was. So even if you would've modified the controller mapping for this it would have in best case moved the mouse sideways and not perform any actual scroll action. There's no real use case for scroll in the mouse pointer how it's implemented now as we're snapping to all the interactable objects - without this feature it would make sense to move the cursor in depth with the scroll wheel.  Which input event would you recommend to send for the wheel that makes sense?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
thalbern: Can you give me more info on this? I'm not quite sure I understand what I'm supposed to do here?

Issue #3957, opened 2019-04-12
Proposal: Make it easier to obtain the position of the primary active cursor


provencher:
## Overview
Currently, it is rather difficult to obtain the position of the active cursor in the scene. 

The focus handler system provides many pointers, and it is unclear which one is driving the active cursor. Further, there is no obvious means of gaining access to the active cursor in the scene.

Given that only one UIRayCasting camera is active at once in the scene, so it would make sense to be able to at least obtain information about the pointer driving that camera's position and orientation.

## Expected Behavior
MixedRealityToolkit.InputSystem should contain an easy getter for obtaining information about the primary cursor in the scene. If there are two active cursors (one for each hand), it should return the cursor drawn by the dominant hand.

## Actual Behavior
The MixedRealityToolkit.InputSystem can provide a list of many pointers and does not provide any information about the active cursor.
## Unity Editor Version
Unity 2018.3.8f1

## Mixed Reality Toolkit Release Version
v2 RC1

provencher:
@luis-valverde-ms what do you think about this?

luis-valverde-ms:
@provencher , it doesn't sound unreasonable. Can I ask about your particular use cases?

provencher:
@luis-valverde-ms My use case involves interfacing with an existing UI interaction system, without rendering my scenes incompatible with MRTK interactibles.

I have an existing UI system which requires a Pointer origin and direction to draw a ray

That said, getting a general pointer origin and direction from the MRTK has required some clever hacking around.

The origin seems to be consistently represented by `MixedRealityToolkit.InputSystem.FocusProvider.UIRaycastCamera.transform.position;`

The cursor has been obtainable by writing code that checks for active hoverlight, as follows:
![image](https://user-images.githubusercontent.com/7420990/56051494-c65d9500-5d1c-11e9-91b1-c1dcd92197b9.png)

Finally, the direction I obtain by finding the vector drawn from the origin to the active hoverlight, which represents the active cursor.

luis-valverde-ms:
> getting a general pointer origin and direction from the MRTK has required some clever hacking around

Did you try using IMixedRealityPointer.Position/Rotation for that?

> The origin seems to be consistently represented by MixedRealityToolkit.InputSystem.FocusProvider.UIRaycastCamera.transform.position;

I would advise against relying on that as it will contain information for the last pointer to raycast that frame.

> The cursor has been obtainable by writing code that checks for active hoverlight

[...]

PR #3983, opened 2019-04-16
Modified input system to keep track of last pointer down

luis-valverde-ms:
Keep track of the last pointer to raise a pointer down event to help on choosing the primary pointer in two handed setups.

- Fixes: #3957 .

### DIFF
--- a/Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
+++ b/Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
@@ -53,6 +53,11 @@ namespace Microsoft.MixedReality.Toolkit.Input
         IMixedRealityEyeGazeProvider EyeGazeProvider { get; }
 
         /// <summary>
+        /// Last pointer to raise a pointer down event. May be useful for choosing a primary pointer in two handed setups.
+        /// </summary>
+        IMixedRealityPointer LastPointerDown { get; }
+
+        /// <summary>
         /// Indicates if input is currently enabled or not.
         /// </summary>
         bool IsInputEnabled { get; }
--- a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
@@ -60,6 +60,9 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// <inheritdoc />
         public IMixedRealityEyeGazeProvider EyeGazeProvider { get; private set; }
 
+        /// <inheritdoc />
+        public IMixedRealityPointer LastPointerDown { get; private set; }
+
         private readonly Stack<GameObject> modalInputStack = new Stack<GameObject>();
         private readonly Stack<GameObject> fallbackInputStack = new Stack<GameObject>();
 
@@ -868,6 +871,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// <inheritdoc />
         public void RaisePointerDown(IMixedRealityPointer pointer, MixedRealityInputAction inputAction, Handedness handedness = Handedness.None, IMixedRealityInputSource inputSource = null)
         {
+            LastPointerDown = pointer;
             pointer.IsFocusLocked = (pointer.Result?.Details.Object != null);
 
             pointerEventData.Initialize(pointer, inputAction, handedness, inputSource);


provencher:
Also, @luis-valverde-ms I'm seeing many MissingReferenceExceptions trying to query this LastPointerDown.
It seems the object gets destroyed when the hand goes out of view in the Simulated inputs, as the hand pointer gets destroyed but a pointer to it remains. Given that it's now possible to reference it.

It might be best to create an abstraction to the pointer, and only expose things like Pointer position and rotation, without exposing the pointer reference itself. This way, the Input system can check for validity and prevent leaking a reference.

provencher:
`MissingReferenceException: The object of type 'ShellHandRayPointer' has been destroyed but you are still trying to access it.`

I tried adding some checks for null and active, but it seems that missingreferences don't count as null, which means they are let through this check, and then raise errors.

provencher:
Ok the missing references are because you have to cast the interface to a MonoBehavior before doing the null check or it simply doesn't work. Having the interface itself removes the overloaded null check behavior.

Second, I noticed that switching between pointer modes by toggling space bar in the editor destroys the hand, so the pointer being referenced by the hand is a new instance. This means that you have to click every time the pointer changes in order to access it. It would be good to address this.

I think simply retrieving the pointer passed through the OnPointerDown event is insufficient.

davidkline-ms:
@provencher, great catches!

luis-valverde-ms:
@provencher  Excellent, this is the kind of feedback I was hoping for. I don't think I'll be able to spend time on this for the next week or so, so if you feel like giving it a go, be my guest.

jbienzms:
Our team provided feedback very similar to #3957 last week. Knowing which device last did a pointer down might be helpful in some cases, but I don't think it would have solved ours.

We were working on an example scene to show off eye tracking. What we ended up creating was a "See-king" rocket. A voice command fires the rocket and the rocket is supposed to fly toward the eye gaze.

We noticed that on HoloLens 2, if Eye Gaze isn't setup correctly (e.g. the permission hasn't been provided in the package.appxmanifest) then the GazeProvider will default back to head tracking. This worked fairly well for our scenario. However, this doesn't work well when you go cross-platform.

If we wanted to take our little demo to immersive headset, then the most natural "target" would be the motion controller pointer. So, what we intended to do was just have our rocket seek to cursor location of the _current pointer_. This would do the "right thing" cross platform.

The problem I have with OnPointerDown is that (unless I'm missing something) it will never trigger for head tracking or eye gaze. Also, our application never did anything with pointer down so there would never be a reason for user input to raise this event.

We really feel there needs to be some core idea of the "last active pointer" or "primary pointer" and a way to get the cursor position (or hit position) from that pointer. 

I know when there are multiple controllers active there are two pointers / cursors. And I know that will be difficult to provide a concept of "primary". 

Rather than OnPointerDown, maybe the concept of "primary" should be the pointer that last caused something to receive focus? I think that would be a better approach, but there still needs to be an intelligent "default" at startup. That should probably be either eye gaze or head gaze depending on the "prefer eye gaze" setting on GazeProvider.

These are just my thoughts. Thanks for working on this one @luis-valverde-ms.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
```diff
@@ -52,6 +52,11 @@ public interface IMixedRealityInputSystem : IMixedRealityEventSystem
         /// </summary>
         IMixedRealityEyeGazeProvider EyeGazeProvider { get; }
 
+        /// <summary>
+        /// Last pointer to raise a pointer down event. May be useful for choosing a primary pointer in two handed setups.
+        /// </summary>
+        IMixedRealityPointer LastPointerDown { get; }
```
jbienzms: Please consider implementing this as LastFocusPointer instead of triggering on the InputDown event.

luis-valverde-ms: The issues I see with using focus are that:
- You don't have a primary pointer until you actually point at something that can receive focus
- The primary pointer could change while you're actively using it. E.g. you're dragging something with your right hand and accidentally point at a focusable object with your left.
I'm currently prototyping something along the lines of using the first active pointer in the last controller detected or gaze if none.

The difficulties we're having an coming up with a good criteria for primary pointer makes me think that, besides having some sort of default, we need to make sure that that logic can be customized easily.

jbienzms: > The difficulties we're having an coming up with a good criteria for primary pointer makes me think that, besides having some sort of default, we need to make sure that that logic can be customized easily.

I totally agree @luis-valverde-ms. Maybe we should spin off a short design call and invite anyone that wants to join to discuss? I'd be more than happy to participate.

luis-valverde-ms: @jbienzms, good idea. I would be most interested in what _primary pointer_ means for people, the use cases they have in mind for it and how they would like the API to get access to it to look like. What would be the right place to schedule that call?

jbienzms: Great question @luis-valverde-ms . @davidkline-ms and @Yoyozilla, is there a process for spinning up community discussion calls around a feature like this? We can of course use time in the ship room if there's time available, but it might be a good idea to have a process for one-off collaborative design discussion calls around specific subjects where community members might want to participate.

provencher: So on the topic of determining the best pointer, I've been doing some tests, and I found a good criteria.

Basically, just grabbing the UIRayCastCamera's forward vector, and then checking the dot product against forward direction of all the pointers in the FocusProvider. The dot product with the highest value (closest to 1), seems to best represent the pointer I'm looking for.

However, with this in mind there are a few caveats. You need to check if the pointer you're seeking is null every frame - this requires casting to monobehavior in order to get the unity overridden null check.

Moreover, the camera gaze pointer seems to be set to null every single frame when no hand is present in view. This seems to be a bug with handling of the gaze pointer.

luis-valverde-ms: Thanks @provencher , do you mind opening a bug with repro instructions for that issue with the gaze pointer?

It sounds like a method based on the pointer angle with the forward camera direction could be a useful option for some use cases. I'm concerned about two handed manipulation scenarios where you can potentially switch main pointer in the middle of a manipulation, but I think we could/should support the forward method as an option depending on your findings.

Let me know how it goes in your use case and if you find any more caveats.

provencher: I put some code forward for my method
https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3957#issuecomment-488834026

luis-valverde-ms:
Closing this one as it was just a quick test. Let's continue the discussion in #3957 .

PR #3984, opened 2019-04-16
removed external folder and moved root level .md files

thalbern:
moved images from all readMeImages to Documentation/Images
moved all docs from external and root (except main readme) to documentation or subfolders of documentation
patched up links


wiwei:
Change looks sane to me - full transparency here I didn't go to the generated docs. I can take a pass-post checkin.

PR #3985, opened 2019-04-16
Improve TCPConnectionManager monobehaviour

chrisfromwork:
Overview
---
1) Removed structs that weren't used
2) Ensured socket endpoints disconnected correctly
3) Removed the constructor for TCPConnectionManager now that its a monobehaviour
4) Updated the example scene to conduct a disconnect after sending/receiving messages.

Changes
---
- Fixes: # .


Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/TCPConnectionManager.cs
```diff
@@ -12,6 +12,13 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.Socketer
     /// </summary>
     public class TCPConnectionManager : MonoBehaviour
     {
+        /// <summary>
+        /// Timeout interval for the socket endpoints
```
wiwei: What are the units of the timeout (miliseconds, seconds?)

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/TCPConnectionManager.cs
```diff
@@ -25,15 +32,14 @@ public class TCPConnectionManager : MonoBehaviour
         /// </summary>
         public event Action<IncomingMessage> OnReceive;
 
-        private readonly ConcurrentQueue<SocketEndpoint> newConnections = new ConcurrentQueue<SocketEndpoint>();
-        private readonly ConcurrentQueue<SocketEndpoint> oldConnections = new ConcurrentQueue<SocketEndpoint>();
-        private readonly ConcurrentDictionary<int, SocketEndpoint> serverConnections = new ConcurrentDictionary<int, SocketEndpoint>();
+        private ConcurrentQueue<SocketEndpoint> newConnections;
```
provencher: Why not keep these auto initialized and readonly?

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/SocketEndpoint.cs
```diff
@@ -40,7 +40,11 @@ public bool IsConnected
         /// </summary>
         public void Disconnect()
         {
-            State = ConnectionState.Disconnected;
+            if(State != ConnectionState.Disconnected)
```
wiwei:
```suggestion
            if (State != ConnectionState.Disconnected)
```

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/TCPConnectionManager.cs
chrisfromwork: These constructors weren't seeming to work correctly for the concurrent data structures when moving to a monobehaviour. Having now removed the constructor call from the test script, I'll give this another shot.

davidkline-ms: readonly fields can only be initialized in their declarations or in the constructor. by making this a monobehavior (and initializing them in awake) they cannot be readonly fields.

to preserve the previous behavior, consider initializing in the declarations (like previously) and not in Awake.

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/TCPConnectionManager.cs
chrisfromwork: I'm going to remove this inspector behavior for now. We can make it public as needed/when having an inspector element that isn't a TimeSpan is tested.

Issue #3918, opened 2019-04-10
Eye tracking demo does not work due to bug in Microsoft.MixedReality.Toolkit.Examples.Demos.EyeTracking.LoadAdditiveScene


LocalJoost:
## Overview
When you try to click one of the four options in mrtk_eyes_00_RootScene nothing happens 

## Expected Behavior
When you try to click one of the four options in mrtk_eyes_00_RootScene you get one of the four test scenes

## Actual Behavior
Console shows:
ArgumentException: Scene to unload is invalid
UnityEngine.SceneManagement.SceneManager.UnloadSceneAsync (System.String sceneName)
 
## Steps to reproduce
- Double click mrtk_eyes_00_RootScene to open
- Start Play mode
- Click any of the four options

## Unity Editor Version
2018.3.11.f1

## Mixed Reality Toolkit Release Version
2.0 RC1+, pulled from MRTK_development of April 10

## Analysis 
The code in method LoadNewScene of Microsoft.MixedReality.Toolkit.Examples.Demos.EyeTracking.LoadAdditiveScene contains a bug. It   unloads an addivite scene whether one is loaded or not.

## Solution
More statements need to be moved into the if statement and the check for empty scene name can go. Correct code should be:
```C#
private IEnumerator LoadNewScene(string sceneName)
{
    AudioFeedbackPlayer.Instance.PlaySound(audio_OnSelect);

    // Let's find out the name of the currently loaded additive scene to unload
    if (SceneManager.sceneCount > 1))
    {
        lastSceneLoaded = SceneManager.GetSceneAt(1).name;
        Debug.Log($"Last scene name: {lastSceneLoaded}");

        // Let's wait in case we don't want to switch scenes too abruptly 
        yield return new WaitForSeconds(waitTimeInSecBeforeLoading);
        SceneManager.UnloadSceneAsync(lastSceneLoaded);
    }

    Debug.Log($"New scene name: {SceneToBeLoaded}");
    lastSceneLoaded = SceneToBeLoaded;
    SceneManager.LoadSceneAsync(sceneName, LoadSceneMode.Additive);
}
```

wiwei:
This *might* have been addressed by @sostel's change (which was checked in one day after this was filed):

https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3870

sostel:
[...]

PR #3986, opened 2019-04-16
Sostel patch 3918

sostel:
Overview
---
- Fixing some TextMeshPro items that were only rendered in one eye.
- Addressing issue #3918: Fixing loading additive scene.

Changes
---
- Fixes: #3918 .


ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2018.3.9f1
```
wiwei: I would probably leave this one as-is (or at least not downgrade the editor version).

I need to chat with the others to see how this managed otherwise.

ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
...
+m_EditorVersion: 2018.3.8f1
```
wiwei:
```suggestion
m_EditorVersion: 2018.3.9f1
```

wiwei: I think it might be best to leave this as-is (or at least not turn back the editor version clock)

sostel: Yeah sorry, that snug in :/

PR #3991, opened 2019-04-16
Updated with a new way of loading resources from project folders

andreiborodin:
This work is needed when MRTK is split into separate NuGet packages and may contain assets split across those package folders.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
@@ -24,8 +24,6 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
         private const string IgnoreKey = "_MixedRealityToolkit_Editor_IgnoreSettingsPrompts";
         private const string SessionKey = "_MixedRealityToolkit_Editor_ShownSettingsPrompts";
 
-        private static string mixedRealityToolkit_RelativeFolderPath = string.Empty;
-
         [Obsolete("Use the 'MixedRealityToolkitFiles' APIs.")]
         public static string MixedRealityToolkit_AbsoluteFolderPath
         {
@@ -53,7 +51,7 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
         [Obsolete("Use the 'MixedRealityToolkitFiles' APIs.")]
         public static string MixedRealityToolkit_RelativeFolderPath
         {
-            get { return MakePathRelativeToProject(MixedRealityToolkit_AbsoluteFolderPath); }
+            get { return MixedRealityToolkitFiles.GetAssetDatabasePath(MixedRealityToolkit_AbsoluteFolderPath); }
         }
 
         static MixedRealityEditorSettings()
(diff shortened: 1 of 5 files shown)


Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -0,0 +1,131 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
+{
+    /// <summary>
+    /// API for working with MixedRealityToolkit folders contained in the project.
+    /// </summary>
+    [InitializeOnLoad]
+    public static class MixedRealityToolkitFiles
+    {
+        private class AssetPostprocessor : UnityEditor.AssetPostprocessor
```
wiwei: >private class AssetPostprocessor : UnityEditor.AssetPostprocessor [](start = 8, length = 65)

Nit: It might be helpful to have a doccomment like:

/// <summary>
/// A helper class that ensures that the list of mrtkFolders is updated as assets are imported,
/// deleted, or moved.
/// </summary>

andreiborodin: I was debating, it's private cause it only serves the purpose of being an AssetImporter; will add this comment.

wiwei: Yep, could also not be a doccomment (i.e. just a regular comment). Whether or not it shows up in docfx generated stuff is of questionable value (because it's private as you say)

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
```diff
@@ -22,27 +24,34 @@ public MixedRealityEditorSettings()
...
             {
-                if (string.IsNullOrEmpty(mixedRealityToolkit_RelativeFolderPath))
+                if (MixedRealityToolkitFiles.AreFoldersAvailable)
                 {
-                    if (!FindDirectory(Application.dataPath, "MixedRealityToolkit", out mixedRealityToolkit_RelativeFolderPath))
+#if UNITY_EDITOR
+                    if (MixedRealityToolkitFiles.MRTKDirectories.Count() > 1)
                     {
-                        Debug.LogError("Unable to find the Mixed Reality Toolkit's directory!");
+                        Debug.LogError($"A deprecated API '{nameof(MixedRealityEditorSettings)}.{nameof(MixedRealityToolkit_AbsoluteFolderPath)}' " +
```
davidkline-ms: Is this to inform of deprecated method use in the core toolkit or is this something that app / extension code may hit?

andreiborodin: app/extension code may potentially hit this, although I don't know how many would or do rely on MixedRealityToolkt_AbsoluteFoldersPath currently.

---
In reply to: [276630039](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3991#discussion_r276630039) [](ancestors = 276630039)

Issue #3225, opened 2018-12-05
EditorUtility.FocusProjectWindow() throws an Exception


Jarodshow:
**Does this affect the legacy HoloToolkit (master) or the Mixed Reality Toolkit (mrtk_release)?**
Mixed Reality Toolkit

**Describe the bug**
Whenever any of add profile buttons are selected, it throws an error along with creating the new profile.

**To Reproduce**
1. Create new scene
2. Go to the DefaultMixedRealityToolkitConfigurationProfile asset and select "Create new profiles"
3. Select any of the add profile buttons to see the error.
[Stack Trace](https://gist.github.com/Jarodshow/8bc315f378b453146ad0f528ec3cb82b)

**Expected behavior**
No error being thrown and a new profile being created.

**Actual behavior**
A new profile being created but an error being thrown.

**Unity Editor Version**
2018.2.16f1

**Mixed Reality Toolkit Release Version**
latest mrtk_development

**Additional context**
The underlying problem is that a TargetInvocationException is being thrown on FocusProjectWindow(), if this line is commented out, the exception goes away.

![scriptableobjecterror](https://user-images.githubusercontent.com/9560490/49486453-22129180-f804-11e8-80a4-a475f74bcc45.png)

[Stack Trace](https://gist.github.com/Jarodshow/38366a1df8ff1ddc9663121e72b73e4a)

StephenHodgson:
Thanks for the report.

PR #3995, opened 2019-04-16
Fix a couple of inspector related bugs and exceptions

wiwei:
Fixes two things, which kinda are fixed together because one of them blocks the other. The main thing this addresses is the cloning of profile's failure to clone subprofiles correctly (i.e. if you do:

Click 'Copy & Customize' in Mixed Reality Toolkit component
Click 'Clone' by Input System Profile
Select 'Clone Existing' in Pointer Profile and click 'Clone'
Navigate to newly created pointer profile and observe how the Pointer Options array is empty)

1) Removed line in ScriptableObjectExtensions was causing exceptions in the case where you cloned a profile. This actually was causing the sub-profile cloning code to hit an exception and then never actually copy things over. This also manifests itself when you just hit "copy and customize" for the root profile. AFAICT I can see that this has existed for ~5 months now since the report here: #3225.
2) For MixedRealityProfileCloneWindow, it was just passing the wrong variable (i.e. it was passing the empty thing instead of the non-empty thing).


wiwei:
The CI is red but it's because it ran on the machine that's currently misconfigured - x86 and arm64 built fine:

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=1074

PR #3996, opened 2019-04-16
Remove references to dead field showCursorWithEyeGaze.

wiwei:
This was introduced as part of our large payload but was actually never hooked up to anything - it shows up as an option in the inspector but is misleading (i.e. you think it should show a cursor with eye gaze, but it doesn't)

### DIFF
--- a/Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Profiles/EyeTrackingDemoPointerProfile.asset
+++ b/Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Profiles/EyeTrackingDemoPointerProfile.asset
@@ -24,7 +24,6 @@ MonoBehaviour:
     type: 3}
   gazeProviderType:
     reference: Microsoft.MixedReality.Toolkit.Input.GazeProvider, Microsoft.MixedReality.Toolkit.Services.InputSystem
-  showCursorWithEyeGaze: 0
   pointerOptions:
   - controllerType: 256
     handedness: 7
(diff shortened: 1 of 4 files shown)


wiwei:
I should call out before I press a button that this is technically a breaking change, in that a property on an object is being removed. However, this property never worked (or I think more accurately was deferred work that hasn't been done yet), so it is exceedling unlikely that a dependency could have rationally been taken here.

davidkline-ms:
This change feels safe.

jbienzms:
Just want to write that I support this change as it stands (in fact I reported the issue). However, I'd love to see this work completed rather than just have the property removed. I see that @davidkline-ms added it to the To Do list and the RC2 milestone. Does that mean that the actual implementation of the property will be worked on in that timeframe or does that mean that the property will just be removed in that timeframe?

I don't know where we should discuss the "proper" remediation / implementation of this feature, but I'd like to mention that what I'd like to see is the ability to specify cursor by controller type. This way Gaze, Controller and Eye could each have distinct cursors. Then, hiding the cursor for eye simply becomes setting the "eye cursor" to null.

Having said that, the implementation would need to handle changes to this configuration at runtime. For example, there could be a specific scenario where visualizing of eye tracking is valuable but in the rest of the application it's not needed. Or, there could be scenarios where the controller cursor needs to change over time based on which virtual tool the user is holding.

davidkline-ms:
> I see that @davidkline-ms added it to the To Do list and the RC2 milestone. Does that mean that the actual implementation of the property will be worked on in that timeframe or does that mean that the property will just be removed in that timeframe?

It was added to the milestone for cleanup. I defer to @keveleigh and @sostel to determine if the implementation was planned or reworked (and cleanup forgotten).

sostel:
Based on a discussion with @keveleigh and @wiwei, the plan was to remove the checkbox for now. I do agree with @jbienzms though. 
1. An eye cursor is required in some instances. Currently you would need to drop in a prefab to the scene to support that (it's part of the ET demo folder). For consistency and handling multiple cursors from various inputs, it would be preferable though to have a more consistent way of handling cursors. 

2. It is required to be able to change the state of the cursor (show/hide) at runtime.

wiwei:
Note on the failing build - one of our build machines looks to be misconfigured so it's chronically red. Looking at the build logs it looks like this change is good (built x86/arm)

Issue #3971, opened 2019-04-15
Problem with MixedRealityInputSimulationProfile serialization


genereddick:
## Overview
The error message below is showing up on a deploy to the WMR Headset (but not on HoloLens on the in the Unity editor). I've separated it from #3944.

From @lukastoenneMS's comment in #3944 

The problem with the MixedRealityInputSimulationProfile seems to be that the data providers in InputSystem use a generic BaseMixedRealityProfile. When the standalone app is built the InputSimulation assembly is excluded since it's editor-only. The final type of the data provider's profile asset (MixedRealityInputSimulationProfile, 300 bytes) is then missing and the serialization interprets the asset as a BaseMixedRealityProfile (76 bytes).

@davidkline-ms is looking into it.

---

From #6054:

>The DefaultMixedRealityInputSimulationProfile is part of the the simulation data provider in the MRKT input profile. But at the same time it's AMSDEF is Editor and UWP only, causing a deserialization problem on all other platforms.
Since it's the only service (I think) that's only being used in Editor and the fact that it also compiles for UWP makes me believe that either this could very well be just compiled for all platforms and we're done with it, or the profile needs to be extracted out, which I tried and stopped after the 3rd file.

## Expected Behavior
No error

## Actual Behavior
Error message:

A scripted object (probably Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSimulationProfile?) has a different serialization layout when loading. (Read 76 bytes but expected 300 bytes)
Did you #ifdef UNITY_EDITOR a section of your serialized properties in any of your scripts?

## Steps to reproduce
_(Links to sample github project preferred)_

## Unity Editor Version
2018.3.12f1

## Mixed Reality Toolkit Release Version
MRTK latest

davidkline-ms:
@genereddick, have you noticed any side effects other than the message in the debugger?

Currently working on a possible solution, just wondering if there are any ill effects.

genereddick:
@davidkline-ms 
In addition to the head-locked (well, one cross-eyed eye-locked) error window
Selection no longer works, both in app but also can't even exit to the MR Portal homepage (so have to shutdown the MR Portal to get out)
Can still look around, but screen is inconsistently frozen / distorted -- seems similar to the effects of extremely low frame rate jerkiness

genereddick:
@davidkline-ms 
[...]

PR #4002, opened 2019-04-17
Eliminate serialization warning message from the input simulation profile

davidkline-ms:
The input simulation service is only valid when running in the editor. However, since it uses a custom profile, it gets deserialized on the target runtime platform as part of loading the mixed reality toolkit profile.

The root cause of the issue was that the profile class was not known by the serializer because the assembly containing it was not in the runtime image. This is likely to be true for other, platform specific extensions and providers (guidance on how to resolve this warning needs documentation).

The solution was to create a types assembly for the input simulation service that is marked as all platforms. This assembly contains the profile and the required data types and is referenced by the other input simulation assemblies.

To avoid a circular asm reference, an interface was created (and located in the new assembly) from which the input simulation service now inherits.

Tested in editor (confirm simulation remains functional) and on device (confirm no serialization warning messages in the debugger)

- Fixes: #3971 .


lukastoenneMS:
Instead of moving the non-editor-only parts into /Core, would it make sense to make the main InputSimulation assembly platform independent and in turn move the InputSimulationService into a new editor-only sub-assembly "InputSimulation/Service"? Seems that is the only class actually left in the editor part, and a bit more intuitive than nesting a platform independent assembly inside an editor-only one (even though technically they just live side-by-side).

davidkline-ms:
It actually didnt move the non-editor parts into the core it just was a name i picked :) I am good with using a new name for the editor components (in fact, we would potentially merge the inspector assembly and the editor only class into the same assembly.

Thoughts?

davidkline-ms:
@lukastoenneMS, updated the asmdef relationships per our discussion. the nicest thing about this is that we were able to go back to just two assemblies

PR #4013, opened 2019-04-18
Pointer dragged event

luis-valverde-ms:
Added a pointer updated event to be raised every frame a pointer is down. The intent is to provide the means to implement drag behaviors based on events instead of forcing client code to poll the pointer every frame.


Railboy:
Not sure I have a better alternative, but IMixedRealityPointerUpdatedHandler sounds like something that would receive pointer updates regardless of whether it's down or not. MixedRealityPointerDownUpdateHandler is a mouthful.

luis-valverde-ms:
I'm not too happy with pointer updated either. Pointer dragged?

Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerUpdatedHandler.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine.EventSystems;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Interface to implement to react to simple pointer input and pointer updates.
+    /// </summary>
+    public interface IMixedRealityPointerUpdatedHandler : IMixedRealityPointerHandler
```
julenka: I like IMixedRealityPointerDragHandler. What are your thoughts on just adding OnPointerUpdated to the IMixedRealityPointerHandler interface? I can see a downside of this being a breaking change. The upside is it does not add another interface to MRTK.

julenka: I like OnPointerDragged instead of OnPointerUpdated as well. So IMixedRealityPointerDraggedHandler, OnPointerDragged.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -877,6 +877,27 @@ public void RaisePointerDown(IMixedRealityPointer pointer, MixedRealityInputActi
...
+            delegate (IMixedRealityPointerUpdatedHandler handler, BaseEventData eventData)
+            {
+                var casted = ExecuteEvents.ValidateEventData<MixedRealityPointerEventData>(eventData);
+                handler.OnPointerUpdated(casted);
+            };
+
+        /// <inheritdoc />
+        public void RaisePointerUpdated(IMixedRealityPointer pointer, MixedRealityInputAction inputAction, Handedness handedness = Handedness.None, IMixedRealityInputSource inputSource = null)
+        {
+            pointer.IsFocusLocked = (pointer.Result?.Details.Object != null);
```
julenka: Why are we setting IsFocusLocked here?

Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerUpdatedHandler.cs
luis-valverde-ms: I would like to add it to IMixedRealityPointerHandler (OnPointerDragged?) but it is a breaking change and I don't know how much wiggle room we have there. It is the best type of breaking change though, in the sense that in produces a build time error (missing implementation) that is very straightforward to fix. @wiwei , any thoughts on that?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
luis-valverde-ms: Mindless copy-paste. Good catch!

Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerUpdatedHandler.cs
wiwei: So we're in an odd state. Because we haven't clearly defined the delineation between what is internal vs what is external, technically speaking every single change that we make is technically a breaking change, because someone could have taken a dependency on a utility class that is nested deep within our code, but is still technically public because we distribute as scripts.

This is a problem that we need to solve, but I figured it was interesting to write here because it highlights that there won't be a clear easy answer due to that.

That being said, because of the timing of where we are right now, it feels we might still have some time to do the right long term thing (i.e. IMixedRealityPointerHandler), assuming that we make it very explicit (i.e. we write a doc that highlights a limited set of changes between RC1->) maybe this is okay. We can followup with product on this to see what the agreement is.

wiwei: Okay circling back here, it sounds like we're comfortable with making this one change given the tradeoffs (i.e. no incurring immediate interface technical debt going forward), and we will author a changes doc (in release notes) that describes this change, how it will surface as an error, and provide clear guidance on how to fix it.

wiwei:
>                 {

I think that this comment will be mooted by the update to move the new function into IMixedRealityPointerHandler. For the code as-is (as of the time I'm reviewing this), it looks like this function is still around even though this object no longer implements the interface that declares it (i.e. OnPointerUp is part of IMixedRealityPointerHandler, but this class no longer implements that interface). 

As such, it seems like this would be a candidate for deletion, but maybe it should be kept around after doing the work to move OnPointerUpdated into IMixedRealityPointerHandler.

Either way, just a candidate to take another look at after your interface rework changes.

---
Refers to: Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs:423 in 88ee4a9. [](commit_id = 88ee4a9dd42171fd0ab13f860a950e456f1661ab, deletion_comment = False)

wiwei:
>                 {

Oh doh, missed the obvious fact that the interface extended the old one, please ignore :)

---
In reply to: [484599126](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4013#issuecomment-484599126) [](ancestors = 484599126)

---
Refers to: Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs:423 in 88ee4a9. [](commit_id = 88ee4a9dd42171fd0ab13f860a950e456f1661ab, deletion_comment = False)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
```diff
@@ -162,7 +162,10 @@ public override int GetHashCode()
 
         public void OnPostSceneQuery()
         {
-
+            if (isSelectPressed && IsInteractionEnabled)
```
wiwei: >IsInteractionEnabled [](start = 35, length = 20)

For my own knowledge, what's the reason that only the GGV pointer has the IsInteractionEnabled check?

Documentation/Input/Pointers.md
```diff
@@ -14,5 +14,6 @@ Handler | Events | Description
 --- | --- | ---
 [`IMixedRealityFocusChangedHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityFocusChangedHandler) | Before Focus Changed / Focus Changed | Raised on both the game object losing focus and the one gaining it every time a pointer changes focus.
 [`IMixedRealityFocusHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityFocusHandler) | Focus Enter / Exit | Raised on the game object gaining focus when the first pointer enters it and on the one losing focus when the last pointer leaves it.
-[`IMixedRealityPointerHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityPointerHandler) | Pointer Down / Up / Clicked | Raised to report pointer input.
+[`IMixedRealityPointerHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityPointerHandler) | Pointer Down / Up / Clicked | Raised to report pointer presses and releases.
+[`IMixedRealityPointerUpdatedHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityPointerUpdatedHandler) | Pointer Updated | Raised every frame a pointer is down. Useful to implement drag-like behaviors.
```
wiwei: Can we open an issue (somewhere) to track the doc writing/release notes writing so we don't lose track of this?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
luis-valverde-ms: For consistency with the way we raise the down and up events in OnInputDown/Up(). Maybe @julenka can give background on why it is done that way.

Documentation/Input/Pointers.md
luis-valverde-ms: #4025

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/PointerClickHandler.cs
```diff
@@ -47,6 +47,8 @@ public void OnPointerDown(MixedRealityPointerEventData eventData)
             }
         }
 
+        public void OnPointerDragged(MixedRealityPointerEventData eventData) { }
```
luis-valverde-ms: Add <inheritdoc />. Same anywhere else, matching what is already in the file.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
julenka: Just because we don't want to raise pointer down or up events if the pointer is not enabled, meaning in GGVPointer case to not raise events when the object is not active. For other interactions for example hand rays, we don't want to raise events if say the hands are facing the user and the rays should be off

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/LinePointer.cs
```diff
@@ -143,6 +143,8 @@ public override void OnPreSceneQuery()
         /// <inheritdoc />
         public override void OnPostSceneQuery()
         {
+            base.OnPostSceneQuery();
```
julenka: Interesting, sounds like we weren't doing this before. Was this a bug?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -877,6 +877,25 @@ public void RaisePointerDown(IMixedRealityPointer pointer, MixedRealityInputActi
 
         #endregion Pointer Down
 
+        #region Pointer Updated
```
julenka: Pointer Dragged.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/LinePointer.cs
luis-valverde-ms: It was fine because the base implementation was empty.

PR #4016, opened 2019-04-18
App State Sharing Extension

Railboy:
Overview
---
A stripped-down, MRTK-service-ified version of the AppState system proposed in #3210

The AppState service provides a means to synchronize a set of structs among connected users. That's it - it doesn't know or care about scene objects, and it's up to the dev to decide what info to synchronize and when.

If you want examples of other systems that can build on AppState to generate / manipulate scene objects, there are several in #3210.

Feedback Goals
---
This is an exploratory PR that won't be ready for general consumption until our networking interface is complete. Until then we hope to incorporate feedback on:
- General clarity. Does the service make its purpose and requirements clear with its inspector and profile?
- Workflow for setting up state structs for syncing - looking for friction / confusion.
- Going through the process of connecting two editor instances and/or devices - looking for friction / confusion.

AppStateData / AppStateDataSubscriptions
---
This is a Photon-specific implementation of the IAppStateData interface. IAppStateData is the conduit through which the AppState synchronizes data, and IAppStateDataSubscriptions is used to configure which users receive which data. I would like to remove the Photon dependency and use an MRTK networking service in place of these components.

Profile Setup
---
The profile inspector helps you search for and configure app state structs and (in the case of IL2CPP builds) state array types.

AppState

Other Issues
---
- The connection process handled by PhotonServiceConnector is opaque and flakey. This is an issue that will hopefully be solved by using an MRTK networking service.


julenka:
This looks related to sharing, @andreiborodin

Railboy:
In the long run this extension will be consuming @andreiborodin's interface. We've been syncing offline.

luis-valverde-ms:
@Railboy , is this PR still alive? Are you waiting on reviews or other PRs?

Railboy:
@luis-valverde-ms I'm mainly waiting on our networking service interface to be finalized so this extension can consume that instead of Photon. Reviews in the meantime would be welcome.

lukastoenneMS:
Looks like this needs a merge from mrtk_development if it should be reviewed now, otherwise i would suggest closing it and opening a new PR when the networking service interface is available (to shorten our PR list).

PR #4017, opened 2019-04-18
Update the Pull Request template to include a "Verification" section

wiwei:
One idea point of feedback around pull requests is that it's often not clear the level of verification expected from reviewers - should reviewers be patching all changes to be validating them, only specific ones. For new features, it's also sometimes unclear how a reviewer can functionally test things (i.e. exactly what should be tested, how to engage the scenarios, etc). One idea is to provide an optional section on the PR template where people can call out specifics they want reviewed (i.e. please patch my change and test ABC out, or please pay attention to this specific thing)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,12 +7,12 @@
 
 ## Verification
 > This optional section is a place where you can detail the specific type of verification 
-> you want from reviewers. For example, if you specifically want reviewers to checkout 
-> the PR locally and validate the functionality of specific scenarios, provide instructions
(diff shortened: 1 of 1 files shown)


wiwei:
> Can we provide example instructions like command line commands for how to check out the fork? I feel like people will be more likely to check out the fork if they have steps that can copy/paste

Done - I spent some time looking to see if github markdown supports variables (in that, is there some macro/variable I would use such that {PR_ID} would be auto-replaced by the actual ID of the pull request, so that people who were reviewing could literally copy paste and the commands would just work without their intervention in substituting values)

I didn't have any luck so I think this is the next best thing.

Issue #3987, opened 2019-04-16
Proposal: Require / enforce additive scene loading via scene service


Railboy:
## Overview
Unity suggests using a [base manager scene](https://docs.unity3d.com/Manual/MultiSceneEditing.html) and additively loading / unloading content on top of that. In theory we recommend this approach but in practice our story is muddled.

Docs and editor prompts should make it clear that single-scene loading is **not officially supported** and that MRTK instances / services will be (gracefully) destroyed by single loads. DontDestroyOnLoad should never be used in any case by core services or scripts.

## Scene service / Manager vs Content scenes
A scene management core service should be created to enforce this convention. Like all core services it should be possible to disable and ignore, but using it would make loading / unloading additive scenes simpler both at runtime and in editor.

The profile for this service would require devs to specify a 'Manager' scene - the presence of an MRTK instance would be enforced in this scene.

All other scenes would be considered 'Content' scenes - these can contain MRTK instances if devs want to quickly test them in isolation, but those instances would be ignored when additively loaded.

The service should also help devs avoid common multi-scene gotchas like enforcing lighting settings across multiple scenes, navmesh settings, etc.

## Multiple MRTK instances in editor
To avoid confusion in cases where multiple MRTK instances are loaded across multiple scenes in the editor, one instance would be visibly 'active' (either an arbitrary instance or the Manager instance if present) while the rest would be visibly 'disabled.'

## Add 'OnSceneAdded / OnSceneRemoved' to service interfaces
Services should be able to handle the loading and unloading of content gracefully. Adding these methods to the base service interface would set that expectation.

Related: #3639, #3601, #3599, #3261, #3246, #3941

chrisfromwork:
Assigning to @Yoyozilla to schedule the associated work.

cre8ivepark:
We've seen many customers struggling with multi-scenes applications in HoloToolkit (Camera management, Cursor/Controller/Focus management, Services management...) Definitely, it would be really helpful to provide clear guideline and methods for it.

StephenHodgson:
I think creating a scene loading service is overkill and a bit over engineered. `SceneManager` should be enough and we shouldn't rewrite(overwrite) how the editor handles this functionality.

The only real problem is during development, when devs are moving between scenes in the editor.

I've already had plans to support this for some time and in my fork, I've set it up to that developers need only press play and the base scene automagically loads in for them and they don't have to worry about anything.

[...]

PR #4035, opened 2019-04-19
Feature: multiple MRTK instances

Railboy:
Overview
---
Adds support for multiple MRTK instances as well as hot-swapping of instances.

First step of the scene-related plan laid out in #3987. This is an alternative to the approach in #3941. The final result may be a blend of the two. We should discuss.

Changes
---
- `DontDestroyOnLoad` is no longer used to protect instances.

- `MixedRealityToolkit.IsInitialized` simply checks if `activeInstance` is null. It no longer generates an instance.

- `MixedRealityToolkit.Instance` will generate an instance if none is found. If the active instance is null and multiple instances exist in the loaded scene(s), one is arbitrarily chosen as the active instance. All other instances are disabled and renamed. These inactive instances can be deleted / duplicated / whatever to no effect.

MultipleInstances

- An inactive instance displays an inspector letting you know you're looking in the wrong place.

InactiveMessage

- Destroying the active instance results in the next available inactive instance being set active. If all instances are destroyed, all services are disabled. (Note that this doesn't necessarily mean the services handle this well or clean up after themselves.)

- Newly detected instances (eg from an additive load) will be disabled in favor of the active instance.

Issues
---
- We're assuming each instance will be using the same configuration profile. I suppose it's possible to detect newly loaded instances with a different profile, shut down existing services, then re-initialize with the new profile. But I doubt we want to go there.

- This setup suffers from the CameraCache and MixedRealityPlayspace issues we're working out in #3890

Testing Requirements
---
We want to try destroying / creating MRTK instances in every possible combination. Multiple instances loaded via multiple additive scenes, multiple instances in one scene, unloading, deleting manually, duplicating manually - every permutation we can think of.

In all cases there should be uninterrupted continuity of services provided at least one instance exists and has a configuration profile.

We also need to be testing for race conditions during instance initialization.

---
- First step in fixing: #3987


Railboy:
Been working with this locally for about a week without incident. Setting this as ready for review.

Railboy:
Labeling this to match SceneSystem service since it's a prerequisite.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -745,75 +728,124 @@ public Transform MixedRealityPlayspace
...
+                MixedRealityToolkit.activeInstance = toolkitInstance;
+                toolkitInstances.Add(toolkitInstance);
+                toolkitInstance.gameObject.SetActive(true);
+                toolkitInstance.InitializeInstance();
+                toolkitInstance.name = ActiveInstanceGameObjectName;
+                return;
+            }
+
+            // If we're already registered, no need to proceed
+            if (!toolkitInstances.Add(toolkitInstance))
```
julenka: nit: can we please put brackets with our if statements, to follow the coding guidelines? You do this above.

julenka:
> Adds support for multiple MRTK instances as well as hot-swapping of instances.

I read the proposal and also the PR description, and I'm not sure I fully understand what this change enables, from the developer perspective. Could you please give an example? Reason I ask for this detail is because it sounds like a big change to how MRTK work, at least from the description.

For example, does this now mean that I could have two instances of MRTK in my scene, and I can set one instance of MRTK active? Out of curiosity, why would I want to do that?

julenka:
> We want to try destroying / creating MRTK instances in every possible combination. Multiple instances loaded via multiple additive scenes, multiple instances in one scene, unloading, deleting manually, duplicating manually - every permutation we can think of.

> In all cases there should be uninterrupted continuity of services provided at least one instance exists and has a configuration profile.

> We also need to be testing for race conditions during instance initialization.

Great writeup about what we want to do regarding tests! If we are not going to make these tests in this PR (which would be amazing, by the way, and a great example of what we should strive for), can we please make a task to write these tests? That way these tests can run as part of CI and ensure that this change or future changes to scene loading don't break adding / removing scenes.

Railboy:
>For example, does this now mean that I could have two instances of MRTK in my scene, and I can set one instance of MRTK active? Out of curiosity, why would I want to do that?

Having two instances in one scene isn't so useful, but having _multiple_ instances in _multiple_ scenes is very useful:

MultipleInstances

Two of the three instances have quietly disabled themselves and will not update services, so I can hit Play and preview the content from all three scenes simultaneously.

This means you can plan around additively loading your content in separate scenes while still keeping an instance of the toolkit in each scene for testing content individually and/or in any combination. This is true both in the editor and at runtime.

Railboy:
> If we are not going to make these tests in this PR (which would be amazing, by the way, and a great example of what we should strive for)

I've got some tests made locally. I didn't want to commit them and dirty up the history.

But if we're all cool with committing temp testing stuff that gets nuked prior to merge, I'd be happy to include them.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -26,12 +26,17 @@ namespace Microsoft.MixedReality.Toolkit
     /// The Profile can be swapped out at any time to meet the needs of your project.
     /// </summary>
     [DisallowMultipleComponent]
+    [ExecuteAlways]
```
julenka: What is the consequence of setting MRTK class to ExecuteAlways? Why do we need this?

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
julenka: HI @Railboy  could you please add the brackets around if statements? Just a reminder.

julenka:
Hi @StephenHodgson, I noticed you downvoted this change, I was wondering could you please explain why? I'm new to MRTK internals still and you created them, so I wonder if you see something wrong with this change that we are missing.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
Railboy: This ensures that its OnEnable / OnDisable will be called when you're not in playmode, and I'm using those to detect and register instances.

Railboy:
@julenka I've committed a folder called TemporaryTesting that includes my additive and single load tests. We'll have to delete this folder before merging. I'm adding a DO NOT MERGE tag until that's been done.

So far I've been doing editor testing by hand but I'll try to add a few automatic tests to EditorTesting.cs soon.

julenka:
> @julenka I've committed a folder called TemporaryTesting that includes my additive and single load tests. We'll have to delete this folder before merging. I'm adding a DO NOT MERGE tag until that's been done.

Thanks, I ran these tests today. When I ran the tests, I got the following errors. I just opened the "AdditiveLoadTesting" scene and hit play.

Error #1:
```
There's already a UIRaycastCamera in the scene. It will be ignored, so please delete it to avoid confusion.
UnityEngine.Debug:LogError(Object, Object)
Microsoft.MixedReality.Toolkit.Input.FocusProvider:Initialize() (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:427)
```

Error #2:

```
Gaze was never registered with the Input Manager!
UnityEngine.Debug:Assert(Boolean, String)
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:RaiseSourceLost(IMixedRealityInputSource, IMixedRealityController)
```

I do not see this normally when I run a scene with MRTK in it.

julenka:
I was also able to add multiple scenes in the scene viewer, which is great! I was also able to successfully run all the unit tests. 

I think it would be great to add these additive scene loading tests to the unit tests, simply because it will ensure that adding / editing multiple scenes won't get broken by other changes. This can be done in separate PR if this one needs to be completed urgently.

julenka:
I ran into one issue when testing multiple scene loading, but it is not a regression because the behavior also happens without your change.

To reproduce, please do the following:

1. Check out this branch: https://github.com/julenka/MixedRealityToolkit-Unity/tree/multiscene_mrtk_bug 
2. In build window, remove all scenes, then add Assets/Scene{0,1,2} to scenes in build settings
3. Open Assets/Scene 2
4. Hit "Play", and press the 'next scene' button using mouse button. This will attempt to load Scene0 using `SceneManager.LoadScene`.

Expected: the new scene 0 opens and no errors occur. I expect this to happen because even though scene0 has a second MRTK instance, I thought the PR you had would support multiple instances of MRTK.

Actual: When scene0 loads, I get the following error:

> Unable to find IMixedRealityInputSystem service.
UnityEngine.Debug:LogError(Object)
Microsoft.MixedReality.Toolkit.MixedRealityToolkit:GetService(String, Boolean) (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:271)

Followed by many of these errors:

> Some objects were not cleaned up when closing the scene. (Did you spawn new GameObjects from OnDestroy?)
The following scene GameObjects were found:
MixedRealityToolkit


and:

> NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.Input.GenericPointer..ctor (System.String pointerName, Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputSource inputSourceParent) (at Assets/MixedRealityToolkit/Providers/GenericPointer.cs:23)
Microsoft.MixedReality.Toolkit.Input.GazeProvider+InternalGazePointer..ctor (Microsoft.MixedReality.Toolkit.Input.GazeProvider gazeProvider, System.String pointerName, Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputSource inputSourceParent, UnityEngine.LayerMask[] raycastLayerMasks, System.Single pointerExtent, UnityEngine.Transform gazeTransform, Microsoft.MixedReality.Toolkit.Physics.BaseRayStabilizer stabilizer) (at Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs:179

Assets/TemporaryTesting/AdditiveLoadTesting.cs
```diff
@@ -0,0 +1,116 @@
+﻿using Microsoft.MixedReality.Toolkit;
```
julenka: Reminder to delete this file and everything else under TemporaryTesting before checking in.

julenka: And then put it in a Unity test if possible, can be done in separate PR.

ProjectSettings/EditorBuildSettings.asset
```diff
@@ -26,4 +26,13 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Scenes/EyeTrackingDemo-05-Visualizer.unity
     guid: 2d6c43a82f3a88c4dbdc3b5e99a68d8a
+  - enabled: 1
```
julenka: Reminder to revert this before merging the PR

julenka:
@luis-valverde-ms  you might want to have a look at this PR, I feel it's related to the PR you're working on #3941

davidkline-ms:
@Railboy, this needs to be validated with the changes from #4099. I will work with you on that :)

davidkline-ms:
>  I suppose it's possible to detect newly loaded instances with a different profile, shut down existing services, then re-initialize with the new profile. 

This feels like the primary benefit of this work. Easy switching between different configurations.

Railboy:
@julenka Now that @davidkline-ms's playspace changes are in, I'm working on removing those final errors.

Assets/TemporaryTesting/AdditiveLoadTesting.cs
Railboy: Removed.

wiwei:
At this point this is good to go right? Do you just need someone to do one more pass or do you just need a button presser?

davidkline-ms:
> Do you just need someone to do one more pass or do you just need a button presser?

The code looks good. Before we merge this change, I would like to meet with @Railboy and discuss implications on the service registry.

ProjectSettings/ProjectSettings.asset
```diff
@@ -3,7 +3,7 @@
 --- !u!129 &1
 PlayerSettings:
```
Railboy: I'm not loving all the changes being made to player settings here. I'm thinking we should revert all of this + quality settings before merge.

Railboy: Restored.

Railboy:
@wiwei At this point we're good to go. There may still be unanticipated side effects but we can deal with them as they happen.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -526,99 +538,68 @@ private void EnsureMixedRealityRequirements()
...
-                }
-
-                if (Application.isPlaying && !searchForInstance)
-                {
-                    return null;
+                    return activeInstance;
                 }
 
-
                 var objects = FindObjectsOfType<MixedRealityToolkit>();
```
provencher: I think we should remove this check. 
If there is no instance in the scene, this check will run each time we do a null check against the instance, and no instance will ever be added.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -526,99 +538,68 @@ private void EnsureMixedRealityRequirements()
...
-#endif
-                        Debug.Assert(!newInstanceBeingInitialized, "We shouldn't be initializing another MixedRealityToolkit unless we errored on the previous.");
-                        newInstanceBeingInitialized = true;
-                        newInstance = new GameObject(nameof(MixedRealityToolkit)).AddComponent<MixedRealityToolkit>();
                         break;
+
                     case 1:
-                        newInstanceBeingInitialized = false;
-                        newInstance = objects[0];
+                        RegisterInstance(objects[0]);
```
provencher: I don't see the need to register the instance in the getter. When the instance is created, it should register itself then, not when someone calls it.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -526,99 +538,68 @@ private void EnsureMixedRealityRequirements()
...
-                    // Setup any additional things the instance needs.
-                    newInstance.InitializeInstance();
-                }
-                else
+                if (activeInstance == null)
                 {
-                    // Don't do any additional setup because the app is quitting.
-                    instance = newInstance;
-                    newInstanceBeingInitialized = false;
+                    Debug.LogError("Didn't find / couldn't create an active instance of mixed reality toolkit. You can add one to your scene via 'Mixed Reality Toolkit->Add to Scene and Configure'");
```
provencher: This is not an error in my opinion. If we try and use a null instance, we'll get a null reference exception, but if it's null and we null check it, we have enough information there that we don't need to throw an additional error log.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -526,99 +538,68 @@ private void EnsureMixedRealityRequirements()
...
         /// If no instance is found, then we search for an instance in the scene.
-        /// If more than one instance is found, we log an error and no instance is returned.
+        /// If more than one instance is found, only the first instance is activated. All others are deactivated.
         /// </summary>
         public static MixedRealityToolkit Instance
         {
             get
             {
-                if (IsInitialized)
+                if (activeInstance != null)
```
provencher: A null check on a monobehavior is actually non-trivial due to the overridden checks.
This gets called enough that we shouldn't null check it.
https://docs.unity3d.com/Manual/BestPracticeUnderstandingPerformanceInUnity7.html

Since we no longer create any instances in the getter, I think we should slim this down and simply return whatever the active registered instance is, rather than adding any logical overhead.

provencher:
@Railboy sorry for the late review. Please address my concerns before merging this.

Railboy:
>The instance getter should not have any logic in it since we no longer create instances in it.

@provencher I'm going to make this change and remove the find by type failsafe but I'll emphasize the need for testing one more time. Depending on unity object lifecycles always makes me nervous even if they should work perfectly in theory.

provencher:
The only case where the unity lifecycle may pose issues is if the only MRTK instance is introduced in an additively loaded scene. This is because the Awake called in the same blocking process as the scene load.

I recommend against loading the MRTK, or any central singleton this way anyway.

The changes seem great otherwise. I'll try doing some testing further this week, but I can guarantee it won't cause issues in my usecase.

Railboy:
>The only case where the unity lifecycle may pose issues is if the only MRTK instance is introduced in an additively loaded scene. This is because the Awake called in the same blocking process as the scene load.

This PR is a rereq for the SceneSystem service, which is focused on additive scene loading, and which expects multiple instances to be loaded. So that's _exactly_ my concern. But we can wait for it to be a problem before we try to fix it.

provencher:
@Railboy I understand what you're saying but it depends on how you go about it.

If your base scene has an MRTK instance in it, and you additively add scripts that depend on the MRTK instance, all is well. 

The issue arrises when the base scene does not have any MRTK instance in it, and you additively load it after, while calling MRKT.Instance in awake. I recommend against this from my experience with Unity.

provencher:
The way I solved this issue of loading in dependencies at runtime is by having a script that loads prefabs.

The prefab loaded has a scriptable object with a list of prefabs that need to be loaded, and that get loaded on Awake, with the highest priority in the script execution list. This ensures that any prefabs needed in the scene are instantiated before anything else can call it.

The scene loading system has no guarantees on timing, so using it to handle dependencies is poor practice.

Railboy:
@provencher The prefab approach sounds pretty bulletproof but I suspect most devs would not be on board. Let's discuss further in the SceneSystem PR.

davidkline-ms:
From the description:
>  one is arbitrarily chosen as the active instance

Non-deterministic behavior concerns me. Once this is in, we should design a reliable active instance selection pattern.

Railboy:
@davidkline-ms We could preserve the load order on registration and go with that.

davidkline-ms:
@Railboy or heirarchy order, maybe?

_Not_ for this change, though. It can be a follow-on PR

Assets/MixedRealityToolkit.Generated.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
davidkline-ms: Were the *Generated/* meta files added intentionally?

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs
```diff
@@ -55,6 +74,7 @@ public override void OnInspectorGUI()
                         profile.CreateAsset("Assets/MixedRealityToolkit.Generated/CustomProfiles");
                         activeProfile.objectReferenceValue = profile;
                         Selection.activeObject = profile;
+                        Debug.Log("display dialog");
```
davidkline-ms: it: do we need this (and related) log message?

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -526,99 +531,31 @@ private void EnsureMixedRealityRequirements()
...
         /// <summary>
-        /// Lock property for the Mixed Reality Toolkit to prevent reinitialization
+        /// Returns the Singleton instance of the classes type.
         /// </summary>
-        private static readonly object initializedLock = new object();
+        public static MixedRealityToolkit Instance => activeInstance;
 
         private void InitializeInstance()
         {
-            lock (initializedLock)
```
davidkline-ms: do we no longer need this lock?

ProjectSettings/XRSettings.asset
```diff
@@ -0,0 +1,10 @@
+{
```
davidkline-ms: Were these changes intentional?

Railboy:
@davidkline-ms Are you seeing that error in:
- a sample scene
- single scene with no instance
- single scene with multiple instances
- multiple scenes with no instance
- multiple scenes with single instance
- multiple scenes with multiple instances?

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
Railboy: It seems to me we could remove it, but I didn't want to take that step until someone else had looked at it.

provencher: Well the instance is a monobehavior, and those fail when being called from other threads, so the lock is entirely superfluous

provencher: Well the instance is a monobehavior, and those fail when being called from other threads, so the lock is entirely superfluous

provencher: Well the instance is a monobehavior, and those fail when being called from other threads, so the lock is entirely superfluous

davidkline-ms:
@Railboy, looks like this change broke some of the tests... Can you take a look at them?

Railboy:
Tests updated. MixedRealityToolkit instances are not auto-generated by calls to Instance any more, so we had to add an extra step.

davidkline-ms:
Thanks @Railboy!

PR #4039, opened 2019-04-19
Manipulation event data and clipping primitive fix

radicalad:
Overview
---
A few minor updates I'd like to add in before too much time passes.

Changes
Exposed a few extra properties on the `ManipulationEventData` class and updated `ManipulationHandler` accordingly. Currently there is no way to know what `ManipulationHandler` sent the ManipulationEvent. While I was in there I exposed a few other things that looked useful such as Pointer _position_, _velocity_, and _angular velocity_. 

Fixes
---
Added missing `OnEnable` preRender check on `ClippingPrimitive.cs`. #3886

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Events/ManipulationEventData.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Events/ManipulationEventData.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using UnityEngine;
+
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/Input/Events/ManipulationEventData.cs
```diff
@@ -1,13 +1,38 @@
...
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using UnityEngine;
+
 namespace Microsoft.MixedReality.Toolkit.UI
 {
     /// <summary>
     /// Information associated with a particular manipulation event.
     /// </summary>
     public class ManipulationEventData
```
Railboy: This class makes more sense now IMO.

PR #4041, opened 2019-04-20
Update README_Interactable.md

cre8ivepark:
Overview
---
Adding overview description to Interactable README file. (mrtk_release)


keveleigh:
Since docs are currently published out of mrtk_development, I think it makes sense to make this change there.

Issue #4031, opened 2019-04-19
Proposal: Create a core camera management service which can be disabled / configured to work with AR


Railboy:
## Describe the problem

The MRTK exercises a great deal of control over the main camera. This can lead to problems when using ARKit / ARCore / ARFoundation.

## Describe the solution you'd like

A camera management service should be written. The default service should mimic the toolkit's current functionality. Alternate versions can be written to accommodate AR solutions and/or more complex camera setups. And like any core service, it should be possible to disable camera control altogether.

This has the added benefit of making it easier to use the toolkit as a pure service provider, if desired.

Related issues: #3616, #3948, #3762

chrisfromwork:
Assigning to @Yoyozilla  and @wiwei to try and find an appropriate owner

PR #4047, opened 2019-04-22
Moves camera management into core CameraService

Railboy:
Overview
---
Re-creates existing MRTK camera management as a core service.

This makes it easy for devs to replace the camera system with something that doesn't interfere with ARCore / ARKit camera setups.

The new service uses the existing camera profile and should be minimally disruptive.

Changes
---
- Created CameraSystem service which is now registered alongside other core services on startup.
- Removed camera management from MRTK singleton.
- Added cameraSystemType field to MRTK config profile.
- Removed isCameraPersistent option from profile - this is at odds with our [additive loading 
recommendations.](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/3987) This is something devs should be doing in their own service implementations.
- Set default config profile to use MixedRealityCameraSystem.

Testing Requests
---
Verify that camera behavior is identical to previous setup (with the exception of persistent camera).

---
- Implements #4031


Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -0,0 +1,91 @@
...
+            else
+            {
+                ApplySettingsForTransparentDisplay();
+            }
+        }
+
+        /// <inheritdoc />
+        public override void Update()
+        {
+            if (cameraProfile.IsOpaque)
```
StephenHodgson: We should probably check if this has changed, instead of setting it each update.

Assets/MixedRealityToolkit/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.CameraSystem
+{
+    public interface IMixedRealityCameraSystem : IMixedRealityEventSystem, IMixedRealityEventSource, IMixedRealityDataProvider { }
```
StephenHodgson: All the public methods and properties should be defined here.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -0,0 +1,99 @@
...
+    public class MixedRealityCameraSystem : BaseCoreSystem, IMixedRealityCameraSystem
+    {
+        public MixedRealityCameraSystem(
+            IMixedRealityServiceRegistrar registrar,
+            BaseMixedRealityProfile profile = null) : base(registrar, profile)
+        {
+            cameraProfile = (MixedRealityCameraProfile)profile;
+        }
+
+        private MixedRealityCameraProfile cameraProfile;
```
StephenHodgson: We should probably move the `IsOpaque` property into the instance.

Assets/MixedRealityToolkit/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
Railboy: At the moment there aren't any. Are you thinking we should cache the profile settings on initialize and then let devs modify those settings through the interface?

StephenHodgson: Yes there is: `IsOpaque` It's defined in the profile, but shouldn't be.

StephenHodgson:
@Railboy great changes. I think it was always intended to have a camera service, but at the time it wasn't needed. Great work

Assets/MixedRealityToolkit/Definitions/MixedRealityToolkitConfigurationProfile.cs
```diff
@@ -64,6 +67,20 @@ public MixedRealityCameraProfile CameraProfile
...
+        public SystemType CameraSystemType
+        {
+            get { return cameraSystemType; }
+            internal set { cameraSystemType = value; }
+        }
+
+        [SerializeField]
+        [Tooltip("Camera System Class to instantiate at runtime.")]
+        [Implements(typeof(IMixedRealityCameraSystem), TypeGrouping.ByNamespaceFlat)]
+        private SystemType cameraSystemType;
```
luis-valverde-ms: >cameraSystemType [](start = 27, length = 16)

Will this not be better placed in the camera profile? The way I see it, the less information about the camera that leaks into the general MRTK profile the better.

Assets/MixedRealityToolkit/Definitions/MixedRealityToolkitConfigurationProfile.cs
```diff
@@ -38,16 +40,17 @@ public ExperienceScale TargetExperienceScale
...
 
         /// <summary>
         /// Enable and configure the Camera Profile for the Mixed Reality Toolkit
         /// </summary>
-        public bool IsCameraProfileEnabled
+        public bool IsCameraSystemEnabled
         {
-            get { return CameraProfile != null && enableCameraProfile; }
-            internal set { enableCameraProfile = value; }
+            get { return CameraProfile != null && enableCameraSystem; }
```
StephenHodgson:
```suggestion
            get { return CameraProfile != null && cameraSystemType != null && cameraSystemType.Type != null && enableCameraSystem; }
```

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -1252,6 +1246,35 @@ public static IMixedRealityBoundarySystem BoundarySystem
 
         private static bool logBoundarySystem = true;
 
+        private static IMixedRealityCameraSystem cameraSystem = null;
```
StephenHodgson: There's a clear cache method where this needs to get added to as well.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -407,6 +391,16 @@ private void InitializeServiceLocator()
                 }
             }
 
+            // If the Camera system has been selected for initialization in the Active profile, enable it in the project
+            if (ActiveProfile.IsCameraSystemEnabled)
+            {
+                object[] args = { this, ActiveProfile.CameraProfile };
+                if (!RegisterService<IMixedRealityCameraSystem>(ActiveProfile.CameraSystemType, args: args) || CameraSystem == null)
```
StephenHodgson: What the heck is going on here?
Is this a product of the new registrar system?

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -1252,6 +1246,35 @@ public static IMixedRealityBoundarySystem BoundarySystem
 
         private static bool logBoundarySystem = true;
 
+        private static IMixedRealityCameraSystem cameraSystem = null;
+
+        /// <summary>
+        /// The current Camera System registered with the Mixed Reality Toolkit.
+        /// </summary>
+        public static IMixedRealityCameraSystem CameraSystem
```
StephenHodgson: Pretty sure this also needs to be added to the core system check

Assets/MixedRealityToolkit/Definitions/MixedRealityToolkitConfigurationProfile.cs
StephenHodgson: No, it needs to be here, just like all the rest of the service types.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
Railboy: This is how other core services are registered. @davidkline-ms could probably tell you more.

Assets/MixedRealityToolkit/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.CameraSystem
+{
+    public interface IMixedRealityCameraSystem : IMixedRealityEventSystem, IMixedRealityEventSource, IMixedRealityDataProvider
```
luis-valverde-ms: I'm not sure we should be using an interface here. Nothing is calling IsOpaque via the interface and I don't think it sufficiently describes what a type must implement to be a camera system.

Assets/MixedRealityToolkit/Definitions/MixedRealityToolkitConfigurationProfile.cs
Railboy: IMO all the service types should be moved into their respective profiles. But this is how it's done right now.

Assets/MixedRealityToolkit/Definitions/MixedRealityToolkitConfigurationProfile.cs
Railboy: Is there a way to initialize a system type with a default value? Would be nice if we didn't have to worry about them blowing up.

Assets/MixedRealityToolkit/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
Railboy: It's thin at the moment but this is a baby step. The long term goal is to move ALL camera-related babysitting into this service.

If we don't want it to be this thin up front maybe we could expose profile-driven settings like camera background color, clear flags, etc?

luis-valverde-ms: I'd say let's make it an interface when/if we actually use the camera system through its interface and if the interface defines well what is expected of a camera system. I.e. by implementing the interface I should have a reasonable chance of creating a type that can work well in place of MixedRealityCameraSystem.

StephenHodgson: @luis-valverde-ms it's obvious you don't understand how the services work.

It's required to define an interface so to set it up in the configuration profile.

This PR effectively uses the camera service via interface.

Railboy: @luis-valverde-ms I agree with what you're saying but our hands are tied in the sense that an interface is required for this to be a service. And if it's not a service then the main benefit of making this move - the ability to swap it out for something custom - is lost.

Here's one idea. This PR leaves out two of the biggest camera-related behaviors in the toolkit - finding/creating the main camera and play space transform. Right now they're handled in CameraCache.cs and MixedRealityToolkit.cs. My plan was always to move those responsibilities into the camera service, but I left them out to keep the initial PR small.

If we move those right away, the interface would look (roughly) like this:

```
public interface IMixedRealityCameraSystem 
{
    // Supplies a main camera, either by finding or creating it.
    public Camera MainCamera { get; }

    // Supplies the parent transform for the main camera. If you want to move the player around, use this transform.
    public Transform PlayspaceTransform { get; }
}
```

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// <summary>
+        /// Is the current camera displaying on an Opaque (AR) device or a VR / immersive device
+        /// </summary>
+        public bool IsOpaque
+        {
+            get
+            {
+                currentDisplayType = DisplayType.Opaque;
+#if UNITY_WSA
+                if (!UnityEngine.XR.WSA.HolographicSettings.IsDisplayOpaque)
```
julenka: I wonder if you are building for ARCore / ARKit the display type should be transparent. Also...magic leap also should be transparent.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -0,0 +1,124 @@
...
+                {
+                    currentDisplayType = DisplayType.Transparent;
+                }
+#endif
+                return currentDisplayType == DisplayType.Opaque;
+            }
+        }
+
+        /// <inheritdoc />
+        public uint SourceId { get; } = 0;
```
julenka: What does SourceID mean here and why is it always zero

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        /// <summary>
+        /// Is the current camera displaying on an Opaque (AR) device or a VR / immersive device
+        /// </summary>
+        public bool IsOpaque
+        {
+            get
+            {
+                currentDisplayType = DisplayType.Opaque;
```
julenka: it's a little funny that you are changing currentDisplaytype in the getter. What if IsOpaque never gets called? Will currentDisplayType be invalid until IsOpaque is called? Is that a dependency you want to maintain?

Assets/MixedRealityToolkit/Interfaces/CameraSystem/IMixedRealityCameraSystem.cs
julenka: Could you please comment the interface, like what it should be used for so others can understand why it exists?

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -1252,6 +1248,35 @@ public static IMixedRealityBoundarySystem BoundarySystem
...
+                    return null;
+                }
+
+                if (cameraSystem != null)
+                {
+                    return cameraSystem;
+                }
+
+                cameraSystem = Instance.GetService<IMixedRealityCameraSystem>(showLogs: logCameraSystem);
+                // If we found a valid system, then we turn logging back on for the next time we need to search.
```
julenka: Interesting...what bug is this fixing, out of curiosity? I wonder if it could get turned off and then not back on.

julenka:
I pulled this code, tried out the test scene, and created a new scene. I wasn't able to tell any difference in experience from before, so that's great! @Railboy have you had a chance to test this on HoloLens and VR to make sure things look okay?

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
Railboy: I'm not sure, to be honest. This was a copy-paste of the existing methods from IMixedRealityToolkit.cs. I wanted to disrupt things as little as possible.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
Railboy: I'm leaving that up to developers - they're now free to write a camera service that better accommodates AR settings.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
Railboy: It has to do with being a data provider. Core services are expected to be data providers by default, even if they're never expected to provide data. When they don't expect to provide data they set their SourceID to 0. Not sure what motivates this - @davidkline-ms may be able to weigh in.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
Railboy: Not sure - again I'm just duping the registration method I see other core services use. I haven't delved into the logging system yet.

Railboy:
Everything seems fine in VR and HL1 - haven't tested on HL2.

julenka:
@Railboy it looks like building Unity NuGet packages is failing for this PR. I will start a thread with you, Niall and Andrei to learn how to fix this.

julenka:
> @Railboy it looks like building Unity NuGet packages is failing for this PR. I will start a thread with you, Niall and Andrei to learn how to fix this.

It looks like the CI build is currently broken already, so this change is not regressing the unity nuget package generation. I don't see anything in this change that would break nuget packaging, so I'll just complete the PR and we will fix any breaks as part of fixing CI. All other steps passed.

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build?definitionId=15

Issue #4043, opened 2019-04-21
No Active Toolkit Found Dialog shows up during build


genereddick:
## Describe the bug
For awhile now when attempting to build in Unity, it will create and open an empty scene. When the build is complete it will return to the scene that was open when the build started and the temporary scene will vanish. 

Now, in addition to this odd behavior, while the build is ongoing, for this empty scene the MRTK will open up a No Active Toolkit Found dialog (see screenshot below).

This dialog does not automatically close when the build is done and it returns to the original scene. It does not seem to impact anything, it doesn't stop the build. It is just odd / annoying.

## To reproduce
?

## Expected behavior
Assuming the empty scene is some Unity bug, this dialog seems overly aggressive in any case, though maybe it is ncessary? 

## Screenshots
[https://photos.app.goo.gl/iV22FVHHESkjHTio9](https://photos.app.goo.gl/iV22FVHHESkjHTio9)

## Your Setup (please complete the following information)
- Unity Version: 2018.3.12f1
- MRTK Version: latest MRTK_developmernt

## Target Platform (please complete the following information)
- HoloLens
- WMR immersive

julenka:
Hi, I am not hitting these issues when I try to compile a UWP in Unity 2018.3.11f1. Perhaps your build steps are different? Could you please detail your build steps?

For example, here's what I do:

1. File -> Build Settings...
2. Ensure UWP is selected. I usually set the Scripting Backend as .NET.
2. Press "Build" button.

genereddick:
I created a new project this morning from MRTK_development (as of yesterday), and tested with both .NET scripting backend and IL2CPP on 2018.3.13f1 with the same result. During the build Unity creates a new scene ("Untitled" in the screenshot below) and switches to it from the default Sample Scene. 

At which point the MRTK apparently checks the scene and, as there is no Toolkit installed, pops up the dialog. 

When the build is done the temporary scene goes away and switches back to the correct scene. The dialog stays up and has to be closed.

[https://photos.app.goo.gl/gjKvfkc2pcYWyw918](https://photos.app.goo.gl/gjKvfkc2pcYWyw918)

genereddick:
OK, the first issue is Unity related, in that Unity is creating a backup scene in the temp folder 
( /Temp/__Backupscenes/0.backup -- If I delete this it gets recreated during the build) and temporarily redirecting to it during the build before switching back to the actual scene.

[...]

PR #4048, opened 2019-04-22
Disable MRTK scene checker window while build is active

Railboy:
Overview
---
This stops the scene checker window from opening during / immediately after a build as described in #4043

Requests for validation
---
Verify that this fixes the repro case described in #4043

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
+++ b/Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
@@ -29,23 +29,11 @@ namespace Microsoft.MixedReality.Toolkit.Utilities
 
         private static void SceneOpened(Scene scene, OpenSceneMode mode)
         {
-            // If building, don't perform this check
-            if (BuildPipeline.isBuildingPlayer)
-            {
-                return;
-            }
-
             CheckMixedRealityToolkitScene();
         }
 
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
```diff
@@ -29,11 +29,23 @@ static MixedRealityToolkitSceneChecker()
 
         private static void SceneOpened(Scene scene, OpenSceneMode mode)
         {
+            // If building, don't perform this check
+            if (BuildPipeline.isBuildingPlayer)
+            {
+                return;
+            }
+
```
wiwei: Nit: Could also move into CheckMixedRealityToolkitScene to handle two internet birds with one internet stone, unless the goal here is to explicitly only handle SceneOpened/NewSceneCreated (i.e. I could imagine in Unity 2019 maybe there's some other hook we listen to and then we have to fix it again by adding this check to that hook too).

Could in this context means you can go either way and this looks good to me as well

julenka:
I'm still seeing the build window popping up after pulling these changes.

Assets/MixedRealityToolkit/Utilities/MixedRealityToolkitSceneChecker.cs
Railboy: Done.

PR #4049, opened 2019-04-22
Interactable Optimization

Cameron-Micka:
Overview
---
Changing the Interactable system to use the shared material to avoid creating new material instances and breaking batching. Because MaterialPropertyBlocks are used for setting material properties the shared material is used to set the initial state of the MaterialPropertyBlock(s) before mutating state.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableThemeShaderUtils.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableThemeShaderUtils.cs
@@ -98,7 +98,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
         /// <summary>
         /// Grab the shared material to avoid creating new material instances and breaking batching.
-        /// Because MaterialPropertyBlocks are used for setting material properties the shadered material is
+        /// Because MaterialPropertyBlocks are used for setting material properties the shared material is
         /// used to set the initial state of the MaterialPropertyBlock(s) before mutating state.
         /// </summary>
         /// <param name="renderer"></param>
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableThemeShaderUtils.cs
```diff
@@ -107,18 +109,7 @@ public static Material GetValidMaterial(Renderer renderer)
...
-                    material = renderer.sharedMaterial;
-                }
-                else
-                {
-                    material = renderer.material;
-                }
-#else
-                material = renderer.material;
-#endif
+                material = renderer.sharedMaterial;
```
keveleigh: My guess is that this was written the original way because, in the editor while playing, making changes to the `sharedMaterial` propagates those changes back the the material file. Not sure we should update by just removing the original logic.

keveleigh: It's also the Unity recommendation to [not modify materials returned by sharedMaterial](https://docs.unity3d.com/ScriptReference/Renderer-sharedMaterial.html), but as long as the caller of this method doesn't actually modify the material, we might be fine.

keveleigh: Oh though I misread the original version, which didn't use `sharedMaterial` outside the editor.

Cameron-Micka: Yep! All callers of this method only read from the material properties.

PR #4051, opened 2019-04-22
MRTK NuGet with Compiled Binaries

andreiborodin:
This change brings in the work that was done to enable generation of compiled binary NuGet packages. In step with this change, we will update the build pipeline that will also publish packages to a temporary internal feed (for testing).

Shortly thereafter we will be setting up the complete publishing pipeline.


Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+    //                EditorGUILayout.PropertyField(settings.FindProperty("m_SomeString"), new GUIContent("My String"));
+    //            },
+
+    //            // Populate the search keywords to enable smart search filtering and label highlighting:
+    //            keywords = new HashSet<string>(new[] { "Number", "Some String" })
+    //        };
+
+    //        return provider;
+    //    }
+    //}
```
wiwei: Do you know what the plan is for uncommenting this code (i.e. is it better to: 1) delete this commented out code or 2) having the TODO reference a github issue with a description on how to reintroduce this code?)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+
+        private const string YamlPrefix = "%YAML 1.1";
+
+        private static readonly Dictionary<string, string> sourceToOutputFolders = new Dictionary<string, string>
+        {
+            {"Library/PlayerDataCache/WindowsStoreApps", "UAPPlayer" },
+            {"Library/PlayerDataCache/Win", "StandalonePlayer" },
+        };
+
+        private static readonly HashSet<string> ExcludedYamlAssetExtensions = new HashSet<string> { ".jpg", ".csv", ".meta", ".pfx", ".txt", ".nuspec", ".asmdef", ".yml", ".cs", ".md", ".json", ".ttf", ".png", ".shader", ".wav", ".bin", ".gltf", ".glb", ".fbx", ".FBX", ".pdf", ".cginc" };
```
wiwei: >FBX [](start = 264, length = 3)

Nit: Do you know why FBX is the only one that is special cased where both the upper case and lower case ones are done? If we do want to generally handle upper/lower it seems like it would be okay to just store the lower version here and then to-lower the comparison strings down below where things are used.

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+        private static readonly HashSet<string> ExcludedSuffixFromCopy = new HashSet<string>() { ".cs", ".cs.meta", ".asmdef", ".asmdef.meta" };
+
+        private static Dictionary<string, string> nonClassDictionary = new Dictionary<string, string>(); //Guid, FileName
+
+        private const string ScriptFileIdConstant = "11500000";
```
wiwei: >private const string ScriptFileIdConstant = "11500000"; [](start = 8, length = 55)

Nit: Is this an arbitrarily chosen number/constant (or is this dependent on something else?)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+        {
+            try
+            {
+                Debug.Log("Starting to Retarget Assets.");
+                RunRetargetToDLL();
+                Debug.Log("Complete.");
+            }
+            catch (Exception ex)
+            {
+                Debug.LogError("Failed.");
```
wiwei: >Failed. [](start = 32, length = 7)

Nit: "Failed to retarget assets"

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                return false;
+            }
+
+            using (StreamReader reader = new StreamReader(filePath))
+            {
+                return reader.ReadLine().StartsWith(YamlPrefix);
+            }
+        }
+
+        private static Dictionary<string, ClassInformation> ProcessScripts(string[] allFilePaths)
```
wiwei: >private static Dictionary<string, ClassInformation> ProcessScripts(string[] allFilePaths) [](start = 8, length = 89)

Nit: This is a general comment for a number of these functions which return dictionaries (especially because they return general types i.e. string), it would be helpful to have a comment saying what the semantics of the returned values are (i.e. "returns a dictionary of asset full assembly names to ClassInformtion values"

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                    Debug.LogWarning($"Can't find a compiled version of the script: {pair.Key}; guid: {pair.Value.Guid}");
+                }
+            }
+
+            ProcessYAMLAssets(allFilesUnderAssets, Application.dataPath.Replace("Assets", "NuGet/Content"), remapDictionary, asmDefMappings);
+        }
+
+        private static Dictionary<string, string> RetrieveAsmDefGuids(string[] allFiles)
+        {
+            int lengthOfPrefix = Application.dataPath.IndexOf("Assets");
```
wiwei: >int lengthOfPrefix = Application.dataPath.IndexOf("Assets"); [](start = 12, length = 60)

Super nit: This is shared between a couple of places, consider having a helper function/constant so that we don't have to have multiple instances of hardcoded "Assets" strings.

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+            Dictionary<string, string> dllGuids = new Dictionary<string, string>();
+            foreach (string asmdefFile in allFiles.Where(t => t.EndsWith(".asmdef")))
+            {
+                string asmdefText = AssetDatabase.LoadAssetAtPath<AssemblyDefinitionAsset>(asmdefFile.Substring(lengthOfPrefix)).text;
+                string dllName = JsonUtility.FromJson<AssemblyDefinitionStub>(asmdefText).name;
+                string guid = File.ReadAllLines($"{asmdefFile}.meta")[1].Substring(6);
```
wiwei: >string guid = File.ReadAllLines($"{asmdefFile}.meta")[1].Substring(6); [](start = 16, length = 70)

Do you know if this structure of the .meta file will be consistent/solid going forward (i.e. we can take this dependency) or if not, if there's any way we can error out/assert if this isn't actually a valid GUID anymore? i.e. they changed the data structure now and the guid is on line 3 instead of line 2

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: I will delete the code.

---
In reply to: [277821264](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277821264) [](ancestors = 277821264)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                        guidBuilder.Append('0');
+                        break;
+                    case '9':
+                        guidBuilder.Append('a');
+                        break;
+                    default:
+                        guidBuilder.Append((char)(hexit + 1));
+                        break;
+                }
+            }
```
wiwei: Do you know if we can depend on the guids in the meta file to always be lowercase? If there's a uppercase F I think we'll just go to G.

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: Good point, will do that.

---
In reply to: [277822172](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277822172) [](ancestors = 277822172)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                        metaFileContent = File.ReadAllLines(uapMetaFile.FullName);
+                        isUAP = true;
+                    }
+
+                    FileInfo[] files = directory.GetFiles("*.dll", SearchOption.TopDirectoryOnly);
+                    string metaFilePath = string.Empty;
+                    foreach (FileInfo file in files)
+                    {
+                        //Editor is guid + 1
+                        //Standalone is guid + 2
```
wiwei: >//Standalone is guid + 2 [](start = 24, length = 24)

Is it the case that the editor guide + 1 case has already been done (i.e. it's already been +1ed before entering this function? My confusion comes from seeing comments that say there can be up to 3 cycles applied, but there's only two calls to CycleGuidForward here.

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: Added comment, it's the Unity defined script file id.

---
In reply to: [277822892](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277822892) [](ancestors = 277822892)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                    {
+                        File.Copy(filePath, targetPath);
+                    }
+                }
+            }
+
+            foreach (var extension in foundNonYamlExtensions)
+            {
+                Debug.Log($"Not a YAML extension: {extension}");
+            }
```
wiwei: Do you know if the point of this logging was mostly for development of this code, or does this debug output help with long term debuggability as well? It seems like given there's a .cs (and some other somewhat common extensions) this would probably log a bunch of lines.

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: Yeah I know, but it's a minor in terms of time, and I would rather keep it here for easier code understanding/readability as to how the value is gotten.

---
In reply to: [277826520](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277826520) [](ancestors = 277826520)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: This can only come with Unity upgrades, at which time we will test all this; but I will do a TryParse to validate we did get a guid indeed.

---
In reply to: [277827565](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277827565) [](ancestors = 277827565)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                string dllName = JsonUtility.FromJson<AssemblyDefinitionStub>(asmdefText).name;
+                string guid = File.ReadAllLines($"{asmdefFile}.meta")[1].Substring(6);
+                guid = CycleGuidForward(guid);
+                dllGuids.Add($"{dllName}.dll", guid);
+            }
+
+            return dllGuids;
+        }
+
+        private static void ProcessYAMLAssets(string[] allFilePaths, string outputDirectory, Dictionary<string, Tuple<string, long>> remapDictionary, Dictionary<string, string> dllGuids)
```
wiwei: >private static void ProcessYAMLAssets(string[] allFilePaths, string outputDirectory, Dictionary<string, Tuple<string, long>> remapDictionary, Dictionary<string, string> dllGuids) [](start = 8, length = 178)

Nit: It would be helpful to have a comment about the incoming dictionary types (i.e. remapDictionary is a map of what looks to be raw GUIDs i.e. non-cycled to file IDs, similar thing for the dllGuids).

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: The way I have it setup is that I assume all assets are YAML but the ones in ExcludedYamlAssetExtensions; and then I double check by opening the file and looking for the header. If the header is missing, it isn't YAML and we should add that extension to ExcludedYamlAssetExtensions.

YAML can have arbitrary named extensions.

---
In reply to: [277830136](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277830136) [](ancestors = 277830136)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                            Match regexResults = Regex.Match(line, @"guid:\s*([0-9a-fA-F]*)");
+                            if (!regexResults.Success || regexResults.Groups.Count != 2 || !regexResults.Groups[1].Success || regexResults.Groups[1].Captures.Count != 1)
+                            {
+                                throw new InvalidDataException($"Failed to find the guid in line: {line}.");
+                            }
+
+                            string guid = regexResults.Groups[1].Captures[0].Value;
+                            if (remapDictionary.TryGetValue(guid, out Tuple<string, long> tuple))
+                            {
+                                line = $"  m_Script: {{fileID: {tuple.Item2}, guid: {tuple.Item1}, type: 3}}";
```
wiwei: >{ [](start = 53, length = 1)

Question for my own understanding - is this extra '{{' character designed to trigger some string replacement behavior or something? I see that in the YAML files that are checked in they're just like:

 m_Script: {fileID: 11500000, guid: e534d3d8fbada634785d567d8fe8562c, type: 3}

Instead of 

 m_Script: {{fileID: 11500000, guid: e534d3d8fbada634785d567d8fe8562c, type: 3}}

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                            }
+                            else if (nonClassDictionary.ContainsKey(guid))
+                            {
+                                throw new InvalidDataException($"A script without a class ({nonClassDictionary[guid]}) is being processed.");
+                            }
+                            else
+                            {
+                                // Switch to error later
+                                Debug.LogWarning($"Couldn't find a script remap for {guid} in file: '{filePath}' at line '{lineNum}'.");
```
wiwei: >LogWarning [](start = 38, length = 10)

Nit: Maybe LogError and delete the comment on the line above.

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                foreach (Assembly dll in dlls)
+                {
+                    if (dll.name.Contains("MixedReality"))
+                    {
+                        File.Copy(dll.outputPath, Path.Combine(tmpDirPath, $"{dll.name}.dll"), true);
+                    }
+                }
+
+                // Load these directories
+                // TODO maybe we don't need to
```
wiwei: >// TODO maybe we don't need to [](start = 16, length = 30)

NIt: Should we just delete this TODO?

andreiborodin: Let me try without it :)

---
In reply to: [277834891](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277834891) [](ancestors = 277834891)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: I'll make it lower for here.

---
In reply to: [277828536](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277828536) [](ancestors = 277828536)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: Yes, it has been done when we copied editor dlls, will update.

---
In reply to: [277829124](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277829124) [](ancestors = 277829124)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: It is the escape string for string format done through $. The actual string representation is a single '{'.

---
In reply to: [277833931](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277833931) [](ancestors = 277833931)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: Not yet, this will give lot's of false negatives still.

---
In reply to: [277834376](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277834376) [](ancestors = 277834376)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+            RecursiveFolderCleanup(outputDirectory);
+            CopyIntermediaryAssemblies(Application.dataPath.Replace("Assets", "NuGet/Plugins"));
+            UpdateMetaFiles(dllGuids);
+        }
+
+        private static void CopyIntermediaryAssemblies(string outputPath)
+        {
+            List<DirectoryInfo> outputDirectories = new List<DirectoryInfo>();
+            string playerDataCachePath = Application.dataPath.Replace("Assets", "Library/PlayerDataCache");
+            DirectoryInfo playerDataCahce = new DirectoryInfo(playerDataCachePath);
```
wiwei: >playerDataCahce [](start = 26, length = 15)

Nit playerDataCache (but also I think that this isn't used, so maybe this line should just be deleted

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
andreiborodin: Will remove Todo keep the refresh.

---
In reply to: [277835987](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277835987) [](ancestors = 277835987,277834891)

Assets/MixedRealityToolkit/Utilities/AssetScriptReferenceRetargeter.cs
```diff
@@ -0,0 +1,593 @@
...
+                            {
+                                foundMatch = true;
+                                break;
+                            }
+                        }
+                    }
+                    if (!foundMatch)
+                    {
+                        file.Delete();
+                    }
```
wiwei: Nit: I think we can just use a single if (!foundMatch) scope

andreiborodin: Actually we can't, but I will update it

---
In reply to: [277841425](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4051#discussion_r277841425) [](ancestors = 277841425)

Issue #4026, opened 2019-04-19
Input simulation doesn't work if speech is not supported on dev machine


matkrzyztt:
## Describe the bug

Camera controls (rotation, movement) is not working in editor despite Emulation Mode being set no "none".

## To reproduce

1. Create an empty project in unity 2018.3.12
2. Change platform to UWP
3. Enable VR with Mixed Reality Toolkit
4. Import MRTK v2 Core
5. Click "Add to scene and configure"
6. Launch project
7. Hold right mouse button and move 

## Expected behavior

Instead of camera rotating, nothing happens. 

## Your Setup (please complete the following information)

Windows 10 Pro
Unity 2018.3.x (not working on any version)
MRTK v2 RC

## Additional context

This is however working on my home machine. The only difference is my work computer (the one it's not working on) I don't have speech recognition, and thus get an error on play.

> UnityException: Speech recognition is not supported on this machine.
> UnityEngine.Windows.Speech.KeywordRecognizer..ctor (System.String[] keywords, UnityEngine.Windows.Speech.ConfidenceLevel minimumConfidence) (at C:/buildslave/unity/build/Runtime/Export/Windows/Speech.cs:221)
> Microsoft.MixedReality.Toolkit.Windows.Input.WindowsSpeechInputProvider.Enable () (at Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs:95)
> Microsoft.MixedReality.Toolkit.MixedRealityToolkit.RegisterServiceInternal (System.Type interfaceType, Microsoft.MixedReality.Toolkit.IMixedRealityService serviceInstance) (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:843)
> Microsoft.MixedReality.Toolkit.MixedRealityToolkit.RegisterServiceInternal[T] (T serviceInstance) (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:858)
> Microsoft.MixedReality.Toolkit.MixedRealityToolkit.RegisterService[T] (System.Type concreteType, Microsoft.MixedReality.Toolkit.Utilities.SupportedPlatforms supportedPlatforms, System.Object[] args) (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:199)
[...]

PR #4054, opened 2019-04-23
Catch Unity exception in speech input provider.

lukastoenneMS:
This is necessary so other providers can still be enabled after the speech input provider fails.

Overview
---

Changes
---
- Fixes: #4026 .

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
@@ -96,9 +96,9 @@ namespace Microsoft.MixedReality.Toolkit.Windows.Input
                 {
                     keywordRecognizer = new KeywordRecognizer(newKeywords, (ConfidenceLevel)RecognitionConfidenceLevel);
                 }
-                catch (UnityException e)
+                catch (UnityException ex)
                 {
-                    Debug.LogError($"Could not start WindowsSpeechInputProvider: {e.Message}");
+                    Debug.LogError($"Failed to start keyword recognizer. Are microphone permissions granted? Exception: {ex}");
                     keywordRecognizer = null;
                     return;
                 }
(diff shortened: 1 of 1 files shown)


lukastoenneMS:
Note that the dictation provider already has a similar try-catch:
https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/4506767641a25fac53809218b7100f6a9b894f43/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs#L114

StephenHodgson:
I don't think this was handled properly.

We def want to prevent the service from being instanced by throwing the exception, but likely what we'd want to do is handle the specific case of the machine not supporting speech better.

StephenHodgson:
Because now you just have an instance of a service running that's not doing anything.

lukastoenneMS:
@StephenHodgson It doesn't look like there is any function in the Unity API to actually verify that KeywordRecognizer or DictationRecognizer will work, before it throws that exception in the ctor.

lukastoenneMS:
That being said, it might be good to have a poll function in services/data providers to determine if they can be enabled, or return a value from Enable() to indicate success/failure gracefully, and destroy failing services.

StephenHodgson:
IMHO, the service should never be created if it fails.

keveleigh:
> It doesn't look like there is any function in the Unity API to actually verify that KeywordRecognizer or DictationRecognizer will work, before it throws that exception in the ctor.

We may be able to use [`PhraseRecognitionSystem.isSupported` for that](https://docs.unity3d.com/ScriptReference/Windows.Speech.PhraseRecognitionSystem-isSupported.html).
Although the dictation recognizer isn't a phrase recognizer (just the keyword recognizer and grammar recognizer are), we may still be able to reason to some degree about speech support in general.
Of course, this is all Windows-specific, so it's not necessarily a solution for checking pre-constructor unless we add something like your solution above for services determining if they can be enabled.

StephenHodgson:
Agreed with @keveleigh  here, I think we should use some windows api's to verify.

Still not sure how we should handle the fact that the service shouldn't be enabled.

Would it create the instance anyway, then destroy itself shortly after?
Ideally we should throw the exception and catch and handle it properly in the place that it's being created.

PR #4056, opened 2019-04-23
Use pointer pose for one handed manipulation in manipulation handler

### LINKED ISSUE #3863: ManipulationHandler should use pointer pose instead of SpatialGrip interaction
ManipulationHandler moves the manipulated object using the first interaction (input) of SpatialGrip type in the controller that is driving the pointer that is driving the handler. We should be using the pointer pose (pointer.Position and .Rotation) instead to isolate us from the details of the controller driving the pointer.

An immediate consequence of the current code is that we can't manipulate an object with an Xbox controller because, even though it is used to drive a pointer in conjunction with gaze, it doesn't have a SpatialGrip interaction.

Hit with rc1_stabilization at commit 3f92b5fbbef98045d378d63963a89d5dc9ee1955

luis-valverde-ms:
Fixes #3863. See issue for background details on the change.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
@@ -501,6 +501,8 @@ namespace Microsoft.MixedReality.Toolkit.UI
             hostTransform.rotation = Quaternion.Lerp(hostTransform.rotation, targetRotationTwoHands, lerpAmount);
             hostTransform.localScale = Vector3.Lerp(hostTransform.localScale, targetScale, lerpAmount);
         }
+        
+        public bool usePointerPose = false;
 
         private void HandleOneHandMoveUpdated()
         {
@@ -508,7 +510,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
             IMixedRealityPointer pointer = pointerIdToPointerMap.Values.First();
 
             var interactionMapping = GetSpatialGripInfoForController(pointer.Controller);
-            if (interactionMapping != null)
+            if (usePointerPose || interactionMapping != null)
             {
                 Quaternion targetRotation = Quaternion.identity;
                 RotateInOneHandType rotateInOneHandType = isNearManipulation ? oneHandRotationModeNear : oneHandRotationModeFar;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -501,80 +491,76 @@ private void HandleTwoHandManipulationUpdated()
...
+                targetRotation = Quaternion.LookRotation(-directionToTarget);
+            }
+            else if (rotateInOneHandType == RotateInOneHandType.FaceAwayFromUser)
+            {
+                Vector3 directionToTarget = hostTransform.position - CameraCache.Main.transform.position;
+                targetRotation = Quaternion.LookRotation(directionToTarget);
+            }
+            else
+            {
+                targetRotation = pointer.Rotation * objectToHandRotation;
```
julenka: Don't you also need to update objectToHandRotation to use the pointer pose?

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -501,80 +491,76 @@ private void HandleTwoHandManipulationUpdated()
...
                 {
-                    if (oneHandRotationModeNear == RotateInOneHandType.RotateAboutGrabPoint)
-                    {
-                        targetPosition = (interactionMapping.PoseData.Rotation * objectToHandTranslation) + interactionMapping.PoseData.Position;
-                    }
-                    else // RotateAboutCenter or DoNotRotateInOneHand
-                    {
-                        targetPosition = objectToHandTranslation + interactionMapping.PoseData.Position;
-                    }
+                    targetPosition = (pointer.Rotation * objectToHandTranslation) + pointer.Position;
```
julenka: Will this correctly use the grip pose rotation instead of the pointer pose rotation?

julenka:
Make sure to check whether HandInteractionExampleScene needs to be updated.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
luis-valverde-ms: I believe it shouldn't matter because we're only interested in how the rotation changes over time, not the absolute value. In practice, it seems to work fine for 6 DOF controllers and near interaction but hand rays don't produce any rotation.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
luis-valverde-ms: It seems that is was already using it:

            // Calculate relative transform from object to hand.
            Quaternion worldToPalmRotation = Quaternion.Inverse(pointer.Rotation);
            objectToHandRotation = worldToPalmRotation * hostTransform.rotation;

PR #4057, opened 2019-04-23
Auto camera creation

Alexees:
Overview
At least I had the problem with existing scenes that had no camera I wanted to test the toolkit in. I found no real reason why not creating a default main camera if there is none instead of throwing errors and screwing up the setup.
Any other setup might just be unique enough that you need to lay hands on anyways.

Changes
Auto creates a default main camera if there's none when adding the toolkit to a scene that has none


Railboy:
>I found no real reason why not creating a default main camera if there is none instead of throwing errors and screwing up the setup.

I agree this should be handled better. We've got a PR open that covers this problem here: #3890

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs
```diff
@@ -99,11 +100,24 @@ public static void CreateMixedRealityToolkitGameObject()
         {
             Selection.activeObject = MixedRealityToolkit.Instance;
             Debug.Assert(MixedRealityToolkit.IsInitialized);
+            AssureCamera();
             var playspace = MixedRealityToolkit.Instance.MixedRealityPlayspace;
             Debug.Assert(playspace != null);
             EditorGUIUtility.PingObject(MixedRealityToolkit.Instance);
         }
 
+        private static void AssureCamera()
```
StephenHodgson: This should be handled in [CameraCache](https://github.com/XRTK/XRTK-Core/pull/94/files?file-filters%5B%5D=.asset&file-filters%5B%5D=No+extension#diff-17eef5d7dd71af91d052c7e5097ab064R23)

Alexees:
#3890 handles this in CameraCache

PR #4058, opened 2019-04-23
Fix the formatting of the configuration guide's "Back to Configuration Profile" messages.

wiwei:
Without a newline after the > line, it was rendering the line as a header, which was super super super super super weird.


wiwei:
To see what this addresses, look at this screenshot:

image

PR #4060, opened 2019-04-23
Fix the reference to the 'Configuring the Spatial Awareness Mesh Observer' page.

wiwei:
This particular entry in the TOC wasn't hooked up even though the underlying page was already created.

### DIFF
--- a/Documentation/toc.yml
+++ b/Documentation/toc.yml
@@ -85,6 +85,9 @@
         href: README_Solver.md     
     - name: Spatial Awareness
       href: SpatialAwareness/SpatialAwarenessGettingStarted.md
+      items:	
+      - name: Configuring the Spatial Awareness Mesh Observer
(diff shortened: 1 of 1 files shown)


davidkline-ms:
That's odd. I recall writing that article...

Documentation/toc.yml
```diff
@@ -85,8 +85,6 @@
         href: README_Solver.md     
     - name: Spatial Awareness
       href: SpatialAwareness/SpatialAwarenessGettingStarted.md
-      items:
-      - name: Configuring the Spatial Awareness Mesh Observer
```
davidkline-ms: looks like the issue is actually a missing href line.

please add:
href: SpatialAwareness/ConfiguringSpatialAwarenessMeshObserver.md

wiwei: Done.

PR #4063, opened 2019-04-24
Standardize on the way we wait for InputSystem validity.

wiwei:
There are multiple ways we wait for InputSystem validity, some of which
check for the current state of the object and some of which use the await directly.

AFAICT it looks like we should have moved to a common helper (i.e. in this PR: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3462/files), so this common-izes this action into the base class

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.MixedReality.Toolkit.Utilities;
+using System.Threading.Tasks;
 using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Input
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
```diff
@@ -13,6 +14,10 @@ public class InputSystemGlobalListener : MonoBehaviour
     {
         private bool lateInitialize = true;
 
+        /// <summary>
+        /// Prefer using EnsureInputSystemValid(), which will only await if the input system isn't
+        /// yet valid.
+        /// </summary>
         protected readonly WaitUntil WaitUntilInputSystemValid = new WaitUntil(() => MixedRealityToolkit.InputSystem != null);
```
luis-valverde-ms: Looks good. Should we mark this one as deprecated so users don't take dependencies on it?

wiwei: Done

wiwei:
I'm waiting on the reporter to verify the fix.

wiwei:
Reporter verified it, thanks!

PR #4070, opened 2019-04-24
FocusProvider not properly handling pointer removal and double exit

Alexees:
Overview
The FocusProvider had a couple of issues with handling focus properly:

- Removing two pointers from a focuser during the same frame fired the exit event once for the first, which resulted in an Interactable to not getting deactivated due to the pointer counting it does itself.
- Disabling a Pointer while gazing on a Gameobject did not execute the right events in the right order.


Alexees:
@luis-valverde-ms ok, this is about it. I think now I covered everything.

luis-valverde-ms:
Thanks @Alexees . I need more detail on the two issues this PR addresses.

> Removing two pointers from a focuser during the same frame fired the exit event once for the first

I can't see in code how that happens. Can you walk me through the execution sequence in the focus provider that leads to this?

> Disabling a Pointer while gazing on a Gameobject did not execute the right events in the right order

Same as the previous one.

In general, always try to create first the bugs that describe each individual issue you are trying to solve. It helps us a lot when it comes to review PRs and prioritize work.

Alexees:
I will try splitting and writing more issues in the future.

> Removing two pointers from a focuser during the same frame fired the exit event once for the first

In the old line 1069, pendingOverallFocusExitSet is checked for the existence of a focuser. Since both pointers have this as their last focuser, the first iterated over will remove it from this list and fire PointerExit. So instead of firing Exit for the last cursor, in this particular case it fires for the first of as many as exited the focuser during this frame.

> Disabling a Pointer while gazing on a Gameobject did not execute the right events in the right order

The initial oder was "OnFocusChange" and then "OnFocusExit", which does not make sense at all. It now uses the same exit logic the regular focusing does, resulting in "OnPreFocusChange", "OnFocusExitOnFocusChange" and then "OnFocusChange".

luis-valverde-ms:
Understood. For the sake of simplicity, I'd keep the current code in UnregisterPointer() just moving the call to RaisePreFocusChanged before the one to RaiseFocusExit. In UpdateFocusedObjects() I would also keep the current code but modifying the focus exit set to be a map that keeps count of the number of exits per game object, so we can use it to decide whether to raise the exit event.

Alexees:
> For the sake of simplicity, I'd keep the current code in UnregisterPointer() just moving the call to RaisePreFocusChanged before the one to RaiseFocusExit
When you say "I'd keep", does that mean I'm still going to change that, or will you do that?

> I would also keep the current code but modifying the focus exit set to be a map that keeps count of the number of exits per game object
Have been there, but since the events require pointers and gameobjects to be passed along, you can't simply count focuser for each. You need to keep track of each pair because over the course of the three pointer loops, you loose who did what for that otherwise.

Alexees:
We could very well remove the pointer info in Enter and Exit because they make no sense in those cases where two events would happen during the same frame. You must not rely on them in there.

Alexees:
@luis-valverde-ms my last comment is 5 days old and nothing happend since then. How do you guys make sure pull request get processed? Is it up to me to ask every now and then, or is such a delay just a matter of more important things being done at the moment?

luis-valverde-ms:
@Alexees , it's totally my fault. I got sidetracked but I should at least have mentioned it here and set an expectation on when I could go back to this. I'll have a look today.

luis-valverde-ms:
So, I'd say in UnregisterPointer() just swap the order of events. In UpdateFocusedObjects(), go ahead with the changes as they are if you're happy with them.

Alexees:
I reverted UnregisterPointer , swapped the calls, but also added the missing OnFocusChange. I also reverted some renames that weren't necessary. Done.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -630,10 +631,12 @@ public bool UnregisterPointer(IMixedRealityPointer pointer)
 
                 if (!objectIsStillFocusedByOtherPointer)
                 {
+                    MixedRealityToolkit.InputSystem?.RaisePreFocusChanged(pointer, unfocusedObject, null);
```
luis-valverde-ms: We should raise pre focus changed/focus changed regardless of the value of objectIsStillFocusedByOtherPointer

luis-valverde-ms:
>         public bool UnregisterPointer(IMixedRealityPointer pointer)

We are not updating FocusReceiverPointers when unregistering a pointer, is that correct? 

---
Refers to: Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:610 in 640a7ce. [](commit_id = 640a7ce8410c3f435f1064e215d54f1e5b67fa8b, deletion_comment = False)

luis-valverde-ms:
@Alexees , can you have a quick look at the changes in https://github.com/luis-valverde-ms/MixedRealityToolkit-Unity/tree/update_focused_objects? That is the way I would keep track of pointer exits. Do you think it would work?

Alexees:
> @Alexees , can you have a quick look at the changes in https://github.com/luis-valverde-ms/MixedRealityToolkit-Unity/tree/update_focused_objects? That is the way I would keep track of pointer exits. Do you think it would work?

The link just leads to your fork. Is it broken?

luis-valverde-ms:
This should be better: https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4136

Alexees:
> @Alexees , can you have a quick look at the changes in https://github.com/luis-valverde-ms/MixedRealityToolkit-Unity/tree/update_focused_objects? That is the way I would keep track of pointer exits. Do you think it would work?

This works except for when there's more than 2 pointers. EnterSets also need to count their enters so that an exiting pointer won't delete the one reference in case of a two pointer enter and one pointer exit during the same frame. This block needs to account for multipointers:

```
foreach (var pointer in pointers)
{
    pendingOverallFocusExitSet.Remove(pointer.CurrentPointerTarget);
    pendingOverallFocusEnterSet.Remove(pointer.PreviousPointerTarget);
}
```

Other than that, you could add an early return in case pendingPointerSpecificFocusChange is empty after the first collection loop. Then, this PullRequest would essentially do the same thing that I did, except for a little different :)

luis-valverde-ms:
> two pointer enter and one pointer exit during the same frame

I don't see how that use case is broken. I would expect no focus enter/exit events to be raised and it seems the code will achieve that. Can give more detail there?

> early return in case pendingPointerSpecificFocusChange is empty

Good idea

> this PullRequest would essentially do the same thing that I did, except for a little different :)

The main difference for me is that I find this version simpler, and I think that keeping things as simple as possible is very important.

> the question if entering or exiting pointer should be part of the enter and exit calls

I agree in principle but enter/exit still report the right pointer when there's just one and client code may have already taken dependencies on that, so that would be a breaking change which we're trying to avoid before the v2 release. If I were to break things I would go all the way and have focus enter/exit being called for every pointer instead, but that's for another discussion.

Alexees:
> If I were to break things I would go all the way and have focus enter/exit being called for every pointer instead, but that's for another discussion
I took part in a long discussion about this and voted for removing preFocusChange and FocusChange for receiving all enter and exit. But the conclusion was to keep all 4 and have enter and exit only be called once... well

> two pointer enter and one pointer exit during the same frame
2 enter and 1 exit should leave 1 enter to happen. But the following code will remove even that, no matter the amount of enters and exits:

```
foreach (var pointer in pointers)
{
    pendingOverallFocusExitSet.Remove(pointer.CurrentPointerTarget);
    pendingOverallFocusEnterSet.Remove(pointer.PreviousPointerTarget);
}
```
> keeping things as simple as possible
That's true. Only because I was maintaining state this got a little out of hand. This is cleaner for sure.

luis-valverde-ms:
> But the conclusion was to keep all 4 and have enter and exit only be called once

If I remember correctly, the conclusion was that we valued more not breaking the current behavior before v2 release than implementing what we both seem to agree could be a better behavior. 

> I would actually expect one enter. 2 enter and 1 exit should leave 1 enter to happen

The way I understand it an enter event, in its current definition, should only be raised if the target has gone from having no focus in the previous frame to having focus in this one. Do you see it differently?

Alexees:
> before v2 release

Ah, I did not get that

> The way I understand it an enter event, in its current definition, should only be raised if the target has gone from having no focus in the previous frame to having focus in this one

Wow, I don't know why I have so much trouble keeping the cases in my head. You're correct. That's would be a change vor V2 if atl all.
Ok, done with this, I think that's it.

Alexees:
@luis-valverde-ms is handling both of these problems seperately in two new PRs

Issue #4052, opened 2019-04-22
Manipulation handler constraints not applied when set in code


provencher:
## Describe the bug

When modifying the constraints of a manipulation handler via code, the default settings are still applied.

## To reproduce

1. Add a manipulation handler to the scene
2. At runtime, try and change the constraints.
3. The manipulation handler will be effectively unchanged.

## Expected behavior

The manipulation handler should adapt to the changed constraints.

## Your Setup (please complete the following information)

- Unity Version 2018.3.8f1
- MRTK Version v2 RC1

## Target Platform (please complete the following information)

- HoloLens 2

chrisfromwork:
Assigning to @julenka and @yoyozilla to find an owner

julenka:
Thanks for the issue! I'll see if I can take a look at it this week.

provencher:
@julenka @Yoyozilla I've done some work tracking this down.

It seems to be a mix of two things.
Changing handle visibility in the bounding box for scaling does not restrict two handed manipulation in the manipulation handler. With that said, the controls do work as expected.

Second, the handle visibility property changes via code do not persist with the default box activation mode. I exposed those in a PR
https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4071

julenka:
Hi @provencher I'm a little confused about how the two things you are mentioning (changing handle visibility and persisting property changes) are related to the issue.

As I understand it, this bug is saying: "if i change manipulationhandler to rotate only about Y axis in code as opposed to asset, the constraint is not actually applied". 

So in other words if I do something like:

```
var mh = GetComponent<ManipulationHandler>();
mh.Rotation.ConstraintOnRotation = XAxisOnly
```

and then try to rotate an object, the object still rotates about all axes.

provencher:
@julenka it's a weird bug.

So the jist is that I was setting which handles the bounding box would show, as your code example pointed out, but when it came time to hover over the box, no handles would show up at all.

I looked through the code, and this is because the focus change function in the bounding box returns before changing handle visibility if the enum I exposed was set to manual, which is the default value. This early return prevents the handles that should be visible from showing, as the code that determines visibility does a check against the show wireframe property, which is set on the focus change function.

Perhaps this exposes a deeper bug, but I didn't want to change core functionality of the box as that could have far reaching consequences. Simply changing the enum I exposed stops the focus change function from returning early, and the wireframe bool is properly set.

[...]

PR #4071, opened 2019-04-24
User/provencher/exposed box activation setting

provencher:
Overview
---
Previously, the box activation mode was set to Manual, which meant that changing handle visibility settings would not be persisted as the pointer focus change listener would prevent handle activation depending on the box activation mode.

Changes
---
- Fixes: # https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/4052
-> The above issue is resolved when setting the activation type to ActivateByProximityAndPointer


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -87,7 +87,7 @@ private enum BoundsCalculationMethod
             Renderers,
             MeshFilters
         }
-        private enum BoundingBoxActivationType
+        public enum BoundingBoxActivationType
```
Jarodshow: Since this is public now shouldn't it get its own file?

StephenHodgson: Agreed

provencher: Can we just expose all these enums then?

I've exposed what I want, but I'd like anything serializable to be exposed.

The issue with making this it's own file as well is that it's only used in BoundingBox. Is there really a need to split that from the source file?

StephenHodgson: Mainly I don't want to have the hassle of writing the `Class.EnumType` every time. 
But it's also a legibility issue and part of the coding guidelines.

StephenHodgson: If it were possible to make the enum a bit more generic, that'd be great too

StephenHodgson: This enum _does_ seem rather useful in more than just bounding box activations

provencher: Well a few thoughts on that.
`Class.EnumType` helps with consistency regarding where the enum type is used. That's actually a pro to me, if it's only used in one spot.

Regarding the re-usability of the enum however, I think we could move it out once it's needed elsewhere. As in, if someone PR's a change that makes use of it elsewhere, then it should be moved to another time. Until then, I think the way it currently is setup is sufficient.

wiwei:
I want to give @luis-valverde-ms a chance to look at this.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
luis-valverde-ms: > Regarding the re-usability of the enum however, I think we could move it out once it's needed elsewhere. As in, if someone PR's a change that makes use of it elsewhere, then it should be moved to another time. Until then, I think the way it currently is setup is sufficient.

+1

luis-valverde-ms:
The intent of the changes looks reasonable to me but the component owners (@cre8ivepark and @johnppella?) should have a look too.

The changes seem to deal with being able to change the activation type at runtime (from code) but the PR description doesn't make it clear. Can we rework that a bit and make it clear too how it solves the connected issue?

provencher:
@luis-valverde-ms this is from the referenced issue 

"So the jist is that I was setting which handles the bounding box would show, as your code example pointed out, but when it came time to hover over the box, no handles would show up at all.

I looked through the code, and this is because the focus change function in the bounding box returns before changing handle visibility if the enum I exposed was set to manual, which is the default value. This early return prevents the handles that should be visible from showing, as the code that determines visibility does a check against the show wireframe property, which is set on the focus change function.

Perhaps this exposes a deeper bug, but I didn't want to change core functionality of the box as that could have far reaching consequences. Simply changing the enum I exposed stops the focus change function from returning early, and the wireframe bool is properly set.

Part of the issue with my bug report is that I conflated manipulation constraints with bounding box handles thinking they were more related.

Perhaps constraints on the manipulation handler should affect which handles the bounding box shows, as the handles permit transformations the manipulation handler might forbid."

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
StephenHodgson: No worries, I understand :)

Just wanted to keep the project consistent and in line with the established patterns and in line with the coding guidelines. Just have to speak up and have it noted for the record.

imho it's better to make this change now than to have a breaking change later when we decide to make that kinda change.

provencher: Yeah that makes sense. I didn't consider it breaking code down the line. It's a tradeoff

wiwei:
I'll followup with @cre8ivepark to see if he can look at this.

PR #4077, opened 2019-04-25
Fix "Buttons can be pressed by pinching when hand is near button, even when hand rays are not present"

julenka:
# Overview
This PR fixes a bug where it is possible to click on an Interactable by air tapping when your finger is right next to a button, even though no hand ray or other far pointer is active. The expected behavior is that pinching near a button when no far pointer is active should not press the button, because no pointer that should respond to Select should be sending OnPointerDown.

The root cause is that the PokePointer is senting OnPointerDown/Up events when a user pinches/unpinches. Additionally, Interactable is listening for OnInputUp and firing click events whenever it detects an input up, including when from a PokePointer.

It is not expected for a PokePointer to send OnPointerDown/Up events in response to a pinch. Instead it should send OnPointerDown/Up when it penetrates a surface.

Additional fixes:
- Update InteractableExample scene to work for near interaction (hand pressing the buttons).
- Update PokePointer so that devs can use Interactable to press buttons without needing PressableButton
- Fix bug where dialog "you must pick MRTK configuration" would pop up every time build completes.

# Review Requests
I would love feedback on this approach, specifically:
- I'm curious why Interactable was handling OnInputDown / OnInputUp as well as OnPointerDown and OnPointerUp. This seems like it would cause duplicate events. Additionally, I couldn't think of a way to not respond to OnInputUp from a PokePointer in Interactable::OnInputUp short of specifically ignoring that type. 
- I'd like to know if people think it's reasonable for the PokePointer to not listen to OnInputDown/Up, or if there is another way.

# Fixes
This fixes an issue that is being tracked here: https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/21333770


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -680,66 +679,6 @@ protected void StartInputTimer(bool isInput = false)
 
         #region MixedRealityInputHandlers
 
-        /// <summary>
-        /// Used for click events for actions not processed by pointer events
-        /// </summary>
-        /// <param name="eventData"></param>
-        public void OnInputUp(InputEventData eventData)
```
julenka: I removed OnInputUp/OnInputDown from interactable because I don't see a case where OnPointerDown / OnPointerUp does not handle this, and because I couldn't think of a way to not respond to OnInputUp from PokePointer without specifically ignoring that type

luis-valverde-ms: Agreed. If it breaks something at least we will understand the use case that is meant to cover.

luis-valverde-ms:
Agreed as well with not listening to InputDown/Up in poke pointer. I would even say that poke pointer shouldn't raise pointer down/up events at all and leave it to the UI to implement their logic based on the pointer position. But that's for another PR.

julenka:
Before I check this in I'd like to test it more thoroughly on HoloLens 2, VR, and in more scenes.

killerantz:
> Just tested this in VR and everything still seems to work as expected.
> 
> I'd like to know if people think it's reasonable for the PokePointer to not listen to OnInputDown/Up, or if there is another way.
> 
> Makes sense to me - @killerantz was having both was some kind of now-unnecessary workaround from an older version of the input system? This issue stirs a dim memory from the early days of this class.

Indeed, in order to get HoloLens, VR and everything else to work as expected, I had to implement OnInputDown and OnInputUp. Then there were issues where pointer and input events firing at the same time, so I added a check to take one event type and ignore the other. I can review in a little bit and see what else we can clean up in the pointer handlers.

julenka:
> > Just tested this in VR and everything still seems to work as expected.
> > I'd like to know if people think it's reasonable for the PokePointer to not listen to OnInputDown/Up, or if there is another way.
> > Makes sense to me - @killerantz was having both was some kind of now-unnecessary workaround from an older version of the input system? This issue stirs a dim memory from the early days of this class.
> 
> Indeed, in order to get HoloLens, VR and everything else to work as expected, I had to implement OnInputDown and OnInputUp. Then there were issues where pointer and input events firing at the same time, so I added a check to take one event type and ignore the other. I can review in a little bit and see what else we can clean up in the pointer handlers.

@killerantz do you remember what bug you had that required using OnInputDown/OnInputUp? I'd like to test that case now to make sure that I didn't break any behavior.

julenka:
I believe this PR will also fix #4129

killerantz:
> Just tested this in VR and everything still seems to work as expected.
> I'd like to know if people think it's reasonable for the PokePointer to not listen to OnInputDown/Up, or if there is another way.
> Makes sense to me - @killerantz was having both was some kind of now-unnecessary workaround from an older version of the input system? This issue stirs a dim memory from the early days of this class.
> 
> Indeed, in order to get HoloLens, VR and everything else to work as expected, I had to implement OnInputDown and OnInputUp. Then there were issues where pointer and input events firing at the same time, so I added a check to take one event type and ignore the other. I can review in a little bit and see what else we can clean up in the pointer handlers.
> 
> @killerantz do you remember what bug you had that required using OnInputDown/OnInputUp? I'd like to test that case now to make sure that I didn't break any behavior.

I could have to do with getting the Global button to work. The ability for the Interactable to listen to global input like a modal or click anywhere. There may be some buttons on the VR controllers that were not working with the previous system. I should be able to set an action that is mapped to an input type or button and that should work.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: Okay, I think I figured out why @killerantz  is listening to OnInputUp. Based on the comments it seems like previously the only way to globally listen for inputs was to listen for OnPointerUp events, and then fire OnClicked. Now, Pointer events are also dispatched globally. Since Interactable is registered as a global listener, it should also be able to receive all global pointer events.

julenka:
> > Just tested this in VR and everything still seems to work as expected.
> > I'd like to know if people think it's reasonable for the PokePointer to not listen to OnInputDown/Up, or if there is another way.
> > Makes sense to me - @killerantz was having both was some kind of now-unnecessary workaround from an older version of the input system? This issue stirs a dim memory from the early days of this class.
> > Indeed, in order to get HoloLens, VR and everything else to work as expected, I had to implement OnInputDown and OnInputUp. Then there were issues where pointer and input events firing at the same time, so I added a check to take one event type and ignore the other. I can review in a little bit and see what else we can clean up in the pointer handlers.
> > @killerantz do you remember what bug you had that required using OnInputDown/OnInputUp? I'd like to test that case now to make sure that I didn't break any behavior.
> 
> I could have to do with getting the Global button to work. The ability for the Interactable to listen to global input like a modal or click anywhere. There may be some buttons on the VR controllers that were not working with the previous system. I should be able to set an action that is mapped to an input type or button and that should work.

Great, Ill make sure to verify that having global interactables work with various different button inputs.

julenka:
@killerantz  The following test is working for me, with my changes to remove OnInputUp listener in Interactable.

1. Add a new Input Action "JuliaTest"
2. Update Xbox controller mapping to map "Y" to "JuliaTest"
3. Create interactable that is a global listener, listening for new action JuliaTest

image

The interactable reacts as expected.

julenka:
> Before I check this in I'd like to test it more thoroughly on HoloLens 2, VR, and in more scenes. This is why I didn't complete the PR :-)

Verified works on HoloLens, and @Railboy verified works on VR. Thank you!

provencher:
@julenka this fix makes the issue worse than how it was previously.

I opened another issue again.
https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/4170

As I mentioned in the original issue, the issue gets resolved by reordering the poke and grab pointers.

killerantz:
Without the IMixedRealityInputHandlers, we do not get events for other controller buttons anymore, like the Menu or Grip press.

julenka:
> Without the IMixedRealityInputHandlers, we do not get events for other controller buttons anymore, like the Menu or Grip press.

Interesting, if you read my comment above, you'll see that I tested what I thought was the case you had raised. Can you provide an example that is breaking for you?:

@killerantz  The following test is working for me, with my changes to remove OnInputUp listener in Interactable.

1. Add a new Input Action "JuliaTest"
2. Update Xbox controller mapping to map "Y" to "JuliaTest"
3. Create interactable that is a global listener, listening for new action JuliaTest

image

The interactable reacts as expected.

killerantz:
Using the existing actions (Menu and Grip Press) and the MR controllers I get nothing, even with IsGlobal set to true or false. I wonder way the Xbox controller is working and not the MR controller. WIth the input handlers added, I can get down and up events for those buttons actions on the MR controller.

image

image

julenka:
> Using the existing actions (Menu and Grip Press) and the MR controllers I get nothing, even with IsGlobal set to true or false. I wonder way the Xbox controller is working and not the MR controller. WIth the input handlers added, I can get down and up events for those buttons actions on the MR controller.
> 
> image
> 
> image

Thank you @killerantz for reporting this. I think this may be a bug, regression from this commit. Could you please open an issue so that we can track reproducing and fixing this? Hoping to get this on the books for the next release.

killerantz:
I created issue #4492

Issue #4080, opened 2019-04-25
AnimatedCursor jumps back and forth infront of the MotionController


Alexees:
## Describe the bug
Just pointing into nowhere results in the Cursor switching position like crazy, bewteen linerenderer tip and somewhere along that line.

## To reproduce

Just watch your UWP Motion Controller cursor

## Expected behavior

Pointer is stable at the linerenderer tip

## Your Setup (please complete the following information)

- UNITY Version 2018.3.13f1
- MRTK Version v2.0

## Target Platform (please complete the following information)

- WMR immersive

Alexees:
The issue was that there was no proper case for distances longer than the raysteps total length. And in case of an exact same lenght, floating point precision coused the raysteps length to sometimes be shorter than the distance.

PR #4081, opened 2019-04-25
Mrtk Raystep Fix

Alexees:
- Fixes: #4080 .


Railboy:
Thank you this was bugging me.

Assets/MixedRealityToolkit/Definitions/Physics/RayStep.cs
```diff
@@ -155,33 +155,33 @@ public static Vector3 GetPointByDistance(RayStep[] steps, float distance)
...
+            Debug.Assert(steps != null && steps.Length > 0);
 
-            int numSteps = steps.Length;
+            float traveledDistance = 0;
             float stepLength = 0;
+            RayStep currentStep = default;
+
 
-            for (int i = 0; i < numSteps; i++)
+            foreach (var step in steps)
```
Railboy: Looks good. Works as expected.

julenka:
azp /run mrtk_pr

julenka:
Thank you for the fix! Next time, it would be great if you could explain in the PR the bug that you are fixing and the changes you made so we don't need to click through to the bug.

Alexees:
@julenka I thought it was sufficient to link the issue, which explains the bug. I guess instead of commenting the issue with what's actually going on, I put that into the PR next time.

julenka:
Hey @Alexees actually, looking more at what e.g. @luis-valverde-ms  is doing, it looks like we are following patterns that we do just refer to bugs, if the bugs are detailed enough. So, no worries, I think what you did here was fine.

PR #4085, opened 2019-04-25
Add definitions of AzurePipelines builds

macborow:
We should switch to YAML build definitions to leverage the benefits of having "configuration-as-code" (as previously discussed in an internal email thread and offline). 

This change adds YAML configuration files that correspond to the existing MRTK_CI and MRTK_PR jobs.
Switching from visual designer to YAML will require setting up new builds (it should be straightforward although we will lose build history and we need to pay close attention to setting up triggers, etc.) and I can take care of it afterwards.
The CI config references a new variable "NuGetFeedId" (which is a GUID of the feed that we use to publish the packages internally).

I verified the set up by creating personal test builds on my fork of the repo:
https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=1292
https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=1290

### DIFF
--- a/pipelines/ci.yaml
+++ b/pipelines/ci.yaml
@@ -1,0 +1,197 @@
+pool:
+  name: On-Prem Unity
+
+steps:
+- powershell: |
+   # Find unity.exe as Start-UnityEditor currently doesn't support arbitrary parameters
+   $editor = Get-ChildItem ${Env:Unity2018.3.7f1} -Filter 'Unity.exe' -Recurse | Select-Object -First 1 -ExpandProperty FullName
+   
+   $outDir = "$(Build.ArtifactStagingDirectory)\build"
+   $logFile = New-Item -Path "$outDir\build\build_x86.log" -ItemType File -Force
+   
+   $sceneList = "Assets\MixedRealityToolkit.Examples\Demos\HandTracking\Scenes\HandInteractionExamples.unity"
+   
+   $proc = Start-Process -FilePath "$editor" -ArgumentList "-projectPath $(Get-Location) -executeMethod Microsoft.MixedReality.Toolkit.Build.Editor.UnityPlayerBuildTools.StartCommandLineBuild -sceneList $sceneList -logFile $($logFile.FullName) -batchMode -x86 -buildOutput $outDir -buildTarget WSAPlayer -buildAppx" -PassThru
+   $ljob = Start-Job -ScriptBlock { param($log) Get-Content "$log" -Wait } -ArgumentList $logFile.FullName
+   
+   while (-not $proc.HasExited -and $ljob.HasMoreData)
+   {
+       Receive-Job $ljob
+       Start-Sleep -Milliseconds 200
+   }
+   Receive-Job $ljob
+   
+   Stop-Job $ljob
+   
+   Remove-Job $ljob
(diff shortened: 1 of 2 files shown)


pipelines/ci.yaml
```diff
@@ -0,0 +1,197 @@
...
+   $logFile = New-Item -Path "$outDir\build\build_x86.log" -ItemType File -Force
+   
+   $sceneList = "Assets\MixedRealityToolkit.Examples\Demos\HandTracking\Scenes\HandInteractionExamples.unity"
+   
+   $proc = Start-Process -FilePath "$editor" -ArgumentList "-projectPath $(Get-Location) -executeMethod Microsoft.MixedReality.Toolkit.Build.Editor.UnityPlayerBuildTools.StartCommandLineBuild -sceneList $sceneList -logFile $($logFile.FullName) -batchMode -x86 -buildOutput $outDir -buildTarget WSAPlayer -buildAppx" -PassThru
+   $ljob = Start-Job -ScriptBlock { param($log) Get-Content "$log" -Wait } -ArgumentList $logFile.FullName
+   
+   while (-not $proc.HasExited -and $ljob.HasMoreData)
+   {
+       Receive-Job $ljob
```
wiwei: Is it possible to use yaml templates to avoid some of the duplication in these files:

https://github.com/Microsoft/azure-pipelines-agent/issues/1686

i.e. I started out wondering about if we could share like "{Env:Unity2018.3.7f1}" but then noticed there were larger swaths of things that we could possibly template-ize.

Or if this isn't possible, I guess it would be good for our collective knowledge to know if this isn't possible and if so for what reason.

pipelines/ci.yaml
```diff
@@ -0,0 +1,197 @@
...
+   Stop-Job $ljob
+   
+   Remove-Job $ljob
+   Stop-Process $proc
+  displayName: 'Build UWP ARM'
+
+- task: PublishBuildArtifacts@1
+  displayName: 'Publish ARM'
+  inputs:
+    PathtoPublish: '$(Build.ArtifactStagingDirectory)\build\AppPackages\MixedRealityToolkit\MixedRealityToolkit_2.0.0.0_ARM_Master_Test'
```
wiwei: It's interesting seeing a hardcoded version number here, but that also makes me wonder if this is even necessary (i.e. for this validation does the version # even matter?). I assume that whatever our publish/step would handle appending the right version at that point.

macborow: This is a first draft - I basically took the existing pipelines and committed with minimum clean up. There's definitely a lot we can and should do with templates/variables but I haven't looked into it yet.

wiwei:
Per the other comment, if the goal of this review is to be a copy over the existing config (but YAML-ified) then this is good - looking at mrtk_CR/mrtk_PR and this they look to have the same stages and output.

pipelines/ci.yaml
wiwei: Per your other comment, feel free to resolve this as work to do in the future. I didn't understand the scope of what this review was for (i.e. UI -> YAML), where the next step is YAML -> better YAML.

Issue #4084, opened 2019-04-25
Misordered MixedRealityInputModule and OnSourceDetected invocation


wiwei:
We've seen a couple of instances now where the MixedRealityInputModule and various providers (i.e. GazeProvider) are getting initialized in non-deterministic fashion. If the GazeProvider starts first, it will raise its source detected, and then when MixedRealityInputModule::ActivateModule is invoked, it re-invokes OnSourceDetected, which tries to add the same pointer to the pointer dictionary.

This throws an exception (bad). We want to update the OnSourceDetected function to only try to add it to the dictionary if it doesn't exist, but at the same time we don't want to lose track of what appears to be a relatively interesting (and possibly with other ramifications) issue.

wiwei:
Okay, it turns out we figured out what's going on here - the change https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4063 caused this OnSourceDetected issue to bubble up because GazeProvider's await was no longer yielding execution.

It turns out that GazeProvider.cs will run first (Start) and previous to my change it would micro-sleep (enough for MixedRealityInputModule::ActivateModule to come in and do its work).

With my change, since GazeProvider no longer micro-sleeps, we suddenly are actually hit by the mis-ordered startup of these two objects.

wiwei:
I'm going to close this out - it looks like the code in ActivateModule probably doesn't need to be re-doing source detected, but at the same time we don't have any test coverage on this (i.e. doing this could break things in inadvertent ways) and the benefit here is purely cleanup (i.e. there's no end-user/consumer benefit)

tomyeti:
Also had this issue. Is it pretty safe to just not add the new source if the key already exists? or should we replace the item in the dictionary with the new source? 

Interestingly enough, in our builds this issue only ever occurred after the first playthrough, every time after that the input was broken.

PR #4086, opened 2019-04-25
Workaround for mis-ordered initialization of some input providers.

wiwei:
We've seen a couple of instances now where the MixedRealityInputModule and various providers (i.e. GazeProvider) are getting initialized in non-deterministic fashion. If the GazeProvider starts first, it will raise its source detected, and then when MixedRealityInputModule::ActivateModule is invoked, it re-invokes OnSourceDetected, which tries to add the same pointer to the pointer dictionary.

This throws an exception (bad). We want to update the OnSourceDetected function to only try to add it to the dictionary if it doesn't exist to work around this and unblock a few consumers.

https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/4084


wiwei:
Waiting on reporter to validate this change.

wiwei:
Because none of us have an actual repro of the issue.

wiwei:
Okay, managed to actually hit this pretty reliably, this works good.

PR #4087, opened 2019-04-25
Delete ControllerElement.cs.meta

keveleigh:
Overview
---
The script was deleted in #4076, but the meta lived on.


wiwei:
Oh my.

PR #4088, opened 2019-04-25
Fix using MRTK in Unity editors on non-Windows platforms (part 1)

keveleigh:
Overview
---
Reported on Slack: opening the MRTK caused issues on non-Windows platforms (in this case, Mac).
This PR fixes most of them, but I'll need to get my hands on a Mac to repro the glTF-specific ones.

The fix: I moved the methods required by the interfaces outside of the Windows-platform `#if`s, where possible, or created a second version for non-Windows platforms where not possible.

Report: https://holodevelopers.slack.com/archives/C2H4HT858/p1556141818234600

image
image

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
@@ -3,11 +3,11 @@
 
 using Microsoft.MixedReality.Toolkit.Input;
 using Microsoft.MixedReality.Toolkit.Utilities;
-using System.Text;
 using System.Threading.Tasks;
 using UnityEngine;
 
 #if UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
+using System.Text;
 using UnityEngine.Windows.Speech;
 #endif // UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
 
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
```diff
@@ -41,6 +41,124 @@ public class WindowsDictationInputProvider : BaseInputDeviceManager, IMixedReali
...
+        }
+
+        /// <inheritdoc />
+        public async void StopRecording()
+        {
+            await StopRecordingAsync();
+        }
+
+#if UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
+        /// <inheritdoc />
```
andreiborodin: I would scope these tighter, inside the function; at least to avoid two function declarations, as when the signature changes we will hit this issue again if the second is forgotten to be updated.

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
```diff
@@ -50,15 +50,41 @@ public class WindowsSpeechInputProvider : BaseInputDeviceManager, IMixedRealityS
...
-
         /// <inheritdoc />
         public bool IsRecognitionActive
         {
+#if UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
             get { return keywordRecognizer != null && keywordRecognizer.IsRunning; }
+#else
+            get { return false; }
+#endif
+        }
```
andreiborodin: Could do ;)
public bool IsRecognitionActive =>
#if UNITY_STANDALONE_WIN || UNITY_WSA || UNITY_EDITOR_WIN
        keywordRecognizer?.IsRunning ??
#endif
        false;

andreiborodin:
In general, I like to scope the #if to as tight as possible, this way it eliminates issues when we change an API of another class that this class uses. Although in this case the majority of devs are on Windows, and we shouldn't have the issue, still a good practice.

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
keveleigh: That was how I had it originally, to match the SpeechInputProvider, but I couldn't think of another way to resolve this warning without just defining a non-async version in the other function declaration: 
image

keveleigh: Ah, I suppose I could define a no-op await or something.

StephenHodgson:
I fixed this in my fork by making a base class for the dictation providers.

Then made the interface methods virtual and overrode them in the platform specific implementations.

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsDictationInputProvider.cs
keveleigh: Updated to tighten the scope.

wiwei:
By the way, was this verified directly on mac hardware or using that win->osx compiling extension thing?

keveleigh:
This one will only repro directly on mac hardware.
I haven't had access to a mac yet, but I reproduced the original issue with `#if`s (since those were causing the problem in the first place).

PR #4099, opened 2019-04-26
Introduce service registry, static playspace and update boundary demo to use the registry

davidkline-ms:
This is part 4 of the changes for #3545.

This change does the following:
- Implements an adapted version of @Alexees' static playspace PR (#3624) and removes the need for passing the playspace transform to services
- Adds a static MixedRealityServicesRegistry class and registers each core or extension service (data providers are not part of the registry)
- Update the boundary demo to use the registry to acquire the instance of the boundary system

NOTE: This change does _NOT_ break the ability for application code to call MixedRealityToolkit.<service> to acquire the service instance (presuming the MRTK object is in the scene).

Other services and demos will be updated to use the new registry pattern over the next few PRs.


Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityBoundaryVisualizationProfileInspector.cs
```diff
@@ -94,12 +94,6 @@ public override void OnInspectorGUI()
             EditorGUILayout.Space();
             EditorGUILayout.LabelField("Boundary Visualization Options", EditorStyles.boldLabel);
             EditorGUILayout.HelpBox("Boundary visualizations can help users stay oriented and comfortable in the experience.", MessageType.Info);
-            // Boundary settings depend on the experience scale
-            if (MixedRealityToolkit.Instance.ActiveProfile.TargetExperienceScale != ExperienceScale.Room)
-            {
-                EditorGUILayout.Space();
-                EditorGUILayout.HelpBox("Boundary visualization is only supported in Room scale experiences.", MessageType.Warning);
-            }
```
wiwei: Will a future change re-introduce this warning (it seems helpful), or is this something that will be too difficult to support going forward?

davidkline-ms: Good question. Something similar may be introduced elsewhere (ex: in the main profile when selecting the experience scale) that better describes the expected behaviors.

davidkline-ms: Part of the intent with removing this warning is to avoid reaching across profiles for information (better separation)

wiwei:
I'm working through a build call of the stabilization branch right now, and will take an in-depth look at this in the afternoon.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -0,0 +1,130 @@
...
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public static class MixedRealityPlayspace
```
Railboy: I like this new class a lot.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        private const string Name = "MixedRealityPlayspace";
+
+        private static Transform mixedRealityPlayspace;
+
+        /// <summary>
+        ///
```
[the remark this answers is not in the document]
davidkline-ms: good catch! i can add summaries.

davidkline-ms:
> I'm fuzzy on the ultimate fate / purpose of MixedRealityToolki.ActiveSystems / MixedRealityToolkit.RegisteredMixedRealityServices but it sounds like that will be cleared up in upcoming PRs.

Those will remain for the near future as one of the primary goals was to leave the MRTK object's functionality alone as much as possible. Very open to thoughts, however.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
davidkline-ms: You can thank @Alexees for most of it! I just did some minor reformatting and tweaking.

Railboy:
>Those will remain for the near future as one of the primary goals was to leave the MRTK object's functionality alone as much as possible. Very open to thoughts, however.

If we want to avoid a muddled registration story without blowing stuff up this might work:

```
public class MixedRealityToolkit : MonoBehaviour, IMixedRealityServiceRegistrar
{
     [Obsolete("Use MixedRealityServiceRegistery.ActiveSystems.")]
     public IReadOnlyDictionary<Type, IMixedRealityService> ActiveSystems => MixedRealityServiceRegistry.ActiveSystems;

     [Obsolete("Use MixedRealityServiceRegistery.RegisteredMixedRealityServices.")]
     public IReadOnlyList<Tuple<Type, IMixedRealityService>> RegisteredMixedRealityServices => MixedRealityServiceRegistry.RegisteredMixedRealityServices;

     ...

    [Obsolete("Use MixedRealityServiceRegistery.RegisterService directly.")]
    public bool RegisterService<T>(
        Type concreteType,
        SupportedPlatforms supportedPlatforms = (SupportedPlatforms)(-1),
        params object[] args) where T : IMixedRealityService
    {
        MixedRealityServiceRegistery.RegisterService(concreteType, supportedPlatforms, args);
    }

}
```

So the ONLY place services are actually registered / stored is in the MixedRealityServiceRegistry class.

And the ONLY thing the MixedRealityToolkit instance would be responsible for is relaying MonoBehaviour events to the registered services (which is the one thing a static class can't do).

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -233,6 +233,11 @@ public void ResetConfiguration(MixedRealityToolkitConfigurationProfile profile)
             if (registeredMixedRealityServices.Contains(registryInstance))
             {
                 registeredMixedRealityServices.Remove(registryInstance);
+                if (!(serviceInstance is IMixedRealityDataProvider))
+                {
+                    // Only remove IMixedRealityService or IMixedRealityExtensionService (not IMixedRealityDataProvider)
```
wiwei: >// Only remove IMixedRealityService or IMixedRealityExtensionService (not IMixedRealityDataProvider) [](start = 20, length = 100)

Nit: Can you update this comment to explain the 'why' part (instead of the 'what') - it is clear from the code that it's not doing things to data providers (this is what the comment also says) but what's actually valuable is to leave a note about why this is being done.

wiwei:
>                 activeSystems.Add(interfaceType, serviceInstance);

Is there going to be a corresponding RegisterService<> call here as well (to mirror the UnregisterService change that's below or above I forget which direction)

---
Refers to: Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:794 in d821506. [](commit_id = d8215063910828983f1d4ba67af0b364dfdbc816, deletion_comment = False)

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
Railboy: Thanks @Alexees

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// </returns>
+        public static bool AddService<T>(T serviceInstance, IMixedRealityServiceRegistrar registrar) where T : IMixedRealityService
+        {
+            Type interfaceType = typeof(T);
+            T existingService;
+
+            if (TryGetService<T>(out existingService, serviceInstance.Name))
+            {
+                return false;
+            }
```
wiwei: Do we want to depend on callers to filter out dataproviders from coming here? i.e. I see MixedRealityToolkit doing things where it filters out data providers, but if this is something we want to prevent from ever happening it would probably be good to enforce at a lower level.

davidkline-ms: Good call!

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -865,7 +822,17 @@ private bool RegisterServiceInternal(Type interfaceType, IMixedRealityService se
         private bool RegisterServiceInternal<T>(T serviceInstance) where T : IMixedRealityService
```
wiwei: >RegisterServiceInternal [](start = 21, length = 23)

Can we have some comments about why only this flavor of RegisterServiceInternal is also calling MixedRealityServiceRegistry.AddService<T>(serviceInstance, this)?

For example, there are comments saying that core services are always removable but I see them get added like:


#region Services Initialization

            var orderedCoreSystems = activeSystems.OrderBy(m => m.Value.Priority).ToArray();
            activeSystems.Clear();

            foreach (var system in orderedCoreSystems)
            {
                RegisterServiceInternal(system.Key, system.Value);
            }

            var orderedServices = registeredMixedRealityServices.OrderBy(service => service.Item2.Priority).ToArray();
            registeredMixedRealityServices.Clear();

            foreach (var service in orderedServices)
            {
                RegisterServiceInternal(service.Item1, service.Item2);
            }

            InitializeAllServices();


And these wouldn't get added to the registry because they don't call this overload of RegisterServiceInternal.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: Sounds good. I will clarify the comment when i move the check to AddService

davidkline-ms:
> So the ONLY place services are actually registered / stored is in the MixedRealityServiceRegistry class.

I like that. How about once we get all services updated, we revisit this. @Railboy, Can you file an issue to track the work?

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -0,0 +1,130 @@
...
+            transform.SetParent(Transform);
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public static void DetachChildren()
+        {
+            Transform.DetachChildren();
+        }
```
wiwei: The API surface exposed here feels a bit interesting - there's a public one to add a specific child (i.e. as a consumer I can invoke that to add a specific child) but there's no way for me to remove that child without totally messing up all other callers (i.e. I can't remove just my child, I have to remove everything).

I'm guessing that maybe this could be cleared up with a comment, or perhaps is coming in the next PR, but this API assymetry should be cleared up.

davidkline-ms: This is based, a bit on Unity's transform API. there is a RemoveChildAt(index), but no RemoveChild(transform).

@WiWei, can you file an issue to add better symetry?

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,222 @@
...
+    /// <summary>
+    /// Static class that represents the Mixed Reality Toolkit service registry.
+    /// </summary>
+    /// <remarks>
+    /// The service registry is used to enable discovery of and access to active Mixed Reality Toolkit services at
+    /// runtime without requiring direct code reference to a singleton style component.
+    /// </remarks>
+    public static class MixedRealityServiceRegistry
+    {
+        // todo: is this the data type we want?
```
wiwei: > // todo: is this the data type we want? [](start = 7, length = 40)

Nit: What's the resolution on this TODO?

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: you bet!

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
wiwei: I guess what I don't understand is, is my question even a problem (i.e. will Detach children ever be called by multiple people, or just a single person managing some greater lifecycle?) I'm trying to understand the calling mechanics of this function.

---
In reply to: [279107558](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4099#discussion_r279107558) [](ancestors = 279107558)

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// <summary>
+        /// The service registry store where the key is the Type of the service interface and the value is
+        /// a pair in which they key is the service instance and the value is the regisrar instance.
```
wiwei: >regisrar [](start = 82, length = 8)

registrar

wiwei: >pair [](start = 14, length = 4)

I think this is a list of pairs - TBH it feels like this datastructure is interesting enough to have its own class abstraction. Anytime you start going 3 layers deep (i.e. a hashtable of a list of pairs) it probably means it should be abstracted a little different.

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,222 @@
...
+
+        /// <summary>
+        /// Removes an <see cref="IMixedRealityService"/> instance from the registry.
+        /// </summary>
+        /// <typeparam name="T">The interface type of the service being removed.</typeparam>
+        /// <param name="serviceInstance">Instance of the service to remove.</param>
+        /// <returns>
+        /// True if the service was successfully removed, false otherwise.
+        /// </returns>
+        public static bool RemoveService<T>(T serviceInstance) where T : IMixedRealityService
```
wiwei: As a caller, why would I need both of these flavors if it looks like the second one does things automatically for me?) Is it just for effiicency (i.e. some callers know the registrar?)

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: Willing to offline chat about your thoughts? We can update here with the result.

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: Correct, the overload that takes a registrar parameter allows the code to bypass the loop.

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// <typeparam name="T">The interface type of the service being requested.</typeparam>
+        /// <param name="serviceInstance">Output parameter to receive the requested service instance.</param>
+        /// <param name="registrar">Output parameter to receive the registrar that loaded the service instance.</param>
+        /// <param name="name">Optional name of the service.</param>
+        /// <returns>
+        /// True if the requested service is being returned, false otherwise.
+        /// </returns>
+        public static bool TryGetService<T>(
+            out T serviceInstance,
+            out IMixedRealityServiceRegistrar registrar,
```
wiwei: It's weird seeing outparams being first, instead of last - in other languages that support this type of mechanic most of the coding guidelines I've seen have been to put the input first, output last.

Other examples of code from warnings/guidelines:
[https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1021-avoid-out-parameters?view=vs-2019](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1021-avoid-out-parameters?view=vs-2019)
[https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier)

Seem to list it last as well.

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: haha! forgot to remove that one :)

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: nice catch. thanks!

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// <param name="registrar">Instance of the registrar manages the service.</param>
+        /// <returns>
+        /// True if the service was successfully removed, false otherwise.
+        /// </returns>
+        private static bool RemoveServiceInternal(
+            Type interfaceType,
+            IMixedRealityService serviceInstance,
+            IMixedRealityServiceRegistrar registrar)
+        {
+            List<KeyValuePair<IMixedRealityService, IMixedRealityServiceRegistrar>> services = registry[interfaceType];
```
wiwei: I think this will throw a KeyNotFound exception if this doesn't exist. AFAICT there are public->private paths that don't guard against this.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
davidkline-ms: it >might< be called by multiple people. i will remove for now until more thinking can be applied

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
wiwei: I guess in this case it's interesting because name is optional.

---
In reply to: [279109153](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4099#discussion_r279109153) [](ancestors = 279109153)

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: you're right! i'll check for that

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,222 @@
...
+                    serviceInstance = default(T);
+                    registrar = null;
+                    return false;
+                }
+            }
+            else
+            {
+                // Use the first service found
+                registryIndex = 0;
+            }
```
wiwei: Can you help me understand a bit about this name concept - it looks like it lets us register multiple services of the same type (i.e. multiple boundary systems?) What are the scenarios that we're trying to enable with this? The reason I ask is this name concept appears to add some additional complexity to this code and I"m trying to understand the value/use cases that it enables.

wiwei: If there are two services and no name is passed, shouldn't we log error /return false if there's more than one service? It feels like it's unpredictable otherwise and we don't know what the consumer actually wanted.

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,222 @@
...
+        /// <returns>
+        /// True if the requested service is being returned, false otherwise.
+        /// </returns>
+        public static bool TryGetService<T>(
+            out T serviceInstance,
+            string name = null) where T : IMixedRealityService
+        {
+            return TryGetService<T>(
+                out serviceInstance,
+                out _,                  // The registrar out param is not used, i tcan be discarded.
```
wiwei: >i tcan be discarded. [](start = 80, length = 20)

spelling : "it can be discarded"

davidkline-ms: fixed. thanks!

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: the multiple of the same type support is for extension services where it may be possible to load two of the same type. we should think about this to see if it is the desired behavior

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: good question. elsewhere in the toolkit a similar pattern is applied. intended to match that behavior. perhaps a name should be required?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/TeleportPointer.cs
```diff
@@ -362,7 +362,11 @@ public override void OnInputChanged(InputEventData<Vector2> eventData)
                             {
                                 canMove = false;
                                 // Rotate the camera by the rotation amount.  If our angle is positive then rotate in the positive direction, otherwise in the opposite direction.
-                                MixedRealityToolkit.Instance.MixedRealityPlayspace.RotateAround(CameraCache.Main.transform.position, Vector3.up, angle >= 0.0f ? rotationAmount : -rotationAmount);
+                                MixedRealityPlayspace.PerformTransformation(
+                                    p =>
+                                    {
+                                        p.RotateAround(CameraCache.Main.transform.position, Vector3.up, angle >= 0.0f ? rotationAmount : -rotationAmount);
+                                    });
```
wiwei: Is OnInputChanged called in the update loop (i.e. every frame?) If so we should avoid using lambdas here because they contribute to frame-by-frame memory alloc (and then eventually GC thrashing)

davidkline-ms: good question. it might be when the pointer is displayed. 

tbh, i adapted this code from a previous PR that used the lambda.

Assets/MixedRealityToolkit.Services/TeleportSystem/MixedRealityTeleportSystem.cs
```diff
@@ -283,11 +267,18 @@ private void ProcessTeleportationRequest(TeleportEventData eventData)
...
-            cameraParent.RotateAround(CameraCache.Main.transform.position, Vector3.up, targetRotation.y - CameraCache.Main.transform.eulerAngles.y);
+            MixedRealityPlayspace.PerformTransformation(
+                p =>
+                {
+                    p.position = targetPosition;
+                    p.RotateAround(
+                        CameraCache.Main.transform.position, 
+                        Vector3.up, 
+                        targetRotation.y - CameraCache.Main.transform.eulerAngles.y);
+                });
```
wiwei: Same comment from the other section - if this is going to be done in any sort of inner loop it would be best to avoid lambdas (which may mean we have to change the function definition of PerformanceTransformation

wiwei:
Alrighty, done taking a look

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: added comment as recommended

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: we discussed internally and decided to leave as is

davidkline-ms: we discussed internally and decided to leave as is

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: we discussed internally and decided to leave as is

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
davidkline-ms: we discussed internally and decided to leave as is

davidkline-ms:
@WiWei, can you take a look at the changes i made to the playspace class and the teleport pointer/system? they address your lambda performance concerns.

if you feel this is ready, go ahead and merge (i'll be out of the office) thanks!

Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
```diff
@@ -0,0 +1,231 @@
...
+                    // Failed to find the requested service.
+                    serviceInstance = default(T);
+                    registrar = null;
+                    return false;
+                }
+            }
+            else
+            {
+                // Use the first service found
+                registryIndex = 0;
```
wiwei: >registryIndex = 0; [](start = 16, length = 18)

Can you still add a Debug.Assert or warn or log or return failure if there are more than two services? This will be an undefined thing that could really bite people in the future (i.e. things could 'just work' luckily because it happens to be that the first dude always asks but the second dude doesn't, but if the second dude ever asks it's gonna be bad)

wiwei:
Just had one more comment about the registerIndex = 0 thing - I do feel fairly passionate about that because it's a case where there is non-determinism in what we return and this can lead to really hard to debug/diagnose things down the line.

Approving (so you can merge in after making that change)

davidkline-ms:
> Just had one more comment about the registerIndex = 0 thing - I do feel fairly passionate about that because it's a case where there is non-determinism in what we return and this can lead to really hard to debug/diagnose things down the line.

Change will be to return false with a log message stating that multple were found and to please query using the name

davidkline-ms:
if (services.Count > 0)
                {
                    Debug.LogWarning("Multiple instances of the requested service were found. Please re-call this method and provide a value for the name parameter.");
                    serviceInstance = default(T);
                    registrar = null;
                    return false;
                }
                registryIndex = 0;

Railboy:
I would love to incorporate the static playspace into #4035. When are we planning to merge this?

PR #4101, opened 2019-04-26
Fixed MousePointer

thalbern:
Overview
readded changes from https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3982 and addressed issues raised:

New Changes
--- readded scroll and original input mappings
--- implemented spatial ptr 
--- simplyfied position calculation
--- added gameobject position setting to conform with posesynchronizer
--- removed spatial grip input mapping -> will be fixed in the manipulation handler by @luis-valverde-ms 

------------------------------------------------
Original Changes
--- mouse cursor not stuck anymore in center of screen
--- only one mouse cursor gameobject is created / controller won't be enabled more than once
--- screenspace cursor will be hidden when spatial cursor is used
--- fixed other controllers interferring with mouse cursor visibility (hands / camera)
--- removed spatial mouse cursor (retrieved the mouse cursor position and sent it back to the mouse cursor to set the position of the cursor to that same position again)
--- removed mouse wheel scrolling that modified mouse cursor position (not needed as spatial cursor is snapping to interactable objects)
--- added ray to world space position calc to enable object manipulation (bounding box)
--- added spatial grab to controller mapping to support dragging of objects
--- added mouse cursor speed adjustment setting

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
@@ -46,7 +46,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         float IMixedRealityMousePointer.HideTimeout => hideTimeout;
 
         [SerializeField]
-        [Range(0.1f, 1f)]
+        [Range(0.1f, 10f)]
         [Tooltip("Mouse cursor speed that gets applied to the mouse delta.")]
         private float speed = 0.25f;
 
@@ -60,8 +60,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
         public override bool IsInteractionEnabled => isInteractionEnabled;
 
         private IMixedRealityController controller;
-
-        private Vector3 cachedWorldSpacePosition;
 
         /// <inheritdoc />
         public override IMixedRealityController Controller
@@ -102,14 +100,14 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
             // The RayCastCamera is placed so that the current cursor position is in the center of the camera's view space.
             Vector3 viewportPos = new Vector3(0.5f, 0.5f, 1.0f);
-            cachedWorldSpacePosition = RaycastCamera.ViewportToWorldPoint(viewportPos);
+            gameObject.transform.position = RaycastCamera.ViewportToWorldPoint(viewportPos);
         }
 
         public override Vector3 Position
         {       
             get
             {
-                return cachedWorldSpacePosition;
+                return gameObject.transform.position;
             }
         }
 
@@ -179,9 +177,23 @@ namespace Microsoft.MixedReality.Toolkit.Input
         {
             if (eventData.SourceId == Controller?.InputSource.SourceId)
             {
-                if (PoseAction == eventData.MixedRealityInputAction)
+                if (PoseAction == eventData.MixedRealityInputAction && !UseSourcePoseData)
                 {
-                    UpdateMousePosition(eventData.InputData.x, eventData.InputData.y);
+                    Vector3 mouseDeltaRotation = Vector3.zero;
+                    mouseDeltaRotation.x += eventData.InputData.x;
+                    mouseDeltaRotation.y += eventData.InputData.y;
+                    UpdateMouseRotation(mouseDeltaRotation);
+                }
+            }
+        }
+
+        public override void OnInputChanged(InputEventData<MixedRealityPose> eventData)
+        {
+            if (eventData.SourceId == Controller?.InputSource.SourceId)
+            {
+                if (UseSourcePoseData)
+                {
+                    UpdateMouseRotation(eventData.InputData.Rotation.eulerAngles);
                 }
             }
         }
@@ -227,15 +239,13 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
         #endregion MonoBehaviour Implementation
 
-        private void UpdateMousePosition(float mouseX, float mouseY)
+        private void UpdateMouseRotation(Vector3 mouseDeltaRotation)
         {
-            float scaledMouseX = mouseX * speed;
-            float scaledMouseY = mouseY * speed;
-            if (Mathf.Abs(scaledMouseX) >= movementThresholdToUnHide ||
-                Mathf.Abs(scaledMouseY) >= movementThresholdToUnHide)
-            {
+            if (mouseDeltaRotation.magnitude >= movementThresholdToUnHide)
+            { 
                 if (isDisabled)
                 {
+                    // if cursor was hidden reset to center
                     BaseCursor?.SetVisibility(true);
                     transform.rotation = CameraCache.Main.transform.rotation;
                 }
@@ -248,10 +258,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 timeoutTimer = 0.0f;
             }
 
-            var newRotation = Vector3.zero;
-            newRotation.x += scaledMouseX;
-            newRotation.y += scaledMouseY;
-            transform.Rotate(newRotation, Space.World);
+            transform.Rotate(mouseDeltaRotation, Space.World);
         }
     }
 }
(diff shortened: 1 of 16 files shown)


Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
```diff
@@ -130,6 +126,16 @@ public void Update()
                         }
                     }
                 }
+
+                if (Interactions[i].InputType == DeviceInputType.Scroll)
```
StephenHodgson: What's the benefit of moving this block?

thalbern: no benefit. just happened while merging the old with the new commits. if that's an issue I'll move it back? let me know

StephenHodgson: Just easier to review without as much churn. No worries.

Alexees:
would love to see this being approved as not being able to use the mouse slows down development quite a bit

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
```diff
@@ -65,32 +64,39 @@ public void Update()
...
+                    mouseDeltaRotation.y += UInput.GetAxis("Mouse X");
+                    IMixedRealityMousePointer mousePointer = (IMixedRealityMousePointer)InputSource.Pointers[0];
+                    if (mousePointer != null)
+                    {
+                        mouseDeltaRotation *= mousePointer.Speed;
+                    }
+
+                    MixedRealityPose controllerPose = MixedRealityPose.ZeroIdentity;
+                    controllerPose.Rotation = Quaternion.Euler(mouseDeltaRotation);
                     Interactions[i].PoseData = controllerPose;
```
luis-valverde-ms: >Interactions[i].PoseData = controllerPose [](start = 20, length = 41)

We seem to be storing just a rotation delta here. It should be a proper world space pose instead. The idea is to use that directly in the mouse pointer as ray cast origin.

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
```diff
@@ -65,32 +64,39 @@ public void Update()
...
             {
+                
+
                 if (Interactions[i].InputType == DeviceInputType.SpatialPointer)
                 {
+                    // add mouse delta as rotation
+                    var mouseDeltaRotation = Vector3.zero;
+                    mouseDeltaRotation.x += -UInput.GetAxis("Mouse Y");
+                    mouseDeltaRotation.y += UInput.GetAxis("Mouse X");
+                    IMixedRealityMousePointer mousePointer = (IMixedRealityMousePointer)InputSource.Pointers[0];
```
luis-valverde-ms: >IMixedRealityMousePointer mousePointer = (IMixedRealityMousePointer)InputSource.Pointers[0]; [](start = 20, length = 92)

Controllers shouldn't know about pointers. We can configure mouse speed either through a profile for the mouse device manager or by consuming just mouse deltas in the mouse pointer and applying it there.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
```diff
@@ -167,10 +171,23 @@ public override void OnInputChanged(InputEventData<Vector2> eventData)
...
-                    UpdateMousePosition(eventData.InputData.x, eventData.InputData.y);
+                    Vector3 mouseDeltaRotation = Vector3.zero;
+                    mouseDeltaRotation.x += eventData.InputData.x;
+                    mouseDeltaRotation.y += eventData.InputData.y;
+                    UpdateMouseRotation(mouseDeltaRotation * speed);
+                }
+            }
+        }
+
+        public override void OnInputChanged(InputEventData<MixedRealityPose> eventData)
```
luis-valverde-ms: >OnInputChanged [](start = 29, length = 14)

ControllerPoseSynchronizer is already doing this for us.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
```diff
@@ -75,15 +82,27 @@ public override IMixedRealityController Controller
         /// <inheritdoc />
         public override void OnPreSceneQuery()
         {
+            // screenspace to ray conversion
             transform.position = CameraCache.Main.transform.position;
 
-            Ray ray = new Ray(Position, Rotation * Vector3.forward);
+            Ray ray = new Ray(transform.position, transform.forward);
```
luis-valverde-ms: >Ray ray = new Ray(transform.position, transform.forward); [](start = 12, length = 57)

As far as I understand, the idea for pointers derived from controller pose synchronizer is that they will take the origin and direction from the pose reported by the controller. In that sense, here we should be using directly Position and Rotation * Vector3.forward for the ray.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
```diff
@@ -75,15 +82,27 @@ public override IMixedRealityController Controller
...
+
+            // ray to worldspace conversion
+            gameObject.transform.position = transform.position + transform.forward * DefaultPointerExtent;
+        }
+
+        public override Vector3 Position
+        {       
+            get
+            {
+                return gameObject.transform.position;
```
luis-valverde-ms: >gameObject.transform.position [](start = 23, length = 29)

By overriding the base implementation here we are ignoring raycastOrigin. I think we could be better off having mouse pointer not derive from BaseControllerPointer / ControllerPoseSynchronizer.

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
thalbern: The problem I have here is that I don't have access to the previous mouse position - I shouldn't query the pointer so I can't provide the position here if we want to keep using the delta

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
thalbern: the controllerPoseSynchronizer is not calling our method for calculating the rotation - I have to keep this method

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
thalbern: in the latest update there's a config profile for mouser input for setting the speed - that should solve this issue

Assets/MixedRealityToolkit/Providers/UnityInput/MouseController.cs
```diff
@@ -65,32 +65,40 @@ public void Update()
...
+                
+
                 if (Interactions[i].InputType == DeviceInputType.SpatialPointer)
                 {
+                    // add mouse delta as rotation
+                    var mouseDeltaRotation = Vector3.zero;
+                    mouseDeltaRotation.x += -UInput.GetAxis("Mouse Y");
+                    mouseDeltaRotation.y += UInput.GetAxis("Mouse X");
+                    
+                    var mouseInputProfile = MixedRealityToolkit.Instance?.GetService<MouseDeviceManager>()?.MouseInputProfile;
```
lukastoenneMS: GetService is surprisingly expensive, it does a whole bunch of allocations and linear string search. Typical usage is to have an internal ref and lazy-initialize it once, e.g.: [https://github.com/microsoft/MixedRealityToolkit-Unity/blob/acee452b82ffbfe6f48c0c4b1086058496d94161/Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs#L145](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/acee452b82ffbfe6f48c0c4b1086058496d94161/Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs#L145)

thalbern: caching profile now in controller and pointer

thalbern:
added follow up task for removing posesynchronizer from mouse pointer https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4251

PR #4110, opened 2019-04-28
Fix/clone settings should dirty scene

### LINKED ISSUE #4109: Copy & Customize of MRTK settings should dirty the scene so that profile change is saved
## Describe the bug

When MRTK applies changes via *Copy & Customize*, it should dirty the scene so that the scene can be saved.

## To reproduce

1. Create new project, add MRTK foundation .unitypackage
2. Click on Mixed Reality Toolkit > Add to Scene and Configure
3. Save scene
4. Select MixedRealityToolkit game object
5. In Inspector, click Copy & Customize button
6. View MixedRealityToolkit game object in inspector, notice that Active Profile = MixedRealityToolkitConfigurationProfile (the settings that you just cloned)
7. Exit Unity
8. Reopen project

## Actual behavior
MixedRealityToolkit game object Active Profile = DefaultMixedRealityToolkitConfigurationProfile.

## Expected behavior
* After step 5, asterisk (*) is shown in the Unity title bar, indicating that the scene is dirty
* As part of step 7, you are prompted to save changes to the scene before exiting
* After step 8, MixedRealityToolkit game object Active Profile = MixedRealityToolkitConfigurationProfile (the settings that you just cloned)

- Unity Version 2019.1.0f2
- MRTK Version v2.0.0 RC1

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

ForrestTrepte:
Call Undo.RecordObject so that editor will dirty the scene when ActiveProfile setting is changed.
Fixes #4109.

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
@@ -117,6 +117,7 @@ namespace Microsoft.MixedReality.Toolkit.Editor
                 {
                     ScriptableObject profile = CreateInstance(nameof(MixedRealityToolkitConfigurationProfile));
                     var newProfile = profile.CreateAsset("Assets/MixedRealityToolkit.Generated/CustomProfiles") as MixedRealityToolkitConfigurationProfile;
+                    UnityEditor.Undo.RecordObject(MixedRealityToolkit.Instance, "Create new profiles");
                     MixedRealityToolkit.Instance.ActiveProfile = newProfile;
                     Selection.activeObject = newProfile;
                 }
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
@@ -186,6 +186,7 @@ namespace Microsoft.MixedReality.Toolkit.Editor
                 // Sub-profiles are easy to update in the master configuration inspector.
                 if (MixedRealityToolkit.Instance.ActiveProfile.GetType() == profile.GetType())
                 {
+                    UnityEditor.Undo.RecordObject(MixedRealityToolkit.Instance, "Copy & Customize Profile");
                     MixedRealityToolkit.Instance.ActiveProfile = profile as MixedRealityToolkitConfigurationProfile;
                 }
             }


StephenHodgson:
Awesome stuff :)

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
```diff
@@ -117,6 +117,7 @@ public override void OnInspectorGUI()
                 {
                     ScriptableObject profile = CreateInstance(nameof(MixedRealityToolkitConfigurationProfile));
                     var newProfile = profile.CreateAsset("Assets/MixedRealityToolkit.Generated/CustomProfiles") as MixedRealityToolkitConfigurationProfile;
+                    UnityEditor.Undo.RecordObject(MixedRealityToolkit.Instance, "Create new profiles");
```
StephenHodgson: I think actually putting this in the `CreateAsset` overload would be a little better as we'd then get undo capabilities for each created profile as well.

ForrestTrepte: My intent was to cover the assignment to the ActiveProfile field, so I think the Undo.RecordObject needs to be here in the outer function.

ForrestTrepte: @StephenHodgson raises an interesting point about undo and creation of the asset itself. I was just interested in the side-effect of the undo, which is that it tells Unity the the object is dirty so that it can be saved. But if you actually want to be able to fully undo the whole shebang then I suppose you'd also want to delete the asset that's created. Perhaps that could be accomplished via Undo.RegisterCreatedObjectUndo inside the AssetDatabase.CreateAsset extension method, but I'm not sure if that works with assets and what all the consequences of that could be. I see several calls in MRTK to CreateAsset that don't seem to be creating an Undo record and I'm hesitant to wade into that territory.

For now, I'd like to stick with this fix as-is to ensure that the change is saved as described in #4109. @StephenHodgson, if you like you could open a new issue to suggest that calls to AssetDatabase.CreateAsset should be keeping Undo records. But I'm not sure if that's sufficiently important.

ForrestTrepte: Resolving the conversation. (Please let me know if that's not the etiquette for PR creator to resolve.)

StephenHodgson: I think that's fine :) Your reply makes sense to me.

I'm not sure it warrants an new issue, as I think it's just a nice to have feature.

Thanks @ForrestTrepte!

wiwei:
@keveleigh lemme know if you have background/time for this one, otherwise I can take a look at this today.

PR #4112, opened 2019-04-29
Enable the input simulation assembly for all platforms.

lukastoenneMS:
The service itself is still editor-only. Enabling the main assembly for
all platforms removes a warning when the input system tries to
deserialize the input sim profile on a non-editor platform.

Overview
---

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityToolkit.Services.InputSimulation.asmdef
+++ b/Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityToolkit.Services.InputSimulation.asmdef
@@ -5,9 +5,7 @@
         "Microsoft.MixedReality.Toolkit.SDK"
     ],
     "optionalUnityReferences": [],
-    "includePlatforms": [
-        "Editor"
-    ],
+    "includePlatforms": [],
     "excludePlatforms": [],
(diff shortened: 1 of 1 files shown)


SimonDarksideJ:
A quick interjection point, adding the InputSimulation to ALL platforms is the wrong way to go imho.  We ONLY need to simulate in the editor, on device is pointless as you HAVE the bits you wanted to simulate in the editor available.

The issue above seems to be more with what the framework does when the simulation is NOT available, which is what should be fixed. NOT putting it everywhere as it breaks the abstraction.

lukastoenneMS:
I agree, the input simulation assembly *should* not be required on non-editor platforms. The problem is that the profile type in that assembly is needed to fully deserialize any profile asset that is linked by the input system in its data providers list. @davidkline-ms is more familiar with services and data providers and concluded that, at least for the time being, there is no good alternative to having the profile type accessible on all platforms. So the compromise was to have the input sim *serivce* in its own editor-only assembly, but make the rest of the classes all-platforms.

lukastoenneMS:
Ok, i'm going to retract this. The warning isn't critical and we all seem to agree that a better solution for optional assemblies is needed. @andreiborodin offered to look into a better solution on the data providers side of things.

Issue #4114, opened 2019-04-29
Some Vector2 input changes are not reported correctly 


luis-valverde-ms:
The code used to set Vector2 values in interaction mappings (MixedRealityInteractionMapping.Vector2Data setter) performs the following check:
```
Changed = !vector2Data.x.Equals(value.x * invertXAxisFactor) &&
          !vector2Data.y.Equals(value.y * invertYAxisFactor);
```
Which only returns true if both x and y components have changed.

wiwei:
Closing, I believe this was fixed by #4116

PR #4116, opened 2019-04-29
Fix check for value change when setting Vector2 values in interaction mappings

luis-valverde-ms:
Fixes: #4114 .

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
@@ -212,6 +212,7 @@ namespace Microsoft.MixedReality.Toolkit.Tests.InputSystem
             var testValue1 = Vector2.one;
             var testValue2 = Vector2.zero;
 
+
             var initialValue = interaction.Vector2Data;
 
             Assert.True(initialValue == Vector2.zero);
@@ -234,6 +235,13 @@ namespace Microsoft.MixedReality.Toolkit.Tests.InputSystem
 
             Assert.True(setValue2 == testValue2);
             Assert.IsFalse(interaction.Changed);
+
+            // Make sure changes to a single component are detected correctly when only one axis is inverted (bug #4114)
+            interaction.InvertXAxis = true;
+            interaction.InvertYAxis = false;
+            interaction.Vector2Data = Vector2.one;
+            interaction.Vector2Data = new Vector2(1, -1);
+            Assert.IsTrue(interaction.Changed);
         }
 
         [Test]
--- a/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
@@ -206,73 +206,47 @@ namespace Microsoft.MixedReality.Toolkit.Tests.InputSystem
         #region Vector2
 
         [Test]
-        public void Test07_TestVector2Changed()
+        public void Test07_TestVector2()
         {
-            var interaction = new MixedRealityInteractionMapping(1, string.Empty, AxisType.DualAxis, DeviceInputType.None, MixedRealityInputAction.None);
-            var testValue1 = Vector2.one;
-            var testValue2 = Vector2.zero;
+            bool[] invertAxisValues = { true, false };
+            Vector2[] vectorValues = { new Vector2(1, 1), new Vector2(1, -1) };
 
-
-            var initialValue = interaction.Vector2Data;
-
-            Assert.True(initialValue == Vector2.zero);
-            Assert.IsFalse(interaction.Changed);
-
-            interaction.Vector2Data = testValue1;
-
-            Assert.IsTrue(interaction.Changed);
-
-            var setValue1 = interaction.Vector2Data;
-
-            Assert.True(setValue1 == testValue1);
-            Assert.IsFalse(interaction.Changed);
-
-            interaction.Vector2Data = testValue2;
-
-            Assert.IsTrue(interaction.Changed);
-
-            var setValue2 = interaction.Vector2Data;
-
-            Assert.True(setValue2 == testValue2);
-            Assert.IsFalse(interaction.Changed);
-
-            // Make sure changes to a single component are detected correctly when only one axis is inverted (bug #4114)
+            foreach (var invertXAxis in invertAxisValues)
             {
-                interaction.InvertXAxis = true;
-                interaction.InvertYAxis = false;
-                interaction.Vector2Data = Vector2.one;
-
-                Vector2 oneMinusOne = new Vector2(1, -1);
-                interaction.Vector2Data = oneMinusOne;
-                Assert.IsTrue(interaction.Changed);
-
-                interaction.Vector2Data = oneMinusOne;
-                Assert.IsFalse(interaction.Changed);
+                foreach (var invertYAxis in invertAxisValues)
+                {
+                    foreach (var value in vectorValues)
+                    {
+                        Test07_TestVector2_internal(invertXAxis, invertYAxis, value);
+                    }
+                }
             }
         }
-
-        [Test]
-        public void Test08_TestVector2NoChange()
+            
+        private static void Test07_TestVector2_internal(bool invertXAxis, bool invertYAxis, Vector2 vectorValue)
         {
-            var interaction = new MixedRealityInteractionMapping(1, string.Empty, AxisType.DualAxis, DeviceInputType.None, MixedRealityInputAction.None);
-            var testValue = Vector2.one;
+            string msg = string.Format("invertXAxis: {0}, invertYAxis: {1}, vectorValue: {2}", invertXAxis, invertYAxis, vectorValue);
 
-            var initialValue = interaction.Vector2Data;
+            var i = new MixedRealityInteractionMapping(
+                1, string.Empty, AxisType.DualAxis, DeviceInputType.None, MixedRealityInputAction.None, invertXAxis: invertXAxis, invertYAxis: invertYAxis);
 
-            Assert.True(initialValue == Vector2.zero);
-            Assert.IsFalse(interaction.Changed);
+            // Test initial state
+            Assert.AreEqual(Vector2.zero, i.Vector2Data, msg);
+            Assert.IsFalse(i.Changed, msg);
 
-            interaction.Vector2Data = testValue;
+            // Test change
+            i.Vector2Data = vectorValue;
+            Vector2 invertedValue = vectorValue * new Vector2(invertXAxis ? -1f : 1f, invertYAxis ? -1f : 1f);
+            Assert.AreEqual(invertedValue, i.Vector2Data, msg);
+            Assert.IsTrue(i.Changed, msg);
 
-            Assert.IsTrue(interaction.Changed);
+            // Test Changed gets reset after querying
+            Assert.IsFalse(i.Changed, msg);
 
-            // Make sure the second time we query it's false
-            Assert.IsFalse(interaction.Changed);
-
-            interaction.Vector2Data = testValue;
-
-            // Make sure if we set the same value it's false
-            Assert.IsFalse(interaction.Changed);
+            // Test no change
+            i.Vector2Data = vectorValue;
+            Assert.AreEqual(invertedValue, i.Vector2Data, msg);
+            Assert.IsFalse(i.Changed, msg);
         }
 
         #endregion Vector2
--- a/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
@@ -237,11 +237,18 @@ namespace Microsoft.MixedReality.Toolkit.Tests.InputSystem
             Assert.IsFalse(interaction.Changed);
 
             // Make sure changes to a single component are detected correctly when only one axis is inverted (bug #4114)
-            interaction.InvertXAxis = true;
-            interaction.InvertYAxis = false;
-            interaction.Vector2Data = Vector2.one;
-            interaction.Vector2Data = new Vector2(1, -1);
-            Assert.IsTrue(interaction.Changed);
+            {
+                interaction.InvertXAxis = true;
+                interaction.InvertYAxis = false;
+                interaction.Vector2Data = Vector2.one;
+
+                Vector2 oneMinusOne = new Vector2(1, -1);
+                interaction.Vector2Data = oneMinusOne;
+                Assert.IsTrue(interaction.Changed);
+
+                interaction.Vector2Data = oneMinusOne;
+                Assert.IsFalse(interaction.Changed);
+            }
         }
 
         [Test]
--- a/Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
+++ b/Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
@@ -401,22 +401,9 @@ namespace Microsoft.MixedReality.Toolkit.Input
                     Debug.LogError($"SetVector2Value is only valid for AxisType.DualAxis InteractionMappings\nPlease check the {inputType} mapping for the current controller");
                 }
 
-                if (invertXAxis || invertYAxis)
-                {
-                    float invertXAxisFactor = invertXAxis ? -1f : 1f;
-                    float invertYAxisFactor = invertYAxis ? -1f : 1f;
-
-                    Changed = !vector2Data.x.Equals(value.x * invertXAxisFactor) &&
-                              !vector2Data.y.Equals(value.y * invertYAxisFactor);
-
-                    vector2Data.x = value.x * invertXAxisFactor;
-                    vector2Data.y = value.y * invertYAxisFactor;
-                }
-                else
-                {
-                    Changed = vector2Data != value;
-                    vector2Data = value;
-                }
+                Vector2 newValue = value * new Vector2(invertXAxis ? -1f : 1f, invertYAxis ? -1f : 1f);
+                Changed = vector2Data != newValue;
+                vector2Data = newValue;
             }
         }
 


Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
```diff
@@ -401,22 +401,9 @@ public Vector2 Vector2Data
...
-
-                    vector2Data.x = value.x * invertXAxisFactor;
-                    vector2Data.y = value.y * invertYAxisFactor;
-                }
-                else
-                {
-                    Changed = vector2Data != value;
-                    vector2Data = value;
-                }
+                Vector2 newValue = value * new Vector2(invertXAxis ? -1f : 1f, invertYAxis ? -1f : 1f);
```
StephenHodgson: creating a new `Vector2` here will impact performance.

SimonDarksideJ:
Can you elaborate on what this is trying to fix, as sadly it's not an improvement.
As @StephenHodgson points out, creating a new vector 2 every frame, even though it's a struct, is going to impact performance.

Would be good to have a little more detail in the PR (filling in the PR form) explaining what is being impacted and what is being changed and more importently, why?
(sorry if that sounds abrupt, but this is why we talk about issues instead of just posting "fixes")

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
```diff
@@ -401,22 +401,9 @@ public Vector2 Vector2Data
                     Debug.LogError($"SetVector2Value is only valid for AxisType.DualAxis InteractionMappings\nPlease check the {inputType} mapping for the current controller");
                 }
 
-                if (invertXAxis || invertYAxis)
-                {
-                    float invertXAxisFactor = invertXAxis ? -1f : 1f;
-                    float invertYAxisFactor = invertYAxis ? -1f : 1f;
-
-                    Changed = !vector2Data.x.Equals(value.x * invertXAxisFactor) &&
```
StephenHodgson: How come we're not just updating the `&&` to an `||`? Seems like the only change that needed to happen based on the issue you've opened.

Hard to tell exactly how much this change will impact.

StephenHodgson:
Also, let's update the unit test to validate the change we expect to see.

SimonDarksideJ:
Agreed, the Interactions are the most heavily unit tested feature in the MRTK to date.  So if there is an issue, there should be a Unit Test to confirm the issue, so that the fix can be measured and ensure it doesn't happen again.  Good point @StephenHodgson 

@luis-valverde-ms can you add unit tests to highlight the issue and that this fix resolves it.

StephenHodgson:
Agreed, here's what it should likely look like: https://github.com/XRTK/XRTK-Core/pull/138

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
wiwei: Yep - I believe this gets referenced in the update loop and will end up leading to small allocs (which then trigger GC down the line). There's a couple possibilities here:

1) Switch to ||.

Advantages being less churn, but also possibly less readable code (i.e. it's easy to look at vector2Data != newValue, still easy to read the above but a little less natural)

2) Have a class-member variable for temp space to avoid alloc.

Slightly readable but now we have a what like 16 byte increase in class memory usage.

I don't really have a preference.

---
In reply to: [279413081](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4116#discussion_r279413081) [](ancestors = 279413081)

luis-valverde-ms: I'm quite ignorant about the life-cycle of temporary value-type objects in C#, but I would have expected the compiler being able to do something reasonable there. Can you elaborate a bit more on where that performance impact comes from?

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
luis-valverde-ms: The simpler the code, the better. We you talk about impact you mean performance or are you worried about a change in behavior?

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
StephenHodgson: > 1. Switch to ||.

💯 that was also my original suggestion.

> I would have expected the compiler being able to do something reasonable there

That's partly why we're being pretty explicit about how we're handling these value types here.

> Can you elaborate a bit more on where that performance impact comes from

This entire class was written to be highly efficient because it's **_very_** hot code, and we were trying to make sure all the memory was packed in tight. You're creating a `new` reference type that will need to be collected later and throws off the tightly packed memory layout. Why waste the time writing to a new memory address, when the one you have works just fine?

luis-valverde-ms: This makes me think that the Vector2 creation is not that expensive, but I'd be super happy if someone has some more info about what code ends up being generated.

> Both value-type objects such as structs and reference-type objects such as classes are destroyed automatically, but value-type objects are destroyed when their containing context is destroyed, whereas reference-type objects are destroyed by the garbage collector at an unspecified time after the last reference to them is removed

From https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-operator.

luis-valverde-ms:
@SimonDarksideJ , the detail about the issue this PR is trying to solve is in the issue itself (#4114 ). Let me know if you find the description there is not clear enough.

There's no harm in adding a unit test to make sure that changes to Vector2 affecting just one of the components are reported correctly, I'll add that.

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
luis-valverde-ms: > You're creating a new reference type

As far as I can tell Vector2 is a value type. Are you refering to a different type or am I missing something?

Roughly, how many times per frame do you expect this setter to be called? Do you have a use case in mind where this code path would be specially hot?

StephenHodgson: Sorry, you're correct. value type. I updated my comment to reflect.

All of the input mapping code is hot. Input has and always will be a performance critical area.

Railboy: These perf concerns come up often enough that I think a burden-of-proof approach is warranted. Otherwise they just turn into a (usually unresolved) back-and-forth about hypotheticals.

If we think this approach will cause a measurable performance problem, great - measure it. Profile and/or stopwatch both approaches and post the results. If the difference is negligible, discussion over. If it isn't, discussion over.

StephenHodgson: I mean @SimonDarksideJ went back and forth for weeks testing the perf, and I'm pretty confident there's results in the GitHub history somewhere in here.

Let's keep it the way it is and make minimal changes to fix the existing code because we know, without a shadow of a doubt it works and is performance tested.

IMHO The burden of proof falls on the one opening the PR making these kinds of sweeping changes.

But then again, it's not like you guys are keen on listening to us anyway, so at this point, I'm not even sure why I bother anymore. Your loss.

Railboy: Don't misunderstand me - I concede there could be an issue here. My suggestion is more about reducing the mental load of discussing it. Plus for hot code like this we should really be in the habit of profiling anyway.

wiwei: The conversation really went on here, but I wanted to mention:

1) Unity's own guidelines recommend against allocations in Update/LateUpdate (i.e. inner loop stuff)

https://unity3d.com/learn/tutorials/topics/performance-optimization/optimizing-garbage-collection-unity-games?playlist=44069

See "Don’t allocate in functions that are called frequently"

2) Once things go down to IL2CPP, I have no idea if any of the .NET rules for GC still apply given that it's some unity IL2CPP stuff that's going on. Maybe it's smart enough to actually cache that new value once (and thus it will keep using the same thing on the heap with no issues). The problem is that we don't know (or if someone does know and can point to data, please do!)

3) Given Unity's own recommendation, if we trust that Unity's recommendation is reasonable (which I believe is the case) and we don't yet have the cycles to run our own profiling to double check it, it seems reasonable to err on the side of following Unity's performance guidelines in how we approach our problems.

StephenHodgson: Look mom, not more allocations!
```suggestion
                Vector2 newValue = value * (Vector2.one * (invertXAxis ? -1f : 1f, invertYAxis ? -1f : 1f));
```

keveleigh: I'm not sure this suggestion compiles.
image

It doesn't appear better than using new:
image

Explanation of the above:
I ran three variations 1,000,000 times per update loop.
1. **TestOneMult** is a simplified, but compilable, version of the above suggestion: `var test = Vector2.one * 1;`.
2. **TestOne** is...`var test = Vector2.one;`.
3. **TestNew** is `var test = new Vector2(1, 1);`

| | TestOneMult| TestOne| TestNew|
| ------| ------------- |-------------| -----|
| Time (ms) | 494.19      | 171.82 | 161.86 |
|GC Alloc | 0     | 0|  0 |

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
keveleigh: I ran through perf testing the proposed new method (`TestNewM()`) and the existing method (`TestExisting()`).

Looks like using Vector2 equality checks and multiplication is much more expensive than the existing method of dealing entirely with float comparisons.

Value types (like Vector2) don't allocate on the heap, so there's no GC either way, but it does measure more expensive for other reasons.

The internals of my test methods are functionally identical to the code here (but feel free to run your own tests to double check), just run 1,000,000 times per update. Obviously that's way higher than the scale we're dealing with here, but it helps to measure the comparison.

image

| | TestExisting| TestNewM|
| ------| ------------- |-------------| 
| Time (ms) | 173.93      | 1394.3 | 
|GC Alloc (B) | 0     | 0|

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
SimonDarksideJ: Those results are very curious @keveleigh , can you share the unit test you wrote them in.
Especially the .One vs new result as .One is a static constant in the struct

StephenHodgson: Agreed. We have no way to validate your results.

>"Here guys! take these results and believe everything I tell you"

Not very scientific method of you mate. Easy way to win over someone who takes things at face value.

keveleigh: I pasted the tests above exactly as they're written. I thought that was interesting as well. When you're measuring a difference of 10ms over the span of a million runs, there may be some unexpected results with rounding and whatnot. I don't really think this test says anything specific on the comparisons between those two.
Testing only 1000 times led to One being faster than New _most_ of the time.

image

image

keveleigh: If you've run these tests and see different results, let me know.

keveleigh: >Not very scientific method of you mate. Easy way to win over someone who takes things at face value.

I posted the exact tests I ran in my original post:

image

StephenHodgson: Agreed, without being able to see the test setup we can very well assume you've rigged the results in the favor of your expected result. 

Also the commit suggested was supposed to be a play on "Look Mom, no hands!" 
I'm sad no one laughed.

StephenHodgson: Sure, but not the exact text procedure, which is important. Ask any scientist.

StephenHodgson: Also posting a picture of the code makes it a lot harder to copy and test locally as well.

Seems like you're intentionally trying to make it difficult to validate imho.

keveleigh: Here's a gist of the exact script I ran with the deep Unity profiler attached.

https://gist.github.com/keveleigh/05c22aa9c6609841636f40b823c791c6

Here's a second script I just ran to measure the methods without the profiler (I originally attached to profiler to try to detect allocations):

https://gist.github.com/keveleigh/e2ecd3b4a07c68397a6d66138678ca25

Please feel free to run both of this scripts and report back if your data appears different.

Second script data (still run 1 million times, but measured with stopwatch instead of Unity deep profiler):

| | TestOneMult| TestOne| TestNew|
| ------| ------------- |-------------| -----|
| Time (ms) | 74.3      | 26.76 | 25.04 |

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
keveleigh: For completeness.

Here's this gist I ran under the Unity deep profiler to get the screenshot and data above:
https://gist.github.com/keveleigh/e5b312e6bba6b472d6b724705884d20d

Here's this test with stopwatch instead of profiler:
https://gist.github.com/keveleigh/46e205c2550442298fe4364945615250

| | TestExisting| TestNewM|
| ------| ------------- |-------------| 
| Time (ms) | 32.1      | 204.9 |

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
Railboy: Vector2.one just returns a new Vector2. This is from source decompiled with Resharper:

```
    /// <summary>
    ///   <para>Shorthand for writing Vector2(1, 1).</para>
    /// </summary>
    public static Vector2 one
    {
      get
      {
        return new Vector2(1f, 1f);
      }
    }
```

StephenHodgson: Oh man, I thought Unity was doing the right thing and cashing that value like they do in their other static properties. They do it for `Vector3` so I assumed it also was the same for `Vector2`. I'd like to see the results of one that's using the cached value.

StephenHodgson: Actually, looking at the source, they do, in fact use a cached field value:

```C#
    public static Vector2 one
    {
      get
      {
        return Vector2.oneVector;
      }
    }
```

Railboy: The decompiled code is from 2016 so it must be out of date. Directional vectors got changed, too. Pretty strange that a cached field value is consistently slower in the tests.

StephenHodgson: Haven't had a chance to test definitively and get some results. I'll post my results when I get some time later today.

StephenHodgson: @keveleigh wow! I can't believe `new` is faster! I wonder what's even the point of using the static accessors, as I always was taught that's the benefit of even having them. 😆 

But I did add one more test to the mix. Caching that Vector2.one value locally:
```C#
[Test]
public void Test_04_01_01_Vector2_One()
{
    var watch = Stopwatch.StartNew();

    for (int i = 0; i < 10000000; i++)
    {
        var test = Vector2.one;
    }

    watch.Stop();
    Debug.Log($"Test Vector2.one execution Time: {watch.Elapsed.TotalMilliseconds} ms");

    watch = Stopwatch.StartNew();
    var vector2One = new Vector2(1f, 1f);

    for (int i = 0; i < 10000000; i++)
    {
        var test = vector2One;
    }

    watch.Stop();
    Debug.Log($"Test cached Vector2 execution Time: {watch.Elapsed.TotalMilliseconds} ms");

    watch = Stopwatch.StartNew();

    for (int i = 0; i < 10000000; i++)
    {
        var test = new Vector2(1f, 1f);
    }

    watch.Stop();
    Debug.Log($"Test new Vector2 execution Time: {watch.Elapsed.TotalMilliseconds} ms");
}
```
image

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
StephenHodgson: I think the tests in https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4116#discussion_r279865650 are a bit biased as they're not testing the validity of the vector2, instead the extra equality checks are really what the difference here is you're seeing.

I agree the latter is much easier to read and def performs better, but the point is the `new` vs cached Vector2 value.

A better objective test would be:

```C#
est]
public void Test_04_01_02_Checks()
{
    var iterations = 10000000;
    var value = new Vector2(1, 1);
    var vector2Data = new Vector2(2, 2);
    var changed = false;

    var cachedVector2 = new Vector2(1f, 1f);

    var watch = Stopwatch.StartNew();

    for (int i = 0; i < iterations; i++)
    {
        var invertMultiplier = Vector2.one;

        if (false)
        {
            invertMultiplier.x = -1f;
        }

        if (true)
        {
            invertMultiplier.y = -1f;
        }

        var newValue = value * invertMultiplier;
        changed = vector2Data != newValue;
        // use the internal reading for changed so we don't reset it.
        vector2Data = newValue;
    }

    watch.Stop();
    Debug.Log($"Test Vector2.one execution Time: {watch.Elapsed.TotalMilliseconds} ms");

    watch = Stopwatch.StartNew();

    for (int i = 0; i < iterations; i++)
    {
        var invertMultiplier = cachedVector2;

        if (false)
        {
            invertMultiplier.x = -1f;
        }

        if (true)
        {
            invertMultiplier.y = -1f;
        }

        var newValue = value * invertMultiplier;
        changed = vector2Data != newValue;
        vector2Data = newValue;
    }

    watch.Stop();
    Debug.Log($"Tes cached Vector2 execution Time: {watch.Elapsed.TotalMilliseconds} ms");

    watch = Stopwatch.StartNew();

    for (int i = 0; i < iterations; i++)
    {
        var newValue = value * new Vector2(false ? -1f : 1f, true ? -1f : 1f);
        changed = vector2Data != newValue;
        vector2Data = newValue;
    }

    watch.Stop();
    Debug.Log($"Tes new Vector2 execution Time: {watch.Elapsed.TotalMilliseconds} ms");
}
```
image

keveleigh: These tests were intended to test the existing code vs the new proposed code, not really anything about the validity of the Vector2. That was what [the other tests](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4116#discussion_r279860251) were for.

My tests showed that the existing code (`TestExisting`) was more performant than the proposed change (`TestNewM`). It seems to me that if changing the `&&` to `||` fixes the original bug (the proposal at the beginning of this comment chain), that'd be preferred, perf-wise.

StephenHodgson: Nobody questioned the performance between the two. Mainly the case was made against a smaller change vs a complete rewrite. Personally I agree, less code is better and the updated code likely performs better just from less to do, obviously.

The question came up against the `new Vector2` usage, and I think it's pretty clear, using a cached value is preferable for performance.

StephenHodgson: > When you talk about impact you mean performance or are you worried about a change in behavior?

Mainly I was just worried about the change in behavior, but that's unclear without first updating the unit tests to confirm the bug.

luis-valverde-ms:
Added that unit test. Regarding performance, thanks for those timings @keveleigh . Did you take them in a release build?

I have an alternative test that I think may be more representative of the actual performance difference in practice, see https://github.com/luis-valverde-ms/MixedRealityToolkit-Unity/tree/vector2_test. When I run the test scene in the editor I see the new setter producing times twice as long. If I run the same code as a standalone UWP Il2CPP app compiled in release I see the times being roughly the same.

I'd rather check the Vector2 version in because it is simpler and perf doesn't seem much different. Thoughts?

Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
```diff
@@ -234,6 +235,13 @@ public void Test07_TestVector2Changed()
 
             Assert.True(setValue2 == testValue2);
             Assert.IsFalse(interaction.Changed);
+
+            // Make sure changes to a single component are detected correctly when only one axis is inverted (bug #4114)
+            interaction.InvertXAxis = true;
+            interaction.InvertYAxis = false;
+            interaction.Vector2Data = Vector2.one;
+            interaction.Vector2Data = new Vector2(1, -1);
+            Assert.IsTrue(interaction.Changed);
```
StephenHodgson: Could we format this test like the others above where we're also testing the return value?

Typically it's best to also test setting the value with the same value again to ensure that not change has happened as well.

Thanks :)

luis-valverde-ms: > Could we format this test like the others above where we're also testing the return value?

I find this formatting quite natural and clear. Can you paste the test formatted as you'd like to see it?

StephenHodgson: Sure take a look [here](https://github.com/XRTK/XRTK-Core/pull/138/files)

I added an additional test case that was specifically designed to test the inverted axis types, and updated the tests to also validate two additional value changes. 

We're testing:

1. set initial value
2. did value change? `Asset.False`
3. set to value 1
4. did value change? `Assert.True`
5. set value 1 again
6. did value change? `Asset.False`
7. repeat for each additional value change (we used 4 additional values)
8. set initial value
9. did the value change? `Assert.True`
10. set initial value again
11. did the value change? `Assert.False`

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
StephenHodgson:
```suggestion
               var newValue = value;
               
               if (invertXAxis)
               {
                   newValue.x *= -1f;
               }
               
               if (invertYAxis)
               {
                   newValue.y *= -1f;
               }
```

Railboy: Very interesting results. Glad we kept testing.

luis-valverde-ms:
@StephenHodgson , I updated the Vector2 test to provide more coverage. Let me know if you think I left anything out.

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
luis-valverde-ms: What's the reasoning behid the proposed change?

Issue #4117, opened 2019-04-29
Null reference exception when clicking on manipulation handler with gaze pointer


luis-valverde-ms:
1. Open HandInteractionExamples scene
2. Play scene with an Xbox controller connected
3. Look at an object with a manipulation handler and press the A button in the controller
4. Observe following error:

```
NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.UI.ManipulationHandler.GetPointersVelocity () (at Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs:232)
```

This is happening because manipulation handler assumes that all pointers have a valid controller, which is not the case for the gaze pointer.

StephenHodgson:
You're right, not all pointers have controllers, but they all do have valid `InputSources`

luis-valverde-ms:
Fixed by #4118

PR #4118, opened 2019-04-29
Correctly handle pointers with no controller attached (i.e. gaze pointer) in manipulation handler

luis-valverde-ms:
Fixes: #4117  .

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
@@ -227,21 +227,31 @@ namespace Microsoft.MixedReality.Toolkit.UI
         private Vector3 GetPointersVelocity()
         {
             Vector3 sum = Vector3.zero;
+            int numControllers = 0;
             foreach (var p in pointerIdToPointerMap.Values)
             {
-                sum += p.Controller.Velocity;
+                if (p.Controller != null)
+                {
+                    numControllers++;
+                    sum += p.Controller.Velocity;
+                }
             }
-            return sum / Math.Max(1, pointerIdToPointerMap.Count);
+            return sum / Math.Max(1, numControllers);
         }
 
         private Vector3 GetPointersAngularVelocity()
         {
             Vector3 sum = Vector3.zero;
+            int numControllers = 0;
             foreach (var p in pointerIdToPointerMap.Values)
             {
-                sum += p.Controller.AngularVelocity;
+                if (p.Controller != null)
+                {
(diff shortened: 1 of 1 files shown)


StephenHodgson:
Why isn't this checking for a valid `InputSource` instead of a controller?

luis-valverde-ms:
AFAIK, a pointer may have a valid input source but no controller (i.e. gaze pointer). In this case, a valid controller seems to be required to obtain linear and angular velocities.

StephenHodgson:
I mean, you still can't get that data form an Xbox controller or similar traditional controller.

Not sure how this actually fixes anything.

luis-valverde-ms:
Not to sound self-referential but this PR fixes the issue it fixes, i.e.: #4117. I agree ManipulationHandler could do with a general review (e.g. use of controller velocities) but in the meantime fixing those null references does no harm and help us with upcoming changes to gaze pointer.

julenka:
> Not to sound self-referential but this PR fixes the issue it fixes, i.e.: #4117. I agree ManipulationHandler could do with a general review (e.g. use of controller velocities) but in the meantime fixing those null references does no harm and help us with upcoming changes to gaze pointer.

I agree that using velocities in ManipulationHandler is something we can remove. This is to support a specific behavior of throwing things, which I think is a great idea to have as a separate component, but doesn't seem critical for manipulation.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -227,21 +227,31 @@ private Vector3 GetPointersCentroid()
         private Vector3 GetPointersVelocity()
         {
             Vector3 sum = Vector3.zero;
+            int numControllers = 0;
             foreach (var p in pointerIdToPointerMap.Values)
             {
-                sum += p.Controller.Velocity;
+                if (p.Controller != null)
```
julenka: Would be nice to explain why we are doing this null check here instead of just using pointerIdToPointerMap.Count, it's not obvious to me :-)

julenka: > Not to sound self-referential but this PR fixes the issue it fixes, i.e.: #4117. I agree ManipulationHandler could do with a general review (e.g. use of controller velocities) but in the meantime fixing those null references does no harm and help us with upcoming changes to gaze pointer.

I think Stephen actually has a point. This will unblock people yes, but really we should remove all that velocity code from ManipulationHandler. I created a separate bug to track that. It sounds like this at least temporarily unblocks you though, which is why I approved this PR.

PR #4131, opened 2019-04-30
Updates Create Extension Service Wizard to generate service inspectors

Railboy:
Overview
---
Updates the wizard to generate an optional service inspector (#3773) for the new service.

ServiceInspectors

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/ExtensionTemplates/ExtensionServiceCreator.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/ExtensionTemplates/ExtensionServiceCreator.cs
@@ -49,9 +49,11 @@ namespace Microsoft.MixedReality.Toolkit.Editor
         {
             public string ServiceName;
             public bool UsesProfile;
+            public bool UsesInspector;
             public SupportedPlatforms Platforms;
             public CreationStage Stage;
             public string ServiceFolderPath;
+            public string InspectorFolderPath;
             public string InterfaceFolderPath;
             public string ProfileFolderPath;
             public string ProfileAssetFolderPath;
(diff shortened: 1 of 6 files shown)


StephenHodgson:
Awesome stuff as always @Railboy. I approve in spirit.

ExtensionTemplates/ExtensionInterfaceTemplate.txt
```diff
@@ -3,8 +3,8 @@ using Microsoft.MixedReality.Toolkit.Extensions;
 
 namespace #NAMESPACE#
 {
-   public interface #INTERFACE_NAME# : IMixedRealityExtensionService
-    {
-        // Expose service features and abilities here
-    }
+	public interface #INTERFACE_NAME# : IMixedRealityExtensionService
```
keveleigh: Do these need to be tabs? Or can they be spaces, like the code formatting?

Railboy: I was getting some funky results with spaces. Not sure why.

wiwei: Any chance we could do a tabs -> spaces conversion in CreateTextAssetFromTemplate to fix that part up?

Assets/MixedRealityToolkit/Inspectors/ExtensionTemplates/ExtensionServiceCreator.cs
```diff
@@ -401,6 +442,12 @@ public async Task BeginAssetCreationProcess()
             if (Result == CreateResult.Error)
                 return;
 
+            await Task.Delay(100);
```
wiwei: Some of the Task.Delays below make sense to me (i.e. waiting for compile) but these here I don't fully understand - CreateTextAssetFromTemplate appears to just be doing string replacement and WriteTextAssetToDisk LOOKS to be the sync flavor - Also it looks like each section is doing an unrelated thing (i.e. one for inspector, one for template, etc.)

What's the wait for (and could it be commented?). If it not necessary, can we remove it?

Usually anytime you see random 10-1000ms awaits in code it's a red flag.

ExtensionTemplates/ExtensionInterfaceTemplate.txt
wiwei: Or I guess we just depend on peoples' text editors to convert the automagically.

Assets/MixedRealityToolkit/Inspectors/ExtensionTemplates/ExtensionServiceCreator.cs
Railboy: Those waits serve no purpose other than to provide visual flow as assets are being created.

wiwei: Can you add a comment about this?

wiwei:
I think I just want to get that minor comment in, and once that's done I'll handle pushing this into mrtk_development. Lemme know when I can handle that part.

Railboy:
@wiwei Done.

PR #4132, opened 2019-04-30
Updates in the getting started

meulta:
Overview
---
Adding an explicit step to ask people to switch to the correct platform so the changes applied by adding the package are made to the correct one


wiwei:
Thanks again for making our docs better.

meulta:
My pleasure @wiwei

PR #4136, opened 2019-05-01
Raise focus exit event for last pointer leaving when two or more leave in the same frame

### LINKED ISSUE #4160: Pointer exit event can be raised before the last pointer leaves the target
If two pointers leave a given target in the same frame, FocusProvider.UpdateFocusedObjects() will raise the pointer left event for the first one leaving. We should instead keep track of the number of pointers leaving the target that frame and raise it for the last one.

luis-valverde-ms:
Modified FocusProvider.UnregisterPointer() to keep track of the number of pointers leaving a target in a given frame and raise the pointer left event for the last one, instead of the first.

Fixes #4160


luis-valverde-ms:
@Alexees , same story, do you mind having a quick look at this PR and at the corresponding issue?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -1050,11 +1058,20 @@ private void UpdateFocusedObjects()
...
+            {
+                return;
+            }
+
             // ... but now we trim out objects whose overall focus was maintained the same by a different pointer:
 
             foreach (var pointer in pointers)
             {
-                pendingOverallFocusExitSet.Remove(pointer.CurrentPointerTarget);
+                if (pointer.CurrentPointerTarget != null)
```
Alexees: Is this null check necessary? It's not performed for the pendingOverallFocusEnterSet.

luis-valverde-ms: It is. I found out the hard way: Dictionary will throw an exception if you try to remove a null key.

PR #4144, opened 2019-05-01
PressableButton fixes

thalbern:
* Use local space push distances, so that pressable buttons can be scaled
  - Add experimental versioning to PressableButton to not make this a breaking change
  - Adapt plane gizmos for local space unit distances

Overview
---

Changes
---
- Fixes: # .

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
@@ -4,6 +4,7 @@
 using Microsoft.MixedReality.Toolkit.Input;
 using Microsoft.MixedReality.Toolkit.Utilities;
 using System.Collections.Generic;
+using UnityEditor;
 using UnityEngine;
 using UnityEngine.Events;
 using UnityEngine.Serialization;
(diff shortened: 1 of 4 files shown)


Railboy:
The version check is pretty gnarly.

You could do this instead:

```
public enum ButtonSpaceMode 
{
    World
    Local
}

public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler 
{
    ...

    [SerializeField]
    private ButtonSpaceMode spaceMode = ButtonSpaceMode.World;

    ...
}
```

And then execute different press logic based on spaceMode.

This would enable local space distances without breaking existing buttons and without the need for a version check.

I see both modes being useful so having an option would be nice anyway.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -14,17 +15,93 @@ namespace Microsoft.MixedReality.Toolkit.UI
     /// A button that can be pushed via direct touch.
     /// You can use <see cref="Microsoft.MixedReality.Toolkit.Examples.Demos.PhysicalPressEventRouter"/> to route these events to <see cref="Interactable"/>.
     ///</summary>
+    [InitializeOnLoad]
     [RequireComponent(typeof(BoxCollider))]
     public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
     {
         const string InitialMarkerTransformName = "Initial Marker";
 
+        #region Experimental Versionining
```
julenka: Hm...this versioning stuff is interesting, but is it possible to move somewhere outside of the button class, so the button class can be focused on just that, pressing a button? Otherwise if we add this it's another thing we need to keep in our heads when using buttons, and coudl set a precedent for people putting unrelated code into classes.

Also, regarding versioning....is it required for this PR, or could we do the versioning experiment later? It would be too bad to prevent a bug fix from getting in just because we are experimenting with versioning.

julenka:
One reason we may not need to do such an advanced version upgrade for the buttons is that we are now very clearly visualizing the butotn push planes in the editor. So, as long as we clearly state "after you upgrade, double check the push planes for your custom buttons", we should be fine.

Also, we should update the pressable button prefabs, since I think most people just use the PressablebuttonPlaced prefab anyway.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -213,6 +314,8 @@ void IMixedRealityTouchHandler.OnTouchCompleted(HandTrackingInputEventData event
             {
                 touchPoints.Remove(eventData.Controller);
 
+                IsTouching = (touchPoints.Count > 0);
```
julenka: Why directly set the properties here and not use the old method to determine IsTouching? touchPoints.Count > 0.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -221,60 +324,65 @@ void IMixedRealityTouchHandler.OnTouchCompleted(HandTrackingInputEventData event
...
 
         private void UpdateMovingVisualsPosition()
         {
             if (movingButtonVisuals != null)
             {
-                Debug.Assert(initialTransform != null);
-                movingButtonVisuals.transform.position = initialTransform.position + WorldSpacePressDirection * currentPushDistance;
+                Debug.Assert(pushSpaceTransform != null);
+                movingButtonVisuals.transform.position =
+                    pushSpaceTransform.position + (Vector3)pushSpaceTransform.localToWorldMatrix.GetColumn(2) *
```
julenka: Is this guaranteed to be normalized? Is it okay that it is not normalized?

julenka:
It would be very useful to have a test scene committed (perhaps in playmode tests) that verifies that after this change we can press buttons whose global scale is at widely diffrent values, like 100x, 10x, 1x, 0.1x, 0.01x

julenka:
Isn't there a bug that this is actually fixing? It's not mentioned in the PR.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: > Isn't there a bug that this is actually fixing? It's not mentioned in the PR.

Yes there's at least 2 tickets that are fixed with this PR (https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4219, https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4216) - but I just wanted to open this as a draft for  having a discussion about it. If possible I'd like to avoid having the migration step (if we have to keep it in it should be moved out of the class). 
Lars had a good idea by having a flag that would allow users to switch the mode between world and local space coordinates. that way we wouldn't break anything that is already set up and customers using scaled versions of their buttons can just switch to the new mode where the distances are in local space

radicalad:
> Use local space push distances, so that pressable buttons can be scaled.
I haven't pulled to test, but I'm hoping this fixes a related issue I'm running into where I change the size of the pressable-button collider (not the object scale) and end up with wonky states pretty reliably.

image

image

thalbern:
closing this draft PR for now - will reopen a new PR with the changes we want to keep (and addressing the feedback) and link it here soon

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: hmm i'm not sure i see the previous change you're referring to. we used to set IsTouching in update which would be fine, but by setting it in the touch completed we can safe the per frame call of the setter. Am I missing something here?

julenka: never mind.

Issue #4009, opened 2019-04-17
Need documentation on how to trigger interactables from code


genereddick:
## Describe the problem
A couple of patterns for Interactables need documentation:
a) How to change state in code: 
- Example use case: Checking a checkbox triggers an OnClick Event set up in Inspector and hides an object. Tapping the object also hides it, and should cause the checkbox to show as checked. How would this be done in code?

b) Setting / changing defaults at start. Ideally this would be done by exposing some settings in the Inspector (see #3603), but alternately an explanation of how to accomplish this either in code, or by changing the Profile templates.

cre8ivepark:
@killerantz Could you please help with this?

killerantz:
We can show some better examples for things like the checkbox.

Just to break a few things apart to make sure we are talking about the same things.
State in the Interactable refers to things like Focus, Press, visited, disabled.
These can be set through code using SetFocus(true) or SetFocus(false). All the supported states have a setting. 

This doesn't seem to be what @genereddick  is referring to with the checkbox state. In order to not confused things like checked, toggled, selected with the basic interactive states and because the way the theme work in these alternative state I came up with Dimensions. Meaning a button can be focused and checked, in this case we may want to use a different theme to visually reflect this combination.

This is weird, but to change the Dimension of the Interactable, which is more of a cyclical thing, use the method SetDimensionIndex(0) for normal, or SetDimensionIndex(1) for checked. This whole system is controlled by the property Dimensions, which enables the Interactable to assign two themes to each target, one for each dimension. There are properties for Can Select and Can Deselect, but we could also add a default value.

![image](https://user-images.githubusercontent.com/22085553/56381656-5e9ac480-61ca-11e9-89ad-6b5c370de824.png)

To handle the toggle events, click the Add Event button at the bottom of the component and select "InteractableToggleReceiver" from the drop down. This will expose the OnSelect and OnDeselect events to make something turn on when the check box is checked or turn off when it is unchecked.

![image](https://user-images.githubusercontent.com/22085553/56381560-167ba200-61ca-11e9-9cc0-3a1ad2ffe16c.png)

The InteractableToggleCollection component on the RadialSet example has code that controls the radial buttons toggled state (or DimensionIndex) and setting the currentIndex will set the default selected radial button.

![image](https://user-images.githubusercontent.com/22085553/56381987-2e075a80-61cb-11e9-9c3c-742dc52f16eb.png)

[...]

PR #4145, opened 2019-05-01
Update Interactable docs with toggle and dimensions clarification

killerantz:
## Overview
A question about how the check box example in the InteractiveExamples scene. #4009 
How to trigger the toggled state.

Added some details to the Toggles section.


killerantz:
I redirected this PR to MRTK_developement instead of MRTK_release.

julenka:
@killerantz  It looks like there is a broken image, could you please fix? Other than that looks good to me and I'll merge it in.

image

Documentation/README_Interactable.md
```diff
@@ -55,41 +55,64 @@ Themes work a lot like materials. They are scriptable objects that contain a lis
...
+SetDimensionIndex(1);
+
+//Set the DimensinIndex - Untoggled
+SetDimensionIndex(0);
+```
 
-You can use an interactable to detect input events other than just OnClick. The events feature provides a way to enable functionality to extend a button, but not really visual or needed to provide feedback based on state changes.
+Every click event will advance the DimensionIndex which will increase until the set Dimensions value is reached then cycle or reset to 0. A good example of working with Dimensions with code is the InteractiveToggleCollection found in the InteractableExamples demo scene on the RadialSet object.
+
+<img src="../Documentation/ReadMeImages/Interactable/InteractableToggleCollection.png" width="450">
```
keveleigh:
```suggestion
<img src="../Documentation/Images/Interactable/InteractableToggleCollection.png" width="450">
```

killerantz: Thanks for that. Totally thought I tested the link.

Issue #4147, opened 2019-05-01
MixedRealityToolkitFacadeHandler Destroy all children creating MRTK instance


provencher:
## Describe the bug

The MixedRealityToolkitFacadeHandler has a function called DestroyAllChildren that gets called when the playmode state changes, among other times, which calls MixedRealityToolkit.Instance to grab it's children and destroy them.

However, when this happens, if there is not already an MRTK instance in the scene, it will create one and then destroy it. This is is very highly undesirable.

## To reproduce

Open a scene without the mrtk, but with the mrtk in the project. Hit play, and then stop playmode. DestroyAllChildren will be called and this behavior will be hit.

## Expected behavior

DestroyAllChildren is not executed if no MRTK instance exists in the scene, and cleanup scripts should not add garbage to the scene.

## Your Setup (please complete the following information)

- Unity Version 2018.3.8f1
- MRTK Version v2.0 latest dev

## Target Platform (please complete the following information)

- HoloLens 2

julenka:
This is an excellent bug description, thank you!

Railboy:
Yeah @provencher your description is why this was literally fixed in 15 minutes, thank you.

PR #4148, opened 2019-05-01
Stops FacadeHandler from accessing MRTK instance in DestroyAllChildren

Railboy:
## Overview
- Fixes: #4147


Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitFacadeHandler.cs
```diff
@@ -25,20 +25,19 @@ static MixedRealityToolkitFacadeHandler()
 
         private static void OnPlayModeStateChanged(PlayModeStateChange obj)
         {
-            if (!MixedRealityToolkit.IsInitialized)
```
julenka: Curious, why did we remove this segment here?

Railboy: It's already guaranteed to get called at least once by later methods so it was overkill.

PR #4152, opened 2019-05-02
Fix NearInteractionTouchable interference.

MenelvagorMilsom:
NearInteractionTouchable.DistanceToSurface was getting the distance to a rectangle in 3D space. This was causing issues where nearby touchable surfaces were interfering with eachother. This fix disregards points that don't project into the rectangle bounds.

I'm sure there's a better way to do this too, but this was a quick fix to existing code that fixed our repro case.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
@@ -218,10 +218,13 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 Vector3.Dot(pointOnPlane, LocalRight),
                 Vector3.Dot(pointOnPlane, localUp));
 
-            // Clamp to bounds
-            planeSpacePoint = new Vector2(
-                Mathf.Clamp(planeSpacePoint.x, -bounds.x / 2, bounds.x / 2),
-                Mathf.Clamp(planeSpacePoint.y, -bounds.y / 2, bounds.y / 2));
+            if (planeSpacePoint.x < -bounds.x / 2 ||
+                planeSpacePoint.x > bounds.x / 2 ||
+                planeSpacePoint.y < -bounds.y / 2 ||
+                planeSpacePoint.y > bounds.y / 2)
+            {
+                return float.PositiveInfinity;
+            }
 
             // Convert back to 3D space
             Vector3 clampedPoint = transform.TransformPoint((LocalRight * planeSpacePoint.x) + (localUp * planeSpacePoint.y) + localCenter);
--- a/Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
```diff
@@ -218,10 +218,13 @@ public virtual float DistanceToSurface(Vector3 samplePoint)
                 Vector3.Dot(pointOnPlane, LocalRight),
                 Vector3.Dot(pointOnPlane, localUp));
 
-            // Clamp to bounds
-            planeSpacePoint = new Vector2(
-                Mathf.Clamp(planeSpacePoint.x, -bounds.x / 2, bounds.x / 2),
-                Mathf.Clamp(planeSpacePoint.y, -bounds.y / 2, bounds.y / 2));
+            if (planeSpacePoint.x < -bounds.x / 2 ||
```
julenka: Could you please add a comment here explaining why we are doing this in the code, so that when somebody new to the code reads it they know why we did it?

julenka:
hm...for the case of buttons, what happens if you push through the button, but then your finger goes outside the bounds of the touch rectangle? Would you always get a touch up then, because the DistanceToPlane function would return infinity?

julenka:
I'm not clear on how returning the distance to the clamped rectangle was causing interference, could you please provide a bit more detail?

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
MenelvagorMilsom: I want to give the code a second look. If we are doing things this way there's a better way to do it. But I'll make sure to explain reasoning.

MenelvagorMilsom:
> hm...for the case of buttons, what happens if you push through the button, but then your finger goes outside the bounds of the touch rectangle? Would you always get a touch up then, because the DistanceToPlane function would return infinity?
Yeah, I haven't tested this, but I imagine that this would be the behaviour. To me, if your finger leaves the touchable volume, you should get a touch up.

MenelvagorMilsom:
> I'm not clear on how returning the distance to the clamped rectangle was causing interference, could you please provide a bit more detail? Want to better understand what was happening in case there's a better solution, since I'm worried that if we use this approach we will hit other problems when peoples' fingers move outside the bounds of the rectangle.

It's hard to put to words, I'll see if I can make a little diagram of the problem as I see it. I'll post that here soon.

MenelvagorMilsom:
IMG_20190502_093458
So I've drawn this top down diagram of our repro scene. Hopefully it helps me get the ideas across. In this image, the X is the fingertip. In both diagrams. the horizontal line on the left is a button and the horizontal line on the right is a just a touchable surface.

We'll discuss the diagrams with the logic as it exists now, without changes.

In diagram 1, everything is working as expected, the button receives the events as you start pushing it as it is the nearest surface. However, in diagram 2, before you reach the button's back plate, your fingertip comes closer to the surface on the right, preventing the button from being pressable this close to the other surface.

I hope this explains what I mean by interference.

julenka:
Have you tested this change with the touchable cubes? These ones in HandInteractionExamples scene: 
image

With those cubes, I no longer see the finger cursor showing up when I approach the cubes from the side.

I did actually come up with a workaround, which is to add a touchable for each side of the cube:

image

julenka:
I think we don't want to add 6 near interaction touchables to each cube to make it touchable from all sides though, since I believe pokepointer queries all touchables every frame. Instead, we probably want to have a 'near interaction touchable type' like 'quad', 'plane', 'bounding box' and compute the distances based on the type.

I think we can do that in a separate PR.

julenka:
I think we don't want to add 6 near interaction touchables to each cube to make it touchable from all sides though, since I believe pokepointer queries all touchables every frame. Instead, we probably want to have a 'near interaction touchable type' like 'quad', 'plane', 'bounding box' and compute the distances based on the type.

I think we can do that in a separate PR.

julenka:
> I think we don't want to add 6 near interaction touchables to each cube to make it touchable from all sides though, since I believe pokepointer queries all touchables every frame. Instead, we probably want to have a 'near interaction touchable type' like 'quad', 'plane', 'bounding box' and compute the distances based on the type.
> 
> I think we can do that in a separate PR.

I filed #4174 to fix the issue of not being able to use volumes

Issue #4158, opened 2019-05-02
Focus provider does not raise pointer events correctly when unregistering a pointer with focus


luis-valverde-ms:
When FocusProvider.UnregisterPointer() is called for a pointer that is currently targeting something, it raises the pre focus changed event after raising the focus exit one (should be the other way around) and it does not raise the focus changed event at all.

luis-valverde-ms:
Fixed by #4159

PR #4159, opened 2019-05-02
Raise pre focus changed event before focus exit when unregistering pointer

luis-valverde-ms:
Also, raise pointer changed event.
Fixes: #4158


luis-valverde-ms:
@Alexees , I don't seem to be able to add you as reviewer. Can you have a quick look here and also make sure that the associated bug correctly describes the problem? I'll deal with issue with the pointer exit event in UpdateFocusedObjects() in a separate PR.

Alexees:
looks good :)

PR #4171, opened 2019-05-03
Fix grab pointer is disabled when poke pointer near touchable.

julenka:
## Overview
- Fixes: #4170

The root cause of this bug is that the grab pointer was getting disabled whenever the poke pointer was near a surface, this was a bug in the input mediator.

In addition, I made a few other changes:
- Only render debug pointer rays if the rays are active (makes them easier to view)
- Update default pointer profile to include several colors of pointer rays so we can tell apart poke pointer from grab pointer. The new colors are as follows:

image

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
@@ -703,7 +703,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
                     MixedRealityRaycaster.DebugEnabled = pointerProfile.DebugDrawPointingRays;
 
                     Color rayColor;
-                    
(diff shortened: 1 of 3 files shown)


provencher:
Thanks for the quick turnaround on this! I'll give it a go tomorrow and let you know if it solves my issue.

Cheers for improving the debugging as well!

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -703,7 +703,7 @@ private void UpdatePointers()
                     MixedRealityRaycaster.DebugEnabled = pointerProfile.DebugDrawPointingRays;
 
                     Color rayColor;
-
+
```
keveleigh:
```suggestion

```

provencher:
Didn't get a chance to try it, but I'm glad it's been merged. I'll pull the latest dev branch next week and get you some feedback @julenka.

Happy you managed to sort this out amid Mr dev days

Issue #4179, opened 2019-05-04
ManipulationHandler transform not settable at runtime


randystiles:
## Describe the bug

ManipulationHandler does not allow runtime setting of hostTransform for manipulation.

Works in editor, but not in a runtime script.

## To reproduce

1. Add manipulation handler to your GameObject in script using AddComponent
2. Try to set the hostTransform
3. It is private and there is no set method for it.

## Expected behavior

In script (runtime) set proper transform to use for Manipulation

Need method for setting hostTransform as part of class

## Your Setup (please complete the following information)

- Unity Version [2018.3.11f1]
- MRTK Version [ v2.0 RC Refresh]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive

chrisfromwork:
It appears that julia has fixed this issue. If this bug exists with the fix in 4184, please reopen.

Issue #4173, opened 2019-05-03
null reference when creating ManipulationHandler at runtime


lukastoenneMS:
## Describe the bug
I am creating a number of ManipulationHandler components _at runtime_ (because there are too many parts to set up manually in the editor). I'm getting null references on the 4 ManipulationEvents in the component (ManipulationStarted/Ended, HoeverEntered/Exited).

The event fields are never explicitly assigned in the code, it seems they "usually" are created automatically somehow, but this seems to fail at runtime.

## To reproduce

1. Create a new scene with MRTK
2. Add a cube
3. Use a script component to create the ManipulationHandler, like so:
```

using Microsoft.MixedReality.Toolkit.UI;
using UnityEngine;

public class ManipTest : MonoBehaviour
{
    void Start()
    {
        gameObject.AddComponent<ManipulationHandler>();
    }
}

```
4. Enter play mode
5. Look at the Cube
6. Hit null reference in ManipulationHandler.OnFocusEnter

## Expected behavior

Events are either initialized properly or checked for null.

## Your Setup (please complete the following information)

- Unity Version 2018.3.9f1
- MRTK RC1 Refresh

## Additional context

The null reference doesn't always seem to be logged in Unity, but is hit consistently when putting a breakpoint or debug print. This may be due to uninitialized memory (the events are not initialized to null either).

randystiles:
This same runtime issue with events is probably present in the other interaction types, where they presume things are set up in the editor, instead of created at runtime.

chrisfromwork:
A fix has been checked into mrtk_development. Please reopen this issue if additional problems occur or if the fix does not suffice.

PR #4181, opened 2019-05-05
ManipHandler fixes when using programmatically, and two playmode manipulation tests!

julenka:
## Overview
A few maniphandler fixes, and adding a few simple tests for ManipHandler!
- Expose all serialized fields as properties so they can be set programmatically (we had missed a few)
- ManipHandler check for null before invoking events.
- Initialize events handlers so that events like hover, manipstart can be accessed programmatically. 

Also, adding a couple playmode test changes to actually test manip handler!
- Add ManipHandler test for instantiating programmatically, and seeing basic hover events.
- Fix bug that was preventing multiple playmode tests from running sequentially. CleanupScene() would create a new scene every run, however in PlayMode this would cause MRTK to not be initialized. Instead, let the PlayMode tests instantiate the scenes (they do this by default, it seems).

## Fixes
- Fixes: #4179
- Fixes: #4173 

## Verification
- I verified this by running all EditMode and PlayMode tests. Also verified by testing the repro case in #4173.
- Also tested the HandInteractionExample scene in editor. 
- Verified that code compiles for UWP


Assets/MixedRealityToolkit.Tests/PlayModeTests/TestFixture_02_ManipulationTests.cs
```diff
@@ -0,0 +1,94 @@
...
+using NUnit.Framework;
+using System.Collections;
+using UnityEditor.SceneManagement;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+using UnityEngine.TestTools;
+using Microsoft.MixedReality.Toolkit;
+namespace Microsoft.MixedReality.Toolkit.Tests
+{
+    public class TestFixture_02_ManipulationTests
```
wiwei: I see that this follows the edit mode naming scheme - do you know if the TestFixture and # actually is necessary here? If not, they are arbitrary and non-value accruing noise (i.e. the numbering is arbitrary based on when things are being added, and TestFixture prefixes everything and isn't super helpful). If this works without the naming, I would rather we move our tests to be named based on what they are testing.

Assets/MixedRealityToolkit.Tests/PlayModeTests/TestFixture_02_ManipulationTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Test creating adding a ManipulationHandler to GameObject programmatically.
+        /// Should be able to run scene without getting any exceptions.
+        /// </summary>
+        /// <returns></returns>
+        [UnityTest]
+        public IEnumerator Test01_ManipulationHandlerInstantiate()
+        {
+            TestUtilities.InitializeMixedRealityToolkitScene(true);
+
+            var playspace = MixedRealityToolkit.Instance.MixedRealityPlayspace;
```
wiwei: This playspace setup is common to a number of they playmode tests, and might be worth having a TestUtilities.InitializePlayspace().

Note that it's possible to go either way here - having code written out this way has an advantage in that the entire test is clearly readable (in terms of how it is setting up the playspace, and then where the cube gets put, etc.

wiwei: At the same time if playspace creation (i.e. the light location + playspace lookat) will basically be the same for our playspace tests then having a single thing would probably reduce the noise in the tests to just what is being tested.

Assets/MixedRealityToolkit.Tests/PlayModeTests/TestFixture_02_ManipulationTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            var testObject = GameObject.CreatePrimitive(PrimitiveType.Cube);
+            testObject.transform.localScale = Vector3.one * 0.2f;
+
+            var manipHandler = testObject.AddComponent<ManipulationHandler>();
+            // Wait for two frames to make sure we don't get null pointer exception.
+            yield return null;
+            yield return null;
+
+            GameObject.Destroy(testObject);
```
wiwei: Do you know why testLight doesn't need to be destroyed? Or is it that this is also intending to test that component creation -> manipulation -> destruction's entire lifecycle works as expected?

wiwei:
It looks like the mrtk_PR pipeline doesn't do play mode tests, so my theory that it should be run in PRs didn't pan out. @MenelvagorMilsom do you know if there were specific things that made it so that we weren't able to do the play mode tests in CI?

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -41,18 +41,6 @@ public static void CleanupScene()
                 return;
             }
 #endif
-
-            if (testScene.IsValid() && testScene.isLoaded)
-            {
-                SceneManager.UnloadSceneAsync(testScene);
-            }
-
-            testScene = SceneManager.CreateScene("TestScene");
```
keveleigh: Is this not needed? The comment above mentions that the remaining lines work in edit mode (and are blocked by a `!EditorApplication.isPlaying` check), but that we need to set the scene up manually in play mode.

wiwei:
> It looks like the mrtk_PR pipeline doesn't do play mode tests, so my theory that it should be run in PRs didn't pan out. @MenelvagorMilsom do you know if there were specific things that made it so that we weren't able to do the play mode tests in CI?

Okay hopefully this will work:

https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4185

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
julenka: I saw those comments too, and I didn't initially have it. However, when I had the code above I noticed that when I ran all new 3 playmode tests that I had, the first test passed, but then the last few tests would fail (I forget which error now).

After debugging for a bit I realized that when we create a new scene for every tests, the following happens:

1. We create a scene, with one version of MRTK A for scene A.
2. We unload / remove scene A, but MRTK for scene A is still around.
3. We create a new scene B, but the MRTK singleton that we are using is still from Scene A.
4. As we use scene B, but MRTK doesn't get any events from scene B because it seems to still be tied to scene A.

When I removed the code to actually create this test scene, I actually managed to get all tests to pass. The reason I think this happens is that each test class creates and then properly tears down its own scene, including tearing down MRTK properly. So we get MRTK A with Scene A, MRTK B with Scene B.

wiwei:
> > It looks like the mrtk_PR pipeline doesn't do play mode tests, so my theory that it should be run in PRs didn't pan out. @MenelvagorMilsom do you know if there were specific things that made it so that we weren't able to do the play mode tests in CI?
> 
> Okay hopefully this will work:
> 
> #4185

Ah, that one didn't work because of some other issues, anyways, let's just move forward with this PR and I'll handle the build part.

Assets/MixedRealityToolkit.Tests/PlayModeTests/TestFixture_02_ManipulationTests.cs
keveleigh: Yeah, I'm not aware of the naming being a requirement. TestFixture is NUnit's term, and it's an optional attribute that can be added to the class.

If anything, the numbering helps organize the test fixtures in the Test Runner window in a specific order (instead of alphabetically).

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
julenka: Okay, I am wrong. Here is what I think happens in Play Mode Tests:

1. At start of test run, test scene is created, with MRTK A
2. All tests run in that one scene. 

If we want to ensure that for each class / each test, we use a new scene, we need to destroy / create the scene in cleanup. Only problem is that when you create a new scene, the old MRTK instance is still sticking around. So if you create Scene B, the MRTK from Scene A is still around, it seems.

I see two approaches:

1. Destroy the scene for each test.
2. In the cleanup for each scene, destroy all gameObjects in the scene. Could be very expensive.

Assets/MixedRealityToolkit.Tests/PlayModeTests/TestFixture_02_ManipulationTests.cs
julenka: It does, my tests are currently adding a bunch of lights to the tests, which is not correct.

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
julenka: Upon even further investigation (I tried adding back the technique of unloading old scene, and creating new one), I have realized why I removed the code in the first place. Two reasons:

First, when I initially tried using this code and running two tests, I ran into the following error: "A scene named 'TestScene' already exists". After I resolved that error by waiting for the async op to finish before loading a new scene, I ran into my second issue.

The second issue is what I mentioned above: that the MRTK from the old scene was still sticking around. The main symptom of this was that MRTK was throwing errors saying a main camera existed when it didn't. Also, after unloading a scene I would check in debugger in MRTK was still initialized, and it was.

To avoid going down a rabbit hole, I think for now a good policy is just to have all playmode tests run in the same scene. This currently means that all tests need to make sure to remove all objects they add. It's not great and I think we should solve the problem that MRTK is still around even when a scene is destroyed. However, I don't want to block completing this PR to do that, since this PR fixes bugs to manipulationhandler and does add new tests.

@Railboy and @lukastoenneMS  I'd love to hear your thoughts on this.

Assets/MixedRealityToolkit.Tests/PlayModeTests/TestFixture_02_ManipulationTests.cs
julenka: lights are not always needed, but I did create an InitializePlayspace method.

Assets/MixedRealityToolkit.Tests/PlayModeTests/TestFixture_02_ManipulationTests.cs
julenka: Removed the TestFixture_ prefxi from PlayMode tests. I think we could also remove them from EditoMode Tests.

MenelvagorMilsom:
> It looks like the mrtk_PR pipeline doesn't do play mode tests, so my theory that it should be run in PRs didn't pan out. @MenelvagorMilsom do you know if there were specific things that made it so that we weren't able to do the play mode tests in CI?

Sorry, I experimented with this before, but there were issues with the build machines. It looks like you've hit a different issue in #4185. We should definitely look at getting these tests running on CI and for PR validation.

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
lukastoenneMS: I think running on the same scene for these particular tests is fine.

Once we actually make runtime modifications to scenes it becomes more difficult, because to reliably run the tests we'd want to start from a clean slate and (re)load a new scene.

PR #4189, opened 2019-05-06
Add calibration process based on qr codes/aruco markers

chrisfromwork:
## Overview
This review add some logic that will allow dslr cameras to be calibrated using a chessboard and qrcodes + aruco markers.

1) A new scene is added for calculating camera intrinsics based on detecting a chessboard calibration board.
2) A new scene is added for calculating camera extrinsics based on detecting qr codes and aruco markers on the same calibration board

Note: This functionality requires opencv logic built out of the mixed reality toolkit repo. It also requires qr code detection dlls that are not yet available to the public. It also requires the spectator view pro compositor which is not immediately intuitive to use/is still under active refactoring.

## Changes
- Fixes: # .


Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
```diff
@@ -89,8 +87,17 @@ public bool TryGetLocationForQRCode(Guid id, out Matrix4x4 location)
...
-            var coordinateSystem = SpatialGraphInteropPreview.CreateCoordinateSystemForNode(id);
-            return TryGetLocationForQRCode(coordinateSystem, out location);
+            try
+            {
+                var coordinateSystem = SpatialGraphInteropPreview.CreateCoordinateSystemForNode(id);
+                return TryGetLocationForQRCode(coordinateSystem, out location);
+            }
+            catch (Exception e)
+            {
+                Debug.LogError($"Exception thrown creating coordinate system for qr code id: {e.ToString()}");
```
wiwei: >qr code id: [](start = 81, length = 11)

I think either update this string to be "exception: " or also log the qr code id as well.

Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs
```diff
@@ -89,8 +87,17 @@ public bool TryGetLocationForQRCode(Guid id, out Matrix4x4 location)
...
+                return false;
+            }
+
 #else
             Debug.LogError("Failed to create coordinate system for qr code id");
```
wiwei: >id" [](start = 75, length = 3)

Nit: Unrelated to this PR, but might be helpful to list the ID here (it seems like it was intended to list the ID but maybe it was missed)

wiwei:
> // #define QRCODESTRACKER_BINARY_AVAILABLE

Is this commented out define useful for some other stuff?

---
Refers to: Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs:7 in e7bc5ca. [](commit_id = e7bc5ca6da570cdfeff9307d0fcd4ea121fe15a6, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/Socketer/Scripts/SocketEndpoint.cs
```diff
@@ -40,7 +40,7 @@ public bool IsConnected
         /// </summary>
         public void Disconnect()
         {
-            if (State != ConnectionState.Disconnected)
+            if(State != ConnectionState.Disconnected)
```
wiwei: >f( [](start = 13, length = 2)

Super nit: space between if (State != ConnectionState.Disconnected)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
```diff
@@ -0,0 +1,389 @@
...
+        {
+            try
+            {
+                if (InitializeCalibrationNative())
+                {
+                    initialized = true;
+                    return;
+                }
+            }
+            catch { }
```
wiwei: >catch { } [](start = 12, length = 9)

Are the exceptions generated from this not helpful (to catch + log)?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
```diff
@@ -0,0 +1,389 @@
...
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
+{
+    internal class CalibrationAPI
+    {
+#if UNITY_WSA
+        [DllImport("SpectatorViewPlugin", EntryPoint = "InitializeCalibration")]
+#else
+        [DllImport("SpectatorViewPlugin.Editor", EntryPoint = "InitializeCalibration")]
+#endif
+        internal static extern bool InitializeCalibrationNative();
```
wiwei: >InitializeCalibrationNative [](start = 36, length = 27)

Question about the mechanics of this (and I guess, maybe the answer is I just get pointed to docs on how this works). Is this a "one global per process initialization step" or is it possibly thread dependent as well? Since we do it in response to newing up the CalibrationAPI I asssume that the underlying APIs as well must be thread-safe (either that or whatever it's doing must set things up for the entire process)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
```diff
@@ -0,0 +1,389 @@
...
+
+            int sizeExtrinsics = 7;
+            float[] extrinsics = new float[numExtrinsics * sizeExtrinsics];
+            if (!ProcessIndividualArUcoExtrinsicsNative(inputIntrinsics, extrinsics, numExtrinsics))
+            {
+                PrintLastError();
+                return null;
+            }
+            else
+            {
```
wiwei: Nit: This else isn't strictly necessary, and is also not consistent with how things are done in the rest of this file.

i.e. CalculateGlobalArUcoExtrinsics has something similar and does without the else.

While I think I'd prefer the model that CalculateGlobalArUcoExtrinsics has, I think it's more important to be consistent here so it's up to you which model you pick.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
```diff
@@ -0,0 +1,389 @@
...
+                List<CalculatedCameraExtrinsics> output = new List<CalculatedCameraExtrinsics>();
+                for (int i = 0; i < numExtrinsics; i++)
+                {
+                    int offset = i * sizeExtrinsics;
+                    bool succeeded = (extrinsics[offset] > 0.01f);
```
wiwei: >(extrinsics[offset] > 0.01f) [](start = 37, length = 28)

This may be answered in some docs though I'm not familiar - what's the special reason why 0.01f is chosen here as "success"?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
```diff
@@ -0,0 +1,389 @@
...
+            inputIntrinsics[11] = 0.0f; // reprojection error, unused
+
+            int sizeExtrinsics = 7;
+            int numExtrinsics = 1;
+            float[] extrinsics = new float[sizeExtrinsics];
+            if (!ProcessGlobalArUcoExtrinsicsNative(inputIntrinsics, extrinsics, numExtrinsics))
+            {
+                PrintLastError();
+                return null;
+            }
```
wiwei: This code has some copy-paste with lines 211-224, consider having a helper.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
```diff
@@ -0,0 +1,389 @@
...
+            leftHandedMatrix.m32 *= -1.0f;
+
+            var inverse = leftHandedMatrix.inverse;
+            var inversePosition = inverse.GetColumn(3);
+            var inverseRotation = Quaternion.LookRotation(inverse.GetColumn(2), inverse.GetColumn(1)) * Quaternion.Euler(0, 0, 180);
+
+            var calcExtrinsics = new CalculatedCameraExtrinsics();
+            calcExtrinsics.ViewFromWorld = Matrix4x4.TRS(inversePosition, inverseRotation, Vector3.one);
+            calcExtrinsics.Succeeded = succeeded;
+            return calcExtrinsics;
```
wiwei: This section also has some copiage from the function above, consider making some of this common as well.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
```diff
@@ -0,0 +1,389 @@
...
+            int numIntrinsics = 1;
+            int intrinsicsSize = 12;
+            float[] output = new float[numIntrinsics * intrinsicsSize];
+            if (ProcessChessboardIntrinsicsNative(chessSquareSize, output, numIntrinsics))
+            {
+                List<CalculatedCameraIntrinsics> intrinsics = new List<CalculatedCameraIntrinsics>();
+
+                // Note: The undistorted projection matrix is not calculated
+                for (int i = 0; i < numIntrinsics; i++)
+                {
```
wiwei: Nit: It seems a bit odd having a for loop over something of size 1 - if this is because we expect to change this very soon that's fine, but if numIntrisics is only set to 1 to make things more readable (as I assume it is) then probably not doing the for loop would make things clearer about the intent of that variable.

wiwei: Unless I'm missing something here and numIntrinsics is also an outparam for the native API?

---
In reply to: [281373726](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4189#discussion_r281373726) [](ancestors = 281373726)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationDataHelper.cs
```diff
@@ -0,0 +1,250 @@
...
+    internal class CalibrationDataHelper
+    {
+        private const string ChessboardImageDirectoryName = "CHESS";
+        private const string ChessboardDetectedImageDirectoryName = "CHESS_DETECTED";
+        private const string DSLRArUcoDirectoryName = "DSLR_ARUCO";
+        private const string DSLRArUcoDetectedDirectoryName = "DSLR_ARUCO_DETECTED";
+        private const string HeadsetDataDirectory = "META";
+        private const string RootDirectoryName = "Calibration";
+
+        public static void Initialize(out int nextChessboardImageId, out int nextArUcoImageId)
```
wiwei: >nextChessboardImageId [](start = 46, length = 21)

Can you add some comments about the usage/intention of this next ID mechanism - AFAICT it looks like it's based on the number of files in a directory, which doesn't seem to be unique, so if this is truly an identifier I'm a bit confused about how we would prevent ID collisions.

wiwei: Ah, reading more it makes sense what it's doing (i.e. monotonically increasing ID based on the written out filenames).

---
In reply to: [281374506](https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/4189#discussion_r281374506) [](ancestors = 281374506)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationDataHelper.cs
```diff
@@ -0,0 +1,250 @@
...
+            }
+        }
+
+        private static void InitializeDirectory(string rootFolder, string directoryName, string fileExtensionExpected, out int nextImageId)
+        {
+            nextImageId = 0;
+            string directory = Path.Combine(rootFolder, directoryName);
+            if (Directory.Exists(directory))
+            {
+                nextImageId = Math.Max(Directory.GetFiles(directory, fileExtensionExpected, SearchOption.TopDirectoryOnly).Length, nextImageId);
```
wiwei: >Math.Max(Directory.GetFiles(directory, fileExtensionExpected, SearchOption.TopDirectoryOnly).Length, nextImageId); [](start = 30, length = 114)

This could be simplified to nextImageId = Directory.GetFiles(directory, fileExtensionExpected, SearchOption.TopDirectoryOnly).Length.

It's not possible for nextImageId to be anything greater than 0 in this case.

The main reason why I think it this should be simplified is that this code suggests that it's possible for nextImageId to be meaningful up until this point (that it could be > 0) - this is not the case

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CompositorWrapper.cs
```diff
@@ -0,0 +1,49 @@
...
+            return ShaderManager.Instance.compositeTexture; // TODO - figure out what is the correct texture to use here in the new compositor wrapper
+#else
+            return null;
+#endif
+        }
+
+        public static Texture2D GetDSLRTexture()
+        {
+#if COMPOSITOR_PLUGIN_AVAILABLE
+            // Obtain DSLR Imag
```
wiwei: >Imag [](start = 27, length = 4)

Super Nit: Image

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CompositorWrapper.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
+{
+    public class CompositorWrapper
+    {
+        public static RenderTexture GetDSLRFeed()
+        {
+#if COMPOSITOR_PLUGIN_AVAILABLE
+            // Obtain DSLR Feed
+            return ShaderManager.Instance.compositeTexture; // TODO - figure out what is the correct texture to use here in the new compositor wrapper
```
wiwei: >// TODO - figure out what is the correct texture to use here in the new compositor wrapper [](start = 60, length = 90)

Nit: If this TODO is going to be addressed very soon this is fine, if not it's gonna be a dead TODO and future folks probably won't know what to do with it. I guess either expand on it (to explain how to figure this out) or file an issue to track it.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -0,0 +1,305 @@
...
+        public DebugVisualHelper cameraVisualHelper;
+
+        private CalibrationAPI calibration = null;
+        private CalculatedCameraIntrinsics dslrIntrinsics;
+        private Queue<HeadsetCalibrationData> dataQueue = new Queue<HeadsetCalibrationData>();
+        private int nextArUcoImageId = 0;
+        private List<CalculatedCameraExtrinsics> cameraExtrinsics;
+        private CalculatedCameraExtrinsics globalExtrinsics;
+        private List<GameObject> parentVisuals = new List<GameObject>();
+        private int expectedMarkers = 18;
```
wiwei: >private int expectedMarkers = 18; [](start = 8, length = 33)

Nit: Could probably be a static/const (i.e. it's mutable)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -0,0 +1,305 @@
...
+            {
+                matchMakingService.Connect();
+            }
+
+            if (headsetCalibration != null)
+            {
+                headsetCalibration.Updated += OnHeadsetCalibrationUpdated;
+            }
+
+            for(int i = 0; i < nextArUcoImageId; i++)
```
wiwei: >for( [](start = 12, length = 4)

Nit: Can you do a:

s/for\(/for (/g
s/if\(/if (/g

In this file.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -0,0 +1,305 @@
...
+            {
+                dataQueue.Enqueue(headsetData);
+            }
+        }
+
+        private void Update()
+        {
+            if (feedImage != null &&
+                feedImage.texture == null)
+                feedImage.texture = CompositorWrapper.GetDSLRFeed();
```
wiwei: Nit: 

if (feedImage != null &&
                feedImage.texture == null)
{
    feedImage.texture = CompositorWrapper.GetDSLRFeed();
}

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -0,0 +1,305 @@
...
+            {
+                var dslrTexture = CalibrationDataHelper.LoadDSLRArUcoImage(i);
+                var headsetData = CalibrationDataHelper.LoadHeadsetData(i);
+
+                if (dslrTexture == null ||
+                    headsetData == null ||
+                    !ProcessArUcoData(headsetData, dslrTexture))
+                {
+                    Debug.LogWarning($"Failed to process dataset: {i}");
+                }
```
wiwei: How often do you expect these things to actually fail? It looks like if it does fail, we could end up with unused IDs in the middle, where the next time we do some operations we end up re-using old IDs because we're basing the next ID on the number of files in a directory, rather than the actual filenames themselves.

For example, imagine that we have 10 things, we only load the 10th (so let's say now there is file_9.png in the directory).

The next time this runs this will say that the nextID is 1, and will try to use ID space 1-11, and if it succeeds this time it will blow away the old.

Maybe this is fine, but anytime I see ID reuse that is scary.

If in doubt just use the millis timestamp since epoch or something, which is unique and easily understandable.

chrisfromwork:
> // #define QRCODESTRACKER_BINARY_AVAILABLE

Yes, it prevents this script from being compiled by default. This script requires a nuget package that is not yet available for public consumption.

---
In reply to: [489783140](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4189#issuecomment-489783140) [](ancestors = 489783140)

---
Refers to: Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs:7 in e7bc5ca. [](commit_id = e7bc5ca6da570cdfeff9307d0fcd4ea121fe15a6, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -0,0 +1,305 @@
...
+                    Debug.Log("Requesting headset calibration data from VR Headset");
+                    headsetCalibration.UpdateHeadsetCalibrationData();
+                }
+            }
+
+            while (dataQueue.Count > 0)
+            {
+                var data = dataQueue.Dequeue();
+
+                if (dataQueue.Count == 0)
```
wiwei: If we're only ever going to use the most recently queued thing's data, then we probably shouldn't store a queue - just store the last thing.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -0,0 +1,305 @@
...
+            for (int i = 0; i < unityPixels.Length; i++)
+            {
+                unityPixels[i] = pixels[i];
+            }
+
+            dslrTexture.Apply();
+
+            if (lastArUcoImage)
+                lastArUcoImage.texture = dslrTexture;
+
```
wiwei: Nit: 

if (lastArUcoImage)
{
                lastArUcoImage.texture = dslrTexture;
}

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorIntrinsicsCalibration.cs
```diff
@@ -0,0 +1,180 @@
...
+#if UNITY_EDITOR
+        private void Start()
+        {
+            CalibrationDataHelper.Initialize(out nextChessboardImageId, out var nextArUcoImageId);
+            calibration = new CalibrationAPI();
+
+            for (int i = 0; i < nextChessboardImageId; i++)
+            {
+                var texture = CalibrationDataHelper.LoadChessboardImage(i);
+                if(texture == null ||
```
wiwei: > if(texture == null || [](start = 15, length = 22)

Same general comment about if( -> if (

And also down below having:

            if (chessboardHeatmap)
                CalibrationDataHelper.SaveImage(chessboardHeatmap, "ChessboardHeatmap");

            if (chessboardCorners)
                CalibrationDataHelper.SaveImage(chessboardCorners, "ChessboardCorners");

Become:


if (chessboardHeatmap)
{
                CalibrationDataHelper.SaveImage(chessboardHeatmap, "ChessboardHeatmap");
}

if (chessboardCorners)
{
                CalibrationDataHelper.SaveImage(chessboardCorners, "ChessboardCorners");
}

I think this is the last I'll leave of this but I will hope that this can be fixed up in the rest of the codebase

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorIntrinsicsCalibration.cs
```diff
@@ -0,0 +1,180 @@
...
+            int imageWidth = texture.width;
+            int imageHeight = texture.height;
+
+            if (chessboardCorners == null)
+            {
+                chessboardCorners = new Texture2D(imageWidth, imageHeight, TextureFormat.RGB24, false);
+                for (int y = 0; y < chessboardCorners.height; y++)
+                {
+                    for (int x = 0; x < chessboardCorners.width; x++)
+                    {
```
wiwei: This is a really weird question but maybe mentally I'm so used to going where I iterate over x first, then y, then z (or i -> j -> k) - is this order primarily because you get some sort of perf benefits of memory localility, or is it also equally arbitrary like what I usually do?

chrisfromwork: I think that these values are stored in rows compared to columns, so my assumption would be that accessing subsequent x values would have better performance than different y values. That going said, this was just an arbitrary choice.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs
```diff
@@ -0,0 +1,267 @@
...
+        [SerializeField]
+        protected DebugVisualHelper arucoDebugVisualHelper;
+
+        private bool markersUpdated = false;
+        private Dictionary<int, Marker> qrCodeMarkers = new Dictionary<int, Marker>();
+        private Dictionary<int, GameObject> qrCodeDebugVisuals = new Dictionary<int, GameObject>();
+        private Dictionary<int, GameObject> arucoDebugVisuals = new Dictionary<int, GameObject>();
+        private readonly float markerPaddingRatio = 34f / (300f - (2f * 34f)); // padding pixels / marker width in pixels
+        private Dictionary<int, MarkerCorners> qrCodeMarkerCorners = new Dictionary<int, MarkerCorners>();
+        private Dictionary<int, MarkerCorners> arucoMarkerCorners = new Dictionary<int, MarkerCorners>();
```
wiwei: This is a nit and now that it's been written feel free to wontfix/ignore this - in general if we're gonna be storing multiple things on the same key, we should probably just store all of the things associated with that key on a sturct thing - so that we don't have to do 2-3 different hashes to do the samething (see lines 66-67 for where we are doing multiple hashes).

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
chrisfromwork: on this should just be zero

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/HeadsetCalibration.cs
```diff
@@ -0,0 +1,267 @@
...
+        private static void MergeDictionaries(Dictionary<int, Marker> dictionary, Dictionary<int, Marker> update)
+        {
+            HashSet<int> observedMarkers = new HashSet<int>();
+            foreach(var markerUpdate in update)
+            {
+                dictionary[markerUpdate.Key] = markerUpdate.Value;
+                observedMarkers.Add(markerUpdate.Key);
+            }
+
+            RemoveItems(dictionary, observedMarkers);
```
wiwei: >RemoveItems [](start = 12, length = 11)

The naming of this function is extremely confusing.

It says "remove items" so I would expect this would remove observedMarkers from dictionary.

However, observedMarkers is actually kept.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
chrisfromwork: this is old behavior from when i was testing multiple intrinsic calculation processes. For now this should be reduced to one. I likely should remove the num instances for multiple plugin related items.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationDataHelper.cs
chrisfromwork: given this is an internal class I likely won't spend a great deal of time on updating this documentation. I don't think I will address this for now.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/HeadsetRequestHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+        [Tooltip("Used to display the last request timestamp.")]
+        [SerializeField]
+        private Text lastRequestTimestampText;
+
+        private bool updateData = false;
+        private HeadsetCalibrationDataRequest request = null;
+
+        private void OnValidate()
+        {
+#if UNITY_EDITOR
```
wiwei: Thoughts on moving this around OnValidate() so that it's purely an editor thing?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CompositorWrapper.cs
chrisfromwork: I"ll change it to note. This has to be changed before its used by anyone externally. The shader manager may not exist in the future.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        /// <summary>
+        /// Timeout interval for establishing connections
+        /// </summary>
+        [Tooltip("Timeout interval for establishing connections")]
+        [SerializeField]
+        protected TimeSpan timeoutInterval = TimeSpan.Zero;
+
+        private bool runAsUser = false;
+        private Dictionary<string, string> playerIds = new Dictionary<string, string>();
```
wiwei: >private Dictionary<string, string> playerIds = new Dictionary<string, string>() [](start = 8, length = 79)

Nit: Can you add a comment about what this dictionary is? i.e. seeing a string to string map and I wonder what it is.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        /// <inheritdoc />
+        public event DataHandler DataReceived;
+
+        private void Start()
+        {
+#if UNITY_WSA && !UNITY_EDITOR
+            runAsUser = true;
+#elif UNITY_ANDROID || UNITY_IOS || UNITY_EDITOR
+            runAsUser = false;
```
wiwei: I guess this list could grow in the future - I guess maybe just get rid of this because we default to false? Otherwise it's a bit weird in the future where we aren't handling certain flags but really the intent is to only have runAsUser on WSA

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
```diff
@@ -0,0 +1,188 @@
...
+        public void Connect()
+        {
+            if (net == null)
+            {
+                Debug.LogError("TCPConnectionManager not set for SocketNetworkingService. Unable to establish connection");
+                return;
+            }
+            else
+            {
+                net.DisconnectAll();
```
wiwei: > net.DisconnectAll(); [](start = 15, length = 21)

Super dumb question for me - does this also drop the OnConnected registration below (or does the eventing mechanism prevent double/triple registrations?)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        private void OnDisconnected(SocketEndpoint obj)
+        {
+            string playerId = string.Empty;
+            lock (playerIds)
+            {
+                if (playerIds.ContainsKey(obj.Address))
+                {
+                    playerId = playerIds[obj.Address];
```
wiwei: >playerId = playerIds[obj.Address]; [](start = 20, length = 34)

Should this also be getting removed from the dictionary at this point? Or do you want to be storing them forever for other reasons (i.e. still want to handle OnRecieve after disconnection?)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
```diff
@@ -0,0 +1,188 @@
...
+                {
+                    playerId = playerIds[obj.Endpoint.Address];
+                    validId = true;
+                }
+            }
+
+            if (validId)
+            {
+                DataReceived?.Invoke(playerId, obj.Data);
+            }
```
wiwei: Why does this function use a validId bool (but other ones above check against string.Empty)? 

Also, this uses String.Empty but things above use string.Empty.

wiwei:
Generally looks good to me, let me know when you want me to look again.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
chrisfromwork: I will address this in a future review. We need to start enabling UNITY_WSA as a spectator in some instances in the immediate future.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
chrisfromwork: In some instances the same address will reconnect and send payloads before the connection event occurs. To avoid having these messages dropped, I intentionally kept id definitions for the same address.

Assets/MixedRealityToolkit.Extensions/PhotoCapture/CameraExtrinsics.cs
```diff
@@ -16,5 +16,12 @@ public class CameraExtrinsics
         /// Camera's view from world matrix
         /// </summary>
         public Matrix4x4 ViewFromWorld = Matrix4x4.identity;
+
+        public override string ToString()
+        {
+            var position = ViewFromWorld.GetColumn(3);
```
andreiborodin: >var [](start = 12, length = 3)

nit: Use explicit types.

Assets/MixedRealityToolkit.Extensions/PhotoCapture/CameraExtrinsics.cs
```diff
@@ -16,5 +16,12 @@ public class CameraExtrinsics
...
+            var rotation = Quaternion.LookRotation(ViewFromWorld.GetColumn(2), ViewFromWorld.GetColumn(1));
+            return $"Position: {position.ToString()}, Rotation: {rotation.ToString()}";
```
andreiborodin: >ToString() [](start = 41, length = 10)

Consider specifying precision for displaying, ie. "G4"

andreiborodin: >rotation.ToString() [](start = 65, length = 19)

Consider precision, but also whether you want to use .eulerAngles

Assets/MixedRealityToolkit.Extensions/PhotoCapture/CameraIntrinsics.cs
```diff
@@ -80,5 +80,21 @@ public class CameraIntrinsics
...
+            PrincipalPoint = Vector2.zero;
+            RadialDistortion = Vector3.zero;
+            TangentialDistortion = Vector2.zero;
+            UndistortedProjectionTransform = Matrix4x4.identity;
+        }
+
+        public override string ToString()
+        {
+            return $"Focal Length:{FocalLength}, Principal Point:{PrincipalPoint}, Image Width:{ImageWidth}, Image Height:{ImageHeight}, Radial Distortion:{RadialDistortion}, Tangential Distortion:{TangentialDistortion}";
+        }
```
andreiborodin: Consider precisions here.

andreiborodin:
> // #define QRCODESTRACKER_BINARY_AVAILABLE

Ok, but don't we generally want to specify the PREPROCESSORS in Unity player settings?

---
In reply to: [489801408](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4189#issuecomment-489801408) [](ancestors = 489801408,489783140)

---
Refers to: Assets/MixedRealityToolkit.Extensions/QRCodesTracker/Scripts/QRCodesManager.cs:7 in e7bc5ca. [](commit_id = e7bc5ca6da570cdfeff9307d0fcd4ea121fe15a6, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -0,0 +1,341 @@
...
+        private void Update()
+        {
+            if (feedImage != null &&
+                feedImage.texture == null)
+            {
+                feedImage.texture = CompositorWrapper.GetDSLRFeed();
+            }
+
+            if (Input.GetKeyDown(KeyCode.Space))
```
provencher: Not a fan of hardcoding inputs like this.

The MRTK uses a platform input data provider abstraction. It might be best to run this stuff on an event response.

Further, this code will run continuously if you hold the space bar. That might not be ideal.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
```diff
@@ -0,0 +1,341 @@
...
+                    GameObject hololens = null;
+                    cameraVisualHelper.CreateOrUpdateVisual(ref hololens, Vector3.zero, Quaternion.identity);
+                    hololens.name = "Global HoloLens";
+                }
+            }
+        }
+
+        private void OnDataReceived(string playerId, byte[] payload)
+        {
+            Debug.Log($"Received payload of {payload.Length} bytes");
```
provencher: Might be good to funnel logs through a method and add a property that allows you to toggle them

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
chrisfromwork: no we should unregister here as well.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SocketerNetworkingService.cs
chrisfromwork: I'm going to keep this for OnReceive after disconnection. If we unsubscribe we hopefully won't hit this issue, but i have seen some stranger behavior here where messages get received before a reconnect if the client/server disconnects.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
chrisfromwork: If i'm interpreting Unity's documentation correctly, GetKeyDown will only get called for the first frame the button is pressed: https://docs.unity3d.com/ScriptReference/Input.GetKeyDown.html

Right now this code is intended to run in the editor and isn't meant for an end user application. We also don't want this code to take any dependencies on the foundational/core MRTK components, which may be required for the MRTK input abstraction.

I think for now this change may be beyond the scope of this review. This initial checkin will still need some reworking once qr code detection dlls are available to the public.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalibrationAPI.cs
chrisfromwork: Hm, subsequent initializations clear the data. We should probably change this to be a static access that has a clear data plugin function compared to muddling the two functionalities.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
chrisfromwork: This seems like a good suggestion for the future but it is probably beyond the scope of what i would like to get in with this review.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/EditorExtrinsicsCalibration.cs
chrisfromwork: i have it at seconds right now. we can change it to something more robust if needed, but we should now avoid using random indices.

PR #4191, opened 2019-05-06
Code cleanup, part 1 of N.

wiwei:
I'm going through the code to better understand things. As part of reading the code, I'm seeing some occasional typos/issues/areas for cleanup (i.e. stale comments, dead code).

Changes below include:

- Adding some comments to things like Distorters (some of them aren't used and *could* be deleted, but they seem to be useful to have as a package)
- Something are deleted that are not used (i.e. some Interlator* things, some utilites and CalibrationSpace things)
- Some code cleanup standardizing on Mathf.Clamp (instead of doing manual if/else checks).


wiwei:
Okay good to go.

julenka:
Why are we deleting DebugUtilities.cs?

wiwei:
> Why are we deleting DebugUtilities.cs?

It's not used in the codebase.

julenka:
Oh no please don’t delete that, I find it super useful to draw the debug points when I’m debugging stuff.

Sent from phone, please excuse typos.

keveleigh:
>I find it super useful to draw the debug points when I’m debugging stuff.

In the same vein, the interpolator scripts seem like useful utilities for devs building on the Toolkit, even if they aren't used directly within the MRTK.

wiwei:
> I avoided using Mathf.Clamp because there was a minor perf hit. Same with using PointCount instead of a const. But it was very minor and the cleaner code is probably worth the hit.

Do you have a pointer to more info on that? i.e. was it like single digit percent difference, or was it significant?

wiwei:
> > I avoided using Mathf.Clamp because there was a minor perf hit. Same with using PointCount instead of a const. But it was very minor and the cleaner code is probably worth the hit.
> 
> Do you have a pointer to more info on that? i.e. was it like single digit percent difference, or was it significant?

i.e. I'd like to keep the code as-is if this if minor is still noticeable, and then update the coding guidelines to mention this.

wiwei:
> > I find it super useful to draw the debug points when I’m debugging stuff.
> 
> In the same vein, the interpolator scripts seem like useful utilities for devs building on the Toolkit, even if they aren't used directly within the MRTK.

I'll keep the debug utilities (I will delete the DebugUtilites.Log* because they don't really provide value as-is).

For the interpolators, they seem useful, but without documentation or usage in our code, I don't really understand how someone will want to take the time to figure out how the code works and then start using it.

The fundamental problem with dead code is that it will generally be "this could be useful" but without any coverage (i.e. actual usage) or documentation (describing how it works) it's super easy to let it rot (and then when someone tries to use it later it could not work (and at that point no one may know how it was supposed to work because didn't have coverage or docs)

wiwei:
Removed the DebugUtilities.cs* files from the commits, so those have been left alone

PR #4225, opened 2019-05-07
update spatial observers and demo to use registry

davidkline-ms:
## Overview
- Update the base spatial observer to honor the spatial awareness system instance passed in the constructor.
- Remove the duplicate SpatialAwarenessSystem property from the WindowsMixedRealitySpatialMeshObserver.
- Update the spatial awareness demo to retrieve the service instance from the registry,

## Changes
This change builds upon #4099 and is related to #3545.

### DIFF
--- a/Assets/MixedRealityToolkit/Providers/BaseSpatialObserver.cs
+++ b/Assets/MixedRealityToolkit/Providers/BaseSpatialObserver.cs
@@ -24,17 +24,17 @@ namespace Microsoft.MixedReality.Toolkit.SpatialAwareness
             uint priority = DefaultPriority, 
             BaseMixedRealityProfile profile = null) : base(registrar, spatialAwarenessSystem, name, priority, profile)
         {
-            if (MixedRealityToolkit.SpatialAwarenessSystem != null)
-            {
-                SourceId = MixedRealityToolkit.SpatialAwarenessSystem.GenerateNewSourceId();
-            }
-            else
-            {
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit/Providers/BaseSpatialObserver.cs
```diff
@@ -24,18 +24,18 @@ public class BaseSpatialObserver : BaseDataProvider, IMixedRealitySpatialAwarene
...
-            {
-                SourceId = MixedRealityToolkit.SpatialAwarenessSystem.GenerateNewSourceId();
-            }
-            else
-            {
-                Debug.LogError($"A spatial observer is registered in your service providers profile, but the spatial awareness system is turned off. Please either turn on spatial awareness or remove {name}.");
-            }
+            SpatialAwarenessSystem = spatialAwarenessSystem;
 
+            SourceId = (SpatialAwarenessSystem != null) ? SpatialAwarenessSystem.GenerateNewSourceId() : 0;
```
wiwei: This drops the LogError - was this old error not helpful/we don't want to preserve it? Also, is there a concept of a globally non-valid ID (i.e. is zero just understood to be a non-valid ID)?

davidkline-ms: from what i have seen elsewhere in the code, 0 is essentially the non-valid ID.

and the LogError was not useful in this case

PR #4227, opened 2019-05-07
Code cleanup, 2 of N

wiwei:
Was going through the code to learn more, found some other small things for cleanup, so cleaning them up here.

General changes:
m_* -> not m_* (this is a C++ like coding convention which doesn't follow our C# guidelines)
InteractableStateModel.cs -> comments removed because they looked like there were notes-to-self and didn't help a reader understand what was going on (I'm more confused reading it).
BaseCursor -> It powers non-gaze cursors as well.
Everything else - general comment cleanup

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/InteractableStateModel.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/InteractableStateModel.cs
@@ -8,18 +8,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
     /// <summary>
     /// State data model, state management and comparison instructions
     /// </summary>
-
-    /*
-     * Have an enum with all the button states -
-     * Create a list using the enums as the state type -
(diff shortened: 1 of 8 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/InteractableStateModel.cs
```diff
@@ -8,18 +8,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
     /// <summary>
     /// State data model, state management and comparison instructions
     /// </summary>
-
-    /*
-     * Have an enum with all the button states -
```
keveleigh: This feels like information that might be useful to have *somewhere*.

wiwei: @killerantz Is there a rewording here that would be a little more helpful as a doccomment? The text here reads somewhat as dev notes to self, and as a reader I don't really know what to make of this.

PR #4230, opened 2019-05-07
Activate PressableButton on press by default, not on release

julenka:
## Overview
Our default button prefabs currently invoke Interactable.OnClick events when the button gets unpressed ("On Click Completion"), instead of when a button gets pressed.

This is currently causing many buttons to appear difficult to press by partner teams, for a few reason:
1. People do not realize that they need to 'unpress' the button to activate (evidence: user research in OOBE).
2. We have some bugs in pressable button. For example, if the finger exits the collider on the same frame as an unpress happens, Interactable.OnClick does not get invoked.

Activating buttons on release is only necessary in some cases, such as for buttons on scrollable panels, or for buttons that support press and hold. To unblock customers and allow for most reliable click experience, allow for the easiest possible button click experience which activates buttons on press, not on release.

This PR is based on feedback from partner teams, Kevin Collins, and Eric Havir.

# Changes
Prefabs
- Modify PressableButton and PressableButtonPlated event router to activate on press, not release.
- Modify HandInteractionExample scene to have buttons activate on press.

Misc
- Add function in ColorChanger to change object to random color. This allows for testing of multiple 'click' type behavior since it is user visible change that will be different every time.

# Validation
- Manually tested in editor, and via example scene.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButtonPlated.prefab
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButtonPlated.prefab
@@ -19,6 +19,10 @@ PrefabInstance:
       propertyPath: source
       value: 
       objectReference: {fileID: 2204069621426241314}
+    - target: {fileID: 316800720, guid: 45fd0ad89d6d17b4fbe68eb48dbe9de9, type: 3}
+      propertyPath: InteractableOnClick
+      value: 1
+      objectReference: {fileID: 0}
     - target: {fileID: 316800721, guid: 45fd0ad89d6d17b4fbe68eb48dbe9de9, type: 3}
       propertyPath: handlerTarget
       value: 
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButtonPlated.prefab
```diff
@@ -128,7 +132,7 @@ PrefabInstance:
         type: 3}
       propertyPath: m_Materials.Array.data[0]
       value: 
-      objectReference: {fileID: 2100000, guid: e676303390b2ee941b658c1d04ccdc57, type: 2}
```
keveleigh: Is this a reference we don't need anymore? I'm not sure exactly what it's referring to.

julenka: The following command yields no results, so I believe that the reference is no longer valid.
> git grep -i e676303390b2ee941b658c1d04ccdc57

Issue #4195, opened 2019-05-06
Inconsistency in eye tracking properties


jbienzms:
## Describe the bug

When viewing GazeProvider in the Unity inspector there is a option labeled "Prefer Eye Tracking". However, in the [GazeProvider](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs) code this field is exposed as the property **UseEyeTracking**. These are inconsistent.

MRTK will fall back from eye tracking to head tracking if:

- The application is missing permissions (see #4194)
- The user did not approve permissions
- The device is not calibrated for the user

Since MRTK will silently fall back to head tracking in these situations, the property should be renamed to **PreferEyeTracking** to match the inspector.

## Additional context

I'm aware this is a breaking change and I am labeling this issue as such. However, it's unlikely many developers are currently setting this property in code and I believe consistency with the inspector and clarity around what the property actually does justify the breaking change.

wiwei:
Got around to looking at this one - as mentioned above, this is a breaking change, not just because it's a public property on GazeProvider.cs, but also because it's a public property on the interface IMixedRealityEyeGazeProvider.cs

Given that the value is primarily in consistency (and yes, prefer definitely a more correct term to use), it's somewhat difficult to argue the value vs break proposition here. The chance here is super low, but we could also make this consistent just by naming the underlying property useEyeTracking (and give it a better tooltip).

I'm probably going to go with that safer option here.

wiwei:
Removed the breaking change label, because we took a non-breaking route here.

PR #4232, opened 2019-05-08
Update the "Use Eye Tracking" properties to be consistent with each other

wiwei:
As https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4195 calls out, the inspector has a term called "Prefer Eye Tracking" but elsewhere it's referred to as "UseEyeTracking." Renaming everything to PreferEyeTracking would actually be more proper (because it's a more accurate description of how things work)

However, renaming things to PreferEyeTracking would be a breaking change (where the main value here is consistency/term correctness). It's safer for us to keep things consistent by renaming the serialized property (which supports backwards compat and is not a breaking change)

https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4195


Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityEyeGazeProvider.cs
```diff
@@ -17,8 +17,13 @@ public interface IMixedRealityEyeGazeProvider : IMixedRealityGazeProvider
         bool IsEyeGazeValid { get; }
 
         /// <summary>
-        /// True to provide eye tracking, when available.
+        /// If true, eye-based tracking will be used when available.
         /// </summary>
+        /// <remarks>
+        /// The usage of eye-based tracking depends on having the Gaze Input permission set
+        /// and user approved, along with proper device eye calibration. This willback to head-based
```
keveleigh:
```suggestion
        /// and user approved, along with proper device eye calibration. This will fallback to head-based
```

Issue #4239, opened 2019-05-08
Remove SupportedControllerEnum


Alexees:
## Describe the problem

This enum has a fixed number of controllers. Any new device (Oculus Go) needs to define its own controller there. increasing the list of controllers is crucial when you want to add more controllers.

## Describe the solution you'd like

Like #4239, the enum should be replaced with classes instead. A lot of the code is converting between enum and actual class type already, so why not just go with classes altogether?

## Describe alternatives you've considered

None considered, seems the best option.

Railboy:
@Alexees I think you meant to link to #4235 

Really like this proposal as well.

keveleigh:
#9106 made it in as a first step in this direction for 2.6, adding the concept of "controller definitions" to eventually replace this enum.

A reworked profile using these definitions and deprecation of this enum will come in a future release.

davidkline-ms:
Excellent news! MRTK3 no longer uses a SupportedControllerEnum! We are using Unity's input system and supporting a new controller is effectively automatic.

PR #4240, opened 2019-05-08
Removed SupportedControllerType enum

Alexees:
## Overview
Removes the SupportedControllerType enum altogether and uses the actual classes everywhere. A system for serializing System.Type has already been established and most uses are statically typed into the controllers corresponding attributes.

## Changes
- Fixes: #4239 .


chrisfromwork:
Thank you for putting out this review.

This change seems like it is an improvement and will make the MRTK more extensible, but removing the use of this enum type feels like a breaking change. Custom MixedRealityController definitions will likely have to be updated based on this refactor.

Right now with RC1 and RC2, its a bit unclear whether we will be taking any breaking changes. I'll make sure that this pull request gets addressed by contributors with more context so that we can get this change the proper attention it deserves.

davidkline-ms:
I definitely like the change to using typeof(class). Perhaps we should consider that change and add the obsolete attribute to the enum.

StephenHodgson:
@Alexees I'm more than happy to take this change in the xrtk asap.

Would you like to open a similar PR there?

Alexees:
if you could give me a hand on how I get all changed files over there, I would. I'm not the most advanced git dude.

StephenHodgson:
The git histories aren't the same so a straight cherry pick wouldn't work.

Easiest would probably be to do a diff against each file with a tool, and commit that to a new branch.

If you feel like it's too much work, I could do it no problem, but I'd prefer your name to be on the commit changes.

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkit.Core.Inspectors.asmdef
```diff
@@ -6,7 +6,8 @@
         "Microsoft.MixedReality.Toolkit.Core.Build",
         "Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities.Editor",
         "Microsoft.MixedReality.Toolkit.Core.Utilities.Editor",
-        "Microsoft.MixedReality.Toolkit.Core.Utilities.Async"
+        "Microsoft.MixedReality.Toolkit.Core.Utilities.Async",
+        "Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality"
```
julenka: I'm noticing that after this change, several .asmdef files are including a dependency on Providers.WindowsMixedReality @davidkline-ms  and @keveleigh I don't know much the remifications of including this dependency, is this something that could bite us down the line?

Assets/MixedRealityToolkit.Providers/OpenVR/OculusTouchController.cs
```diff
@@ -8,7 +8,7 @@
 namespace Microsoft.MixedReality.Toolkit.OpenVR.Input
 {
     [MixedRealityController(
-        SupportedControllerType.OculusTouch,
+        typeof(OculusTouchController),
```
StephenHodgson: Kinda odd that it's passing it's own type to itself.

Assets/MixedRealityToolkit/Definitions/InputSystem/PointerOption.cs
```diff
@@ -19,23 +19,23 @@ public struct PointerOption
...
-        private SupportedControllerType controllerType;
+        [Implements(typeof(IMixedRealityController), TypeGrouping.ByNamespaceFlat)]
+        private SystemType controllerType;
 
         /// <summary>
         /// The type of Controller this pointer will attach itself to at runtime.
         /// </summary>
         /// <remarks>If <see cref="Microsoft.MixedReality.Toolkit.Utilities.Handedness.None"/> is selected, then it will attach to any controller type</remarks>
-        public SupportedControllerType ControllerType => controllerType;
+        public Type ControllerType => controllerType;
```
StephenHodgson: This should by `SystemType` as well to match the backing field.

StephenHodgson: Likely the same goes for all the usages of this property as well.

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkit.Core.Inspectors.asmdef
StephenHodgson: Agreed, I'm also not sure this dependency is a good one.

Assets/MixedRealityToolkit.Providers/OpenVR/OculusTouchController.cs
Alexees: I know, just converted it in a first step. If you know how to get the attribute decorated type I'd gladly implement it.

Assets/MixedRealityToolkit/Definitions/InputSystem/PointerOption.cs
Alexees: working on it ;)

Assets/MixedRealityToolkit.Providers/OpenVR/OculusTouchController.cs
StephenHodgson: I'm not sure adding this attribute here was even a good idea. The whole point of the profiles was to hold the data for the implementation, not decorate the implementation with data 😅 

Awe well. ¯\_(ツ)_/¯

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkit.Core.Inspectors.asmdef
davidkline-ms: @julenka, there should be zero dependencies from the core to any provider or service. Dependencies should go the other direction.

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/MixedRealityToolkit.Services.InputSimulation.Editor.asmdef
```diff
@@ -4,7 +4,8 @@
         "Microsoft.MixedReality.Toolkit",
         "Microsoft.MixedReality.Toolkit.Core.Inspectors",
         "Microsoft.MixedReality.Toolkit.Core.Utilities.Editor",
-        "Microsoft.MixedReality.Toolkit.Services.InputSimulation"
+        "Microsoft.MixedReality.Toolkit.Services.InputSimulation",
+        "Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality"
```
davidkline-ms: there should not be dependencies between services and platforms. was there a reason the WMR provider needed to be added?

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityToolkit.Services.InputSimulation.asmdef
```diff
@@ -2,7 +2,8 @@
     "name": "Microsoft.MixedReality.Toolkit.Services.InputSimulation",
     "references": [
         "Microsoft.MixedReality.Toolkit",
-        "Microsoft.MixedReality.Toolkit.SDK"
+        "Microsoft.MixedReality.Toolkit.SDK",
+        "Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality"
```
davidkline-ms: same comment re: WMR provider dependency

wiwei:
Thanks for sending out this PR!

A few comments here

I pulled the PR locally to try things out, and one of the things that I was trying to keep an eye out for was how these controller types are configured in the inspector:

image

It should look something like this:
image

I also tried just opening up some common scenes we use for testing (i.e. HandInteractionExamples) and tried to see if some basic scenarios (i.e. pressing play  -> pressing spacebar to invoke hand simulation -> seeing if hand simulation works). This was broken, I assume because with this change, it necessitates regenerating/updating all of the .unity and .asset files that have references to these previous types.

If we wanted to proceed with this change, we'd probably need to spend some time iterating on creating an inspector drawer (i.e. guess like Assets\MixedRealityToolkit\Inspectors\PropertyDrawers\EnumFlagsAttributeDrawer.cs) that will properly render these options in the inspector.

Since this is a breaking change in more than a code basis (i.e. existing customers who have created profiles will now see their own profiles broken, and fixing that will be a highly manual process for them especially if they've taken a binary form of their stored .unity files/.asset files), I think before this went in, we'd also want to have a plan for how we could migrate those existing customers so that they don't have to experience that pain (Having done this before, it is SUPER easy to mess this up when done manually, and not something to be taken lightly).

We've been in this "no breaking change" mode for a couple of months now, and although we have not enforced that 100%, we've gotten stronger about doing this as time goes on (because we'll be doing RC2 soon, more customers are onboarding, and the cost of breaking changes is actually getting higher now, especially if the breaking changes are in profiles they may have cloned). I'm hesistant to take this change because it's not just a code break (i.e. class deleted) but will also require customers that we know that have already set up their own profiles to take a fairly painful change (unless again, we do the work to do auto migration).

Here is concretely what I think the next steps should be:

1) First recognize that it is amazing and that I am 100% thankful that you're taking the time to contribute here. Please don't take this as any reflection on anything you're doing, as there is literally a timing aspect that is outside of your control (and FWIW to some degree outside of our control because of how folks are starting to use this).

2) I would like to close to this PR, discuss on the issue how we can move forward with this. I think there may be a solution where we have a 'parallel' stopgap (i.e. have another attribute/section of the pointer profile that lets you do the extension stuff you want, while not breaking existing customers). We could then, once breaking changes are no longer a problem, migrate the old stuff to the new stuff, and everyone is happy.

3) Once everyone is happy with the tradeoffs, we can then move onto the PR phase and we'll help with testing and validation like I've done here.

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkit.Core.Inspectors.asmdef
Alexees: It was due to this change here:
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/246dce4c233aaa6350393a5e2a70efeaa94a6887/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityControllerMappingProfileInspector.cs#L299

It was previously using the removed enum to identify MixedRealityControllers, which is now the concrete type. It's only used to give a different controller name in the inspector in case the handedness is none, then it's supposed to be a Hololens.
I wasn't able to find where to put that distinction but in my opinion this shouldn't be part of any inspector.
Maybe the Hololens is just a different controller instead, then this wasn't necessary.

Assets/MixedRealityToolkit.Providers/OpenVR/OculusTouchController.cs
Alexees: I removed all these in the latest commit

Assets/MixedRealityToolkit/Definitions/InputSystem/PointerOption.cs
Alexees: done

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/MixedRealityToolkit.Services.InputSimulation.Editor.asmdef
Alexees: resolved itself

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityToolkit.Services.InputSimulation.asmdef
Alexees: same here

Alexees:
> I assume because with this change, it necessitates regenerating/updating all of the .unity and .asset

yes, I did that for the DefaultMixedRealityInputPointerProfile, which I assume wasn't ready the moment you checked it, sorry for that.

> we'd probably need to spend some time iterating on creating an inspector drawer

This was the reason all of the pointers were empty. I did a first version of it, you can check it out now.

> We've been in this "no breaking change" mode for a couple of months now

I really like and need to contribute to the toolkit. Since even the mrtk_development branch does not take breaking changes, which branch does?

> I would like to close to this PR, discuss on the issue how we can move forward with this

This actually frustrates me. Then at least provide a branch where I and everyone who is interested in fast changes can contribute to (Like the Windows 10 insider preview.... what was is called, fast lane?)

Alexees:
> The git histories aren't the same so a straight cherry pick wouldn't work.
> 
> Easiest would probably be to do a diff against each file with a tool, and commit that to a new branch.
> 
> If you feel like it's too much work, I could do it no problem, but I'd prefer your name to be on the commit changes.

@StephenHodgson According to @wiwei this doesn't seem to be the accepted solution so moving it over into the XRTK might not be a good idea then. Since I have no idea who is in charge, I just assume that that's how it's going to be.

luis-valverde-ms:
> @luis-valverde-ms do you remember why we might have wanted to stick with the SupportedControllerType enum?
I do not @julenka , maybe to avoid breaking changes.

wiwei:
Closing this PR out per the comments above.

Because this change involves breaking changes to profiles (which are extremely hard for customers to fix up, speaking from personal experience doing merges and from direct feedback from other customers.)

We'll want to re-open this once we're done with V2 and open again to breaking changes.

davidkline-ms:
For additional context, we are actively striving to avoid changes that break customer's customized profiles in v2. For the next major release we are considering revising the configuration profile system and providing tooling to migrate from v2 to the new version.

Issue #4175, opened 2019-05-03
TryGetJointPose only works on Handedness.Right/Handedness.Left but not Any


oitoito:
## Describe the bug
TryGetJointPose only gives a result on Handedness.Right/Handedness.Left but not Handedness.Any.

Using Handedness.Any as parameter will always return false. In the documentation it is pointed out that it should be placed in an if statement implying that true means the joint pose has been acquired and false otherwise. 

## To reproduce
``
// is always false
  if (HandJointUtils.TryGetJointPose(TrackedHandJoint.IndexTip, Handedness.Any, out pose))
            {
              //..

            }

           // will always go here
            else
            {
               // ..

            }

// trying again with right and left will never return false when hand is visible
            if ( HandJointUtils.TryGetJointPose(TrackedHandJoint.IndexTip, Handedness.Right, out pose))
            {
              //..
            }

            else if (HandJointUtils.TryGetJointPose(TrackedHandJoint.IndexTip, Handedness.Left, out pose))
            {
              //..

            }
``

## Expected behavior

Handedness.Any should find the first hand available (or all?), return true if anything is found, and populate out variable, same as Handedness.Left/Right do

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.11f1]
- MRTK Version [e.g. v2.0]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

chrisfromwork:
@wiwei is going to update the documentation for this behavior. It's unlikely that there will be a change in this behavior.

wiwei:
The intent of this function as I understand it was to provide the joint information for a specific hand (i.e. user wants the right hand, user wants the left hand). 

Handling "Any" or "Both" puts it in this weird case where it's non-deterministic which hand gets returned (I guess we could always default to a specific one, but without a good understanding of the use cases it's tough to say whether or not that would actually make sense).

In order to address this, we'll update the docs to describe what they actually take so there's no future confusion about the set of supported values.

If there's a specific behavior you're looking for (i.e. I want this API to support Any/Both, and default to left or right hand, or the last one that sent input, or the 'primary' one), it would be good to open a new issue describing what that request is and what it would enable.

wiwei:
Alrighty, it looks like we updated this to work with Any as well, so this should be good.

PR #4244, opened 2019-05-08
Update the TryGetJointPose docs to accurately describe the set of inputs

wiwei:
These functions only work if you pass in the specific Handedness.Left or Handedness.Right. They should be documented as such.

https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4175


lukastoenneMS:
IMO ideally we should have a pure `enum Handedness { Left, Right }`, and then a `enum HandednessFlags {...}` with mask bits where it makes sense. I.e. use the strict left/right enum when either one or the other is expected (which is the majority of cases), and the flag where any combination is possible. That should avoid a lot of the current confusion when `Handedness.Both` etc. don't work.

PR #4248, opened 2019-05-09
Reset touchPoints on Enable on PressableButtons

AsoboStudio:
If a pressable button is disabled before the OnTouchCompleted event, the button remains in an unusable state as the touchPoints controller is never resetted

to be sure, the button is always in a usable state, the touchpoint list is always resetted on OnEnable


julenka:
Thank you for making this PR! Actually this fix has already been made by @thalbern in the mrtk_development branch, but it was done in OnDisable.

Have you tried pulling the latest mrtk_development code to see if the issue still exists?

wiwei:
Closing this out given @julenka's comments above. Feel free to reopen if @thalbern's fix didn't address what you were looking for

PR #4253, opened 2019-05-09
Add class for reading new calibration data

chrisfromwork:
This review contains the following logic:
1) calculated camera intrinsics/extrinsics serialization is moved within their class
2) An argument name that was a bit unclear has been renamed in ICalibrationData
3) A new class implementing ICalibrationData is added for applying camera intrinsics/extrinsics obtained through the new spectator view calibration process.

### DIFF
--- a/Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalculatedCameraExtrinsics.cs
+++ b/Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalculatedCameraExtrinsics.cs
@@ -28,7 +28,7 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
         public byte[] Serialize()
         {
             var str = JsonUtility.ToJson(this);
-            var payload = Encoding.ASCII.GetBytes(str);
+            var payload = Encoding.UTF8.GetBytes(str);
             return payload;
         }
 
@@ -38,7 +38,7 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
 
             try
             {
-                var str = Encoding.ASCII.GetString(payload);
+                var str = Encoding.UTF8.GetString(payload);
                 extrinsics = JsonUtility.FromJson<CalculatedCameraExtrinsics>(str);
                 return true;
             }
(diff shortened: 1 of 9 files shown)


Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalculatedCameraExtrinsics.cs
```diff
@@ -22,5 +24,28 @@ public override string ToString()
         {
             return $"Succeeded: {Succeeded} {base.ToString()}";
         }
+
+        public byte[] Serialize()
+        {
+            var str = JsonUtility.ToJson(this);
+            var payload = Encoding.ASCII.GetBytes(str);
```
matthejo: To avoid surprises if content ends up containing Unicode characters, it might be safer to use Encoding.UTF8 for all of these string transfers. UTF8 is the default encoding that BinaryWriter and BinaryReader use for string encoding.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Compositor/CalibrationData.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        /// <inheritdoc />
+        public void SetUnityCameraExtrinstics(Transform cameraTransform)
+        {
+            Matrix4x4 headFromCamera = cameraExtrinsics.ViewFromWorld;
+            cameraTransform.transform.localPosition = cameraExtrinsics.ViewFromWorld.GetColumn(3);
+            cameraTransform.transform.localRotation = Quaternion.LookRotation(cameraExtrinsics.ViewFromWorld.GetColumn(2), cameraExtrinsics.ViewFromWorld.GetColumn(1));
+
+            // Magic offset from Unity's underlying coordinate frame (WorldManager.GetNativeISpatialCoordinateSystemPtr()) and the head pose used for the camera.
+            // Poses are sent in the coordinate frame space because the camera position uses prediction.
```
matthejo: Maybe expand the comment to say "Poses are sent in the coordinate frame space because the Unity camera position uses prediction"

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CalculatedCameraExtrinsics.cs
```diff
@@ -22,5 +24,28 @@ public override string ToString()
...
+        {
+            extrinsics = null;
+
+            try
+            {
+                var str = Encoding.ASCII.GetString(payload);
+                extrinsics = JsonUtility.FromJson<CalculatedCameraExtrinsics>(str);
+                return true;
+            }
+            catch
```
wiwei: It might be worth logging the exception (in case the JSON library has some helpful message about what part failed deserialization)

chrisfromwork: i'll add this to both extrinsics/intrinsics

Issue #4174, opened 2019-05-03
NearInteractionTouchable cannot be pressed from the side


julenka:
## Describe the bug

Near interaction touchables can only be configured to be touched within a quad region, from one direction.

## To reproduce

1. Create a box collider with cube on it.
2. Add near interaction touchable, set local forward to -1
3. Try to touch the cube from the side

## Expected behavior

1. As I approach the cube from side, finger cursor shows up correctly angled toward the cube
2. When I touch, I get a touch event

## Actual behavior

When I approach, I do not see finger cursor until I'm touching the object.
Some people have reported not getting any events at all.

## Your Setup (please complete the following information)

- Unity Version 2018.3.9f1
- MRTK Version RC1 Refresh 1

## Target Platform (please complete the following information)

- Editor

## Additional context

I believe one approach to fixing this would be to modify NearInteractionTouchable to support different touchable types. Currently, we just suport a 'quad', we coudl also support 'box volume', and 'plane'. The DistanceToSurface calculation would differ based on the touch type.

PR #4263, opened 2019-05-10
Touch NearInteractionTouchable volumes from the side

MenelvagorMilsom:
Changes NearInteractionTouchable.TouchableSurface to allow for quads or boxes. Quads use the old logic for DistanceToSurface. Boxes use shortest distance to a cuboid in space for DistanceToSurface.

This change may be breaking, as the TouchableSurface enum has changed, meaning that the value that was previously UnityUI is now Box. The UI will still function, but the distance to surface may be strange at the edges. There is a helper in the inspector that will amend this.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
@@ -25,9 +25,11 @@ namespace Microsoft.MixedReality.Toolkit.Input
         
         private float closestDistance = 0.0f;
 
+        private Vector3 closestNormal = Vector3.forward;
+
         // The closest touchable component limits the set of objects which are currently touchable.
         // These are all the game objects in the subtree of the closest touchable component's owner object.
-        private NearInteractionTouchable closestProximityTouchable = null;
+        private BaseNearInteractionTouchable closestProximityTouchable = null;
         // The current object that is being touched. We need to make sure to consistently fire 
         // poke-down / poke-up events for this object. This is also the case when the object within
         // the same current closest touchable component's changes (e.g. Unity UI control elements).
@@ -54,35 +56,31 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 Rays = new RayStep[1];
             }
 
+            closestNormal = Rotation * Vector3.forward;
+
             // Check proximity
-            NearInteractionTouchable newClosestTouchable = null;
+            BaseNearInteractionTouchable newClosestTouchable = null;
             {
                 closestDistance = distFront; // NOTE: Start at distFront for cutoff
-                foreach (var prox in NearInteractionTouchable.Instances)
+                foreach (var prox in BaseNearInteractionTouchable.Instances)
                 {
                     if (prox.ColliderEnabled)
                     {
-                       float dist = prox.DistanceToSurface(Position);
-                       if (dist < closestDistance)
-                       {
-   
-                           closestDistance = dist;
-                           newClosestTouchable = prox;
-                       }
+                        Vector3 normal;
+                        float dist = prox.DistanceToSurface(Position, out normal);
+                        if (dist < closestDistance)
+                        {   
+                            closestDistance = dist;
+                            newClosestTouchable = prox;
+                            closestNormal = normal;
+                        }
                     }
                 }
             }
 
-            // Determine ray direction
-            Vector3 rayDirection = Rotation * Vector3.forward;
-            if (newClosestTouchable != null)
-            {
-                rayDirection = -newClosestTouchable.Forward;
-            }
-
             // Build ray (poke from in front to the back of the pointer position)
-            Vector3 start = Position - distBack * rayDirection;
-            Vector3 end = Position + distFront * rayDirection;
+            Vector3 start = Position - distBack * -closestNormal;
+            Vector3 end = Position + distFront * -closestNormal;
             Rays[0].UpdateRayStep(ref start, ref end);
 
             // Check if the currently touched object is still part of the new touchable.
@@ -212,7 +210,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
             }
         }
 
-        private static bool IsObjectPartOfTouchable(GameObject targetObject, NearInteractionTouchable touchable)
+        private static bool IsObjectPartOfTouchable(GameObject targetObject, BaseNearInteractionTouchable touchable)
         {
             return targetObject != null && touchable != null &&
                 (targetObject == touchable.gameObject ||
@@ -239,7 +237,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// <inheritdoc />
         bool IMixedRealityNearPointer.TryGetNormalToNearestSurface(out Vector3 normal)
         {
-            normal = (closestProximityTouchable != null) ? closestProximityTouchable.Forward : Vector3.forward;
+            normal = closestNormal;
             return true;
         }
 
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
@@ -55,7 +55,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
                     if (prox.ColliderEnabled)
                     {
                         Vector3 normal;
-                        float dist = prox.DistanceToSurface(Position, out normal);
+                        float dist = prox.DistanceToTouchable(Position, out normal);
                         if (dist < closestDistance)
                         {   
                             closestDistance = dist;
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
@@ -109,16 +109,24 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
             if (Result?.CurrentPointerTarget != null)
             {
-                float dist = Vector3.Distance(Result.StartPoint, Result.Details.Point) - distBack;
-                bool newIsDown = (dist < debounceThreshold);
+                float distToFront = Vector3.Distance(Result.StartPoint, Result.Details.Point) - distBack;
+                bool newIsDown = (distToFront < 0);
+                bool newIsUp = (distToFront > debounceThreshold);
 
                 if (newIsDown)
                 {
                     TryRaisePokeDown(Result.CurrentPointerTarget, Position);
                 }
-                else
+                else if (currentTouchableObjectDown != null)
                 {
-                    TryRaisePokeUp(Result.CurrentPointerTarget, Position);
+                    if (newIsUp)
+                    {
+                        TryRaisePokeUp(Result.CurrentPointerTarget, Position);
+                    }
+                    else
+                    {
+                        TryRaisePokeDown(Result.CurrentPointerTarget, Position);
+                    }
                 }
             }
 
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
@@ -9,15 +9,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
     public class PokePointer : BaseControllerPointer, IMixedRealityNearPointer
     {
         [SerializeField]
-        protected float distBack;
-
-        [SerializeField]
-        protected float distFront;
-
-        [SerializeField]
-        protected float debounceThreshold;
-
-        [SerializeField]
         protected LineRenderer line;
 
         [SerializeField]
@@ -37,9 +28,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
         protected void OnValidate()
         {
-            Debug.Assert(distBack > 0, this);
-            Debug.Assert(distFront > 0, this);
-            Debug.Assert(debounceThreshold > 0, this);
             Debug.Assert(line != null, this);
             Debug.Assert(visuals != null, this);
         }
@@ -61,7 +49,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
             // Check proximity
             BaseNearInteractionTouchable newClosestTouchable = null;
             {
-                closestDistance = distFront; // NOTE: Start at distFront for cutoff
+                closestDistance = float.PositiveInfinity;
                 foreach (var prox in BaseNearInteractionTouchable.Instances)
                 {
                     if (prox.ColliderEnabled)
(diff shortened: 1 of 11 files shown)


MenelvagorMilsom:
@julenka, in #4174 you mention a "plane" touchable. What distinguishes the "plane" from the "quad"? Is there any other touchable volumes\surfaces that would be worth supporting?

julenka:
> @julenka, in #4174 you mention a "plane" touchable. What distinguishes the "plane" from the "quad"? Is there any other touchable volumes\surfaces that would be worth supporting?

To me, the difference between a plane and a quad is that the plane has no bounds, whereas a quad has a width and height. You could approximate a plane using a very large quad, of course.

Actually, since a plane is just a very large quad I think simplifying to the following would work:
- **TouchableVolume** - defined by a {Mesh,Box,Capsule,etc.}Collider, entering the volume from anywhere generates a touch down, exiting anywhere generated exit.
- **FrontFaceTouchableSurface** - a touchable surface, touchable just from one direction (the front face). Defined by a normal + bounds, in local space. Touch down generated when you enter from front surface. 
  - Touch up generated when you exit from the front within some tolerance "touchUpToleranceFront". 
  - Touch up also generated if you exit from the back within some tolerance "touchUpToleranceBack"

One more thought:
  - I think the tolerances for surfaces should be configurable per-surface, not globally in the poke pointer like it is now. touch up and touch down should be visualized in an inspector, much like we do for buttons. Something like this:
image

julenka:
> Changes NearInteractionTouchable.TouchableSurface to allow for quads or boxes. 

Regarding shapes to support, in my comment above it would be really cool to support an arbitrary mesh collider if possible. So just "make the collidable that's on the game object I'm on (or provide reference to collidable) and use that to determine if i'm touching the volume".

Also, in addition to boxes, we could pretty easily support spheres without needing to do the full mesh collider.

BUT...looking at the types of objects our customers support (3D models of anatomy and 3D parts), my intuition is they will want arbitrary mesh collidables.

julenka:
While this is a breaking change, I see it as a very positive change since it greatly increases the types of touchables we can have. However, I think we want to make this breaking change only once, so it would be good to finalize the touchable types. I think touchable volume that supports a collider passed to it, FrontTouchableSurface, and then FrontTouchablSurfaceUnbounded would be good.

MenelvagorMilsom:
Thanks for the feedback @julenka. There's one thing I want to clarify. I've started implementing some of your changes under the assumption that **FrontFaceTouchableSurface**, **FrontFaceTouchablePlane** and **TouchableVolume** will all be implemented in much the same way as I implemented **Quad** and **Box** (i.e. using an enum in NearInteractionTouchable to distinguish between behaviours). I was going to use a custom GUI to expose relevant properties for each enum value.

It occurred to me that maybe you meant separating out the behaviours into multiple distinct classes. This may be preferable as the properties and inspector logic required vary quite a lot depending on the enum value you have selected.

I guess that separating things out into multiple scripts will require some further changes in the PokePointer maybe. Which way of doing things did you have in mind in your description above?

julenka:
> Thanks for the feedback @julenka. There's one thing I want to clarify. I've started implementing some of your changes under the assumption that **FrontFaceTouchableSurface**, **FrontFaceTouchablePlane** and **TouchableVolume** will all be implemented in much the same way as I implemented **Quad** and **Box** (i.e. using an enum in NearInteractionTouchable to distinguish between behaviours). I was going to use a custom GUI to expose relevant properties for each enum value.
> 
> It occurred to me that maybe you meant separating out the behaviours into multiple distinct classes. This may be preferable as the properties and inspector logic required vary quite a lot depending on the enum value you have selected.
> 
> I guess that separating things out into multiple scripts will require some further changes in the PokePointer maybe. Which way of doing things did you have in mind in your description above?

Up to you what you think is better, but one advantage of creating different classes for NearInteractionTouchableVolume and NearInteractionTouchableUnboudnedSurface would be that this would remove the need for a breaking change. The disadvantage might be that the classes would be less discoverable (which component do you add), however since they are all prefixed with same "NearInteractionTouchable", it shouldn't be a huge problem.

MenelvagorMilsom:
There appears to be an issue with PokePointer.debounceThreshold, which I imagine is what @julenka describes as _front tolerance (touch up when finger in front of this)_ in her diagram. If this value is increased, it has the effect of moving the touch surface forward, rather than increasing the front tolerance. Seeming as I'm making changes in this area, I'll see if I can fix that as part of this PR.

julenka:
Awesome! Thank you Niall.

MenelvagorMilsom:
Sorry it's taken this long to update this PR, been bouncing between this and other things. I'll see if I can get visualization of the touch planes in.

MenelvagorMilsom:
@julenka myself and Maciej have been focusing on signing recently. I'm not sure I will have time to get these changes in with visualization of the planes in editor this week. Is it worth checking these changes in before Friday and we can get the visualizations in at a later date?

MenelvagorMilsom:
I believe that @thalbern has made changes to the button planes so that you can switch between world and local space. Perhaps we should do something similar here too. At the moment the NearInteractionTouchable distances are all in world space.

thalbern:
> I believe that @thalbern Bernadette Thalhammer FTE has made changes to the button planes so that you can switch between world and local space. Perhaps we should do something similar here too. At the moment the NearInteractionTouchable distances are all in world space.

i've opened a pr for the buttons that has the local / world space switch https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4527

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -54,36 +44,38 @@ public override void OnPreSceneQuery()
...
                     {
-                       float dist = prox.DistanceToSurface(Position);
-                       if (dist < closestDistance)
-                       {
-   
-                           closestDistance = dist;
-                           newClosestTouchable = prox;
-                       }
+                        Vector3 normal;
+                        float dist = prox.DistanceToSurface(Position, out normal);
```
julenka: Now I'm realizing a better name for this would be DistanceToTouchable, since not all touchables are surfaces.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionTouchable.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            distBack = Math.Max(distBack, 0);
+            distFront = Math.Max(distFront, 0);
+            debounceThreshold = Math.Max(debounceThreshold, 0);
+
+            touchableCollider = GetComponent<Collider>();
+            usesCollider = touchableCollider != null;
+        }
+
+        public abstract float DistanceToSurface(Vector3 samplePoint, out Vector3 normal);
```
julenka: Recommend changing this to DistanceToTouchable, since not all of these are surfaces (at least one is a volume).

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnboundedPlane.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    /// <summary>
+    /// Add a NearInteractionTouchableUnboundedPlane to your scene and configure a touchable volume
+    /// in order to get PointerDown and PointerUp events whenever a PokePointer touches the
+    /// attached collider through this surface.
+    /// </summary>
+    public class NearInteractionTouchableUnboundedPlane : BaseNearInteractionTouchable
+    {
+#if UNITY_EDITOR
+        private readonly Color handleColor = Color.white;
```
julenka: You could actually move this to NearInteractionTouchableUnboundedPlaneEditor.cs and move into an Editor folder to remove the #if UNITY_EDITOR. Can do in a later PR.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
```diff
@@ -138,35 +166,14 @@ private enum TouchableSurface
...
-            instances.Remove(this);
-        }
-
-        protected void OnValidate()
-        {
             Debug.Assert(localForward.magnitude > 0);
             Debug.Assert(localUp.magnitude > 0);
             string hierarchy = gameObject.transform.EnumerateAncestors(true).Aggregate("", (result, next) => next.gameObject.name + "=>" + result);
-            Debug.Assert(Vector3.Angle(localForward, localUp) > 80, $"localForward and localUp not perpendicular for object {hierarchy}. Did you set Local Forward correctly?");
+            Debug.Assert(Vector3.Dot(localForward, localUp) == 0, $"localForward and localUp not perpendicular for object {hierarchy}. Did you set Local Forward correctly?");
```
julenka: I'd recommend only checking this if the length of the vector is 1, since i often erroneously get this error message as I'm changing the value from -1 to 1

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableVolume.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    /// <summary>
+    /// Add a NearInteractionTouchableVolume to your scene and configure a touchable volume
+    /// in order to get PointerDown and PointerUp events whenever a PokePointer collides with this volume.
+    /// </summary>
+    public class NearInteractionTouchableVolume : BaseNearInteractionTouchable
+    {
+        public override float DistanceToSurface(Vector3 samplePoint, out Vector3 normal)
+        {
+            if (usesCollider)
```
julenka: what about if the volume does not use the collider? It would be good in the tooltip or class description to say that this class requires that a collider exists on the component and that the collider is used to compute the touchable volume.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableVolume.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+using UnityEngine.Serialization;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
```
julenka: @wiwei I wish we had a way to mark this component as "BETA" or "EXPERIMENTAL". It will be the first time we are showing it in a release and I could see the component becoming subject to change as people put it through its paces.

julenka:
> I believe that @thalbern has made changes to the button planes so that you can switch between world and local space. Perhaps we should do something similar here too. At the moment the NearInteractionTouchable distances are all in world space.

How would that help for near interaction touchables? Could you provide an example?

julenka:
> @julenka myself and Maciej have been focusing on signing recently. I'm not sure I will have time to get these changes in with visualization of the planes in editor this week. Is it worth checking these changes in before Friday and we can get the visualizations in at a later date?

Thank you for the great work to get this in, thus far! I would prioritize signing and just verifying that these changes work (via say an example scene) over the visualization. It's fine to do the visualization later, as long as we have a work item to track it.

MenelvagorMilsom:
Thanks for the review @julenka. I just added some small changes. I should have more time to give this tomorrow.

MenelvagorMilsom:
image
Sample scene has objects in the foreground, with different combinations of touchables and colliders. They rotate when you touch them so you can observe how they all behave from different angles.
There's also an unbounded plane below the rest of the scene content that changes color on touch.

MenelvagorMilsom:
I just created an issue to track the plane visualizations:
#4632

julenka:
Nice! By the way it looks like touchableSurfaceUnbounded is not unbounded -- i get a touch up when i exit the bounds of the plane
unboundplane

julenka:
Playing with this myself I was a bit confused...what do Dist Back and Dist Front really mean for the Near Interaction Touchable Volume? Maybe we should hide those fields.

image

julenka:
Oh wow I LOVE the way the finger cursor matches the normal of the sphere. Awesome!

touchnormal

MenelvagorMilsom:
> Nice! By the way it looks like touchableSurfaceUnbounded is not unbounded -- i get a touch up when i exit the bounds of the plane
Yeah, this is still depending on the collider. If we want a truly unbounded touchable plane I can create another issue to track this.

MenelvagorMilsom:
> Playing with this myself I was a bit confused...what do Dist Back and Dist Front really mean for the Near Interaction Touchable Volume? Maybe we should hide those fields.

Yeah, I'm actually a bit confused what DistFront is really doing. I'll hide it alright. As for DistBack, if the fingertip moves further than DistBack from the surface (be it a plane or the surface of a volume), you'll get a touch up event. I'll make sure that DistBack and DebounceThreshold have descriptive tooltips for the time being. I imagine that this will become clearer to the user when we have visualizations of these planes.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -54,36 +44,38 @@ public override void OnPreSceneQuery()
...
 
+            closestNormal = Rotation * Vector3.forward;
+
             // Check proximity
-            NearInteractionTouchable newClosestTouchable = null;
+            BaseNearInteractionTouchable newClosestTouchable = null;
             {
-                closestDistance = distFront; // NOTE: Start at distFront for cutoff
-                foreach (var prox in NearInteractionTouchable.Instances)
+                closestDistance = float.PositiveInfinity;
```
keveleigh: With this change, it looks like there will always be a `BaseNearInteractionTouchable` within "touching" distance. Previously, using `distFront` as the cutoff, we'd only determine the closest touchable object if the pointer was within a specific distance from it (which means there might be no closest touchable in many cases). Were we missing cases with that method?

keveleigh: An example of a usage of `distFront` that we should make sure is accounted for, if we remove it:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4640#issuecomment-497514917

julenka: @keveleigh actually the comment in that issue was not that they are using distFront explicitly -- it was more like they had undone their original workaround to go back to the default value. I'm actually not clear how their original workaround got them to fix the issue they were having, to be honest.

keveleigh: Ohh interesting. Thanks for the clarification!

julenka:
Documentation says the following about DistFront:

distFront: How far input point needs to be in front of surface before press event occurs.

For volumes, we could adapt this to be:

distFront: Distance from volume before press event occurs.

I would be okay to check this in as is and add follow up item to resolve this.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
julenka: However, I think you do have a point Kurtis that I'd be curious to hear Niall's response to. I think this can go in and we can solve it in a separate issue, just in the interest of moving things forward. I'm sure once this is in we will discover other bugs.

MenelvagorMilsom: Good point, I think I actually intended on trying something here that I didn't get around to. I've created an issue and assigned it to myself #4699.

Issue #3876, opened 2019-04-06
Inconsistencies between profile window messages


SimonDarksideJ:
## Overview
New Profiles, like the Input Simulation profile have additional "helpful" messages on them.  However, this is now inconsistent with the rest of the profiles in the MRTK.

Should pick a standard and update all screens to match.

Pointer Profile
![image](https://user-images.githubusercontent.com/1793042/55669585-1518b400-5871-11e9-9e71-1abc8339751e.png)

Input Simulation Profile
![image](https://user-images.githubusercontent.com/1793042/55669594-26fa5700-5871-11e9-8ba4-4c9b8db8c9f5.png)

## Expected Behavior
Screen behaviour / look and feel should be the same across all profiles.

## Actual Behavior
Inconsistent behaviour

## Steps to reproduce
Look at all profiles

## Unity Editor Version
Unity 2018.3.11

## Mixed Reality Toolkit Release Version
MRTK-RC1

davidkline-ms:
Will investigate and address while making related profile inspector changes. thanks!

chrisfromwork:
@SimonDarksideJ and @Railboy our expectation for fixing this is that we should add these messages to all profiles. Would this suggested changes suffice in your twos minds?

davidkline-ms:
The images as shown above are encountered only when the MRTK object is not in the scene. The next part of the #3545 work modifies the behavior and eliminates this specific issue.

Thanks!

Yoyozilla:
@davidkline-ms   Is this addressed in the latest profile updates and service facades?

Railboy:
@Yoyozilla no we didn't tackle this issue in the service facade PR.

I'm still waiting on an official style guide a la #3908 to come together (of which this issue would be a part) before I try to homogenize everything. So this might be a chicken-and-egg problem.

PR #4264, opened 2019-05-10
Config Profile Homogenization

Railboy:
## Overview
 Makes the look of headers in profiles more consistent and adds some functions to encourage more consistency in the future.

ProfileConsistency

## Changes
- Adds a `RenderTitleDescriptionAndLogo` function to MixedRealityInspectorUtility. This handles the business of drawing a title and helpbox with consistent spacing etc.
- When an instance of the MRTK isn't found, profile inspectors now provide a button that will add one to your scene.
- Most profiles will now render their contents even if a MixedRealityToolkit instance isn't found. The only exceptions are profiles that require calls to MixedRealityToolkit.Instance to function. In a future PR it would be nice if these could be made to at least render partial contents. Will open an issue for this later.

- Fixes: #3876

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/Utilities/MixedRealityInspectorUtility.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Utilities/MixedRealityInspectorUtility.cs
@@ -22,19 +22,19 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
         /// Check and make sure we have a Mixed Reality Toolkit and an active profile.
         /// </summary>
         /// <returns>True if the Mixed Reality Toolkit is properly initialized.</returns>
-        public static bool CheckMixedRealityConfigured(bool showHelpBox = true)
+        public static bool CheckMixedRealityConfigured(bool renderEditorElements = true, bool showCreateButton = false)
         {
             if (!MixedRealityToolkit.IsInitialized)
             {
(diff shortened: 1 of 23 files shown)


Assets/MixedRealityToolkit/Inspectors/Utilities/MixedRealityInspectorUtility.cs
```diff
@@ -22,19 +22,19 @@ public static class MixedRealityInspectorUtility
         /// Check and make sure we have a Mixed Reality Toolkit and an active profile.
         /// </summary>
         /// <returns>True if the Mixed Reality Toolkit is properly initialized.</returns>
-        public static bool CheckMixedRealityConfigured(bool showHelpBox = true)
+        public static bool CheckMixedRealityConfigured(bool renderEditorElements = true, bool showCreateButton = false)
```
wiwei: Do you know if we plan on calling this with renderEditorElements = false ever? I know that prior to this change it looks like it was used to check for the MRTK object without rendering the help box in certain cases - with this change it looks like all callers I can see within the context of this PR had their false-passing instances removed.

Railboy: It often gets called by inspectors in OnEnable when gui stuff isn't getting rendered.

Issue #4176, opened 2019-05-03
Head cursor on by default in scene


erhavir:
## Overview
We are seeing the head cursor being on by default. It turns off when hands enter the scene, and then turns back on when hands go away. This isn't the behavior in the shell, where head cursor appears only before first hand interaction and disappears and then only appears again when voice has been used and then goes away upon hand interaction and stays away.

## Expected Behavior
If the user says "Select" or other voice commands, the head cursor is activated until their next hand interaction, at which time the head cursor goes away.

## Actual Behavior
Head cursor is always on.

## Steps to reproduce
_(Links to sample github project preferred)_

## Mixed Reality Toolkit Release Version
MRTK RC1

Yoyozilla:
@julenka this seems like a shell parity issue, could you confirm?

julenka:
Yeah, by default shell does not show gaze cursor.

julenka:
Started on this yesterday.

wiwei:
Closing this out now that the fix was merged

Issue #4223, opened 2019-05-07
Pointers should not be destroyed when source is lost


provencher:
## Describe the problem

Currently, pointers are destroyed and re-instantiated when their source is lost.
This has a number of side effects:

- It is not possible to cache them, since new ones take their place and the old ones go stale quickly.
- If attempting to cache a pointer, to detect that it has gone stale, it is imperative to first cast it to a monobehavior to then null check it using Unity's overridden null check. 

## Describe the solution you'd like

Pointers should be set inactive, perhaps with a tracking state change, but should not be destroyed. This will alleviate performance considerations as recreating pointers is no longer necessary, and when they return into view, the cached instances will not be stale.

SimonDarksideJ:
Agreed. I the source is "Lost" then everything should be cleaned up as you cannot guarantee the state of the controller if it is discovered again.

If however, it needs to be retained, better to look at ways to hold the controller and not fire the source lost in those instances.

Yoyozilla:
This is a performance issue where pointer is destroyed everytime when hand is not in view, and recreated when it comes back. 

@provencher  could we get performance profiling data on perf gain when not destroying pointers?

provencher:
@Yoyozilla I don't think I should be assigned to solve this.
I'm personally not very familiar with how pointers are managed at a data provider level, especially in regards to WMR platforms which our apps don't support.

I haven't benchmarked how this performs, but I know from much experience with unity that instantiating a prefab can take a non-trivial amount of CPU time and block the Unity thread each time it's done.
By not destroying these pointers, we save that compute time. 

Right now, to get around losing access to the pointers and not being able to cache them, I have to iterate over the list of all pointers and do a series of checks, including string comparisons, which generates garbage, each frame that neither hand or controller has an active source.


@SimonDarksideJ I don't think you should clean up the pointer the moment the source is lost.

There can never be more than 2 hands, or 2 controllers at any given moment. With that said, while you cannot guarantee the state of the controller when it returns, you can easily query it before setting it active and exposing that state to the API user.

In regards to controllers, even if their source is lost, as in they're outside the field of view, they're still paired and provide acceleration and rotation information that can be useful.

As for hands, they come drop in and out constantly, and doing all these checks around finding active pointers in view can get expensive, without mentioning the impact of disposing and re-instantiating the prefab what could be several times per second depending on the user's hand movement.

StephenHodgson:
> As for hands, they come drop in and out constantly

[...]

PR #4270, opened 2019-05-10
Gaze cursor activation behavior that matches the shell

julenka:
## Overview
Implement gaze cursor activation behavior that matches the HoloLens 2 and Windows Mixed Reality (WMR) shell:

- Head cursor appears when app start
- Cursor disappears when articulated hands or motion controllers show up (but NOT GGV hands as for hololens 1, or for mouse / gamepad).
- Gaze cursor only appears again when voice wake word has been used (for WMR this is the word 'select').

## Changes
- Fixes: #4176
- Add speech command "select" to the speech commands profile, activatable by Alpha1
- Remove "toggle diagnostics" and "toggle profiler" speech commands because they are never used.
- **Small Breaking Change** IFocusProvider implements IMixedRealitySpeechHandler. I don't expect that anybody has built a custom focusprovider yet, and if they have the compile error provides obvious solution (implement the interface). I couldn't think of any other way to listen for a speech command that wouldn't result in adding extra game objects that I could listen to events for. 

## Verification
Tested in editor the following cases:
- Articulated hands
- GGV Hands
- Gamepad
- Mouse

Need to still test:
- HL1 device
- HL2 Device
- Using WMR controllers (after voice mode, clicking controller should move back to controller input)

>
>


Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -412,6 +412,92 @@ public override int GetHashCode()
             }
         }
 
+        /// <summary>
+        /// Helper class for managing the visibility of the gaze pointer to match windows mixed reality and HoloLens 2
+        /// When application starts, gaze pointer is visible. Then when articulate hands / motion controllers
+        /// appear, hide the gaze cursor. Whenever user says the voice wake word, make the 
+        /// gaze controller appear.
+        /// </summary>
+        private class GazePointerStateMachine
```
wiwei: Code looks good to me, only comment here is that this code looks super testable (i.e. it takes deterministic inputs, and then its state changes) - this would be a great candidate for a unit test.

wiwei:
Something I think that I also don't fully grok is that this is a breaking change (an interface change), and I'm curious to understand the dialogue around the tradeoffs/benefits/risks to doing this.

i.e. is parity in this aspect worth the pain of breaking changes for this interface, or do we believe that this particular thing has a 0% chance of being extended, so the net pain is basically zero and thus the breaking change is okay?

We've been reducing the amount of breaking changes as we stabilize, and generally: 1) we shouldn't do it and 2) if there is significant benefit and we can prove insigifinicant harm or provide a migration script, then maybe it's okay?

julenka:
> Something I think that I also don't fully grok is that this is a breaking change (an interface change), and I'm curious to understand the dialogue around the tradeoffs/benefits/risks to doing this.
> 
> i.e. is parity in this aspect worth the pain of breaking changes for this interface, or do we believe that this particular thing has a 0% chance of being extended, so the net pain is basically zero and thus the breaking change is okay?
> 
> We've been reducing the amount of breaking changes as we stabilize, and generally: 1) we shouldn't do it and 2) if there is significant benefit and we can prove insigifinicant harm or provide a migration script, then maybe it's okay?

I've updated the change to have it be a smaller breaking change: just have IFocusProvider implement the MRSpeechHandler interface in addition to sourcestatehandler. The chance that IFocusProvider has been extended is small (<10%), however even if somebody extended it, the way to fix the break is obvious from the compile error: implement the missing interface. The migration script in this case would just be manual and should not take long "if you implemented your own FocusProvider, add an implementation of the new method".

I actually couldn't think of a way to not require FocusHandler to listen to a voice command without modifying the interface or having to add a fairly big hack (add game object that implements interface, expose unity event, and listen for that event. Or cast IMRFocusProvider into concrete type and call a public method directly...which seems against the approach of trying to allow others to extend the system). AFAIK there is no way to say "InputSystem.SpeechRecognized += OnSpeechRecognized". You need to implement an interface to get events.

The change fixes a user visible defect "cursor is always visible when hands are down", and helps us get closer to shell parity. I think it is worth the risk of having some small percentage of people add another method to their FocusProvider in this case.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -1294,6 +1294,8 @@ public void RaiseSpeechCommandRecognized(IMixedRealityInputSource source, Recogn
             // Create input event
             speechEventData.Initialize(source, confidence, phraseDuration, phraseStartTime, command);
 
+            FocusProvider?.OnSpeechKeywordRecognized(speechEventData);
```
julenka: This approach matches the approach used for OnSourceDetected, above.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -412,6 +412,97 @@ public override int GetHashCode()
...
+                get { return gazePointerState != GazePointerState.GazePointerInactive; }
+            }
+
+            public void UpdateState(HashSet<PointerData> pointers, GenericPointer currentGazePointer)
+            {
+                int numFarPointers = 0;
+                int numNearPointersActive = 0;
+                foreach (var p in pointers)
+                {
+                    if (p.Pointer != null)
```
provencher: It's not enough to do a null check on a pointer.
This is because the pointer is stored as an interface, but many of them are actually monobehaviors.

When a pointer is destroyed, it will pass a regular null check, but throw a MissingReferenceException. You must also cast it to monobehavior to null check it properly. This unfortunately is much more expensive, but safer.

This is why I'm so adamant pointers should never be destroyed, so that we don't ever hold a list of null pointers.
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4223

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -412,6 +412,97 @@ public override int GetHashCode()
...
+                    {
+                        if (p.Pointer is IMixedRealityNearPointer)
+                        {
+                            if (p.Pointer.IsInteractionEnabled)
```
provencher: You could reduce nesting by doing and && here with the condition above.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -412,6 +412,97 @@ public override int GetHashCode()
...
+                            {
+                                numNearPointersActive++;
```
provencher: If you're counting the active near pointers, you could expose this at a higher level. This seems like information that might be important.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -412,6 +412,97 @@ public override int GetHashCode()
...
+                            }
+                        }
+                        else if (p.Pointer.BaseCursor != null 
+                            && !(p.Pointer == currentGazePointer)
+                            && p.Pointer.IsInteractionEnabled)
+                        {
+                            // We ignore the currentGazePointer here because for cases like HoloLens 1
+                            // hand input or the gamepad, we want to show the cursor still.
+                            numFarPointers++;
```
provencher: Same as above. I can see a few cases where knowing the number of far pointers can be very useful.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -412,6 +412,97 @@ public override int GetHashCode()
...
+                    if (p.Pointer != null)
+                    {
+                        if (p.Pointer is IMixedRealityNearPointer)
+                        {
+                            if (p.Pointer.IsInteractionEnabled)
+                            {
+                                numNearPointersActive++;
+                            }
+                        }
+                        else if (p.Pointer.BaseCursor != null
```
provencher: Same as above. Be careful doing null checks with monobehaviors that are stored as interfaces.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -412,6 +412,97 @@ public override int GetHashCode()
...
+            }
+            private GazePointerState gazePointerState = GazePointerState.Initial;
+            private bool activateGazeKeywordIsSet = false;
+
+            public bool IsGazePointerActive
+            {
+                get { return gazePointerState != GazePointerState.GazePointerInactive; }
+            }
+
+            public void UpdateState(HashSet<PointerData> pointers, GenericPointer currentGazePointer)
```
provencher: This whole loop has a ton of complexity just to gauge the state of a cursor. I understand why it's happening, but I think it should be heavily profiled before being merged in.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
provencher: It's not safe to use `?` on monobehaviors. They do not override the null check properly.

julenka: Oh interesting, I didn't know that. I simply wrote it this way to stay consistent with OnSourceDetected behavior. We may want to fix this in a separate issue. I'm curious, could you please share a reference explaining why we cannot use ? operator with monobehaviors?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: Hm, interesting. Since this does not regress prior behavior (we did the same null check previously), I think we could keep this in for now, and then fix it and other null checks later, as part of a separate issue. Trying to keep this scoped to implementing the specific behavior to match the shell.

provencher: I understand wanting to keep the same behavior.

My experience with pointers however has shown that if the pointer is destroyed, it won't be considered null until unity's native side cleans it up.

This is why you want to know if the reference is still valid, which can only be done with the overridden monobebehavior null check, which isn't called unless you cast the pointer reference to monobebehavior.

provencher: This is the same reason the `?` fails on monobebehaviors. This check is actually somewhat new since it wasn't possible before dot net 4.x. It works the same way as the ternary operator or the ?? Null check however.

You can read about how unity handles null checks here

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: I could, but then the meaning of the else clause below would mean "it is not near and enabled", meaning that if the pointer was NearPointer and disabled, that it would fall into the else clause. Even though it might be okay, I think I'd prefer to keep it this way for now, just so it's more clear what what else clause captures.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: Okay, I'll expose the number of near and far active pointers as public property in FocusProvider. I won't make it part of the interface to avoid making a breaking change.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: Just added some playmode tests to validate the state of cursor giving hand states.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: Removed this null check, because the pointer should never be null at this point in time. Other code in update loop already assumes the pointer is not null. It's appropriate in this case to crash/throw exception because this reflects program being in incorrect state.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -104,7 +104,7 @@ public virtual void OnSourceLost(SourceStateEventData eventData)
 
                 if (destroyOnSourceLost)
                 {
-                    if (Application.isEditor)
+                    if (!Application.isPlaying)
```
luis-valverde-ms: >if (!Application.isPlaying) [](start = 20, length = 27)

No biggie and thanks for spotting it but this seems to fix a different bug. If that is the case I would have rather seeing this change in a separate PR and linking to its own bug.

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerStateMachine.cs
```diff
@@ -0,0 +1,74 @@
...
+                    break;
+            }
+            gazePointerState = newState;
+        }
+
+        public void OnSpeechKeywordRecognized(SpeechEventData eventData)
+        {
+            if (eventData.Command.Keyword.Equals("select", StringComparison.CurrentCultureIgnoreCase))
+            {
+                activateGazeKeywordIsSet = true;
```
luis-valverde-ms: >activateGazeKeywordIsSet = true [](start = 16, length = 31)

What happens if we say select while the gaze pointer is active, does this not remain set and re-activates the pointer immediately after the next deactivation? We may want to set this only if the current state is inactive.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
luis-valverde-ms: >FocusProvider?.OnSpeechKeywordRecognized(speechEventData); [](start = 12, length = 58)

It would be cleaner if our FocusProvider implementation received speech events registering as a global listener instead.

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerStateMachine.cs
```diff
@@ -0,0 +1,74 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{ 
+    /// <summary>
+    /// Helper class for managing the visibility of the gaze pointer to match windows mixed reality and HoloLens 2
+    /// When application starts, gaze pointer is visible. Then when articulate hands / motion controllers
+    /// appear, hide the gaze cursor. Whenever user says the voice wake word, make the
```
wiwei: >wake word [](start = 67, length = 9)

Nit: specify what the wake word is here (I know it's mentioned deeper in the code, as a reader of docs I would read this first and since this is hardcoded to "select" and not configurable in any way, it's probably better to have it mentioned as early as possible.

Assets/MixedRealityToolkit.Tests/PlayModeTests/FocusProviderTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Tests
+{
+    public class FocusProviderTests
+    {
+        /// <summary>
+        /// </summary>
+        /// <returns></returns>
+        [UnityTest]
+        public IEnumerator Test01_GazeCursorArticulated()
```
wiwei: >Test01_ [](start = 27, length = 7)

Nit: Delete the 'Test01' prefix

Assets/MixedRealityToolkit.Tests/PlayModeTests/FocusProviderTests.cs
```diff
@@ -0,0 +1,62 @@
...
+            yield return null;
+            // Gaze cursor should not be visible
+            Assert.IsFalse(MixedRealityToolkit.InputSystem.GazeProvider.GazePointer.IsInteractionEnabled, "Gaze cursor should not be visible when one articulated hand is up");
+            inputSimulationService.HandDataRight.Update(false, false, handOpenPose);
+            yield return null;
+
+            // Say "select" to make gaze cursor active again
+            // Really we need to tear down the scene and create it again but MRTK doesn't support that yet
+            var gazeInputSource = MixedRealityToolkit.InputSystem.DetectedInputSources.Where(x => x.SourceName.Equals("Gaze")).First();
+            MixedRealityToolkit.InputSystem.RaiseSpeechCommandRecognized(gazeInputSource, RecognitionConfidenceLevel.High, new System.TimeSpan(), System.DateTime.Now, new SpeechCommands("select", KeyCode.Alpha1, MixedRealityInputAction.None));
```
wiwei: >MixedRealityToolkit.InputSystem [](start = 12, length = 31)

You and dave should chat and make sure that one of you is on point to update this test in response to his PR:

[https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4315](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4315)

Assets/MixedRealityToolkit.Tests/PlayModeTests/FocusProviderTests.cs
wiwei: Mentioned in another PR, but tests should be named based on what they're testing, and arbitrary numbered prefixes don't add value.

That being said "TestGazeCursorArcticulated" would also be fine (because it's testing the gaze cursor articulated interaction).

---
In reply to: [283890681](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4270#discussion_r283890681) [](ancestors = 283890681)

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.MixedReality.Toolkit.Input;
+using Microsoft.MixedReality.Toolkit.Utilities;
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
```
wiwei: Nit: Place at top of file, before usings

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerStateMachine.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{ 
+    /// <summary>
+    /// Helper class for managing the visibility of the gaze pointer to match windows mixed reality and HoloLens 2
+    /// When application starts, gaze pointer is visible. Then when articulate hands / motion controllers
+    /// appear, hide the gaze cursor. Whenever user says the voice wake word, make the 
+    /// gaze controller appear.
+    /// </summary>
+    class GazePointerStateMachine : IMixedRealitySpeechHandler
```
Railboy: Nit: The name threw me on this one because it sounds like a generic pointer utility. Maybe GazePointerSpeechStateMachine?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -32,6 +32,18 @@ public class FocusProvider : BaseDataProvider, IMixedRealityFocusProvider
...
+        /// <summary>
+        /// Number of IMixedRealityNearPointers that are active (IsInteractionEnabled == true).
+        /// </summary>
+        public int NumNearPointersActive { get; private set; }
+
+        /// <summary>
+        /// The number of pointers that support far interaction (like motion controller rays, hand rays) that 
+        /// are active (IsInteractionEnabled == true), excluding the gaze cursor
+        /// </summary>
+        public int NumFarPointersActive { get; private set; }
```
Railboy: Nit: I understand the motivation here but I'll just say it hurts a bit to see the non-universal concepts of near and far pointers hard-coded into the focus provider. I had hoped the pointer mediator could handle the business of near and far pointers somehow. Ah well.

wiwei:
I'm glad that @provencher mentioned the null-comparison-against-interfaces-that-are-monobehaviors-under-the-covers issue.

@provencher do you happen to know of any analyzers that will detect this type of stuff? If not it might be worth it for me to write something like this because having this be called out automatically (instead of having to have someone having to have a super careful eye on all of these reviews calling out those sorts of things)

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
julenka: Actually, without this fix the playmode tests will fail, caused by a null reference exception that happens when removing the controller for a simulated hand.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
julenka: Will also suggested this, I didn't do this because I wanted to follow the existing pattern. But it looks like the existing pattern is incorrectly checking null anyway I'll make the change

julenka: Okay, I see why it was not created as a global listener. The FocusProvider is actually not a gameObject, so cannot be registered as a global listener using `MixedRealityToolkit.InputSystem.Register(gameObject);`. Will keep it as is since I don't know of a solution to register the MRTK input system with a class that is not a game object.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
julenka: Oh wow, sorry.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: @provencher asked to expose the number of near and far pointers. 

I had also thought about putting the near and far pointer management in the mediator, but wanted to avoid the number of classes that knew about / managed the gaze pointer.

If we put the near / far pointer counts in the mediator, how would a dev access it? MRTK.InputSystem.PointerMediator.NumFarPointers? Then we would need to add PointerMediator to the IMixedRealityInputSystem interface, or we could do same trick I do here and cast it to MixedRealitYinputSystem.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
keveleigh: >It's not safe to use ? on monobehaviors.

While true, FocusProvider isn't a MonoBehaviour.

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerStateMachine.cs
julenka: Added test case for this also.

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerStateMachine.cs
julenka: Good catch, I forgot to update the docs.

Assets/MixedRealityToolkit.Tests/PlayModeTests/FocusProviderTests.cs
julenka: Fixed, except the registry gave me null pointers when I tried to get the InputSimulationService, so I kept that the old way. @davidkline-ms any idea why TryGetService<InputSimulationService> would give me null?

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerStateMachine.cs
julenka: Renamed to GazePointerVisibiiltyStateMachine

PR #4291, opened 2019-05-10
Improving alignment of rotation handles on bounding box

djohnsomsft:
The current way that rotation handles are aligned calculates an axis-to-axis alignment from the Y axis to the axis the edge runs along. For models that are symmetric across the Y axis, this looks fine, but for asymmetric gizmos (like the example corrected in the image below), the X and Z axis aren't oriented consistently relative to the box.

This change orients the Y axis in the same way, and the X and Z axis pointing away from the bounding box. The image shows the result of this change when used with a sample arrowed arc model common in rotation widgets, with the arc going from +X to +Z in the prefab's space.

gizmofix


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -749,6 +737,22 @@ private void AddLinks()
                 }
             }
 
+            // Aligns each rotation handle with the Y axis along the edge and the X and Z axis pointing
```
julenka: It would be useful to describe alignment of the prefabs in a tooltip for the rotation prefab, since some people will not have easy access to the raw code and may not see this comment.

djohnsomsft: Done.

Issue #2463, opened 2018-07-19
Update Standard shader to be LRP compatible


SimonDarksideJ:
## Overview
More of an experiment as I haven't even tried the standard shader in the Lightweight Render Pipeline.  But it would be good to ensure the Standard shader is either supported in the LRP, or make a LRP Standard shader

## Expected Behavior
Standard Shader works in LRP

Cameron-Micka:
Sounds like a good idea to test this! I can role this into my next PR if anything needs to change.

isurfraz:
@SimonDarksideJ @Cameron-Micka I am actually doing experiments right now with LWRP and trying to make a slimline LWRP compatible hololens shader. I will update you with progress. BUt would be very nice if on your end you could get the holotoolkit working with the LWRP standard shader. From looking at its code it has some optimisations for XR, and the rendering pipeline also seems to be in general optimised for AR and VR as well as mobile hardware (such as the hololens has).

SimonDarksideJ:
Out LWRP efforts will likely be solely focus on the vNext project, as it is more streamlined and memory efficient already @isurfraz 
We welcome any contribution towards the goal of making MRTK even better :D

isurfraz:
@SimonDarksideJ in that case if I manage to get any decent headway on this in the meantime I will update you both regarding this :)

On a side note, does the vNext project offer any optimisations itself as we are using the release map for this project and hoping to incorporate it when it drops!

isurfraz:
@SimonDarksideJ @Cameron-Micka @StephenHodgson  does a pull of vNext cause any errors or breaking changes? or is it possible to test this in projects now?

SimonDarksideJ:
Best to have this conversation on the Slack channel, in the holodevelopers chat.
vNext hasn't yet reached the Alpha stage, it's close but not there yet.

Cameron-Micka:
"Bumping" this since is something that should be tackled soon, especially once we move to Unity 2019.1 or greater since it takes the LWRP out of preview.

Issue #3953, opened 2019-04-12
Mr. Teekay's Standard.shader shows up as pink (without issues) in Unity 2019.2.0a11 (LWRP)


quakeulf:
## Overview
The newest **LightweightRenderPipelineAsset** is incompatible with the shader. **Removing** the _reference_ to **it** in **_Project Settings > Graphics > Scriptable Render Pipeline Settings_** makes it return to _normal._
## Expected Behavior
A nice feel of **_hologrammatic_** shade.
## Actual Behavior
Plain **_pink._**
## Steps to reproduce
1. Import **MRTK v2 RC1** into the specified version of _Unity._
2. Make _sure_ that the **Scriptable Render Pipeline Settings** in **_Project Settings > Graphics > Scriptable Render Pipeline Settings_** is _set_ to **LWRP_Asset(LightweightRenderPipelineAsset)**
3. Open a **scene** and use the **_Standard.shader_** on a _material on an object_ on the **scene.**
4. It's hereby _pink._

## Unity Editor Version
**Unity** _2019.2.0a11_
## Mixed Reality Toolkit Release Version
**MRTK** _v2_ RC**1**

provencher:
@quakeulf This is not a bug.

The MRTK standard shader is built against a different rendering pipeline. Any shader in the MRTK needs to be completely rebuilt for the LWRP. If you'd like to volunteer such a contribution, I'm sure the community would appreciate it. Until then, if you want the LWRP, you'll need to write new shaders, or use the default unity ones and hope they run well enough.

Cameron-Micka:
Hi @quakeulf, the LWRP is something I hope to support. But, I have yet to evaluate what work needs to be done. I'll be sure to post back here as I know more. :)

Cameron-Micka:
A pull request is up to address this: https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4311

PR #4311, opened 2019-05-13
MRTK/Standard Shader Updates + Lightweight Scriptable Render Pipeline Support

Cameron-Micka:
## Overview
This pull request adds a few improvements to the MRTK/Standard shader (listed below) and also introduces support for Unity's Lightweight Scriptable Render Pipeline.

## Changes

- Added a upgrade path to allow developers to utilize Unity's Lightweight Scriptable Render Pipeline (LWRP) with MRTK shaders. Tested in Unity 2019.1.1f1 and the "com.unity.render-pipelines.lightweight": "5.7.2" package on Windows PC, HoloLens, and HoloLens 2. For instructions on getting started with the LWRP please see [this page.](https://docs.unity3d.com/Packages/com.unity.render-pipelines.lightweight@6.5/manual/installing-lwrp-into-an-existing-project.html) To perform the MRTK upgade select: Mixed Reality Toolkit -> Utilities -> Upgrade MRTK Standard Shader for Lightweight Render Pipeline
Upgrade

- Adding custom depth offset support to avoid z-fighting and allow for hand tuned depth LSR values.
- Added minimum near fade clamp to control the minimum allowable fade.
- New feature that supports vertex extrusion in world space. Useful for visualizing extruded bounding volumes or transitions in/out meshes. (With new example in the MaterialGallery scene.)
VeretxExtrusion

- Optimization, moved spherical harmonics calculations for ambient lighting to vertex shader. Can be used to mimic vertex lighting given a suitable skybox.
- Bug fix, enabled instancing to disable batching for local space triplanar mapping which requires object scale.
- Improvements to the channel packer tool (now called the Texture Combiner) based on feedback from  Chidi Mbadugha. Thank you Chidi! The tool can now export to multiple texture formats.

- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/2463 https://github.com/microsoft/MixedRealityToolkit-Unity/issues/3953

## Verification
I would like reviewers who currently use the LRWP in their Unity project to verify these changes work based on their render pipeline configuration, and for developers who are new to the Unity's scriptable renderer pipeline to report back any issues they encounter.


Cameron-Micka:
I'm going to push in README updates soon to document the LWRP upgrade, document the MRTK/Standard shader and Texture Combiner. Just wanted to get some eyes ASAP on these changes.

Assets/MixedRealityToolkit/Inspectors/MixedRealityStandardShaderGUI.cs
```diff
@@ -1018,5 +1063,36 @@ protected static void SetColorProperty(Material material, string propertyName, C
                 material.SetColor(propertyName, propertyValue.Value);
             }
         }
+
+        [MenuItem("Mixed Reality Toolkit/Utilities/Upgrade MRTK Standard Shader for Lightweight Render Pipeline")]
+        protected static void UpgradeShaderForLightweightRenderPipeline()
+        {
```
Railboy: Odds are people will be clicking this just to see what happens.

If they do this in a project that isn't using the lightweight render pipeline, their shaders will stop working and render as pink, correct?

We should confirm their choice with a yes/cancel dialog, and we should add a menu operation that reverses the procedure.

Cameron-Micka: That's a good idea! I'll add that.

Jarodshow: You could also confirm they have the LWRP dependency before upgrading.

Cameron-Micka: @Jarodshow Great idea! I just added a menu item validator to make sure they at minimum have a renderPipelineAsset specified in their project settings. (Just validating any SRP asset in case a developer has a custom RP based on the LWRP.)

julenka:
>  For instructions on getting started with the LWRP please see this page. To perform the MRTK upgade select: Mixed Reality Toolkit -> Utilities -> Upgrade MRTK Standard Shader for Lightweight Render Pipeline

Would be good to share this info more broadly than just in the PR. For example, might be something we want to share in release notes.

Assets/MixedRealityToolkit/Inspectors/MixedRealityStandardShaderGUI.cs
```diff
@@ -1018,5 +1063,49 @@ protected static void SetColorProperty(Material material, string propertyName, C
                 material.SetColor(propertyName, propertyValue.Value);
             }
         }
+
+        [MenuItem("Mixed Reality Toolkit/Utilities/Upgrade MRTK Standard Shader for Lightweight Render Pipeline")]
+        protected static void UpgradeShaderForLightweightRenderPipeline()
```
julenka: I'd like to make sure we have a plan to communicate this out to our devs that they need to do this. Cameron, since you made the PR, could you please be on point to communicate out that people may need to click this upgrade button? You could work with @wiwei  since he is managing the release

Assets/MixedRealityToolkit/Inspectors/TextureCombinerWindow.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.MixedReality.Toolkit.Editor
 {
-    public class ChannelPackerWindow : EditorWindow
+    public class TextureCombinerWindow : EditorWindow
```
davidkline-ms: Does this class name change have anticipated customer impact?

Assets/MixedRealityToolkit/StandardAssets/Shaders/MixedRealityStandard.shader
```diff
@@ -319,8 +344,11 @@ Shader "Mixed Reality Toolkit/Standard"
 #if defined(_VERTEX_COLORS)
                 fixed4 color : COLOR0;
 #endif
+#if defined(_SPHERICAL_HARMONICS)
+                fixed3 ambient : COLOR1;
+#endif
 #if defined(_IRIDESCENCE)
-                fixed3 iridescentColor : COLOR1;
+                fixed3 iridescentColor : COLOR2;
```
davidkline-ms: Does moving this from COLOR1 to COLOR2 have any impact on pre-existing materials?

Assets/MixedRealityToolkit/Inspectors/MixedRealityStandardShaderGUI.cs
Cameron-Micka: Of course! I agree we should get this in the release notes and maybe bubble it up higher in the docs. Unity has a similar [obfuscated menu item ](https://docs.unity3d.com/Packages/com.unity.render-pipelines.lightweight@4.0/manual/images/MaterialUpgraders.png) for upgrading a project. I thought about adding the MRTK menu item there too, but Unity keeps changing the location with each release.

Assets/MixedRealityToolkit/Inspectors/TextureCombinerWindow.cs
Cameron-Micka: Good question, it does not unless someone changed the private methods to public and incorporated into some pipeline. But, I don't think many developers know this tool exists yet.

Assets/MixedRealityToolkit/Inspectors/MixedRealityStandardShaderGUI.cs
Cameron-Micka: Maybe once we make the Lightweight Scriptable Render Pipeline is more standard (when we upgrade to Unity 2019+) we can build some tooling that auto detects shaders need to be upgraded? But, Unity doesn't even do this. You just end up with a bunch of magenta materials.

Assets/MixedRealityToolkit/StandardAssets/Shaders/MixedRealityStandard.shader
Cameron-Micka: It does not since this is just changing the interpolator semantic which doesn't have a connection to serialized data. Thanks for checking!

Assets/MixedRealityToolkit/Inspectors/TextureCombinerWindow.cs
davidkline-ms: feels safe.

PR #4312, opened 2019-05-13
Both and Any support in HandJointUtils' FindHand

Ziugy:
## Overview
Support for `Both` and `Any` bit flags in `HandJointUtils`.

## Changes
- Updates `FindHand` in `HandJointUtils` to properly return a hand if `Both` or `Any` flags are past in.

### DIFF
--- a/Assets/MixedRealityToolkit/Providers/Hands/HandJointUtils.cs
+++ b/Assets/MixedRealityToolkit/Providers/Hands/HandJointUtils.cs
@@ -54,7 +54,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 var hand = detectedController as T;
                 if (hand != null)
                 {
-                    if (detectedController.ControllerHandedness == handedness)
+                    if ((detectedController.ControllerHandedness & handedness) != 0)
                     {
                         return hand;
                     }


wiwei:
@lukastoenneMS can you take a peek at this?

@Ziugy, can you add a little more info on how you're planning on using this (i.e. passing in Both/Any ends up being non-deterministic in terms of what gets returned - Both especially doesn't make sense, because there's no way for this function to return the hand joint information for "both" hands at the same time)

lukastoenneMS:
Here's what passing in the various enum values would cause:
* `None`: is 0, so would always fails
* `Left`/`Right`: as expected
* `Both`: either hand will be accepted, first registered controller is returned. Afaik it's not rigorously specified, but the returned hand will be the one that started tracking first. Basically means "i don't care".
* `Other`: this is an awkward value, basically a 3rd hand type that isn't used anywhere to my knowledge. Not sure why it exists in the first place, probably was supposed to allow opposite hand selection somehow. Would always fail because we never register a controller with that value.
* `Any`: same as `Both`, including the `Other` value (which will never be set).

So the only meaningful addition is to support the `Other = Left | Right` combination. I'm fine with that, but would like to redesign the Handedness flags at some point. IMO there should be a plain `enum Handedness {Left, Right}` and a separate mask enum like

> HandednesFlags
> {
>     Left = (1 << Handedness.Left),
>     Right = (1 << Handedness.Right),
>     Both = Left | Right,
> }

That way it's unambiguous if a specific handedness or a combination is expected.

lukastoenneMS:
Logged as feature request here: #4333

Assets/MixedRealityToolkit/Providers/Hands/HandJointUtils.cs
```diff
@@ -54,7 +54,7 @@ public static IMixedRealityHand FindHand(Handedness handedness)
                 var hand = detectedController as T;
                 if (hand != null)
                 {
-                    if (detectedController.ControllerHandedness == handedness)
+                    if ((detectedController.ControllerHandedness & handedness) != 0)
```
wiwei: Please also update the remarks on FindHand above (The given handeness should be either Handedness.Left or Handedness.Right.) - I would just delete the remarks. 

I don't have the ability to suggest the change because that line is in the unchanged section.

wiwei:
Otherwise lgtm too

wiwei:
I'll merge this in and update the comments afterward.

PR #4315, opened 2019-05-13
Update systems and demos to reference the input system via the service registry

davidkline-ms:
This change completes the migration of MRTK code from directly calling into the MixedRealityToolkit object to obtain instances of core services. While this is still a viable pattern for client code, the prefered method is to use MixedRealityServiceRegistry.TryGetService<T>().

This completes the work started with #4099 and is related to the ongoing effort to fully implement #3545.


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/CursorModifier.cs
```diff
@@ -164,7 +181,8 @@ public Vector3 GetModifiedPosition(IMixedRealityCursor cursor)
             }
 
             FocusDetails focusDetails;
-            if (MixedRealityToolkit.InputSystem != null && MixedRealityToolkit.InputSystem.FocusProvider.TryGetFocusDetails(cursor.Pointer, out focusDetails))
+            if (InputSystem?.FocusProvider != null &&
```
wiwei: >InputSystem?.FocusProvider != null &&  [](start = 16, length = 38)

This seems slightly different compared to the past (i.e. there wasn't a FocusProvider null check) - was this intentional to add this focus provider null check? Is that a side effect of all this work?

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/CanvasUtility.cs
```diff
@@ -61,8 +78,8 @@ private void Start()
             {
                 if (canvas.worldCamera == null)
                 {
-                    Debug.Assert(MixedRealityToolkit.InputSystem.FocusProvider.UIRaycastCamera != null, this);
-                    canvas.worldCamera = MixedRealityToolkit.InputSystem.FocusProvider.UIRaycastCamera;
+                    Debug.Assert(InputSystem?.FocusProvider?.UIRaycastCamera != null, this);
```
wiwei: >? [](start = 59, length = 1)

Same question in other places (i.e. why does this add new null checks against FocusProvider?)

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
```diff
@@ -15,18 +15,28 @@ public class InputSystemGlobalListener : MonoBehaviour
...
 
+        private IMixedRealityInputSystem inputSystem = null;
+
         /// <summary>
-        /// Prefer using EnsureInputSystemValid(), which will only await if the input system isn't
-        /// yet valid.
+        /// The active instance of the input system.
         /// </summary>
-        [Obsolete("Prefer using EnsureInputSystemValid()")]
-        protected readonly WaitUntil WaitUntilInputSystemValid = new WaitUntil(() => MixedRealityToolkit.InputSystem != null);
```
wiwei: This is technically a breaking change, and also a break that doesn't accrue any value to other developers.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
```diff
@@ -43,13 +53,13 @@ protected virtual async void Start()
                 }
 
                 lateInitialize = false;
-                MixedRealityToolkit.InputSystem.Register(gameObject);
+                InputSystem?.Register(gameObject);
```
wiwei: >? [](start = 27, length = 1)

Is this ? necessary? It looks like it should be waiting until the InputSystem is non-null (which is why I assume the previous change didn't have the null check)

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
```diff
@@ -273,7 +270,7 @@ public override void OnPostSceneQuery()
 
                 if (isDown)
                 {
-                    MixedRealityToolkit.InputSystem.RaisePointerDragged(this, MixedRealityInputAction.None, currentHandedness, currentInputSource);
```
wiwei: >InputSystem [](start = 40, length = 11)

Any idea why this didn't have a null check before? It looks like this system in Start() waits for input to become valid, so I'm curious how we need to keep checking that it's valid later.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityGazeProvider.cs
```diff
@@ -15,11 +15,6 @@ public interface IMixedRealityGazeProvider
         /// </summary>
         bool Enabled { get; set; }
 
-        /// <summary>
-        /// The <see cref="IMixedRealityInputSystem"/> instance for which this object is providing gaze data.
-        /// </summary>
-        IMixedRealityInputSystem InputSystem { set; }
-
```
wiwei: This is a breaking change as well - thoughts on leaving this obsolete, and then just not using it? Or you guys think we should just axe it and it's okay? 

It feels like there is no customer value here to removing this (it will break them - unless the value is the explicit notification that they need to access the input system separately?)

Assets/MixedRealityToolkit/Providers/Hands/BaseHandVisualizer.cs
```diff
@@ -18,14 +18,31 @@ public class BaseHandVisualizer : MonoBehaviour, IMixedRealityHandVisualizer, IM
...
+            get
+            {
+                if (inputSystem == null)
+                {
+                    MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem);
+                }
+                return inputSystem;
+            }
+        }
+
```
wiwei: This overall change looks fine, it's is weird to me though that we go from one singleton (i.e. MixedRealityToolkit) to another (MixedRealityServiceRegistry) - in this case now we put the onus of caching and reaching out to all of the consumers (i.e. this exact code is copy pasted in a lot of the things here)

If every single caller is handling caching it almost feels like MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem> should cache as well (i.e. why not share some of this code at a lower level)

Assets/MixedRealityToolkit/Providers/Hands/HandJointService.cs
```diff
@@ -33,9 +33,12 @@ public class HandJointService : BaseInputDeviceManager, IMixedRealityHandJointSe
         /// <inheritdoc />
         public override void LateUpdate()
         {
+            if (InputSystem == null) { return; }
```
wiwei: >InputSystem [](start = 16, length = 11)

What's the case where InputSystem is null? It looks like in BaseInputDeviceManager it will set this in the constructor, so this one doesn't appear to be one that tries to get things later.

wiwei:
Alrighty, done with this set of comments.

Assets/MixedRealityToolkit/Providers/Hands/BaseHandVisualizer.cs
davidkline-ms: > we go from one singleton (i.e. MixedRealityToolkit) to another (MixedRealityServiceRegistry)

MixedRealityServiceRegistry is a static c# class that has no scene presence (it is not a Unity object).

> If every single caller is handling caching it almost feels like MixedRealityServiceRegistry.TryGetService should cache as well (i.e. why not share some of this code at a lower level)

Interesting thought. This is actually very similar to the original pattern where many scripts would attempt to get the service if a previous instance had not yet been found (ex: was null)

Assets/MixedRealityToolkit/Providers/Hands/HandJointService.cs
davidkline-ms: This protects against the input device manager being passed a null instance. Perhaps this should be an assert instead.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityGazeProvider.cs
davidkline-ms: This was a set-only property to get the input system into the gaze provider. Client code relying on this would, in general, cause unexpected effects in the system.

We >could< mark it obsolete, but as it was for internal bookkepping, it felt prudent to remove.

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
davidkline-ms: Good question. I have a habit of checking for null.... i can revert.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
davidkline-ms: probably not :)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/CursorModifier.cs
davidkline-ms: it was intentional. we would hit a null reference if the focus provider were ever not setup

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/CanvasUtility.cs
davidkline-ms: to protect against null-reference exceptions if the FP is ever not set

davidkline-ms: i am fine removing them. it just felt wrong to not have them when the code is dereferencing FocusProvider

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
davidkline-ms: i can put something back. the new pattern does not work with a readonly WaitUntil instance. also, it was used in exactly one place (that has been updated)

davidkline-ms: Just confirmed. There are no references in the MRTK codebase other than the call within EnsureInputSystemValid().

Not sure if there is a clean path to keep this... We can discuss

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/CanvasUtility.cs
davidkline-ms: in retrospect, i would like to keep these checks to allow us to avoid expensive nullref exceptions.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
davidkline-ms: removed :)

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
davidkline-ms: removed

Assets/MixedRealityToolkit/Providers/Hands/HandJointService.cs
davidkline-ms: removed the check. it was inconsistent compared to other providers.

Assets/MixedRealityToolkit/Providers/Hands/BaseHandVisualizer.cs
davidkline-ms: As an aside, caching is an optimization and is not required. Any script can request a system at the time it is needed.

Assets/MixedRealityToolkit/Providers/Hands/HandJointService.cs
wiwei: Thanks - this was a case where it felt like it makes no sense for a hand joint service to start up without an input system backing it - it's a fundamental dependency of the class, and we should fail earlier (i.e. in the constructor) if we're not given required dependencies.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs
davidkline-ms: We did a cost/benefit analysis and feel this change is the correct thing to do.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityGazeProvider.cs
davidkline-ms: We did a cost/benefit analysis and feel this change is the correct thing to do.

PR #4318, opened 2019-05-13
SpectatorView camera status monitoring and world anchor persistence

matthejo:
There are two main changes here:
1. Status UI in the Compositor window to tell the user the state of world tracking and world anchor locatability on the camera HoloLens, as well as the name and IP address of the device.
2. Code in the HoloLens camera app to persist and restore WorldAnchors from the device's local anchor store. In the next PR the WorldAnchor will be able to be located from an ArUco marker.


Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/HolographicCamera/HolographicCameraOriginAnchor.cs
```diff
@@ -0,0 +1,57 @@
...
+            using (BinaryReader reader = new BinaryReader(stream))
+            {
+                string command = reader.ReadString();
+
+                switch (command)
+                {
+                    case "CreateSharedAnchor":
+                        {
+                        }
+                        break;
```
chrisfromwork: Should this do anything? or is this just a first pass for future functionality?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/HolographicCamera/HolographicCameraOriginAnchor.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        private void Start()
+        {
+            worldAnchorManager = GetComponent<WorldAnchorManager>();
+            worldAnchorManager.AttachAnchor(gameObject, AnchorName);
+
+            tcpConnectionManager = GetComponent<TCPConnectionManager>();
+            tcpConnectionManager.OnReceive += TcpConnectionManager_OnReceive;
+        }
```
chrisfromwork: nit: based on other subscriptions, it seems this should have an unsubscription on destroy.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Editor/CompositorWindow.cs
```diff
@@ -106,6 +110,35 @@ private void NetworkConnectionGUI()
...
+                    else
+                    {
+                        GUILayout.Label($"Connected to {cameraNetworkManager.HoloLensName} ({cameraNetworkManager.ConnectedIPAddress} -> {cameraNetworkManager.HoloLensIPAddress})");
+                    }
+
+                    string anchorStatusMessage;
+                    if (!cameraNetworkManager.HasTracking)
+                    {
+                        anchorStatusMessage = "Tracking lost";
+                    }
```
chrisfromwork: nit: make these messages constants at the top of the file.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Compositor/HolographicCameraNetworkManager.cs
```diff
@@ -8,6 +8,7 @@
 using System.Collections.Generic;
 using System.IO;
 using UnityEngine;
+using UnityEngine.XR.WSA;
```
chrisfromwork: does this need to be in UNITY_WSA? If you switch platforms to UNITY_ANDROID or UNITY_IOS, does this still resolve?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/HolographicCamera/HolographicCameraOriginAnchor.cs
provencher: I don't like the design of this.

There should be a protocol that handles sorting of commands in a more standardized way.

Personally, I'm a fan of Google's protobuf protocol, as the protocol templates it produces are compiled natively in all languages it supports, including c++ and c#, but also java.

This means you can parse it in a native app on any platform supported by Azure spatial anchors.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/HolographicCamera/HolographicCameraOriginAnchor.cs
provencher: Especially when subscribing to a singleton.

Another nit: initialization should happen on awake, as start is much later in the lifecycle.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Compositor/HolographicCameraNetworkManager.cs
provencher: It does require unity wsa

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/HolographicCamera/HolographicCameraOriginAnchor.cs
matthejo: This is just a first pass for future functionality, the next PR I have pending has this block filled in.

RE: protobof. This is definitely worth looking into. This PR is just the tip of porting a large amount of currently-working code from an internal repository, but I'll investigate what work would be required to change the networking solution to something like protobuf. We also have some tasks to improve performance of the state synchronization, and this could tie into that work as well.

Issue #4334, opened 2019-05-14
MRTK > Add to Scene and Configure menu does not add MixedRealityPlayspace


cre8ivepark:
## Describe the bug
MRTK > Add to Scene and Configure menu does not add MixedRealityPlayspace

## To reproduce

1. File > New Scene
2. Mixed Reality Toolkit > Add to Scene and Configure

![2019-05-15 00_49_40-Unity 2018 3 13f1 Personal - Untitled - MRTK-GitHub - PC, Mac   Linux Standalone](https://user-images.githubusercontent.com/13754172/57712437-6a40a600-7625-11e9-9f8b-c6f8d6dd9ea1.png)

## Result
Only **MixedRealityToolkit** is added. **MixedRealityPlayspace** is missing.

![2019-05-15 00_50_02-Unity 2018 3 13f1 Personal - Untitled - MRTK-GitHub - PC, Mac   Linux Standalone](https://user-images.githubusercontent.com/13754172/57712527-8e9c8280-7625-11e9-80c0-39c23a1d05eb.png)

## Expected behavior
**MixedRealityPlayspace** should be added together with **MixedRealityToolkit** just like other existing example scenes.

![2019-05-15 00_53_15-Unity 2018 3 13f1 Personal - HandInteractionExamples unity - MRTK-GitHub - PC, M](https://user-images.githubusercontent.com/13754172/57712610-bc81c700-7625-11e9-8d3f-2dfa5e7e6b9d.png)

## Setup 

- Unity Version: 2018.3.13f1
- MRTK: mrtk_development branch

Railboy:
This is a change introduced by the new static playspace class.

Is this considered a bug, or is it just new expected behavior? Does anyone know if people make use of the playspace transform prior to runtime?

cre8ivepark:
@Railboy Oh, if that's the case, we should update the existing example scenes and documentation. I can do it once it is clarified.

Railboy:
If people depend on the playspace being there to do setup work then I think I'd call it a bug. I'll start asking around.

cre8ivepark:
Currently, since we have MixedRealityPlayspace in our documentation:
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/GettingStartedWithTheMRTK.html

keveleigh:
Fixed via #4338

PR #4338, opened 2019-05-14
Ensures playspace transform is created in editor

Railboy:
## Overview
Creates a playspace transform and parents main camera under playspace transform in editor.

## Changes
- Fixes: #4334

### DIFF
--- a/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
@@ -488,6 +488,11 @@ namespace Microsoft.MixedReality.Toolkit
             // We'll enforce that here, then tracking can update it to the appropriate position later.
             CameraCache.Main.transform.position = Vector3.zero;
 
+            // This will create the playspace
+            Transform playspace = MixedRealityPlayspace.Transform;
+            // Parent the camera under the playspace
+            CameraCache.Main.transform.parent = playspace;
+
             bool addedComponents = false;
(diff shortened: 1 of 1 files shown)


wiwei:
@Railboy do you know the history for why this got removed in the first place and why that's not valid?

Railboy:
@wiwei This behavior changed when the static MixedRealityPlayspace class was introduced. At that point Playspace creation was removed from MixedRealityToolkit.EnsureMixedRealityRequirements, since it was no longer the MixedRealityToolkit instance's job to actually create the transform. But we should have kept a call to the transform getter in the instance.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -488,6 +488,11 @@ private void EnsureMixedRealityRequirements()
             // We'll enforce that here, then tracking can update it to the appropriate position later.
             CameraCache.Main.transform.position = Vector3.zero;
 
+            // This will create the playspace
+            Transform playspace = MixedRealityPlayspace.Transform;
+            // Parent the camera under the playspace
+            CameraCache.Main.transform.parent = playspace;
```
keveleigh: Looks like this should already happen in the `Transform` call. Were you hitting cases where it wasn't?

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs#L35

Issue #3950, opened 2019-04-12
Adding and configuring components like BoundingBox, ManipulationHandler at runtime?


mtaulty:
## Overview
Adding a behaviour like BoundingBox or ManipulationHandler to an object in the designer seems pretty good - I can simply drag/drop and then I can change 'properties' such as BoundingBox.activation or ManipulationHandler.allowFarManipulation.

However, I am creating objects dynamically and adding these behaviours from code and I find that the various properties that I need to set to 'configure' the behaviour are all private fields.

Is this by design or am I just missing the right way to do this? In the previous toolkit scripts like TwoHandManipulatable.cs had public access to properties to change behaviour when dynamically adding the script and I was making use of them.

## Expected Behavior
I'd like to be able to add components like BoundingBox, ManipulationHandler and there could perhaps be others dynamically to an object created at runtime.

If the fields were protected, I could just derive and I'd be happy with that but, right now, they are private.

## Actual Behavior
I don't seem to be able to dynamically change much on these types if I have added them through code rather than through the editor.

## Steps to reproduce
N/A

## Unity Editor Version
2018.3.11f1

## Mixed Reality Toolkit Release Version
2.0.0-RC1

SimonDarksideJ:
It's certainly not the case that these should be less accessible than the versions in the htk, although some patterns might have changed slightly.
Will try and look later. 

@cre8ivepark  any thoughts why the access pattern might have changed

SimMue:
Similar problem here.
I'm trying to add the interactable script dynamically and this gives me a null-reference-exception in the Update-Method, because the StateManager is not initialized.
It seems like you have to do a lot of manual configuration. That's fine, however as @mtaulty mentioned, most of the fields are private.

## Steps to reproduce
You can simulate this behaviour, by adding the component manually via the editor while in play-mode.

julenka:
> I'd like to be able to add components like BoundingBox, ManipulationHandler and there could perhaps be others dynamically to an object created at runtime.
> If the fields were protected, I could just derive and I'd be happy with that but, right now, they are private.

All ManipulationHandler fields should now be publicly settable as of #4181

julenka:
Looks like the remaining work is:

- verify that manipulationhandler can be added programmatically, and different fields set.
- verify that bounding box can be added programmatically, and configured programatically.

julenka:
[...]

PR #4352, opened 2019-05-14
Expose ability to get and set scaling limits on bounding box at runtime

djohnsomsft:
We're making use of bounding boxes to manipulate dynamic content. Part of this involves dynamically querying the size characteristics of the content and deciding at runtime what the minimum and maximum scale factors should be. Another part of this scenario is detecting when the minimum/maximum are reached to display information to the user.

Exposing these values and the means to update them at runtime (and recalculate the cached parameters based on them) allows completion of this scenario. This also exposes the ability to have more precise control of scaling by setting it in either absolute or relative (default) terms.

partially Fixes:#3950


julenka:
@djohnsomsft thanks for the PR! We are trying to sure that each PR references an issue, could you please reference the issue you are fixing, or create a new one & reference it?

PR #4353, opened 2019-05-14
IMixedRealitySceneSystem interface and definitions

Railboy:
## Overview

Interface for the IMixedRealitySceneSystem service described in #3987

An complete implementation is in progress in draft PR #4107

While we wait for that PR to exit the draft stage @davidkline-ms suggested we discuss the interface, which is ready to be evaluated.

Looking forward to your comments and suggestions!

### DIFF
--- a/Assets/MixedRealityToolkit.Examples/Demos/Audio/Scripts/obj.meta
+++ b/Assets/MixedRealityToolkit.Examples/Demos/Audio/Scripts/obj.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 8cf37e0305e24734f8655565c7063879
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Assets/MixedRealityToolkit/Definitions/SceneSystem/LightingSceneTransitionType.cs
+++ b/Assets/MixedRealityToolkit/Definitions/SceneSystem/LightingSceneTransitionType.cs
@@ -1,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.SceneSystem
+{
+    /// <summary>
+    /// Used by scene service to control how to transition from one lighting scene to another.
+    /// </summary>
+    public enum LightingSceneTransitionType
+    {
+        None,           // Previous lighting scene is unloaded, new lighting scene is loaded. No transition.
+        FadeToBlack,    // Previous lighting scene fades out to black. New lighting scene is loaded, then faded up from black. Useful for smooth transitions between locations.
+        CrossFade,      // Previous lighting scene fades out as new lighting scene fades in. Useful for smooth transitions between lighting setups in the same location. NOTE: This will not work with different skyboxes.
+    }
+}
--- a/Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
+++ b/Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
@@ -1,0 +1,186 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using UnityEngine.SceneManagement;
+
+namespace Microsoft.MixedReality.Toolkit.SceneSystem
+{
+    /// <summary>
+    /// Interface for managing scenes in Unity.
+    /// Scenes are divided into three categories: Manager, Lighting and Content.
+    /// 
+    /// The Manager scene is loaded first and remains loaded for the duration of the app.
+    /// Only one Manager scene is ever loaded, and no scene operation will ever unload it.
+    /// 
+    /// The Lighting scene is a largely empty scene which controls lighting settings.
+    /// Ambient lighting, skybox, sun direction, etc. A default lighting scene is loaded on initialization.
+    /// After that the active lighting scene may be changed at any time via SetLightingScene.
+    /// Only one lighting scene can ever be loaded at a time.
+    /// 
+    /// Content scenes are everything else. These can be loaded and unloaded at will in any combination.
+    /// 
+    /// The scene actions provided improve on unity's SceneManagement events by ensuring that scenes
+    /// are considered valid before the action is invoked.
+    /// </summary>
+    public interface IMixedRealitySceneSystem : IMixedRealityEventSystem, IMixedRealityEventSource, IMixedRealityDataProvider
+    {
+        #region Actions
+        
+        /// <summary>
+        /// Called when scene is has been loaded, activated and is valid.
+        /// Called for all scene types (content, lighting & manager)
+        /// </summary>
+        Action<Scene> OnSceneLoaded { get; set; }
+
+        /// <summary>
+        /// Called when a content scene has been loaded, activated and is valid.
+        /// If multiple scenes have been loaded in one operation, this is called for each loaded scene.
+        /// </summary>
+        Action<Scene> OnContentLoaded { get; set; }
+
+        /// <summary>
+        /// Called when a lighting scene has been loaded, activated and is valid.
+        /// </summary>
+        Action<Scene> OnLightingLoaded { get; set; }
+
+        /// <summary>
+        /// Called just before a content scene unload operation begins.
+        /// If multiple scenes have been unloaded in one operation, this is called for each scene.
+        /// </summary>
+        Action<Scene> OnWillUnloadContent { get; set; }
+
+        /// <summary>
+        /// Called just before a lighting scene unload operation begins.
+        /// </summary>
+        Action<Scene> OnWillUnloadLighting { get; set; }
+
+        /// <summary>
+        /// Called after a content scene has been completely unloaded.
+        /// </summary>
+        Action<string> OnContentUnloaded { get; set; }
+
+        /// <summary>
+        /// Called after a lighting scene has been completely unloaded.
+        /// </summary>
+        Action<string> OnLightingUnloaded { get; set; }
+
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// True if the scene system is loading or unloading content scenes.
+        /// Manager and lighting scenes are ignored.
+        /// </summary>
+        bool SceneOpInProgress { get; }
+
+        /// <summary>
+        /// Progress of the current scene operation, from 0-1.
+        /// A scene operation may include multiple concurrently loaded scenes.
+        /// </summary>
+        float SceneOpProgress { get; }
+
+        /// <summary>
+        /// True when content has been loaded with an activation token and AllowSceneActivation has not been set to true.
+        /// Useful for existing entities that shouldn't act until a newly loaded scene is actually activated.
+        /// </summary>
+        bool WaitingToProceed { get; }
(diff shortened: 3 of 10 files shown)


SimonDarksideJ:
Looking good so far @Railboy

Assets/MixedRealityToolkit.Examples/Demos/Audio/Scripts/obj.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
davidkline-ms: This (and other non-scene system meta files) are unrelated and should not be getting created / modified.

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+    ///
+    /// The Manager scene is loaded first and remains loaded for the duration of the app.
+    /// Only one Manager scene is ever loaded, and no scene operation will ever unload it.
+    ///
+    /// The Lighting scene is a largely empty scene which controls lighting settings.
+    /// Ambient lighting, skybox, sun direction, etc. A default lighting scene is loaded on initialization.
+    /// After that the active lighting scene may be changed at any time via SetLightingScene.
+    /// Only one lighting scene can ever be loaded at a time.
+    ///
+    /// Content scenes are everything else. These can be loaded and unloaded at will in any combination.
```
davidkline-ms: Can there be more than one content scene loaded at any give time?

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+    /// </summary>
+    public interface IMixedRealitySceneSystem : IMixedRealityEventSystem, IMixedRealityEventSource, IMixedRealityDataProvider
+    {
+        #region Actions
+
+        /// <summary>
+        /// Called just before a set of content scenes is loaded.
+        /// Includes names of all scenes about to be loaded.
+        /// </summary>
+        Action<IEnumerable<string>> OnWillLoadContent { get; set; }
```
davidkline-ms: Should this (and similar) be OnBegin* instead of OnWill*?

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        /// <summary>
+        /// Called when scene has been loaded, activated and is valid.
+        /// Called for all scene types (content, lighting & manager)
+        /// Includes scene name
+        /// </summary>
+        Action<string> OnSceneLoaded { get; set; }
+
+        /// <summary>
+        /// Called jus before a scene will be unloaded unloaded
```
davidkline-ms:
```suggestion
        /// Called just before a scene will be unloaded unloaded
```

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+        Action<string> OnWillUnloadLighting { get; set; }
+
+        /// <summary>
+        /// Called after a lighting scene has been completely unloaded.
+        /// Includes scene name.
+        /// </summary>
+        Action<string> OnLightingUnloaded { get; set; }
+
+        /// <summary>
+        /// Called just prior to a scene being loaded.
```
davidkline-ms:
```suggestion
        /// Called just before to a scene being loaded.
```
This is a nit, but adds consistency in the docs :)

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        #endregion
+
+        #region Properties
+
+        /// <summary>
+        /// True if the scene system is loading or unloading content scenes.
+        /// Manager and lighting scenes are ignored.
+        /// </summary>
+        bool SceneOpInProgress { get; }
```
davidkline-ms:
```suggestion
        bool SceneOperationInProgress { get; }
```

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        /// <summary>
+        /// Progress of the current scene operation, from 0-1.
+        /// A scene operation may include multiple concurrently loaded scenes.
+        /// </summary>
+        float SceneOpProgress { get; }
```
davidkline-ms:
```suggestion
        float SceneOperationProgress { get; }
```

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+        Task LoadContentByTag(string tag, LoadSceneMode mode = LoadSceneMode.Additive, SceneActivationToken activationToken = null);
+
+        /// <summary>
+        /// Async method to unload scenes by name.
+        /// If a scene is not loaded, it will be ignored.
+        /// If a scene operation is in progress, no action will be taken.
+        /// </summary>
+        /// <param name="tag">Scene tag</param>
+        /// <returns>Task</returns>
+        Task UnloadContentByTag(string tag);
```
davidkline-ms: If the scene is not tagged, is it unable to be unloaded?

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        /// <summary>
+        /// Returns true if a content scene is fully loaded.
+        /// </summary>
+        /// <param name="sceneName"></param>
+        bool IsContentLoaded(string sceneName);
```
davidkline-ms: Unload uses Tag, but this method implies a scene name... are they the same?

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
Railboy: Correct, at least not by this method. You can still use:
```
Task UnloadContent(string sceneToUnload);
Task UnloadContent(IEnumerable<string> scenesToUnload);
```

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
Railboy: You can't check whether a tag is loaded, only a scene. You may load 5 scenes by tag "foo", then unload 1 scene by name. At that point I'm not sure whether to say whether scenes with tag "foo" are still loaded. So I left it as an individual check.

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
Railboy: Yes, you can load as many content scenes as you want. They're unrestricted.

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
Railboy: Not sure. I'm (roughly) following the OnWillRenderObject pattern but it may not be appropriate here.

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+
+        /// <summary>
+        /// Called after a lighting scene has been completely unloaded.
+        /// Includes scene name.
+        /// </summary>
+        Action<string> OnLightingUnloaded { get; set; }
+
+        /// <summary>
+        /// Called just before a scene is loaded.
+        /// Called for all scene types (content, lighting & manager)
```
StephenHodgson: might wanna use the escaped xml version of the `&` here so it's rendered properly in the documentation.

Assets/MixedRealityToolkit/Definitions/SceneSystem/LightingSceneTransitionType.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.SceneSystem
+{
+    /// <summary>
+    /// Used by scene service to control how to transition from one lighting scene to another.
+    /// </summary>
+    public enum LightingSceneTransitionType
+    {
+        None,           // Previous lighting scene is unloaded, new lighting scene is loaded. No transition.
```
julenka: What is a lighting scene?

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystem.cs
```diff
@@ -0,0 +1,245 @@
...
+        /// If a scene operation is in progress, no action will be taken.
+        /// </summary>
+        /// <param name="scenesToLoad">Names of content scenes to load. Invalid scenes will be ignored.</param>
+        /// <param name="mode">Additive mode will load the content additively. Single mode will first unload all loaded content scenes first.</param>
+        /// <param name="activationToken">
+        /// Optional token for manual scene activation. Useful for loading screens and multiplayer.
+        /// If not null, operation will wait until activationToken's AllowSceneActivation value is true before activating scene objects.
+        /// </param>
+        /// <returns>Task</returns>
+        Task LoadContent(IEnumerable<string> scenesToLoad, LoadSceneMode mode = LoadSceneMode.Additive, SceneActivationToken activationToken = null);
```
julenka: Can you give an explanation of why I would want to use MRTK scene system instead of Unity APIs to load and unload scenes?

Railboy: Sure. First, unity's basic scene manager is full of [little gotchas](https://issuetracker.unity3d.com/issues/scenemanager-dot-getscenebybuildindex-dot-name-returns-an-empty-string-if-scene-is-not-loaded) that nibble away at productivity and cause hard-to-diagnose bugs. This service protects you from those.

Second, keeping a manager scene loaded while you additively load / unload other scenes (which is what both we and Unity recommend you do) can't be done without some sort of scene management system written on top of Unity's basic system. Using our service means you won't have to reinvent that wheel for every project.

Assets/MixedRealityToolkit/Definitions/SceneSystem/LightingSceneTransitionType.cs
Railboy: The Lighting scene is a largely empty scene which controls lighting settings.
Ambient lighting, skybox, sun direction, etc. A default lighting scene is loaded on initialization.
After that the active lighting scene may be changed at any time via SetLightingScene.
Only one lighting scene can ever be loaded at a time.

Railboy:
>It might be out of scope for this review but it would be helpful to provide an example of how the scene system would be used

@julenka I'm saving that for the scene system implementation.

PR #4355, opened 2019-05-14
Indeterminate loader

radicalad:
## Overview
Addition of a simple Indeterminate Loader based on HoloLens 2 system design. Easily tunable for personal needs, the default settings match what the current design is.

Simply enable / disable the indeterminate loader prefab to visualize a loading state in your app.

This addresses feature request: #4408 

New files:
Indeterminate loader prefab
Loader controller script
MRTK standard shader material - Unlit white

image

image

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Loader.meta
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Loader.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 7f3dcf209068f224bb8d4a88407e916f
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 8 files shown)


Railboy:
@radicalad How would you feel about pulling in the IProgressIndicator interface from the scene transition service PR and implementing it here? Others have asked for the progress indicator assets to be pulled into standard assets, and this is a good one to start with.

radicalad:
@Railboy great idea. The scope of this is really just a dynamically animated prefab. I'd suggest fleshing out the loader controls and having multiples - makes sense to have one be driven from the scene transition service.

As far as where it lives - I'd defer to @davidkline-ms. The material is in standard assets. IMO this is a control and not an asset since it is not art driven - just primitives and a script driving the behavior.

julenka:
Thanks @radicalad. We are actually trying to ensure that there is an issue for each PR, could you please reference the issue / feature you are fixing here, or create a new one?

radicalad:
@julenka added a feature request. :)

Issue #4331, opened 2019-05-14
Radial button selection doesn't visually change when set from script


chbecker-ms:
(Ported to GitHub, reference 21387842)

Using either the CurrentIndex member or the SetSelection(int idx) set the current index and generate appropriate callbacks, but don't call SetDimensionIndex on the the ToggleList elements so that they can visually update. That happens in OnSelection(), which is protected.

It seems like ideally SetSelection(idx) should call OnSelection(idx, true).

This would also get around having to call both "set.CurrentIndex = idx" and "set.SetSelection(idx)" to set the current index.

But I may be missing some subtlety here.

killerantz:
The correct approach is to call SetSelection(idx) which calls into the Interactable and sets it to toggled, which should call back to the InteractableToggleCollection to change the Current Index. Changing the current index only effects the collection when enabled, it's a way to set the default selected value through the inspector. **Some recent updates to Interactable broke the round trip and actually no longer executes any code in Interactable.** 

A fix would be to change the SetSelection(index) to call OnSelection(index, true). This will set the visual state of the radial button or toggle, but will not execute any OnClick events, which I think is preferred. The original intention would have triggered the OnClick event for the selected toggle, which doesn't offer much flexibilty if we only want to change the visual state of the radials. I can test and push something up in the next day or so.

keveleigh:
Fixed by #4356.

PR #4356, opened 2019-05-15
Fixed issue where SetSelection was not triggering an update to the ra…

killerantz:
…dial colleciton.

## Overview
There were changed in Interactable that effects how this script changes which radial button is selected. InteractableToggleColleciton.SetSelection(index) used to call into Interactable's pointer click event, but some of the checks do not allow this call to pass, execute a click and call back to the InteractableToggleCollection to set the new selected index.

This fix just brute forces the visual state change without triggering any onClick events in the interactable, which is I think is preferred.

## Changes
- Fixes: #4331  .

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/InteractableToggleCollection.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/InteractableToggleCollection.cs
@@ -45,7 +45,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


julenka:
@killerantz, we are trying to ensure that each PR has an issue assigned to it, could you please create or reference and issue for this bug?

killerantz:
Issue #4331, I put it in the Fixes section.

keveleigh:
@killerantz was this meant to be closed without merging?

killerantz:
Sorry, it should be merged. I thought the close option was for the comment not the PR.

Issue #3897, opened 2019-04-09
Add hand interaction state (touch, grab) to Interactable's states


cre8ivepark:
## Overview
Current Interactable has four states available for HoloLens v1 and WMR controllers: Default, Focus, Pressed, Disabled. By adding additional states for the hand interactions, developers and creators can define visual/audio feedback without writing additional script. Touch and Grab event could be useful. 
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/README_Interactable.html

## Expected Behavior
Developers can define visual/audio feedback for Touch and Grab state in the theme of Interactable without writing code.

## Actual Behavior
Currently, there is no way to define visual/audio feedback in the Interactable. Developers have to write additional code with hand interaction related interfaces.

killerantz:
Adding the states to show up in the themes is easy, but wiring up the input to drive those states will require code. I'm working on some updates for press right now.

cre8ivepark:
@killerantz I guess two additional states 'Touch' and 'Grab' could be added for hand interactions. For hand interaction press, I think we can use existing 'Press' state - However, I am not clear how we can drive hand press which is currently in PressableButton script.

killerantz:
There's a PhysicalTouch state that was added recently, it's in there and will listen for touch events I'll try to make it less reliant on other components. The grab will be something new. I saw in another script, I think the Manipulation Handler, that looks for grab and checks for the literal string of the action description. That seems breakable if the actions are changed. Is there guidance for action label/descriptions to enable grab functionality?

julenka:
@cre8ivepark I think it would be good to clarify the meaning of grabbed and touched. Touched is easy: a near pointer is physically touching the Interactable's collider. Grabbed is less clear. An intercatable will already be in the Pressed state if a near pointer grasps the object. So why couldn't a designer just use the Pressed state? Is it really that you want to distinguish "near pressed" from "far pressed"?

julenka:
Fixed by #4357

PR #4357, opened 2019-05-15
Interactable: Implement near interaction grab states with examples.

killerantz:
## Overview
The latest focus of this work is added the grab state from a near interaction. There was a PhysicalTouch State previously, so I wanted to add examples for that as well.

Working on issue #3897 

1. Added the Grab state to the InteractableStates enum and state list.
2. Fixed a bug where the state list was not getting updated properly to import all the states at runtime, so the grab state was not triggering. At least one state that was added previously was not added to the states list, so I added that - Voice Command.
3. Created a theme for grab, InteractableGrabScaleTheme which will provide a squeeze effect when an item is grabbed using a near interaction.
4. To enable the theme functionality properly, I implemented a custom settings feature in themes, something that was stubbed in, but was not functional. A theme will assign one or more properties, like a color or Vector3 to each state, but custom settings allows for exposing properties that control how the overall theme applies those state values. For the grab scale theme, there is a magnifier value that controls how the scale is applied to give the squeeze effect.
5. Created an OnTouch and OnGrab receiver, for the event system to expose those events and provided an example for them.
6. Created a new states object containing states for Default, Focus, Press, Touch, Grab and Gesture. There is an example of this theme and state system. 
7. As I worked on the themes, when a new theme property was added, the theme settings was closed by default. This was annoying, so I added a line in the inspector code to open the settings by default.
8. While updating the examples scene, I added some audio receivers to show examples of playing audio on click, this was existing functionality, but was not shown anywhere.
9. I also noticed the words on the see-it-say-it label for the holographic buttons were not hooked up. The voice command functionality has been changed recently, so this caused some confusion with what I thought I was doing, which seemed simple enough, but is a much bigger deal. I opened an issue on this. #4598 Updated the Example scene to use the Speech Command listed on the See-it-say-it-label.
10. Updated documentation on states, themes and events.

The listed work started with adding the grab state, but what does this mean for Grip Press from a controller?

In this case, we listening for the specified Input Action (Select) but also looking for physical touch and grab from near interaction. To easily have near and far interactions, with potentially extra feedback for near interaction, this is a nice interaction. Even if we are not using the Grab or Physical Touch states, these interactions now enable feedback, which would have been ignored previously. A NearInteractionGrabbable uses the GripPress action, but Interactable would have ignored it if Select was the chosen Input Action.

The preferred process is to use the Input Action drop down to target the action Interactable will listen to, like grip from a controller or menu button.


julenka:
> Added Grip to the states list

I don't actually understand how I would use 'grip' in Interactable, could you please explain?

julenka:
> routed pointer data to the themes so we can get information about hands and touches.

Could you please explain why somebody might want to get this data? An example to help me understand why this change would be useful would help.

julenka:
> added a drop-down to trigger OnClick using near interaction touch.
I believe I already have implemented this behavior in PR #4077 , could you please explain what you did differently / what is new?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/HolographicButton.prefab
```diff
@@ -389,8 +389,9 @@ MonoBehaviour:
   Dimensions: 1
   CanSelect: 1
   CanDeselect: 1
-  VoiceCommand: 
+  VoiceCommand: Button
   RequiresFocus: 1
+  PhysicalPressOnClickTrigger: 0
```
julenka: I noticed that the interactable has a behavior that's now quite similar to PushButton.cs:
image

In particular, the interactable is now supporting a push-button behavior that's the same as PressableButton.

Could you provide some explanation about why you'd like to have this behavior directly in Interactable? There is an advantage of having all the behavior stored in a single script, but a disadvantage I see is that with two pieces of code implementing the same behavior, any improvements to PressableButton will now also need to be made to Interactable.cs.

Additionally, we would need to all agree on whether we are recommending customers to use Interactable, or PressableButton.cs, or all agree / commit that either component can be used and behavior will be identical. 

Since I could see the decision of whether to implement pressable button directly in interactable could create a bit of discussion, it might make sense to pull that change out of this PR and create a separate PR just for that behavior, so we can have that discussion there. For example, one way I could see to solve this is to abstract away the push button logic into a class, and then have interactable and pressablebutton use the same underlying logic. But that might be a bigger task than you want to do in this PR alone, since you would want to work with @thalbern on this.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PhysicalPressButton.prefab
```diff
@@ -0,0 +1,412 @@
+%YAML 1.1
```
julenka: Interesting to see that you're adding a second pressable button to MRTK SDK. It might be good to think about the customer perspective here.

Putting myself into shoes of external customer, I would be confused about whether I should use PressableButtonPlated.prefab (old prefab) or PhysicalPressButton.prefab.

Assuming that our goal is to make implementing UX as smooth as possible, I wonder if adding another option would actually create confusion / stress for the customer, because they may not know what to do.

A few options I see here:
- Agree that we have two types of buttons and in the prefab naming make it clear what each one is for (e.g. InteractablePhysicalPressbutton.prefab).
- Pull back the change from SDK, and just put this in examples, so that there is just one kind of button in SDK.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnGripReceiver.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities.Editor;
+using UnityEngine.Events;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// A basic grip event receiver
```
julenka: It would be good to explain what the grip receive actually does. I can tell from the class name that this is a basic grip receiver, but don't really know what that means.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnGripReceiver.cs
```diff
@@ -0,0 +1,47 @@
...
+            Name = "OnGrip";
+        }
+
+        public override void OnUpdate(InteractableStates state, Interactable source)
+        {
+            bool changed = state.CurrentState() != lastState;
+
+            bool hasGrip = state.GetState(InteractableStates.InteractableStateEnum.Grip).Value > 0;
+
+            if (hadGrip != hasGrip && changed)
```
julenka: nit: I'm not sure if you need to track changed. I think it's enough to just check if (hadGrip != hasGrip)

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnTouchReceiver.cs
```diff
@@ -0,0 +1,47 @@
...
+            Name = "OnTouch";
+        }
+
+        public override void OnUpdate(InteractableStates state, Interactable source)
+        {
+            bool changed = state.CurrentState() != lastState;
+
+            bool hadTouch = state.GetState(InteractableStates.InteractableStateEnum.PhysicalTouch).Value > 0;
+
+            if (this.hadTouch != hadTouch && changed)
```
julenka: nit again: Also here, not sure if you need changed, hadTouch != this.hadTouch should be enough.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -115,6 +130,11 @@ public class Interactable :
         protected float rollOffTime = 0.25f;
         protected float rollOffTimer = 0.25f;
 
+        // handle near touch interaction
```
julenka: I'm seeing some common code here between PressableButton and Interactable.

We need to ensure that the behavior of pressable button matches Interactable, so therefore any changes / improvements we make to pressable button will need to also be duplicated here, which will make this code quite hard to maintain and likely to diverge from pressable button.

This seems like a very good candidate for removing the code duplication, either by:

1) Instantiating the pressable button at runtime and forwarding / listening to events, so that the logic lives in pressable button.
2) Abstracting out the state machine / logic for PressableButton, and then using it both here and in pressable button.


My intuition is option 2 will pay off more in the end, since then we could re-use that code even for other game engines.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -510,8 +589,51 @@ private void RemovePointer(IMixedRealityPointer pointer)
...
 
+        #region SourceManagement
+
+        /// <summary>
+        /// Adds a source to sources
+        /// </summary>
+        /// <param name="source"></param>
+        private void AddSource(SourceStateEventData source)
+        {
+            if (!sources.Contains(source))
```
julenka: Why not just use a hashset, if you want to have a list of unique sources? A hashset will allow you to add and check for membership in O(1) time, where as this approach is O(N). Note it doesn't really matter for a list this small, but it will also simplify your code (you won't need these two methos, just you just do sources.Add() and sources.Remove() directly.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -510,8 +589,51 @@ private void RemovePointer(IMixedRealityPointer pointer)
...
+        /// <summary>
+        /// Removes a source, lost
+        /// </summary>
+        /// <param name="source"></param>
+        private void RemoveSource(SourceStateEventData source)
+        {
+            sources.Remove(source);
+        }
+
+        /// <summary>
```
julenka: nit: Curious why this is being implemented as a method and not a property. In MRTK we tend to use properties.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -989,18 +1131,111 @@ protected int GetVoiceCommandIndex(string command)
...
 
+        #region NearInteractionHandlers
+        /// <summary>
+        /// Is the eventData a grip action or near pointer?
+        /// </summary>
+        /// <param name="eventData"></param>
+        /// <returns></returns>
+        protected bool IsGrip(IMixedRealityPointer pointer, MixedRealityInputAction action)
+        {
+            if (action.Description.ToLower().IndexOf("grip") > -1 || pointer as IMixedRealityNearPointer != null)
```
julenka: Careful, we are highly discouraging people to do direct string comparisons to determine if a pointer is a grip. Instead, the pattern we try to recommend is expose a mixedrealityaction gripAction, and set that to "Grip". @luis-valverde-ms  FYI as another example of this pattern.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -989,18 +1131,111 @@ protected int GetVoiceCommandIndex(string command)
...
+            {
+                return true;
+            }
+
+            return false;
+        }
+
         void IMixedRealityTouchHandler.OnTouchStarted(HandTrackingInputEventData eventData)
         {
             SetPress(true);
```
julenka: It's curious that both press and physical touch are set to true here. It seems like we should distinguish press from physical touch, or document that the two are the same.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -989,18 +1131,111 @@ protected int GetVoiceCommandIndex(string command)
...
+            if (PhysicalPressOnClickTrigger != PhysicalPressClickBehavior.None)
+            {
+                SetPhysicalTouch(true);
+            }
+
             eventData.Use();
+
+            pressStartPosition = eventData.InputData;
```
julenka: Again, this sort of pointer management and state management should be shared between the pressable button, so that we don't have duplicate state management code in our MRTK (for when we need to fix bugs).

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerData.cs
```diff
@@ -0,0 +1,149 @@
...
+
+    public class InteractablePointerData
+    {
+        public IMixedRealityPointer Pointer;
+        public HandTrackingInputEventData Hand;
+        public InteractablePointerTransform CurrentTransform;
+        public InteractablePointerTransform StartTransform;
+        public bool HasPress { get; protected set; }
+        public bool HasFocus { get; protected set; }
+        public int ActionScore { get; protected set; }
```
julenka: What does actionscore mean

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerData.cs
```diff
@@ -0,0 +1,149 @@
...
+        public Vector3 Direction;
+        public float Distance;
+        public float Percentage;
+        public bool HasTouch;
+        public float MaxDistance;
+        public Vector3 ProjectedDirection;
+        public Vector3 PressedValue;
+    }
+
+    public class InteractablePointerData
```
julenka: I don't understand what this class is used for, could you please explain via a comment?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerDataManager.cs
```diff
@@ -0,0 +1,257 @@
...
+
+using Microsoft.MixedReality.Toolkit.Input;
+using Microsoft.MixedReality.Toolkit.Utilities;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// Manages InteractablePointerData for Interactive
```
julenka: I don't understand this comment.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerDataManager.cs
```diff
@@ -0,0 +1,257 @@
...
+    /// Converts HandTrackingInputEventData, input data and pointer data to InteractablePointerData
+    /// </summary>
+    public class InteractablePointerDataManager
+    {
+        protected Dictionary<uint, InteractablePointerData> pointerData = new Dictionary<uint, InteractablePointerData>();
```
julenka: Explain what the map keys are. Is it pointer id to pointer?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerDataManager.cs
```diff
@@ -0,0 +1,257 @@
...
+
+        /// <summary>
+        /// Get the current list of pointer data
+        /// </summary>
+        /// <returns></returns>
+        public Dictionary<uint, InteractablePointerData> GetPointerData()
```
julenka: I stopped here, want to get feedback on the first part of the review before continuing.

Assets/MixedRealityToolkit.Examples/Demos/UX/Interactables/Scenes/InteractablesExamples.unity
```diff
@@ -38,7 +38,7 @@ RenderSettings:
   m_ReflectionIntensity: 1
   m_CustomReflection: {fileID: 0}
   m_Sun: {fileID: 0}
-  m_IndirectSpecularColor: {r: 0.44657898, g: 0.4964133, b: 0.5748178, a: 1}
+  m_IndirectSpecularColor: {r: 0.44657844, g: 0.49641222, b: 0.57481694, a: 1}
```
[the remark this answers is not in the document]
killerantz: Thanks for all the great comments and questions, there are a lot of great points here. I'll start implementing the actionable feedback and try to answer what I can.

The main purpose of this work was to try to get more feedback options into the Interactable Theme system, like grip (grab) and physical press. The PhysicalPressEventRouter and PressableButton can use Interactable themes for simple feedback, if what we have now is sufficient, then we can reduce the scope of this PR. I was contemplating adding the ability to turn off all interaction handling, so that other components could handle that.

The grip/grab functionality is purely visual - to drive themes, it doesn't trigger an OnClick event. Though the event/receiver system is built to look for state patterns and trigger events, like the InteractableGripReceiver. It provides a Unity event for OnGrip and OnRelease (I should change the current naming).

In order to figure out if a grip is happening, I needed to add a bunch of code though, so I like the idea of abstracting some of this stuff out, especially to reduce the duplicated code. I'm also really interested in anything we can do to abstract pointer, hand and controller data and make that much easier to pass around and decipher intent. There was an effort to create a base button class that we could extend, the idea was to have Interactable extend that class and focus on the themes and feedback.

I wasn't thinking about adding pressable button functionality to Interactable as much as trying to make it easy for a button that already works with air tap and controller interaction, to also understand touch. It's a subtle, but the outcome was the same, duplicated code.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/HolographicButton.prefab
killerantz: That's a good idea. Things will change over time or other use cases will popup that will require changes to button tech or new components to be written, I'm all for keeping that stuff condensed.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PhysicalPressButton.prefab
killerantz: Agreed, if we figure out the above discussions about what functionality needs be here, that will effect which samples we create and what they will be named.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: Just old habit, I'll update this. Also, I should the source stuff to it's own PR. It could be self contained and we can discuss if this is needed, because it doesn't work consistently using the source handlers unless the button is set to global. This is part of an older request to support HoloLens interaction states, Interactive, Observation, Observation Targeted and Targeted, where source status is required.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: I was hoping that's the case, it's really problematic as is. One issue I thought about was users could create their own actions so they could name them anything, so going by name or existing actions could break. On the MR controller side, we would get a mixed reality action for grip, but for HoloLens, we just get Select, but at least we can cast to a near pointer to check intent. I'll reach out to @luis-valverde-ms

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: When the PhysicalPressEventRouter was created, some of this code was added, so I tried to make sure I wasn't breaking any existing patterns. What I got from it was Press was like a fallback state if we weren't supplying feedback for physical touch. So at least something was happen on contact even thought the button doesn't move. Documentation here would help.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerData.cs
killerantz: Early in MRTKv2 I was looking at supporting gestures with hands or controllers. The way the data was structured at the time did not link a MR InputAction to a controller button in the interaction mappings, so there was no way to monitor the controller for the button associated with that action to be released. I believe it's still not as easy as it could be.

I came up with ActionScore as a temporary fix, it could be outdated. A use case was using grip on a controller to pick up an oil can, then while gripping and moving the can around, using trigger to drop oil. We did not want to drop the can prematurely, so action score would store all the values in the controller interaction mappings, at the beginning of an interaction, then wait till the score dropped below that number to handle the release. It's flawed but worked at the time. I need to do another pass here, but was hoping for some guidance.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerDataManager.cs
killerantz: Thanks for the time looking so far.

Assets/MixedRealityToolkit.Examples/Demos/UX/Interactables/Scenes/InteractablesExamples.unity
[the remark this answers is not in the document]
julenka: > In order to figure out if a grip is happening, I needed to add a bunch of code though,

I wan to understand this better. Could you point me to the code in your PR that you are using to determine if Grip is happening?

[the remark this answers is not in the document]
killerantz: Updated the info panel and simplified how grab is being detected. Right now we are detecting NearInteractionGrabbable interaction.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnGripReceiver.cs
killerantz: Added clarification

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/HolographicButton.prefab
killerantz: removed the press OnClick trigger and detecting touch start and deltas

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PhysicalPressButton.prefab
killerantz: removed pressable button examples and prefabs

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: removed pressable button code from Interactable

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: using hashSet now.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: These have been updated, or removed.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: I'm putting this as a TODO, will research later.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: I'll update the documentation. Press is a fallback to Physical Press, so we get some feedback if physical press is not being monitored.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: removed

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerData.cs
killerantz: this class was removed. it was an attempt to capture generic gesture data.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerDataManager.cs
killerantz: this class was removed

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/PointerData/InteractablePointerDataManager.cs
killerantz: yes and removed.

killerantz:
This should be cleaned up and ready for another review.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/HolographicButton.prefab
```diff
@@ -389,8 +389,9 @@ MonoBehaviour:
   Dimensions: 1
   CanSelect: 1
   CanDeselect: 1
-  VoiceCommand: 
+  VoiceCommand: Button
```
julenka: Does this mean that saying "Button" will press the button? That doesn't seem like something we want in our core button, since we don't share that anywhere, and the standard way to press a button with voice is to use "select"

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/HolographicButton.prefab
```diff
@@ -409,7 +410,124 @@ MonoBehaviour:
       m_Calls: []
     m_TypeName: UnityEngine.Events.UnityEvent, UnityEngine.CoreModule, Version=0.0.0.0,
       Culture=neutral, PublicKeyToken=null
-  Events: []
+  Events:
```
julenka: How are these changes to add audio events related to adding the ability to respond to grip?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnClickEffect.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.MixedReality.Toolkit.Utilities.Editor;
+using UnityEngine;
+using UnityEngine.Events;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// A basic receiver for detecting clicks
+    /// </summary>
+    public class InteractableOnClickEffect : ReceiverBase
```
julenka: It's not clear to me from the description + name what this does. I don't think the class only receives clicks, it seems to also perform some effect in response to the clicks. It seems like this is something that spawns a gameobject in response to a click. Is that something we want in the core SDK, or something that would make more sense as an example?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnTouchReceiver.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// A basic touch event receiver for detecting Physical Touch state changes in the Interactable
+    /// When the physical touch states change, these events are triggered.
+    /// </summary>
+    public class InteractableOnTouchReceiver : ReceiverBase
+    {
+        [InspectorField(Type = InspectorField.FieldTypes.Event, Label = "On Touch Off", Tooltip = "Touch has left the object")]
+        public UnityEvent OnTouchOff = new UnityEvent();
```
julenka: Why is there OnTouchOff but on OnTouchOn or OnTouchBegin?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Simulation/ScaleFadeAndDestroy.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    public class ScaleFadeAndDestroy : MonoBehaviour
```
julenka: Can you explain what this class does?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Simulation/ScaleFadeAndDestroy.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        [Tooltip("The value to scale to")]
```
julenka: This seems like a useful script, but I'm not sure why it is in a PR that adds a touch and grip state to interactable. Did you add it by mistake?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/InteractableStateModel.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Collections.Generic;
+using UnityEngine;
```
julenka: Extra using?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/InteractableStates.cs
```diff
@@ -95,7 +99,10 @@ public enum InteractableStateEnum
             new State(){ Index = 10, Name = "Gesture", ActiveIndex = -1, Bit = 0, Value = 0},
             new State(){ Index = 11, Name = "GestureMax", ActiveIndex = -1, Bit = 0, Value = 0},
             new State(){ Index = 12, Name = "Collision", ActiveIndex = -1, Bit = 0, Value = 0},
-            new State(){ Index = 13, Name = "Custom", ActiveIndex = -1, Bit = 0, Value = 0}
+            new State(){ Index = 13, Name = "VoiceCommand", ActiveIndex = -1, Bit = 0, Value = 0},
```
julenka: Did you intend to also add the vice command here? It wasn't in the PR description, was this extra code that snuck in?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableCustomSetting.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
```
julenka: What is this for?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableThemePropertySettings.cs
```diff
@@ -47,6 +47,8 @@ public struct InteractableThemePropertySettings
         public InteractableThemeBase Theme;
         public List<InteractableThemeProperty> Properties;
         public List<InteractableThemeProperty> History;
+        public List<InteractableCustomSetting> CustomSettings;
```
julenka: What are these for? They seem unrelated to the PR description.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -316,7 +316,12 @@ public virtual void RenderCustomInspector()
                             EditorGUI.indentLevel = indentOnSectionStart + 3;
 
                             string prefKey = themeItem.objectReferenceValue.name + "Profiles" + i + "_Theme" + t + "_Edit";
+                            bool hasPref = EditorPrefs.HasKey(prefKey);
                             bool showSettings = EditorPrefs.GetBool(prefKey);
+                            if (!hasPref)
```
julenka: This also seems like an unrelated change to the PR

Documentation/README_Interactable.md
```diff
@@ -109,6 +109,10 @@ You can use Interactable to detect input events other than just OnClick. The Eve
 
 *Example of a hold event*
 
+<img src="../Documentation/Images/Interactable/Event_onClickEffect.png" width="450">
```
julenka: Since you are adding grip states, it would be nice to also document that in the README.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/HolographicButton.prefab
killerantz: The See-it, Say-it label on the Holographic button says, Say "Button". So it seemed like a bug for it not to listen for the voice command so I updated this voice command field to make it match the label. This is a discussion we were just having on a project where we have a toolbar with a set of commands that you can use voice to activate. Though it is typical to say select to trigger a button you are gazing or pointing at, if the command is global, then you want to train the users on what to say to trigger the button when they are not looking at it, that case we would want consistency between the two context.

killerantz: It's been documented. https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Documentation/README_Interactable.md#voice-commands

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/HolographicButton.prefab
killerantz: I was updating the scene to include more info on the panel and looking at documented existing functionality better, we have an audio event that was documented but it was never demoed, now that we have MRTK audio files, I added and audio event to the holographic button, and took a snapshot to replace the one in the docs. This could be a separate PR, it just felt weird to revert it.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnClickEffect.cs
killerantz: This should be another PR and definitely failed on the description.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnTouchReceiver.cs
killerantz: There is an OnTouch event. All ReceiverBase classes come with a Unity event which in this case is called OnTouch. OnTouchOff is the second event. Maybe OnTouchStart and OnTouchEnd would be better, or start/complete?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Simulation/ScaleFadeAndDestroy.cs
killerantz: This is an effect script that that will fade and scale a gameObject over the Lifespan then destroy itself at the end. I use it to provide interaction feedback. Like OnClick and burst will happen.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/InteractableStates.cs
killerantz: This was a bug I noticed while adding the grip state, VoiceCommand was added previously to the enum above, but was never added to the states list.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableCustomSetting.cs
killerantz: I updated the comments. It's used by the themes to expose properties for additional theme settings. Allows more customization of themes.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableThemePropertySettings.cs
killerantz: It was included in the list, but could have been separate. This was something that was stubbed in a few places but never finished and I use them to do the InteractableGrabScaleTheme, plus a couple other themes that have been removed from this PR.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: This was a bug fix made as I was working on the themes, every time a new theme is added the properties were closed by default, it was getting annoying.

killerantz: I'll add to the bug fix section.

killerantz:
I'm going to removed all the source states from this PR as well. There's more discussion to be had about the states source should support, plus the fact that sources are not detected unless the button is set to global. Unfortunately this will not reduce the file count.

killerantz:
I removed half the files, hopefully this will be more palatable. I know it's still a lot though.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/HolographicButton.prefab
killerantz: I can change the see-it-say it label if we want to go back to select, but we would need to make sure the Select keyword is working.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnClickEffect.cs
killerantz: removed

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Simulation/ScaleFadeAndDestroy.cs
killerantz: removed

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Simulation/ScaleFadeAndDestroy.cs
killerantz: This should have been put in the Utilities folder or something, it's a general temporary glow or burst effect. I removed it from this PR.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableCustomSetting.cs
killerantz: Used in the GrabScaleTheme

Documentation/README_Interactable.md
killerantz: Good call, added some documentation about states that were not previously mentioned.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnTouchReceiver.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.MixedReality.Toolkit.Utilities.Editor;
+using UnityEngine.Events;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// A basic touch event receiver for detecting Physical Touch state changes in the Interactable
+    /// When the physical touch states change, these events are triggered.
+    /// </summary>
+    public class InteractableOnTouchReceiver : ReceiverBase
```
julenka: I'm not able to create a receiver that responds to touch events. Could you provide and example of this in the scene? I was able to get grab receiver working.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -585,9 +644,17 @@ public void OnPointerUp(MixedRealityPointerEventData eventData)
                 return;
             }
 
-            if (ShouldListen(eventData.MixedRealityInputAction))
+            bool isGrab = IsGrab(eventData.Pointer, eventData.MixedRealityInputAction);
+
+            if (ShouldListen(eventData.MixedRealityInputAction) || isGrab)
```
julenka: why not just put isGrab logic into ShouldListen? Why is isGrab special?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1003,20 +1035,61 @@ protected int GetVoiceCommandIndex(string command)
...
-        #region IMixedRealityTouchHandler
+        #region NearInteractionHandlers
+        /// <summary>
+        /// Is the eventData a grab action or near pointer?
+        /// </summary>
+        /// <param name="eventData"></param>
+        /// <returns></returns>
+        protected bool IsGrab(IMixedRealityPointer pointer, MixedRealityInputAction action)
+        {
+            if (pointer as IMixedRealityNearPointer != null)
```
julenka: All this will return is whether the interaction is a near pointer, not if it is a grab action.

If you wanted to check for a grab action the way to do it would be to define a sister field to "Input Actions" and compare against that action. Something like this:

```
// Rename "InputActions" to "select action" since this is the MR action we want the interactable to respond to
MixedRealityAction selectAction;
// All events that match the grab actions will be considered grabs
MixedRealityAction grabAction;

bool isGrab(pointer)
{
    return pointer.action == grabAction;
}
```

However, it's possible that you don't actually want to distinguish between grab and select actions like this, since if I understand Interactable correctly, the way to distinguish this would just be to have two different interactables -- one for grab action and one for the select action. Happy to explain more my understanding.

Perhaps what you reall want here is to distinguish between near and far interactions, and to have corresponding "near interaction states" instead of "grip states"

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1003,20 +1035,61 @@ protected int GetVoiceCommandIndex(string command)
...
+            }
+
+            return false;
+        }
+
+        protected bool IsControllerGrip(IMixedRealityPointer pointer)
+        {
+            bool hasGrip = false;
+
+            // TODO : detect a controll grip press or grip pose action
```
julenka: It's generally not great practice in MRTK to have commented out code. Is there a reason you need to keep this here?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: I would like Grab to reflect near grab or grab on controller, otherwise, we are a far interaction press (unless controller). Having two MixedRealityAction dropdowns on the Interactable could add confusion for what may be a minimal use-case. I was thinking about having another component handle the grab from a controller and routing that in. I'm worried about getting too descriptive of what a grab is and is not, because I could be too limiting. I would love to expose more control over how these states are defined or let people route them in.

julenka: When you say "grab on controller", do you mean "the grip button on the controller is pressed when using far interaction"? Interestingly, the current design of the BaseControllerPointer is actually *only* letting the select button to go through the pointer system, and I think very few people know that. The fact that pointers only dispatch when select button is pressed is also causing #4492.

This method as written will actually return false if you try to interact with pointer via the grab button. This is because the method currently just returns true if the pointer is a near pointer. And until #4492 is fixed you won't get events when you press the grab button anyway. 

When I understood the original description from @cre8ivepark , I actually thought when he said "grab" he was meaning "near interaction", as in "hand / controller was near object when the 'select' button pressed". If dev configured interactable to respond to a button other than select, for example the menu button, then it would read "hand/controller was near object when the 'menu' button was pressed".

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: This could have been a remnant of testing a keeping it separate until everything worked.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: I may be looking at this too deeply. The currently implimentation is looking at Grab by testing if a near interaction is in the pressed state. It only works if the NearInteractionGrabbable is present, so this is working as expected.

Do we really need to differentiate the grab on a controller from any other action or will a press event be enough? As @julenka mentioned, it would have the selected action from the Input Action drop-down. I don't like the idea of trying to detect mulitple input types if we have an input action selector to filter input. In this case we are detected near interactions touch and grab, not buttons on a controller, using the select action. This makes sense to me. Is it OK to say Grab for near Interaction Grab, and not be confused with Grip from a controller? We could say Physical Grab, like Physical Touch? I don't like adding words if they are not needed, just throwing that out there for consistency.

Thanks for all the great feedback, you are definitely asking all the right questions.

I'll remove all the controller grip stuff and assume that if someone wants grip they will select it from the input action drop-down and use the press state, like all the other input types. This clears up things for me at least.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: Oh, wait. I'm using the IsGrab value to set the grab state. I'm also using it for handling the event. I could create an out value for grab then combine them.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnTouchReceiver.cs
killerantz: I added an InteractableOnTouchReciever to the NearInteractionSample, the cube with the white faceted sphere inside.

As for the voice command stuff, I was totally confused. There used to be voice recognition code in Interactable that was removed, but I didn't know that. The voice command field would enable a voice command directly from Interactable, but now that it's gone the feilds are confusing. Like the Requires Gaze doesn't really do anything now, so it needs to be removed. The Voice Command field does filter the voice commands, but it should be a drop down pulling a list from the speech input profile. I'll open an issue for this.

killerantz: The documentation for Voice Commands in Interactable is totally wrong now and needs to point to the Speech Input setup.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: It is interesting that the Near Interaction Grab sends a Grip Press action though, but I filter this by checking for the NearPointer.

julenka:
> I noticed when IsGlobal is checked, Interactable gets two press events. There was code that ignored the double fires, but it was mainly to stop InputHander events from repeating pointerHandler events. I think we should have something back in place. There was a coroutine to pause input for a tenth of a second, any better ideas?

Coroutine or keeping track of last time fired both sound fine to me. Sorry I think I removed that in my previous change to interactable. I didn't realize it broke global manip.

killerantz:
I think I removed the coroutine check actually. It was entwined with some code that would check which events fired first btween the InputHandler and the PointerHandler, I'll put that back in.

Issue #4235, opened 2019-05-08
Replace PlatformUtility with class instances


Alexees:
## Describe the problem

At the moment, the MRTK uses its own RuntimePlatform enum to distinguish between platforms. It neither lists all platforms that Unity supports nor does it distinguish between "platforms" on the same platform (OculusGo and mobile devices are both Android platform).

## Describe the solution you'd like

Instead of ever extending this enum with new platforms and thinking about a way to even distinguish between devices on the same platform, this can be replaced by a class deriving mechanism.
Every script that makes use of the enum now makes use of System.Type implementing a simple check interface. Implementing this interface one can provide any logic that defines if conditions for a platform are met, like: "Is the Oculus SDK active AND the platform Android" in which case the class could itself derive from an Android only checker class and extend it to only meet Oculus Go requirements.

## Describe alternatives you've considered

I thought about ScriptableObjects, but they would work the same with the overhead of first creating an instance of any of the derived classes of them. This would bloat the project and is just unnecessary.

Alexees:
Same goes for SupportedControllerType actually. It's listing specific controllers and the base class you need to derive from to get a DeviceManager returns it. This would also be better if each device could provide its own controllers not based on an enum.

Alexees:
I added some pseudo code to clarify my idea better. You basically lay out the current 5 supported platforms the way they are already checked anyways. Everyone can then refine the check to whatever platform specific he needs.

Railboy:
Love this.

StephenHodgson:
Agreed, I like this idea as well.

Having a base device class and interfaces defining each platform and device is an interesting idea.

I'm unsure how it'll play out in the UI for selecting one, and how platform capabilities will be enabled.

We should definitely discuss this idea more.

Alexees:
Adding #5771 because of relation

PR #4358, opened 2019-05-15
Replaced PlatformSupport with a type based approach

Alexees:
BETA:

This Pull Request is a first draft of how a type based platform support approach could look like. It's definitely a breaking change affecting several files listed below.

The main goal was to be able to more granularity define criteria for certain platforms. This goes as far as e.g.:

"I want to have this service under UWP, but only on Desktop and only when I have a controller connected"

MixedRealityDataProvider now uses classes that return a boolean based on the criteria defined in them. The first reason to do this was that Android supports mobile, Oculus Go and who knows what other platforms.
The current classes, deriving from IPlatformSupport, mimic the current implementations of platform checks, so there is no degradation there (at least ;) )

### Affected files:

- **IMixedRealityServiceConfiguration** and everything that derives from it:

Pros:
- Every configuration that can be checked now can be checked
- Is configurable the same as an enum, except for that you can just add classes and it still works

Cons:
- IMixedRealityServiceConfiguration implementors need to rewrite and reserialize a lot
- The new system doesn't make it obvious, that IPlatformSupport has a preprocessor directived method to be implemented, which in turn needs to be wrapped as well (was formerly distinguished in the MixedRealityToolkit check)
- You must not create more than 32 derived classes

I'm not sure if it's a pro or con, but the MaskFields Nothing and Everything fields act the same, because no platform is not permitted and therefore null is also every platform.

Adresses #4235


davidkline-ms:
While this is a breaking change, it has significant value to help make platform extensibility easier. Not requiring a core enum modification to enable platform extensions is a great win.

I will pull this down and try it out today.

Alexees:
that's awesome, let me know what you think

davidkline-ms:
i definitely see the issues with the inspectors. looking forward to seeing your solution

davidkline-ms:
getting the following when trying to enter play mode (uwp build target)

NullReferenceException: Object reference not set to an instance of an object
IPlatformSupportExtension.IsPlatformSupported (IPlatformSupport[] platformSupports) (at Assets/MixedRealityToolkit/Definitions/SupportedPlatforms/IPlatformSupport.cs:22)
Microsoft.MixedReality.Toolkit.MixedRealityToolkit.RegisterService[T] (System.Type concreteType, IPlatformSupport[] supportedPlatforms, System.Object[] args) (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:172)
Microsoft.MixedReality.Toolkit.MixedRealityToolkit.InitializeServiceLocator () (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:371)
Microsoft.MixedReality.Toolkit.MixedRealityToolkit.ResetConfiguration (Microsoft.MixedReality.Toolkit.MixedRealityToolkitConfigurationProfile profile) (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:121)
Microsoft.MixedReality.Toolkit.Editor.MixedRealityToolkitInspector.OnInspectorGUI () (at Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs:105)
UnityEditor.InspectorWindow.DoOnInspectorGUI (System.Boolean rebuildOptimizedGUIBlock, UnityEditor.Editor editor, System.Boolean wasVisible, UnityEngine.Rect& contentRect) (at C:/buildslave/unity/build/Editor/Mono/Inspector/InspectorWindow.cs:1625)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr)

davidkline-ms:
standalone / openvr was able to run in play mode, though some boundary visualization components were not visible... i did not notice the visualization profile having the floor and play area disabled.

Alexees:
yeah, I thought you were going to fix what you needed. I thought I made it clear that it's broken.
I'd like to know if this approach is worth completing before I go in and put effort in repairing the rest.

Alexees:
@davidkline-ms Hey, if you want you can give it another go. I think I completely fixed everything.

davidkline-ms:
Excellent! Thank you again for working on this!

Alexees:
Things to improve:
- the bitmask is based on the sorted list of classes. therefore it's a consecutive matching. right now it's a double for loop that's unnecessary
- the SystemType class should probably be extended to provide the serialization capabilities on it's own
- the lists of class types and names could probably be generated statically on editor launch and recompile

davidkline-ms:
> one question though:
do the 3 current derivates of IMixedRealityProfileConfiguration necessarily have to be structs? if they weren't a lot of their stuff could go into a base class .
same goes for their custom inspectors, that show a lot of the same data.

Good question. Will look into that.

davidkline-ms:
@Alexees, Apologies for taking so long to get back to this PR (was out of the office for a time).

This is a very promising solution! 

As there have been some significant changes to the development branch, it may require some reworking. Do you mind me doing some integration work in a separate branch?

Thanks!

davidkline-ms:
Wondering if there is a hybrid solution that can preserve existing functionality (and profiles), reduce required changes,  while enabling easy expansion..... Investigating

Assets/MixedRealityToolkit/Definitions/SupportedPlatforms/IPlatformSupport.cs
```diff
@@ -0,0 +1,78 @@
...
+                from type in assembly.GetTypes()
+                where typeof(IPlatformSupport).IsAssignableFrom(type)
+                where type != typeof(IPlatformSupport)
+                orderby type.Name
+                select type).ToArray();
+    }
+
+    public static string[] GetSupportedPlatformNames()
+    {
+        return (from assembly in AppDomain.CurrentDomain.GetAssemblies()
```
davidkline-ms: this may not work with the .net backend... will verify

Assets/MixedRealityToolkit/Definitions/SupportedPlatforms/IPlatformSupport.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Collections.Generic;
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
+
+public interface IPlatformSupport
+{
+    bool IsEditorOrRuntimePlatform();
+
+    bool IsCurrentRuntimePlatformSameAs(RuntimePlatform runtimePlatform);
```
davidkline-ms: nit: consider renaming to IsCurrentRuntimePlatform()

Assets/MixedRealityToolkit/Definitions/MixedRealitySpatialObserverConfiguration.cs
```diff
@@ -30,11 +30,25 @@ public struct MixedRealitySpatialObserverConfiguration : IMixedRealityServiceCon
...
         /// <inheritdoc />
-        public SupportedPlatforms RuntimePlatform => runtimePlatform;
+        private IPlatformSupport[] _runtimePlatform;
+
+        /// <inheritdoc />
+        public IPlatformSupport[] RuntimePlatform
+        {
+            get
+            {
+                if (_runtimePlatform == null)
```
davidkline-ms: based on the description, null is the same as everything. null should be the equivalent of "none"

davidkline-ms:
@Alexees, thank you again for the great thought and effort put into this PR.

Closing as we are not taking changes of this magnitude for version 2.0. I will be creating issues tracking this (and #4240) for consideration in MRTK's next major update.

#4691 adds SupportedPlatform enum values for iOS, Web and Lumin. If there are any others that should be added, please file an issue to track.

Thanks!

davidkline-ms:
For additional context, we are actively striving to avoid changes that break customer's customized profiles in v2. For the next major release we are considering revising the configuration profile system and providing tooling to migrate from v2 to the new version.

Assets/MixedRealityToolkit/Definitions/MixedRealitySpatialObserverConfiguration.cs
Alexees: you sure? none wasn't available before. It's the only option to even have an "all platforms" choice, because otherwise classes using the new system would need to include all of them manually.
Alternatively, of course there could be an "All Platforms" class that just returns true no matter what. But what's the use of none?

Alexees:
@davidkline-ms I don't mind you doing integration work, at all ;)

Alexees:
> > one question though:
> > do the 3 current derivates of IMixedRealityProfileConfiguration necessarily have to be structs? if they weren't a lot of their stuff could go into a base class .
> > same goes for their custom inspectors, that show a lot of the same data.
> 
> Good question. Will look into that.

@davidkline-ms did you get an answer on this one? Would make for a good unifying task

Issue #4201, opened 2019-05-07
Can't interact with UI components using a gamepad controller


chbecker-ms:
Run HandInteractionExamplesScene in the editor with an Xbox gamepad connected.
Click on any UI element
You'll hit the following assert:
`
KeyNotFoundException: The given key was not present in the dictionary.
System.Collections.Generic.Dictionary`2[TKey,TValue].get_Item (TKey key) (at <d7ac571ca2d04b2f981d0d886fa067cf>:0)​
Microsoft.MixedReality.Toolkit.Input.DefaultPointerMediator.UpdatePointers () (at Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPointerMediator.cs:111)​
Microsoft.MixedReality.Toolkit.Input.FocusProvider.UpdatePointers () (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:681)​
Microsoft.MixedReality.Toolkit.Input.FocusProvider.Update () (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:443)​
Microsoft.MixedReality.Toolkit.MixedRealityToolkit.UpdateAllServices () (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:994)​
Microsoft.MixedReality.Toolkit.MixedRealityToolkit.Update () (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:777)
`

This is caused by the fact that the gamepad controller is using the pointer created by the GazeProvider. When the controller is created it triggers registration of the pointer, which at the time it has its InputSourceParent set to null, in the mediator via FocusProvider.ResgisterPointers(). It is only a bit later in the same function that the call to MixedRealityToolkit.InputSystem.GazeProvider.GazeInputSource lazily creates the gaze input source and sets it in the gaze pointer. This can be worked around by, for example, changing DefaultPointerMediator.UpdatePointers() to ignore pointers whose input source is not in the registry.

The next problem we hit is that both the gamepad controller and the gaze provider set the gaze pointer pose every frame, leading to unusable interactions. 

I see two possible solutions:
Create a pointer that is driven by both gaze and the gamepad in a coordinated way, in the style of GGVPointer.
Create a virtual controller that produces the right combined inputs to drive any pointer via gaze and gamepad.
[...]

PR #4360, opened 2019-05-15
Report active inputs (non-zero/true) as changed

### LINKED ISSUE #3253: Some controllers don't send input events on each frame if input is active.

luis-valverde-ms:
This aims to fix #3253 . It is a breaking change because it changes the current behavior but I don't think in practice would cause much trouble.

For clarity, I think we should also update naming throughout to talk in terms input Updated or Active instead of input Changed. This will be quite disruptive as it will affect all implementations IMixedRealityInputHandler so we probably want to defer that to v3.

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
@@ -138,7 +138,13 @@ namespace Microsoft.MixedReality.Toolkit.Tests.InputSystem
 
             interaction.BoolData = testValue;
 
-            // Make sure if we set the same value it's false
+            // Make sure if we set again the true value it's true
+            Assert.IsTrue(interaction.Changed);
+
+            // Make sure setting the false value twice reports no change
+            interaction.BoolData = false;
+            Assert.IsTrue(interaction.Changed);
+            interaction.BoolData = false;
             Assert.IsFalse(interaction.Changed);
         }
 
@@ -197,7 +203,13 @@ namespace Microsoft.MixedReality.Toolkit.Tests.InputSystem
 
             interaction.FloatData = testValue;
 
-            // Make sure if we set the same value it's false
+            // Make sure if we set the same non-zero value it's true
+            Assert.IsTrue(interaction.Changed);
+
+            // Make sure setting a zero value twice reports no change
+            interaction.FloatData = 0f;
+            Assert.IsTrue(interaction.Changed);
+            interaction.FloatData = 0f;
             Assert.IsFalse(interaction.Changed);
         }
 
@@ -209,7 +221,7 @@ namespace Microsoft.MixedReality.Toolkit.Tests.InputSystem
         public void Test07_TestVector2()
         {
             bool[] invertAxisValues = { true, false };
-            Vector2[] vectorValues = { new Vector2(1, 1), new Vector2(1, -1) };
+            Vector2[] vectorValues = { Vector2.zero, new Vector2(1, 1), new Vector2(1, -1) };
 
             foreach (var invertXAxis in invertAxisValues)
             {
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
```diff
@@ -138,7 +138,13 @@ public void Test04_TestBoolNoChange()
 
             interaction.BoolData = testValue;
```
wiwei:
```suggestion
            interaction.BoolData = true;
```

wiwei: There's not a lot of value having a bool that is assigned to true, and used whereever the 'true' value is desired - just use true directly (same goes for other parts of this file)

Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_02_InteractionDefinitionTests.cs
```diff
@@ -238,15 +250,29 @@ private static void Test07_TestVector2_internal(bool invertXAxis, bool invertYAx
             i.Vector2Data = vectorValue;
             Vector2 invertedValue = vectorValue * new Vector2(invertXAxis ? -1f : 1f, invertYAxis ? -1f : 1f);
             Assert.AreEqual(invertedValue, i.Vector2Data, msg);
-            Assert.IsTrue(i.Changed, msg);
+            if (vectorValue != Vector2.zero)
+            {
+                Assert.IsTrue(i.Changed, msg);
+            }
+            else
```
wiwei: Nit: It might be helpful to have a comment that this is specifically verifying that changing a value from 0->0 is the single exception that doesn't trigger another change

Assets/MixedRealityToolkit/Definitions/Devices/MixedRealityInteractionMapping.cs
```diff
@@ -372,16 +372,9 @@ public float FloatData
...
-                    Changed = !floatData.Equals(value * -1f);
-                    floatData = value * -1f;
-                }
-                else
-                {
-                    Changed = !floatData.Equals(value);
-                    floatData = value;
-                }
+                float newValue = (invertXAxis ? -1f : 1f) * value;
+                Changed = newValue != 0f || newValue != floatData;
```
wiwei: I'm not very familiar with floating point comparison issues in C# - I know that with C++ this sort of comparison is dangerous because you could get floating point comparison issues (i.e. something that is very near zero reports as zero, something)

I'm curious if you know about this stuffs

luis-valverde-ms: I don't have additional information there but I'd be very surprised if floating point operations in C# were implemented differently to C/C++ in this regard. If `value` is exactly zero, `newValue` should be too and compare as equal against `0f`. Does that address your concern?

luis-valverde-ms:
@wiwei  I'm having seconds thoughts about this change. This is not needed anymore for #4201 , because we're accesing the axes values directly in the controller, and I think we could do with a bit more testing and documentation. I'd also be interesting in @StephenHodgson thoughts, as I think he may have solved this issue already in XRTK.

StephenHodgson:
Yeah I fixed this a long time ago with pretty much the same changes.

I added a second flag so I can tell the difference between a change and an update.

luis-valverde-ms:
Thanks @StephenHodgson , did you keep a single, shared event for changes and updates? Did you hit any issues with code that wasn't prepared to receive input down every frame instead of just the first?

StephenHodgson:
I actually have them as two events.

`Updated` and `Changed`.

That way whomever was doing the implementation of the specific device class could choose which worked best for them.

luis-valverde-ms:
Closing this as I can't spend more time on it and I think we need more careful consideration about what would happen if we were to raise input down events every frame.

PR #4362, opened 2019-05-15
Use pointer rotation too for scaling and rotation manipulations in bounding box

### LINKED ISSUE #4361: Bounding box ignores pointer rotation when scaling or rotating
Unlike in the Shell, only the pointer position is used to drive the scaling or rotation.

luis-valverde-ms:
This fixes #4361


julenka:
FYI: The shell implementation is still different from what this does: it takes the pointer ray, then projects that ray onto the scaling  / rotation axis, and then uses that projected position to figure out the new scale.

However, this is still an improvement, thanks for making the change!

PR #4374, opened 2019-05-15
Merging latest from MRTK_Development

andreiborodin:
This merges latest from mrtk_development into features/spectatorView


chrisfromwork:
we should probably make sure that this doesn't break .net compilation. I think for the most part if you don't open a mrtk scene/configure a mrtk scene, it won't matter.

Issue #4381, opened 2019-05-15
MRTK instance can be initialized twice in editor


Railboy:
## Describe the bug

MRTK instance can be initialized twice in quick succession when entering playmode in editor. This can cause a null reference exception in GazeProvider, which cascades into a series of errors about missing cursors.

## Steps to Reproduce

- Open HandInteractionExamples scene, or any scene with a main camera that already has a GazeProvider component.
- Hit Play.
- Observe errors.

## Expected behavior

MRTK instance should be initialized once. No errors should occur.

gilbdev:
other repro:
- play
- edit some code, unity will auto recompile code and try to keep it played ("edit and continue")
=> mrtk is initialized again (you can see diagnostic twice)

PR #4382, opened 2019-05-15
Prevents multiple MRTK instance initializations on enter playmode

Railboy:
## Overview
- Fixes: #4381


Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -686,8 +686,11 @@ private static void RegisterInstance(MixedRealityToolkit toolkitInstance)
                 // Set the instance to active
                 MixedRealityToolkit.activeInstance = toolkitInstance;
                 toolkitInstances.Add(toolkitInstance);
```
wiwei: Per https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4381, given that the same instance is being initialized multiple times, does this also imply that toolkitInstances will have a duplicate (i.e. two of the same instances?)

Railboy: We're using a hashset to avoid duplicates.

PR #4384, opened 2019-05-15
Removes ExecuteAlways from MixedRealityToolkit

### LINKED ISSUE #4392: In-Editor input simulation is not working
Pressing Play in MRTK scene and then pressing space bar does not bring up hands.

Using mouse to move camera around does not work.

Railboy:
## Overview

Removes the [ExecuteAlways] attribute from MixedRealityToolkit instance to prevent it from being destroyed on entering play mode in editor. This was causing some services like camera and input simulation to not function.

Fixes #4392
Follow-up fix for #4381

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
@@ -16,7 +16,8 @@ namespace Microsoft.MixedReality.Toolkit.Tests.Core
         public void Test_01_InitializeMixedRealityToolkit()
         {
             TestUtilities.CleanupScene();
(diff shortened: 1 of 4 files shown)


wiwei:
Just pulled this locally, it fixes the input simulation in editor now, thanks!

julenka:
@railboy it looks like the tests are failing after this change, could you please fix?

MenelvagorMilsom:
@Railboy there appears to be an issue with PR validation builds at the moment. We're looking into it.

wiwei:
I'm going to pull this in, because it unblocks everyone working out of mrtk_development. I just spun up another PR build on another PR https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4385 which succeeded yesterday (so it's a known good reference). That PR also had the same issue, so I'm assuming it's not related here.

Yes there's always the chance that this introduces something else under the covers but we think that this is important enough to get in to improve the quality of the dev branch.

PR #4387, opened 2019-05-15
Update the MRTK NuGet packages to have only a Foundation and Examples package.

wiwei:
The current NuGet package count is somewhat of a reflection of the folder structure of the MRTK, and ultimately in order to build interesting things you need to have all four MRTK packages (i.e. Core, Providers, SDK, Services) in order to build anything interesting (for example, things in Examples). Rather than ship four packages that all must be included, we'll avoid the overhead of having to juggle four different things and just ship those as a single package.

Note that this also:

1) Still ships Examples as its own package (now with a dependency on Foundation)
2) Removes the test NuGet package (it's not interesting to ship to other customers our tests, which we run internally).


wiwei:
Sending out a PR as non-draft so that I can run this through CI, please don't look yet.

wiwei:
Finally got a build with this working after the great build outage of 2019.

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=2136

MenelvagorMilsom:
@wiwei @macborow, we're going to have to rewrite our signing config xml's with these changes. Unfortunately, they're not checked in to git.

luis-valverde-ms:
@wiwei , is this one still alive? Do you have a dependency on some other issue?

macborow:
Signing config will need to be updated if this PR is completed. The config files are now checked into an internal repo although my [PR](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4996) that switches the builds to use those configs is still in review. Both I and Niall know the details and can fill you in.

wiwei:
This is still on me. There's some internal communication that needs to happen here, because this NuGet work is primarily an internal consumption mechanism that needs some degree of staging. I know this PR is somewhat old now, I'll close this out while I deal with this internally.

PR #4388, opened 2019-05-15
Diagnostics Visual Profiler Improvements

Cameron-Micka:
## Overview
- The Visual Profiler now supports user defined colors for performance thresholds on the frame info panel. Defaults to green when at 95%+ of the target framerate, yellow when at 75%+ of the target framerate, and red otherwise. (A request from multiple users.)

- Portions of the profiler can be dynamically toggled on/off, such as the frame info panel and memory stats.

VisualProfilerChange

- Fixed an issue where required shader permutations for rendering the frame info items using instancing were not included in player builds. This will result in improved rendering performance for the Visual Profiler.

## Verification
Verify the Visual Profiler displays in editor and player .

Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4414


Cameron-Micka:
> Looks good! Looks like there's a merge conflict though.

Thank you Kurtis! Just fixed the merge.

Assets/MixedRealityToolkit/Interfaces/Diagnostics/IMixedRealityDiagnosticsSystem.cs
```diff
@@ -23,6 +23,16 @@ public interface IMixedRealityDiagnosticsSystem : IMixedRealityEventSystem, IMix
         /// </summary>
         bool ShowProfiler { get; set; }
 
+        /// <summary>
```
julenka: @wiwei  FYI this is a breaking change technically, since anybody implementing the MRDiagnosticsSystem would now need to implement these two properties. I think it's fine, but we should add it to the release notes.

Cameron-Micka: Nice catch, thank you Julia!

PR #4389, opened 2019-05-15
Update the Unity Package build step to accurately named.

wiwei:
I think that this step used to build nuget packages as well, until we did the later NuGet work (of packing it up at the end of CI). However, this text appears to have stuck around.

Note that the command it runs actually says "NoNuget"

I noticed this as I was doing some nuget/packaging work, figured I should clean it up while I'm in there.


macborow:
The build definitions have changed since your commit. I think the easiest way to resolve this would be to just reapply the edits in the new files as this looks like a very small update. Sorry about that.

PR #4390, opened 2019-05-16
Reworked Alignment Abstraction

andreiborodin:
This change brings in some of the changes from Build Demo, and reworks a bit the alignment abstraction. Next I will work with Chris to get this hooked up and tested with markers and spectator view. After that is add the ASA implementation.


Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common.Unity/SpatialCoordinateServiceUnityBase.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Numerics;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.MixedReality.SpatialAlignment.Common
```
chrisfromwork: >SpatialAlignment [](start = 33, length = 16)

we should add experimental to this namespace for the time being.

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/QR.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
+guid: 1f2ff87d1bbb54148a536756bcfd486b
```
chrisfromwork: if nothing is in these directories, they probably aren't worth adding yet.

If its felt that these directories need to exist, i would prefer if this were called QRCode compared to QR

chrisfromwork:
> {

This namespace should be fixed.

Do we feel that this logic and the marker visualizer coordinate service logic should be in alignment compared to spectator view? It may not matter for now, but i'm curious where this lives long term.

---
Refers to: Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerDetectorCoordinateService.cs:13 in 2dc1368. [](commit_id = 2dc13685bdcd199aa8c4d88ae95d58003ce49a12, deletion_comment = False)

andreiborodin:
> {

Ideally should be in alignment. I haven't moved in later, trying to keep alignment folder clean.

---
In reply to: [493151565](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4390#issuecomment-493151565) [](ancestors = 493151565)

---
Refers to: Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerDetectorCoordinateService.cs:13 in 2dc1368. [](commit_id = 2dc13685bdcd199aa8c4d88ae95d58003ce49a12, deletion_comment = False)

chrisfromwork:
Adding Eric and Jared for visibility. We are on a time crunch to get these changes in and get out a release of spectator view, so these changes may end up in the feature branch rather quickly. That going said, we appreciate all feedback.

andreiborodin:
> %YAML 1.1

These snuck in as part of my last commit. Do we want to just check them in, appears to be an update to the schema of the .meta? Always have these pending changes.

---
Refers to: Assets/MixedRealityToolkit/StandardAssets/FontsSDFTextures/segoeui SDF.asset:1 in 206adc5. [](commit_id = 206adc5093f13a122498abb877a95e19954ec85a, deletion_comment = True)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerVisualizerCoordinateService.cs
```diff
@@ -48,56 +47,60 @@ public SpatialCoordinate(int id)
...
-        public void UpdateTick()
-        {
-            ThrowIfDisposed();
+            markerCoordinate.ShowMarker();
 
-            if (markerCoordinate != null)
+            while (cancellationToken.IsCancellationRequested)
             {
                 markerCoordinate.WorldToCoordinate = markerInWorldSpace.worldToLocalMatrix;
+                await Task.Delay(1, cancellationToken).IgnoreCancellation(); // Wait a frame
```
matthejo: This won't wait for a frame, it will wait for 1ms, correct? Or will it wait for a full frame because of the synchronization context of being on Unity's main thread? If that's the case, probably worth clarification about why in a comment here.

andreiborodin: Yeah, it will wait a frame because of Unity's synchronization context.

---
In reply to: [284813578](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4390#discussion_r284813578) [](ancestors = 284813578)

andreiborodin: I will add a comment.

---
In reply to: [284814467](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4390#discussion_r284814467) [](ancestors = 284814467,284813578)

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/ISpatialCoordinate.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        event Action StateChanged;
+
+        /// <summary>
+        /// Gets the Id representing this coordinate.
+        /// </summary>
+        string Id { get; }
+
+        // TODO agree with Jared
+        //Vector3 Accuracy { get; }
```
provencher: Seems like finding an accuracy metric might be tough, but it could be very useful

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/ISpatialCoordinate.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        /// <summary>
+        /// Occurs when the value of the <see cref="State"/> property has changed.
+        /// </summary>
+        event Action StateChanged;
+
+        /// <summary>
+        /// Gets the Id representing this coordinate.
+        /// </summary>
+        string Id { get; }
```
provencher: Does the Id need to be a string? Why not a unique int / uid?
It could have a name as well, but I think the core unique id should be an int for fast comparisons.

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/ISpatialCoordinate.cs
jbienzms: In my [Spatial Alignment](http://aka.ms/SpatialAlignment) reference architecture, I [define](https://github.com/jbienzms/SpatialAlignment/blob/Dev/SpatialAlignment-Unity/Assets/SpatialAlignment/Strategies/IAlignmentStrategy.cs) accuracy as a metric of drift in meters. Think of it as "I am confident that this point in Unity space matches the same point in real-world space, plus or minus (x, y, z) meters".

For something like QR, this value could just be hard-coded. It's basically knowing the best possible spatial resolution using the devices camera. (RGB camera frustum projected into real-world space). For something like GPS, this value can come directly from devices which support it. Many bluetooth GPS units report HDOP and VDOP (horizontal and vertical dilution of precision) NMEA sentences. Garmin GPS units have a PGRE NMEA sentance. These can all be used to provide approximate error in meters on all axis in real-time.

provencher: Do you think it would make more sense to set an enum to define orders of magnitude of precision?

1CM, 10CM, 1M, 10M, etc? I think this would be easier to work with than a Vector3 I'll have to process math on

jbienzms: Honestly I'm not a fan of the enum-only approach. Especially if it's only represented as a single unified axis. And if you present the enum on all 3 axis (which I think is important) you're really just trading one type of number for another. One which has less accuracy.

If you feel strongly about the enum though you could easily write an extension method that converts Vector3 to your enum type. It's always possible to go from a higher fidelity type to a lower fidelity one, just not the other way around.

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/ISpatialCoordinate.cs
jbienzms: There's a lot of precident for a string ID. Azure Mobile Services ended up going with string and the new Cloud Spatial Anchors [uses](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.spatialanchors.cloudspatialanchor.identifier#Microsoft_Azure_SpatialAnchors_CloudSpatialAnchor_Identifier) string.

I know strings aren't fast, but they're flexible. A string can represent an int, a GUID, or really any type of key. This provides some flexibility in how the coordinate abstraction system maps into various underlying stores.

provencher: They're flexible but they spit out garbage every time you use them. I understand the precedent, but I still think an int is a better choice.

You can hash other keys and store them as an int.

Regardless, I understand this isn't changing.

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/ISpatialCoordinate.cs
provencher: You can use enum flags to register the different axies, and store a composite enum that way.

I feel strongly about it not just for ease of use, but for performance and storage reasons.

A vector3 is 32bit * 3, while an enum is a byte or at worse 32bit.
You get everything you need in 99% of usecase, for far less data. All this needs to be networked as well, so keeping data down is important.

Beyond that just parsing the tolerance becomes a lot cheaper than doing vector size checks, which require square root operations and consume a lot of operations.

Issue #4183, opened 2019-05-06
Null reference when creating Interactable at runtime


dl4mmers:
## Describe the bug

I'm getting a NullReferenceException @ Interactable.cs:272 when i'm creating gameobjects with  interactable components at runtime. This exception does not occur when i'm creating the gameobjects before runtime in editor. 

## To reproduce

1. Create a new MRTK scene. 
2. Add an empty gameobject and assign the following script:

```
using Microsoft.MixedReality.Toolkit.UI;
using UnityEngine;

public class InteractableTest : MonoBehaviour
{
    void Start()
    {
        GameObject test = GameObject.CreatePrimitive(PrimitiveType.Cube);
        test.AddComponent<Interactable>();
    }
}
```
3. Hit play button
4. Console should print following NullReferenceException

```
NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.UI.Interactable.SetupStates () (at Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs:272)
Microsoft.MixedReality.Toolkit.UI.Interactable.Awake () (at Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs:192)
UnityEngine.GameObject:AddComponent()
InteractableTest:Start() (at Assets/InteractableTest.cs:9)
```

## Expected behavior

States should be setup at runtime without any exception. 

## Your Setup (please complete the following information)

- Unity Version 2018.3.14f1
- Mixed Reality Toolkit 2.0.0-RC1

## Target Platform (please complete the following information)

- HoloLens

SimMue:
Same problem here. Haven't found a nice way how to achieve this yet.
My current workaround copies the interactable settings from a gameobject which I added to the scene via editor.

>`var copiedSettings = GetComponent<Interactable>();
var settings = loadedObj.AddComponent<Interactable>();
settings.enabled = true;
settings.States = copiedSettings.States;
settings.Events = copiedSettings.Events;
settings.OnClick = copiedSettings.OnClick;
return settings;`

This works, however it would be really great if anyone could show me how to do this the right way.

julenka:
Thank you! I'd like to take a look at this one. FYI @thalbern since you are on buttons, and Interactable is like a button :-)

julenka:
[...]

PR #4391, opened 2019-05-16
Fix NullRef when Interactable created at runtime

julenka:
## Overview
Fixes: #4183 by loading the default interactable states if Interactable.States is null on Awake(). This can happen when Interactable is instantiated at runtime.

## Verification
Verified by running repro steps at #4183 and seeing that Interactable is correctly instantiated.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
@@ -927,7 +927,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
         public void LoadStates(string statesAssetName)
         {
             string[] stateLocations = AssetDatabase.FindAssets(statesAssetName);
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -913,6 +919,19 @@ protected IEnumerator InputDownTimer(float time)
...
 
+        /// <summary>
+        /// Sets the states of the interactable to be asset that matches the given asset name
+        /// If more than one asset name is found, uses the first found asset.
+        /// </summary>
+        /// <param name="statesAssetName">The name of the interactable states asset</param>
+        public void LoadStates(string statesAssetName)
+        {
+            string[] stateLocations = AssetDatabase.FindAssets(statesAssetName);
+            Debug.Assert(stateLocations.Length > 0, $"Interactable.LoadStates loading {statesAssetName} but no no {statesAssetName}.asset found");
```
thalbern:
```suggestion
            Debug.Assert(stateLocations.Length > 0, $"Interactable.LoadStates loading {statesAssetName} but no {statesAssetName}.asset found");
```

PR #4398, opened 2019-05-16
Enable UI interactions using an Xbox controller

luis-valverde-ms:
## Overview

Changes in this PR:

- Gave the Xbox controller a pose in the form of a new spatial pointer interaction mapping
- Modified Unity Joystick Manager to create pointers for controllers if they have any assigned in the pointer profile.
- Added an entry in the default pointer profile to create GGV pointers for Xbox controllers.

## Changes
- Fixes: #4201  .


julenka:
Cool! Clicking worked well, bounding box felt okay. Movement felt pretty off to me.

- When I move my joystick left/right, the object moved forward/back
- I couldn't figure out how to get the object to move left / right instead of forward back.
- Suggest implementing the object movement in camera space, as follows:
  - Joystick moves the object up/down/left/right
  - Bumpers move the object forward/back
  - Object can be rotated using secondary bumpers

To see the expected behavior, I recommend bringing up windows mixed reality home and moving objects around using the gamepad there.

julenka:
I also noticed that when controller is in resting position the object moves around a bit, most noticeable with bounding box. I think you need a bit of a deadzone for the joystick.

luis-valverde-ms:
> Movement felt pretty off to me

Was the controller also moving the camera when you tried it? If you disable the input simulation service  objects with a `ManipulationHandler` should move fine left to right and in a wonky way up and down. That is down to the handler script itself, the controller in moving in a plane perpendicular to the camera.

> Bumpers move the object forward/back

Can we leave that for another PR? I'm not planing on spending more time on this.

> when controller is in resting position the object moves around a bit

That should be fixed by #4367

julenka:
> Can we leave that for another PR? I'm not planing on spending more time on this.

Sure, though we should file an issue so we don't forget.

julenka:
> Was the controller also moving the camera when you tried it? If you disable the input simulation service, objects with a ManipulationHandler should move fine left to right and in a wonky way up and down. That is down to the handler script itself, the controller is moving in a plane perpendicular to the camera.

No, I don't think it was. I was using a VR headset, and did not notice any strange camera movement. Perhaps the missing piece is that I was using VR headset + gamepad, and you are just using editor simulation + gamepad?

Assets/MixedRealityToolkit/Providers/UnityInput/XboxController.cs
```diff
@@ -48,12 +48,42 @@ public XboxController(TrackingState trackingState, Handedness controllerHandedne
             new MixedRealityInteractionMapping(13, "B", AxisType.Digital, DeviceInputType.ButtonPress,KeyCode.JoystickButton1),
             new MixedRealityInteractionMapping(14, "X", AxisType.Digital, DeviceInputType.ButtonPress,KeyCode.JoystickButton2),
             new MixedRealityInteractionMapping(15, "Y", AxisType.Digital, DeviceInputType.ButtonPress,KeyCode.JoystickButton3),
+            new MixedRealityInteractionMapping(16, "Spatial Pointer", AxisType.SixDof, DeviceInputType.SpatialPointer, MixedRealityInputAction.None)
```
StephenHodgson: Why would an xbox controller ever have a spatial pointer?

Shouldn't the pointer just default to the user's gaze, and current selections be wired up properly via UI navigation events?

luis-valverde-ms: The idea was that by making the gamepad emulate a six DOF controller we would get interaction with our UI components for free, as they already consume the pointer position and rotation. The alternative is to add logic where required to explicitly handle actions like rotate, scale or move based on deltas. That being said, the more I look at it the more I like the second option.

The PR doesn't seem ready for check in and I can't spend more time on it so I'll close it. Anyone is welcome to take over either resurrecting this or starting anew.

PR #4402, opened 2019-05-16
Move pipeline steps into templates and add internal Release build

macborow:
Extract common parts of build definitions into templates.
Add ci-release template (for the build that creates official packages).

Some things to note or to improve further:
- The standard extension for YAML is .yml, not .yaml. Unfortunately I used the latter for the original pipeline definitions and I did not rename those files in this PR (mrtk_CI and mrtk_PR pipelines point to them, so we would need to switch on the Azure DevOps side). I used .yml extension for any new files.
We could actually change to a different convention - use .yml for top level configs (Azure DevOps UI will automatically locate them and display in a drop down when creating a new build) and .yaml for internal templates (won't be picked up by the UI). Perhaps something to consider later on?
- I tried to move UnityVersion and MRTKVersion variables into a separate template. This works but unfortunately pool:demands does not perform variable expansion, so we would still end up with multiple places where the Unity version needs to be updated. I reverted that change for now - we can revisit that later.
- PR validation now also gained a step to compile with .NET backend. It was convenient to do that instead of splitting the "common.yml" template but I can change that based on feedback.


MenelvagorMilsom:
@macborow has this been tested in pipelines?

macborow:
> @macborow has this been tested in pipelines?

Yes, all 3 of them (in both projects).

macborow:
FYI, another PR check (with unrelated changes) has just failed with the exact same error (using the old build definitions). Perhaps machine got in bad state??

Issue #4419, opened 2019-05-16
'Select' input action in 'Select' speech command does not match the one in the input actions profile


luis-valverde-ms:
## Describe the bug

Using the default profiles, the speech command events raised on response to 'Select' contain a "Toggle Diagnostics" input action even though the speech commands profile show it as "Select". This is caused by the default speech commands profile containing outdated input action information (an example of #4083 ). Fix coming soon.

luis-valverde-ms:
Resolved via #4421

PR #4421, opened 2019-05-16
Updated outdated description and axis constraint for 'Select' input action in default speech commands profile

luis-valverde-ms:
This was causing 'Select' speech comands being ignored when filtering using the 'Select' input action.

Fixes: #4419  .


julenka:
Hm... I just made a change a few days ago to do this exact thing (add 'select') command. How is this different?

luis-valverde-ms:
@julenka , you didn't update the axis constraint which is used when comparing for equality (see MixedRealityInputAction.Equals()).

julenka:
ah okay, thank you!

PR #4424, opened 2019-05-16
Adds multi-scene editor tests / tests for multiple MRTK instances

Railboy:
## Overview
- Updates TestUtilities to support tests with multiple scenes.
- Adds editor tests for multiple MRTK instances.
- Updates Playmode tests for multiple MRTK instances.


Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -594,10 +595,90 @@ public void Test_07_02_DisableServicesByType()
             }
         }
 
+        #region Multiple Instances Tests
+
+        [Test]
+        public void Test_08_01_CreateMultipleInstances()
```
julenka: Per @wiwei suggestion I think we are trying to move away from numberint tests, just use the test name without numbers.

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -22,23 +21,76 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+                if (additiveTestSceneAsset != null)
+                {
+                    AssetDatabase.DeleteAsset(path);
+                }
+            }
+            AssetDatabase.Refresh();
+#endif
+        }
+
+        public static void CreateScenes(int numScenesToCreate = 1)
```
julenka: Can you please comment what this method does?

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
Railboy: Do you think we should make that change in this PR, or in a future PR?

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -22,23 +21,76 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+
+            // Create default test scenes.
             // In the editor this can be done using EditorSceneManager with a default setup.
             // In playmode the scene needs to be set up manually.
 
 #if UNITY_EDITOR
             if (!EditorApplication.isPlaying)
             {
-                EditorSceneManager.NewScene(NewSceneSetup.DefaultGameObjects, NewSceneMode.Single);
+                List<Scene> additiveTestScenesList = new List<Scene>();
```
julenka: nit: why not do something like:

additiveTestScenes = new Scene[numScenesToCreate]

and then fill in the array?

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -22,23 +21,76 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+                if (numScenesToCreate == 1)
+                {   // No need to save this scene, we're just creating one
+                    primaryTestScene = EditorSceneManager.NewScene(NewSceneSetup.DefaultGameObjects, NewSceneMode.Single);
+                }
+                else
+                {
+                    // Make the first scene single so it blows away previously loaded scenes
+                    primaryTestScene = EditorSceneManager.NewScene(NewSceneSetup.DefaultGameObjects, NewSceneMode.Single);
+                    // Save the scene (temporarily) so we can load additively on top of it
+                    EditorSceneManager.SaveScene(primaryTestScene, primaryTestSceneTemporarySavePath);
```
julenka: Where do you later on unsave / remove the scene? Would this cause lots of scenes to show up saved in your assets folder?

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
Railboy: Habit. No real reason.

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
Railboy: They get destroyed in `TestUtilities.TearDownScenes()`

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -54,16 +106,19 @@ public static void InitializePlayspace()
...
-
-            MixedRealityToolkit mixedRealityToolkit = new GameObject("MixedRealityToolkit").AddComponent<MixedRealityToolkit>();
-            MixedRealityToolkit.SetActiveInstance(mixedRealityToolkit);
+            CreateScenes(numScenesToCreate);
+            InitializeMixedRealityToolkit(useDefaultProfile);
+        }
 
+        public static void InitializeMixedRealityToolkit(bool useDefaultProfile = false)
+        {
             if (!MixedRealityToolkit.IsInitialized)
```
julenka: It's suspicious that MRTK can be initialized once, but then that multiple scenes with MRTK can be loaded. What is the determininstic outcome of this method? If one scene with MRTK exists, then MRTK for that scene will be loaded (hopefully, actualy may not be true).

If there are several scenes with MRTK, does this mean that the version of MRTK that will be loaded / initialized is the MRTK from the very first scene?

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -22,23 +21,76 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+                    EditorSceneManager.SaveScene(primaryTestScene, primaryTestSceneTemporarySavePath);
+
+                    for (int i = 1; i < numScenesToCreate; i++)
+                    {
+                        string path = additiveTestSceneTemporarySavePath.Replace("#", additiveTestScenesList.Count.ToString());
+                        // Create subsequent scenes additively
+                        Scene additiveScene = EditorSceneManager.NewScene(NewSceneSetup.DefaultGameObjects, NewSceneMode.Additive);
+                        additiveTestScenesList.Add(additiveScene);
+                        // Save the scene (temporarily) so we can load additively on top of it
+                        EditorSceneManager.SaveScene(additiveScene, path);
```
julenka: Same here, will the scenes get deleted at the end of the test?

Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_03_InputSystemTests.cs
```diff
@@ -78,7 +78,7 @@ public void Test04_TestMixedRealityInputSystemExists()
         [TearDown]
         public void CleanupMixedRealityToolkitTests()
         {
-            TestUtilities.CleanupScene();
+            TestUtilities.CreateScenes();
```
julenka: Do you also need to tear down the scenes in the playmode tests? Interesting that there are only two locations where we create / tear down scenes in edito mode tests.

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
Railboy: >If there are several scenes with MRTK, does this mean that the version of MRTK that will be loaded / initialized is the MRTK from the very first scene?

Yes, the first MRTK instance initialized will remain the active instance until you manually switch it with `MixedRealityToolkit.SetActiveInstance.` All other instances will be inactive.

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -594,10 +595,90 @@ public void Test_07_02_DisableServicesByType()
...
+                Scene scene = SceneManager.GetSceneAt(i);
+                SceneManager.SetActiveScene(scene);
+
+                TestUtilities.InitializeMixedRealityToolkit();
+            }
+
+            foreach (MixedRealityToolkit instance in GameObject.FindObjectsOfType<MixedRealityToolkit>())
+            {
+                MixedRealityToolkit.SetActiveInstance(instance);
+
```
julenka: Does it make sense to also test that the other instances are inactive?

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -594,10 +595,90 @@ public void Test_07_02_DisableServicesByType()
...
+        public void Test_08_04_CreateMultipleInstancesInMultipleScenes()
+        {
+            TestUtilities.CreateScenes(3);
+
+            for (int i = 0; i < SceneManager.sceneCount; i++)
+            {
+                Scene scene = SceneManager.GetSceneAt(i);
+                SceneManager.SetActiveScene(scene);
+
+                TestUtilities.InitializeMixedRealityToolkit();
```
julenka: Why are you initializing MRTK SceneManager.sceneCount times here?

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -594,10 +595,90 @@ public void Test_07_02_DisableServicesByType()
...
+            MixedRealityToolkit.SetActiveInstance(thirdInstance);
+
+            Assert.AreEqual(thirdInstance, MixedRealityToolkit.Instance);
+            Assert.IsTrue(thirdInstance.IsActiveInstance);
+        }
+
+        [Test]
+        public void Test_08_03_DestroyActiveInstance()
+        {
+            TestUtilities.InitializeMixedRealityToolkitScene();
```
julenka: Ooh, this seems like a way that we can actually load / destroy scenes in the playmode tests.

Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/TestFixture_03_InputSystemTests.cs
Railboy: Methods marked with [TearDown] get called after each test is run so it's destroying scene assets every time.

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
Railboy: I've created 3 scenes, and for each loaded scene, I'm setting that scene to active and creating an MRTK instance in it. It's to make sure the instances behave as expected across multiple scenes.

Railboy:
Added comments and made some steps more explicit in the multiple instances tests. Also renamed some methods to make what they're doing clearer.

Railboy:
@wiwei Would like to get this merged so I can start writing tests for #4455

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -22,23 +21,83 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+        public static Scene[] additiveTestScenes = new Scene[0];
 
         public static void InitializeMixedRealityToolkit()
         {
             MixedRealityToolkit.ConfirmInitialized();
         }
 
-        public static void CleanupScene()
+        /// <summary>
+        /// Destroyes all scene assets that were created over the course of testing
```
julenka: nit: typo

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -22,23 +21,83 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+                {
+                    AssetDatabase.DeleteAsset(path);
+                }
+            }
+            AssetDatabase.Refresh();
+#endif
+        }
+
+        /// <summary>
+        /// Creates a number of scenes and loads them additively for testing. Must create a minimum of 1.
```
julenka: Which scene becomes the active scene, after this is loaded? Is it the last one? Should we make it configurable.

Railboy: The latest loaded is always the active scene unless you set it manually.

julenka:
@Railboy looks like the CI check failed. Can you take a look?

Railboy:
@julenka looking into it. Builds on my end and there are no failure details. Maybe a problem with the CI check?

Railboy:
Updated latest tests, also merged in latest mrtk_development and updated those tests as well.

keveleigh:
Merged after local test run and build check. Automated CI is currently down.

Issue #4245, opened 2019-05-08
Simplify enabling/disabling eye tracking feature support with MRTK


Troy-Ferrell:
## Describe the problem

The user has to go to three different components to enable Eye tracking support from MRTK in their app (not counting the need to also enable the capability in every VS build). 

## Describe the solution you'd like

Turning on the feature should be consolidated to one checkbox or at the very least one section for the user to control. 

## Describe alternatives you've considered

It would be awesome if this was just a package to install and by installing, you are auto-enabling the feature support

Yoyozilla:
Active thread discussing the right approach. We should fix this soon to reduce confusion

wiwei:
@Yoyozilla do you have any new information from that thread? Ideally we could have that conversation on this issue.

Yoyozilla:
Added Will and Luis to the thread to discuss

Yoyozilla:
yoyo to follow up on if the fixes we already did are good enough for GA 
Also needs to get the Unity bug number

wiwei:
So this one was interesting - a few things have changed since this was originally filed, and now we're in a slightly better world because of:

1) We have a default HL2 profile, which has the profile changes there by default. 
2) Made the "use eye tracking" toggle on by default.

Note that the gaze capability is still kinda weird - it's still only something that exists in VS, despite having a coming soon status in Unity. That being said, we also added an update to the MRTK build tooling to add the manifest capability in automatically afterward, so that should help.

PR #4427, opened 2019-05-16
Simplify enabling/disabling eye tracking feature support with MRTK, Part 1

wiwei:
There are three steps that you need to take to enable eye tracking, and one of them is flipping a checkbox on the GazeProvider. The interesting thing is that even if this is set to true, the gaze provider will end up defaulting back to HL1 behavior anyway if eye tracking capabilities aren't set up, so there's no harm in having the setting default to true - for folks who have HL1, they will still get HL1 behavior. For folks building on HL2, they will have one less step to take.

Note that this does increase the work for folks who have HL2 and actively want HL1 behavior, however.

https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4245


Documentation/EyeTracking/EyeTracking_BasicSetup.md
```diff
@@ -7,9 +7,11 @@ Alternatively, you can check out our already configured [MRTK Eye Tracking Examp
 For Eye Tracking to work correctly, the following requirements must be met:
 
 1. An _'Eye Gaze Data Provider'_ must be added to the input system. This provides eye tracking data from the platform.
-2. Eye Tracking must be set as the preferred source in the _'Gaze Provider'_. This setting can be toggled at runtime.
-3. The _'Gaze Input'_ capability must be enabled in the application manifest. **Currently this is only available in Visual Studio.**
-4. The HoloLens **must** be calibrated for the current user under system settings.
+2. The _'Gaze Input'_ capability must be enabled in the application manifest. **Currently this is only available in Visual Studio.**
+3. The HoloLens **must** be calibrated for the current user under system settings.
+4. The GazeProvider must have its "Use Eye Tracking" property set to true. Note that true is the default value (so no special
```
sostel: Hm, the order is a bit weird. Maybe we should address the initial things to check in the Unity Editor first, then VS, then the generic hint that the user has to be calibrated. By the way, statement number 3 sounds like that you must calibrate via System Settings, which is incorrect. The calibration should be automatically triggered once a new user puts on the device. Only in case the user skipped that step for some reason, they would need to call it up via System Settings.

sostel:
I see that the simulated eye gaze isn't addressed. Is the plan to have it disabled by default?

wiwei:
> I see that the simulated eye gaze isn't addressed. Is the plan to have it disabled by default?

I'm not handling it in this change - this change doesn't touch any profiles of any sort

Documentation/EyeTracking/EyeTracking_BasicSetup.md
wiwei: Fair point, switched back to original order.

I had done things this way because it was the least important thing for people to be aware of (i.e. it's a no-op) so having it in the middle actually causes them to have to spend attention on something that isn't actually important.

sostel: Yeah, I understood where you're coming from. We could have two sections: 1. Assuming you start out with our default setup. 2. "Troubleshoot" section: "What to check if your Eye Targeting isn't working?"

PR #4433, opened 2019-05-16
Take 2: Fix NullRef when Interactable created at runtime 

julenka:
## Overview
Second attempt after reverting PR #4391 which broke UWP builds.

Fixes: #4183 

Changes:
- Add new method States.GetDefaultInteractableStates(), use those states if States field is null.

## Verification
- Verified by running repro steps at #4183 and seeing that Interactable is correctly instantiated.
- Tested HandInteractionExample in editor
- Added playmode test to verify.


julenka:
Note: this puts the Interactable states into a Resources folder, which will cause the assets to be packaged at runtime so they can be loaded at runtime. MRTK's only other Resources folder is in TextMesh Pro. I'm curious why we don't have other assets as part of resources, for example the prefabs in the MRTK.SDK folder.

Railboy:
>Note: this puts the Interactable states into a Resources folder, which will cause the assets to be packaged at runtime so they can be loaded at runtime.

>I'm curious why we don't have other assets as part of resources, for example the prefabs in the MRTK.SDK folder.

Using Resources is not a good practice. I've had many, many bad experiences with it. It's brittle and it can cause memory leaks.

I think it's fine to use it as a stopgap for one asset to get this problem solved in the short term, but in the future we should probably revisit this. We could store these assets in a scriptable object that interactable prefabs carry a reference to, and we could modify the inspector to enforces that reference.

julenka:
> I think it's fine to use it as a stopgap for one asset to get this problem solved in the short term, but in the future we should probably revisit this. We could store these assets in a scriptable object that interactable prefabs carry a reference to, and we could modify the inspector to enforces that reference.

I'm not sure how that would work. The problem I'm trying to solve in this PR is to enable Interactable to be instantiated in code. The states are already stored as a scriptable object, so storing the info in another scriptable object means we need to load it again.

I'm open to other suggestions / approaches.

Another option is to resolve the bug as "By Design" and say "you cannot instantiate Interactable in Code, instead always instantiate a prefab that already has an interactable on it", however I'd like to avoid that if possible.

julenka:
Reading [this reference](https://blog.theknightsofunity.com/unity-resources-folder-how-to-use-it/), it seems like resources folder may not be a great idea. I'll think about another way to solve this.

julenka:
Seems like it could be easy to re-make the asset in code, to avoid the pattern of using Resources folder

provencher:
One thing to note about the ressources system, is that it's the only way to unload data from memory when something isn't used.

Anything passed in a serialized reference stays in memory as long as the object holding the reference is alive.

For small data however, I'm with Lars that ressources should be avoided.

julenka:
Update: I updated the change to no longer use Resources folder. I think it would be useful in future to provide an example of how to instantiate Interactable at runtime, since even though the null ref is fixed, the resulting interactable is not very useful.

Railboy:
>I'm not sure how that would work. The problem I'm trying to solve in this PR is to enable Interactable to be instantiated in code. Loading scriptable objects at runtime requires them to be in resources folder, if I understand correctly.

You're right, my approach was trying to solve the wrong problem.

The new method works great.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/States.cs
```diff
@@ -23,12 +23,30 @@ public class States : ScriptableObject
         /// </summary>
         private static readonly List<Type> candidateStateTypes = new List<Type>() { typeof(InteractableStates) };
 
+        public static States GetDefaultInteractableStates()
+        {
+            States result = new States();
+            List<State> defaultStates = new List<State>();
+
+            InteractableStates allInteractableStates = new InteractableStates();
+            defaultStates.Add(allInteractableStates.GetState(InteractableStates.InteractableStateEnum.Default));
```
provencher: It would be nice to move the state collection for the default states into a separate function that is virtual, in order to allow someone to override which enums are added to the list in their interactible.

julenka:
Verified works on HoloLens 2, and in editor, and that handinteractionexamples scene still works.

PR #4438, opened 2019-05-16
Calibration touchup - use new compositor.

chrisfromwork:
The original calibration work used some old code behind preprocessor directives. This change moves the calibration scripts to use the new dslr compositor logic.

In this change:
1) The CompositorWrapper script is updated to use CompositionManager and TextureManager for obtaining dslr frames
2) Calibration scenes are updated to use the CompositionWrapper. Note, the VR headset calibration scene has been removed for now. Once we have qr code dlls, work will be done to add this scene back.
3) Plugin dlls were updated based on the UNITY_WSA preprocessor directive. It seems that this can still be true in the editor, which makes for a weird experience. I've updated api scripts to use WINDOWS_UWP instead so that if the editor is building for UNITY_WSA, things still work correctly.


Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CompositorWrapper.cs
```diff
@@ -1,47 +1,66 @@
...
     {
-        public static RenderTexture GetDSLRFeed()
+        /// <summary>
+        /// CompositionManager used to obtain a DSLR stream
+        /// </summary>
+        [Tooltip("CompositionManager used to obtain the DSLR stream")]
+        [SerializeField]
+        private CompositionManager compositionManager = null;
+
+        public RenderTexture GetDSLRFeed()
```
matthejo: >GetDSLRFeed [](start = 29, length = 11)

I think I'd been trying to call this the "VideoCamera" elsewhere

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/SpectatorViewPluginAPI.cs
```diff
@@ -15,14 +15,14 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView.M
     /// </summary>
     public class SpectatorViewPluginAPI
     {
-#if UNITY_WSA
+#if WINDOWS_UWP
```
matthejo: >WINDOWS_UWP [](start = 4, length = 11)

When is this defined compared to UNITY_WSA?  The structure of this file actually confuses me a little bit. In the past when I've had an editor and a WSA version of a plugin, the plugin DLLs have had the same name with .meta files that correctly load them in editor versus UWP.  Maybe SpectatorViewPlugin.Editor should just be a different build of SpectatorViewPlugin with appropriate .meta information?

chrisfromwork: Yeah, i'm not sure the best practice for maintaining this. We could definitely have these plugins built with the same name. But they are built/configured through different visual studio projects because one is uwp friendly while the other is not.

With this specific fix, UNITY_WSA seems to be enabled when runniing things in the editor when set to build for UWP. WINDOWS_UWP is only picked up when generating the actual uwp visual studio solution. So, this fix will make it easier to not get dll missing exceptions in the editor based on build configurations. If its felt that we should name these dlls the same, we can do that, but i'd prefer to check this in now and fix that later. It may be more practical to fix the dll naming when all of this code is in a shared repo.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Calibration/CompositorWrapper.cs
chrisfromwork: Sounds good to me. I'll change this now.

Issue #4261, opened 2019-05-10
Unity editor pauses for a significant amount of time when ending Play mode in a scene when using MRTK 2.0


shaddadjr:
## Describe the bug

The Unity Editor freezes for a significant amount of time when ending Play mode in scene

This has been corroborated by several developers in the HoloDevelopers slack channel:
https://holodevelopers.slack.com/archives/C2H4HT858/p1557497162412800

## To reproduce

1. Import MRTK2.0
2. Configure a scene to use MRTK 2.0
3. Play the scene
4. Exit play mode and observe editor freeze for several seconds (some time as long as 30 seconds)

## Expected behavior

The editor should not freeze when exiting play mode.

## Your Setup (please complete the following information)

- Unity Version [2018.3.11f1]
- MRTK Version [v2.0 RC1 and RC1 refresh]

## Target Platform (please complete the following information)

Unity Editor

StephenHodgson:
See https://github.com/microsoft/MixedRealityToolkit-Unity/issues/3026

julenka:
Note: you can work around this by creating a custom MRTK profile, then creating a custom input profile, and removing the following two Data Providers from the list:

![image](https://user-images.githubusercontent.com/168492/57565822-35102b80-7379-11e9-8500-360881043c2c.png)

julenka:
Closing as duplicate of #3026. Please use the workaround mentioned above. One way we could fix this would be to remove speech input from the default providers, but that would have the disadvantage of removing speech support from MRTK.

julenka:
Re opening based on suggestion from eprovencher that we should avoid running the speech code in edit mode, seems like a path to a solution

ForrestTrepte:
I am seeing lengthy delays in both startup and shutdown as described in #3026, but I couldn't quite tell what the conclusion was to that issue. Disabling speech in the editor sounds like a good interim solution, but would it also be possible to fix the underlying solution? (If it's a Unity issue, make sure they know that it's still a problem?)

StephenHodgson:
Just FYI, I was able to fix this in my fork by not destroying the `DictationRecognizer` & `KeywordRecognizer`s when entering and leaving playmode.

wiwei:
Marked as Urgency-Soon, because it's a huge pain to deal with daily development.

wiwei:
Should be good now, please reopen if folks still hit this with the latest mrtk_dev

chbecker-ms:
[...]

PR #4439, opened 2019-05-16
Move the disposal of the dictation and speech providers off the main Unity thread.

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4261

Starting and stopping play mode can be excruciatingly slow because of the time spent waiting for the dictation and speech providers to shut down. These are ultimately due to slowdowns in the Unity wrapped APIs themselves (and ultimately the UWP APIs)

Note that the "dispose" function on these services isn't a standard monobehavior function - it's specially invoked by the finalizer of the BaseService - this happens off of the main Unity thread.

## Verification
Please fetch this change locally by running:

> git fetch origin pull/4439/head:verifyslowdownpr
> git checkout verifyslowdownpr


wiwei:
So I have to confess that I haven't hit this myself - can folks who have hit this pull this down (see instructions in the "Verification" section") and try this out?

wiwei:
Thanks @cre8ivepark for pulling this down and verifying!

Issue #4351, opened 2019-05-14
Add a default HoloLens 2 profile


wiwei:
A common ask from a number of customers is to have a common HL2 profile, which does things like disabling boundaries, disabling hand visualizations (for perf), etc.

wiwei:
Should have teleport off as well.

julenka:
@wiwei any reason this is still open or should I close it?

keveleigh:
Added via #4442

Issue #4431, opened 2019-05-16
Eye targeting not working with near hand pointers enabled anymore


sostel:
# Describe the bug

Hand focus breaks BaseEyeFocusHandler in Unity Editor (did not test on device).

**Edit**: I noticed that after I remove all of the "Articulated Hand" pointers, namely PokePointer and GrabPointer, the expected behavior would work again: Look at a target and use airtap to select it. I had already removed the DefaultPointer (causing the hand ray to show up) a while ago, because it interferred with eye targeting as soon as a hand would be detected. However, so far, it was still possible to use "Look & tap" for far and poke and grab for near interactions. This is not possible anymore.

## To reproduce

1. Load EyeTrackingDemo-00-RootScene
2. Hit play
3. Move eye cursor over a gem by pressing right mouse button and moving mouse. The gem should start rotating.
4. Press Space to show hands. The rotation stops. An airtap does not cause the gem to be selected.
5. Release Space to hide hands. Eye targeting via BaseEyeFocusHandler remains broken. 

## Expected behavior

Look at a gem and select it via airtap.

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.4.0f1]
- MRTK Version [e.g. last updated 05/16/19 - 1:43 PM]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

## Additional context

Current Input Profile setting that seems to work for "look & tap": 
![image](https://user-images.githubusercontent.com/32180789/57889796-92c5cd00-77ea-11e9-8059-6344f7ee0c66.png)

wiwei:
Okay, figured out what's going on here - after this PR went in:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4270

It looks like this was also causing the new-eye-tracking-style-gaze-provider to also get disabled once hands were in the picture, and you had to say "select" in order to bring them back.

It looks like the intent here is for eye gaze scenarios, we want to go back to what it was before (i.e. if any far interactions active, disable eye gaze, near interactions don't really matter)

PR #4442, opened 2019-05-17
Add a default set of profiles that are optimized for getting started with HoloLens2

wiwei:
The default MRTK profile is good at showing off a lot of the features in VR and AR, but has some extra settings that either aren't necessary in AR (teleportation, boundary) or aren't performant (hand mesh visualization). Furthermore, there are also some settings that must be enabled specially in order to use HL2 features (such as Eye Tracking), which should really be set by default.

This copies a subset of the configuration profiles with deltas as follows:

1. Boundary system is disabled.
2. Teleport system is disabled.
3. Spatial awareness system is disabled.
4. Eye tracking provider is added (https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/EyeTracking/EyeTracking_BasicSetup.html#create-an-eye-gaze-data-provider)
5. Eye simulation is enabled by default (for input simulation).
6. Hand mesh visualization is disabled.
7. Camera profile settings are set to match, so that visual quality in the editor and device will look similar.

#4351 

For reviewers, please comment on where you think this should actually go - I opted not to put it in examples because I felt like for folks who just grab the foundation package, this should also be available to them.

Also if there are other defaults that should be set/updated, let me know.


julenka:
Should we make this the default profile in HandInteractionExamples scene?

sostel:
> Should we make this the default profile in HandInteractionExamples scene?

Good question. There is a bug that prevents hands and eyes to be used together right now [1]. If this is the default profile and it includes both near and far hand interactions, eye input won't be of much use and we would still need a different profile to use eye tracking. Not sure how we want to address this. Thoughts? 

[1] https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4431

davidkline-ms:
> Spatial awareness system is disabled.

Curious as to why this system is disabled?

wiwei:
> > Should we make this the default profile in HandInteractionExamples scene?
> 
> Good question. There is a bug that prevents hands and eyes to be used together right now [1]. If this is the default profile and it includes both near and far hand interactions, eye input won't be of much use and we would still need a different profile to use eye tracking. Not sure how we want to address this. Thoughts?
> 
> [1] #4431

We should fix that other bug separately, so yeah, I can make the change to swap hand interaction examples to use this as its default profile

wiwei:
> > Spatial awareness system is disabled.
> 
> Curious as to why this system is disabled?

I'm open to bringing it back - some of our other interactables and example scenes also had this disabled. Lemme know what folks want to do here.

Cameron-Micka:
> Spatial awareness system is disabled.
> 
> Curious as to why this system is disabled?
> 
> I'm open to bringing it back - some of our other interactables and example scenes also had this disabled. Lemme know what folks want to do here.

We usually disable the spatial awareness system to avoid the spatial mesh being calculated/rendered/etc.

wiwei:
@julenka @keveleigh I updated the hand interaction examples scene to use the new default HL2 profiles

Issue #4397, opened 2019-05-16
Box collider should not be required on BoundingBox script


julenka:
BoxCollider is "required" by the BoundingBox script, but it isn't strictly needed since it can reference another object. 

To reproduce, add a bounding box which references a different box collider not on the game object. I expect that I can do this, but actually because the component requires box collider component, I cannot do this without also adding a dummy box collider component.

julenka:
@johnppella is this something you can take a look at?

julenka:
Looks like the component is no longer required.

julenka:
Correction, it looks like you cannot add bounding box in editor without a collider -- you will get an error when adding the component.

julenka:
Fixed by #4443

Issue #4394, opened 2019-05-16
Having the setting to toggle bounding box visualization enabled in editor will be helpful in debugging


ritijain:
## Describe the problem

While using the bounding box, it would be helpful to have the rigidbody created for bounding box to be visible in inspector and the hierarchy. (I had my collider configuration messed up and so i was not able to get bounding box working, since it used to pop up at a different location in the 3d space.  Enabling this property helped me debug this issue faster)

## Describe the solution you'd like

set the 
hideElementsInInspector  property to be configurable in the inspector.

djee-ms:
I agree, this is a very simple yet valuable thing to have. I would like to understand the rationale of hiding those things in the first place. Creating some `GameObject` at runtime already makes it more difficult to reason about the `BoundingBox`; hiding those in addition increases the burden.

julenka:
@johnppella wrote the Bounding Box, John could you please comment?

julenka:
I will expose this in an upcoming PR

julenka:
I've got a PR that will address this.

wiwei:
I believe this has been addressed by the referenced PR, closing this out.

Issue #4393, opened 2019-05-16
BoundingBox script doesn't work  when ManipulationHandler component is also added to the same gameobject


ritijain:
## Describe the bug

I attempted leveraging the Bounding Box + ManipulationHandler capabilities. Turns out BoundingBox doesn’t work reliably when ManipulationHandler is also active in the same hierarchy. 

## To reproduce

1. Create an object in the MRTK enabled scene
2. Add the ManipulationHandler script to the gameobject.
3. Add the BoundingBox script to the gameobject. (script order is important)
4. Hit play in unity editor
5. try to use the bounding box handles. - Nothing happens.

## Expected behavior
Bounding Box works as expected and allows rotation and scale operations based on handles.

## Screenshots

![image](https://user-images.githubusercontent.com/17660747/57820466-c698e800-7741-11e9-8245-cb8d67a8ecd5.png)

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.10f1]
- MRTK Version [e.g. v2.0 - RC1-Refresh]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- UnityEditor

## Additional context
I tracked down the reason to the pointer event that BoundingBox receives has been marked “used” by manipulation handler sometimes.

wiwei:
@julenka @cre8ivepark what's the intent here? Was this ever designed to be supported, or should we have a more formal system by which there can only be a single component of type "manipulator" or something

julenka:
Yes, bounding box and manipulation handler should work together -- design is for you to be able to manipulate by grabbing the box itself (manipulationhandler), or by using affordances on corners/edges (boundingbox)

cre8ivepark:
Currently, it only works when the Bounding Box is above the Manipulation Handler in the inspector.
Our examples in the HandInteractionExample scene has this order. Need to support both features regardless of the script order.

![2019-05-15 18_28_01-Me - DongYoon Park@microsoft com - Outlook](https://user-images.githubusercontent.com/13754172/58051146-49d18980-7b06-11e9-88a7-7b3bf54293c2.png)

julenka:
Revisting this, I realize this is actually by design. Bounding Box should be above manipulation handler so that it can handle events before manipultion handler gets this. If we allowed both manipulation handler and bounding box to handle events, the object with both move from manipulation handler and scale.

[...]

PR #4443, opened 2019-05-17
Add ability to instantiate and configure BoundingBox from code

julenka:
## Overview
Fixes some bugs in BoundingBox that prevented it from being instantiated at runtime, and expose properties so that all bounding box fields can be changed in code.

Also create a test scene under Examples/UX/BoundingBox that tests instantiating a bounding box and setting various properties, as well as a test 

Additional fixes:
- Don't require collidable on bounding box by no longer extending BaseFocusHandler (it is not needed, we don't use any functionality in BaseFocusHandler).
- Expose property "show debug fields" in editor so people can debug the bounding box.
- Correctly compute bounds of a bounding box when encapsulating many children.

## Changes
- Fixes: #3950 #4394 #4397

## Verification

> git fetch origin pull/4443/head:julenka/bbox
> git checkout julenka/bbox

- Tested HandInteractionExamples scene
- Tested BoundingBoxExamples scene
- Verified manually that UWP builds.


julenka:
One open issue when running through the tests: When you have bounding box added as parent to many children, the bounds are not correct.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -158,54 +158,242 @@ public float ScaleMinimum
...
+
+        public bool ShowWireFrame
+        {
+            get { return showWireframe; }
+            set
+            {
+                if (showWireframe != value)
+                {
+                    showWireframe = value;
+                    CreateRig();
```
djohnsomsft: One consequence of calling CreateRig again is that it will recalculate minimumScale/maximumScale (in CaptureInitialState). Because by default it is set relative to the current scale, if the scale of the object has changed from when it was initialized or SetScaleLimits is called, this will have an indirect consequence of changing the scale limits which is likely not intentional when showing/hiding elements.

djohnsomsft: It looks like this may also destroy the bounding box collider override? DestroyRig destroys cachedTargetCollider, which is set from boundsOverride in SetBoundingBoxCollider.

julenka: > It looks like this may also destroy the bounding box collider override? DestroyRig destroys cachedTargetCollider, which is set from boundsOverride in SetBoundingBoxCollider.

I read through the code and I think this one is okay. The cached target collider (which can be either created in code, or assigned to be the box collider override) only gets destroyed if it was created in code.

julenka: > One consequence of calling CreateRig again is that it will recalculate minimumScale/maximumScale (in CaptureInitialState). Because by default it is set relative to the current scale, if the scale of the object has changed from when it was initialized or SetScaleLimits is called, this will have an indirect consequence of changing the scale limits which is likely not intentional when showing/hiding elements.

Good catch.

julenka:
> Good direction but there are a lot of sideeffects of calling CreateRig as it currently alters a lot of things based on states not fixed to the programmatic properties of the bounding box (see comments inline). It would be a good idea to audit the whole method if used as an update callback to identify where it is taking dependencies on non-explicit inputs to recalculate internal state (i.e. the current transform) and where it might be mutating other properties (i.e. destroying the cached collider or recalculating scale) otherwise we'll get weird behavior.

Great suggestion, I audited the whole method for unintended side effects. Aside from updating the minimum and maximum scales, I actually didn't find anything. Here are my notes. The highlighted areas are things that I noticed were code smells / suggestions for improvements.

image

julenka:
Related issue: #4393

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1,7 +1,8 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See LICENSE in the project root for license information.
+﻿// Licensed under the MIT License. See LICENSE in the project root for license information.
```
julenka: TODO: Fix the removed copyright.

julenka:
We can complete #4461 once this goes in.

julenka:
Cannot run CI checks because the base branch is the stabilization branch.

julenka:
The CI build step to generate an appx package is passing locally for me, but does not pass on CI.

I am running the following command:

```
& 'C:\Program Files\Unity\Hub\Editor\2018.3.9f1\Editor\Unity.exe'
-projectPath
F:\julenka.mrtk\
-executeMethod
Microsoft.MixedReality.Toolkit.Build.Editor.UnityPlayerBuildTools.StartCommandLineBuild
-sceneList
Assets\MixedRealityToolkit.Examples\Demos\HandTracking\Scenes\HandInteractionExamples.unity
-logFile
$logPath
-batchMode
-x86
-buildOutput
f:\julenka.mrtk\\build\UWP_x86_default
-buildTarget
WSAPlayer
-buildAppx
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -405,6 +655,7 @@ public BoxCollider TargetBounds
 
         // True if this game object is a child of the Target one
         private bool isChildOfTarget = false;
+        private readonly string rigRootName = "rigRoot";
```
djohnsomsft: Nit: static readonly

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -693,24 +945,36 @@ private void AddCorners()
...
-                    }
 
                     corners.Add(corner.transform);
                 }
             }
         }
 
+        private Bounds GetMaxBounds(GameObject g)
+        {
+            var b = new Bounds(g.transform.position, Vector3.zero);
```
djohnsomsft: Can't assume the center of the object is necessarily part of its geometry.

julenka: Good call.

julenka:
CI is still failing but I ran it locally and everything looks good. Completing.

Issue #4450, opened 2019-05-17
MRTK doesn't deterministically select a MainCamera / EventSystem.


Railboy:
## Describe the bug

When multiple scenes with MixedRealityToolkit instances and MainCameras are loaded, warnings  and errors are issued about event systems. At runtime, selection of a MainCamera is non-deterministic.

## To reproduce

- Create and save two scenes containing a MixedRealityToolkit instance, MixedRealityPlayspace and MainCamera.
- Open both scenes in editor simultaneously.
- Switch between sections in MixedRealityToolkit inspector or toggle services.
- Observe warnings / errors about multiple EventSystems.

`Too many event systems in the scene. The Teleport System requires only one.`

`Found an existing event system in your scene. The Mixed Reality Toolkit requires only one, and must be found on the main camera.`

`For Mixed Reality Toolkit input to work properly, please remove your other input module(s) and add a MixedRealityInputModule to your main camera.`

## Expected behavior

The toolkit should deterministically select a MainCamera and an accompanying EventSystem without displaying errors. Devs should not be expected to delete anything.

- Unity Version 2019.1.1
- MRTK Version 2.0 mrtk_development branch

PR #4455, opened 2019-05-17
Updates MixedRealityPlayspace and CameraCache to handle multiple scenes

Railboy:
## Overview
Updates MixedRealityPlayspace to handle multiple scenes in editor and at runtime.

The MixedRealityPlayspace script has been converted to a MonoBehaviour. Static methods have not been altered so this is not a breaking change.

When multiple playspaces are detected, the first loaded is given priority and the rest are disabled. If the Camera service or CameraCache detects that a main camera has been disabled, that camera is discarded and a new camera is found or created.

The Camera service is now responsible for supplying a main camera to CameraCache. This service is given the highest service priority to ensure that other services don't pre-empt camera creation. (If the Camera service is disabled, CameraCache returns a default camera.)

## MixedRealityPlayspace Inspector

The playspace inspector now offers the choice to activate a playspace, much like the ability to set the active MixedRealityToolkit instance:

InactivePlayspace

When active, the inspector offers some brief documentation about its purpose and capabilities:

ActivePlayspace

## Changes
- Fixes: #4450

## Known Issues

The creation of a camera and playspace may occasionally throw [this warning.](https://issuetracker.unity3d.com/issues/unity-throws-sendmessage-cannot-be-called-during-awake-checkconsistency-or-onvalidate-warning-when-manipulating-the-ui) The warning is harmless.

## Questions (Asked during Draft phase)
This only searches root game objects for playspace transforms. Can playspace transforms be parented? Do we need to enforce keeping them in the root of the scene?

Do we need to extend our search to main cameras as well?

Should we worry about users enabling / disabling playspace transforms manually?

## Verification
- MixedRealityPlayspace now searches root game objects in loaded scenes. This is potentially expensive - we need to create tests for performance impact.


fast-slow-still:
On the question of whether playspaces can be parented, and enforcing them being at root of scene, we currently use a parent of the playspace node to adjust the camera transform, just as MRTK uses the playspace node to implement teleport and camera movement. While it would be mathematically possible to have both systems using the same playspace node for adjustments, in practice it would lead to drift through floating point error, not to mention the other complications of multiple independent systems trying to set the same transform.

I feel other applications are likely to have other needs to insert transforms into the camera's tree as well, for reasons difficult to anticipate. It seems an important development tool for MRTK to support.

I also wonder if, rather than co-opting and renaming the camera's parent when there is no playspace node, it might be better to insert a playspace node in between the camera's parent and the camera.

julenka:
> Should we worry about users enabling / disabling playspace transforms manually?

I would say "yes" because my experience so far has been that people have tried to enable / disable UX components / MRTK components in ways I didn't expect in the past.

julenka:
> Do we need to extend our search to main cameras as well?

I would assume yes, for the same reason as above.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -39,34 +39,43 @@ public static Transform Transform
...
+                {   // If the camera system is enabled, let it handle parenting the camera.
+                    // Just create a transform.
                     GameObject mixedRealityPlayspaceGo = new GameObject(Name);
                     mixedRealityPlayspace = mixedRealityPlayspaceGo.transform;
-                    CameraCache.Main.transform.SetParent(mixedRealityPlayspace);
                 }
                 else
                 {
-                    if (CameraCache.Main.transform.parent.name != Name)
+                    if (CameraCache.Main.transform.parent == null)
```
SimonDarksideJ: We should be careful about blanket == null operators like this as Unity "may" remove them
It's unclear as to what is the best alternative as yet.

Railboy: I've been taking the (possibly wreckless) approach of ignoring this possibility until they actually follow through. Because the consequences would be so far-reaching that I see no point in trying to get ahead of cleaning up the rubble.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -63,12 +67,39 @@ public MixedRealityCameraProfile CameraProfile
...
             }
         }
+        
+        /// <inheritdoc/>
+        public Camera Main
+        {
+            get
+            {
+                if (main == null || !main.isActiveAndEnabled)
+                {
```
chrisfromwork: This seems like it may break if someone actively changes the main camera without disabling cameras. Compared to checking isActiveAndEnabled, should we check whether the camera is rendering to the main screen? It seems like we may also want to pick the last camera in the render pass because I think you can have multiple cameras so that the first cameras pass is composited into the second.

chrisfromwork: we may also need to check things like [https://docs.unity3d.com/ScriptReference/Camera-targetDisplay.html](https://docs.unity3d.com/ScriptReference/Camera-targetDisplay.html)

It may be worth avoiding obtaining a camera main for android and ios until we have an understanding of what it should return. I'm not sure how ARCore/ARKit transition from the default unity main camera the tracking camera.

---
In reply to: [293876592](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4455#discussion_r293876592) [](ancestors = 293876592)

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -117,13 +148,28 @@ private void ApplySettingsForOpaqueDisplay()
...
+        private void FindOrCreateMainCamera()
+        {
+            // If the cached camera is null, search for main
+            main = Camera.main;
+
+            if (main == null || !main.isActiveAndEnabled)
+            {   // If no main camera was found, create it now
+                Debug.LogWarning("No main camera found. The Mixed Reality Toolkit requires at least one camera in the scene. One will be generated now.");
+                main = new GameObject("Main Camera", typeof(Camera)) { tag = "MainCamera" }.GetComponent<Camera>();
+            }
```
chrisfromwork: Forcing a camera creation seems like it could really disrupt a scene if somehow this function is called while still configuring a scene. I think it may be better practice to document that the camera may be null.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -137,4 +183,5 @@ int IEqualityComparer.GetHashCode(object obj)
             return Mathf.Abs(SourceName.GetHashCode());
         }
     }
-}
\ No newline at end of file
+}
+
```
chrisfromwork: nit: remove extra space.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -63,12 +67,39 @@ public MixedRealityCameraProfile CameraProfile
...
+        private static Camera main;
 
         /// <inheritdoc />
         public override void Initialize()
+        {
+            FindOrCreateMainCamera();
+
+            // Ensure that the camera is parented under the mixed reality playspace
+            main.transform.SetParent(MixedRealityPlayspace.Transform);
+
```
chrisfromwork: In some scenarios, such as spectator view, cameras may already have parent transforms influencing their positions. If a MixedRealityCameraSystem is dynamically added/initialized, I think this will break those experiences.

It seems like we actually need to set the mixed reality playspace transform as the parent of any main camera parent transforms. This should probably be a search up the parenting hierarchy compared to a replacement.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -63,12 +67,39 @@ public MixedRealityCameraProfile CameraProfile
...
+            // There's lots of documented cases that if the camera doesn't start at 0,0,0, things break with the WMR SDK specifically.
+            // We'll enforce that here, then tracking can update it to the appropriate position later.
```
chrisfromwork: If this is WMR SDK specific, this forced position should likely be contained in UNITY_WSA || UNITY_STANDALONE_WINDOWS stub.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
```diff
@@ -63,12 +67,39 @@ public MixedRealityCameraProfile CameraProfile
...
+            main.transform.position = Vector3.zero;
+        }
```
chrisfromwork: Does WMR sdk require the global camera position to be 0,0,0 or the local camera position to be 0,0,0? If we account for any preexisting parent game object transforms, we may need to make sure that the global position is zero here (i'm not sure if setting position sets a true position relative to all parent transforms).

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
```diff
@@ -1,49 +1,84 @@
...
+        /// <summary>
+        /// Returns true if cached camera exists and is enabled.
+        /// </summary>
+        public static bool MainExists
+        {
+            get
+            {
+                if (MixedRealityToolkit.IsCameraSystemEnabled)
+                {   // The camera system is guaranteed to create a camera.
+                    return true;
```
chrisfromwork: I think we should have a more robust check for this compared to always assuming it will be true. I'm not sure if the long term plans are for MixedRealityCameraSystems to be used only in wmr experiences. We may as well add an additional flag in MixedRealityToolkit to check for this compared to building out the assumption that IsCameraSystemEnabled == Camera exists

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
```diff
@@ -1,49 +1,84 @@
...
+            if (mainCamera == null || !mainCamera.isActiveAndEnabled)
+            {   // If no main camera was found, create it now
+                Debug.LogWarning("No main camera found. The Mixed Reality Toolkit requires at least one camera in the scene. One will be generated now.");
+                mainCamera = new GameObject("Main Camera", typeof(Camera)) { tag = "MainCamera" }.GetComponent<Camera>();
+            }
+
+            // Cache the main camera
+            cachedCamera = mainCamera;
+            // Parent the camera under our play space
+            cachedCamera.transform.SetParent(MixedRealityPlayspace.Transform);
```
chrisfromwork: >MixedRealityPlayspace [](start = 45, length = 21)

We should check whether the playspace exists here, someone may try and use this class without an mrtk configured scene.

Railboy: Calling MixedRealityPlayspace.Transform automatically generates the transform so it's guaranteed to exist.

chrisfromwork: Since this class is static, it reads like it should work independent of mrtk playspaces/components. If there is no camera system, whats the fallout of not creating a playspace/not adding the camera to the playspace? It feels like parenting to a playspace should happen in MixedRealityToolkit.CameraSystem.Main's getter compared to in this class.

---
In reply to: [293890000](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4455#discussion_r293890000) [](ancestors = 293890000)

chrisfromwork:
Changing interface definitions is a breaking change, updated labels.

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
Railboy: I think it would remove a lot of red tape to make this a requirement of the service interface. The Main property is simply expected to deliver a camera if the service is enabled. Not enabling the service is equivalent to saying 'I don't want MRTK to create / manage a camera for me.'

There are still questions about whether it should be the same camera every time, or how to deal with multiple cameras, or whether the main camera is the one rendering, etc.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
Railboy: Let's start an issue where we track these experience requirements and how our current setup breaks them. We can use it to discuss whether custom camera system implementations can accommodate them / whether changes need to be made to the camera cache / playspace.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
Railboy: Not sure. This has been the default behavior up to this point so I don't want to change it without a lot of testing. Probably belongs in another PR.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
Railboy: See below.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -166,6 +152,101 @@ public static void PerformTransformation(Action<Transform> transformation)
             transformation?.Invoke(Transform);
         }
 
+        public static void SetActivePlayspace(MixedRealityPlayspace playspace)
```
chrisfromwork: nit: add summary

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -166,6 +152,101 @@ public static void PerformTransformation(Action<Transform> transformation)
...
+            if (mixedRealityPlayspace != null)
+            {
+                mixedRealityPlayspace.gameObject.SetActive(false);
+            }
+
+            mixedRealityPlayspace = playspace;
+            mixedRealityPlayspace.gameObject.SetActive(true);
+        }
+
+        public static bool IsActivePlayspace(MixedRealityPlayspace playspace)
```
chrisfromwork: nit: add summary

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
Railboy: I agree - IMO the camera cache class should not exist. The camera system should be the only thing responsible for creating or parenting cameras. Creating the Main property on CameraSystem is a baby step in that direction.

Removing CameraCache altogether is beyond the scope of this PR. The goal here is just to ensure that you can additively load multiple cameras / playspace transforms without breaking the system.

We should open a proposal for removing CameraCache so we can discuss impact.

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
Railboy: Shelving this discussion for now but let's bring it back up in the camera setup issue we open. Finding the 'right' main camera is a big topic.

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
Railboy: Created: #4896

Assets/MixedRealityToolkit.Services/CameraSystem/MixedRealityCameraSystem.cs
Railboy: Let's discuss in #4896

andreiborodin:
It looks good, but I haven't done a super deep dive to understand the implications. The biggest issue with multi-scene is the possible expectation by the developer for the configuration to be picked up from the scene that is most relevant to the object requesting it. If we enforce the singleton, we don't open this can of worms, but I understand why people may ask for this. Have you considered mapping access to the instances based on the ".scene" of the gameobject requesting it?

Railboy:
>Have you considered mapping access to the instances based on the ".scene" of the gameobject requesting it?

@andreiborodin I have. I think there's potential there, though it gets a bit tricky to enforce when you can move playspace transforms across scenes. I've also considered mapping them to specific MRTK instances with a serialized field. If we ever implement a 'favor latest loaded MRTK instance' option in the scene system one of these will probably be necessary.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
davidkline-ms: @SimonDarksideJ, that Unity blog post was from five years ago. I concur with @Railboy that we should not over index on something that was asked and no related product change for that time span.

davidkline-ms:
> The MixedRealityPlayspace script has been converted to a MonoBehaviour.

Question regarding this change. What requires the playspace to be a monobehavior? I'm certain there is a good reason behind it, i would just like to understand before approving this change.

Thx!

Railboy:
It makes searching for instances simpler, and it lets us display some in-editor documentation.

In the future I want to search for playspaces by script instead of by name. This is an intermediate non-breaking step in that direction.

SimonDarksideJ:
@Railboy a little concerned by your statement that you are making this a MonoBehaviour to be searchable, that is one of the primary reasons for introducing the Service Locator pattern, to avoid searching using monolithic features of Unity which are not performant.

Would it not be better to register the playspace in the runtime component registry and allow any component to quickly search for it there as a registered component?

Railboy:
@SimonDarksideJ The issue is that we can have several playspace transforms loaded at one time, and we only want one to be active, just like MRTK instances.

Searching by name / main camera parent for the transform is reliable when you're guaranteed to have only one transform with that name and / or only one main camera.

When you have many transforms with that name, or many main cameras, it's no longer reliable.

Making the playspace a Monobehavior hasn't altered its static methods. From the perspective of services consuming the playspace transform, nothing has changed. There just happens to be a script attached to the transform we were already creating.

davidkline-ms:
>Would it not be better to register the playspace in the runtime component registry and allow any component to quickly search for it there as a registered component?

That is an interesting thought @SimonDarksideJ. That would require making the playspace into a service to support registration. Not hard, but a significant change (primarilly with regards to accessing the object).

Railboy:
@SimonDarksideJ @davidkline-ms Just to keep our eye on the ball, would the runtime component registry approach also help us solve the problem of detecting and disabling / enabling multiple playspace transforms during scene loading / unloading?

davidkline-ms:
> would the runtime component registry approach also help us solve the problem of detecting and disabling / enabling multiple playspace transforms during scene loading / unloading?

That would need to be investigated. It _might_ if it were a named extension service.

To me, this is more of a thought experiment. Not something that should derail this PR.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -122,6 +122,43 @@ public override void Initialize()
                 CameraCache.Main.transform.position = Vector3.zero;
                 CameraCache.Main.transform.rotation = Quaternion.identity;
 
+                // Ensure there's an event system attached to the main camera.
+                bool addedComponents = false;
```
davidkline-ms: This feels like a major behavior change (it doesnt look like this code was moved from elsewhere in the input system).

It also does not feel like the responsibility of the input system to enforce this (read: what if the input system is not being used?).

davidkline-ms: Shouldnt this be the responsiblity of the camera system?

Assets/MixedRealityToolkit/Inspectors/MixedRealityPlayspaceInspector.cs
```diff
@@ -0,0 +1,42 @@
...
+using UnityEngine;
+using UnityEditor;
+
+namespace Microsoft.MixedReality.Toolkit
+{
+    [CustomEditor(typeof(MixedRealityPlayspace))]
+    public class MixedRealityPlayspaceInspector : UnityEditor.Editor
+    {
+        private static string description =
+          "This is where the Toolkit instantiates objects like cursors, pointers and boundary visualizations." +
```
davidkline-ms: This feels like non-deterministic behavior from an application stand point (read: an app may not always load scenes in the same order and may get different playspaces each time).

davidkline-ms: I realize we discussed this a bit yesterday in person. It just hit me this morning while re-reviewing.

Railboy: >This feels like non-deterministic behavior

The first playspace you load is treated as the active playspace. Once that's set, any others that are loaded get disabled. So you should be able to load scenes in any order and reliably predict which playspace will be active.

It starts getting weird if you unload the scene with the active playspace. Our workflow recommendations advise against this (keep your manager scene loaded etc) but if you go against that grain, the next active playspace selection will be non-deterministic. We could potentially tie playspace selection to scene build index or something like that, but it's not something I've tried.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
Railboy: The camera system doesn't know or care about input, so it felt odd for the camera system to handle the business of managing input-related components. Imagine a dev writing a custom implementation of the camera service - it wouldn't naturally occur to them to include this step when creating their main camera.

>It also does not feel like the responsibility of the input system to enforce this (read: what if the input system is not being used?).

This is true, if the input system isn't being used then the event system is not added. That seemed like a good thing to me, since if I've shut off the input system I assume I'd be doing it to have total control over what input-related components are being added to the scene. Are there are side effects I'm not considering here? Ie does anything wonky happen if the input system is disabled and there's no event system added to the scene?

davidkline-ms: > if the input system isn't being used then the event system is not added

In the current implementation, which component is adding the event system to the camera?

Railboy: The MRTK instance adds it under `EnsureMixedRealityRequirements`.

luis-valverde-ms:
@Railboy and @davidkline-ms , just checking if this is still active. Are we waiting on further review or pending changes?

SimonDarksideJ:
> @SimonDarksideJ @davidkline-ms Just to keep our eye on the ball, would the runtime component registry approach also help us solve the problem of detecting and disabling / enabling multiple playspace transforms during scene loading / unloading?

Yes, that is the purpose of the runtime component locator, to allow scene items to be registered and found anywhere, even by name.

Railboy:
@luis-valverde-ms Nobody has made any specific requests.

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
```diff
@@ -1,49 +1,84 @@
...
+        /// Returns a cached reference to the main camera. Uses the camera system's main camera, if available. Otherwise finds or creates a main camera.
         /// </summary>
         public static Camera Main
         {
             get
             {
-                if (cachedCamera != null)
+                // If the camera system is enabled, use the main camera.
+                // This allows developers to control exactly which camera will be returned by the cache.
+                if (MixedRealityToolkit.IsCameraSystemEnabled)
```
davidkline-ms: We cannot have direct references to the MixedRealityToolkit class in our systems. They need to use the MixedRealityServiceRegistry

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
```diff
@@ -1,49 +1,84 @@
...
-                    Debug.LogWarning("No main camera found. The Mixed Reality Toolkit requires at least one camera in the scene. One will be generated now.");
-                    mainCamera = new GameObject("Main Camera", typeof(Camera)) { tag = "MainCamera" }.GetComponent<Camera>();
+        /// <summary>
+        /// Returns true if cached camera exists and is enabled.
+        /// </summary>
+        public static bool MainExists
+        {
+            get
+            {
+                if (MixedRealityToolkit.IsCameraSystemEnabled)
```
davidkline-ms: We cannot have direct references to the MixedRealityToolkit class in our systems. They need to use the MixedRealityServiceRegistry

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
Railboy: Good catch, will change.

davidkline-ms:
I believe we are close on this one

Assets/MixedRealityToolkit/Inspectors/MixedRealityPlayspaceInspector.cs
```diff
@@ -0,0 +1,42 @@
...
+          "\n- Your main camera should be kept here (unless you are using a custom CameraSystem implementation.)" +
+          "\n- This transform can be moved, but it should not be parented under another object." +
```
fast-slow-still: >but it should not be parented under another object [](start = 44, length = 50)

Why this advice that the playspace shouldn't be parented under another object? Does that actually break anything? An additional transform in the scene graph is an important capability for the app, since they can't directly manipulate the camera's transform (it will be overwritten), nor the playspace transform (it's already in use for teleport etc.).

Railboy: I'm probably being over-cautious. I'll remove that part of the warning message. I've tested parented playspace transforms and they do work.

davidkline-ms:
I'll pull these changes and test them out today.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -14,21 +15,32 @@
 namespace Microsoft.MixedReality.Toolkit
 {
     /// <summary>
-    /// A static class encapsulating the Mixed Reality playspace.
+    /// A class encapsulating the Mixed Reality playspace.
     /// </summary>
-    public static class MixedRealityPlayspace
+    [ExecuteAlways]
+    public partial class MixedRealityPlayspace : MonoBehaviour
```
keveleigh: What's the benefit of turning this into a `MonoBehaviour` with a scene requirement? It's been working well so far without that requirement.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -14,21 +15,32 @@
...
+        private static IMixedRealityCameraSystem cameraService;
 
         public static void Destroy()
         {
-            // Playspace makes main camera dependent on it (see Transform initialization),
-            // so here it needs to restore camera's initial position. 
-            // Without second parameter camera will not move to its original position.
-            CameraCache.Main.transform.SetParent(null, false);
-            UnityEngine.Object.Destroy(mixedRealityPlayspace.gameObject);
+            // This will destroy any main camera parented under this playspace.
```
keveleigh: This seems risky. What if I've customized my camera with additional, non-MRTK scripts that the camera service doesn't know about?

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -254,42 +358,63 @@ private static void SearchForAndDisableExtraPlayspaces(IEnumerable<GameObject> r
...
-                if (rootGameObject == mixedRealityPlayspace.gameObject)
-                {   // Don't disable our existing playspace
-                    continue;
+                if (rootGameObject.name.Equals(NameEnabled) || rootGameObject.name.Equals(NameDisabled))
+                {
+                    MixedRealityPlayspace playspace = rootGameObject.EnsureComponent<MixedRealityPlayspace>();
                 }
 
-                if (rootGameObject.name.Equals(Name))
+                foreach (MixedRealityPlayspace playspace in rootGameObject.GetComponentsInChildren<MixedRealityPlayspace>(true))
```
keveleigh: Instead of searching for this script, can't we continue making the current assumption about camera parents? Or naming? Both of which seem to still be enforced by this script.

Railboy: That's the point of this PR. Searching by name or parent is no longer reliable when you're dealing with multiple scenes and multiple main cameras.

keveleigh: But you're still using the main camera to [ensure](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4455/files#diff-a2a2bab0eb44c8a46d3c07c490881396R231) or [set](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4455/files#diff-a2a2bab0eb44c8a46d3c07c490881396R214) the playspace.

Assets/MixedRealityToolkit/Utilities/CameraCache.cs
```diff
@@ -1,49 +1,86 @@
...
+            if (mainCamera == null || !mainCamera.isActiveAndEnabled)
+            {   // If no main camera was found, create it now
+                Debug.LogWarning("No main camera found. The Mixed Reality Toolkit requires at least one camera in the scene. One will be generated now.");
+                mainCamera = new GameObject("Main Camera", typeof(Camera)) { tag = "MainCamera" }.GetComponent<Camera>();
+            }
+
+            // Cache the main camera
+            cachedCamera = mainCamera;
+            // Parent the camera under our play space
+            cachedCamera.transform.SetParent(MixedRealityPlayspace.Transform);
```
keveleigh: Won't this already be set, if needed, in the `Transform` getter? https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4455/files#diff-a2a2bab0eb44c8a46d3c07c490881396R214

Railboy: It only parents the main camera if no camera system is found. The camera cache doesn't assume that a main camera delivered by the service belongs under the playspace transform.

In our default implementation we do parent the camera under the playspace, but custom implementations may do something different.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
Railboy: Since there could be other playspaces loaded when you destroy the active playspace, you don't want to un-parent the camera and leave it hanging. Because another playspace with another main camera may get activated, and then you've got two main cameras to deal with.

It's sort of moot because this destroy method never gets used by any of our services or other scripts. I can revert it if it moves this PR along.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
Railboy: It works fine when you're only dealing with one scene and one playspace transform. But once you've got multiple scenes with multiple playspace transforms, a MonoBehaviour makes searching for / switching between active playspace transforms a lot simpler. Plus you can display in-editor documentation with a custom inspector so people understand what's going on.

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
Railboy: You're right, it shouldn't do that unless the camera system is disabled. I'm pushing a change with cleaner logic.

Railboy:
After discussing with @wiwei @keveleigh and @davidkline-ms I've decided to close this for now.

We can re-open later once community feedback makes the need for these changes more evident.

Some changes may be broken out into separate PRs and dealt with individually.

Issue #4167, opened 2019-05-02
Slate prefab box collider performance issue


carmines:
The box collider on the slate prefab is using a sphere, where the grabbable visual is a box 
![image](https://user-images.githubusercontent.com/36190349/57110908-11047880-6cef-11e9-9ffb-ac783214d2d9.png)

![image](https://user-images.githubusercontent.com/36190349/57110934-34c7be80-6cef-11e9-8fa1-748f3caa4982.png)

julenka:
@cre8ivepark , I believe you worked on the slate prefab, could you please take a look?

wiwei:
I've got a fix for this.

wiwei:
A couple of notes for future folks, especially people who don't know that much about unity physics debugging guides.

Note that the screenshot above was taken in play mode with the physics analysis tools enabled - that is, you'll want to:

1) Go to Window->Analysis->Physics Debugger
2) Check this box

![image](https://user-images.githubusercontent.com/5840182/57956092-a76a9980-78ac-11e9-84e8-76a7ec64a17e.png)

3) Add the slate prefab to an empty scene
4) Enter play mode, and then press spacebar to use input simulation to move your hand ray pointer over the slate - you'll see the colliders show up.

julenka:
Fixed by #4669

PR #4461, opened 2019-05-17
Slate prefab box collider performance issue

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4167

Slates use the bounding box's rotate colliders in order to handle rotation - note that even though the slate provides its own box-like prefab, the bounding box still creates a sphere collider (which has wayyyyy more vertices than the simple box), even though a basic box would suffice. This change addresses the issue by adding an option to either use a box or sphere collider when a custom prefab is provided for bounding box rotate handles.

This is done in this way to avoid being a breaking behavior change - other folks can choose to use a more performant collider, and the slate is the only thing updated to use this behavior for now.

For the prefab, all I did was update the dropdown select button to choose "box" - everything else is auto-generated I think from new types/fields that have been introduced since the last serialization.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
@@ -95,6 +95,16 @@ namespace Microsoft.MixedReality.Toolkit.UI
             ActivateByProximityAndPointer,
             ActivateManually
         }
+
+        /// <summary>
+        /// This enum defines the type of collider in use when a rotation handle prefab is provided.
+        /// </summary>
+        public enum RotationHandlePrefabCollider
+        {
+            Sphere,
+            Box
+        }
+
         #endregion Enums
 
         #region Serialized Fields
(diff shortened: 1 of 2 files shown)


wiwei:
image

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -207,6 +217,25 @@ public float ScaleMinimum
...
+            get
+            {
+                return rotationHandlePrefabColliderType;
+            }
+            set
+            {
+                if (rotationHandlePrefabColliderType != value)
+                {
+                    rotationHandlePrefabColliderType = value;
+                    CreateRig();
```
julenka: In a separate pr #4443  @djohnsomsft mentioed that CreateRig() has unintended consequences, at least in BoundingBox. For example, CreateRig() updates the min and max sizes in bounding box based on the object's current size.

wiwei: I'll hold off here until the dialog on that PR is resolved - I'd rather be consistent with how we're doing everything and I was using that PR as a reference point for consistency.

julenka: Hey will, this is done now! Should be good to check in.

PR #4464, opened 2019-05-17
Fix spatial coordinate scene and remove some exceptions

chrisfromwork:
1) The state synchronization demo scene was missing some serialized fields.
2) After completing detection, you can could get into a state where the marker spatial coordinate service states its no longer tracking. This throws an exception if a spatial coordinate is reported at the same time.

### DIFF
--- a/Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/SpatialCoordinateServiceBase.cs
+++ b/Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/SpatialCoordinateServiceBase.cs
@@ -161,7 +161,7 @@ namespace Microsoft.MixedReality.Experimental.SpatialAlignment.Common
             {
                 UnityEngine.Debug.LogWarning($"Exception thrown when trying to discover coordinate: {e.ToString()}");
                 isTracking = false;
-                return false;
+                throw e;
             }
         }
 
--- a/Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/SpatialCoordinateServiceBase.cs
+++ b/Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/SpatialCoordinateServiceBase.cs
@@ -84,7 +84,8 @@ namespace Microsoft.MixedReality.Experimental.SpatialAlignment.Common
             {
                 if (!isTracking)
                 {
-                    throw new InvalidOperationException("We aren't tracking, and shouldn't expect additional coordinates to be discovered.");
+                    UnityEngine.Debug.LogWarning("New coordinate discovered, but spatial coordinate service was not tracking");
+                    return;
                 }
             }
 
@@ -94,7 +95,7 @@ namespace Microsoft.MixedReality.Experimental.SpatialAlignment.Common
             }
             else
             {
-                throw new InvalidOperationException($"Coordinate with id '{id}' already discovered.");
+                UnityEngine.Debug.LogWarning($"Unexpected behavior, coordinate {id} was rediscovered.");
             }
         }
 
@@ -156,9 +157,11 @@ namespace Microsoft.MixedReality.Experimental.SpatialAlignment.Common
 
                 return true;
             }
-            finally
+            catch (Exception e)
             {
+                UnityEngine.Debug.LogWarning($"Exception thrown when trying to discover coordinate: {e.ToString()}");
                 isTracking = false;
+                return false;
             }
         }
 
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/SpatialCoordinateServiceBase.cs
```diff
@@ -156,9 +157,11 @@ public async Task<bool> TryDiscoverCoordinatesAsync(CancellationToken cancellati
 
                 return true;
             }
-            finally
+            catch (Exception e)
             {
```
andreiborodin: We shouldn't swallow exceptions here, that is a very bad practice. We can log, and then rethrow by doing "throw;"

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/SpatialCoordinateServiceBase.cs
```diff
@@ -84,7 +84,8 @@ protected void OnNewCoordinate(TKey id, ISpatialCoordinate spatialCoordinate)
...
                 if (!isTracking)
                 {
-                    throw new InvalidOperationException("We aren't tracking, and shouldn't expect additional coordinates to be discovered.");
+                    UnityEngine.Debug.LogWarning("New coordinate discovered, but spatial coordinate service was not tracking");
```
andreiborodin: Yeah, this is interesting. Maybe we can do a silent return, or should we actually process since these were as part of previous detection.

I was to strict missing a clear race condition here, strangely I didn't hit it when testing.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerDetectorCoordinateService.cs
```diff
@@ -160,7 +160,10 @@ protected override async Task OnDiscoverCoordinatesAsync(CancellationToken cance
                         await Task.Delay(TimeSpan.FromSeconds(1), cancellationToken).IgnoreCancellation();
                     }
 
-                    DebugLog("We have been told to stop");
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        DebugLog("We have been told to stop");
+                    }
```
andreiborodin: Yeah, good catch.

andreiborodin:
>                 }

Here is the interesting thing, when markers aren't seen, this is set to null. So, I think there should be a fix in this class like this:
- two cached fields: lastMarkerRotation, lastMarkerPosition
- Inside Marker.set if marker is not null (doesn't matter if it equals current value or not), update the lastMarkerRotation, lastMarkerPosition
- Create two private properties:
          private Vector3 Position => marker?.Position ?? lastMarkerPosition;
          private Quaternion Rotation => marker?.Rotation ?? lastMarkerRotation;
- Update the conversion methods below to use these properties in-place of marker.Position and marker.Rotation

---
Refers to: Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerDetectorCoordinateService.cs:36 in 73eb1e5. [](commit_id = 73eb1e5020635f28eba5d9da1e251a3baa4baa40, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/MarkerDetectorCoordinateService.cs
chrisfromwork: FWIW, I'll probably clear out a lot of this logging at some point. But for now this was kind of confusing.

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/SpatialCoordinateServiceBase.cs
chrisfromwork: sounds good.

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/SpatialCoordinateServiceBase.cs
chrisfromwork: I think for now ill log this warning and return. It may be that we need to process this in the future, but right now it seems like processing old logic isn't necessary.

PR #4466, opened 2019-05-17
Add IMixedRealityDataProviderAccess and update how data providers are retrieved

davidkline-ms:
## Overview
This change introduces IMixedRealityDataProviderAccess, which allows system implementations a standard contract through which they can allow components access to data provider instances.

As an example, the solver handler now accesses the hand joint service using the following pattern

``
handJointService = (InputSystem as IMixedRealityDataProviderAccess)?.GetDataProvider<IMixedRealityHandJointService>()
``

This change also adds IMixedRealityMouseDeviceManager to allow for easy recognition of mouse data providers.

To ensure consistency across services with data providers, the spatial awareness system's GetObserver* methods have been marked as obsolete, with a message to check for implementation of IMixedRealityDataProviderAccess and to call GetDataProvider*.

Note: The existing GetObserver* implementations now call the GetDataProvider* equivalents.

This change is part of the #3545 series and gets non-editor code (with a handful of small exceptions) entirely on to the new, flexible pattern.

## Testing performed
- Run edit mode test automation
- UWP build target, Play mode, temporarilly add mouse data provider and move it in 3D space.
- Added temporary code to request spatial observer, verified the obsolete attribute is displayed in Visual Studio


Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
```diff
@@ -44,7 +44,7 @@ public interface IMixedRealityInputSystem : IMixedRealityEventSystem
 
         /// <summary>
         /// The current Focus Provider that's been implemented by this Input System.
-        /// </summary>
+        /// </summary>
```
keveleigh:
```suggestion
        /// </summary>
```

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityMouseDeviceManager.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Microsoft.MixedReality.Toolkit.Input;
```
keveleigh: Missing header

Assets/MixedRealityToolkit/Interfaces/Services/IMixedRealityDataProviderAccess.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System.Collections.Generic;
```
keveleigh: Missing header

davidkline-ms: added

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityMouseDeviceManager.cs
davidkline-ms: added

Assets/MixedRealityToolkit/Interfaces/Services/IMixedRealityDataProviderAccess.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.MixedReality.Toolkit
+{
+    public interface IMixedRealityDataProviderAccess
```
wiwei: > public interface IMixedRealityDataProviderAccess [](start = 3, length = 49)

Nit: Add a doccomment for this interface.

Assets/MixedRealityToolkit/Interfaces/Services/IMixedRealityDataProviderAccess.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        /// <summary>
+        /// Gets the collection of registered data providers.
+        /// </summary>
+        /// <returns>
+        /// Read only copy of the list of registered observers.
```
wiwei: >observers [](start = 53, length = 9)

Should this be "observer" or data provider? I know that the word observer has a different meaning in the spatial awareness system world, and I haven't seen this referenced in the other input parts before.

keveleigh:
```suggestion
        /// Read only copy of the list of registered data providers.
```

wiwei: Same comment for the function down below as well.

---
In reply to: [285701349](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4466#discussion_r285701349) [](ancestors = 285701349)

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityEyeGazeDataProvider.cs
```diff
@@ -40,6 +40,8 @@ public class WindowsMixedRealityEyeGazeDataProvider : BaseInputDeviceManager, IM
 
         public bool SmoothEyeTracking { get; set; } = false;
 
+        public IMixedRealityEyeSaccadeProvider SaccadeProvider => this as IMixedRealityEyeSaccadeProvider;
```
wiwei: > as IMixedRealityEyeSaccadeProvider [](start = 70, length = 35)

I think this suffix isn't necessary right (unless you're explicitly writing it out for readability or something?)

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityToolkitConfigurationProfile.asset
```diff
@@ -47,3 +47,4 @@ MonoBehaviour:
       Microsoft.MixedReality.Toolkit.Services.DiagnosticsSystem
   registeredServiceProvidersProfile: {fileID: 11400000, guid: efbaf6ea540c69f4fb75415a5d145a53,
     type: 2}
+  useServiceInspectors: 1
```
wiwei: >useServiceInspectors: 1 [](start = 2, length = 23)

Intentional change here? Feels like if we do this we should do it in its own change (more easily solo-revertable, not piggybacking on another change etc)

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -272,6 +272,66 @@ public override void Disable()
...
+        /// <inheritdoc />
+        public IReadOnlyList<IMixedRealityDataProvider> GetDataProviders()
+        {
+            return new List<IMixedRealityInputDeviceManager>(deviceManagers) as IReadOnlyList<IMixedRealityInputDeviceManager>;
+        }
+
+        /// <inheritdoc />
+        public IReadOnlyList<T> GetDataProviders<T>() where T : IMixedRealityDataProvider
+        {
+            if (!(typeof(T) is IMixedRealityInputDeviceManager)) { return null; }
```
wiwei: >if (!(typeof(T) is IMixedRealityInputDeviceManager)) { return null; } [](start = 12, length = 69)

I think that here and below we want to be checking for subtypes as well - i.e. it should be something that is an implementor of device manager. So for example, I would expect to be able to ask for IMixedRealityEyeGazeDataProvider, but I can't with this code.

I tried this out locally (i.e. just having a cube that has the PanZoomBase.cs code) and without this subtype check it will basically end up returning null all the time.

wiwei: Same comment for the GetDataProvider overload down below.

---
In reply to: [285707829](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4466#discussion_r285707829) [](ancestors = 285707829)

Assets/MixedRealityToolkit/Interfaces/Services/IMixedRealityDataProviderAccess.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// </summary>
+        /// <typeparam name="T">The desired data provider type.</typeparam>
+        /// <param name="name">The friendly name of the data provider.</param>w
+        /// <returns>
+        /// The requested data provider, or null if one cannot be found.
+        /// </returns>
+        /// <remarks>
+        /// If more than one data provider is registered under the specified name, the first will be returned.
+        /// </remarks>
+        T GetDataProvider<T>(string name = null) where T : IMixedRealityDataProvider;
```
wiwei: I don't fully understand why one of these supports taking a null name, but the other doesn't - what's the rationale behind this non-consistent API story?

wiwei:
@davidkline-ms @keveleigh change generally seems okay, though there are a few questions I had before I hit the approve button.

Assets/MixedRealityToolkit/Interfaces/Services/IMixedRealityDataProviderAccess.cs
keveleigh: Are you referring to `IMixedRealityDataProvider GetDataProvider(string name);` vs `T GetDataProvider<T>(string name = null)`?

The reason, as I understand it, is that both calls need something to actually let you define what you're looking for. The first call lets you ask by name, while the second one lets you ask by type and an optional name.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
keveleigh: There were a few things going on here:
1. The in-editor simulation wasn't properly registering itself as an `IMixedRealityEyeGazeDataProvider`
2. You're right that this check wasn't sufficient to catch those cases anyway

I've updated both.

davidkline-ms:
replaced by #4593

Issue #4452, opened 2019-05-17
Configuration profile inspector doesn't constrain profile types for camera and diagnostics systems.


Railboy:
## Describe the bug

When selecting profiles for the camera system and diagnostics system the object picker doesn't constrain available profile types.

## To reproduce

- Select a MixedRealityToolkit instance
- Try to select a CameraSystem profile
- Observe that all scriptable objects are available

## Expected behavior

Profiles should be constrained to service type.

## Your Setup

- MRTK Version 2.0 mrtk_development

Troy-Ferrell:
Harden MRTK inspectors & increase re-usability of code #4467

Issue #4449, opened 2019-05-17
Clean up MRTK inspector dependencies


provencher:
## Describe the bug

Prior to merging the new UI, it was possible to edit MRTK profiles without an MRTK instance in the scene.

This is no longer the case since https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4314 was merged yesterday
![image](https://user-images.githubusercontent.com/7420990/57940844-5a31fc00-789b-11e9-91c5-16c9c25ba526.png)

## To reproduce

Try to edit and MRTK profile without an active instance in the scene.

## Expected behavior

It should be possible to edit MRTK profiles without having an instance in the scene.

## Your Setup (please complete the following information)

- Unity Version 2018.3f1
- MRTK Version v2 latest dev

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

Troy-Ferrell:
Also need to revert other edge cases to allow editing (i.e no inputsystem enabled, etc)

Troy-Ferrell:
To summarize discussion extended in the PR referenced above, the MRTK inspectors need to be updated to only block rendering/editing when that particular profile requires an MRTK instance in scene. Or conversely to allow a design such that ALL profile inspectors do not require a MRTK instance at all. 

Before the PR, some profiles were blocked under certain conditions or although would render, they would not actually drive any editing value as they were not hooked up to their serializedobjects in OnEnable(). The PR made this non-obvious disability 1) universal and 2) more visible to the user.

Railboy:
Some rules of thumb I think we should consider when reworking these profiles:

- First ask ourselves why this profile needs an instance to display its data and whether we can avoid this requirement altogether.
- When an instance really is required for a profile to function, don't just exit early if no instance is found. Display as much of the data as possible.
- Replace sections that can't be displayed with a standardized helpbox warning explaining that this section of the profile requires an instance, and offer a button to create an instance on the spot.

Railboy:
@Troy-Ferrell Classic cross-posting, haha. Clearly we're on the same page.

Troy-Ferrell:
Indeed, also point to mention for record, when the inspectors require a MRTK instance in scene to function properly, the instance many times MUST be it's own. Example: When editing the speech commands profile, this looks at what input action mappings are available....but it needs to look at the input actions for the instance in which it is the assigned speech profile.

Troy-Ferrell:
[...]

PR #4435, opened 2019-05-16
Move MRTK addition to scene to utility class

Troy-Ferrell:
## Overview
Fixes debug assert with multiple MRTK instances and simplifies code to one utility function for adding MRTK to the current scene

## Changes
- Fixes: #4426
Assert failing in MixedRealityToolkitInspector.CreateMixedRealityToolkitGameObject() 

## Verification
Add to scene & configure
built project
Duplicate MRTK with multiple instances

PR #4467, opened 2019-05-17
Harden MRTK inspectors & increase re-usability of code

Troy-Ferrell:
## Overview

This change is another pass on the recent MRTK inspector changes for profiles. Although profile inspectors would render their fields when there was no MRTK instance in scene, they would not populate correctly and any edits wouldn't be driven into the serialized object. This change ensures that the user is notified when certain fields are not editable because of some dependency (whether a MRTK instance is needed, or enabled input system, etc).

Further, turns out the back button for profile inspectors was broken. The button would only go back up the profile tree for the ACTIVE instance in scene, not necessarily the parent of the current profile which actually could be multiple parent profiles referencing that child. The new approach checks that the current profile being viewed is active and assigned in the current scene/MRTK instance and thus it can be known what parent to select & view in inspector. If it is not assigned in scene, then the back button is hidden.

Fixed object fields not having correct filter types based on the type of profile expected or based on a service type/interface. 

Also fixed issue where private local fields would be re-initialized every refresh because the inspector was being drawn as a sub-profile and not it's own window. Thus, it would be impossible to write any values as they were cleared every refresh

Reorganized utility functions for re-use. Deleted unnecessary code. 

Simplified utils
-> Render MRTK logo
-> Render Indented button
-> Wrapper class to simplify enable/disabling sections of GUI rendering

Fixed spatial awareness mesh observer with correct return profile target

## Changes
- Fixes: 
#4452 
Configuration profile inspector doesn't constrain profile types for camera and diagnostics systems.
#4449 
Clean up MRTK inspector dependencies

## Verification
Tried adding new MRTK instance
Tried selecting mrtk profile that is not instanced
Selected sub-profiles assigned and not assigned in instance
delete objectfield in config profile and ensure correct type is specified
duplicated MRTK gameobject in scene
Built unity example scenes into VS solution


wiwei:
>         /// <param name="property">the <see cref="Microsoft.MixedReality.Toolkit.BaseMixedRealityProfile"/> property.</param>

Nit: Update the doccomments to include the new type (for both RenderProfile and RenderProfileInternal)

---
Refers to: Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs:73 in 07bb716. [](commit_id = 07bb7166a9a25bbfde50842f8bc679001294eda8, deletion_comment = False)

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
```diff
@@ -90,12 +88,18 @@ protected static bool RenderProfile(SerializedProperty property, bool showAddBut
...
+            bool showAddButton, bool renderProfileInBox, Type serviceType = null)
         {
             var profile = property.serializedObject.targetObject as BaseMixedRealityProfile;
             bool changed = false;
             var oldObject = property.objectReferenceValue;
 
+            if (profileType != null && !profileType.IsSubclassOf(typeof(BaseMixedRealityProfile)) && profileType != typeof(BaseMixedRealityProfile))
+            {
+                profileType = null;
+            }
```
wiwei: Is the scenario in which this gets passed an incorrect profile type something where the user dragged a non-profile asset onto the profile field on the inspector?

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityGesturesProfileInspector.cs
```diff
@@ -53,18 +56,27 @@ protected override void OnEnable()
             thisProfile = target as MixedRealityGesturesProfile;
             Debug.Assert(thisProfile != null);
 
-            if (MixedRealityToolkit.Instance.ActiveProfile.IsInputSystemEnabled &&
-                MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.InputActionsProfile != null)
+            UpdateGestureLabels();
+
+            if (!MixedRealityToolkit.IsInitialized ||
+                !MixedRealityToolkit.Instance.ActiveProfile.IsInputSystemEnabled ||
+                MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.InputActionsProfile != null ||
```
wiwei: >! [](start = 98, length = 1)

I think this should be == (because it's supposed to be the opposite of what it used to be.

I was noticing a ton of exceptions in the unity editor when scrolling up and down the input profile I think due to this.

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityInputActionRulesInspector.cs
```diff
@@ -76,89 +74,101 @@ protected override void OnEnable()
...
-
             thisProfile = target as MixedRealityInputActionRulesProfile;
 
-            ResetCriteria();
-        }
-
-        public override void OnInspectorGUI()
-        {
-            if (!RenderProfileHeader(ProfileTitle, ProfileDescription, BackProfileType.Input))
+            // Only reset if we have not yet
```
wiwei: >yet [](start = 41, length = 3)

Nit: I think incomplete comment.

wiwei:
I'm also seeing this weird bug with this change where pressing "Copy & Customize" will, in fact, copy and customize the profile, but won't switch the inspector to the newly cloned profile.

i.e. I press the button, I can see in the assets that it's actually creating a cloned profile, but the inspector is still on the default one (so as a user i think it's done nothing and I keep pressing the button).

If I then go and click the main MRTK object I see that it's actually updated things.

Packages/manifest.json
```diff
@@ -3,10 +3,10 @@
     "com.unity.ads": "2.3.1",
     "com.unity.analytics": "3.2.2",
     "com.unity.collab-proxy": "1.2.15",
-    "com.unity.package-manager-ui": "2.0.3",
+    "com.unity.package-manager-ui": "2.0.7",
```
wiwei: Probably wanna leave the manifest.json and ProjectVersion.txt changes out of this one (I assume these should generally be updated in their own smaller change, unless the other changes in this PR require this update)

wiwei:
@Troy-Ferrell generally looks good, lemme know when you want me to take another look.

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityToolkitConfigurationProfileInspector.cs
```diff
@@ -130,73 +117,63 @@ protected bool DrawBacktrackProfileButton(string message, UnityEngine.Object act
...
+                if (MixedRealityToolkit.IsInitialized)
+                {
+                    if (IsProfileInActiveInstance())
+                    {
+                        DrawBacktrackProfileButton(returnProfileTarget);
+                    }
+                    else if (!isProfileInitialized)
+                    {
+                        EditorGUILayout.HelpBox("This profile is not assigned to the active MRTK instance in this scene. Some properties may not be editable", MessageType.Error);
+                    }
```
Railboy: I assume we don't include a button to add the instance here because we're waiting on #4435 to be merged.

Troy-Ferrell:
Closing this PR since completed by @Railboy (Thanks!) 
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4504

provencher:
@Troy-Ferrell does your PR make it so that you can modify a profile without having any MRTK scene instance?

Troy-Ferrell:
@provencher & @luis-valverde-ms, yes you can now edit MRTK profiles without an instance in the scene. For some reason though in the #4504  PR, there were certain files modified to not render unless an MRTK instance was in the scene. It was only a hand full of files (PointerClickHandlerInspector, SpeechInputHandlerInspector, BaseInputHandlerInspector, MixedRealityInputActionRulesInspector, and MixedRealityGesturesProfileInspector). I don't know why these were reverted to use MixedRealityInspectorUtility.CheckMixedRealityConfigured which enforces MRTK instance in the scene. @Railboy, was this on purpose?

Troy-Ferrell:
Ah the handler input inspectors rely on MRTK instance data (input actions list or speech keywords, etc). Short term we can fix it such that it is more obvious these settings can't be edited without a valid MRTK active instance. Longer term we may want to think on this architecture dependency

PR #4468, opened 2019-05-17
update minimum unity version from 2018.3 to 2018.4

davidkline-ms:
Updates the project version and the minimum Unity version check to 2018.4.


Assets/MixedRealityToolkit/Utilities/Editor/EditorProjectUtilities.cs
```diff
@@ -37,7 +39,7 @@ private static void DisplayIncorrectEditorVersionDialog()
         {
             if (!EditorUtility.DisplayDialog(
                 "Mixed Reality Toolkit",
-                "The Mixed Reality Toolkit requires Unity 2018.3 or newer.\n\nUsing an older version of Unity may result in compile errors or incorrect behavior.",
+                $"The Mixed Reality Toolkit requires Unity {MinUnityVersion} or newer (2018.3 will likley continue to work).\n\nUsing an older version of Unity may result in compile errors or incorrect behavior.",
```
keveleigh:
```suggestion
                $"The Mixed Reality Toolkit requires Unity {MinUnityVersion} or newer (2018.3 will likely continue to work).\n\nUsing an older version of Unity may result in compile errors or incorrect behavior.",
```

Assets/MixedRealityToolkit/Utilities/Editor/EditorProjectUtilities.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
     [InitializeOnLoad]
     public class EditorProjectUtilities
     {
+        private static string MinUnityVersion = "2018.4";
```
keveleigh:
```suggestion
        private static readonly string MinUnityVersion = "2018.4";
```

davidkline-ms:
@keveleigh, something happened with the CI... can you investigate while i am out?  thanks!

davidkline-ms:
this is likely failing due to the ci system not having 2018.4 installed... @MenelvagorMilsom, can you check to see if this theory is correct?

MenelvagorMilsom:
@davidkline-ms unfortunately the errors aren't giving me a huge amount of info to work out what caused the failure, but the build machines do not have 2018.4 installed. I'll let the right people know and see what we can do.

MenelvagorMilsom:
I spoke with Ron yesterday and sent a mail this morning. Hopefully they can add the new version soon.

wiwei:
I think the CI machines have been updated with 2018.4, trying again.

thalbern:
we should upgrade the mrtk project and add that as well to the PR if we move this to stabilization

MenelvagorMilsom:
@davidkline-ms should we use this PR to try upgrading the project version without upgrading the version for CI? You just need to undo the changes made to the pipelines YAML definitions, no?

MenelvagorMilsom:
Some of these changes have been made in #4756. But we can't make the CI changes because we don't have the newer version on the build agents. We should take a second look at this PR once the build agents are updated.

wiwei:
This change has been around for a while - I think that we might still be waiting for our CI machines to get .4, but in the interim I'm closing this change out. It's already gotten a few merge conflicts and other things that would probably be handled cleaner just by doing a s/2018.3/2018.4/g

Issue #4276, opened 2019-05-10
Spatial mapping visualization doesn't work with non-identity playspace transform


chbecker-ms:
Easiest repro:

1) Enable MRTK spatial mapping visualization (on by default in default MRTK profile)
2) Add 1 meter offset on Y in MixedRealityPlayspace position
3) Deploy and run on HL

Note that all the spatial mapping wireframes are one meter low.

This is significant in any situation where it is desired to effectively apply a transform to the camera. This includes situations such as teleport, camera fly mode, and grabbing the entire contents of the world and manipulating them as a unit.

chbecker-ms:
Comment from David:

This is likely due to the parenting of the meshes... I will investigate.

chbecker-ms:
Ported from internal database

fast-slow-still:
I have had to address this in other work. @davidkline-ms  I would be happy to help save some time for you here, either working up a fix or pointing out failure point. Just let me know how I can best assist.

wiwei:
@fast-slow-still if you have cycles to look at this, this would be great.

I don't believe that @davidkline-ms is actively investigating this right now.

PR #4474, opened 2019-05-20
Fix spatial mapping to work with teleport and camera control.

fast-slow-still:
## Overview
Modification to allow spatial mapping to work with teleport and camera control via the MixedRealityPlayspace.

Rather than having a WorldAnchor directly control the spatial mesh's transform, read the WorldAnchor's transform and concatenate with the playspace transform to get the mesh's transform.

## Changes
- Fixes: #4276  .

>
>

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
@@ -416,20 +416,12 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.SpatialAwareness
             /// <summary>
             /// Compute concatenation of lhs * rhs such that lhs * (rhs * v) = Concat(lhs, rhs) * v
             /// </summary>
-            /// <remarks>
-            /// Start defining pose * vector = pose.p + pose.r * vector
-            /// lhs * (rhs * v)
-            /// = lhs * (rhs.p + rhs.r * v)
-            /// = lhs.p + lhs.r * (rhs.p + rhs.r * v) 
-            /// = lhs.p + lhs.r * rhs.p + lhs.r * rhs.r * v
-            /// = Pose(lhs.p + lhs.r * rhs.p, lhs.r * rhs.r) * v
-            /// </remarks>
             /// <param name="lhs">Second transform to apply</param>
             /// <param name="rhs">First transform to apply</param>
             /// <returns></returns>
             private static Pose Concatenate(Pose lhs, Pose rhs)
             {
-                return new Pose(lhs.position + lhs.rotation * rhs.position, lhs.rotation * rhs.rotation);
+                return rhs.GetTransformedBy(lhs);
             }
 
             /// <summary>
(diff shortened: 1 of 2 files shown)


julenka:
From the bug:

> Add 1 meter offset on Y in MixedRealityPlayspace position

I thought that you are not supposed to move the playspace, according to @simondarkside. Am I misunderstanding something?

SimonDarksideJ:
This is a long running debate @julenka 
* Immersive HAS to start at 0 as the height is determined by the framework
* AR requires the head position to be relative to the players head, manual.

@Railboy what are your thoughts, did we ever get the head setup thing sorted?

fast-slow-still:
As far as whether it’s okay to change the transform on the playspace, I believe that is the reason we have the playspace. Note that teleport is implemented by adjusting the transform of the playspace. Manual camera control is generally enabled by putting a ManualCameraControl component on the playspace. (Putting the camera control component on the camera would be ineffective since the camera’s transform is overwritten each frame to track the head).

I think the misconception here stems from thinking of a transform on the MRTK Playspace as moving the camera. In AR, that’s obviously not what we want. The camera’s transform is already keeping the camera aligned with the head, where it should be.

It might help to think of the MRTK Playspace transform as repositioning everything in the scene relative to the head. Moreover, it repositions the contents of the scene without changing their global scene coordinates, so it won’t break any physics simulation going on.

This is a very powerful and commonly used mechanism that seems important to support. 

@julenka , @SimonDarksideJ , It seems like there's an ongoing conversation I haven't seen. I'd be interested in understanding your thoughts better. Is there a link? Or could you summarize? I'm wondering why we wouldn't want to support this, especially since we clearly do need to support it in immersive.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
```diff
@@ -422,7 +452,18 @@ private void RequestMesh(SurfaceId surfaceId)
...
-                worldAnchor = newMesh.GameObject.AddComponent<WorldAnchor>();
+                // The WorldAnchor component places its object where the anchor is in the same space as the camera. 
+                // But since the camera is repositioned by the MixedRealityPlayspace's transform, the meshes' transforms
+                // should also the WorldAnchor position repositioned by the MixedRealityPlayspace's transform.
+                // So rather than put the WorldAnchor on the mesh's GameObject, the WorldAnchor is placed out of the way in the scene,
+                // and its transform is concatenated with the Playspace transform to compute the transform on the mesh's object.
+                // That adapting the WorldAnchor's transform into playspace is done by the internal PlayspaceAdapter component.
+                // The GameObject the WorldAnchor is placed on is unimportant, but it is convenient for cleanup to make it a child
+                // of the GameObject whose transform will track it.
+                GameObject anchorHolder = new GameObject(meshName + "_anchor");
```
StephenHodgson: Do we ever clean up this GameObject when the spatial object is recycled?

fast-slow-still: The anchorHolder is attached as a child node to the newMesh.GameObject. It is cleaned up as part of the newMesh.GameObject tree.

Issue #4006, opened 2019-04-17
Generic Input Action Handler


jbienzms:
## Describe the problem

On the surface, Input Actions _appear_ to be a centralized place for defining user intents. For example, the application may have actions for things like 'Fire', 'Save' and 'Add a Note'. 

Inside the MRTK configuration settings, actions can be triggered by any number of input sources. For example, the 'Fire' intent can be triggered by speaking the word "Fire" as well as by pressing the 'X' button on a controller. 

Conceptually, a developer would then assume that they could handle the Action (the intent) regardless of the source that triggered it. However, in the current design that is not the case.

In the current design, in order to handle the 'Fire' intent the developer must handle each source event type. In this example they would need to handle the Speech Recognized event and check the event payload for the action as well as handle the Controller Pressed event and check the event payload for the action. This not only requires each intent handler to implement many interfaces, but it means that new sources added via  configuration won't trigger until code is added to the handler.

## Describe the solution you'd like

We would like a generic way to handle an Action regardless of the source that triggered it. Since many of the current sources have "starting" and "ending" events, this should be reflected too. The generic action handler would have **Action_Started** and **Action_Ended**. These would map to source events like **Button_Pressed** and **Button_Released**. A strategy would need to be defined for sources that do not provide both events. For example, a voice command would obviously trigger the **Started** event, but what should be done about **Ended**? 

## Describe alternatives you've considered

The alternate, as described above, requires implementing each and every source event type interface.

luis-valverde-ms:
I agree on the convenience of having a mechanism to subscribe to input actions directly. We should look into it.

I don't fully get what you mean with:
> implementing each and every source event type interface
Following the voice command example above, if you had a digital input action mapped to both a voice command and to binary controller inputs I think you would have to implement just two interfaces: IMixedRealityInputHandler and IMixedRealitySpeechHandler. Do you have use case in mind were you would need to implement more?

I also don't get:
> new sources added via configuration won't trigger until code is added to the handler
If you have a global listener implementing IMixedRealityInputHandler for example, it will receive callbacks for all input sources raising those events. What did you mean there?

jbienzms:
> I don't fully get what you mean with: 
> - implementing each and every source event type interface

In the scenario above, at the end of the **Day 1** the application has two sources that can trigger the same action:

1. Voice Command
2. Controller Input

This required the application developer to implement two interfaces:

1. IMixedRealitySpeechHandler
[...]

PR #4475, opened 2019-05-20
Added input action handler interface to receive action events regardless of the input source

luis-valverde-ms:
## Overview

Added a new input action handler interface to listen for input action events regardless of input source: 

```
public interface IMixedRealityInputActionHandler : IMixedRealityBaseInputHandler
{
	void OnActionStarted(BaseInputEventData eventData);
	void OnActionEnded(BaseInputEventData eventData);
}
```
Also added an input action handler script for convenience:

InputActionHandler

And an input actions example:

InputActionsExample

## Changes
- Fixes: #4006  .


luis-valverde-ms:
@jbienzms , do you mind having a look at it and tell me what you think? I'll add the interfaces for actions that contain data soon but it should be easy enough to imagine how they will look like.

julenka:
Talking to @killerantz Interactable isn’t actually meant to be used as the main way to listen to events. As I understand, the main power of interactable is in configuring visual states of buttons in response to events, not to hook up functionality in response to events.

Glad to see there is a way people may use instead of interactable.

killerantz:
I think this is what I've been looking for all along. As long as I can get the source of the event, than this is what I was expecting from the Input Actions system. I wanted to build Interactable off of action events.

It would be great for Interaction Mappings to have a better link to the actions they invoke. or the eventData to include the InteractionMapping that invoked it as a property.

jbienzms:
> @jbienzms , do you mind having a look at it and tell me what you think? I'll add the interfaces for actions that contain data soon but it should be easy enough to imagine how they will look like.

Soo sorry this has taken me so long. I'm in Berlin at hacks this week and didn't have a chance to look. Reviewing now.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -944,6 +944,25 @@ public void RaisePointerUp(IMixedRealityPointer pointer, MixedRealityInputAction
...
+                Debug.Assert(inputData.MixedRealityInputAction != MixedRealityInputAction.None);
+
+                var inputHandler = handler as IMixedRealityInputHandler;
+                if (inputHandler != null)
+                {
+                    inputHandler.OnInputDown(inputData);
+                }
+
+                var actionHandler = handler as IMixedRealityInputActionHandler;
+                if (actionHandler != null)
```
jbienzms: Do we also need to check if inputData.MixedRealityInputAction isn't **null**? I know we're checking to make sure it's not InputAction.None, but is there any point at which it could be **null**?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -967,6 +993,25 @@ public void RaiseOnInputDown(IMixedRealityInputSource source, Handedness handedn
...
+                Debug.Assert(inputData.MixedRealityInputAction != MixedRealityInputAction.None);
+
+                var inputHandler = handler as IMixedRealityInputHandler;
+                if (inputHandler != null)
+                {
+                    inputHandler.OnInputUp(inputData);
+                }
+
+                var actionHandler = handler as IMixedRealityInputActionHandler;
+                if (actionHandler != null)
```
jbienzms: Same question: Do we need to check for InputAction to be **null**?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -1103,12 +1155,39 @@ public void RaisePoseInputChanged(IMixedRealityInputSource source, Handedness ha
...
+                Debug.Assert(inputData.MixedRealityInputAction != MixedRealityInputAction.None);
+
+                var gestureHandler = handler as IMixedRealityGestureHandler;
+                if (gestureHandler != null)
+                {
+                    gestureHandler.OnGestureStarted(inputData);
+                }
+
+                var actionHandler = handler as IMixedRealityInputActionHandler;
+                if (actionHandler != null)
```
jbienzms: Same question: Do we need to check for InputAction to be **null**?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -1193,12 +1272,39 @@ public void RaiseGestureUpdated(IMixedRealityController controller, MixedReality
...
+                Debug.Assert(inputData.MixedRealityInputAction != MixedRealityInputAction.None);
+
+                var gestureHandler = handler as IMixedRealityGestureHandler;
+                if (gestureHandler != null)
+                {
+                    gestureHandler.OnGestureCompleted(inputData);
+                }
+
+                var actionHandler = handler as IMixedRealityInputActionHandler;
+                if (actionHandler != null)
```
jbienzms: Same question: Do we need to check for InputAction to be **null**?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -1287,6 +1393,26 @@ public void RaiseGestureCanceled(IMixedRealityController controller, MixedRealit
...
+                Debug.Assert(speechData.MixedRealityInputAction != MixedRealityInputAction.None);
+
+                var speechHandler = handler as IMixedRealitySpeechHandler;
+                if (speechHandler != null)
+                {
+                    speechHandler.OnSpeechKeywordRecognized(speechData);
+                }
+
+                var actionHandler = handler as IMixedRealityInputActionHandler;
+                if (actionHandler != null)
```
jbienzms: Same question: Do we need to check for InputAction to be **null**?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -1287,6 +1393,26 @@ public void RaiseGestureCanceled(IMixedRealityController controller, MixedRealit
...
+                {
+                    actionHandler.OnActionStarted(speechData);
+                    actionHandler.OnActionEnded(speechData);
```
jbienzms: Excellent. Yes, I agree. Speech should do Start and then End successively.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
luis-valverde-ms: `MixedRealityInputAction` is a value type so we're safe there.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/InputActionHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+        [SerializeField]
+        [Tooltip("Input Action to handle")]
+        private MixedRealityInputAction InputAction = MixedRealityInputAction.None;
+
+        [SerializeField]
+        [Tooltip("Whether input events should be marked as used after handling so other handlers in the same game object ignore them")]
+        private bool MarkEventsAsUsed = false;
+
+        [System.Serializable]
+        public class InputActionUnityEvent : UnityEvent<BaseInputEventData> { }
```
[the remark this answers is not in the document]
luis-valverde-ms: Good idea. Changes coming.

luis-valverde-ms:
> In this PR it looks like this could be accomplished in 2 ways:
> 
> Look at BaseInputEventData.Source
> Try cast the BaseInputEventData to the source-specific event data.
> 
> Am I correct about both of those working? If so, I think we're set in this regard.

@jbienzms , you're totally correct.

luis-valverde-ms:
> It would be great for Interaction Mappings to have a better link to the actions they invoke. or the eventData to include the InteractionMapping that invoked it as a property.

@killerantz , agreed. Do you mind opening a feature request for that?

luis-valverde-ms:
Checking this in with support only for digital input actions. Created #4576 to add support for non-digital ones, as there's a bit more work involved than expected.

Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityGestureHandler.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// <summary>
     /// Interface to implement for generic gesture input.
     /// </summary>
-    public interface IMixedRealityGestureHandler : IEventSystemHandler
+    public interface IMixedRealityGestureHandler : IMixedRealityBaseInputHandler
```
StephenHodgson: Why was this added?

luis-valverde-ms: Have a look at the documentation of IMixedRealityBaseInputHandler and the implementation of OnGestureStartedWithAction() in MixedRealityInputSystem and let me know if that leaves you with any doubts.

Issue #4448, opened 2019-05-17
Articulated hand pose loading is broken is broken in editor if MRTK not in Assets root


provencher:
## Describe the bug

ArticulatedHandPose.cs is expects the hand poses to be stored in "Assets/MixedRealityToolkit.Services/InputSimulation/ArticulatedHandPoses/"

That said, I stored the MRTK in a sub folder so as to not clutter my project with all the MRTK folders at the root. Because of this, the articulated hands fail to load.

Moreover, `private static ArticulatedHandPose LoadGesturePose(GestureId gesture, string filePath)` does not consider the possibility that the file might not exist, and so an error is thrown while that could be avoided by checking if the file it's trying to open exists in the specified path.

## To reproduce

1. Move the MRTK root folder to a sub folder
2. Try and load a scene where the articulated hand joints should appear, like the hand interaction scene.
3. Articulated hand pose loading will throw an error, and hand joints will not appear in the editor.

## Expected behavior

Articulated hand poses should not have a hardcoded path to loading the json with the hand data.

Rather, StreamingAssets should be used, as it allows you to specify a path relative to the it, without preloading anything into memory. StreamingAssets can be arbitrarily nested in hierarchies, and you can have more than one such folder. It is the preferred method for dynamically loading in non-unity assets at runtime.

Here is the documentation on streaming assets
https://docs.unity3d.com/Manual/StreamingAssets.html

## Your Setup (please complete the following information)

- Unity Version 2018.3.81
- MRTK Version v2.0 latest dev as of May 17 2019

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

wiwei:
We should fix this soon, because we know that lots of customers import MRTK into the non-root folder

CoPrez:
When consuming the NuGetized version of MRTK I'm seeing this same issue. It looks like it stems from this line in [ArticulatedHandPose.cs](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4448).

PR #4476, opened 2019-05-20
Extend the Files utility to support non-core modules and loading.

lukastoenneMS:
## Overview

Articulated hand pose data loading from json files depends on finding paths in the InputSimulation folder. This was using a hardcoded relative path that can easily be broken by installing MRTK into a non-conventional directory.

The fix involves using the MixedRealityToolkitFiles utility class that registers the MRTK directory and provides convenient path construction. However, the utility was only registering the core MixedRealityToolkit path, so it has to be extended to track all the major modules in MRTK.

## Changes
- Fixes: #4448

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -10,6 +11,19 @@ using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
 {
+    /// <summary>
+    /// Base folder types for module searched by the MixedRealityToolkitFiles utilty.
+    /// </summary>
+    public enum MixedRealityToolkitModuleType
+    {
+        Core,
+        Providers,
+        Services,
+        SDK,
+        Examples,
+        Tests,
+    }
+
     /// <summary>
     /// API for working with MixedRealityToolkit folders contained in the project.
     /// </summary>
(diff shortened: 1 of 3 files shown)


lukastoenneMS:
It just occurred to me that using arbitrary mrtk paths (other than "Assets/MixedRealityToolkit" or "Assets/MRTK") will probably still not work. Not sure if/how the MRTFiles utility was handling that so far.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -29,42 +43,82 @@ public static void OnPostprocessAllAssets(string[] importedAssets, string[] dele
...
 
                 foreach (string asset in deletedAssets.Concat(movedFromAssetPaths))
                 {
                     string folder = asset.Replace("Assets", Application.dataPath);
-                    if (folder.EndsWith(MixedRealityToolkitDirectory))
+                    foreach (MixedRealityToolkitModuleType module in Enum.GetValues(typeof(MixedRealityToolkitModuleType)))
                     {
-                        folder = NormalizeSeparators(folder);
-                        if (mrtkFolders.Contains(folder) && !Directory.Exists(folder))
+                        if (mrtkFolders.TryGetValue(module, out HashSet<string> modFolders))
```
wiwei: >mrtkFolders [](start = 28, length = 11)

Since we're now storing a Dictionary of HashSets, it's possible for the underlying hashsets to end up being empty after this operation - in which case, it's probably right to also delete the entry from the dictionary. 

Otherwise you'll get some functions below returning incorrect information like:

public static bool AreFoldersAvailable
        {
            get
            {
                searchForFoldersTask.Wait();
                return mrtkFolders.Count > 0;
            }
        }

Which will check for the presence of data in the dictionary (which won't be correct) instead of what they want to do (which is check that there is a non-empty hashset contained within that dictionary).

Anyways, just doing a size == 0, then remove, would fix these issues.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -10,6 +11,19 @@
 
 namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
 {
+    /// <summary>
+    /// Base folder types for module searched by the MixedRealityToolkitFiles utilty.
```
wiwei:
```suggestion
    /// Base folder types for modules searched by the MixedRealityToolkitFiles utilty.
```

lukastoenneMS:
@andreiborodin Can we safely assume that the MRTK folders end in either "MixedRealityToolkit" or "MRTK" (with ".Service" etc. suffix)? If so, then the current test can be expected to work. I've tested by making a new project and moving the imported mrtk package folders into a subfolder.

Otherwise we will need a more sophisticated test, such as looking for the assembly that contains a class. Such an assembly/reflection based test would probably also need to run per-assembly because the core assembly doesn't link against any of the others. So we'd need a little test script in each module that registers its own module's path in the MixedRealityToolkitFiles.
https://stackoverflow.com/questions/52797/how-do-i-get-the-path-of-the-assembly-the-code-is-in

lukastoenneMS:
The usual case here seems to be that people import the MRTK package into a subfolder as-is, i don't think we need to support changes the actual "MixedRealityToolkit***" folder names. Please shout if you think otherwise, i'm going to commit the change in the meantime since it fixes the immediate issue.

PR #4480, opened 2019-05-20
Refactor localization logic for clearer class naming/less component coupling

chrisfromwork:
The first wave of alignment changes came in kind of hot/weren't thoroughly reviewed based on some time constraints. This review changes some items in preparation for the final spectator view experience.

1) SpatialCoordinate related classes have been moved out of the state synchronization code base. SpatialCoordinateSystemManager now registers itself with the StateSynchronizationObserver and StateSynchronizationBroadcaster to obtain network events. But now spatial coordinate related classes no longer misuse the broadcaster/observer naming. In the future the spatial coordinate manager will need to have bidirectional communication across devices.

2) *LocalizationMechanism class names were reduced to *SpatialLocalizer for legibility. It also appeared that a great deal of marker detection specific logic was living in a base class. This logic has been moved to ArUcoMarkerDetectorSpatialLocalizer so that work can also begin on ArUcoMarkerVisualizerSpatialLocalizer.

3) ConnectedObserver has been renamed to SpatialCoordinateSystemMember. This class was used by both broadcasters and observers in the state synchronization world to identify members of the coordinate system/shared experience. The name ConnectedObserver seemed inappropriate.

4) A new scene was created for the state synchronized spectator view experience (SpectatorViewStateSynchronization.cs). Requiring alignment logic in order to test state synchronization code is unnecessary coupling. Although most state synchronization will want a shared coordinate system, that may not always be the case in the future for remote experiences/VR.

5) ArUcoMarker*Localizer logic has been refactored to take any marker detector.

Note: This breaks changes the original SpectatorView.cs script to become SpectatorViewOld.cs


Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/ICommandHandler.cs
```diff
@@ -0,0 +1,15 @@
...
+using Microsoft.MixedReality.Toolkit.Extensions.Experimental.Socketer;
+using System.IO;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
+{
+    public interface ICommandHandler
+    {
+        void OnConnected(SocketEndpoint endpoint);
+        void OnDisconnected(SocketEndpoint endpoint);
+        void HandleCommand(string command, SocketEndpoint endpoint, BinaryReader reader);
```
matthejo: >string command, SocketEndpoint endpoint [](start = 27, length = 39)

Nit: I would reorder these parameters so the SocketEndpoint is always the first parameter for these 3 functions

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/ICommandService.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
+{
+    public interface ICommandService
+    {
+        bool Register(string command, ICommandHandler handler);
```
matthejo: >Register [](start = 13, length = 8)

Nit: I'd name these RegisterCommandHandler / UnregisterCommandHandler so that they're a little more clear in context

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationBroadcaster.cs
```diff
@@ -16,11 +16,12 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
     /// </summary>
     public class StateSynchronizationBroadcaster : Singleton<StateSynchronizationBroadcaster>
```
matthejo: >Singleton<StateSynchronizationBroadcaster> [](start = 51, length = 42)

Shouldn't this implement ICommandService?

matthejo: >StateSynchronizationBroadcaster [](start = 17, length = 31)

Shouldn't this implement ICommandService?

matthejo: >Singleton<StateSynchronizationBroadcaster> [](start = 51, length = 42)

Would be nice to have a base class CommandService<T> : Singleton<T> to contain the shared implementation of adding and removing handlers and broadcasting a command to the right ones

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationBroadcaster.cs
```diff
@@ -108,38 +103,30 @@ private void DebugLog(string message)
...
             DebugLog($"Broadcaster received connection from {endpoint.Address}.");
 
-            ConnectedObserver connectedObserver = new ConnectedObserver(Role.Broadcaster, endpoint, () => Instantiate(anchorPrefab), debugLogging);
-            connectedObservers.Add(endpoint, connectedObserver);
             Connected?.Invoke(endpoint);
 
-            DebugLog($"Broadcaster kicking off spatial localization on observer from {endpoint.Address}.");
-
-            if (SpatialLocalizationMechanism != null)
+            foreach (var handler in allHandlers)
```
matthejo: >foreach (var handler in allHandlers) [](start = 12, length = 36)

Right now these command-raising loops are not reentrancy safe. If anyone tries to call Register or Unregister while in an OnConnected/OnDisconnected/HandleCommand, it will modify the collection during enumeration.

You could make this safe by either iterating over a copy of the list each time, or by putting a guard against reentrancy so that calls to Register/Unregister perhaps put the new connections into a "pending changes" list if iteration is happening.

This might be design overkill, but I've been bitten by this particular issue a lot in the past.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationBroadcaster.cs
```diff
@@ -156,17 +143,12 @@ protected void OnReceive(IncomingMessage data)
                             StateSynchronizationSceneManager.Instance.ReceiveMessage(data.Endpoint, reader);
                         }
                         break;
```
matthejo: The StateSynchronizationSceneManager could add itself as a command handler

chrisfromwork: I'm going to avoid this for now. The "SYNC" header is used in different places by different classes. This command registration makes more sense when all headers are managed by the same class.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/ArUcoMarkerVisualizerSpatialLocalizer.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        internal override void ProcessIncomingMessage(Role role, Guid token, BinaryReader r)
+        {
+            throw new NotImplementedException();
+        }
+
+        internal override void Uninitialize(Role role, Guid token)
+        {
+            throw new NotImplementedException();
+        }
```
matthejo: Is this implementation coming in a future PR?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/SpatialCoordinateSystemMember.cs
```diff
@@ -13,62 +13,57 @@
 namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
 {
     /// <summary>
-    /// The SpectatorView helper class for managing the connected observer.
+    /// The SpectatorView helper class for managing a participant in the spatial coordinate system
     /// </summary>
-    internal class ConnectedObserver : DisposableBase
+    internal class SpatialCoordinateSystemMember : DisposableBase
```
matthejo: >SpatialCoordinateSystemMember [](start = 19, length = 29)

Nit: Maybe name this "SpatialCoordinateSystemParticipant" or something? It took me a second to parse "member"

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/SpatialLocalizer.cs
```diff
@@ -25,7 +32,7 @@ internal abstract class SpatialLocalizationMechanismBase : MonoBehaviour
         /// <summary>
         /// The type name of this object instance.
         /// </summary>
-        private string TypeName => typeName ?? (typeName = GetType().Name);
+        public string TypeName => typeName ?? (typeName = GetType().Name);
```
andreiborodin: >TypeName [](start = 22, length = 8)

Why make this public?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/SpatialCoordinateSystemManager.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        protected override void OnDestroy()
+        {
+            UnregisterCommands();
+            CleanUpMembers();
+        }
+
+        private void RegisterCommands()
+        {
+            foreach(var command in supportedCommands)
```
matthejo: >h( [](start = 18, length = 2)

Nit: a few of these are missing a space after foreach

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/SpatialLocalizer.cs
```diff
@@ -14,8 +14,15 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
...
     /// <remarks>In the future this would move to SpatialLocalization in a better form, abstraction-wise.</remarks>
-    internal abstract class SpatialLocalizationMechanismBase : MonoBehaviour
+    internal abstract class SpatialLocalizer : MonoBehaviour
     {
+        /// <summary>
+        /// The spatial coordinate service for sub class to instantiate and this helper base to rely on.
+        /// </summary>
+        protected abstract ISpatialCoordinateService SpatialCoordinateService { get; }
+
+        protected readonly object lockObject = new object();
```
andreiborodin: Why move this here?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/ArUcoMarkerDetectorSpatialLocalizer.cs
```diff
@@ -11,25 +12,33 @@
 namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
 {
     /// <summary>
-    /// Helper spatial localization method where the host finds a coordinate and gives its id to the observer
+    /// SpatialLocalizer that detects an ArUco marker
     /// </summary>
-    internal abstract class BroadcasterCoordinateSpatialLocalizationMechanism : SpatialLocalizationMechanismBase
+    internal class ArUcoMarkerDetectorSpatialLocalizer : SpatialLocalizer
```
andreiborodin: This class was supposed to be generic, I was going to extend it for ASA

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/SpatialCoordinateSystemManager.cs
```diff
@@ -0,0 +1,123 @@
...
+                StateSynchronizationBroadcaster.Instance.Register(command, this);
+                StateSynchronizationObserver.Instance.Register(command, this);
+            }
+        }
+
+        private void UnregisterCommands()
+        {
+            foreach (var command in supportedCommands)
+            {
+                StateSynchronizationBroadcaster.Instance.Unregister(command, this);
```
matthejo: >StateSynchronizationBroadcaster.Instance [](start = 16, length = 40)

These probably need null checks for cases where there's only one of the two in the scene? Or will they always both be there because of the unified prefab?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/ArUcoMarkerDetectorSpatialLocalizer.cs
chrisfromwork: I will push changes to split back this functionality for ASA and will keep ArUcoMarkerVisualizerSpatialLocalizer still inheriting from SpatialLocalizer.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/CommandService.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
+{
+    public class CommandService : Singleton<CommandService>,
+        ICommandService
+    {
+        private readonly object lockObject = new object();
+        private Dictionary<string, List<ICommandHandler>> commandHandlers = new Dictionary<string, List<ICommandHandler>>();
+        private List<ICommandHandler> allHandlers = new List<ICommandHandler>();
```
matthejo: >private List<ICommandHandler> allHandlers = new List<ICommandHandler>(); [](start = 8, length = 72)

One thought I had to make this much more efficient for allocations: use an ImmutableList<ICommandHandler>. The way immutable lists work, they create a new reference to a snapshot of the list when you add or remove items without necessarily needing to perform array allocations or copies. Then the CommandHandlers property could safely return 'allHandlers' without having to worry about collection modification during enumeration.

matthejo: Although I haven't used this in a HoloLens application, not sure which .NET Framework versions support ImmutableList.

---
In reply to: [286255732](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4480#discussion_r286255732) [](ancestors = 286255732)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/CommandService.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        public bool UnregisterCommandHandler(string command, ICommandHandler handler)
+        {
+            if (!commandHandlers.ContainsKey(command) ||
+                !commandHandlers[command].Contains(handler) ||
+                !allHandlers.Contains(handler))
+            {
+                return false;
+            }
```
matthejo: For this class to be thread-safe, any access to those fields should be inside the lock(lockObject) block, including reads from those collections

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
```diff
@@ -209,10 +193,14 @@ private void OnReceive(IncomingMessage data)
...
-                        DebugLog("Passing message to observer");
-                        connectedObserver.ReceiveMessage(reader);
+                    default:
+                        if (CommandService.Instance.CommandHandlerDictionary.TryGetValue(command, out var handlers))
+                        {
+                            foreach (var handler in handlers)
+                            {
+                                handler.HandleCommand(data.Endpoint, command, reader);
+                            }
+                        }
```
matthejo: Maybe this method could just be a method on CommandService so that the logic for how to broadcast the command is contained there?

chrisfromwork:
@matthejo and @andreiborodin I'm going to check this in for now so that others can use this code. I will address the comments here in a follow up review.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/ArUcoMarkerVisualizerSpatialLocalizer.cs
chrisfromwork: yes this will be updated in the future.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/CommandService.cs
chrisfromwork: System.Collections.Immutable doesn't resolve for me in Unity. For now this probably suffices, but I'll sync with you to better understand whether i can change the .net version to pick up the immutable namespace/classes.

Issue #4405, opened 2019-05-16
Cursor disappears when you bring up articulated hands set to GGVPointer


julenka:
## Describe the bug
If you set up articulated hands to use V1 GGV pointer, when you bring up your hands, the cursor will disappear.

Fix is to use the following code in GazePointerVisisbiiltyStateMachine.cs:

```
        public void UpdateState(int numNearPointersActive, int numFarPointersActive)
        {
            GazePointerState newState = gazePointerState;
            bool canGazeCursorShow = numFarPointersActive == 1 && numNearPointersActive == 0;
            switch (gazePointerState)
            {
                case GazePointerState.Initial:
                    if (canGazeCursorShow)
                    {
                        // There is some pointer other than the gaze pointer in the scene, assume
                        // this is from a motion controller or articulated hand, and that we should
                        // hide the gaze pointer
                        newState = GazePointerState.GazePointerInactive;
                    }
                    break;
                case GazePointerState.GazePointerActive:
                    if (canGazeCursorShow)
                    {
                        newState = GazePointerState.GazePointerInactive;
                        activateGazeKeywordIsSet = false;
                    }
                    break;
                case GazePointerState.GazePointerInactive:
                    // Go from inactive to active if we say the word "select"
                    if (activateGazeKeywordIsSet)
                    {
                        newState = GazePointerState.GazePointerActive;
                        activateGazeKeywordIsSet = false;
                    }
                    break;
                default:
                    break;
            }
            gazePointerState = newState;
        }
```

julenka:
Setting as this milestone because it's a regression from my PR #4176

PR #4482, opened 2019-05-20
Eye targeting is not working with near hand pointers

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4431

In a previous change (https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4270) a gaze provider state machine was added to bring the head gaze pointer behavior in-line with how the shell works. This had the side effect of also impacting eye-gaze state. This change brings the old eye-gaze behavior back by updating the state machine to be aware of the type of gaze being provided.

Also added a bunch of tests for this new thing, along with tests that show that transitions between the two (for example, when the eye tracking system gains tracking or loses tracking) works well.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
@@ -878,7 +878,10 @@ namespace Microsoft.MixedReality.Toolkit.Input
             }
             if (gazePointer != null)
             {
-                gazePointerStateMachine.UpdateState(NumNearPointersActive, NumFarPointersActive);
+                gazePointerStateMachine.UpdateState(
+                    NumNearPointersActive,
+                    NumFarPointersActive,
+                    InputSystem.EyeGazeProvider.IsEyeGazeValid);
 
                 // The gaze cursor's visibility is controlled by IsInteractionEnabled
                 gazePointer.IsInteractionEnabled = gazePointerStateMachine.IsGazePointerActive;
--- a/Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
@@ -74,11 +74,11 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
         private void UpdateStateHeadGaze(int numNearPointersActive, int numFarPointersActive)
         {
-            bool isMotionControllerOrHandUp = numFarPointersActive > 0 || numNearPointersActive > 0;
+            bool canGazeCursorShow = numFarPointersActive == 1 && numNearPointersActive == 0;
             switch (gazePointerState)
             {
                 case GazePointerState.Initial:
-                    if (isMotionControllerOrHandUp)
+                    if (!canGazeCursorShow)
                     {
                         // There is some pointer other than the gaze pointer in the scene, assume
                         // this is from a motion controller or articulated hand, and that we should
@@ -87,7 +87,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
                     }
                     break;
                 case GazePointerState.GazePointerActive:
-                    if (isMotionControllerOrHandUp)
+                    if (!canGazeCursorShow)
                     {
                         activateGazeKeywordIsSet = false;
                         gazePointerState = GazePointerState.GazePointerInactive;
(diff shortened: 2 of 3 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
```diff
@@ -4,31 +4,76 @@
...
+
+            gazePointerState = isEyeGazePointerActive ?
+                GazePointerState.GazePointerActive :
+                GazePointerState.GazePointerInactive;
+        }
+
+        private void UpdateStateHeadGaze(int numNearPointersActive, int numFarPointersActive)
         {
-            GazePointerState newState = gazePointerState;
             bool isMotionControllerOrHandUp = numFarPointersActive > 0 || numNearPointersActive > 0;
```
julenka: This reminds me of bug #4405. Mind pulling that change in? It's very short, just replace with 

> bool canGazeCursorShow = numFarPointersActive == 1 && numNearPointersActive == 0;

Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/GazePointerStateMachineTests.cs
```diff
@@ -34,16 +34,122 @@ public void TestHandAndSpeechBehaviour()
...
+            TestUtilities.InitializeMixedRealityToolkitScene(true);
+
+            // Initial state: gaze pointer active
+            var gsm = new GazePointerVisibilityStateMachine();
+            Assert.IsTrue(gsm.IsGazePointerActive, "Gaze pointer should be visible on start");
+
+            // With the hand raised, eye gaze pointer should still exist because only far interaction causes the 
+            // eye gaze pointer to go away.
+            gsm.UpdateState(1 /*numNearPointersActive*/, 0 /*numFarPointersActive*/, true);
+            Assert.IsTrue(gsm.IsGazePointerActive, "With near interaction, gaze pointer should continue to exist");
```
julenka: eye gaze?

Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/GazePointerStateMachineTests.cs
```diff
@@ -34,16 +34,122 @@ public void TestHandAndSpeechBehaviour()
...
+            gsm.UpdateState(0 /*numNearPointersActive*/, 1 /*numFarPointersActive*/, true);
+            Assert.IsFalse(gsm.IsGazePointerActive, "With far interaction, gaze pointer should go away");
+
+            // Reset the state and validate that it goes back to being visible.
+            gsm.UpdateState(0 /*numNearPointersActive*/, 0 /*numFarPointersActive*/, true);
+            Assert.IsTrue(gsm.IsGazePointerActive, "Gaze pointer should be visible when no near or far pointers");
+
+            // Saying "select" should have no impact on the state of eye gaze-based interactions.
+            SpeechEventData data = new SpeechEventData(EventSystem.current);
+            data.Initialize(new BaseGenericInputSource("test input source", new IMixedRealityPointer[0], InputSourceType.Voice),
```
julenka: Maybe pull into utility? We've pasted this a few times now.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -878,7 +878,10 @@ private void ReconcilePointers()
             }
             if (gazePointer != null)
             {
-                gazePointerStateMachine.UpdateState(NumNearPointersActive, NumFarPointersActive);
+                gazePointerStateMachine.UpdateState(
+                    NumNearPointersActive,
+                    NumFarPointersActive,
+                    InputSystem.EyeGazeProvider.IsEyeGazeValid);
```
julenka: Can this be null ever?

Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/GazePointerStateMachineTests.cs
wiwei: Put into a static private, utility is overkill until we need to use it in multiple places.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
wiwei: InputSystem is a prereq to this file (i.e. other places have the same level of dependency)

EyeGazeProvider is set up by the input system uniquely within its startup path and startup will fail if there is no gaze provider given.

So tl;dr Not as far as I can see.

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
wiwei: Done - I had to change up the suggested implementation a bit (i.e. #4405 had a bug in its proposed implementation in the missing negation of canGazeCursorShow within its if checks)

Issue #4470, opened 2019-05-18
Vive Pro Defaults OpenVr Generic Device causing one hand operation errors due to missing Spatial Grip Input Action


JShull:
## Describe the bug
When using the default MRTK profile provided with Vive Pro hardware you cannot use one hand interactions with the controllers.

## To reproduce
**This happens when using Vive Pro hardware**
1.) SteamVR added to the Project
2.) Make sure you have all of the other XR settings on as suggested by the generic toolkit setup.
3.) Run the HandInteractionExample scene and try to pick up an item with one controller
4.) Error 'InvalidOperationException': Sequence contains no matching element' --> ManipulationHandler.cs the line in the code that generates the error is 
`return controller.Interactions?.First(x => x.InputType == DeviceInputType.SpatialGrip);`

## Expected behavior

I would expect the one hand interaction to work with the trigger as I am using the default provided MRTK profiles.

## Screenshots
Slack thread image of the error --> https://holodevelopers.slack.com/files/U8RDSPJR3/FJT9YB71N/image.png

## Your Setup (please complete the following information)

- Unity Version 2019.1.3f1
- MRTK Version 2.0.0-RC1

## Target Platform (please complete the following information)

-Windows Standard Build
- OpenVR
-Vive Pro Hardware
## Additional context
**Solution**
One solution, update the Generic Open VR Left/Right Hand Controller Input Action Assignment to include _Spatial Grip_
Slack thread solution--> https://holodevelopers.slack.com/files/U8RDSPJR3/FJTJXBBEU/inputactionupdate.png

wiwei:
Closing this one out now that it's been merged

PR #4484, opened 2019-05-20
Vive Pro Defaults OpenVr Generic Device causing one hand operation errors due to missing Spatial Grip Input Action

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4470

Adds the Input Action assignment to support Vive Pro controllers


wiwei:
For this change, I applied the fix that was mentioned on slack (and linked in the issue).

Unfortunately I don't have access to Vive Pro hardware - @JShull can you confirm that this addresses the issue that you hit?

JShull:
The Vive Pro defaults to the generic openVR controller but the spatial grip reference wasn't on the generic controller mapping so it throws an error the first time you go to one hand use the controller - not sure why it's looking for the spatial grip when I am using the trigger but it does! When I added in the spatial grip to the generic openVR controller mapping the error went away.

JShull:
Confirmed working with Vive Pro Hardware, the error no longer comes up and one hand pick up and move works as it should.

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityControllerMappingProfile.asset
```diff
@@ -2870,6 +2870,19 @@ MonoBehaviour:
       axisCodeY: AXIS_18
       invertXAxis: 0
       invertYAxis: 0
+    - id: 11
```
keveleigh: We should probably make sure the ID, description, and any other stale data from the create button (axisCodeX/Y as well) are reset for this action.

Might be an unrelated bug in the generic action mapper that everything is copied from the previous action, even if you aren't able to change it in the UI.

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityControllerMappingProfile.asset
```diff
@@ -2870,6 +2870,19 @@ MonoBehaviour:
...
+      description: WMR Touchpad Position
+      axisType: 7
+      inputType: 14
+      inputAction:
+        id: 3
+        description: Grip Pose
```
keveleigh: These should be added to the default interactions in the class as well:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/9ab0e9b69afe45f7856b5731436ae427c3fe7cc4/Assets/MixedRealityToolkit.Providers/OpenVR/GenericOpenVRController.cs#L37

luis-valverde-ms:
@wiwei , is this still alive?

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityControllerMappingProfile.asset
wiwei: Perfect, thanks!

---
In reply to: [286192378](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4484#discussion_r286192378) [](ancestors = 286192378)

wiwei:
Yep, sorry for the delay here. I got pulled into doing a bunch of other random work, but finally got back to this.

SimonDarksideJ:
Since you have made changes to the generic controller, have you confirmed this has no knock on impacts to other controllers?  I would have expected Vive Wand controller updates in the ViveWand controller class
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Assets/MixedRealityToolkit.Providers/OpenVR/ViveWandController.cs

Likely if the Vive Pro is reporting it's controllers differently to the base vive, in which case, it should have it's own definition anyway,

Assets/MixedRealityToolkit.Providers/OpenVR/GenericOpenVRController.cs
```diff
@@ -82,6 +82,7 @@ public GenericOpenVRController(TrackingState trackingState, Handedness controlle
             new MixedRealityInteractionMapping(9, "Unity Button Id 3", AxisType.Digital, DeviceInputType.ButtonPress, KeyCode.JoystickButton3),
             new MixedRealityInteractionMapping(10, "WMR Touchpad Touch", AxisType.Digital, DeviceInputType.TouchpadTouch, KeyCode.JoystickButton18),
             new MixedRealityInteractionMapping(11, "WMR Touchpad Position", AxisType.DualAxis, DeviceInputType.Touchpad, ControllerMappingLibrary.AXIS_17, ControllerMappingLibrary.AXIS_18),
+            new MixedRealityInteractionMapping(12, "WMR Grip Pose", AxisType.SixDof, DeviceInputType.SpatialGrip, MixedRealityInputAction.None),
```
keveleigh: Is this specific to WMR? I believe this should work regardless of controller / platform: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/f3a97916629a965c9fc45e85290c453ebc9b45f3/Assets/MixedRealityToolkit/Providers/UnityInput/GenericJoystickController.cs#L207

SimonDarksideJ: Arguably, it's a mistake when I first put that forward as we only had WMR to test on.  These should all be called "Generic" as they are per the unaltered generic spec, not specific to any device.

keveleigh:
>Since you have made changes to the generic controller, have you confirmed this has no knock on impacts to other controllers? I would have expected Vive Wand controller updates in the ViveWand controller class

Yeah, it's possible this mapping is missing from other OpenVR controllers as well (it does appear to be missing from the Vive Wand class). We should probably log this or go through in a subsequent PR to ensure feature parity across classes that support them.

>Likely if the Vive Pro is reporting it's controllers differently to the base vive, in which case, it should have it's own definition anyway

That makes sense, but this PR is a good first step in that direction by at least unblocking devs from using that controller with the Toolkit.

luis-valverde-ms:
@wiwei , are you awaiting reviews or have pending changes to implement?

wiwei:
Thanks for the reminder :)

PR #4185, opened 2019-05-06
Add PlayMode tests to the PR and CI pipelines.

wiwei:
As we add more playmode tests we want to make sure that other folks aren't breaking them inadvertently. Easiest way of doing this is to enable these tests as part of PR validation (and CI as well)

PR #4498, opened 2019-05-21
Add play mode tests to CI pipelines

googlan:
Modifications to CI profile, taken from PR #4185 by wiwei.
Change Error on initialization of dictation and speech services to warning.


wiwei:
I think that the speech/dictation issues are actually red herrings - I believe that @lukastoenneMS fixed the speech/dictation issues a while back (by making them still log, but no longer die on exception).

I think that you'll actually need to remote into these CI machines to see what is missing to get these play mode tests working. I think that the build machines are actually probably missing some underlying setup (i.e. mixed reality portal/steamvr) which might be required to get things to work here.

googlan:
@wiwei Do you mean this error message? :
"Windows Mixed Reality and Holographic Emulation are not supported on operating systems prior to Windows 10 Fall Creators Update (Redstone 3)."
Do you think it's just missing mixed reality portal or can actually be an outdated Windows version?

Without recognizers the system won't die, but the error message will cause a test failure.

Another issue tests hit is a missing UIRaycastCamera, which seems to be unrelated to CI machines, I'll try to repro that locally.

wiwei:
That's one of them - in the past we've tried doing local repros of things and found that they are to some degree unique to the build setup of the machine (i.e. when doing a local build of the .NET backend stuff we didn't have any issues because we had run VS on the machine which restored nuget packages, and thus we had to actually get onto the machines themselves to further investigate the difference)

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -242,6 +242,9 @@ public override void Reset()
         /// <inheritdoc />
         public override void Disable()
         {
+            MixedRealityInputSystemProfile profile = ConfigurationProfile as MixedRealityInputSystemProfile;
```
thalbern: please add comment why we're doing this weird gaze provider destroy

Assets/MixedRealityToolkit/Utilities/MixedRealityPlayspace.cs
```diff
@@ -17,6 +17,13 @@ public static class MixedRealityPlayspace
 
         private static Transform mixedRealityPlayspace;
 
+        public static void Destroy()
+        {
+            CameraCache.Main.transform.SetParent(null, false);
```
thalbern: add comment

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
```diff
@@ -157,19 +157,24 @@ public override void Disable()
             {
                 StopRecognition();
                 keywordRecognizer.OnPhraseRecognized -= KeywordRecognizer_OnPhraseRecognized;
+
+                keywordRecognizer.Dispose();
```
keveleigh: I'm not sure it makes sense to `Dispose` the keyword recognizer fully when just disabling the system.
This should also be happening already in the system's Dispose call. Are you not seeing that called reliably?

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
```diff
@@ -157,19 +157,24 @@ public override void Disable()
...
             }
+
+            keywordRecognizer = null;
         }
 
-#if UNITY_EDITOR
         /// <inheritdoc />
         public override void Destroy()
         {
+            Disable();
```
keveleigh: `Disable` should only really be called from the service's manager. We shouldn't need to call it here.
Is a manager (maybe `MixedRealityToolkit.cs`?) trying to destroy services that it hasn't disabled yet?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -242,6 +242,12 @@ public override void Reset()
         /// <inheritdoc />
         public override void Disable()
         {
+            // Input System adds a gaze provider component on the main camera, which needs to be removed when the input system is disabled/removed.
+            // Otherwise the component would keep references to dead objects.
+            // Unity's way to remove component is to destroy it. Immediate is important, otherwise order of deinitialization will be arbitrary.
+            MixedRealityInputSystemProfile profile = ConfigurationProfile as MixedRealityInputSystemProfile;
+            UnityEngine.Object.DestroyImmediate(CameraCache.Main.gameObject.GetComponent(profile.PointerProfile.GazeProviderType.Type));
```
keveleigh: [From the Unity docs](https://docs.unity3d.com/ScriptReference/Object.DestroyImmediate.html):

>This function should only be used when writing editor code since the delayed destruction will never be invoked in edit mode.

Also:

>You are strongly recommended to use Destroy instead.

Does the order of deinitialization cause issues that can't be resolved in other ways?

keveleigh: Also, are we not able to destroy using the input system's `GazeProvider` reference? Probably after a cast to `MonoBehaviour` or `UnityEngine.Object`.

googlan: I apprectiate both comments, because they helped me to find another bug, which I "fixed" by adding `Disable` calls into `Destroy` calls :smile: 

Also I didn't notice, that there was actually a call to DestroyImmediate here before, and I just needed to add one for play mode.

---
In reply to: [287528647](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4498#discussion_r287528647) [](ancestors = 287528647)

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
googlan: I could do this in `Destroy` (Dispose is not enough, because I'm not getting a `Dispose` call on a service before the next test starts). I thought it makes sense here to mirror initialization in `Enable`. It will make sense if we will ever want to dynamically add remove voice keywords. Don't know if it makes sense?

---
In reply to: [287527461](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4498#discussion_r287527461) [](ancestors = 287527461)

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
googlan: Yes, that was a problem in `MixedRealityToolkit.cs`. Added Disable call there. Thanks for the comment!

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -242,18 +242,24 @@ public override void Reset()
...
+            // Otherwise the component would keep references to dead objects.
+            // Unity's way to remove component is to destroy it.
+            if (GazeProvider != null)
             {
-                var component = CameraCache.Main.GetComponent<IMixedRealityGazeProvider>() as Component;
-
-                if (component != null)
+                if (Application.isPlaying)
+                {
+                    UnityEngine.Object.Destroy(GazeProvider as Component);
```
keveleigh: This component should be added at edit time, so it seems strange to destroy it at runtime.
Especially in `Disable`, which is the inverse of `Enable`, not `Initialize` (which is where the gaze provider is added).

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
keveleigh: This also might regress #4439. @wiwei?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
googlan: This component is added in both Edit mode and Runtime mode. I've added destroying in runtime case, as destroy in Edit mode has already been here before. 
Part about `Initialize` is true, but the old behavior was to set GazeProvider to null in `Disable`. It isn't set to anything in `Enable`, so input system won't work correctly unless `Reset` is called. I preserved this behavior.

keveleigh: >I've added destroying in runtime case, as destroy in Edit mode has already been here before.

I had thought this was going to be a problem, not knowing whether the provider was added at runtime or edit time, but from some testing it looks like I was wrong and the editor preserves state well.

>but the old behavior was to set GazeProvider to null in Disable

Good catch. Feels like there's a bug in there.

PR #4500, opened 2019-05-21
Enable .NET scripting backend support in CI, part 1/2

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/3656

This change updates the code and build scripts to the point where .NET backend builds will work in CI. There were a number of issues in getting to this point, and note that this change alone isn't enough to get things working in CI (we need to also update the pipelines YAML files to build the .NET backend)

This PR still lists all of the things needed to be changed, and **bolds** the ones that this one takes care of:

- **The play mode tests** have a dependency on the input simulation asmdef, but the input simulation asmdef isn't built on all of the the platforms that the playmodetests are run on (in particular, the one the that's interesting is WSA). This solution isn't super optimal but it's workable for getting this CI working.
- **UnityBuildPlayerTools** - this wasn't strictly necessary, but is something that users do in order to debug/build - typically we will enable C# projects when building .NET to be able to iterate faster within VS itself, instead of having to rebuild from within Unity each time. This makes it so that building the .NET flavor also flips this switch to on, to simulate what we use this scripting backend for.
- **NuGet restore** - when you build via visual studio, VS will actually restore nuget packages for you - without it random stuff will fail (i.e. you'll get things like "System" namespace wasn't found). Since we're not using VS to build (we're using msbuild directly) we need to do this step ourselves.

Things that will come in the next PR (to keep our pipeline and code changes separate)

- The folder destination of the appx package and its naming are also different when building C# projects, so instead of being in the build/AppPackages/MixedRealityToolkit folder, it's actually in the build/MixedRealityToolkit/AppPackages. And instead of the name MixedRealityToolkit_2.0.0.0_Win32_Master_Test, it's MixedRealityToolkit_2.0.0.0_x86_Master_Test. This will need to be a followup YAML pipeline switch on scripting backend.


Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityToolkit.Services.InputSimulation.asmdef
```diff
@@ -5,7 +5,8 @@
     ],
     "optionalUnityReferences": [],
     "includePlatforms": [
-        "Editor"
+        "Editor",
+        "WSA"
```
keveleigh: This should ease #3971 on WSA, but we'll still need a proper fix for all platforms.

Issue #4410, opened 2019-05-16
[MRTK Shell Parity] [Button] Fit & finish: proximity light intensity, distance, pulse effect polish, etc...


cre8ivepark:
Part of #4200 UX Controls Parity with HoloLens 2 Shell

cre8ivepark:
Merged.

Issue #4200, opened 2019-05-07
[MRTK Shell Parity] UX Controls Parity with HoloLens 2 Shell


chbecker-ms:
# RC2
## Cursor/Pointer
- [x] [#4417][PR #4446][Yoon] Asset: Replace current MRTK cursor with shell-parity cursor.
- [x] [#4418][PR #4446][Yoon] Asset: Contextual cursor icons
- [x] [#4407][PR #4526][Lars] Tether line visual fit & finish

## Button
- [x] [#4410][PR #4501][Yoon] Visual fit & finish: proximity light intensity, distance, pulse effect polish, etc...

## Bounding Box
- [x] [#4416][PR #4503][Yoon] Visual fit & finish: proximity light intensity, distance, handle color, etc.

## Slate
- [x] [#4412][PR #4514][Yoon] Audio feedback is missing
- [x] [#4513][PR #4514][Yoon] Touch point visual fit & finish

---

# GA
## Button
- [x] [#4979][**PR** #5024]Button highlight for far interaction not defined (hand ray or gaze + air-tap)
- [x] [#4977][**PR** #5049] Button's border highlighting need to be adjusted (highlight only focused one)

## Bounding Box
- [x] [#4415][**PR** #4895][John] Handles should be scaled up on hand proximity 

## Fingertip visualization
- [x] [#4411][**PR** #4836][Cameron] Match the final shell design spec
- [x] [#4561][**PR** #4836][OOBE] Fingertip cursor should converge at point of activation for buttons

## Contextual Cursor Icon
- [x] [#4332][**PR** #4995][Niall] Feature: Contextual cursor icon visualization capability
- [x] [#4471][**PR** #4995][Niall] Replace current MRTK cursor with shell-parity cursor (DefaultCursor prefab)

## Slate
- [x] [#4413][**PR** #4836][Cameron] Fingertip ring shadow on surface needed

---

#  GA+
## Button
- [x] [#4978][#5168] Button's front cage should be compressed. Currently, it moves back and penetrates through
- [x] [#4562][#5168][OOBE] Compressible button visuals tuning
- [x] [#4559][#5244][OOBE] Poke pointer position does not match shell
- [x] [#4563][OOBE] Proximity Shader GlobalProximityAssigner should use point of interaction (poke pointer position) instead of fingertip joint
- [x] [#4564][OOBE] Button 'splash visual' needs to trigger not on touch but on button press
[...]

PR #4501, opened 2019-05-21
[MRTK Shell Parity] [Button] Fit & finish: proximity light intensity, distance, pulse effect polish

cre8ivepark:
## Overview
Fit & finish: proximity light intensity, distance, pulse effect polish
Part of #4200 [MRTK Shell Parity] UX Controls Parity with HoloLens 2 Shell  

## Changes
- Fixes: #4410 

## Updates to match shell's design spec
- Updated front cage's borderline thickness, glow intensity (PressableButton prefab)
- Icon and Text moves the 1/2 distance of the front cage (PressableButton.cs)
- Updated example button sizes (HandInteractionExample scene)
- Updated Toggle Features panel and added it as a prefab

MRTK_Button_Polish_IconText2

MRTK_Button_Polish_Lighting

2019-05-21 12_04_05-Unity 2018 4 0f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof


julenka:
Why wouldn't we make pressablebutton have the nice background the pressablebutton plated has? Right now pressablebutton is transparent. I imagine most people will think they need to add pressablebutton, and not pressablebutton plated.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -23,6 +23,10 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
         [Tooltip("The object that is being pushed.")]
         private GameObject movingButtonVisuals = null;
 
+        [SerializeField]
```
julenka: I can see how you needed this to implement the specific behavior you needed for the shell buttons, but is it something you think should be in the generic pressable button, the thing that most developers use?

It seems like the specific behavior you want is to position the object exactly half way along the press amount. For this very specific behavior it may make more sense to have a component on the moving icon that positions itself based on the press direction of the button and the press amount.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -23,6 +23,10 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
...
+        [Tooltip("The icon and text content moving inside the button.")]
+        private GameObject movingButtonIconText = null;
```
julenka: If you think we should put this in the general pressable button, then make sure to expose this so that developers can configure this at runtime.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
cre8ivepark: Since the intention of this control is to provide the shell-parity, great default experience, we would like to include it here so that developers can get it free when it is used without modification. From the customization standpoint, I agree that we can expose this as a configurable element.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
cre8ivepark: Sounds great!

cre8ivepark:
Having PressableButton without the plate is for the scenarios where we have multiple buttons on the shared plate, leveraging the nested prefab. Might be better to have the name PressableButton as plated default and PressableButtonWithoutPlate.

2019-05-22 17_51_42-Unity 2018 3 8f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
cre8ivepark: Forgot to mention that this fixes the bug of icon & text being occluded by the back plate when pressed fully. Current button has a wrong thickness dimension which made the icon & text still visible on a full press.

cre8ivepark:
I tried renaming current PressableButton.prefab as PressableButtonNoPlate and PressableButtonPlated as PressableButton. However, git recognizes one of these changes as deleting & creating a new file instead of renaming. Not sure what's the best way to rename them without breaking existing users.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
julenka: Hm, I actually didn't realize that the intention of PressableButton.cs was to provide shell parity button. I thought PressableButton.cs intention was to provide a pressable button, meaning a button that has some travel amount when touched, and then fires press and release events based on thresholds. I think that might be where my  comments are coming from. I wonder if the only PressableButton behavior we provide should require to match the shell so strictly in behavior. Instead it could be better to make a  ShellParityButton.cs that extends PressableButton which would have this behavior.


For example, you could do something like the following:


```
public class HoloLens2ShellParityButton : PressableButton
{
    [SerializeField] MovingButtonIconText = null;

    protected void UpdateMovingVisualsPosition()
    {
        base.UpdateMovingVisualsPosition();
        if (movingButtonIconText != null)
        {
            Debug.Assert(initialTransform != null);
            movingButtonIconText.transform.position = initialTransform.position + WorldSpacePressDirection * currentPushDistance / 2;
        }
    }
}
```

Cameron-Micka:
> I tried renaming current PressableButton.prefab as PressableButtonNoPlate and PressableButtonPlated as PressableButton. However, git recognizes one of these changes as deleting & creating a new file instead of renaming. Not sure what's the best way to rename them without breaking existing users.

Hi Yoon, I'm going to pull these changes down now and give them a try. Thanks for doing this! The rename should be fine as long as the GUID stays the same. So, if you do a rename in the Unity editor the GUID should remain the same (you can inspect the associated meta file to verify).

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButton.prefab
```diff
@@ -1400,6 +1400,38 @@ MeshRenderer:
   m_SortingLayerID: 0
   m_SortingLayer: 0
   m_SortingOrder: 0
+--- !u!1 &8352633413104042190
```
Cameron-Micka: One question that has come up with partners and the pressable button is what is the purpose of the "Static Content?" Is that something we can delete while we are working in here? It seems to always be disabled.

image

cre8ivepark: We can delete them. Jonathan placed them as a placeholder when the design was not finalized. I'll update them.

Cameron-Micka: Awesome! Thank you Yoon.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
cre8ivepark: Now I understand. Essentially, we would like to provide a baseline PressableButton which developers can leverage it for implementing any custom buttons. PressableButtonHoloLens2 and PressableButtonRound are great examples. Might be better to rename prefabs to clarify this.

2019-05-23 14_54_06-Unity 2018 3 8f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButton.prefab
cre8ivepark: Removed redundant elements. Updated plate names for easier understanding.
2019-05-23 15_45_30-Unity 2018 3 8f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof

cre8ivepark:
Updates per review feedback:
- Reverted PressableButton.cs and Added new script **PressableButtonHoloLens2**.cs which contains shell design specific elements (icon & text movments) 
- Renamed PressableButtonPlated.prefab to **PressableButtonHoloLens2.prefab**
- Renamed PressableButton.prefab to **PressableButtonHoloLens2Unplated.prefab**
- Updated AppBar prefab which uses PressableButton prefab.
- No GUID changes. Developer should be able to update without breaking anything.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
cre8ivepark: Applied the updates! (description on the main comment thread)

cre8ivepark:
@Cameron-Micka Unchecked both HoverLight and ProximityLight on the backplate.

cre8ivepark:
I'll test these updates with the merge with Bernadette's PR #4527

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -60,7 +60,7 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
         // The maximum distance before the button is reset to its initial position when retracting.
         private const float MaxRetractDistanceBeforeReset = 0.0001f;
 
-        private float currentPushDistance = 0.0f;
+        protected float currentPushDistance = 0.0f;
```
julenka: This actually seems like something that would be useful to expose as public property (getter only) rather than make protected. For example for a script that wants to animate something based on pressed amount.

cre8ivepark:
@julenka I have been noticing the same issue when I keep working on a scene and run play mode multiple times - framerate drops and the camera movement becomes slow. Not sure if it is MRTK issue or Unity issue. It runs smoothly when you restart. (even with all objects are in a viewport)
I'll also check on the device.

2019-05-24 11_48_26-Unity 2018 3 8f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof

cre8ivepark:
Confirmed it runs @ 60fps on the device.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
cre8ivepark: Made it public. Thanks!

cre8ivepark:
## Summary
- Added new script **PressableButtonHoloLens2.cs** which contains shell design specific elements (icon & text movments). (inherits PressableButton.cs)
- Renamed PressableButtonPlated.prefab to **PressableButtonHoloLens2**.prefab
- Renamed PressableButton.prefab to **PressableButtonHoloLens2Unplated**.prefab
- PressableButtonHoloLens2 and PressableButtonHoloLens2Unplated now use **PressableButtonHoloLens2** script.
- Updated AppBar prefab which uses PressableButton prefab.
- **No GUID changes**. Developer should be able to update without breaking anything.
- Existing Piano keys and Round buttons are still using PressableButton.cs script which are good example of customization.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -94,7 +94,8 @@ public SpaceMode DistanceSpaceMode
         // The maximum distance before the button is reset to its initial position when retracting.
         private const float MaxRetractDistanceBeforeReset = 0.0001f;
 
-        private float currentPushDistance = 0.0f;
+        public float currentPushDistance = 0.0f;
```
thalbern: this can be private

cre8ivepark: Made it public per suggestion - it could be useful for the developers who want to leverage current push distance.

julenka:
@Cameron-Micka could you please take a look to see if your blocking issues have been addressed?

cre8ivepark:
Finished updating the Button README file as well to reflect naming updates.

wiwei:
Merging despite @Cameron-Micka's review, will followup afterward.

PR #4505, opened 2019-05-21
Enable .NET scripting backend support in CI, part 2/2

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/3656

The folder destination of the appx package and its naming are also different when building C# projects, so instead of being in the build/AppPackages/MixedRealityToolkit folder, it's actually in the build/MixedRealityToolkit/AppPackages. And instead of the name MixedRealityToolkit_2.0.0.0_Win32_Master_Test, it's MixedRealityToolkit_2.0.0.0_x86_Master_Test. This will need to be a followup YAML pipeline switch on scripting backend.

Note that this change makes it run in CI and PR, which will increase times. Depending on duration we might look to split this step out into its own check (so that we can parallelize across machines). Note that simply replacing one of the other ones I'm not certain will work, because we depend on those existing pipelines to generate assemblies which get bundled into the final NuGet package.

### DIFF
--- a/pipelines/templates/tasks/unitybuild.yml
+++ b/pipelines/templates/tasks/unitybuild.yml
@@ -60,12 +60,12 @@ steps:
   - task: PublishBuildArtifacts@1
     displayName: 'Publish ${{ parameters.Platform }} ${{ parameters.Arch }} (${{ parameters.PackagingDir }}) ${{ parameters.ScriptingBackend }}'
     inputs:
+      ArtifactName: 'mrtk-build-${{ parameters.Arch }}'
       # The final location of the generated package depends on the type of scripting backend it's built against.
       # For the default scripting backend (IL2CPP) the naming of the appx follows the form below:
-      - ${{ if eq(parameters.ScriptingBackend, 'default' }}:
-        PathtoPublish: '$outDir\AppPackages\MixedRealityToolkit\MixedRealityToolkit_$(MRTKVersion).0_${{ parameters.PackagingDir }}_Master_Test'
+      ${{ if eq(parameters.ScriptingBackend, 'default') }}:
+        PathtoPublish: '$(outDir)\AppPackages\MixedRealityToolkit\MixedRealityToolkit_$(MRTKVersion).0_${{ parameters.PackagingDir }}_Master_Test'
       # For .NET scripting backends, the naming is slightly different (mainly the AppPackages and MixedRealityToolkit folder
       # names are reversed, and the Architecture is part of the AppX name)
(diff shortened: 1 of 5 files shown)


pipelines/templates/tasks/unitybuild.yml
```diff
@@ -58,5 +60,12 @@ steps:
   - task: PublishBuildArtifacts@1
     displayName: 'Publish ${{ parameters.Platform }} ${{ parameters.Arch }} (${{ parameters.PackagingDir }}) ${{ parameters.ScriptingBackend }}'
     inputs:
-      PathtoPublish: '$(Build.ArtifactStagingDirectory)\build\AppPackages\MixedRealityToolkit\MixedRealityToolkit_$(MRTKVersion).0_${{ parameters.PackagingDir }}_Master_Test'
+      # The final location of the generated package depends on the type of scripting backend it's built against.
+      # For the default scripting backend (IL2CPP) the naming of the appx follows the form below:
+      - ${{ if eq(parameters.ScriptingBackend, 'default' }}:
```
keveleigh:
```suggestion
      - ${{ if eq(parameters.ScriptingBackend, 'default') }}:
```

pipelines/templates/tasks/unitybuild.yml
```diff
@@ -58,5 +60,12 @@ steps:
...
     displayName: 'Publish ${{ parameters.Platform }} ${{ parameters.Arch }} (${{ parameters.PackagingDir }}) ${{ parameters.ScriptingBackend }}'
     inputs:
-      PathtoPublish: '$(Build.ArtifactStagingDirectory)\build\AppPackages\MixedRealityToolkit\MixedRealityToolkit_$(MRTKVersion).0_${{ parameters.PackagingDir }}_Master_Test'
+      # The final location of the generated package depends on the type of scripting backend it's built against.
+      # For the default scripting backend (IL2CPP) the naming of the appx follows the form below:
+      - ${{ if eq(parameters.ScriptingBackend, 'default' }}:
+        PathtoPublish: '$outDir\AppPackages\MixedRealityToolkit\MixedRealityToolkit_$(MRTKVersion).0_${{ parameters.PackagingDir }}_Master_Test'
+      # For .NET scripting backends, the naming is slightly different (mainly the AppPackages and MixedRealityToolkit folder
+      # names are reversed, and the Architecture is part of the AppX name)
+      - ${{ if eq(parameters.ScriptingBackend, '.NET' }}:
```
keveleigh:
```suggestion
      - ${{ if eq(parameters.ScriptingBackend, '.NET') }}:
```

PR #4507, opened 2019-05-21
Relax the thresholds on the ArUco marker detection

matthejo:
This change reduces the threshold for accepting/rejecting markers, which allows detection to succeed in more cases without significantly affecting the quality of the successfully-detected markers.


Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/SpectatorViewPluginArUcoMarkerDetector.cs
```diff
@@ -360,8 +360,8 @@ private sealed class StationaryMarkerDetectionCompletionStrategy : MarkerDetecti
             private const int _requiredObservations = 5;
             private const int _requiredInlierCount = 5;
             private const int _maximumMarkerSampleCount = 15;
-            private const float _maximumPositionDistanceStandardDeviation = 0.001f;
-            private const float _maximumRotationAngleStandardDeviation = 0.25f;
+            private const float _maximumPositionDistanceStandardDeviation = 0.01f;
+            private const float _maximumRotationAngleStandardDeviation = 0.75f;
```
chrisfromwork: It seems like it may be worth exposing these values as serialized fields. Users may always need to manipulate these values and it will become more difficult when binaries are distributed via nuget.

PR #4521, opened 2019-05-22
Enable .NET scripting backend support in CI, part 2/2

wiwei:
#3656

The folder destination of the appx package and its naming are also different when building C# projects, so instead of being in the build/AppPackages/MixedRealityToolkit folder, it's actually in the build/MixedRealityToolkit/AppPackages. And instead of the name MixedRealityToolkit_2.0.0.0_Win32_Master_Test, it's MixedRealityToolkit_2.0.0.0_x86_Master_Test. This will need to be a followup YAML pipeline switch on scripting backend.

Note that this change makes it run in CI and PR, which will increase times. Depending on duration we might look to split this step out into its own check (so that we can parallelize across machines). Note that simply replacing one of the other ones I'm not certain will work, because we depend on those existing pipelines to generate assemblies which get bundled into the final NuGet package.


wiwei:
Having some issues with getting Azure Pipelines to actually kick off the build in response to /azp run mrtk_pr, so I manually kicked off a build here and it's green

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=1763

Issue #4219, opened 2019-05-07
Make start interaction plane on pressable button configurable


thalbern:
## Describe the problem

currently the point for that plane is the pivot of the movable button part. you can only modify it by putting a parent gameobject to that movable part and moving the pivot throught this hack 

## Describe the solution you'd like

Make the start interaction plane configurable through the debug manipulators (independent from the movable button part)

## Additional context

feature is already implemented in a draft PR from @danielhofmann-ms

wiwei:
I believe this issue was fixed by @thalbern in #4144 (due to the comment left:

_Yes there's at least 2 tickets that are fixed with this PR (#4219, #4216) - but I just wanted to open this as a draft for having a discussion about it. If possible I'd like to avoid having the migration step (if we have to keep it in it should be moved out of the class)._

thalbern:
not fixed! i was asked to close the draft PR - there's a comment in there that says there's going to be a second PR with the actual fix - the draft was just for discussing the fixes

Issue #4216, opened 2019-05-07
Trigger planes in pressable button won't scale with button


thalbern:
## Describe the bug

Planes for defining the distances for a button to trigger / release are not scaled with the button.

## To reproduce

Create a pressable button and up or downscale the button. Observe the debug planes showing the maxDistance / releaseDistance etc won't adjust accordingly

## Expected behavior

Planes should either adjust accordingly to the scaling or have a setting to allow them to be scaled with the button (make distances local checkbox?)

## Your Setup (please complete the following information)

- MRTK Version v2 RC1

## Target Platform (please complete the following information)

- HoloLens 2

wiwei:
I believe that this was fixed by #4144 based on the following comment in that PR:

_Yes there's at least 2 tickets that are fixed with this PR (#4219, #4216) - but I just wanted to open this as a draft for having a discussion about it. If possible I'd like to avoid having the migration step (if we have to keep it in it should be moved out of the class)._

thalbern:
not fixed! i was asked to close the draft PR - there's a comment in there that says there's going to be a second PR with the actual fix - the draft was just for discussing the fixes

PR #4527, opened 2019-05-22
PressableButton fixes and PlayModeTests

thalbern:
## Overview
Changes are based on PR #4144

-Pressable Buttons now have a local and world space distance mode. Local space distance mode will take scaling into account.
-Start plane is now configurable and not automatically bound to the pivot of the moving visual.
-Unified how we calculate world space positions from plane distances so PressableButton and Inspector are using the same method.
-Removed superfluous gameobject that was created to calculate the world space positions of the distance planes
-Added button for converting local to world distances and vice versa in inspector
-Added PlayModeTests for button scaling in the 2 different modes and a test for converting plane distances

## Changes
- Fixes: #4219 , #4216 

## Verification
If you can sync and check if the button still behaves/triggers like you're expecting. Also try switching the modes in the edit mode and scale the button in both modes in z direction
Tests need to be run individually - running a batch of tests will fail due to a bug we have in tearing down the mrtk scene 

git fetch origin pull/4527/head:user/bethalha/button_fixes
git checkout user/bethalha/button_fixes


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -218,69 +319,78 @@ void IMixedRealityTouchHandler.OnTouchCompleted(HandTrackingInputEventData event
...
+        #region public transform utils
 
-        private void FindOrCreatePathMarkers()
+        /// <summary>
+        /// Returns world space position along the push direction for the given local distance
+        /// </summary>
+        /// <param name="localDistance"></param>
+        /// <returns></returns>
+        /// 
+        public Vector3 GetWorldPositionAlongPushDirection(float localDistance)
```
keveleigh: Do these need to be public? They seem like they'd only be interesting within the class itself for figuring out where to position itself.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -125,87 +112,74 @@ private ButtonInfo GatherCurrentInfo()
...
 
+            /// {dahof} NOTE: collider box editing is disabled as we would need to find the min / max point of it along the press direction, which is arbitrary.
+            /// I think for simplicity reasons, we should make editing the collider independent from editing the PressableButton distances.
+
             // BUTTON CONTENT ORIGIN
             // Don't allow editing of button position
-            Handles.color = Color.green;
-            DrawPlaneAndHandle(pressStartPlane, info.TouchCageLocalBounds.size * 0.4f, newInfo.StartPos, info.TouchStartOrigin, "Moving button visuals", false);
-
+            /*Handles.color = Color.green;
```
keveleigh: Should this commented out code just be removed?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -4,6 +4,7 @@
 using Microsoft.MixedReality.Toolkit.Input;
 using Microsoft.MixedReality.Toolkit.Utilities;
 using System.Collections.Generic;
+using UnityEditor;
```
julenka: Does this compile for UWP? In the past when I've done "using UnityEditor" in non-editor inspectors I have had build failures when compiling for UWP. Generally I believe it's good practice to avoid this.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -25,29 +26,72 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
 
         [SerializeField]
         [Header("Press Settings")]
+        [Tooltip("The offset at which pushing starts.")]
```
julenka: Offset from what?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -25,29 +26,72 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
...
+        private float startPushDistance = 0.0f;
+        public float StartPushDistance { get => startPushDistance; set => startPushDistance = value; }
+
+        [SerializeField]
         [Tooltip("Maximum push distance")]
```
julenka: Maximum relative to what?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -25,29 +26,72 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
...
-        [Header("Position markers")]
-        [Tooltip("Used to mark where button movement begins. If null, it will be automatically generated.")]
+        public enum SpaceMode
+        {
+            World,
+            Local
+        }
+
+        [SerializeField]
+        [HideInInspector]
```
julenka: Why do you need to both serialize the field and hide it in the inspector?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -25,29 +26,72 @@ public class PressableButton : MonoBehaviour, IMixedRealityTouchHandler
...
+                {
+                    distanceSpaceMode = value;
+                    float scale;
+                    if (distanceSpaceMode == SpaceMode.Local)
+                    {
+                        scale = transform.InverseTransformVector(WorldSpacePressDirection).magnitude;
+                    }
+                    else
+                    {
+                        scale = transform.TransformVector(WorldSpacePressDirection).magnitude;
```
julenka: Do you really intend to scale the **world**space press direction here? Seems like you are scaling a world space vector by a world space transform.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -98,14 +133,22 @@ private set
...
+                        IsPressing = false;
 
                         TouchEnd.Invoke();
                     }
                 }
             }
         }
 
+        /// <summary>
+        /// Represents the state of whether the button is currently being pressed.
```
julenka: It would be good to define pressing here. Whether pressing is we are touching the button, or pressing means the button has passed some press distance.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -98,14 +133,22 @@ private set
...
+        /// </summary>
+        public bool IsPressing { get; private set; }
```
julenka: Why are we exposing IsPressing but not IsTouching?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -119,9 +162,56 @@ private Vector3 WorldSpacePressDirection
...
+            get { return movingButtonVisuals != null ? movingButtonVisuals.transform : transform; }
+        }
+
+        private float LocalToWorldScale
+        {
+            get
+            {
+                return transform.TransformVector(WorldSpacePressDirection).magnitude;
+            }
+        }
```
julenka: nit: extra newline.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -119,9 +162,56 @@ private Vector3 WorldSpacePressDirection
...
+        private Vector3 LocalForward
+        {
+            get
```
julenka: Interesting that LocalForward is defined as the NearInteractionTouchable instead of z in local space.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -119,9 +162,56 @@ private Vector3 WorldSpacePressDirection
...
+
+            }
+        }
 
         #endregion
 
+
+        private void Awake()
+        {
+            currentPushDistance = startPushDistance;
```
julenka: Do we handle the case where a button gets disabled, moved, and then enabled? Does currentPushDistance get properly re-initialized / zeroed?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
julenka: In fact I'm not sure you need this using.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -134,6 +134,7 @@ public virtual void RenderCustomInspector()
             if (states.objectReferenceValue == null)
             {
                 InspectorUIUtility.DrawError("Please assign a States object!");
+                EditorGUILayout.EndVertical();
```
julenka: hehe, bugfix?

Assets/MixedRealityToolkit.Tests/PlayModeTests/PressableButtonTests.cs
```diff
@@ -0,0 +1,237 @@
...
+
+            return testButton;
+        }
+
+        #endregion
+
+        #region Tests
+
+        [UnityTest]
+        public IEnumerator Test01_PressableButtonInstantiate()
```
julenka: Per suggestion from @wiwei we want to avoid numbering the tests since it's arbitrary. Suggest removing the numbers to stay consistent with his suggestion.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PressableButtonTests.cs
```diff
@@ -0,0 +1,237 @@
...
+            Vector3 startPushDistanceWorldScaled = buttonComponent.GetWorldPositionAlongPushDirection(startPushDistance);
+            Vector3 maxPushDistanceWorldScaled = buttonComponent.GetWorldPositionAlongPushDirection(maxPushDistance);
+            Vector3 pressDistanceWorldScaled = buttonComponent.GetWorldPositionAlongPushDirection(pressDistance);
+            Vector3 releaseDistanceWorldScaled = buttonComponent.GetWorldPositionAlongPushDirection(releaseDistance);
+
+            // compare our distances
+            Assert.IsTrue(startPushDistanceWorld == startPushDistanceWorldScaled, "Start Distance was modified while scaling button gameobject");
+            Assert.IsTrue(maxPushDistanceWorld == maxPushDistanceWorldScaled, "Max Push Distance was modified while scaling button gameobject");
+            Assert.IsTrue(pressDistanceWorld == pressDistanceWorldScaled, "Press Distance was modified while scaling button gameobject");
+            Assert.IsTrue(releaseDistanceWorld == releaseDistanceWorldScaled, "Release Distance was modified while scaling button gameobject");
```
julenka: Since this is actually not relying on running the editor / doing hand movements / getting focus events, could this actually be an editor test?

Assets/MixedRealityToolkit.Tests/PlayModeTests/PressableButtonTests.cs
```diff
@@ -0,0 +1,237 @@
...
+
+            // switch back to world space
+            buttonComponent.DistanceSpaceMode = PressableButton.SpaceMode.World;
+
+            // distances must match up with original values 
+            Assert.IsTrue(startPushDistance == buttonComponent.StartPushDistance, "Conversion from local to world distances didn't return the correct world distances");
+            Assert.IsTrue(maxPushDistance == buttonComponent.MaxPushDistance, "Conversion from local to world distances didn't return the correct world distances");
+            Assert.IsTrue(pressDistance == buttonComponent.PressDistance, "Conversion from local to world distances didn't return the correct world distances");
+            float newReleaseDistance = buttonComponent.PressDistance - buttonComponent.ReleaseDistanceDelta;
+            Assert.IsTrue(releaseDistance == newReleaseDistance, "Conversion from local to world distances didn't return the correct world distances");
```
julenka: Same here, wondering if it could be an editor test instead of a playmode test. Advantage is editor tests run in CI today, playmode tests when they will run in CI is TBD as far as I know.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using Microsoft.MixedReality.Toolkit.Input;
```
julenka: When I make the planes editable, i get the following error:

ArgumentException: Getting control 18's position in a group with only 18 controls when doing repaint
Aborting
UnityEngine.GUILayoutGroup.GetNext () (at C:/buildslave/unity/build/Modules/IMGUI/LayoutGroup.cs:117)
UnityEngine.GUILayoutUtility.DoGetRect (System.Single minWidth, System.Single maxWidth, System.Single minHeight, System.Single maxHeight, UnityEngine.GUIStyle style, UnityEngine.GUILayoutOption[] options) (at C:/buildslave/unity/build/Modules/IMGUI/GUILayoutUtility.cs:461)

image

Assets/MixedRealityToolkit.Tests/PlayModeTests/PressableButtonTests.cs
thalbern: I suppose it could be - the reason I added it as a runtime test was that we had some customers scaling their buttons during runtime, so i wanted to simulate that scenario - in the end it shouldn't matter anymore as the calculations are the same now

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
thalbern: that was part of the ported code. you're right we can remove that - we can always look it up in git history

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: i need them in the inspector class as well

thalbern: > This looks really promising! I did have some questions in the code, and also I foung a few bugs.
> 
> First, the blue push buttons do not move correctly -- they move in Y space when I pulled the change.
> 
> image
> 
> Also, to demonstrate that buttons can work at multiple scales now I added two buttons in PressableButtonExample scene with two scales. I think it would be good to add that so that people can try it out.
> 
> When adding that, I noticed another bug: when I scaled my HolographicButton to be size 0.1, the planes were still rendering quite big:
> 
> image
> 
> The buttons are also not actually pressing. Here is the package for you to test:
> [pressablebuttonexample.zip](https://github.com/microsoft/MixedRealityToolkit-Unity/files/3209978/pressablebuttonexample.zip)

thanks Julia!
I'll look into the push issues right away.

The distance planes only scale in push direction - the scale wouldn't have any effect on the width / height of the planes. I can look into adding that.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: we want to store the value but we don't want it to be visible in the inspector because there's a conversion button for changing the mode.  I wanted it to be more visible that we're actually converting all the distances on switching the mode, that's why i don't directly expose the property and have that button instead.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
thalbern: that was spotted by daniel :)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: Ah I see the problem with the rotation that gets lost - fixing that now

though i can't see the buttons that don't push in your example scene - for me the scene looks identical to the pressable button example scene that is in the demos and the buttons work fine for me.

thalbern: so i couldn't repro a problem with your scene but i found another problem that causes plane distances to be applied incorrectly and i assume that's the problem you were running into - preparing fix now

thalbern: just realized I wasn't seeing your scene but my local button demo scene. for some reason i can't import your scene?!

julenka: Hm, that's weird. All I did was import HolographicButton prefab (which is quite large), then add a pressable button component to it. Then I scaled that to be 0.1 scale. That should repro the issue.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: we are exposing IsTouching . there's a public get for that just above the IsPressing

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: that is currently determined in UpdatePressedState which is also invoking the ButtonPressed and Released signals. do we want to move all that into the getter?

thalbern: I'm not a big fan of getter with side effects

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: I took this over from how this was determined before - we could also just use Vector3.forward instead because that's what nearinteractiontouchable.localforward points to anyways - i'm not sure what the original intent behind this was. I don't want to dig up too many holes at once

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: good point - this is not handled right now

thalbern: we can actually move the content of this function into OnEnable() that should solve the issue - we don't need to do this in Awake()

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: i copied one of the pressable buttons with back plate from the example and scaled it to 0.1  - i can press it though it's very difficult bc it's so hard to catch xD. i will use a scale of 0.3 in the example scene

(tested with the new changes I pushed - rotations and debug visualisation offsets should be fixed by this)

thalbern: image

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: should be fixed

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: actually nevermind - we don't need this at all anymore . i remove it!

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
thalbern: that's caused by hiding/showing ui elements in the OnInspectorGUI method - I've replaced all those checks with disabled groups , so we won't hide the options anymore but instead show them as grayed out (disabled)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
thalbern: just merged in latest changes and the button example scene has been removed - i guess our button examples are part of the hand interaction scene - not sure where to put the buttons there to not make it look messy - will talk to yoon but would be great if we can still get this change in for the current release even without the example

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -122,16 +124,17 @@ private ButtonInfo GatherCurrentInfo()
 
             // All the planes should be drawn within the button cage. But the distance planes are relative to the 
             // initial transform projected onto the box collider.
-            Vector3 initialPositionLocal = transform.InverseTransformPoint(button.InitialPosition);
-
+            Vector3 initialPositionLocal = transform.InverseTransformPoint(button.InitialPosition) ;
```
julenka: nit: extra space before semicolon.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -278,10 +281,49 @@ private float DrawPlaneAndHandle(Vector3[] vertices, Vector3 halfExtents, float
             return distance;
         }
 
+        /// <summary>
+        /// Trigger function for plane distance world to/from local space conversion
+        /// </summary>
+        public void onTriggerPlaneDistanceConversion()
```
julenka: Should this be public method? Is somebody else needing it?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -278,10 +281,49 @@ private float DrawPlaneAndHandle(Vector3[] vertices, Vector3 halfExtents, float
...
+        {
+            useLocalSpaceDistances.boolValue = !useLocalSpaceDistances.boolValue;
```
julenka: nit: first letter of methods normally capitalized.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
thalbern: we're calling it in the inspector class

thalbern: ah ignore me. no it should be private!

cre8ivepark:
I found that the large button's content part starts moving out of the button box when I touch the button and start moving (with ASWD to test the container's RadialView solver)
I see there is a script missing in the container GameObject.
2019-05-24 10_34_45-Unity 2018 3 8f1 Personal - PressableButtonTest unity - MRTK-Public-Microsoft -
2019-05-24 10_43_58-Unity 2018 3 8f1 Personal - PressableButtonTest unity - MRTK-Public-Microsoft -
2019-05-24 10_35_05-Inbox - DongYoon Park@microsoft com - Outlook

cre8ivepark:
Updated the scene while maintaining Huge and Tiny buttons for validation purpose. Renamed the scene as PressableButtonExamples per discussion with Bernadette. Now the scene has both shell-style buttons and custom button examples using PressableButton.cs

2019-05-24 14_07_46-Unity 2018 3 8f1 Personal - PressableButtonExample unity - MRTK-Public-Microsoft

cre8ivepark:
Per customer feedback, I have added examples of 'How to create a button from scratch' to the PressableButtonExample scene and README file.

README Updated:
https://github.com/thalbern/MixedRealityToolkit-Unity/blob/user/bethalha/button_fixes/Documentation/README_Button.md

MRTK_PressableButtonCube0

thalbern:
run CI commands locally (incl docfx) - all checks passed

Shubham7694:
Any idea when these fixes are going to be released?

thalbern:
@Shubham7694  the fixes are available in mrtk_development branch and in the upcoming release RC2 which is currently out for review and should be merged soon https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4834

Issue #4510, opened 2019-05-22
[Cursor] No cursor displayed in HoloLens 1 (mrtk_development)


cre8ivepark:
## Describe the bug
No cursor displayed in HoloLens 1. (for all gesture states idle, ready, air-tap)
The cursor only briefly shows up for the speech command "Select".

## To reproduce
1. Clone mrtk_development branch
2. Build and deploy HandInteractionExample scene

## Expected behavior
Proper cursor display

- Unity Version 2018.3.9f1
- MRTK Version mrtk_development branch (as of 5/21/2019)

## Target Platform (please complete the following information)
- HoloLens

## Additional context
This could be related to #4720

wiwei:
@julenka

julenka:
Looking at this now. It reproduces in editor, you just need to set the hand simulation mode to "Gestures"

![image](https://user-images.githubusercontent.com/168492/58200511-c93a9680-7c87-11e9-9294-8440de05e6c7.png)

PR #4529, opened 2019-05-22
Fix cursor not displaying on HoloLens 1

julenka:
## Overview
Fixes: #4510 while maintaining ability to show cursor when articulated hands are used in GGV mode.

Also add unit tests to cover HoloLens 1 and articulated hands in GGV.


julenka:
Yes that’s a separate issue.

Issue #4531, opened 2019-05-22
Hand Constained UI


Cameron-Micka:
## Describe the problem

Many users have requested guidance and examples of UI constrained to the articulated hands. This is an area which could include a lot of caveats and design guidance.

## Describe the solution you'd like

It would be great if the MRTK contained behaviors and examples to aide in development of hand constrained UI.

## Additional context

Many of the requests have come from partners and HoloLens 2 hacks.

cre8ivepark:
Merged into 2.0.0.rc2_stabilization

PR #4532, opened 2019-05-22
Hand Constrained Solver (useful for Hand UI)

Cameron-Micka:
## Overview
These changes introduce a new type of solver called "HandConstraint." The HandConstraint provides a solver that constrains the tracked object to a region safe for hand constrained content. (Such as hand UI/menus.) Safe regions are considered areas that don't intersect with the hand.

HandUIExample Scene:

handmenu

A derived class of HandConstraint called "HandConstraintPalmUp" is also included to demonstrate a common behavior of activating the solver tracked object when the palm is facing the user. For example use of this behavior please see the HandUIExample scene.

image

Finally, a new utility class called "HandBounds" has been added to help classes access the AABB bounds of the articulated hands. 

## Changes
- Fixes:  https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4531

## Verification
Please check out the HandUIExample scene to verify example use of the solver.


danielescudero:
@Cameron-Micka lets chat :)

Assets/MixedRealityToolkit/Providers/Hands/HandBounds.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.Collections.Generic;
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Utility behavior to access the axis aligned bounds of IMixedRealityHands.
+    /// </summary>
+    public class HandBounds : MonoBehaviour, IMixedRealitySourceStateHandler, IMixedRealityHandJointHandler
```
Railboy: After this PR is merged we should talk about incorporating this into IMixedRealityHands.

julenka:
The name HandUIExample made me think "ui that you interact with using your hands", not "hand locked menu". Perhaps a different name would be more clear, like "HandBasedMenuExample"?

julenka:
Things I especially liked about this change:

- Exposes properties to things can be accessed from code
- It is short, just a few files, makes it much easier to review
- Provides a .gif of the resulting behavior

Assets/MixedRealityToolkit/Providers/Hands/HandBounds.cs
Cameron-Micka: I like that idea! Since you really only need one instance of this.

Cameron-Micka:
> The name HandUIExample made me think "ui that you interact with using your hands", not "hand locked menu". Perhaps a different name would be more clear, like "HandBasedMenuExample"?

I like that idea! It is more descriptive. I will make that change. 😊 Thank you for reviewing!

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/HandConstraintPalmUp.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
julenka: @Cameron-Micka following our principle of "if it's in core SDK then it should be documented and tested", since this doesn't have documentation or tests and I'm not sure you have time to add those, could you please put this in MRTK.SDK/Experimental instead?

Something like:

MRTK.SDK/Experimental/Solvers/HandConstratintPalmUp.cs

Additionally, we might want to move the example scene to 

MRTK.Examples/Experimental/HandTracking/Scenes/HandUIExample.unity

Cameron-Micka:
@julenka Sounds good! I've updated the directory and namespace to include "Experimental." Please let me know if anything else needs to be updated to merge this in.

julenka:
The only other suggestion from Eric was to add "Experimental" to the class name so that people searching for the code in components would know that the class is experimental. I'm on the fence about this. We could always change it later, since experimental code should not follow the 'no breaking changes rule'.

Cameron-Micka:
> The only other suggestion from Eric was to add "Experimental" to the class name so that people searching for the code in components would know that the class is experimental. I'm on the fence about this. We could always change it later, since experimental code should not follow the 'no breaking changes rule'.

I feel like the namespace emphasizes the “experimental-ness” enough. But, I willing to change the class name if needed. I’m not a fan of long class names (especially prefixes since you have to type more to get intellisense to bring up what you want).

julenka:
Cameron, there was a good suggestion to communicate somewhere in the inspector that this is an experimental feature. Something like "This is an experimental feature."
Then, a tooltip / detail view could read: "Some things the MRTK works on appear to have a lot of initial value even if we haven’t fully fleshed out the details. For these types of features, we want the community to get a chance to see them and get value out of them early. Because they are early in the cycle, we label them as experimental to indicate that they are still evolving, and subject to change over time."

Cameron-Micka:
> Cameron, there was a good suggestion to communicate somewhere in the inspector that this is an experimental feature. Something like "This is an experimental feature."
> Then, a tooltip / detail view could read: "Some things the MRTK works on appear to have a lot of initial value even if we haven’t fully fleshed out the details. For these types of features, we want the community to get a chance to see them and get value out of them early. Because they are early in the cycle, we label them as experimental to indicate that they are still evolving, and subject to change over time."

Sounds good! I'll add an attribute and tag the class as experimental.

Cameron-Micka:
@julenka I added an [Experimental] tag decorator drawer so that developers in the futre can easily tag implmentation as experimental without writing custom inspectors. You can see how it looks below:

image

Cameron-Micka:
We would love to get this in for RC2, so please let me know if anything else should be addressed.

julenka:
Thanks Cameron! I've completed this into the rc2 stabilization branch.

Cameron-Micka:
Thank you Julia!

PR #4536, opened 2019-05-23
Added a C# event for handling pointer events to the input system

### LINKED ISSUE #4535: Support pointer event handlers that are not scripts
## Describe the problem

Trying to implement a bit of logic that runs as part of the focus provider I bumped into the issue that I couldn't listen for pointer down events without implementing the `IMixedRealityPointerHandler` interface and registering as an event system listener. You can only register game objects as event system listeners, so effectively you can only handle pointer events from a script, which focus provider is not.

## Describe the solution you'd like

I'd like to be able to use the standard C# event+delegate mechanism for consuming pointer events.

luis-valverde-ms:
This is so we can have global pointer event listeners that are not scripts. This is required for the upcoming PR addressing #3957 .

Fixes #4535 .

### DIFF
--- a/Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerHandler.cs
+++ b/Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerHandler.cs
@@ -36,7 +36,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     }
 
     /// <summary>
-    /// Type of pointer event. Matches the methods defined in <see cref="IMixedRealityPointerHandler"/>
+    /// Type of pointer event. Matches the methods defined in <see cref="IMixedRealityPointerHandler"/>.
     /// </summary>
     public enum PointerEventType
     {
@@ -47,7 +47,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     }
 
     /// <summary>
-    /// Delegate type for handling pointer events. Subscribe via <see cref="IMixedRealityInputSystem.PointerEvent"/>
+    /// Delegate type for handling pointer events. Subscribe via <see cref="IMixedRealityInputSystem.PointerEvent"/>.
     /// </summary>
     public delegate void PointerHandler(MixedRealityPointerEventData eventData, PointerEventType eventType);
 }
--- a/Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerHandler.cs
+++ b/Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerHandler.cs
@@ -34,4 +34,20 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// <param name="eventData"></param>
         void OnPointerClicked(MixedRealityPointerEventData eventData);
     }
+
+    /// <summary>
+    /// Type of pointer event. Matches the methods defined in <see cref="IMixedRealityPointerHandler"/>
+    /// </summary>
+    public enum PointerEventType
+    {
+        Down,
+        Up,
+        Dragged,
+        Clicked
+    }
+
+    /// <summary>
+    /// Delegate type for handling pointer events. Subscribe via <see cref="IMixedRealityInputSystem.PointerEvent"/>
+    /// </summary>
+    public delegate void PointerHandler(MixedRealityPointerEventData eventData, PointerEventType eventType);
 }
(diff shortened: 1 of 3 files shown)


keveleigh:
This feels potentially confusing. All our other events are handled by Unity, but for one specific type we have a duplicate C# event. Questions could come up about why we don't have C# events for everything.

luis-valverde-ms:
> This feels potentially confusing. All our other events are handled by Unity, but for one specific type we have a duplicate C# event. Questions could come up about why we don't have C# events for everything.

Any suggestions on alternative ways of addressing the issue this PR deals with or on avoiding the potential confusion?

StephenHodgson:
There's also the other question of how events are consumed.

using the event data we have currently I suppose we can raise the event, but there's still a question of who gets the events first?

- The C# objects or the scene objects?
- Are the C# objects technically global listeners? (As they could get and consume input before scene elements)
- Lots more questions here as well.

luis-valverde-ms:
> using the event data we have currently I suppose we can raise the event, but there's still a question of who gets the events first?
> 
> The C# objects or the scene objects?
> Are the C# objects technically global listeners? (As they could get and consume input before scene elements)

I can answer you those two: first the scene objects and yes, C# objects are global listeners. We can document that better, I don't see it as an obstacle for this making it in.

> Lots more questions here as well.

I can only answer the questions I know about.

> One easy work around is having a global listener object that just reacts to input, but that's not ideal either.

It is indeed far from ideal.

Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerHandler.cs
```diff
@@ -34,4 +34,20 @@ public interface IMixedRealityPointerHandler : IEventSystemHandler
...
+        Down,
+        Up,
+        Dragged,
+        Clicked
+    }
+
+    /// <summary>
+    /// Delegate type for handling pointer events. Subscribe via <see cref="IMixedRealityInputSystem.PointerEvent"/>.
+    /// </summary>
+    public delegate void PointerHandler(MixedRealityPointerEventData eventData, PointerEventType eventType);
```
wiwei: Did we ever look into somehow keeping just the single interface (i.e. the IMixedRealityPointerHandler) to also be the way that we get these non-Unity UI events? i.e. instead of storing a list of just GameObjects, potentially a list of IEventSystemHandler?

The reason I ask is it seems odd to have duplicate definitions of both a PointerEventType enum and an interface. This will probably set a precedent going forward and if the goal is to make this the way we really wanna split Unity events + c# events and have agreement then maybe it's fine, but if not I'm curious if there was some way to avoid this new concept...

Otherwise, we could always go the route of the default pointer mediator (which is fed information directly from the focus provider), which is also not a necessarily great design (in that it depends on a specific caller to give them the information, rather than generally requesting it as is done here).

Anyways, I guess my thoughts here are:

1) What your thoughts are on being able to keep some consistent interface (i.e. storing gameobjects + raw interfaces that aren't game objects, so that the null checking will work without having to cast to game objects).
2) If that doesn't work, thoughts on following the default mediator approach.
3) If that doesn't work, having it written down somewhere what the exact input dispatch flow is (i.e. so we can point to that in the future when we're creating new things in the future)

I would probably still like to hear @keveleigh's thoughts here - he has a lot more history than I do with this area.

Railboy:
I really don't like this solution but I can't think of a better alternative. The event system has backed us into a weird corner.

>This feels potentially confusing. 

@keveleigh I think as long as we drown users in documentation we'll be okay. It's only confusing when you don't get the motivation.

Railboy:
@luis-valverde-ms Just had a thought. How about creating a global listener object that's hidden in the scene heirarchy, a la the AsyncCoroutineRunner? That object could subscribe via unity's event system, then relay events via C# delegates, which you can access via a singleton... and this suggestion is so awful and gross that I regret even typing it up. Never mind, please don't do this.

Assets/MixedRealityToolkit/Interfaces/InputSystem/Handlers/IMixedRealityPointerHandler.cs
luis-valverde-ms: Thanks for that feedback! I've created a new PR for an alternative approach where we have methods in the input system to subscribe and unsubscribe pointer handlers as global listeners: #4605

julenka:
It looks like this PR hasn't had progress in a couple days, do we have a clear set of next steps? @luis-valverde-ms it looks like you have two different approaches to add ability for non-game object to listen to pointer events. This PR and then #4605. Should we close this PR and then complete #4605 instead?

luis-valverde-ms:
Closed at @julenka 's suggestion as it seems we will go for the option in #4605

PR #4538, opened 2019-05-23
Introduced the primary pointer selector

luis-valverde-ms:
## Overview

Created a new interface, `IMixedRealityPrimaryPointerSelector`, that is used by the focus provider to select the primary pointer every frame. The specific selector implementation to use is configurable via the pointer profile:

PointerProfile

The default selector chooses the primary pointer among all interaction enabled ones using the following rules in order:
1. Currently pressed pointer that has been pressed for the longest
2. Pointer that was released most recently
3. Pointer that became interaction enabled most recently

The focus provider exposes the current primary pointer via a read-only property and a C# event to notify of changes.

All this can be seen in action in the PrimaryPointerExample scene.

## Changes
- Fixes: #3957  .


Assets/MixedRealityToolkit.Examples/Demos/HandTracking/CustomProfiles/HandInteractionAllExampleMixedRealityToolkitConfigurationProfile.asset
```diff
@@ -1,50 +0,0 @@
-%YAML 1.1
```
julenka: Why are you deleting this file?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// </summary>
+    public class DefaultPrimaryPointerSelector : IMixedRealityPrimaryPointerSelector
+    {
+        private readonly Dictionary<IMixedRealityPointer, PointerInfo> pointerInfos = new Dictionary<IMixedRealityPointer, PointerInfo>();
+
+        public void Initialize()
+        {
+            if (MixedRealityToolkit.InputSystem != null)
+            {
+                MixedRealityToolkit.InputSystem.PointerEvent += OnPointerEvent;
```
julenka: Oh this is an interesting pattern! I've never seen it used in MRTK. I like it. When did we add ability to directly listen to pointer events?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
```diff
@@ -0,0 +1,140 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Default primary pointer selector. The primary pointer is chosen among all interaction enabled ones using the following rules in order:
+    ///   1. Currently pressed pointer that has been pressed for the longest
```
julenka: Seems like since there are known inputs and outputs for this behavior it could be good candidate for unit tests.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityPointerProfile.cs
```diff
@@ -92,5 +92,19 @@ public SystemType PointerMediator
...
             get { return pointerMediator; }
         }
+
+        [SerializeField]
+        [Implements(typeof(IMixedRealityPrimaryPointerSelector), TypeGrouping.ByNamespaceFlat)]
+        [Tooltip("")]
+        private SystemType primaryPointerSelector = null;
+
+        /// <summary>
+        ///
```
julenka: nit: missing comment

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityPointerProfile.cs
```diff
@@ -92,5 +92,19 @@ public SystemType PointerMediator
         {
             get { return pointerMediator; }
         }
+
+        [SerializeField]
+        [Implements(typeof(IMixedRealityPrimaryPointerSelector), TypeGrouping.ByNamespaceFlat)]
+        [Tooltip("")]
+        private SystemType primaryPointerSelector = null;
```
julenka: It would be nice to document what this is, either in the tooltip or in the README.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityPrimaryPointerSelector.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Interface used by focus providers to select the primary pointer.
```
julenka: It would be good to define the primary pointer here.

Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
```diff
@@ -0,0 +1,49 @@
...
+    private void OnEnable()
+    {
+        MixedRealityToolkit.InputSystem?.FocusProvider?.SubscribeToPrimaryPointerChanged(OnPrimaryPointerChanged, true);
+    }
+
+    private void OnPrimaryPointerChanged(IMixedRealityPointer oldPointer, IMixedRealityPointer newPointer)
+    {
+        if (CursorHighlight != null)
+        {
+            if (newPointer != null)
```
provencher: A null check is insufficient to detect if the pointer is null. You need to cast it to a monobehavior first, or else it might not be null, but throw a missing reference exception when you try using it

Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                Transform parentTransform = newPointer.BaseCursor?.GameObjectReference?.transform;
+
+                // If there's no cursor try using the controller pointer transform instead
+                if (parentTransform == null)
+                {
+                    var controllerPointer = newPointer as BaseControllerPointer;
+                    parentTransform = controllerPointer?.transform;
```
provencher: ? checks don't work on monobehaviors

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
```diff
@@ -0,0 +1,140 @@
...
+            IMixedRealityPointer primaryPointer = null;
+            PointerInfo primaryInfo = null;
+
+            foreach (var keyValue in pointerInfos)
+            {
+                var pointer = keyValue.Key;
+                var info = keyValue.Value;
+                info.Update(pointer);
+
+                if (info.IsInteractionEnabled &&
```
provencher: It would be fantastic to be able to filter out pointers that are not near interaction pointers.
In most cases, the primary pointer is for interacting with UI, and so it must be a far interaction pointer, especially when driving legacy UIs

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/CustomProfiles/HandInteractionAllExampleMixedRealityToolkitConfigurationProfile.asset
luis-valverde-ms: Because I'm silly. Thanks!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
luis-valverde-ms: That would be PR #4536 , which is not in yet.

Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
luis-valverde-ms: A null check should be fine here as we know we're only dealing with pointers that are valid, i.e. registered with the focus provider.

Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
luis-valverde-ms: Same point as above.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
luis-valverde-ms: Ideally, I would like to have a profile for the default selector with common configuration options. I have very limited time to work on this though, so for the time being I'm going to make DefaultPrimaryPointerSelector.Update virtual so you can implement that behavior easily yourself.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityPrimaryPointerSelector.cs
luis-valverde-ms: Added a bit more detail. Does that look good?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
luis-valverde-ms: It would be nice but I think we would get more value out of a play mode test. I'm quite pressed on time though so I'll leave that out of this PR if you don't mind.

provencher:
@luis-valverde-ms I'm looking through the code, and it seems like the metric for selecting a pointer is entirely based on which one was used to trigger an interaction - IE pointerdown/up

I don't see anything in here that would select a pointer automatically before any interaction is triggered, and I don't see anything in here that will recover a pointer after it has been destroyed, though I may have missed something.

By that last point, I mean that the pointer is the one that interacted with something, but then loses tracking, IE hand moves out of view of the tracking cameras, and then comes back. Will that hand still be the primary pointer?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
provencher: This is a fine solution.

Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
provencher: Interesting. I guess since you're not caching pointers for more than a frame anyway its not a big issue. I don't love the overhead of adding an update loop that needs to process all the pointers every single update. This is where preventing pointers from being destroyed could be very helpful in keeping complexity per update down.

luis-valverde-ms:
> @luis-valverde-ms I'm looking through the code, and it seems like the metric for selecting a pointer is entirely based on which one was used to trigger an interaction - IE pointerdown/up

ReleasedTimestamp doubles as interaction enabled timestamp, that should enable selecting the pointer that "appeared" last. The best way to see it is to play with the PrimaryPointerExample.

Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
luis-valverde-ms: We're caching pointers for more than one frame. We add them to the pointerInfo dictionary when they're registered with the focus provider and remove them when they're unregistered.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityFocusProvider.cs
```diff
@@ -76,5 +86,19 @@ public interface IMixedRealityFocusProvider : IMixedRealitySourceStateHandler, I
         /// <typeparam name="T">The type of pointers to request. Use IMixedRealityPointer to access all pointers.</typeparam>
         /// <returns></returns>
         IEnumerable<T> GetPointers<T>() where T : class, IMixedRealityPointer;
+
+        /// <summary>
+        /// Subscribes to primary pointer changes.
+        /// </summary>
+        /// <param name="handler">Handler to be called when the primary pointer changes</param>
+        /// <param name="invokeHandlerWithCurrentPointer">Whether to invoke the handler with the current primary pointer
```
wiwei: >invoke [](start = 69, length = 6)

Nit: Maybe add the clarification that this is sychronously invoking the handler (instead of just kicking off async work to immediately ping you back). It's possible that the "before subscribing" part is strong enough to describe this, but providing a little more detail here about the callback mechanics probably wouldn't hurt

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityFocusProvider.cs
```diff
@@ -7,6 +7,11 @@
 
 namespace Microsoft.MixedReality.Toolkit.Input
 {
+    /// <summary>
+    /// Delegate type used to handle primary pointer changes
+    /// </summary>
```
wiwei: Thoughts about adding some remarks about possible transitions include null->non-null, pointerA to pointer B, non-null->null? Or is this just generally understood by folks using this?

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityPrimaryPointerSelector.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        /// <summary>
+        /// Unregisters a pointer with the selector.
+        /// </summary>
+        void UnregisterPointer(IMixedRealityPointer pointer);
+
+        /// <summary>
+        /// Called from the focus provider after updating pointers to obtain the new primary pointer.
+        /// </summary>
+        ///<returns>The new primary pointer or null if none.</returns>
```
wiwei: >< [](start = 11, length = 1)

Nit: Space before <returns>

luis-valverde-ms:
Needs updating with the changes in #4716

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            pointerInfos.Remove(pointer);
+        }
+
+        virtual public IMixedRealityPointer Update()
+        {
+            IMixedRealityPointer primaryPointer = null;
+            PointerInfo primaryInfo = null;
+
+            foreach (var keyValue in pointerInfos)
```
wiwei: >foreach (var keyValue in pointerInfos) [](start = 12, length = 38)

Similar comment that I had in another review, just be wary that if this is called in every single Update() loop, that this isn't spawning additional GC garbage. It would be worth doing a quick profile (even in player) to see if this occurs, and if so, it might be worth snapshotting the list on register/unregister to eat the cost there instead of on each update.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                var pointer = keyValue.Key;
+                var info = keyValue.Value;
+                info.Update(pointer);
+
+                if (info.IsInteractionEnabled && 
+                    (primaryInfo == null || 
+                    (info.IsPressed && (!primaryInfo.IsPressed || info.PressedTimestamp < primaryInfo.PressedTimestamp)) ||
+                    (!primaryInfo.IsPressed && info.ReleasedTimestamp > primaryInfo.ReleasedTimestamp)))
+                {
```
wiwei: It might be interesting to add tests for this - these things are somewhat hard to validate manually, and changing this in the future might be scary because validating manually is so hard.

However, manually newing this up in a test and then driving this with calls to OnPointerDown/OnPointerUp (or other mock pointer objects) should be enough to have some basic level of testing for this.

wiwei:
Change seems good to me, just the request to have a test for the state machine of the default primary pointer selector

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
luis-valverde-ms: Added!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
luis-valverde-ms: Good that you bring that up because we need to dispel that myth:

image

Let's go over what happens in that [`foreach`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/foreach-in):

1. [`Dictionary<TKey,TValue>.GetEnumerator`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.getenumerator?view=netframework-4.8) is called, which returns a [`Dictionary<TKey,TValue>.Enumerator`](Dictionary<TKey,TValue>.Enumerator) struct, so no allocation there.
1. [`Dictionary<TKey,TValue>.Enumerator.MoveNext`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.enumerator.movenext?view=netframework-4.8) is called to move to the first element.
1. [`Dictionary<TKey,TValue>.Enumerator.Current`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.enumerator.current?view=netframework-4.8) is called which returns a [`KeyValuePair<TKey,TValue>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.keyvaluepair-2?view=netframework-4.8) struct, again no allocation there.
1. Goto #2 until MoveNext returns false.

Does that sounds alright to you? Where did you think those allocations could be coming from?

wiwei: Awesome, thanks for doing this :)

Yes, I wasn't sure about this because when I was profiling some of the default pointer mediator stuff (which also did foreach over the collection) I found that it did have non-zero GC Alloc on basically every frame.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
wiwei: This is great, thanks!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultPrimaryPointerSelector.cs
luis-valverde-ms: Are you positive those were coming from a foreach? I'd like to have a look if so. At the moment I see DefaultPointerMediator.UpdatePointers() making extensive use of foreach but zero allocations.

Issue #4369, opened 2019-05-15
Nested Prefabs work poorly with MRTK custom inspectors


wiwei:
The new Nested Prefabs have extensive inspector functionality to allow you to apply selective updates to prefabs.
If working directly in a scene on prefabs, you can apply changes piecemeal to the prefabs.

However the custom inspector is nonfunctional or just decently frustrating to work with when using prefabs:

Examples:
The blue line on the left here indicates that there is a pending change that can be applied.
Normally in the inspector you can right click and individually apply changes.
For Unity Events the hitbox is small to nonexistent. I have succeeded occasionally, but I need to apply ~20 changes and it takes 4-12 attempts to hit the TINY hitbox for applying the change. This particular problem could be an issue with out-of-the-box UnityEvents, I haven't looked into whether that is the case.

![19b9591f-7992-42b4-9367-28c14a879fe2](https://user-images.githubusercontent.com/5840182/57795241-4dc26d80-76fa-11e9-9fa5-d3541af9a14a.png)
![441c9683-09bd-498e-a3c2-fd48d17d76b5](https://user-images.githubusercontent.com/5840182/57795243-4f8c3100-76fa-11e9-9c94-5337343e3d4d.png)

Some fields do not even indicate that there are pending changes. See the following two images for an example:
![786fd616-d8b3-480a-b7da-e48251571d02](https://user-images.githubusercontent.com/5840182/57795245-5155f480-76fa-11e9-9023-9c29e251541c.png)

The first one has the pending change. This is a straight hostile nonfunctionality that will create significant interaction errors.
![c6cbf497-9399-41bd-b14d-648b5aa9b959](https://user-images.githubusercontent.com/5840182/57795251-54e97b80-76fa-11e9-9e77-194619a7bfab.png)

Finally, many fields don't even actively display and allow you to apply those prefab changes. If you put the Inspector in Debug Mode, you can see the extent of changes (this is from our custom inspectors which have 0 blue fields unapplied)

The pervasiveness to which this is a problem heavily encourages a developer to never use the inspector in scene, or to constantly switch to debug inspectors to check to make sure they understand the state of their objects (which is a severe problem)

This is likely a widespread issue across ALL custom inspectors and we need a solid solution for our developers.f their objects (which is a severe problem)

[...]

PR #4548, opened 2019-05-23
Fixed nested prefabs not showing pending changes for Interactable.

killerantz:
## Overview
Updated inspector EditorGUILayout fields to use PropertyFields so that pending changes will be shown in prefabs.

## Changes
- Fixes: #4369 .
Nested Prefabs work poorly with MRTK custom inspectors

## Verification
When Interactable is part of a prefab, the inspector should show a blue vertical line to the left of any property and bold text when it has been edited and pending an override. This would mainly cover the main portion above Profiles and the Events section.

- Inspectors should work and look as they did before.
- No data should be lost.
- properties inside of themes would not show a blue mark or bold text when edited, they are ScriptableObjects and their properties are not stored in the prefab.


Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -668,6 +667,15 @@ protected void ChangeEvent(int[] indexArray, SerializedProperty prop = null)
             }
         }
 
+        protected void DrawDrownDownProperty(Rect position, SerializedProperty prop, string[] options, GUIContent label)
```
keveleigh: Not sure if this is what it's supposed to be, but was my best guess :)
```suggestion
        protected void DrawDropDownProperty(Rect position, SerializedProperty prop, string[] options, GUIContent label)
```

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -152,32 +154,30 @@ public virtual void RenderCustomInspector()
             }
             else
             {
-                int newActionId = EditorGUILayout.Popup("Input Actions", actionId.intValue, actionOptions);
-                if (newActionId != actionId.intValue)
-                {
-                    actionId.intValue = newActionId;
-                }
+                position = EditorGUILayout.GetControlRect();
+                DrawDrownDownProperty(position, actionId, actionOptions, new GUIContent("Input Actions", "The input action filter"));
```
keveleigh:
```suggestion
                DrawDropDownProperty(position, actionId, actionOptions, new GUIContent("Input Actions", "The input action filter"));
```

Issue #4541, opened 2019-05-23
Infinite loop caused by InscribedRectangle ctor


kircher1:
## To reproduce

This was reported by a customer, and means the following configuration occured:
1) Tracking space is room scale
2) UnityBoundary.TryGetGeometry returns true
3) The boundary point array is empty

When this happens, the do-while loop inside of the InscribedRectangle ctor never returns.

It should be straight forward enough to add some some input checking to prevent getting into this state. 

- Unity Version: 2018.3
- MRTK Version: RC1 Refresh

keveleigh:
Fixed via #4550

PR #4550, opened 2019-05-23
Fix potential infinite loop caused by InscribedRectangle

keveleigh:
## Overview
In some cases, `UnityBoundary.TryGetGeometry` returns true but returns an empty list of boundary points.
In this case, we shouldn't try to create an inscribed rectangle.

## Changes
- Fixes: #4541


kircher1:
Additionally, throwing an ArgumentException from the InscribedRectangle ctor when the input list is empty would add some extra safety from an infinite loop occurring again, if this class gets used elsewhere. (As far as I can tell, the only client of InscribedRectangle is the boundary system.)

keveleigh:
>Additionally, throwing an ArgumentException from the InscribedRectangle ctor when the input list is empty would add some extra safety

We've tended to log Unity errors instead of throwing Exceptions, so I added an error log in `InscribedRectangle`'s constructor.

Issue #4551, opened 2019-05-24
InteractableColorTheme does not support TextMeshPro


killerantz:
## Describe the bug
InteractableColorTheme supports coloring UI Text and TextMesh components, but does not include support for coloring TextMeshPro objects.

## To reproduce

1. Create an empty with a collider and add the Interactable component.
2. Create a TextMeshPro object
3. Add the TextMeshPro object as a profile target on the Interactable
4. Add a theme that effects color, like InteractableColorTheme or ScaleOffsetColorTheme.
5. Observe that the theme displays the shader property dropdown "_Color", but TextMeshPro does not have a _Color property. It will cause an error when running and not update any colors.

## Expected behavior
The color themes should work on Text, TextMesh, TextMeshPro, TextMeshProUGUI, as well as renderers with materials.

## Screenshots

![image](https://user-images.githubusercontent.com/22085553/58294168-780cce80-7d7d-11e9-82f8-7eb84ca60c3a.png)

cre8ivepark:
MERGED

PR #4552, opened 2019-05-24
Added TextMeshPro support to the InteractableColorTheme

killerantz:
## Overview
The InteractableColorTheme will color meshRenderers, as well as Text objects. Previously it only set the color fields of UI Text and TextMesh objects.

## Changes
- Fixes: #4551 .

## Verification
- Different Text objects will be effected by color changes from the InteractableColorTheme and ScaleOffsetColorTheme.
- Renderers are the fall back option.
- Theme does not show shader property drop down when the target is a text base object.
- Only does the check for each text object once, then will use a delegate to apply colors is a text object exists.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableColorTheme.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableColorTheme.cs
@@ -9,12 +9,17 @@ using UnityEngine.UI;
 
 namespace Microsoft.MixedReality.Toolkit.UI
 {
+    /// <summary>
+    /// A color theme that can set colors on renderers or common text objects
+    /// This theme will try to set color on text objects first, if none can be found,
+    /// then we fall back to renderer color setting using the parent class.
+    /// </summary>
     public class InteractableColorTheme : InteractableShaderTheme
     {
         // caching methods to set and get colors from text object
-        // this will avoid 4 if statements for every set or get
-        private delegate bool SetColorOnText(Color color);
-        private delegate bool GetColorFromText(out Color color);
+        // this will avoid 4 if statements for every set or get - also during animation
+        private delegate bool SetColorOnText(Color color, InteractableThemeProperty property, int index, float percentage);
+        private delegate bool GetColorFromText(InteractableThemeProperty property, out Color color);
         private SetColorOnText SetColorValue = null;
         private GetColorFromText GetColorValue = null;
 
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableColorTheme.cs
```diff
@@ -58,20 +88,211 @@ public override void SetValue(InteractableThemeProperty property, int index, flo
...
+            success = false;
+            return colour;
+        }
+
+        /// <summary>
+        /// Get color from TextMeshProUGUI
+        /// </summary>
+        /// <param name="success"></param>
+        /// <returns></returns>
+        protected Color GetTextMeshProUGUIColor(out bool success)
```
Railboy: Super-duper nit: I've never seen this pattern - typically you'd return success value and use an out for Color.

killerantz: Good point, not sure why it's reversed.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableColorTheme.cs
```diff
@@ -58,20 +83,213 @@ public override void SetValue(InteractableThemeProperty property, int index, flo
...
+            base.SetValue(property, index, percentage);
+
+        }
+
+        /// <summary>
+        /// Get color on UI Text
+        /// </summary>
+        /// <param name="success"></param>
+        /// <returns></returns>
+        protected bool GetTextColor(out Color color)
```
julenka: nit: might want to use the name TryGetTextColor and document what color gets set to if you return false, and what returning false means.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/InteractableColorTheme.cs
```diff
@@ -58,20 +83,213 @@ public override void SetValue(InteractableThemeProperty property, int index, flo
...
+            color = colour;
+            return false;
+        }
+
+        /// <summary>
+        /// Get color from TextMeshPro
+        /// </summary>
+        /// <param name="success"></param>
+        /// <returns></returns>
+        protected bool GetTextMeshProColor(out Color color)
```
julenka: Similarly here, might want to use the name TryGetTextMeshProColor and specify what happens in case of failure. And below.

PR #4588, opened 2019-05-28
Removing build.ps1 step from our pipelines

MenelvagorMilsom:
This step is only creating .unitypackages now. This is no longer necessary for all of our builds. We may need this again for our internal packaging/signing pipeline, but until then we are removing this step.

This will fix an error we are currently seeing on CI.


wiwei:
I'm going to break the usual rule of letting the author merge things in, because this fixes our PR build pipeline and we need it.

Issue #4151, opened 2019-05-01
MRTK v2 Bounding Box handles not displayed correctly with scaled objects


meulta:
## Describe the bug

Bounding Box handles are not displayed at the right position when the object it is applied to has a different scale for at least one axis.

![image](https://user-images.githubusercontent.com/1480170/57049302-4a27e480-6c2c-11e9-9817-bfad60e784b8.png)

Watch the video here: [link](https://1drv.ms/v/s!Ajn1dnU1WwYxu6wjYjt3LsCEQCZmmg)
Note: in that video, I illustrate the normal behavior with an even scaling and then I change the scale in the properties so you can see both behaviors, including the wrong one.

## To reproduce

1. In an empty project: Import MRTK
2. Add any 3D object
3. Add a Bounding Box to the object following [docs](https://docs.microsoft.com/en-us/windows/mixed-reality/mrlearning-base-ch4)
4. Change the scale value in the Transform for the 3D object for ONE axis only (for instance, x)
5. Run the project and move on all axis

You can find an example with that issue here: [https://github.com/meulta/MRTK_BoundingBox](https://github.com/meulta/MRTK_BoundingBox)

(just change the scale of the mountain).

## Expected behavior

Uneven scaling of the object should not impact the Bounding Box handle and they should display properly

## Screenshots

![image](https://user-images.githubusercontent.com/1480170/57049302-4a27e480-6c2c-11e9-9817-bfad60e784b8.png)

## Your Setup (please complete the following information)

- Unity Version 2018.3.12f1
- MRTK Version v2.0 RC1 refresh

## Target Platform (please complete the following information)

- HoloLens 2 (but the issue happens in the player in Unity also)

Railboy:
Editing the object's transform in-editor at runtime while the bounding box is active isn't supported. (Based on your video and scene setup this appears to be what you're doing.) Once the bounding box is active it's the bounding box's job to perform transformations.

We'll make this clearer in the documentation.

Thanks for the detailed bug write-up and example scene.

meulta:
Understand that, but the issue is also happening if you change scale BEFORE running in the editor. The video was just to explain the issue. This is still more than a documentation issue in my opinion.

meulta:
Hello, this is still a bug, can you please keep it opened until it is fixed?

Railboy:
Apologies! Finger slipped on mobile while commenting. We'll take another look and try to repro.

djee-ms:
[...]

PR #4592, opened 2019-05-28
BoundingBox Fix rig when parent of has non uniform scale

julenka:
## Changes
- Fixes: #4151
- Add test to BoundingBox test scene to repro the issue and verify the fix
- Expose method BoundingBox.UpdateRig() developers can use to force the bounding box rig to update.

## Verification

> commands
>
> git fetch origin pull/4592/head:name_of_local_branch
>

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
@@ -705,6 +705,14 @@ namespace Microsoft.MixedReality.Toolkit.UI
             }
         }
 
+        /// <summary>
+        /// Destroys and re-creates the rig around the bounding box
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -705,6 +705,14 @@ public void SetScaleLimits(float min, float max, bool relativeToInitialState = t
             }
         }
 
+        /// <summary>
+        /// Destroys and re-creates the rig around the bounding box
+        /// </summary>
+        public void UpdateRig()
```
luis-valverde-ms: >UpdateRig [](start = 20, length = 9)

For simplicity, I would have just one method instead of having UpdateRig() being a public wrapper around CreateRig().

PR #4595, opened 2019-05-28
Add the "App" network connection box and refactor the CommandService

matthejo:
The primary purpose of this change is to add the "App" connection box in the Compositor scene to provide the network connection to the app providing the content to the compositor.

Part of making this change cleanly involved refactoring the CommandService into a CommandRegistry base class that both SynchronizedStateObserver and HolographicCameraNetworkManager could inherit from. Each one of these connections needs to handle the same command, but in the context of that particular connection. I also fixed what I consider a bug in the implementation of AddCommandHandler where adding the same object as the handler for two different commands did not work (because the object was already in the allHandlers list).

I have not tested out the Android scenario with this change, but I am happy to if given instructions for how to do so.

image


Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
```diff
@@ -11,37 +11,35 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
     /// <summary>
     /// This class observes changes and updates content on a spectator device.
     /// </summary>
-    public class StateSynchronizationObserver : Singleton<StateSynchronizationObserver>
+    public class StateSynchronizationObserver : LocatableDeviceNetworkManager<StateSynchronizationObserver>
```
chrisfromwork: >LocatableDeviceNetworkManager [](start = 48, length = 29)

Is there any instance where a StateSynchronizationBroadcast also needs to report its location?

chrisfromwork:
>                     case "SYNC":

nit: should use SyncCommand definition.

---
Refers to: Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationBroadcaster.cs:131 in b0f1aff. [](commit_id = b0f1aff24b3fbe0dade02b579ee289bf05e12ddf, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
chrisfromwork: >LocatableDeviceNetworkManager [](start = 48, length = 29)

This adds a concept of spatial alignment to these state synchronization classes. It may make more sense for this localization logic to be handled through its own command registry compared to existing by default.

matthejo: I've refactored the "LocatableDevice" logic into a separate class for the next review, so the StateSynchronizationBroadcaster could in theory also have a LocatableDevice associated with it.

matthejo: I created a new base type NetworkManager that StateSynchronizationObserver, StateSynchronizatoinBroadcaster, and HolographicCameraNetworkManager can all inherit from. LocatableDevice is now a separate class that just deals with that aspect.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Compositor/HolographicCameraObserver.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// <summary>
+    /// Component that connects to the HoloLens application on the holographic camera rig for synchronizing camera poses and receiving calibration data.
+    /// </summary>
+    public class HolographicCameraObserver : NetworkManager<HolographicCameraObserver>, ICommandHandler
+    {
+        public const string CameraCommand = "Camera";
+        public const string CalibrationDataCommand = "CalibrationData";
+
+        [SerializeField]
+        private CompositionManager compositionManager = null;
```
chrisfromwork: nit: add summary and tooltip for serialized field.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Compositor/HolographicCameraObserver.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        [SerializeField]
+        private LocatableDeviceObserver appDeviceObserver = null;
```
chrisfromwork: nit: add summary and tooltip for serialized field.

chrisfromwork:
>         private Stopwatch timestampStopwatch;

Right now both the CameraPoseProvider and HolographicCameraBroadcaster take dependencies on this TCPConnectionManager. HolographicCameraBroadcaster is only used for setting up the network connection, which I feel could be done in this class to reduce the number of classes.

Long term, we will likely need a CameraPoseProvider that runs in the editor based on the VR headpose. We may save some work if we keep both classes but have HoloGraphicCameraBroadcaster contain all network functionality. Then we could have the CameraPoseProvider pipe data somewhere else locally for our VR experience.

---
Refers to: Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/HolographicCamera/CameraPoseProvider.cs:34 in ec33bc8. [](commit_id = ec33bc8841114c6ee2474a205d26f1b4ff717c0d, deletion_comment = False)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerAnchor.cs
```diff
@@ -15,15 +15,19 @@
...
 namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView.HolographicCamera
 {
-    [RequireComponent(typeof(WorldAnchorManager))]
-    [RequireComponent(typeof(TCPConnectionManager))]
-    public class HolographicCameraOriginAnchor : MonoBehaviour
+    /// <summary>
+    /// Represents an object that can detect an ArUco marker and persist and restore a WorldAnchor
+    /// to persist the location of that marker across sessions.
+    /// </summary>
+    public class ArUcoMarkerAnchor : MonoBehaviour, ICommandHandler
```
chrisfromwork: >ArUcoMarkerAnchor [](start = 17, length = 17)

Nothing about this is ArUco specific, we should change this to MarkerAnchor so that it can be used with qr codes/HoloLens 2 as well.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/LocatableDeviceBroadcaster.cs
```diff
@@ -18,26 +18,25 @@
...
-    [RequireComponent(typeof(TCPConnectionManager))]
-    [RequireComponent(typeof(HolographicCameraOriginAnchor))]
-    public class HolographicCameraStatusMonitor : MonoBehaviour
+    public class LocatableDeviceBroadcaster : MonoBehaviour
     {
-#if UNITY_WSA
-        TCPConnectionManager tcpConnectionManager = null;
-        HolographicCameraOriginAnchor originAnchor = null;
+        [SerializeField]
+        private TCPConnectionManager connectionManager = null;
```
chrisfromwork: nit: needs tooltip

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/LocatableDeviceBroadcaster.cs
```diff
@@ -18,26 +18,25 @@
...
+
+        [SerializeField]
+        private ArUcoMarkerAnchor originAnchor = null;
```
chrisfromwork: nit: needs tooltip.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/NetworkManager.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
+{
+    public abstract class NetworkManager<TService> : CommandRegistry<TService>, INetworkManager where TService : Singleton<TService>
+    {
+        [SerializeField]
```
chrisfromwork: >SerializeField [](start = 9, length = 14)

nit: needs tooltip

Issue #4600, opened 2019-05-29
[RC2 Stabilization] Pressable Button moves unexpected way when parented by Solver


cre8ivepark:
## Describe the bug
When PressableButton is placed under an object that has Solver (e.g. Radial View Solver), the button's front cage can be separated by the Solver's movement. It seems like while the finger is on the button, Solver’s transform is not applied to the button and makes it broken.

![MRTK_PressableButton_Bug2](https://user-images.githubusercontent.com/13754172/58528941-9a438980-818d-11e9-8b5f-561177851a32.gif)
![MRTK_PressableButton_Bug0](https://user-images.githubusercontent.com/13754172/58528942-9a438980-818d-11e9-8d77-c73a7349fc7f.gif)
![MRTK_PressableButton_Bug1](https://user-images.githubusercontent.com/13754172/58528943-9a438980-818d-11e9-8285-47c507ef86ff.gif)

## To reproduce
Open HandInteractionExample scene
Try the 'Follow Me' button on the slate with direct finger press
Try the 'Pin' button on the toggle profiler/hand mesh/joint panel with direct finger press

## Expected behavior
The front cage should not break

## Branch
**prerelease/2.0.0.rc2_stabilization**

cre8ivepark:
This bug was originally found in the PR #4527 but I couldn't repro it since I have updated the PressableButtonExample scene without the Solver.

thalbern:
fixed with #4604

PR #4604, opened 2019-05-29
fixed moveable button visuals position update while button is moving

thalbern:
## Overview
- we're storing an offset to the parent now instead of the world space position, so we can apply that offset to an up-to-date world space position of the button
- also added proper reset of the moveable button part on disabling of the button

## Changes
- Fixes: #4600  .

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
@@ -201,11 +201,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
         #endregion
 
-        private void Awake()
-        {
-            initialOffsetMovingVisuals = PushSpaceSourceTransform.localPosition;
-        }
-
         private void OnEnable()
         {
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -197,10 +201,14 @@ private Vector3 InitialPosition
 
         #endregion
 
+        private void Awake()
+        {
+            initialOffsetMovingVisuals = PushSpaceSourceTransform.localPosition;
```
luis-valverde-ms: >PushSpaceSourceTransform.localPosition [](start = 41, length = 38)

This transform may come from another game object that has not awaken yet and that may modify its transform on Awake(). To make sure there are no initialization order issues, I think it is better to do this in Start(). There is a good write up about Awake vs Start here: [https://www.tangledrealitystudios.com/development-tips/awake-vs-start-vs-onenable-and-when-to-use-them/](https://www.tangledrealitystudios.com/development-tips/awake-vs-start-vs-onenable-and-when-to-use-them/). The general rule is: things that rely on other game objects are better done in Start().

PR #4605, opened 2019-05-29
Pointer handler subscription

luis-valverde-ms:
## Overview

Added a way to subscribe pointer handlers directly as global listeners instead of subscribing the game object the belong (or not) to. 

I think it should be possible generalize this pattern for other event handler types but I don't have the time to do it.

## Changes
- Fixes: #4535  .


Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -316,18 +321,28 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
...
             Debug.Assert(eventData != null);
-            var baseInputEventData = ExecuteEvents.ValidateEventData<BaseInputEventData>(eventData);
-            DispatchEventToGlobalListeners(baseInputEventData, eventHandler);
+            DispatchEventToGlobalListeners(eventData, eventHandler);
 
-            if (baseInputEventData.used)
+            if (pointerHandlers != null)
+            {
+                eventExecutionDepth++;
+                foreach (var handler in pointerHandlers)
```
wiwei: >foreach [](start = 16, length = 7)

I've avoided using foreach in inner loop stuff before because of past guidance around the speed of for vs foreach, and the fact that foreach generates GC garbage but for does not.

Granted I think that this advice came around from older Unity versions which created significant garbage for each run - I see that there are still some recent guidelines for plain old C# talking about preferring for over foreach, but again because we do IL2CPP it's not clear to me if that actually has any real effect now. So this is probably fine at this point, if we wanted to be paranoid we could do the list thing that BaseEventSystem does.

I don't really have a preference, so feel free to just just ACK this.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -858,6 +873,38 @@ public void RaiseFocusExit(IMixedRealityPointer pointer, GameObject unfocusedObj
...
+        {
+            if (pointerHandlers == null)
+            {
+                pointerHandlers = new HashSet<IMixedRealityPointerHandler>();
+            }
+            // We do this check to guard against changes to the pointer handler collection while we're iterating over it.
+            else if (eventExecutionDepth > 0)
+            {
+                await doneExecutingEvents;
+            }
```
wiwei: If we wanted to be really pendantic here should should probably be its own if, so that changes to the pointerHandlers is also done in a way that is ordered after all event execution. I think the weirdness comes from how we're using doneExecutingEvents and eventExecutionDepth, which were designed to protect a different data structure (i.e. ideally this would be pointerEventExecutionDepth since it doesn't matter if there are other global events firing).

Anyways, just thinking out loud. I think in the end this doesn't matter and I think it's fine as is.

wiwei:
Resolved my other comments, seems fine to me, would be interested to hear @keveleigh's opinion

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
luis-valverde-ms: I don't think `foreach` generates GC here. We can go over it if you want. I think the general guideline should be the opposite: use `foreach` unless you know for a fact the loop is hot and `for` will make a significant difference in performance.

wiwei:
@keveleigh, friendly ping!

luis-valverde-ms:
Closing this as it is superseded by #4716

Issue #4610, opened 2019-05-29
[RC2 Stabilization] Interactable says 'Mixed Reality Toolkit is missing', PressableButtonExamples scene is not working


cre8ivepark:
## Describe the bug

Interactable says 'Mixed Reality Toolkit is missing' even though it is in the Hierarchy.
PressableButtonExamples scene is not working.

![2019-05-29 09_49_29-Unity 2018 3 8f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof](https://user-images.githubusercontent.com/13754172/58575517-17571900-81f7-11e9-88b5-8401d367a8e7.png)

![2019-05-29 09_47_15-Unity 2018 3 8f1 Personal - PressableButtonExample unity - MRTK-Public-Microsoft](https://user-images.githubusercontent.com/13754172/58575459-f393d300-81f6-11e9-9500-efefb2a63178.png)

![2019-05-29 09_47_24-Unity 2018 3 8f1 Personal - PressableButtonExample unity - MRTK-Public-Microsoft](https://user-images.githubusercontent.com/13754172/58575460-f393d300-81f6-11e9-8c1c-90ee0ccdc666.png)

![2019-05-29 10_17_34-Interactable cs - Microsoft Visual Studio](https://user-images.githubusercontent.com/13754172/58577501-728b0a80-81fb-11e9-80d4-2627b48316db.png)

## To reproduce
Open HandInteractionExamples scene or PressableButtonExamples scene
Select PressableButton from the Hierarchy panel. See inspector panel.

## Branch
**prerelease/2.0.0.rc2_stabilization**

julenka:
Taking a look

killerantz:
I saw this when taking the latest from MRTK_development on Friday. I had to restart Unity to get rid of all the errors that flooded the console window. I still get the MRTK is missing message on Interactable, but everything still works and no more errors. I was hoping it was an issue with a recent commit.

julenka:
I have a fix.

PR #4618, opened 2019-05-29
Fix "MRTK not found" errors even when MRTK is present in scene

julenka:
## Overview
Interactable inspector accesses the MRTK.Instance.ActiveProfile to find available input actions. Currently, MRTK.Instance will return null until the MRTK gameobject has actually been selected by the user.

Therefore, you will get an error when trying to edit an inspector until you click MRTK gameObject.

image

Fix this by changing the .Instance accessor to actually search for an instance if one is not found.

Fixes: #4610 

## Verification
- Verified that I don't get errors when inspecting interactable
- PlayMode Tests
- EditMode Tests

### DIFF
--- a/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
@@ -550,18 +550,15 @@ namespace Microsoft.MixedReality.Toolkit
                 //
                 // To avoid returning null in this case, make sure to search the scene for MRTK.
                 // We do this only when in editor to avoid any performance cost at runtime.
-                if (activeInstance == null)
+                if (activeInstance != null)
                 {
-                    var mrtks = FindObjectsOfType<MixedRealityToolkit>();
-                    for (int i = 0; i < mrtks.Length; i++)
-                    {
-                        RegisterInstance(mrtks[i]);
-                    }
+                    return activeInstance;
+                }
 
-                    if (activeInstance == null)
-                    {
-                        Debug.LogError("Didn't find / couldn't create an active instance of mixed reality toolkit. You can add one to your scene via 'Mixed Reality Toolkit->Add to Scene and Configure'");
-                    }
+                var mrtks = FindObjectsOfType<MixedRealityToolkit>();
+                for (int i = 0; i < mrtks.Length; i++)
+                {
+                    RegisterInstance(mrtks[i]);
                 }
                 return activeInstance;
             }
--- a/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
+++ b/Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
@@ -536,6 +536,7 @@ namespace Microsoft.MixedReality.Toolkit
         private static MixedRealityToolkit activeInstance;
         private static bool newInstanceBeingInitialized = false;
 
+#if UNITY_EDITOR
         /// <summary>
         /// Returns the Singleton instance of the classes type.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -539,7 +539,29 @@ private void EnsureMixedRealityRequirements()
...
+                for (int i = 0; i < mrtks.Length; i++)
+                {
+                    RegisterInstance(mrtks[i]);
+                }
+
+                if (activeInstance == null)
+                {
+                    Debug.LogError("Didn't find / couldn't create an active instance of mixed reality toolkit. You can add one to your scene via 'Mixed Reality Toolkit->Add to Scene and Configure'");
+                }
+
```
wiwei: Do you know if this is something that's editor only (i.e. and non-play mode only)? It seems like there's some weird non-determinism here around which would also get added as the activeInstance based on enumeration order of MixedRealityToolkit instances.

I don't really have enough background here to state if this is safe or not for all unity player (i.e. actual device) cases, and I'd probably defer to @Railboy for this

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -539,7 +539,29 @@ private void EnsureMixedRealityRequirements()
         /// <summary>
         /// Returns the Singleton instance of the classes type.
         /// </summary>
-        public static MixedRealityToolkit Instance => activeInstance;
+        public static MixedRealityToolkit Instance
```
Railboy: I recommend:

```
#if UNITY_EDITOR
        public static MixedRealityToolkit Instance
        {
            get
            {
                if (activeInstance != null)
                {
                    return;
                }

                var objects = FindObjectsOfType<MixedRealityToolkit>();

                switch (objects.Length)
                {
                    case 0:
                        break;

                    case 1:
                        RegisterInstance(objects[0]);
                        break;

                    default:
                        foreach (MixedRealityToolkit toolkit in objects)
                        {
                            RegisterInstance(toolkit);
                        }
                        break;
                }

                return activeInstance;
            }
        }
#else
        public static MixedRealityToolkit Instance => activeInstance;
#endif // UNITY_EDITOR
```

This is only a problem in the editor. If we only execute the search in editor, we sidestep complaints about performance.

julenka: Great idea, this feels much safer.

julenka: Fixed. Note that when simulating in editor, the search could technically be run, so it might slightly impact people when they are trying to profile code. However, they can always just delete the search if it's impacting their performance profiling.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
julenka: Fixed by using Lars's suggestion to not actually run this check except when in editor.

julenka:
Use /azp run rc2_stabilization_PR to validate PRs into the rc2_stabilization branch

julesra:
Hi, this issue seems to still be going on in MRTK 2.4.0.
I see the "Mixed Reality Toolkit is missing" error on the Interactable script, even after clicking on MixedRealityToolkit object (see screenshot). I don't know if it makes any difference at runtime.

I'm using MRTK 2.4.0, in Unity 2018.4.21f1, in the HandInteractionExamples, using a custom MRTK profile that doesn't include speech input. 

image

Issue #4579, opened 2019-05-27
unregistering core system does not change associated private member


gilbdev:
## Describe the bug

member spatialAwarenessSystem in MixedRealityToolkit (and other core services) are not set to null when the associated service is unregistered.

This force to reset the whole configuration with ResetConfiguration to have the cache properties cleared.

## To reproduce

i'm trying to change spatial awareness profile by code without changing whole mrtk profil configuration.

so i first disable the current spatailawareness
`MixedRealityToolkit.Instance.UnregisterService<IMixedRealitySpatialAwarenessSystem>()`

then i register my new profile this way:

```
//MyNewSpatialAwarenessProfile is a SerializeField in my component
object[] args = { MixedRealityToolkit.Instance, MyNewSpatialAwarenessProfile };
MixedRealityToolkit.Instance.RegisterService<IMixedRealitySpatialAwarenessSystem>(MixedRealityToolkit.Instance.ActiveProfile.SpatialAwarenessSystemSystemType, args: args)
```
the problem is that `MixedRealityToolkit.SpatialAwarenessSystem` is still pointing to the previous system, not the new registered one.

## Expected behavior
when calling UnregisterService on a CoreSystem, i expect to have the corresponding property set to null

```
public bool UnregisterService<T>(T serviceInstance)
{
    ...
    if (IsCoreSystem(interfaceType))
    {
        ClearCoreSystemCache(); //a bit too much of cache cleaning but avoid testing which system should be cleared :p, could make a bit more code to clean only the corresponding system cache
        ...
    }
    ...
}

```

## Your Setup (please complete the following information)

- Unity Version2018.3.11f1
- MRTK Version 2.0 RC1 Refresh

wiwei:
@davidkline-ms

PR #4619, opened 2019-05-29
MixedRealityToolkit object now properly unregisters services on shutdown

### LINKED ISSUE #4339: MixedRealityToolkit object does not properly unregister services on destroy
## Describe the bug

When the MixedRealityToolkit object is destroyed, it does not remove the registered service instances from the static registry class. This results in destroyed objects being registered as well as the potential for client application confusion due to multiple instances of the same service existing unexpectedly.

- Unity Version 2018.3+
- MRTK Version v2.0 RC1 Refresh

davidkline-ms:
## Overview
Previously, the MixedRealityToolkit object was failing to call UnregisterService in its shutdown path (#4339). Unregistering a service was also not causing the static accessor to be reset to null (#4579)

This change fixes both issues as well as updating the static registry class to remove the dictionary entries when the number of registered services goes to 0.

## Changes
- Fixes #4339, #4579 .

## Testing performed

- Stepped through removal of core services to confirm proper cleanup
- Registered temporary extension services and debugged removal code path
- Repeat start and stop cycles in the Unity editor (using solver scene) to confirm that services are properly loaded and reloaded
    - OpenVR project
    - Windows Mixed Reality UWP project


Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -929,7 +929,54 @@ private void DestroyAllServices()
         {
             if (!ExecuteOnAllServices(service => service.Destroy())) { return; }
 
+            List<Type> serviceTypes = activeSystems.Keys.ToList<Type>();
+            foreach (Type type in serviceTypes)
+            {
+                if (typeof(IMixedRealityBoundarySystem).IsAssignableFrom(type))
+                {
+                    UnregisterService<IMixedRealityBoundarySystem>();
+                    boundarySystem = null;
```
keveleigh: I'm not sure if this PR fully fixes #4579. It seems like this null setting take place within the `UnregisterService` call itself, instead of just when we're destroying every service.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -1108,11 +1155,6 @@ private static bool CheckServiceMatch(Type interfaceType, string serviceName, Ty
         /// <returns></returns>
         private static bool CanGetService(Type interfaceType)
         {
-            if (isApplicationQuitting)
```
keveleigh: What's the benefit of removing this?

davidkline-ms: keeping it prevents us from getting the service on shutdown.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -929,7 +929,54 @@ private void DestroyAllServices()
         {
             if (!ExecuteOnAllServices(service => service.Destroy())) { return; }
 
+            List<Type> serviceTypes = activeSystems.Keys.ToList<Type>();
+            foreach (Type type in serviceTypes)
+            {
+                if (typeof(IMixedRealityBoundarySystem).IsAssignableFrom(type))
```
wiwei: >IsAssignableFrom [](start = 56, length = 16)

Are we looking for assignable, or identity? i.e. UnregisterService<IMixedRealityBoundarySystem> looks like it takes exactly that interface, not just some type that happens to extend/be related to it.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -929,7 +929,54 @@ private void DestroyAllServices()
         {
             if (!ExecuteOnAllServices(service => service.Destroy())) { return; }
```
wiwei: >if (!ExecuteOnAllServices(service => service.Destroy())) { return; } [](start = 12, length = 68)

Should services be unregistered before they are destroyed?

It seems like otherwise there's still technically some time during which services are registered but still are running.

Then again, if the intent is that destroy should happen first because this is actually just a notification to the service that it's being shutdown, then that's fine.

The order here seems important but there's no comments/docs about the mechanics, so I figured it would be good to ask.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
wiwei: If this can be called during runtime (and shoudl probably continue to have this restriction) maybe it makes sense to keep this behavior conditional on some method parameter.

---
In reply to: [288757284](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4619#discussion_r288757284) [](ancestors = 288757284)

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: assignable. just in case a developer extends the interface

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: developers can call unregister explicitly if destroy is not desired. this is in the service locator shutdown path.

good question on the ordering. we can move the destroy call to the end of the method for safety

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: good catch. i misread 4579 when fixing :)

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
wiwei: In that case I don't fully understand how this works - we'll then look to unregister a IMixedRealityBoundarySystem object by looking at the IMixedRealityBoundarySystem type array in the list of services, but if they've registered it as a subtype it won't be there.

I guess the question is, if we have the list of types and values (i.e. the service instances, which I assume are in activeSystems) why not just call UnregisterService passing the service instance directly? It seems a bit roundabout to have to make that function re-find the object we already have here.

---
In reply to: [288762686](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4619#discussion_r288762686) [](ancestors = 288762686)

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: moved destroy call to end of method

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: good point!

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -228,8 +228,17 @@ public void ResetConfiguration(MixedRealityToolkitConfigurationProfile profile)
...
             {
                 activeSystems.Remove(interfaceType);
-                // Core services are always removable
                 MixedRealityServiceRegistry.RemoveService<T>(serviceInstance, this);
+
+                // Reset the convenience properties.
+                if (typeof(IMixedRealityBoundarySystem).IsAssignableFrom(interfaceType)) { boundarySystem = null; }
+                else if (typeof(IMixedRealityCameraSystem).IsAssignableFrom(interfaceType)) { cameraSystem = null; }
+                else if (typeof(IMixedRealityDiagnosticsSystem).IsAssignableFrom(interfaceType)) { diagnosticsSystem = null; }
+                // Focus provider reference is not managed by the MixedRealityToolkit class.
```
keveleigh: Should we clear the input system's reference to the focus provider?

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: > I guess the question is, if we have the list of types and values (i.e. the service instances, which I assume are in activeSystems)

The issue is that to properly call the generic <T> method, we need to know T at compile time.

davidkline-ms: > In that case I don't fully understand how this works - we'll then look to unregister a IMixedRealityBoundarySystem object 

On further reflection, for a system to be a core system, it must implement one of the core interfaces (in your example IMixedRealityBoundarySystem). If it is registered any other way, it would not be a core service.

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
wiwei: Should that happen in the input system itself on its own destroy call?

---
In reply to: [288768978](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4619#discussion_r288768978) [](ancestors = 288768978)

davidkline-ms: This question leads to a longer discussion of why the input system owns the reference to the focus provider, instead of the focus provider being an independently registered system.

davidkline-ms:
investigating edit mode tests...

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: changing the service locator to track the focus provider is too large of a change at this time. also, the input system interface makes the focus provider a read-only property and we are no longer taking breaking changes.

we can revisit how the focus provider is managed in a future release

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
davidkline-ms: @wiwei, agree that at this point in the release, it is prudent to be mindful of unexpected changes. will add an optional param

davidkline-ms: added class global internalShutdown variable. also removed unneeded ExecuteOnAllServices all

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -703,6 +712,9 @@ private static void RegisterInstance(MixedRealityToolkit toolkitInstance)
 
         private static void UnregisterInstance(MixedRealityToolkit toolkitInstance)
         {
+            // We are shutting this instance down.
+            internalShutdown = true;
```
wiwei: >internalShutdown = true; [](start = 12, length = 24)

Should RegisterInstance set this back to true? Otherwise it seems like even the static global will always be shutdown forever...

davidkline-ms: good catch!

PR #4623, opened 2019-05-29
Make pointers dispatch events for all button types

### LINKED ISSUE #4492: Interactable not receiving input from Grip Press or Menu Button on an MR Controller
When testing a controller grip button or the menu button, it does not trigger any state or events on the Interactable. Sometimes I use an Interactable with IsGlobal set to true, then set the Mixed Reality Input Action to Menu, so the menu button on the controller will trigger an action.

1. Place an Interactable on a gameObject.
2. Add a collider
3. Set the Input Action drop down to Menu or Grip Press.
4. Test in Unity with a MR Headset
5. Press the menu button

Interactable Inspector Menu action

Interactable Inspector Grip Press action

## Intended Behavior

MR Controller buttons other than the trigger (select) can send input into and Interactable.

- Unity Version [2018.3.9f1]
- MRTK Version [v2.0]

- WMR immersive

## Additional context

@julenka did a test with the Xbox Controller, it worked mapping buttons from the Xbox controller to an custom action, there may be an issue with existing actions?

julenka:
## Overview
Fixes: #4492 

Any pointer that extends BaseControllerPointer will by default only dispatch pointer events for input actions that match "pointerAction". This behavior is actually not consistent between other pointer types.
For example, SpherePointer overrides OnInputDown and dispatches pointer events for all actions. GazePointer also dispatches pointer events for all actions.

Consolidate pointer behavior by dispatching pointer events even when the input action doesn't match pointerAction. 

Changes:
- Rename "pointerAction" to "selectAction".
- Track all action down events that have been sent by pointer to make sure to send pointer up events correctly when source disappears or is lost.

## Verification
- I verified that this fixed #4492
- Tested that nothing unusual was happening in HandInteractionExample, using editor.
- Tested in VR in the editor
- Tested on HoloLens

>
>

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Pointers/DefaultControllerPointer.prefab
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Pointers/DefaultControllerPointer.prefab
@@ -444,12 +444,12 @@ MonoBehaviour:
     id: 0
     description: None
     axisConstraint: 0
-  pointerAction:
+  selectAction:
     id: 1
     description: Select
     axisConstraint: 2
   requiresHoldAction: 0
-  requiresActionBeforeEnabling: 0
+  requiresSelectBeforeEnabling: 0
   overrideGlobalPointerExtent: 0
   pointerExtent: 10
   defaultPointerExtent: 5.5
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
@@ -4,7 +4,9 @@
 using Microsoft.MixedReality.Toolkit.Physics;
 using Microsoft.MixedReality.Toolkit.Utilities;
 using System.Collections;
+using System.Collections.Generic;
 using UnityEngine;
+using UnityEngine.Serialization;
 
 namespace Microsoft.MixedReality.Toolkit.Input
 {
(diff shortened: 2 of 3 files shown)


keveleigh:
>Consolidate pointer behavior by dispatching pointer events even when the input action doesn't match pointerAction.

What's the difference between the pointer events and the input events now? If you need to listen for all actions (which seems like the issue with #4492), you should be able to receive them with the input events.

Pointer events had been good as a filter for a specific type of action with a specific type of pointer (select for far interaction; grab for near interaction). If we're seeing inconsistent behavior there (like you called out above, with the SpherePointer), it feels to me that we should fix that.

julenka:
> What's the difference between the pointer events and the input events now? If you need to listen for all actions (which seems like the issue with #4492), you should be able to receive them with the input events

The biggest difference is that pointer events to objects focused by a pointer, but input events are dispatched to all objects focused by all objects on an input source. For example, we do not want an air tap to click a button when the hand is near the button -- when hand is near the button should only be pressable via direct touch (see #4077). If Interactable handles OnInputDown and not OnPointerDown, then the select gesture that comes from the hand will get raised, because OnInputDown will get dispatched to all pointers attached to that input source and pokepointer is focusing on the button.

Right now there's no way that I know of to listen of arbitrary buttons pressed from different pointers.

I do see the benefit of controllung which button trigger the "select action" specifically for the case of hand rays so that we can avoid false positives while bringing hands down, but if we want to support this, then perhaps we should have a special HandLinePointer that specifically filters out the action for hands specifically, rather than filtering in the base controller pointer.

julenka:
I chatted with Luis and am going to try a different approach that does not change pointers to dispatch all input types.

Issue #4485, opened 2019-05-20
Can't build UWP APPX from MRTK Build Window in Unity


olga-microsoft:
## Describe the bug

Can't build UWP APPX from MRTK Build Window in Unity

## To reproduce

1. Create a new project and import MRTK; go to MRTK Build Window
2. Click on Unity Build Options and build
3. When build succeeds click on 'Build Appx' in the dialog
4. See error: The specified solution configuration "master|" is invalid. Please specify a valid solution configuration using the Configuration and Platform properties
5. Change configuration to Release
6. See error MSB4019: The imported project "C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\Microsoft\WindowsXaml\v15.0\Microsoft.Windows.UI.Xaml.Cpp.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.

## Expected behavior

Expect the project to build and manage own dependencies or be clear if a dependency is missing
Expect MRTK to not contain build configurations that are not supported

## Your Setup (please complete the following information)

- Unity Version 2018.3.11f1
- MRTK Version v2.0-RC1-Refresh

## Target Platform (please complete the following information)

- UWP/APPX

wiwei:
We can fix this up - note that you can still use the Unity build window to build the VS solution, and then open up that solution in visual studio to deploy to the device

gilbdev:
sure works by hand, but this comes to be problematic when using automatic build step in command line.

PR #4625, opened 2019-05-30
Can't build UWP APPX from MRTK Build Window in Unity

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4485

If you're building the UWP AppX from the build window for the very first time (without having built the Unity project before) this error shows up because the current platform isn't being properly propagated to the second step (i.e. the actual building of the AppX). If it's already been built (which is the case once you've started using your project for a while) you won't hit this codepath.

It's a pretty annoying thing to see as a first time user (i.e. I would expect it to work) so making sure that the right data (platform) is getting piped through is the right thing to do!


SoundGuy:
This also happens if i build the project , both Unity and then in Visual studio, and doesn't go away when I build the project for the first time.

Issue #4582, opened 2019-05-28
ET demo texture size is too big.


cre8ivepark:
## Describe the bug
There are two large files under Eye Tracking examples folder. 36MB and 15MB.
Not sure if the PSD file is used. Map example texture needs optimization.

![2019-05-27 20_27_28-Textures](https://user-images.githubusercontent.com/13754172/58448822-e8895780-80bd-11e9-8911-b9c4043a01f2.png)

## Branch
mrtk_development

keveleigh:
Closed via #4628

PR #4628, opened 2019-05-30
ET demo texture size is too big

sostel:
## Overview
- Addresses issue #4582
- Decreased file size of textures and replaced Photoshop (*.psd) files with *.png for the Eye Tracking navigation demo. 
- The texture for zoom is still relatively big, but if that is reduced further the zoom doesn't make much sense anymore.


wiwei:
I changed the title of this PR to reflect what it's addressing - I think that with this we're still bloating the repo (because we have historical copies of things) so there's still work here that needs to be done in order to scrub the older files away... I think

sostel:
> I changed the title of this PR to reflect what it's addressing - I think that with this we're still bloating the repo (because we have historical copies of things) so there's still work here that needs to be done in order to scrub the older files away... I think

How can we scrub the older files away? Not familiar with this. Do you know anybody who's familiar with that?

keveleigh:
I think we should take this as-is. This will help decrease the size of our release packages. I'm not sure there's a way to scrub older files without rewriting history, which we shouldn't do in a shared repo like this one.

PR #4635, opened 2019-05-30
Make the ICommandRegistry have separate Connected and Disconnected events

matthejo:
When starting on the refactor to use SpatialCoordinateSystemManager, I realized that there's sometimes the need to handle Connected or Disconnected events even without a command handler. I've separated those events out and made ICommandHandler into a delegate CommandHandler, so now Connected, Disconnected, and HandleCommand are all separate events (where HandleCommand is essentially a dictionary of events keyed on command name).


Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/CommandRegistry.cs
```diff
@@ -12,76 +12,66 @@ public abstract class CommandRegistry<TService> : Singleton<TService>,
...
+                CommandHandler commandList;
+                if (!commandHandlers.TryGetValue(command, out commandList))
                 {
-                    commandHandlers[command] = set = new HashSet<ICommandHandler>();
+                    commandHandlers[command] = handler;
+                }
+                else
+                {
+                    commandHandlers[command] = commandList + handler;
                 }
```
chrisfromwork: I think that we should throw if the specific command already has a handler. Handing the same BinaryReader to different handlers doesn't seem safe.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
     /// <summary>
     /// This class observes changes and updates content on a spectator device.
     /// </summary>
-    public class StateSynchronizationObserver : NetworkManager<StateSynchronizationObserver>, ICommandHandler
+    public class StateSynchronizationObserver : NetworkManager<StateSynchronizationObserver>
```
chrisfromwork: >NetworkManager [](start = 48, length = 14)

It's starting to seem like NetworkManager should be its own true singleton. If somehow a StateSynchronizationObserver and StateSynchronizationBroadcaster get enabled, we'll start seeing multiple connections/disconnection events through the CommandRegistry

It also seems like the onconnected/ondisconnected calls in both the state synchronization broadcaster and state synchronization observer could be achieved by registering with a network manager singleton.

chrisfromwork:
I agree with the direction of this change but would appreciate if some of the comments were addressed.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/StateSynchronizationObserver.cs
matthejo: I don't think NetworkManager can't be a singleton, as the HolographicCameraObserver and StateSynchronizationObserver are two NetworkManagers in the same Compositor scene.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/StateSynchronization/CommandRegistry.cs
matthejo: I think that makes sense. I'll have to refactor one place where I'm intentionally handling the "Camera" command in two locations, one to actually read the values and the other to just tick an lastUpdatedTime for connectivity-monitoring purposes

Issue #4279, opened 2019-05-10
Don't dispatch touch events for a hand that is already grabbing something and vice versa


chbecker-ms:
(Ported to GitHub, reference 19739652)

If a hand H1 is grabbing and moving an object O1, then touchable objects should not receive touch events from H1 until H1 has released O1.

For example: if I am moving something that has a touchable surface/button, I should not be accidentally pressing that button while moving.

### To reproduce

Import the unity package below, and then grab the cube, and move finger to touch the sphere.

[touchgrab_repro.zip](https://github.com/microsoft/MixedRealityToolkit-Unity/files/3192273/touchgrab_repro.zip)

Expected behavior is that while hand is grabbing, the sphere should not get touch events (change color)

![GrabTouchConflict](https://user-images.githubusercontent.com/168492/57942192-b558f400-7885-11e9-93c7-6fd7ec218c23.gif)

wiwei:
@julenka can you add some more details on what the request/bug is here? The last mention I've seen here is around the mediator handling this.

julenka:
Fixed by #4645

PR #4645, opened 2019-05-30
Don't allow a hand to dispatch touches if it is already grabbing something

julenka:
## Overview
Fixes: #4279 

The pointer mediator was correctly disabling a grab / touch pointer if some other pointer on a given hand had the lock, however PokePointer was not respecting this disable, it would always call SetActive = IsNearObject, overwriting the mediator's decision. Fix this by not calling SetActive = IsNearObject. This is handled by the mediator.

## Verification
Verified by creating a test scene with a grabbable cube and a touchable cube, and verifying that touchable cube doesn't get activated when the grab is active. Also verified that you can still touch cubes from the other hand.

grabtouch

I would LOVE to create a PlayMode test for this. @lukastoenneMS do you already have code that will allow me to do something like this:

```
SimulatedHand h = new SimulatedHand();
h.MoveTo(0, 0, 1.5);
h.Press();
h.MoveTo(0, 0, 3);
h.Unpress();
```

I'd like to do this in a playmode test so I can add this test case (don't allow for a hand to dispatch touches if it is already grabbing something)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -101,14 +101,18 @@ public override void OnPreSceneQuery()
...
-            IsActive = IsNearObject;
-            visuals.SetActive(IsNearObject);
+            visuals.SetActive(IsActive);
         }
 
         public override void OnPostSceneQuery()
         {
             base.OnPostSceneQuery();
 
+            if (!IsActive)
```
keveleigh: This seems interesting. Do you know if there's a reason we're running `OnPostSceneQuery` on pointers that aren't active?

julenka: No, I'm not really sure why. It looks like the focusprovider will always update all pointer focus info regardless of it it is active or not. This pattern of return early if not active is present in other pointers, for example LinePointer.  

I'd rather not update FocusProvider right now, to avoid regressions. We can open an issue to investigate so we don't forget. Pinging @luis-valverde-ms  in case he has any thoughts.

Issue #4415, opened 2019-05-16
[MRTK Shell Parity] [Bounding Box] Handles should be scaled up on hand proximity


cre8ivepark:
Part of #4200 UX Controls Parity with HoloLens 2 Shell

julenka:
@cre8ivepark  I think you could actually create an Interactable theme that will scale the handles on hover, I think it could be done all in design, no need for code modification.

keveleigh:
@danielescudero had a similar request recently, as far as highlighting the handles on proximity.

cre8ivepark:
PR in-progress as of 7/15

julenka:
merged!

PR #4646, opened 2019-05-30
BoundingBox Script- Proximity Scaling of Handles- Shell Parity

johnppella:
## Overview
changes to Boundingbox.cs script and HandInteractionExamples scene. Added Proximity scaling behaviors to boundingbox handle affordances to match the Shell Boundingbox behavior.

Near interaction
MRTK_BB_HandleScaling3

Far interaction
MRTK_BB_HandleScaling5

## Changes
- Fixes: #4415

## Verification
When either handRay or hand is 'far' away(as defined by an added Property) each handle scales to a 'far' scale- zero by default. When hand is at a medium distance, an initial scale animation plays. Then when close, a final close animation plays. during each animation, the property-specified material is applied. The behavior is intended to mimic the Shell boundingbox behavior - an initial handle increase in scale, followed by a final scale with a 'blue' material applied.
>
Default values in the script are based on the use cases illustrated in the HandInteractionExamples scene.


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1678,6 +1845,27 @@ private HandleType GetHandleType(Transform handle)
             return HandleType.None;
         }
 
+        private bool TryGetPointerPoint(Handedness handed, out Vector3 point)
```
julenka: Will this work for other input types, for example sixdof controllers? It is directly accessing the hand positions, not the pointer data.

johnppella: ahh some more properties have to be made public.

johnppella: ok ive pushed an update to the PR that sets the default values in the event that there is no corner prefab set. THis way, the defaults in the script are set for the shell parity handle prefabs, and if none are specified (the old style cube and sphere handles) the settings get set on Start(). If the BoundingBoxRuntimeExample scene is retried- the blue boxes start up and appear as illustrated above.

cre8ivepark:
John is looking into a material related memory issue.

ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2018.3.9f1
+m_EditorVersion: 2018.3.11f1
```
julenka: Can we keep the Unity version to .9f1 please, for now?

ProjectSettings/QualitySettings.asset
```diff
@@ -4,7 +4,7 @@
 QualitySettings:
   m_ObjectHideFlags: 0
   serializedVersion: 5
-  m_CurrentQuality: 5
+  m_CurrentQuality: 0
```
julenka: Was this change to quality settings intentional?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -2181,6 +2176,7 @@ public void OnSourceDetected(SourceStateEventData eventData)
                 if (sourcesDetected.Count == 0 || sourcesDetected.Contains(eventData.Controller) == false)
                 {
                     sourcesDetected.Add(eventData.Controller);
+                    Debug.Log(eventData.Controller.InputSource.SourceName);
```
julenka: Might be good to remove this before checking in

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -97,6 +96,13 @@ public enum BoundingBoxActivationType
             ActivateByProximityAndPointer,
             ActivateManually
         }
+
+        private enum HandleProximityState
```
Jarodshow: Is this something we could reuse somewhere?

julenka: I was also actually hoping that instead of modifying boundingbox we could create a  new ProximityHandler component that would expose an event like OnPointerProximityChanged(ProximityEventData d), as well as a way to poll for the closest pointer information (which would return Infinity of no pointer is within grabbing distance). You would also probably want to have events like OnNearStateEntered, OnFarStateEntered, OnMediumStateEntered.

ProximityEventData would be something like this:

```
class ProximityEventData
{
    ProximityState oldState;
    ProximityState newState;
    float oldDistance;
    float newDistance;
}

enum ProximityState
{
    FarDistance,
    NearDistance,
    MediumDistance
}
```

I think that with that component you could build the behavior you need for the bounding box handles by adding a ProximityHandler to each of the corner prefabs. 

You could then just listen for those proximity events and change the visuals in the way you're doing here. Or, a designer could hook up the Unity events OnNearStateEntered, etc. to set different states in an Interactable, which would allow for somebody to make really interesting visual feedback based on the proximity of a hand to a corner.

ProjectSettings/ProjectVersion.txt
johnppella: I pushed the fix to the memory leak problem. 
Yes I can put the version at .9f1

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: yup= removed.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: that is a major architecture change. the bb currently centralizes the handle grabbing test in order to avoid a script running on each handle. At this point- with the evolution of the bb from the simple tinkertoy style to the Shell Parity functionality- i think the handle-based script is a great idea but i think its beyond the scope of this bug fix.?

ProjectSettings/QualitySettings.asset
johnppella: no it wasnt.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
julenka: Yes it's probably out of scope, though it would be good to create an issue so we don't forget about this suggestion. We can then decide if that issue is in scope for GA. I created #4689 to track this

julenka:
Looks like there are some merge conflicts -- could you please resolve them?

ProjectSettings/ProjectVersion.txt
johnppella: version changed back to 9f1

ProjectSettings/QualitySettings.asset
johnppella: changed back to 5

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: cool!

julenka:
Hi John, it looks like there are 229 files changed in this PR -- could you please double check that you are merging from the correct branch? It looks like you might be using the wrong base branch.

julenka:
John it loks like there are conflicts again, could you please resolve?

cre8ivepark:
Closing this PR - New PR opened #4895

Issue #4520, opened 2019-05-22
SpeechInputHandlerInspector broken


lukastoenneMS:
## Describe the bug

The SpeechInputHandlerInspector does not show the keyword responses any longer, making it useless:
![image](https://user-images.githubusercontent.com/47210016/58187288-c4410b80-7cb6-11e9-9331-4fe471847c7f.png)

lukastoenneMS:
Workaround: It seems that after opening a project i have to select the MixedRealityToolkit object at least once, so that MixedRealityToolkit.IsInitialized becomes true. After that the speech handler gui works fine.

lukastoenneMS:
Similar issue with `MixedRealityInputAction` fields: They show up with a message: "Missing MixedRealityToolkit", fine after showing the MRTK object once in inspector.

wiwei:
Closing this one out as fixed by #4661

Issue #4654, opened 2019-05-31
Can't bind the keyboard input for voice input


Ailegen:
![image](https://user-images.githubusercontent.com/16714300/58687078-2edb0100-83b3-11e9-843c-f08c6129b8e9.png)

Ailegen:
should work according to the document
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Input/Speech.html

Troy-Ferrell:
@Ailegen see PR referenced above. The dropdown selector UI should still work if you click on it. It's not a functional bug, it's just a visual bug. The PR referenced though fixes the visual issue

wiwei:
Closing this one out as fixed by #4661

PR #4437, opened 2019-05-16
Update runtime components to access system profiles from the system instance

davidkline-ms:
## Overview
As part of #3545, systems are passed their profile via the constructor. With the PR series that started with #4099, client code can access system instances from the service registry.

This change updates runtime and client code (with a small set of exceptions) to access system profiles directly from the system instance. For example:

```
 MixedRealityInputAction[] actions = InputSystem.InputSystemProfile.InputActionsProfile.InputActions;
```

NOTE:
While this _is_ a breaking change, it is purely additive for code that _consumes_ services. Services themselves will require updating to support IMixedRealityService.ConfigurationProfile and <serviceinterface>.<service>Profile properties.

Changes to handle editor code (and the set of remaining runtime files) are forthcoming.

Fixes: #4335

## Testing performed
- Play mode, solver demo scene (select objects, teleport)
    - OpenVR
    - Windows Mixed Reality Immersive
- Edit mode test automation

PR #4661, opened 2019-05-31
Handle additional edge cases for MRTK inspectors

Troy-Ferrell:
## Overview

There are two key dependencies between profiles/inputhandlers: input actions and speech keywords (i.e speechinputhandler). The inspector logic for these impacted components has been updated to be simpler, more consistent, and accurate with useful error messages to the user informing them of the issue (i.e no MRTK in scene, etc etc.). 

Rendering of speech commands when rendered as sub-profile view (i.e within input profile) would not work properly for layout. Switched model to be vertical components to preserve useful formatting

Added forced height on scrollview for input actions inspector when rendered as sub-profile

Fixed edge case where inspector logic does not account for an active profile assigned to the scene MRTK instance (an unlikely edge case)

Fixed DefaultMouseProfile to correctly set IsCustomProfile to false

Fixed DefaultHololens2..profiles to correctly set IsCustomProfile to false

Fixed "Back to.." buttons for profile up the hierarchy to actually render

## Changes
- Fixes: #4520 SpeechInputHandlerInspector broken
#4467 
Harden MRTK inspectors & increase re-usability of code 
#4654 
Can't bind the keyboard input for voice input

## Verification

Removed input actions profile
Removed active profile
Removed MRTK instance
Switched hierarchy with multiple gesture profiles (but only one active in instance)
Deleted MRTK instance & re-added with SpeechInputHandler button when error message shows


Troy-Ferrell:
@luis-valverde-ms and @provencher FYI

Assets/MixedRealityToolkit.SDK/Inspectors/Input/Handlers/BaseInputHandlerInspector.cs
```diff
@@ -23,5 +23,32 @@ public override void OnInspectorGUI()
...
         }
+
+        /// <summary>
+        /// If MRTK is not initialized in scene, then show an error message and button to add MRTK to the scene
+        /// If MRTK does not has an active profile, show an error message
+        /// </summary>
+        /// <returns>true if MRTK is initialized and has active profile, false otherwise</returns>
+        protected bool CheckMixedRealityToolkit()
+        {
+            if (!MixedRealityToolkit.IsInitialized)
```
davidkline-ms: A MixedRealityToolkit, instance is no longer required to use the MRTK (this is in the process of becoming easier to configure).

This method should likely not be added.

Assets/MixedRealityToolkit.SDK/Inspectors/Input/Handlers/PointerClickHandlerInspector.cs
```diff
@@ -26,13 +27,35 @@ public override void OnInspectorGUI()
         {
             base.OnInspectorGUI();
 
-            if (!MixedRealityInspectorUtility.CheckMixedRealityConfigured(false)) { return; }
+            bool enabled = CheckMixedRealityToolkit();
+            if (enabled)
+            {
+                if (!MixedRealityToolkit.Instance.ActiveProfile.IsInputSystemEnabled)
```
davidkline-ms: The new pattern for accessing profiles and systems has changed. Please see #4315 and #4437 for reference and feel free to reach out with questions.

Troy-Ferrell: @davidkline-ms , hmm so what service then contains IsInputSystemEnabled as your commet highlighted(aka what service consumes MixedRealityToolkitConfigurationProfile in it's constructor)? The only method I see for accessing profiles is roughly:

```
IMixedRealityService service;
MixedRealityServiceRegistry.TryGetService<IMixedRealityService>(out service); 

svc.profile.blah.dosomething etc
```

Assets/MixedRealityToolkit.SDK/Inspectors/Input/Handlers/BaseInputHandlerInspector.cs
Troy-Ferrell: @davidkline-ms if there is no MRTK instance in scene, how does one get the list of available inputactions? Looking at the updated Interactable.cs, it still checks if MRTK is initialized. 

       ```
 public static bool TryGetInputActions(out string[] descriptionsArray)
        {
            if (!MixedRealityToolkit.ConfirmInitialized() || !MixedRealityToolkit.Instance.HasActiveProfile)
            {
                descriptionsArray = null;
                return false;
            }

            MixedRealityInputAction[] actions = InputSystem.InputSystemProfile.InputActionsProfile.InputActions;

            descriptionsArray = new string[actions.Length];
            for (int i = 0; i < actions.Length; i++)
            {
                descriptionsArray[i] = actions[i].Description;
            }

            return true;
        }
```

Comment (davidkline-ms) [reply Assets/MixedRealityToolkit.SDK/Inspectors/Input/Handlers/BaseInputHandlerInspector.cs]: The code snippet above uses a cached input system that was acquired using

```
MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>()
```

This pattern is supported for the MixedRealityToolkit object as well as standalone services.

Please let me know if you need more clarity or have questions.

Troy-Ferrell: @davidkline-ms , yes it uses a cached input system. The point though is that the input actions list is only grabbed from the MRTK instance currently in scene. If there is no MRTK instance in scene, then you can't get the input actions list. Thus, my CheckMixedRealityToolkit() is still necessary. There is still a dependency

Assets/MixedRealityToolkit.SDK/Inspectors/Input/Handlers/PointerClickHandlerInspector.cs
davidkline-ms: at this point, there is nothing that consumes that profile, other than the MixedRealityToolkit object.

based on feedback from some customers, there will be scenarios where there isnt a MixedRealityToolkit object in the scene.

many (though not yet all) instances of Is<Service>Enabled are replaceable with a check to see if the system is non-null

Assets/MixedRealityToolkit.SDK/Inspectors/Input/Handlers/BaseInputHandlerInspector.cs
davidkline-ms: The primary concern is to not add any new references to MixedRealityToolkit.Instance. Let's discuss and post the results here.

Troy-Ferrell:
After syncing with @davidkline-ms , the use of MRTK object references are going to remain in the PR. This is because 
1) These are only editor code additions, not runtime use additions
2) The TryGetService & related paradigm needs to be rationalized for editor use since MRTK is not actually running

wiwei:
Thanks!

PR #4662, opened 2019-05-31
Add sdk demand for release pipeline

MenelvagorMilsom:


MenelvagorMilsom:
@wiwei, should we add this for all pipelines?

pipelines/ci-release.yml
```diff
@@ -12,6 +12,7 @@ jobs:
     demands:
     - Unity2018.3.7f1  # variable expansion not allowed here
     - COG-UnityCache-WUS2-01
+    - SDK_18352 -equals TRUE
```
wiwei:
```suggestion
    - SDK_18362 -equals TRUE
```

wiwei:
@MenelvagorMilsom I think that would be best getting this demand added for all pipelines.

MenelvagorMilsom:
@keveleigh yeah, it does that sometimes

macborow:
FYI, I got that "file not found" error when running a cloned mrtk_CI job against my fork. I thought it was due to an expired authorization token as re-authenticating seems to have fixed it. Perhaps it's an intermittent failure though as I think this build recovered after being forced again... I'll mention that to Nick and perhaps he can pass this info on to the ADO team.

PR #4664, opened 2019-05-31
Move package signing into separate template

macborow:
mrtk_CI pipeline is failing due to missing signing task - it appears that it is not enough to disable a task that doesn't exist (in the context of AIPMR project) as perhaps some validation happens before the template is fully evaluated. This change attempts to remove the signing task from the ci.yml template (only left in ci-release.yml).

### DIFF
--- a/pipelines/ci-release.yml
+++ b/pipelines/ci-release.yml
@@ -36,5 +36,7 @@ jobs:
     parameters:
       ConfigName: "MRTKSignConfig.xml"
   - template: templates/package.yml
+  - template: templates/tasks/signing.yml
     parameters:
-      Sign: true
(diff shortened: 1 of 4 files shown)


macborow:
This change affects only mrtk_CI build, so there's no point in running mrtk_PR. I cloned the mrtk_CI job and ran against the fork to make sure that it passes before we merge:
https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=2010&_a=summary
There are 4 jobs queued ahead of this one, so it looks like it won't run before EOD in Dublin.
Will, would you mind completing the PR if that build looks fine to unblock mrtk_CI on mrtk_development?

wiwei:
Will do.

PR #4665, opened 2019-05-31
Mobile recording for spectator view

chrisfromwork:
This review adds mobile recording to the android asa spectator view scene.

1) We now can specify to have mobile recording/mobile recording ui via SpectatorView.cs
2) If mobile recording is enabled, SpectatorView will dynamically create a recording ui prefab and a IRecordingService.

### DIFF
--- a/Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerAnchor.cs
+++ b/Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerAnchor.cs
@@ -111,6 +111,21 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView.H
             markerDetector.MarkersUpdated -= MarkerDetector_MarkersUpdated;
 
         }
+#else
+        public void HandleCommand(SocketEndpoint endpoint, string command, BinaryReader reader, int remainingDataSize)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public void OnConnected(SocketEndpoint endpoint)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public void OnDisconnected(SocketEndpoint endpoint)
+        {
+            throw new System.NotImplementedException();
+        }
 #endif
     }
 }
--- a/Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerAnchor.cs
+++ b/Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerAnchor.cs
@@ -114,17 +114,17 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView.H
 #else
         public void HandleCommand(SocketEndpoint endpoint, string command, BinaryReader reader, int remainingDataSize)
         {
-            throw new System.NotImplementedException();
+            throw new System.PlatformNotSupportedException();
         }
 
         public void OnConnected(SocketEndpoint endpoint)
         {
-            throw new System.NotImplementedException();
+            throw new System.PlatformNotSupportedException();
         }
 
         public void OnDisconnected(SocketEndpoint endpoint)
         {
-            throw new System.NotImplementedException();
+            throw new System.PlatformNotSupportedException();
         }
 #endif
     }
--- a/Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/UI/RecordingServiceVisual.cs
+++ b/Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/UI/RecordingServiceVisual.cs
@@ -23,6 +23,13 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView.U
         }
 
         /// <summary>
+        /// Game object that contains the record and preview buttons
+        /// </summary>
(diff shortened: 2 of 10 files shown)


julenka:
I'm not sure I understand what this PR is doing. It looks like it's creating a UI for recording spectatorview to a video. When I make UI facing changes I'll often record a video of the experience and then convert it to a GIF using screen2gif. Maybe that would be a good use case for this, so we can better see the change? See this for a nice example of a gif showing a feature: #4646

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/UI/RecordingServiceVisual.cs
```diff
@@ -93,17 +93,15 @@ protected void Awake()
...
         }
 
         protected void Update()
         {
             if (_recordingService != null &&
                 _previewButton != null)
             {
-                _previewButton.gameObject?.SetActive(_recordingService.IsRecordingAvailable());
+                bool showPreviewButton = _recordingService.IsRecordingAvailable() && _recordButton.gameObject.activeSelf;
+                _previewButton.gameObject?.SetActive(showPreviewButton);
```
matthejo: >gameObject?. [](start = 31, length = 12)

Because Unity has an override for Equals and operator== to treat a non-null CLR object that represents a destroyed GameObject/Component as 'null', the "?." and "??" operators don't work like you'd expect with Component or GameObject classes (because they explicitly call ReferenceEquals(o, null) and don't call through overrides of Equals or operator==).  You instead have to have a traditional null check for GameObjects or Component.

chrisfromwork:
mobileRecording1
mobileRecording2

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/MarkerDetection/ArUco/HoloLens/ArUcoMarkerAnchor.cs
```diff
@@ -111,6 +111,21 @@ private void OnDestroy()
             markerDetector.MarkersUpdated -= MarkerDetector_MarkersUpdated;
 
         }
+#else
+        public void HandleCommand(SocketEndpoint endpoint, string command, BinaryReader reader, int remainingDataSize)
+        {
+            throw new System.NotImplementedException();
```
andreiborodin: nit: There should be a better exception: PlatformNotSupported I believe, or something of the sort.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/UI/RecordingServiceVisual.cs
```diff
@@ -93,17 +93,18 @@ protected void Awake()
...
-            _recordButton.onClick.AddListener(OnRecordClick);
-            _previewButton.onClick.AddListener(OnPreviewClick);
         }
 
         protected void Update()
         {
             if (_recordingService != null &&
-                _previewButton != null)
+                _recordButton != null &&
+                _recordButton.gameObject != null &&
```
andreiborodin: Why the GameObject checks?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/UI/RecordingServiceVisual.cs
```diff
@@ -93,17 +93,18 @@ protected void Awake()
...
+                _previewButton != null &&
+                _previewButton.gameObject != null)
             {
-                _previewButton.gameObject?.SetActive(_recordingService.IsRecordingAvailable());
+                bool showPreviewButton = _recordingService.IsRecordingAvailable() && _recordButton.gameObject.activeSelf;
```
andreiborodin: >_recordButton.gameObject.activeSelf [](start = 85, length = 35)

Does this signify "not currently recording", is there a better field that could be used?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/UI/RecordingServiceVisual.cs
```diff
@@ -237,13 +238,15 @@ public void Show()
                 StopRecording();
             }
 
-            gameObject.SetActive(true);
+            _recordButton.gameObject.SetActive(true);
+            _previewButton.gameObject.SetActive(true);
```
andreiborodin: >_previewButton.gameObject.SetActive(true); [](start = 12, length = 42)

Seems strange we always enable previewButton here, even if preview is not available.

Perhaps we can toggle on/off the parent of both these buttons, and let the various states (Recording, preview enabled, etc) toggle whether these buttons are enabled/disabled

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/UI/RecordingServiceVisual.cs
chrisfromwork: I guess we don't have to. This is my naiveness to Unity showing. I wasn't sure if button components could have null game objects

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/UI/RecordingServiceVisual.cs
chrisfromwork: No, technically this checks "Has any recording been made". I think that parenting these buttons to toggle their visibility will improve this experience. But I should probably move away from continually calling this api. Once it's true, itll always be true.

chrisfromwork: scratch that, once its true for android it will always be true. For ios it may vary. I'll update the android recording service to reflect this, but will continual check this value each update

PR #4669, opened 2019-05-31
Slate prefab box collider performance issue

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4167

Slates use the bounding box's rotate colliders in order to handle rotation - note that even though the slate provides its own box-like prefab, the bounding box still creates a sphere collider (which has wayyyyy more vertices than the simple box), even though a basic box would suffice. This change addresses the issue by adding an option to either use a box or sphere collider when a custom prefab is provided for bounding box rotate handles.

This is done in this way to avoid being a breaking behavior change - other folks can choose to use a more performant collider, and the slate is the only thing updated to use this behavior for now.

For the prefab, all I did was update the dropdown select button to choose "box"

This is my undoing the revert that did for the merge - this time around I opted to just update the prefab manually (which still works fine).


wiwei:
Yep, this was followup from https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4461 which was causing that scary merge when going from stabilization -> dev

PR #4672, opened 2019-05-31
Update solver logic to update more consistently and respect scale

gejohnst:
## Overview
Changing the solverUpdate methodology to always apply the update from working to goal, regardless of solver. This allows for those creating new solvers to ignore the update process in favor of simply updating the goal position/rotation/scale. Also, by guaranteeing that these are called, maintainScale becomes a functional variable again
Also some minor documentation updates.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Orbital.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Orbital.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using UnityEngine;
+using UnityEngine.Serialization;
 
 namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
 {
@@ -59,6 +60,7 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
         }
 
         [SerializeField]
+        [FormerlySerializedAs(oldName: "UseAngleSteppingForWorldOffset")]
         [Tooltip("Lock the rotation to a specified number of steps around the tracked object.")]
         private bool useAngleStepping = false;
 
--- a/Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Orbital.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Orbital.cs
@@ -27,7 +27,7 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
         }
 
         [SerializeField]
-        [Tooltip("XYZ offset for this object in relation to the TrackedObject/TargetTransform. Mixing local and world offsets is not recommended.")]
+        [Tooltip("XYZ offset for this object oriented to the TrackedObject/TargetTransform's forward. Mixing local and world offsets is not recommended. Local offsets are applied before world offsets.")]
         private Vector3 localOffset = new Vector3(0, -1, 1);
 
         /// <summary>
@@ -43,7 +43,7 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
         }
 
         [SerializeField]
-        [Tooltip("XYZ offset for this object in worldspace, best used with the YawOnly orientationType. Mixing local and world offsets is not recommended.")]
+        [Tooltip("XYZ offset for this object in worldspace, best used with the YawOnly orientationType. Mixing local and world offsets is not recommended. Local offsets are applied before world offsets.")]
         private Vector3 worldOffset = Vector3.zero;
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Orbital.cs
```diff
@@ -60,15 +60,15 @@ public Vector3 WorldOffset
 
         [SerializeField]
         [Tooltip("Lock the rotation to a specified number of steps around the tracked object.")]
-        private bool useAngleSteppingForWorldOffset = false;
+        private bool useAngleStepping = false;
```
keveleigh: Please add the [`FormerlySerializedAs`](https://docs.unity3d.com/ScriptReference/Serialization.FormerlySerializedAsAttribute.html) tag to ensure we don't break previous serialization.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
```diff
@@ -198,23 +228,25 @@ public void SolverUpdateEntry()
         /// </remarks>
         /// <param name="position"></param>
         /// <param name="rotation"></param>
-        public virtual void SnapTo(Vector3 position, Quaternion rotation)
+        public virtual void SnapTo(Vector3 position, Quaternion rotation, Vector3 scale)
```
keveleigh: It might be good to leave the previous version around, with an [`Obsolete` attribute](https://docs.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=netframework-4.8) and guidance to move to the new one, so we don't break any custom solvers that might have overridden the old method signature.

keveleigh: Same for `SnapGoalTo` below.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
```diff
@@ -18,24 +19,23 @@ public abstract class Solver : MonoBehaviour
...
 
         [SerializeField]
         [Tooltip("If true, the Solver will respect the object's original scale values")]
         private bool maintainScale = true;
-
         [SerializeField]
-        [Tooltip("Working output is smoothed if true. Otherwise, snapped")]
-        private bool smoothing = true;
+        [Tooltip("If true, updates are smoothed to the target. Otherwise, they are snapped to the target")]
+        public bool smoothing = true;
```
julenka: Per the [MRTK Coding Guidelines](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/CodingGuidelines.html), it would be great if we could keep this field private, and add a public getter and sett

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -176,11 +176,16 @@ private void LateUpdate()
         {
```
radicalad: Since you're in here, can you update the `SolverHandler.Start()` method to respect the offset values? I found a bug earlier this week, but haven't had time to open a PR.

`        private void Start()`
        `{`
            `// TransformTarget overrides TrackedObjectToReference`
            `if (!transformTarget)`
            `{`
                `AttachToNewTrackedObject();`
            `}`
            `else`
            `{`
                `// we need to ensure the offset gets applied`
                `TrackTransform(transformTarget);`
            }``
        `}`

wiwei: It might make sense to do this in a separate PR (with it's own tests and focus).

This request is generally okay if you have significant test coverage and can be confident about the safety of incremental and tangential changes, but without it, you have to do another manual test pass (potentially holding up this change, etc)

luis-valverde-ms:
@gejohnst , @wiwei , do we think this is ready for commit?

julenka:
@gejohnst I would like to complete this PR, could you please resolve the comment I had above regarding the public smoothing field, per the MRTK coding guidelines?

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
```diff
@@ -18,24 +19,23 @@ public abstract class Solver : MonoBehaviour
...
         private float scaleLerpTime = 0;
 
         [SerializeField]
         [Tooltip("If true, the Solver will respect the object's original scale values")]
         private bool maintainScale = true;
-
         [SerializeField]
-        [Tooltip("Working output is smoothed if true. Otherwise, snapped")]
-        private bool smoothing = true;
+        [Tooltip("If true, updates are smoothed to the target. Otherwise, they are snapped to the target")]
```
julenka:
```suggestion
        [Tooltip("If true, updates are smoothed to the target. Otherwise, they are snapped to the target")]
        private bool smoothing = true;
        public bool Smoothing 
        { 
            get => smoothing; 
            set => smoothing = value; 
        }
```

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
keveleigh: Line 38 contains the private field, so this looks like a duplicately named field and should probably be removed.

Issue #4611, opened 2019-05-29
Interactable Component only applies to host GameObject


NColombeMS:
I'm using a prefab object, composed of sub-objects that I can individually manipulate. 

At some point, I would like to make the whole prefab interactable, in order to select it. So I figured I could use an interactable component that I could enable/disable at the root of the prefab instance.

Problem is, it expects everything to be  set up at its host gameobject and does not seem to care about its children. 

What would be nice is if the interactable component could be applied to the whole hierarchy of objects.

julenka:
@killerantz  I'm not sure if Interactable is designed for this use case, perhaps we need to clarify the documentation?

killerantz:
**Let me try restate the goal.**
There's a prefab with Interactables as children, sub-objects that can be individually manipulated.

**The desire is to:**
1. Add Interactable to the parent and have the other Interactables listen for state changes from the parent as well as their own interaction state changes?
2. Add Interactable to the parent and turn on and off (enable/disable) children Interactables by turning on or off (enable/disable) the parent Interactable?
3. Add Interactable to the parent and have sub objects respond to state changes without connecting them to a profile.

**Some issues to solve:**
1. How to tell when to listen to the parent Interactable and when to listen to the sub-object Interactable.
2. Some of the feedback you could put on a child object is to disable it, but once it's disabled, Update does not run on it anymore, so it won't pickup when it's time to enable itself again. This is one reason for the top down approach, that and the organization. I thought about building a profile that can sit on another object (child or sibling) and point to an Interactable to monitor in some use cases, this could still be done.
3. Making sure the children can still be interacted with. The parent Interactable uses a collider to detect a pointer. If the children Interactables are inside the parent's collider, a pointer will never reach them.

**Some ideas:**
1. This one is pretty complicated.
-- Maybe adding some additional states to the sub-objects (Like Default, Custom, Focus, Press, Disabled). Setup the sub-Interactable to use the new state object at the top of the Interactable. Then use some events on the parent Interactable to set states on the children, like on Parent focus, use the OnFocusReceiver event to set the children's Custom state to true, then add some feedback for that state on the children. When the parent gets focus, all the children could change colors for instance, but their own interaction would still override the parent.

[...]

PR #4673, opened 2019-05-31
Interactable not responding to menu, grip take 2. Also add TouchHandler

julenka:
## Overview
Fixes: #4492 by changing Interactable to listen to OnInputDown instead of OnPointerDown. I did this because pointer events are designed to only be sent for certain buttons, in most cases the 'select' button. So pointers do not send events when menu button, grip button, any other button is pressed. Interactables however want to listen to all possible buttons, so they need to listen to OnInputDown.

I needed to do some potentially confusing logic to ensure that the poke pointer does not cause "select" action to trigger interactable when you are near a button and air tap. This happens because when an input does an input down like air tap, that input gets dispatched to every object focused by every pointer attached to that input, even if the PokePointer shouldn't really be dispatching air taps.

I also made a few other improvements:
- Add a basic TouchHandler class to expose touch events as unity events
- Update interactable button prefabs to use TouchHandler to call SetInputDown and up on Interactable, which will allow the button to trigger clicks.
- Expose SetInputDown() and SetInputUp() events for Interactable to let more complex logic set the states of the Interactable.
- Add more details to the TouchEvent data structure
- Fixes: #4611 by listening for touch events from children
- Update Interactable prefabs and examples examples to work with touch events out of the box (fix up the NearInteractionTouchables).

## Verification
- Tested that Interactables respond to grip, menu buttons in Windows MR with motion controllers.
- Test in editor that handinteractionexamples scene works
- PlayMode tests pass.
- Testing on HoloLens currently

>
> git fetch origin pull/4673/head:name_of_local_branch
>


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -631,139 +625,44 @@ public void OnFocusExit(FocusEventData eventData)
...
-            }
-        }
-
-        public void OnPointerDragged(MixedRealityPointerEventData eventData)
+        protected void StopClickTimer()
         {
-            bool isGrab = false;
-            if (!HasGesture && CanInteract() && (ShouldListen(eventData, out isGrab) || eventData.MixedRealityInputAction.Description == "None") && (dragStart - eventData.Pointer.Position).magnitude > 0.1f)
-            {
-                SetGesture(true);
```
julenka: After this PR interactable would no longer set "is gesture" to true. Need to add that back in.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -887,20 +790,54 @@ protected void FilterThemesByDimensions()
...
+        /// Returns true if the inputeventdata is being dispatched from a near pointer
+        /// </summary>
+        /// <param name="eventData"></param>
+        /// <returns></returns>
+        private bool IsInputFromNearInteraction(InputEventData eventData)
+        {
+            bool isAnyNearpointerFocusing = false;
+            for (int i = 0; i < pointers.Count; i++)
+            {
+                if (pointers[i] is IMixedRealityNearPointer && pointers[i].InputSourceParent.SourceId == eventData.InputSource.SourceId)
```
thalbern: I would turn this if around so you do the source id comparison first and if that one returns false it doesn't have to do the cast

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/TouchHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.MixedReality.Toolkit.UI;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class TouchHandler : MonoBehaviour, IMixedRealityTouchHandler
+{
+
+    #region Event handlers
+    public TouchEvent OnTouchStarted = new TouchEvent();
```
luis-valverde-ms: >TouchEvent [](start = 11, length = 10)

What about having TouchEvent contain HandTrackingInputEventData directly instead of TouchEventData?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1097,7 +1075,180 @@ void IMixedRealityTouchHandler.OnTouchCompleted(HandTrackingInputEventData event
...
+        public void OnInputUp(InputEventData eventData)
+        {
+            if ((!CanInteract() && !HasPress))
+            {
+                return;
+            }
+
+            if (ShouldListenToUpDownEvent(eventData))
+            {
+                SetPress(false);
```
luis-valverde-ms: This is already set in SetInputUp()

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1097,7 +1075,180 @@ void IMixedRealityTouchHandler.OnTouchCompleted(HandTrackingInputEventData event
...
+            }
+        }
+
+        /// <summary>
+        /// Public method that can be used to set state of interactable
+        /// corresponding to an input going down (select button, menu button, touch) 
+        /// </summary>
+        public void SetInputDown()
+        {
+            if (!CanInteract())
```
luis-valverde-ms: This is already checked in OnInputDown()

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1097,7 +1075,180 @@ void IMixedRealityTouchHandler.OnTouchCompleted(HandTrackingInputEventData event
...
+            {
+                return;
+            }
+
+            SetPress(false);
+
+            SetGesture(false);
+            SetPress(false);
```
luis-valverde-ms: Do we need to `SetPress(false)` twice?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -42,9 +47,17 @@ protected static IMixedRealityInputSystem InputSystem
             }
         }
 
-        // list of pointers
-        protected List<IMixedRealityPointer> pointers = new List<IMixedRealityPointer>();
-        public List<IMixedRealityPointer> Focusers => pointers;
+        protected List<IMixedRealityPointer> focusingPointers = new List<IMixedRealityPointer>();
```
luis-valverde-ms: I would make this readonly and clear it as needed instead of recreating it. Same for pressingInputSources.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -42,9 +47,17 @@ protected static IMixedRealityInputSystem InputSystem
...
+        /// <summary>
+        /// Pointers that are focusing the interactable
+        /// </summary>
+        public List<IMixedRealityPointer> Focusers => focusingPointers;
```
luis-valverde-ms: By convention, properties are named as the field backing them but with initial capital letter. In this case it should be `FocusingPointers`. Same for `Pressers` below.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -127,25 +140,38 @@ protected static IMixedRealityInputSystem InputSystem
...
         /// how many times this interactable was clicked
         /// good for checking when a click event occurs.
         /// </summary>
         public int ClickCount => clickCount;
 
+        // 
+        // Variables for determining gesture state
+        //
+        protected Vector3? dragStart3 = null;
+        protected Vector2? dragStart2 = null;
```
luis-valverde-ms: A bit more documentation here and maybe reflect in the naming that they are for 2D and 3D drag.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1097,7 +1075,180 @@ void IMixedRealityTouchHandler.OnTouchCompleted(HandTrackingInputEventData event
...
+        /// corresponding to an input going down (select button, menu button, touch) 
+        /// </summary>
+        public void SetInputDown()
+        {
+            if (!CanInteract())
+            {
+                return;
+            }
+            dragStart3 = null;
+            dragStart2 = null;
```
luis-valverde-ms: If I read it correctly, this seems to imply that pressing a button in any controller focusing the interactable will reset any drag operation progress. Is that right?

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/TouchHandler.cs
julenka: I thought about this, but HandTrackingInputEventData has fields like "InputData" which is hard to interpret. I think TouchPosition is more clearly communicating the underlying data. Think it will be eaiser for people to understand.

julenka: I thought about this, but HandTrackingInputEventData has fields like "InputData" which is hard to interpret. I think TouchPosition is more clearly communicating the underlying data. Think it will be eaiser for people to understand.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -887,20 +790,54 @@ protected void FilterThemesByDimensions()
...
+                    if (pointers[i] is PokePointer)
+                    {
+                        pokePointerCount++;
+                    }
+                }
             }
 
-            bool isListening = HasFocus || IsGlobal;
-            return eventData.MixedRealityInputAction == InputAction && isListening;
+            if (pointerCount > 0 && pointerCount == pokePointerCount)
```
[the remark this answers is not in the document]
julenka: Yes -- if we are listening to global event, then it may be that no pointers are focusing the object.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: Good point. I'll do this above as well in ShouldListen.

julenka: Done.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1012,7 +950,7 @@ protected IEnumerator GlobalVisualReset(float time)
         protected IEnumerator InputDownTimer(float time)
         {
             yield return new WaitForSeconds(time);
-            inputTimer = null;
+            clickValidTimer = null;
```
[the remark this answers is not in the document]
julenka: I think @killerantz has been using this pattern for a while so I'm going to leave it as is for now to not introduce new bugs.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: Yes but I'd like to have this here in case it gets called from some other code.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1097,7 +1035,86 @@ void IMixedRealityTouchHandler.OnTouchCompleted(HandTrackingInputEventData event
...
+            StartClickTimer(true);
+        }
+
+        /// <summary>
+        /// Public method that can be used to set state of interactable
+        /// corresponding to an input going up.
+        /// </summary>
+        public void SetInputUp()
+        {
+            if (!CanInteract())
```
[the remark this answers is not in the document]
julenka: Nice catch. Done.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: Nope, good catch.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: Out of curiosity, why?

julenka: Fixed.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: Done.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/TouchHandler.cs
julenka: I thought about it more and decided to go with your suggestion, since I noticed PointerHandler uses a similar pattern.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -615,7 +649,7 @@ public void OnFocusEnter(FocusEventData eventData)
         {
             if (CanInteract())
             {
-                SetFocus(pointers.Count > 0);
+                SetFocus(focusingPointers.Count > 0);
```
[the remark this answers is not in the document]
julenka: Done.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: For now, yes. Interactable is not really supporting two handed interactions correctly now. I'll leave that for a different PR though.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: Done! I also refactored / cleaned up that code to remove duplication.

julenka:
I think this change could also fix #4611 because the change is updating interactable to listen to events from children.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
luis-valverde-ms: To avoid a new allocation each time.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -220,21 +250,10 @@ private void OnEnable()
...
-                if ((Interactable)pointers[i].FocusTarget != this)
-                {
-                    pointers.RemoveAt(i);
-                }
-                else
-                {
-                    pointersWithFocus++;
-                }
-            }
+            int pointerCount = focusingPointers.Count - 1;
```
luis-valverde-ms: pointerCount is not used anymore

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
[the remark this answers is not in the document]
luis-valverde-ms: I meant that we shouldn't be passing the result of `focusPointers.Count > 0` here, we should be asserting that that is true instead.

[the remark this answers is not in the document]
julenka: I did try this but the assert fired very frequently. I'll look into it and see if this is a valid bug I can fix now.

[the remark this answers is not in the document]
julenka: Okay I realize now that my assert was just wrong. I was using focusingPointers.Count == 0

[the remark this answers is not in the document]
julenka: Fixed!

Issue #4325, opened 2019-05-13
 Support custom raycasting implementations


chbecker-ms:
(Ported to GitHub, reference 20077217)

Some of the MRW apps use "custom" raycasting for certain interactions. It does not appear that there is a way to provide a custom raycaster with the MRTK vNext input system. This poses a problem when trying to adopt MRTK vNext (at least the input system) for MRW.

One of the requirements in was that we override the default RayCast logic that HTK has, and while performing Raycasts we consider whether we hit a hologram, SR mesh or a hole in the SR mesh. If we have a hole in the SR mesh, we still generate a hit point for our annotations to get placed at. Monterrey input system had the ability to allow consuming apps to override the raycast logic. And so this was identified as a potential feature parity gap between the app's existing input system and HoloToolKit/MRTK.

ryantrem:
@wiwei - these folks would be good contacts to validate any proposed solutions to this issue: @Kjakubzak, @rferrese, @ritijain

wiwei:
@Kjakubzak, @rferrese, @ritijain do you guys have some more specific asks here (i.e. are there some specific points in the code that you can help point to that are inflexible, or have snippets of code that show the overall pain points and illustrate things?)

I want to get sufficient detail within this issue that we can both understand what it is we're trying to improve and then also understand how to validate that as well.

ryantrem:
Also @smcbeth may have thoughts here.

kircher1:
This is a feature request that would benefit the [maps sdk project](https://github.com/microsoft/MapsSDK-Unity) as well.

It would be great if the focus/gaze cursors and controller rays would follow the surface of the map; however, the map doesn't generate a collider on the fly (a lot of perf overhead for that), so it's not ray-castable through Unity's physics system.

ritijain:
@wiwei  This ask was due to multiple scenarios that were missing in HTK, and due to some custom solutions that our apps had. Some of the things have definitely been fixed in MRTK and having the ability to write custom Gaze/pointer  systems because of availability  of more interfaces  definitely makes this easier. 

1. The version of HTK that we were using only did Layer prioritization based comparison of hit results, and not distance based comparison, which is why we had to override/swap what HTK did. An ability to override the prioritization logic would help.
2. Sometimes depending on scenario, we might need to prioritize a distance based comparison of raycast hit result rather than a layer based prioritization. So i would like the ability to update raycast hit prioritization logic based on application scenario on the fly.

[...]

PR #4678, opened 2019-05-31
Configurable raycaster

kircher1:
## Overview

Introduces a system type for a Raycast Provider. Allows for a IMixedRealityRaycastProvider to be specified as part of the InputSystem configuration.

NewConfig
_Picture is slightly outdated: Interface name has been changed from Raycast**er** to Raycast._

For objects that are not part of the Unity UI or physics collider systems, this can allow for handling raycasting of those objects. 

As an example, I've been able to use this new configuration to implement pointer handling on the map object provided by [Maps SDK, a Microsoft Garage project](https://github.com/Microsoft/MapsSDK-Unity). The map does not use Unity's physics collider system, but it does provide an API to raycast it.

PointAndDrag

## Changes
- Fixes: #4325.


Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycasterProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// The resulting hit information from an IMixedRealityRaycasterProvider.
+    /// </summary>
+    public struct RaycasterHit
+    {
+        public float Distance;
```
Railboy: I'm confused as to why we need a RaycastHit-analogue since it doesn't add anything new. Why not stick with RaycastHits?

This struct omits a bunch of useful fields like the collider, the rigidbody. triangle index, texture coords - I use this stuff all the time.

kircher1: Good point, those should be added back to have parity with the existing RaycastHitInfo from Unity. Will do that in a future iteration. 

I'm also consider using the Unity property names from RaycastHitInfo exactly, like 'distance' instead of 'Distance', to make the change less breaking.

kircher1: And more broadly: The overall reason for the new RaycasterHit struct is that Unity's RaycastHitInfo is not extensible. Only the position, normal, and distance can be set on it, and important properties that MRTK relies on, like the 'transform' or 'collider', are not settable.

This struct allows for those types to be specified.

Railboy: >This struct allows for those types to be specified.

Ah that makes sense.

In that case, maybe we should go with an extendable struct with a 1:1 parity, including assignment overrides so you can assign all the values from a RaycastHit to a RaycasterHit with one = operation and/or return a RaycastHit as a RaycasterHit.

Prithiaga: Good that this change is happening. We would also want to extend the struct to add additional hit data apart from the unity hit data. This change would be really helpful.

kircher1: To make the hit data extensible, the simplest thing to do would be making the RaycasterHit struct a class so that it can be extended and eventually casted back to the type you expect. The downside in doing that is now the allocation is managed by the GC.

The other option I see is to add some sort of property bag to the struct, which I would do if there was some previous example of this in MRTK. I haven't seen that concept used yet though.

I'm leaning towards simply transforming this struct to a class, so that you can extend and cast it to the expected type. Let me know if that works or if there are other suggestions.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+
+using Microsoft.MixedReality.Toolkit.Physics;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// The resulting hit information from an IMixedRealityRaycastProvider.
+    /// </summary>
+    public class MixedRealityRaycastHit
```
StephenHodgson: This should likely be in it's own file

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycasterProvider.cs
Prithiaga: Is there a way that we could provide our own struct similar to raycaster? Changing it to class is going to be bad on GC as the objects are created in Update loop. If the struct is changed to class, then caching and reusing the references need to be done.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
kircher1: Moved this out to a separate file in Definitions.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycasterProvider.cs
kircher1: While not ideal from the garbage collector perspective, I don't have a better idea to solve this. I'm not going to implement pooling/caching for this type either because it's unlikely the potential additional GC pressure will be an issue, especially given that the event data generated by MRTK is all class based as well.

I'm open to other approaches to make this more performant, but using a class here seems like the best trade-off in order to extend this with more data.

That, or we disallow extension of this type and leave it as a struct. It sounds like for your scenario you plan on implementing a custom FocusProvider as well? Maybe there is some way you can get the additional data from the raycast without having to extend this class. Could it be queried from the GameObject/component directly?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -791,8 +791,8 @@ private void UpdatePointer(PointerData pointer)
                     LayerMask[] prioritizedLayerMasks = (pointer.Pointer.PrioritizedLayerMasksOverride ?? FocusLayerMasks);
 
                     // Perform raycast to determine focused object
-                    hitResult3d.Clear();
```
luis-valverde-ms: Is this not necessary anymore?

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Physics;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Implements the Raycast Provider for handling raycasts into the scene.
```
luis-valverde-ms: Better doc here

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public interface IMixedRealityRaycastProvider : IMixedRealityDataProvider
+    {
+        /// <summary>
+        /// Simple raycasts each physics <see cref="Microsoft.MixedReality.Toolkit.Physics.RayStep"/>.
+        /// </summary>
+        /// <returns>Whether or not the raycast hit something.</returns>
+        bool RaycastSimplePhysicsStep(RayStep step, LayerMask[] prioritizedLayerMasks, out MixedRealityRaycastHit hitInfo);
```
luis-valverde-ms: To keep the interface simple, I would remove the versions that don't receive maxDistance. If we think it is a common use case to want to use the step length as max distance, we can implement that as an extension. See this for a reference: https://stackoverflow.com/a/2567204/3876392

luis-valverde-ms: About the methods names, I would rather shorten them to `Raycast()` and `SphereCast()`.

Assets/MixedRealityToolkit.Services/InputSystem/RaycastProvider.cs
```diff
@@ -0,0 +1,65 @@
+﻿using Microsoft.MixedReality.Toolkit.Physics;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    public class RaycastProvider : BaseDataProvider, IMixedRealityRaycastProvider
```
luis-valverde-ms: I'd rather call this `DefaultRaycastProvider` to make it clear this is the default implementation.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityRaycastHit.cs
```diff
@@ -0,0 +1,41 @@
...
+            barycentricCoordinate = hitInfo.barycentricCoordinate;
+            distance = hitInfo.distance;
+            triangleIndex = hitInfo.triangleIndex;
+            textureCoord = hitInfo.textureCoord;
+            textureCoord2 = hitInfo.textureCoord2;
+            transform = hitInfo.transform;
+            lightmapCoord = hitInfo.lightmapCoord;
+        }
+
+        public static implicit operator MixedRealityRaycastHit(RaycastHit hitInfo)
```
luis-valverde-ms: Is this so we can do `(MixedRealityRaycastHit) aHitInfo`? I'd rather force users to cache the converted value by removing this operator, as it is a relatively expensive casting operation.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
kircher1: Must have removed by mistake. Thanks for catching.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
kircher1: The overloaded variants aren't used currently, so I'll remove.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
kircher1: Will clean up

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycasterProvider.cs
Prithiaga: Though currently I have written a custom FocusProvider, it would be good if we use raycaster class. But right now, I added a property in FocusProvider for additional hit data. The data that we use is more 3D model specific like triangleId, instanceId etc. We cannot use gameobjects, because we render the game objects by doing mesh combining, whereas our hit happens to individual meshes in the model. 
For now, class could be used. I would update if I could think of a better approach. Thanks!

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Physics;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Interface to handle raycasts into the scene.
```
luis-valverde-ms: Add a bit more information to make it clear to implementers where this interface is used by MRTK. I'd at least mention that FocusProvider will use it to perform raycast and sphere cast queries for pointers.

kircher1: Added a blurb about FocusProvider.

davidkline-ms:
Question: Does this change imply that all input system profiles will need to be updated to specify the a raycast implementation? It feels like that is the case.

Thx!

kircher1:
> Question: Does this change imply that all input system profiles will need to be updated to specify the a raycast implementation? It feels like that is the case.
> Thx!

Yes, custom input profiles would require updating after this change in orer to specify the raycast provider.

The default input system config has already been updated, and it brings up a good point: I will double-check the examples for other custom input systems and update them as well.

I can see the concern here from a breaking change perspective... Although if no raycaster has been specified in the config, we could attempt to automatically assign the DefaultRaycasterProvider: Like FocusProvider, this RaycastProvider would be required, i.e. error message is logged if it is not specified in the config. So for these types, I'm not sure why it is alllowed to specify "(None)" as a config option. Maybe we should always fall back to the default implementation for these types if none has been explicitly specified.

davidkline-ms:
Thanks @kircher1! Since this is something that a dev would just need to add to a custom profile, as long as there is a clear error message it might be ok to require specifying a raycast implementation.

kircher1:
Yep @davidkline-ms, the error message is logged from MixedRealityToolkit.cs when registering types. If type is missing or invalid, it shows in the log. (Same as FocusProvider).

The latest changes update remaining custom input profiles in MRTK to use the default RaycastProvider.

Assets/MixedRealityToolkit.Services/InputSystem/DefaultRaycastProvider.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Microsoft.MixedReality.Toolkit.Physics;
```
julenka: Reminder to include these lines:

```
// Copyright (c) Microsoft Corporation. All rights reserved.
```

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -200,7 +200,7 @@ public void Clear()
             /// </summary>
             public void Set(GameObject hitObject, Vector3 hitPointOnObject, Vector4 hitNormalOnObject, RayStep ray, int rayStepIndex, float rayDistance)
             {
-                raycastHit = default(RaycastHit);
+                raycastHit = default;
```
julenka: Will this compile with the .NET backend? I recall a while back having trouble using the default value expression because that's only available in [C# 7.1 IIRC](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/default-value-expressions)

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// in a custom implementation, the raycast does not have to rely only on Unity-based Colliders to provide hit results, e.g. a
+    /// GameObject may use a different mechanism for raycasting, and with a custom implementation, it could be included in the hit result.
+    /// </remarks>
+    public interface IMixedRealityRaycastProvider : IMixedRealityDataProvider
+    {
+        /// <summary>
+        /// Performs a raycast using the specified <see cref="Microsoft.MixedReality.Toolkit.Physics.RayStep"/>.
+        /// </summary>
+        /// <returns>Whether or not the raycast hit something.</returns>
+        bool Raycast(RayStep step, LayerMask[] prioritizedLayerMasks, out MixedRealityRaycastHit hitInfo);
```
ritijain: One more level of configuration that can be helpful is being able to configure our own struct for capturing the RaycastHitResult. Some of our apps also captured the hit type - Hologram, SR, EmptySpace. I am not sure if it makes sense to capture this information in the default MixedRealityRaycastHit struct or just have the option of an interface so that developers creating their custom raycastproviders also have the option of custom raycasthit results.

kircher1: Since the MixedRealityRaycastHit is a class, it could be extended to provide additional hit data. In a custom FocusProvider implementation, you could then cast the hit back to the type you expect and get the extra data off of it.

kircher1:
Looking into test failures.

kircher1:
@wiwei / @davidkline-ms, latest changes address the test failure. Can you kick off the tests? Thanks!

kircher1:
Tests have been fixed.

Issue #4683, opened 2019-06-02
NullPointer in PokePointer.OnPostSceneQuery after pressing a button


julenka:
## Overview
Open the HandInteractionExamples scene and press any button with finger, then move your hand to the right of the button.

After you move your hand off, you will get continuous null pointers, with this call stack:

This is most likely a regression from #4263 

```
NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.Input.PokePointer.OnPostSceneQuery () (at Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs:107)
Microsoft.MixedReality.Toolkit.Input.FocusProvider.UpdatePointer (Microsoft.MixedReality.Toolkit.Input.FocusProvider+PointerData pointer) (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:825)
Microsoft.MixedReality.Toolkit.Input.FocusProvider.UpdatePointers () (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:736)
Microsoft.MixedReality.Toolkit.Input.FocusProvider.Update () (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:475)
```

julenka:
Urgency now because the HandInteractionExamples scene cannot be run.

julenka:
Fixed by #4684

PR #4684, opened 2019-06-02
Fix nullptr in PokePointer when you press any button

julenka:
Fixes: #4683

Also a couple additional fixes:
- Fixes a failing PlayMode button test by passing in updated prefab file name
- Remove some extra print statements from Interactable.


julenka:
Don't complete this PR yet. I'm working on a playmode test that repros the P0 bug (press button with hand and then move hand off of button) that I'd like to check in as part of this PR.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -102,7 +102,7 @@ public override void OnPostSceneQuery()
                 return;
             }
 
-            if (Result?.CurrentPointerTarget != null)
+            if (Result?.CurrentPointerTarget != null && closestProximityTouchable != null)
```
julenka: @MenelvagorMilsom this is the actual fix :-) the other changes in this PR are to fix a playmode test, remove some extra logging, and then add a playmode test that repros this issue so we don't hit it again.

julenka:
@thalbern and @MenelvagorMilsom I'm going to complete this PR because it fixes P0 issue, however keeping it open so we can discuss the tests and the fix. 

I'm not sure if my fix that I mention above (add a null check) has unexpected side effects, @MenelvagorMilsom would love for you to check that in case we need a follow up. I did test the hand interaction examples and touchables examples scenes and everything seemed fine.

I think there may be a better way to write the hand simulation for the tests, please don't consider this final. For example I was thinking to have a SimulatedHand class that would drive a simulated hand for the tests. It would look something like this:

```
class SimulatedHand
public SimulatedHand(Handedness h)
// Shows the hand in the open state
public IEnumerator Show()
// Hides the hand
public IEnumerator Hide()
public IEnumerator SetHandPose(gestureId)

// Couple methods below to move a hand over some number of frames
public IEnumerator MoveTo(fromPosition, toPosition, overHowManyFrames);
public IEnumerator MoveTo(toPosition, overHowManyFrames);
public IEnumerator SetPosition(toPoPosition);
```

julenka:
x86 master build is passing locally for me. Perhaps this is an intermittent failure?

julenka:
Update: I removed the PlayMode tests from this PR since they were breaking CI. will put them in a separate PR.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
MenelvagorMilsom: Makes sense to me! Thanks for catching this.

PR #4685, opened 2019-06-03
Add PlayMode test that presses button with simulated hand

julenka:
## Overview
Adds a playmode test that would have prevented issues #4683 from occurring. The test uses a simulated hand to press a button and verifies that the button gets pressed.

Fixes: #4683

## Changes
- Add couple basic utility methods for moving the simulated hand around in a playmode test
- Also remove SimplePlayModeTests, since they are not really testing anything and take 10 seconds to run. These were valuable at the beginning to prove play mode tests can run and as an example to follow.

## Verification
- Verified that the test fails without the fix for 4683 and then passes with the fix.


julenka:
@lukastoenneMS and @thalbern I think there may be a better way to write the hand simulation for the tests, please don't consider this final. For example I was thinking to have a SimulatedHand class that would drive a simulated hand for the tests. It would look something like this:

```
class SimulatedHand
public SimulatedHand(Handedness h)
// Shows the hand in the open state
public IEnumerator Show()
// Hides the hand
public IEnumerator Hide()
public IEnumerator SetHandPose(gestureId)

// Couple methods below to move a hand over some number of frames
public IEnumerator MoveTo(fromPosition, toPosition, overHowManyFrames);
public IEnumerator MoveTo(toPosition, overHowManyFrames);
public IEnumerator SetPosition(toPoPosition);
```

Not doing that in this PR in case Lukas is already working on this. Lukas, do you already have something like this that you're working on?

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -1,23 +1,82 @@
...
+                float t = 1.0f / numSteps * i;
+                Vector3 handPos = Vector3.Lerp(startPos, endPos, t);
+                var handDataGenerator = GenerateHandPose(
+                        gestureId,
+                        handedness,
+                        handPos);
+                Debug.Assert(handedness == Handedness.Right || handedness == Handedness.Left, "handedness must be either right or left");
+                SimulatedHandData toUpdate = handedness == Handedness.Right ? inputSimulationService.HandDataRight : inputSimulationService.HandDataLeft;
+                bool isPinching = gestureId == ArticulatedHandPose.GestureId.Grab || gestureId == ArticulatedHandPose.GestureId.Pinch || gestureId == ArticulatedHandPose.GestureId.PinchSteadyWrist;
+                inputSimulationService.HandDataRight.Update(true, false, handDataGenerator);
```
julenka: this should actually use isPinching

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -1,23 +1,82 @@
...
+        {
+            for (int i = 0; i < numSteps; i++)
+            {
+                float t = 1.0f / numSteps * i;
+                Vector3 handPos = Vector3.Lerp(startPos, endPos, t);
+                var handDataGenerator = GenerateHandPose(
+                        gestureId,
+                        handedness,
+                        handPos);
+                Debug.Assert(handedness == Handedness.Right || handedness == Handedness.Left, "handedness must be either right or left");
```
julenka: Move this out of the for loop

JonathanPalmerGD:
I cloned the source repo and ran all the tests. 
They all pass but I'm seeing many 'Unable to find ITestDataProvider service.' issues.

image

The first error is related to 'No Mixed Reality Configuration Profile found, cannot initialize the Mixed Reality Toolkit'.
I think that first error is a result of how when it starts running the tests, it starts an Untitled Scene with no content.

These errors might be harmless.

julenka:
> I cloned the source repo and ran all the tests.
> They all pass but I'm seeing many 'Unable to find ITestDataProvider service.' issues.
> 
> image
> 
> The first error is related to 'No Mixed Reality Configuration Profile found, cannot initialize the Mixed Reality Toolkit'.
> I think that first error is a result of how when it starts running the tests, it starts an Untitled Scene with no content.
> 
> These errors might be harmless.

Hm, I'm not able to repro locally. I've seen this error when I pull code without restarting Unity. Could you please try restarting Unity?

keveleigh:
>These errors might be harmless.

Yeah, I believe this is just output from some of the edit mode tests that specifically test cases where things shouldn't be findable (so the errors are expected, and specifically what the test wants to cause).

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -1,23 +1,82 @@
...
+        public static InputSimulationService GetInputSimulationService()
+        {
+            IMixedRealityInputSystem inputSystem = GetInputSystem();
+            InputSimulationService inputSimulationService = (inputSystem as IMixedRealityDataProviderAccess).GetDataProvider<InputSimulationService>();
+            Assert.IsNotNull(inputSimulationService, "InputSimulationService is null!");
+            inputSimulationService.UserInputEnabled = false;
+            return inputSimulationService;
+        }
+
+        internal static IEnumerator MoveHandFromTo(Vector3 startPos, Vector3 endPos, int numSteps, ArticulatedHandPose.GestureId gestureId, Handedness handedness, InputSimulationService inputSimulationService)
```
Jarodshow: Could we make all these async?

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
lukastoenneMS: Yes, the way we determine if simulated hands are pinching/grabbing is not great atm (see also [SimulatedHandState.GestureBlending](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/ae0b1467b564c5b647c5190c84fe725b8ffd6f6a/Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedHandDataProvider.cs#L64). I'm not sure of the best way to tag hand poses for these actions, perhaps add a code comment so we improve it in both places.

lukastoenneMS:
> 
> 
> @lukastoenneMS and @thalbern I think there may be a better way to write the hand simulation for the tests, please don't consider this final. For example I was thinking to have a SimulatedHand class that would drive a simulated hand for the tests. It would look something like this:
> 
> ```
> class SimulatedHand
> public SimulatedHand(Handedness h)
> // Shows the hand in the open state
> public IEnumerator Show()
> // Hides the hand
> public IEnumerator Hide()
> public IEnumerator SetHandPose(gestureId)
> 
> // Couple methods below to move a hand over some number of frames
> public IEnumerator MoveTo(fromPosition, toPosition, overHowManyFrames);
> public IEnumerator MoveTo(toPosition, overHowManyFrames);
> public IEnumerator SetPosition(toPoPosition);
> ```
> 
> Not doing that in this PR in case Lukas is already working on this. Lukas, do you already have something like this that you're working on?

That's fine with me, should be easy to improve this code later as long as it lives only as test utils (i.e. discourage users from building on top of it).

I'm using essentially the same approach for playing back hand animations, using animation curves for position+rotation for each hand joint and then reconstructing simulated hand data from that. The difference is that the recorded input does not use the ArticulatedHandPose to fill in the joint array, but contains animation for each world space joint pose.

JonathanPalmerGD:
I restarted Unity and encountered these problems 2 more times.
I have one last thought on what it could be to validate that there might be a hidden problem

JonathanPalmerGD:
The problem still occurs.

I initially opened the project in 2018.3.7 (instead of 2018.3.9), I checked the release notes and there was a test runner fix. It didn't prove to be the culprit.

This project I set up by just cloning your git repo on that specific branch. That might be the source of it.
It doesn't seem like a common or blocking issue, so proceed as you see appropriate.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
julenka: Hi Jarod, could you please explain why it might be a good idea to do that here? All of the PlayMode tests are currently just using Coroutines without async implementation, so for now I kept it using Coroutines for consistency. However, I'd like to understand why using async would be a better approach. I normally use async when I'm writing asynchronous code such as reading large files or making web / network requests, but not normally when doing playmode tests / in Unity. Perhaps I'm missing some untapped capabilities of using async instead of coroutines in Unity. Definitely willing to learn!

julenka:
> The problem still occurs.
> 
> I initially opened the project in 2018.3.7 (instead of 2018.3.9), I checked the release notes and there was a test runner fix. It didn't prove to be the culprit. I fully recloned the project (with the .zip approach) and first opened it in 2018.3.9 and the problem persists.
> 
> This project I set up by just cloning your git repo on that specific branch. That might be the source of it.
> It doesn't seem like a common or blocking issue, so proceed as you see appropriate.
> 
> Here's the call stack:
> image

Thanks Jonathan. Like Kurtis said I think those are errors from the Edit Mode tests and can be ignored. The PlayMode tests themselves have been passing for me.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -1,23 +1,82 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+#if !WINDOWS_UWP
+// When the .NET scripting backend is enabled and C# projects are built
```
keveleigh: Do the test assemblies need to be included in the build, or can we mark them as editor-only to potentially work around this issue?

julenka: @wiwei  did you try to just modify the .asmdef files to only include these projects in the editor? Was there an issue that you ran into that prevented you from doing this? Reading your comment it seems like there was a build issue when building for UWP .NET backend. I assumed I needed to do the same ifdef here.

julenka: For now I'll check this in as is and maybe we can solve this more elegantly with an asmdef after we hear from will.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
Jarodshow: > Perhaps I'm missing some untapped capabilities of using async instead of coroutines in Unity.

Just one off the top of my head, the ability to return types super easily. While that doesn't have any bearing on this situation, I think that async code in general is much easier to follow. So making these async would not be a better approach *functionally*.

PR #4686, opened 2019-06-03
Merge latest stabilization changes into _development

wiwei:
The new hand-based solver/menu was put into the stabilization branch - we want this to go into the _development branch as well (without waiting for things to hit release first)


julenka:
Looks like mrtk_pr is passing so I'll complete this. I'm not sure why stabilization_pr is not passing though.

Issue #4601, opened 2019-05-29
Input Simulation Does not work on Mac


julenka:
## Describe the bug
If you open an MRTK-Enabled scene on Mac OSX and hit play, the camera will not move when you WASD and hands do not show up.

The expected behavior is that input simulation should work as it does on Windows

I was able to work around this by checking "Mac Standalone" in the Platforms for Input Simulation Service Input Providers: 

![image](https://user-images.githubusercontent.com/168492/58529939-91ed4d80-8191-11e9-93bf-49d19956d4df.png)

## Your Setup (please complete the following information)

- Unity Version 2018.3.14f1
- MRTK Version mrtk_development

## Target Platform (please complete the following information)

- Mac OSX Editor

julenka:
After I apply the workaround above I'm able to use input simulation just fine.

djee-ms:
I can confirm the bug and the workaround, thanks @julenka for the tip!

PR #4691, opened 2019-06-03
add additional supportedplatform enum values for future / expansion use

davidkline-ms:
## Overview
Add iOS, Web and Lumin to the SupportedPlatform enums. This will allow future platform extensions to be created for these without needing to recompile the core assemblies.

This is related to the work by @Alexees (#4358) that has been moved to being under consideration for the next major MRTK release.

Also added Mac and Linux editor platforms to address #4601.

image

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityInputSystemProfile.asset
+++ b/Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityInputSystemProfile.asset
@@ -66,7 +66,7 @@ MonoBehaviour:
       reference: Microsoft.MixedReality.Toolkit.Input.InputSimulationService, Microsoft.MixedReality.Toolkit.Services.InputSimulation.Editor
     componentName: Input Simulation Service
     priority: 0
-    runtimePlatform: 16
+    runtimePlatform: 208
     deviceManagerProfile: {fileID: 11400000, guid: 41478039094d47641bf4e09c20e61a5a,
       type: 2}
   focusProviderType:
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit/Definitions/Utilities/SupportedPlatforms.cs
```diff
@@ -16,6 +16,9 @@ public enum SupportedPlatforms
         LinuxStandalone = 1 << 2,
         WindowsUniversal = 1 << 3,
         WindowsEditor = 1 << 4,
```
keveleigh: Do we want to add `MacEditor` and `LinuxEditor` at this time as well?

davidkline-ms: we might indeed

davidkline-ms: added

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityInputSystemProfile.asset
```diff
@@ -66,7 +66,7 @@ MonoBehaviour:
       reference: Microsoft.MixedReality.Toolkit.Input.InputSimulationService, Microsoft.MixedReality.Toolkit.Services.InputSimulation.Editor
     componentName: Input Simulation Service
     priority: 0
-    runtimePlatform: 16
+    runtimePlatform: 208
```
julenka: Is this change intentional?

julenka: Ah, yes it is: it's binary OR of Linux, Mac, Windows editors. (1 << 4) | (1 << 6) | (1 << 7). Thank you!

Issue #4598, opened 2019-05-29
Interactable Voice Command feature is confusing


killerantz:
## Describe the bug

Interactable has a field for Voice Command, but entering a value there does nothing when running the experience. Older versions of Interactable had the ability to setup a voice command directly from Interactable. Now voice commands need to be setup using the Speech Commands Profile. This should be documented.

Now the Voice Command field just filters the Speech Command that are spoken while the Interaction has focus. This is nice, but should be a drop down instead of a string field to avoid typo bugs.

The Requires Focus property is not hooked up to anything. The ISpeechInputHandler in Interactable does check if Requires Focus is checked and if the Interactable has Focus, but the ISpeechInputHandler does not receive any events unless the Interactable has focus anyway. This property used to let make the Interactable listen globally for a speech commands, but this functionality is not working any more, even with IsGlobal checked. Is there a way to register as a global listener for SpeechCommands?

## To reproduce

1. Create an object with an Interactable.
2. Add a string to the Voice Commands field.
3. observe the voice command does not work.
4. Go to the SpeedCommandsProfile and add a new voice command
5. Enter the voice command from the profile as a string in the Voice Command field of Interactable
6. Observe the Voice Command does work with focus.
7. Uncheck Requires Focus on the Interactable.
8. Observe that the voice command still requires focus.

## Expected behavior

It's clear where and how to setup voice commands
Voice Commands should be a drop down if they are setup in the Speed Command Profile
Change name to Speech Commands for consistency
Requires Focus property should enable a global listener for speed commands if unchecked.

## Screenshots

![image](https://user-images.githubusercontent.com/22085553/58523339-f223c580-8178-11e9-9470-b29ae2890810.png)

julenka:
merged.

PR #4695, opened 2019-06-04
Fix/interactable voice commands

killerantz:
## Overview
Resolves issue #4598 

- Added a drop-down menu pulling speech commands from the MRTK Speed Commands Profile. There is a notice to direct users to the Speed Commands profile if no speech commands are selected to help with visibility.

- Made sure Requires Focus works. When selected, we get normal behavior where an Interactable must have focus to receive a speech command event, when unchecked the Interactable registers at a global listener, this was not hooked up before. The filtering for IsGlobal (referring to pointer events) and Requires Focus (referring to speech commands) should not cause extra clicks to occur even though both toggles essentially register or unregister the Interactable as a global listener.
-- These values only work if they are set before runtime or OnEnable is called. The Interactable does not currently support runtime registering and unregistering.

- Updated terms to use Speech Command, except for the Voice Command state that was setup previously.

- Updated documentation and supporting images to reflect these and other recent changes.


julenka:
> These values only work if they are set before runtime or OnEnable is called. The Interactable does not currently support runtime registering and unregistering.

Do we have a bug to track this?

killerantz:
Added a bug for IsGlobal and Requires Focus only checking on enable. #4703

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -214,7 +214,7 @@ protected virtual void Awake()
 
         private void OnEnable()
         {
-            if (IsGlobal)
+            if (IsGlobal || !RequiresFocus)
```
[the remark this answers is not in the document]
killerantz: It's so we can have something that listens to a global voice command, but not a global click or press. I could have one button for menu on a toolbar, which needs to be focused to click, but a global voice command that I can speak at any time "menu" to show the menu. Without these two properties menu would be clicked whenever another button in the scene was clicked.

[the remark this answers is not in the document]
killerantz: Requires Focus only applies to Speech Commands, I tried to make up a different property name to avoid confusion, though it is confusing :) I'm hoping the context of the Inspector helps.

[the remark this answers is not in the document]
julenka: Okay, I understand. This feels potentially confusing to an external developer. Imagine if you're approaching interactable for the first time and you'd see two fields -- requires focus and Is Global. I would be confused for sure. Even as a developer reading the code it's a bit confusing.

Maybe it would help to instead say "speech command requires focus" so that it's clear that this bool is specifically for speech commands.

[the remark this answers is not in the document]
julenka: My more general aside about the complexity doesn't need to be addressed in this PR, just giving some feedback that I think we could actually re-use some of the code that already exists in MRTK to make Interactable simpler.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -677,7 +707,58 @@ protected void SetupEventOptions()
...
+                return true;
+            }
+            else
+            {
+                keywords = null;
+                return false;
+            }
+        }
+
+		/// <summary>
```
julenka: More issues with incorrect spacing.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -157,15 +156,43 @@ public virtual void RenderCustomInspector()
...
 
+            EditorGUI.indentLevel = indentOnSectionStart + 1;												 
             SerializedProperty isGlobal = serializedObject.FindProperty("IsGlobal");
             EditorGUILayout.PropertyField(isGlobal, new GUIContent("Is Global", "Like a modal, does not require focus"));
-
+            EditorGUI.indentLevel = indentOnSectionStart;
             SerializedProperty voiceCommands = serializedObject.FindProperty("VoiceCommand");
-            EditorGUILayout.PropertyField(voiceCommands, new GUIContent("Voice Command", "A voice command to trigger the click event"));
 
+			if(speechKeywords == null)
```
julenka: Spacing

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -315,6 +334,14 @@ protected virtual void Update()
             }
 
             lastState = StateManager.CurrentState();
+            
+            if (requiresFocusValueCheck != RequiresFocus || isGlobalValueCheck != IsGlobal)
```
julenka: I'm now realizing that RequiresFocus was documented, but I was confused because I didn't see it in our [API docs](https://microsoft.github.io/MixedRealityToolkit-Unity/api/Microsoft.MixedReality.Toolkit.UI.Interactable.html). I think if you use a format like this then the docs will populate, right @thalbern?

```
        /// <summary>
        /// Does the voice command require this to have focus?
        /// </summary>
        public bool RequiresFocus = true;
```

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -678,6 +699,66 @@ protected void SetupEventOptions()
...
+        /// Look for speech commands in the MRTK Speech Command profile
+        /// </summary>
+        /// <param name="keywords"></param>
+        /// <returns></returns>
+        protected bool TryGetSpeechKeywords(out string[] keywords)
+        {
+            List<string> keys = new List<string>();
+            
+            if (MixedRealityToolkit.Instance && (MixedRealityToolkit.Instance.ActiveProfile.IsInputSystemEnabled ||
+                MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.SpeechCommandsProfile != null ||
```
julenka: @davidkline-ms  is this the proper / recommended way to get MRTK profiles now?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -678,6 +699,66 @@ protected void SetupEventOptions()
...
+                MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.SpeechCommandsProfile.SpeechCommands.Length > 0))
+            {
+                int keywordCount = MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.SpeechCommandsProfile.SpeechCommands.Length;
+
+                for (var i = 0; i < keywordCount; i++)
+                {
+                    keys.Add(MixedRealityToolkit.Instance.ActiveProfile.InputSystemProfile.SpeechCommandsProfile.SpeechCommands[i].Keyword);
+                }
+
+                keys.Insert(0, "None");
```
julenka: What do we do if the user selects "none" for the speech commands here?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -678,6 +699,66 @@ protected void SetupEventOptions()
...
+            }
+            else
+            {
+                keywords = null;
+                return false;
+            }
+        }
+        
+        /// <summary>
+        /// Get the index of the speech keyword array item based on it's name, pop-up field helper
```
julenka: grammer nit: its, not it's

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -678,6 +699,66 @@ protected void SetupEventOptions()
...
+        /// </summary>
+        /// <param name="option"></param>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        public int KeywordLookup(string option, string[] options)
```
julenka: You can delete this method and just use [Array.IndexOf](https://docs.microsoft.com/en-us/dotnet/api/system.array.indexof?redirectedfrom=MSDN&view=netframework-4.8#System_Array_IndexOf_System_Array_System_Object_) instead. `Array.IndexOf(options, option);`. Also, any reason you would want this to be public?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -157,15 +156,43 @@ public virtual void RenderCustomInspector()
...
 
+			if(speechKeywords == null)
+            {
+                GUI.enabled = false;
+                EditorGUILayout.Popup("Speech Command", 0, new string[] { "Missing Speech Commands" });
+                GUI.enabled = true;
+            }
+            else
+            {
+                int currentIndex = KeywordLookup(voiceCommands.stringValue, speechKeywords);
```
julenka: Instead, do `currentIndex = Array.IndexOf(speechKeywords, voiceCommands.stringValue)`

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: We set the VoiceCommand property to an empty string. There's code in Interactable and the inspector that checks for an empty string to see if listening for voice commands is active or not.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: There's a slight difference here than Array.IndexOf. This starts at item one in the array because "None" was added at item zero. If the speech command is not in the list, I want it to be zero, which is "None".

I was concerned that someone could create a speech command called "None" for some scenario, like choosing the amount of something via speech command. In that situation, a created speech command would match the inserted "None" value at index zero and turn off  listening for voice commands. It would be weird that there would be two "None" values in the list, but that's an extreme edge case that would not break anything.

I wanted an easy way to turn off speech command handling without adding another toggle to enable/disable it. But any value we insert could potentially be a speech command that someone could create. If this is too weird, I can put in the toggle :)

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: See above comment.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: Should I follow this pattern?
``` C#
if (!MixedRealityToolkit.ConfirmInitialized() || !MixedRealityToolkit.Instance.HasActiveProfile)
            {
                descriptionsArray = null;
                return false;
            }

            MixedRealityInputAction[] actions = InputSystem.InputSystemProfile.InputActionsProfile.InputActions;
```

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: I updated a ton of single line comments to be summary comments. Hopefully this will help with feature discovery.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
julenka: I would say yes, though looking to @Railboy and @davidkline-ms for the blessing since they own that system. But the ConfirmInitialized() has been the pattern I've been following.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: Thanks!

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
julenka: I see. Up to you. Just wanted to make sure you were aware of the tools available. If you wanted to use LINQ you could also do keywords.Skip(1).IndexOf(...) or similar to skip the first element. 

Is it a fair assumption to always assume that the first element is keywords.None? If "None" is not even a valid keyword, why is it even in the keywords list?

It's not a huge deal, just wondering if I can provide some suggestions that could help simplify the code. It was not obvious to me that the first element in the list always represents "no keyword". I would expect if no keywords are registered for the keywords list to be empty.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: Updated to use similar pattern for getting input actions.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: Thanks for the suggestions, I don't use LINQ very much and should learn more about it.

Leaving the first item blank or just spaces does not work in the editor popup, it just ignores the item without valid characters in the drop down so it doesn't should up as an option. I replaced None with -No Selection- as the first item, it's more clear and gives the user the feeling that they can select it as an option to turn the feature off.

luis-valverde-ms:
Pinging @Railboy and @davidkline-ms for their eyes on this.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -212,6 +339,39 @@ public static bool TryGetInputActions(out string[] descriptionsArray)
...
+                return false;
+            }
+
+            List<string> keys = new List<string>();
+            for (var i = 0; i < commands.Length; i++)
+            {
+                keys.Add(commands[i].Keyword);
+            }
+
+            keys.Insert(0, "-No Selection-");
```
Railboy: Super-nit:

```suggestion
            keys.Insert(0, "(No Selection)");
```

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -157,14 +157,46 @@ public virtual void RenderCustomInspector()
...
+                    currentIndex = EditorGUI.Popup(position, label.text, currentIndex, speechKeywords);
+
+                    if (currentIndex > 0)
+                    {
+                        voiceCommands.stringValue = speechKeywords[currentIndex];
+                    }
+                    else
+                    {
+                        voiceCommands.stringValue = "";
+                        InspectorUIUtility.DrawNotice("Create speech commands in the MRTK/Input/Speech Commands Profile");
```
Railboy: I was a little confused by this message showing up when I set the speech command to 'No Selection' - it should probably only show up when there are no options.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -157,14 +157,46 @@ public virtual void RenderCustomInspector()
...
+                //look for items in the sppech commands list that match the voiceCommands string
+                // this string should be empty if we are not listening to speech commands
+                // will return zero if empty, to match the inserted off value.
+                int currentIndex = SpeechKeywordLookup(voiceCommands.stringValue, speechKeywords);
+
+                position = EditorGUILayout.GetControlRect();
+                GUIContent label = new GUIContent("Speech Command", "speech keyword to trigger Interactable");
+                EditorGUI.BeginProperty(position, label, voiceCommands);
+                {
+                    currentIndex = EditorGUI.Popup(position, label.text, currentIndex, speechKeywords);
```
Railboy: >These values only work if they are set before runtime or OnEnable is called. The Interactable does not currently support runtime registering and unregistering.

We should put the dropdown in a disabled block during play mode.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: That makes more sense. I'll also make sure the tooltip is clear where the speech command list is coming from at least.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: I disabled some of the settings during play mode, those that should only be edited in edit mode.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -259,7 +439,19 @@ private void OnEnable()
...
+            {
+                InputSystem.Unregister(gameObject);
+            }
+        }
+
+        private void RegisterGlobalListeners(bool register)
+        {
+            if (register)
+            {
+                InputSystem.Register(gameObject);
```
googlan: Note: this will need to be `RegisterHandler<IMixedRealitySpeechHandler>`. Same with Unregister.

julenka:
Just waiting on @killerantz to fix googlan's one comment and then i'll run pr validation and get this in.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: This should be updated to target the correct input handler types.

Issue #4647, opened 2019-05-30
Add sample for determining if a user is ET calibrated and show a notification


sostel:
## Describe the problem
Based on customer feedback, it is not clear to the user when or for which apps Eye Tracking (ET) is relevant and whether the user should be ET calibrated. In case the user skipped the Eye Tracking calibration initially, it would be nice to show a notification to the user to indicate that the user would have a better experience in this app if being ET calibrated. 

## Describe the solution you'd like
Easy example to allow developers to check if a user is calibrated and if not to trigger a notification.

PR #4697, opened 2019-06-04
Add sample for determining if a user is ET calibrated and show a notification

sostel:
## Overview
Addresses issue #4647. Several devs have mentioned that it would be nice to get a sample of how to determine whether as user is eye calibrated and to show a notification in case the apps requires eye tracking to work properly. 

## Changes
+ Several changes to allow for communicating through whether the current eye calibration is valid
+ Sample for how to use it in EyeCalibrationChecker.cs
+ A few minor clean ups in other files 

image


Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Scripts/Utils/EyeCalibrationChecker.cs
```diff
@@ -0,0 +1,65 @@
...
+                    MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem);
+                }
+                return inputSystem;
+            }
+        }
+
+        // Update is called once per frame
+        void Update()
+        {
+            bool? calibrStatus = InputSystem?.EyeGazeProvider?.IsEyeCalibrationValid;
```
wiwei: Nit: 
```suggestion
            bool? calibrationStatus = InputSystem?.EyeGazeProvider?.IsEyeCalibrationValid;
```

wiwei: May as well keep consistent with fully spelling things out (because you already do this for the calibration status of the local member variable)

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
```diff
@@ -98,6 +98,7 @@ public override void Update()
             if (profile.SimulateEyePosition)
             {
                 InputSystem?.EyeGazeProvider?.UpdateEyeGaze(this, new Ray(CameraCache.Main.transform.position, CameraCache.Main.transform.forward), DateTime.UtcNow);
+                InputSystem?.EyeGazeProvider?.UpdateEyeTrackingStatus(this, true);
```
wiwei: Nit: May as well move this to the line above so that our simulation and actual provider can be consistent (i.e. eye tracking status is always updated first, then eye gaze direction/position).

End effect honestly shouldn't be anything but yeah

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
```diff
@@ -517,12 +518,21 @@ public void UpdateEyeGaze(IMixedRealityEyeGazeDataProvider provider, Ray eyeRay,
...
         /// <summary>
         /// Ensure that we work with recent Eye Tracking data. Return false if we haven't received any
         /// new Eye Tracking data for more than 'maxETTimeoutInSeconds' seconds.
         /// </summary>
         private bool IsEyeTrackingAvailable => (DateTime.UtcNow - latestEyeTrackingUpdate).TotalSeconds <= maxEyeTrackingTimeoutInSeconds;
 
+        /// <summary>
+        /// Boolean to check whether the user went through the eye tracking calibration.
+        /// </summary>
+        public bool? IsEyeCalibrationValid => isEyeCalibrated;
```
wiwei: Do we expect that someone would treat the "null" state any different from the false state? i.e. effectively I'm trying to understand that this API is now a tristate - i.e. null (which currently isn't defined), false (means not calibrated) and true (calibrated). The question someone might have is, what does null mean? Can things go back to null? Should a consumer really just care about true false?

wiwei:
It would also be great to have a small test here (i..e validating that the events will actually fire when eye calibration goes out (using the test property) or that the prefab shows up when the test property goes to false.

wiwei:
Regarding the test thing - if you don't have cycles for this let's chat and see how we can get that coverage and when we could.

wiwei:
Ah, I guess if it's in the demo folder then maybe my test concern isn't AS pressing but would still be good to have.

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
sostel: The problem here was that at the startup of the app, the default would be to return false, but that would trigger the event of "not being calibrated" to fire. Null allows for us to know whether the variables has been initialized appropriately and has received an actual value from the eye tracking system.

sostel: I guess I could have also set the default to "isEyeCalibrated = true", but seems a little less clean to me.

sostel:
> It would also be great to have a small test here (i..e validating that the events will actually fire when eye calibration goes out (using the test property) or that the prefab shows up when the test property goes to false.

Not sure I understand what you mean. How do you want me to document this here? I tested it both in the Editor and several times on the device with clearing my eye tracking calibration data. This is how I noticed that the timestamps didn't match up (last commit).

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Scripts/Utils/EyeCalibrationChecker.cs
sostel: Done.

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
sostel: Done.

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
wiwei: No worries - I think it would be good just to have it commented saying that "hey, a null value means such and such" otherwise it's not clear as a reader why it's not really a bool, but actually a tri-state (and thus, by the typical tristate boolean can't go back to UNDEFINED/null, but again it's just a matter of creating clarity for readers without making them have to dig through the implementation to understand the behavior)

wiwei:
> > It would also be great to have a small test here (i..e validating that the events will actually fire when eye calibration goes out (using the test property) or that the prefab shows up when the test property goes to false.
> 
> Hm, well, I validated that the events are correctly fired on the device by actually changing my eye calibration status. I also tested it in the Editor where it's more convenient of course, because you can use the little checkbox from the EyeCalibrationChecker script to test transitions.
> How else do you want it documented here?

My ask is more around having a play mode test, but since this code isn't in the core package of the MRTK the concern isn't as strong. I see sufficient in value getting this in as-is, I'm mainly trying to also understand how we can prevent regressions here in the future.

wiwei:
Approving conditional on that doccomment update - I'll push this change in after that.

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
sostel: Sounds reasonable. I'll add a comment.

sostel:
> > > It would also be great to have a small test here (i..e validating that the events will actually fire when eye calibration goes out (using the test property) or that the prefab shows up when the test property goes to false.
> > 
> > 
> > Hm, well, I validated that the events are correctly fired on the device by actually changing my eye calibration status. I also tested it in the Editor where it's more convenient of course, because you can use the little checkbox from the EyeCalibrationChecker script to test transitions.
> > How else do you want it documented here?
> 
> My ask is more around having a play mode test, but since this code isn't in the core package of the MRTK the concern isn't as strong. I see sufficient in value getting this in as-is, I'm mainly trying to also understand how we can prevent regressions here in the future.

Ok, makes sense. Thanks, Will! Unfortunately I don't have the bandwidth to do it right now. So, yep, let's either just go ahead with it for now or please assign a dev to create a test for it. 
Thanks!

Issue #4696, opened 2019-06-04
There is no way to access pointer mediators from code


julenka:
## Describe the bug
There is no way to access the pointer mediators in the MRTK input system from code, which makes it difficult to customize which pointers are enabled / disabled at run time. I think we should consider this for GA because it falls under principle of "allow developers to customize behavior".

## Additional Context
Developers can implement their own custom pointer mediators, but if they want to talk to that mediator to say do different behaviors based on application state (let's say you sometimes want to turn all pointers off), there is no way to talk to their custom mediators.

For example, there is no way to turn off the far hand rays at run time without modifying MRTK input system. The following [stackoverflow question](https://stackoverflow.com/questions/56248329/how-to-turn-off-motion-controller-pointer-mediator) is an example of an external user wanting this behavior. 

While it may seem like turning off line pointers could be done by directly accessing `myPointer.IsActive = false`, this actually does not work because the pointer mediators overwrite these values every frame based on the state of other pointers. Therefore, to turn off line pointers a component must talk to the mediator and ask the mediator to turn the line pointers off.

Jarodshow:
> Note: the reason you need to use this approach of modifying the mediator instead of directly setting `myPointer.IsActive = false` is because the default mediator overwrites these values every frame. 

Could we put this in the docs somewhere? :)

PR #4698, opened 2019-06-04
Pointer Mediators accessible from code and extensible

julenka:
## Overview
Fixes: #4696 by making pointer mediators accessible from code. Also allows `DefaultPointerMediator` to be extended by making methods virtual and fields protected.

This enables users to crate solutions like turn off the far hand beams at runtime as described in [this stackoverflow post](https://stackoverflow.com/questions/56248329/how-to-turn-off-motion-controller-pointer-mediator)

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
@@ -32,11 +32,11 @@ namespace Microsoft.MixedReality.Toolkit.Input
         private PointerHitResult hitResult3d = new PointerHitResult();
         private PointerHitResult hitResultUi = new PointerHitResult();
 
-        public ref readonly Dictionary<uint, IMixedRealityPointerMediator> PointerMediators
+        public IReadOnlyDictionary<uint, IMixedRealityPointerMediator> PointerMediators
         {
             get
             {
-                return ref pointerMediators;
+                return pointerMediators;
             }
         }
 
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -32,6 +32,14 @@ public class FocusProvider : BaseDataProvider, IMixedRealityFocusProvider
         private PointerHitResult hitResult3d = new PointerHitResult();
         private PointerHitResult hitResultUi = new PointerHitResult();
 
+        public ref readonly Dictionary<uint, IMixedRealityPointerMediator> PointerMediators
```
luis-valverde-ms: >ref readonly Dictionary [](start = 15, length = 23)

I think this should be IReadOnlyDictionary instead. Dictionary is a reference type (i.e. class) so a ref to a dictionary object is a ref to its address. I believe that what you are returning there in C++ terms is Dictionary* const, instead of Dictionary const *.

luis-valverde-ms: Actually, to be precise, I think `ref readonly Dictionary` translates to `Dictionary** const` in C++ and I don't think C# provides any way to have a `Dictionary const **` or `Dictionary const *` return type.

luis-valverde-ms: I can see no option to approve with comments, so I'm approving it so it can be checked in once you do that change.

luis-valverde-ms: Nth correction, I think `ref readonly Dictionary` translates to `Dictionary * const *` in C++.

PR #4700, opened 2019-06-04
Add missing namespaces to resolve the lack of NuGet packages building.

wiwei:
The addition of this test caused an issue building NuGet packages, because this test was added without a namespace. The asset retargetting step (which is a precursor to building nuget packages) assumes that every single class belongs to the non-global namespace (actually, the MRTK main namespace).

Other followups associated with this issue:

1) Probably some linter that checks that all .cs files actually have a namespace declared.
2) Probably treating warnings from the nuget step/asset retargetting step as error (so that it fails the CI build)


wiwei:
This is being built in CI here:

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=2107

Issue #4699, opened 2019-06-04
There is always a BaseNearInteractionTouchable within "touching" distance


MenelvagorMilsom:
## Describe the bug

Prior to #4263 NearInteractionTouchables were only considered within touching distance, if the fingertip was less than distFront away. Now distFront is per touchable, and this check no longer occurs.

## Additional context
[comment](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4263/files/3d3dabe067a9fb95d684e5cc2ae5f07bd691dd35#r289467678)

PR #4701, opened 2019-06-04
Add check to assure that closestProximityTouchable is at least distFront away

MenelvagorMilsom:
## Overview
distFront is now per BaseNearInteractionTouchable. As such, we need to check distFront for each BaseNearInteractionTouchable.

## Changes
- Fixes: #4699.


julenka:
This fixed a bug I saw where I would never get the far interaction pointer to show up.

Issue #3771, opened 2019-04-02
WindowsMixedRealityArticulatedHand and WindowsMixedRealityGGVHand should not inherit from WindowsMixedRealityController


chrisfromwork:
## Overview
It appears that the HoloLens 2 work introduced WindowsMixedRealityArticulatedHand and WindowsMixedRealityGGVHand. It does not feel appropriate for these classes to inherit from WindowsMixedRealityController.

There was an issue when merging logic into the public github branch where hand mesh rendering broke based on changes to the WindowsMixedRealityController. There is nothing preventing this from continuing to happen. If WindowsMixedRealityController exists specifically for VR controllers, it should be able to contain logic that breaks hand experiences. If WindowsMixedRealityController is confusing naming and just a base class for uwp input in ar/vr space, we should create a new class that is specifically for vr controllers.

keveleigh:
Fixed via #4707

PR #4707, opened 2019-06-04
Fix runtime controller model loading and fix WMRController inheritance

### LINKED ISSUE #4342: Missing Controllers on build
## Describe the bug

When I build my application/game or whatever I cant see the controllers. I can when im on editor mode, but not when built. But if I have something happening when i press a button it works, so the problem its that i cant see or track the controllers. (Last time I tested it they spawned on the ground and wont move).

## To reproduce

1. Make a project
2. Build it
3. Convert into an executable with Visual Studio
4. Open it

## Expected behavior

You should be able to see your hands, but it wont work. They wont be or will be on the ground

## Screenshots

I cant sorry

- Unity Version [latest 2019]
- MRTK Version [v2.0]

- WMR

## Additional context

If I start the application, on the right eye in one corner there is a console with red text and the options close and clear. I cant read what it says because its on the edge. I cant interact with it neither.

On capabilities I have Bluetooth on, i dont know if im missing something else

Im building on release and debug, I tested this both

keveleigh:
## Overview
This PR adds a base class for WMR sources, allowing us to remove the confusing and constraining inheritance on WMR controller. This also allows us to do controller-specific things, like loading the controller model, without affecting work done by other sources, like loading the hand mesh.

The controller model loading regression happened in #3761, which was done to re-enable hand meshes on HL2. As a result, I felt that fixing both of these issues at the same time made the most sense.

## Changes
- Fixes: #3771, fixes #4342

## Verification
1. Verify that input continues to work on all WMR platforms (HL1, HL2, immersive).
2. Verify that motion controller models load at runtime when deployed to a PC
3. Verify that hand meshes continue to display on HL2


julenka:
Did you already complete this in mrtk_development? Very strangely, today I was seeing my WMR controllers as visible right away without needing to press windows button. I figured somebody put a fix in already.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
```diff
@@ -61,443 +61,29 @@ public WindowsMixedRealityController(TrackingState trackingState, Handedness con
...
-                    }
-            }
-        }
-
-        /// <summary>
-        /// Update the Thumbstick input from the device
-        /// </summary>
-        /// <param name="interactionSourceState">The InteractionSourceState retrieved from the platform</param>
-        /// <param name="interactionMapping"></param>
-        private void UpdateThumbStickData(InteractionSourceState interactionSourceState, MixedRealityInteractionMapping interactionMapping)
```
julenka: It was a little surprising to see the thumbstick and touchpad methods moved into the base class, since it feels like those are specific to the motion controller.

julenka: Same with trigger and menu buttons.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs
```diff
@@ -0,0 +1,495 @@
...
+        public virtual void UpdateController(InteractionSourceState interactionSourceState)
+        {
+            if (!Enabled) { return; }
+
+            UpdateControllerData(interactionSourceState);
+
+            if (Interactions == null)
+            {
+                Debug.LogError($"No interaction configuration for Windows Mixed Reality {ControllerHandedness} Source");
+                Enabled = false;
```
wiwei: Could be existing behavior, but curious how this works out - it looks like if Enabled was false, we would have exited out early on and never updated the interactionSourceState. However if Interactions is null here, we would set Enabled to false, but still update interactionSourceState at the end. Is the rest of this stuff depending on this or does it not really matter?

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
keveleigh: Yeah, I waffled a bit on what should go where. I ended up sticking everything from the `InteractionSourceState` in the base class. My rational for some of them was "like we started sending data for hands over the `grasped` state, what if we start sending something for, say, menu in the future".

I agree though, in the current world, I'll move motion controller specific buttons to the controller class explicitly.

keveleigh: Updated

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs
```diff
@@ -0,0 +1,495 @@
...
+
+#if UNITY_WSA
+using UnityEngine;
+using UnityEngine.XR.WSA.Input;
+#endif
+
+namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input
+{
+    /// <summary>
+    /// A Windows Mixed Reality Controller Instance.
```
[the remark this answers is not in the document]
keveleigh: Updated

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityArticulatedHand.cs
```diff
@@ -26,7 +26,7 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input
     [MixedRealityController(
         SupportedControllerType.ArticulatedHand,
         new[] { Handedness.Left, Handedness.Right })]
-    public class WindowsMixedRealityArticulatedHand : WindowsMixedRealityController, IMixedRealityHand
+    public class WindowsMixedRealityArticulatedHand : BaseWindowsMixedRealitySource, IMixedRealityHand
```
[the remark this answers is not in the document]
keveleigh: Confirmed that one and two handed manipulation still works on HL1. It seemed a little inconsistent (the second hand often dropped out seemingly randomly), but I see that without this change as well.

Issue #4705, opened 2019-06-04
Microsoft.MixedReality.Toolkit.UI.States must be instantiated using the ScriptableObject.CreateInstance method instead of new States


keveleigh:
## Describe the bug

When running the HandInteractionExamples scene, a warning is printed: Microsoft.MixedReality.Toolkit.UI.States must be instantiated using the `ScriptableObject.CreateInstance` method instead of `new States`.

## To reproduce

1. Run the HandInteractionExamples scene (since it has interactables set up)
2. Observe the warning in the console

## Expected behavior

No warnings

## Screenshots

![image](https://user-images.githubusercontent.com/3580640/58914315-0749a880-86d3-11e9-987e-659d0eada303.png)

## Your Setup (please complete the following information)

- Unity Version 2018.4.1f1
- MRTK Version - RC2 Stabilization

PR #4709, opened 2019-06-05
Update States.cs to use CreateInstance instead of new

keveleigh:
## Overview
States.cs was using `new` to create an instance of a `ScriptableObject` instead of `CreateInstance`.
I'm not fully sure how I got into this state, since the `GetDefaultInteractableStates` method appears to only be called if you don't have a state specified in the Inspector, but the Inspector seems to try to make sure the default asset is always loaded in. Deleting the default asset is a good way to repro this issue.

Note: this is currently targeting mrtk_development. Feel free to retarget to stabilization if we want to take this there (my branch is based on stabilization, so it won't be bringing over anything extra from dev), but I don't see this as blocking RC2.

## Changes
- Fixes: #4705 

## Verification
1. Ensure that this call still creates `States` and uses it correctly

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/States.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/States/States.cs
@@ -25,7 +25,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
         public static States GetDefaultInteractableStates()
         {
-            States result = new States();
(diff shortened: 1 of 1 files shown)


julenka:
@thalbern @wiwei consider bringing this into rc2 stabilization. I don't expect this to break anything.

wiwei:
Reading over the issue and this code (and the fact that we get warnings in the console without it), retargetting to stabilization. I don't believe this should cause any destabilizing effects on the rc2 branch.

Issue #4511, opened 2019-05-22
[Cursor] Cursor does not represent proper input state


cre8ivepark:
The cursor does not show proper input state visual on both HoloLens 1 and 2.
It seems like some of the updates in the input do not trigger a proper state change. (DefaultCursor prefab has not changed from HoloToolkit)

## Describe the bug
**HoloLens 1 (GGV)**
- Default: Dot cursor is displayed (OK)
- On hand detected (ready gesture): the cursor is updated to ring cursor (OK)
- On pinch: the cursor is updated to bigger dot (OK)
- On pinch release: the cursor is updated to ring (OK)
- **On source lost (move the hand out of GFOV): the cursor stays as ring cursor (BUG: Cursor should return to Dot)**

**HoloLens 2 (At the end of the Hand Ray)**
- Default hand ray: Dot cursor is displayed (OK)
- On pinch: the cursor is updated to dot (OK)
- **On pinch release: the cursor is updated to ring but a few seconds later, it changes to dot (BUG: Cursor should stay as a ring)**

## Expected behavior
HoloLens 1
![2019-05-21 17_34_44-YoonPark_UnityUniteBerlin2018_MixedRealityToolkit_20180621 pptx - PowerPoint](https://user-images.githubusercontent.com/13754172/58139315-0868d900-7bef-11e9-8c46-0557e3caa995.png)

HoloLens 2
Hand ray's default cursor should be a ring. On pinch = big dot.

## Blocking
Without the fix, it would be difficult to properly test out implementations for these items
#4418 and #4417 

## To reproduce
1. Clone mrtk_release branch (or download and import RC1 Refresh)
2. Build and deploy HandInteractionExample scene

## Target Platform (please complete the following information)
- HoloLens

paco-ms:
Hi @cre8ivepark  - this bug is a regression I assume?

MenelvagorMilsom:
@cre8ivepark I don't see any animation for the big dot for when pinching. Did this work before?

cre8ivepark:
Merge

PR #4715, opened 2019-06-05
Fix cursor so it represents the proper input state

MenelvagorMilsom:
## Overview
As far as I can see, the state animations were correct, but BaseCursor.CheckCursorState was returning incorrect state for Interact and InteractHover. This was because GGVPointer and ShellHandRayPointer were not correctly updating BaseCursor's sources count.

As for the "big dot" for Select, I don't see any evidence of this existing before in MRTK. Is this something that should be added in the scope of this PR?

## Changes
- Fixes: #4511.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
@@ -12,8 +12,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// </summary>
     public class ControllerPoseSynchronizer : InputSystemGlobalListener, IMixedRealityControllerPoseSynchronizer
     {
-        protected bool hasSource = false;
-
         #region IMixedRealityControllerPoseSynchronizer Implementation
 
         [SerializeField]
@@ -93,16 +91,11 @@ namespace Microsoft.MixedReality.Toolkit.Input
         #region IMixedRealitySourcePoseHandler Implementation
 
         /// <inheritdoc />
-        public virtual void OnSourceDetected(SourceStateEventData eventData)
-        {
-            hasSource = true;
-        }
+        public virtual void OnSourceDetected(SourceStateEventData eventData) { }
 
         /// <inheritdoc />
         public virtual void OnSourceLost(SourceStateEventData eventData)
         {
-            hasSource = false;
-
             if (eventData.SourceId == Controller?.InputSource.SourceId &&
                 eventData.Controller?.ControllerHandedness == Handedness)
             {
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
@@ -12,6 +12,8 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// </summary>
     public class ControllerPoseSynchronizer : InputSystemGlobalListener, IMixedRealityControllerPoseSynchronizer
     {
+        protected bool hasSource = false;
+
         #region IMixedRealityControllerPoseSynchronizer Implementation
 
         [SerializeField]
@@ -91,11 +93,16 @@ namespace Microsoft.MixedReality.Toolkit.Input
         #region IMixedRealitySourcePoseHandler Implementation
 
         /// <inheritdoc />
-        public virtual void OnSourceDetected(SourceStateEventData eventData) { }
+        public virtual void OnSourceDetected(SourceStateEventData eventData)
+        {
+            hasSource = true;
+        }
 
         /// <inheritdoc />
         public virtual void OnSourceLost(SourceStateEventData eventData)
         {
+            hasSource = false;
+
             if (eventData.SourceId == Controller?.InputSource.SourceId &&
                 eventData.Controller?.ControllerHandedness == Handedness)
             {
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs
@@ -90,7 +90,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 cursorInstance.name = $"{Handedness}_{name}_Cursor";
 
                 BaseCursor oldC = BaseCursor as BaseCursor;
-                if (oldC != null && hasSource)
+                if (oldC != null && enabled)
                 {
                     oldC.VisibleSourcesCount--;
                 }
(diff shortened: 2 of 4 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
```diff
@@ -258,6 +258,17 @@ protected override void Start()
...
         }
 
+        protected override void OnDisable()
+        {
+            base.OnDisable();
+            this.gazeProvider = InputSystem.GazeProvider as GazeProvider;
+            BaseCursor c = gazeProvider.GazePointer.BaseCursor as BaseCursor;
+            if (c != null)
+            {
+                c.VisibleSourcesCount--;
```
wiwei: We might need some other bool state, or a comment saying that this is safe with this line:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/8f29ba829718dc366d9ad2371c21231fa68bf3ba/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs#L322

Otherwise it seems like it would be possible to drive the count down by two.

MenelvagorMilsom: Yeah, I only just noticed that other line. Currently looking into it.

MenelvagorMilsom: Yeah, line 319 should be:
```BaseCursor c = gazeProvider.GazePointer.BaseCursor as BaseCursor;```
That way we don't need the other decrement.

**Comment by MenelvagorMilsom** [comment]
I'm going to look a bit deeper and see if there's a better way of doing this for ShellHandRayPointer.

**Comment by MenelvagorMilsom** [comment]
I moved the sources count update to BaseControllerPointer for ShellHandRayPointer. I'm not sure if this makes sense though. Let me know what you all think.

File: Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs [hunk 2]
```diff
@@ -145,6 +145,12 @@ protected override async void Start()
             }
 
             SetCursor();
+
+            BaseCursor c = BaseCursor as BaseCursor;
```
wiwei: Another thing to consider - it's possible for BaseCursor to change as a result of calling SetCursor - if VisibleSourcesCount has been changed on BaseCursor, then we'd still be off in some way (i..e if we changed it, VisibleSourcesCount would be up by one on the old BaseCursor, and wouldn't have been correctly incremented on the new one)

MenelvagorMilsom: In what situations are SetCursor called?

wiwei: I have noooooo idea, just mentioning this after looking at the code

wiwei: I have a special place in my heart for refcounts.

MenelvagorMilsom:
@julenka, perhaps it makes more sense to increment the counter in OnSourceDetected, rather than in Start? It seems strange to increment before we have a confirmed source.

Also, we should probably decrement OnDisable and increment OnEnable. I guess users would expect disabled sources to no longer have an effect?

These should be simple changes to make, just wanted your opinion.

julenka:
> @julenka, perhaps it makes more sense to increment the counter in OnSourceDetected, rather than in Start? It seems strange to increment before we have a confirmed source.
> 
> Also, we should probably decrement OnDisable and increment OnEnable. I guess users would expect disabled sources to no longer have an effect?
> 
> These should be simple changes to make, just wanted your opinion.

If we both increment in OnSourceDetected and in OnEnable, won't we be double counting? I was also thinking it was weird that we used Start and OnSourceLost, which don't really match. I was doing it this way to match other code so that at least we were wrong in a consistent way :-) 

I think it may be better to do this in OnEnable/Disable because there might be cases where you detect a source but want to ignore it. For example, let's say a hand is flickering in and out and you only want to enable the controller when hand has been visible for 10 frames.

When you make the change can you update all code so it uses the same pattern? We would then need to test it to make sure we didn't regress stuff of course.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -12,6 +12,8 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// </summary>
     public class ControllerPoseSynchronizer : InputSystemGlobalListener, IMixedRealityControllerPoseSynchronizer
     {
+        protected bool hasSource = false;
```
keveleigh: Can we use `IsTracked` for this?

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -91,11 +93,16 @@ public MixedRealityInputAction PoseAction
...
-        public virtual void OnSourceDetected(SourceStateEventData eventData) { }
+        public virtual void OnSourceDetected(SourceStateEventData eventData)
+        {
+            hasSource = true;
+        }
 
         /// <inheritdoc />
         public virtual void OnSourceLost(SourceStateEventData eventData)
         {
+            hasSource = false;
```
keveleigh: This event will fire for all lost sources, not just the one this pose synchronizer is listening for. This should probably be inside the below `if` statement (pending my first comment, though).

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
keveleigh: Or maybe `TrackingState`?

MenelvagorMilsom: Following from Julia's suggestion, we should probably use `enabled`

MenelvagorMilsom:
> When you make the change can you update all code so it uses the same pattern? We would then need to test it to make sure we didn't regress stuff of course.

@julenka other than the changes I made today, is there anywhere else you think this is required? As far as I can see, GGVPointer and BaseControllerPointer (and its subclasses) accounts for all of the pointers with associated cursors. I'm not certain of this though. Let me know if anything else needs a change.

julenka:
> > When you make the change can you update all code so it uses the same pattern? We would then need to test it to make sure we didn't regress stuff of course.
> 
> @julenka other than the changes I made today, is there anywhere else you think this is required? As far as I can see, GGVPointer and BaseControllerPointer (and its subclasses) accounts for all of the pointers with associated cursors. I'm not certain of this though. Let me know if anything else needs a change.

Looks good to me, now it's just a question of if this works when you stress test it. I think a good way to test this is probably to test it on an HL2 and HL1, since with those you'll get lots of input sources aquired and lost.

MenelvagorMilsom:
I tested this on HL1 and HL2 this afternoon. The cursor appears to be correct now on both devices. I tested interactions and losing source with both hands in a number of permutations. We should be able to make play mode tests for this.

PR #4716, opened 2019-06-05
Make global input events registration per-handler instead of per-object

googlan:
## Overview
Global input events are now registered on a per-component (handler) basis instead of per-object.
This PR changes the underlying mechanism of how event system handles event registration. Change preserves behavior of old API and doesn't change registration in most places where it's used. This will be subject of subsequent changes.
Removed async part of registration, as it was hiding at what moment event registration / unregistration would happen.

Problems of old implementation, fixed with the new one:
- If a component registers for global events, it would receive global input events of all types.
- If one of components on an object registers for global input events, all components on this object will receive global events.
- If two components on the same object listen to global events, and one is disabled, second stops receiving global events.


Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -13,56 +13,209 @@ namespace Microsoft.MixedReality.Toolkit
...
+            foreach (IEventSystemHandler handler in handlers)
             {
-                ExecuteEvents.Execute(EventListeners[i], eventData, eventHandler);
+                Debug.Assert(handler is T);
+                eventHandler.Invoke((T)handler, eventData);
             }
 
             eventExecutionDepth--;
+
+            if(eventExecutionDepth == 0 && (handlersToRemove.Count > 0 || handlersToAdd.Count > 0))
```
luis-valverde-ms: I like this better than making it asynchronous but in its current form it does not preserve execution order. E.g. if a handler gets added and then removed while handling this event, we won't produce the right result. I think we can fix that having a single list for add/remove.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -13,56 +13,209 @@ namespace Microsoft.MixedReality.Toolkit
...
+
+                foreach (var handler in handlersToAdd)
+                {
+                    AddHandlerToMap(handler.Item1, handler.Item2);
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public virtual void RegisterHandler<T>(IEventSystemHandler handler) where T : IEventSystemHandler
```
luis-valverde-ms: I think it would be simpler to register handlers only under the type passed in here instead of once for each interface under `IEventSystemHandler` and, when handling an event, traverse all registered types and invoke their handlers if they can be assigned to the specific type that we're handling. 

That sounds more expensive than the current solution but I'd like to get an idea for how much before choosing the more complicated way. Could we maybe profile that in a release build?

googlan: I did a brief profile of it, it seems to be ~2 times slower. Discussed it in person, it would also cause complications in other places. I kept the initial solution with registering multiple entries (if necessary) per handler.

googlan:
Tests are mostly done, need to clean them up and add messages to asserts.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -11,58 +11,220 @@ namespace Microsoft.MixedReality.Toolkit
...
     {
-        #region IMixedRealityEventSystem Implementation
-
         private static int eventExecutionDepth = 0;
-        private readonly WaitUntil doneExecutingEvents = new WaitUntil(() => eventExecutionDepth == 0);
+        private readonly Type eventSystemHandlerType = typeof(IEventSystemHandler);
+
+        private enum Action
+        {
+            Add,
```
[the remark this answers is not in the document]
googlan: I'm using them now for GameObjects as well. I think it's fine as it's private and very local.

---
In reply to: [292971039](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4716#discussion_r292971039) [](ancestors = 292971039)

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -11,58 +11,220 @@ namespace Microsoft.MixedReality.Toolkit
...
-        #endregion IMixedRealityEventSystem Implementation
+        #endregion Registration helpers
+
+        #region Utilities
+
+        private void TraverseEventSystemHandlerHierarchy<T>(IEventSystemHandler handler, Action<Type, IEventSystemHandler> func) where T : IEventSystemHandler
+        {
+            var handlerType = typeof(T);
+
+            func(handlerType, handler);
```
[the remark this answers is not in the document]
googlan: No. Added a comment to clarify that.

---
In reply to: [292965276](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4716#discussion_r292965276) [](ancestors = 292965276)

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            IMixedRealityInputSystem inputSystem = null;
+            MixedRealityServiceRegistry.TryGetService(out inputSystem);
+
+            Assert.IsNotNull(inputSystem, "Input system must be initialized");
+
+            // Let input system to register all cursors and managers.
+            yield return null;
+
+            // Switch off / Destroy all input components, which listen to global events
+            Object.Destroy(inputSystem.GazeProvider.GazeCursor as Behaviour);
```
wiwei: For future reference, would it be possible to have a more deterministic disabling of this (i.e. a test hook for killing off global event listeners). The reason I ask is, this is currently exhaustive of the set of global event listeners, but if we were to add another it would be easy to miss adding it here (and it's likely that unexpected bugs could be difficult to diagnose - i.e. we eventually find out that oh, we missed another input listener here, but it took a while to figure out)

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,300 @@
...
+        [TearDown]
+        public void ShutdownMrtk()
+        {
+            TestUtilities.ShutdownMixedRealityToolkit();
+        }
+
+        /// <summary>
+        /// </summary>
+        [UnityTest]
+        public IEnumerator Test01_TestObjectBasedEventRegistration()
```
wiwei: Nit: Is the Test01 prefix/Test02 prefix needed to ensure some sort of deterministic order? I left similar comments on other folks' reviews - these prefixes don't add a lot of value, and as we add/delete them they will immediately get out order/have numerical gaps.

Imagine we want to add another test that is related to object based event registration, but Test02 is already taken, in which case our naming also gets weird.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            var objectBasedListener = object1.AddComponent<TestInputGlobalListenerObjectBased>();
+            var handlerBasedListener = object1.AddComponent<TestInputGlobalListenerHandlerBasedSpeechHandler>();
+
+            yield return null;
+
+            // Event listener collection is filled for backward compatibility
+            Assert.AreEqual(inputSystem.EventListeners.Count, 1, "Event listener for old event system API hasn't been registered");
+            Assert.Contains(objectBasedListener.gameObject, inputSystem.EventListeners, "Wrong event listener registered for old event system API.");
+
+            Assert.AreEqual(inputSystem.EventHandlersByType.Count, 2, "Input event system doesn't contain expected event handler types.");
```
wiwei: If there's any way for us to use an autoformatter to reign in the line length here it might be helpful (reviewing using github actually is difficult because it's longer that CR viewport width)

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            CollectionAssert.Contains(inputSystem.EventHandlersByType.Keys, typeof(IMixedRealitySpeechHandler), "Input event system doesn't contain IMixedRealitySpeechHandler entry.");
```
wiwei: When it comes to doing collection based comparisons, consider using something like:

https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.collectionassert.areequivalent?view=mstest-net-1.2.0#Microsoft_VisualStudio_TestTools_UnitTesting_CollectionAssert_AreEquivalent_System_Collections_ICollection_System_Collections_ICollection_

This prevents you from having to do a size assert first, followed by an individual assert for each item. 

The only reason you shouldn't do this is if that AreEquivalent actually gives you a really bad diff/error message.

The reason I also ask is, the way it's written, if something is missing (i.e. count is 1) you'll only get an error that the count is 1, and you'll never execute the other stuff which tells you specifically what is missing.

tl;dr when doing collection based comparisons, used collection based compare functions, and if you want to do individual ones prefer doing the size check at the end to get the most helpful error messages.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,300 @@
...
+
+            Assert.AreEqual(inputSystem.EventHandlersByType[typeof(IMixedRealityBaseInputHandler)].Count, 1, "Input event system doesn't contain expected IMixedRealityBaseInputHandler handlers.");
+            CollectionAssert.Contains(inputSystem.EventHandlersByType[typeof(IMixedRealityBaseInputHandler)], handlerBasedListener, "Input event system doesn't contain IMixedRealityBaseInputHandler for new API handler.");
+
+            objectBasedListener.enabled = false;
+
+            // Make sure that disabling global listener doesn't remove the new API one.
+            Assert.AreEqual(inputSystem.EventHandlersByType.Count, 2, "Input event system contains unexpected event handlers.");
+
+            Assert.AreEqual(inputSystem.EventHandlersByType.Count, 2, "Input event system doesn't contain expected event handler types.");
```
wiwei: Lines 97 and 99 are the same thing (I know that there's some copy paste here, but it might be worth just deleting the one on 97 and depending on 99)

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,300 @@
...
+
+            IMixedRealityInputSystem iInputSystem = null;
+            MixedRealityServiceRegistry.TryGetService(out iInputSystem);
+
+            BaseEventSystem inputSystem = (BaseEventSystem)iInputSystem;
+
+            var object1 = new GameObject("Object");
+
+            // These 2 handlers are independent
+            var handlerBasedListener1 = object1.AddComponent<TestInputGlobalListenerHandlerBasedAllHandlers>();
```
wiwei: Nit: It might help with readability to have a comment about what specific handlers each of these components register, so that within the context of this test, you can be like:

#1  = pointer speech base
#2  = speech base

If you read the test from top to bottom you have your mindset now where you know that the diff is just pointer, so when disabling/adding stuff down below it's a bit easier to understand the expected deltas.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            CollectionAssert.Contains(inputSystem.EventHandlersByType.Keys, typeof(IMixedRealityBaseInputHandler), "Input event system doesn't contain IMixedRealityBaseInputHandler entry.");
+
+            Assert.AreEqual(inputSystem.EventHandlersByType[typeof(IMixedRealitySpeechHandler)].Count, 1, "Input event system doesn't contain expected IMixedRealitySpeechHandler handlers.");
+            CollectionAssert.Contains(inputSystem.EventHandlersByType[typeof(IMixedRealitySpeechHandler)], handlerBasedListener2, "Input event system doesn't contain IMixedRealitySpeechHandler for speech-handler component.");
+
+            Assert.AreEqual(inputSystem.EventHandlersByType[typeof(IMixedRealityBaseInputHandler)].Count, 1, "Input event system doesn't contain expected IMixedRealityBaseInputHandler handlers.");
+            CollectionAssert.Contains(inputSystem.EventHandlersByType[typeof(IMixedRealityBaseInputHandler)], handlerBasedListener2, "Input event system doesn't contain IMixedRealityBaseInputHandler for speech-handler component.");
+
+            ///
```
wiwei:
```suggestion
```

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            Assert.Zero(handlerBasedListener1.speechCount,                "Speech event is received by all-handlers component.");
+
+            // No pointer clicked event:
+            Assert.Zero(handlerBasedListener2.pointerClickedCount,        "Pointer clicked event is received by speech-handler component.");
+            Assert.Zero(handlerBasedListener2.pointerDownCount,           "Pointer down event is received by speech-handler component.");
+            Assert.Zero(handlerBasedListener2.pointerUpCount,             "Pointer up event is received by speech-handler component.");
+            Assert.Zero(handlerBasedListener2.pointerDraggedCount,        "Pointer dragged event is received by speech-handler component.");
+            Assert.Zero(handlerBasedListener2.speechCount,                "Speech event is received by speech-handler component.");
+
+            /////
```
wiwei:
```suggestion
```

wiwei: I think there are a few more of these (will probably just leave the rest as a find/replace exercise)

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -13,10 +13,24 @@ namespace Microsoft.MixedReality.Toolkit
...
+        private readonly Type eventSystemHandlerType = typeof(IEventSystemHandler);
+
+        private enum Action
+        {
+            Add,
+            Remove
+        }
+
+        // Lists for handlers which are added/removed during event dispatching.
+        // Lists are processed independently, so
```
wiwei: so!?!?!!!!?!?!?

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,226 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
             for (int i = EventListeners.Count - 1; i >= 0; i--)
             {
                 ExecuteEvents.Execute(EventListeners[i], eventData, eventHandler);
             }
 
+            // Send events to all handlers registered via RegisterHandler API.
+            HashSet<IEventSystemHandler> handlers;
+            if (EventHandlersByType.TryGetValue(typeof(T), out handlers))
+            {
+                foreach (IEventSystemHandler handler in handlers)
```
wiwei: The only thing I'd be careful of here is GC created by using foreach over this enumeration. I know that historically foreach has been slower than for (probably fixed in some .NET or something else) and that this would also actually trigger alloc (because it would allocate an enumerator in order to do the foreach).

It would be worth doing a quick profile of this using the Unity profiler to verify that this doesn't do per-frame GC alloc (i.e. a one time thing is fine, something on every frame or something that happens continuously on every N frames is bad). This is especially interesting because these events CAN be noisy (i.e. joint updates happen each frame) so we basically need to consider this to be happening at a rate Nx of the update loop.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,226 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
+                {
+                    // If handler is a unity component, need to make sure, that it didn't receive an event
+                    // when we were sending it to Game objects above.
+                    // Do that by checking if its parent is registered in EventListeners.
+                    var componentHandler = handler as Component;
```
wiwei: I don't have any data here to back this up, but it feels weird having to do a null check + hashset lookup within each component call.

I would almost rather have us be explicit in being like you can either register for one or the other, so that you have to explicitly move to the new one or stay on the old one.

The only value I can see in having both is if you might use it during migration - but even then it's questionable - if you've registered for both, your "new method" registration is basically useless, and you can't validate it works until you unregister from the old, and if we support both at the same time and then kill off the old in a future release, the customer would have never been forced to evaluate the new method entirely (i.e. will suddenly stop getting messages for another one)

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
wiwei: If this IS happening, we might want to snap a list<> upon each register/unregister call, so that we can for() instead of foreach()

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,226 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
+                        continue;
+                    }
+
+                    eventHandler.Invoke((T)handler, eventData);
+                }
+            }
+
             eventExecutionDepth--;
+
+            if(eventExecutionDepth == 0 && postponedActions.Count > 0)
```
wiwei:
```suggestion
            if (eventExecutionDepth == 0 && postponedActions.Count > 0)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,226 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
+            {
+                foreach(var handler in postponedActions)
```
wiwei:
```suggestion
                foreach (var handler in postponedActions)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,226 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
+                    {
+                        AddHandlerToMap(handler.Item2, handler.Item3);
+                    }
+                    else if(handler.Item1 == Action.Remove)
+                    {
+                        RemoveHandlerFromMap(handler.Item2, handler.Item3);
+                    }
+                }
+
+                foreach(var obj in postponedObjectActions)
```
wiwei:
```suggestion
                foreach (var obj in postponedObjectActions)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,226 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
+                {
+                    if(obj.Item1 == Action.Add)
```
wiwei:
```suggestion
                    if (obj.Item1 == Action.Add)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,226 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
+
+            if(eventExecutionDepth == 0 && postponedActions.Count > 0)
+            {
+                foreach(var handler in postponedActions)
+                {
+                    if (handler.Item1 == Action.Add)
+                    {
+                        AddHandlerToMap(handler.Item2, handler.Item3);
+                    }
+                    else if(handler.Item1 == Action.Remove)
```
wiwei:
```suggestion
                    else if (handler.Item1 == Action.Remove)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
googlan: The problem comes from having 2 different components on the same object. If one of them uses old API, and second uses new API, second will be triggered during `ExecuteEvents` call. We need this check to avoid second invoke of the callback.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
googlan: I don't think I can iterate over HashSet without creating an enumerator. I can do this manually by `GetEnumerator` + `MoveNext`, but it might have the same GC issue.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
googlan: Honestly I've added it after looking at other tests (edit mode, because there were very few play mode). I thought it was intentional. I'll remove them here.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
googlan: I was considering to move this part into a utility function.
It shouldn't be confusing, because it asserts in the end, that not all global input handlers are removed. When someone adds a new global handler, they should immediately hit that.

wiwei:
Alrighty, sorry for the delay (and having reviewed it after @luis-valverde-ms already did). Had some mediumish comments but could go either way on them.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
googlan: Somehow I wasn't shown the second part of the comment. Yeah, list should work here.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
wiwei: right right - I think what I'm saying is, maybe we could just change the Register functions to throw if you tried to register via the old way AND the new way - basically, let's just not support the scenario and error out as early as possible (and do the checking in a place that happens not in every tick), and that we can also reduce some of the complexity here (i.e. if we prevent old + new registration, there's no need to do these checks here)

googlan: This might cause weird behavior, when users just add a new component to an object in existing scene and things stop working because there's an API conflict.
I'll see if I can cache that.

googlan: Actually I'll leave it as is for now, because overhead of `ExecuteEvents.Execute` is at least comparable if not much more.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
googlan: CollectionAssert comparisons are awesome!

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
googlan: I think it should be much better now with Collection comparisons.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -13,10 +13,24 @@ namespace Microsoft.MixedReality.Toolkit
...
+        private readonly Type eventSystemHandlerType = typeof(IEventSystemHandler);
+
+        private enum Action
+        {
+            Add,
+            Remove
+        }
+
+        // Lists for handlers which are added/removed during event dispatching.
+        // Game objects and handlers are processed independently, so be kept in separate lists.
```
luis-valverde-ms: "so they are kept"

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,227 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
         public virtual void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
         {
             Debug.Assert(!eventData.used);
+
             eventExecutionDepth++;
 
+            // First handle events via old API part, via GameObjects.
+            // This sends event to every component of matching type on this object, 
+            // INCLUDING the ones that don't expect it.
```
luis-valverde-ms: To be more clear I'd say: 

> This sends the event to every component that implements the corresponding event handling interface,
> regardless of whether it was the one registering the object as global listener or not.
> We keep this behavior for backwards compatibility. Will be removed together with the `IMixedRealityEventSystem.Register(GameObject listener)` interface.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,227 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
+                        continue;
+                    }
+
+                    eventHandler.Invoke((T)handler, eventData);
+                }
+            }
+
             eventExecutionDepth--;
+
+            if (eventExecutionDepth == 0 && postponedActions.Count > 0)
```
luis-valverde-ms: Need to check `postponedObjectActions.Count` too.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
luis-valverde-ms: I think we should consider an error to register a game object and a handler in one of its components, so I wouldn't add code to make it silently work. Instead I'd check as much as we can on registration, as Will was suggesting.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -25,44 +39,227 @@ public abstract class BaseEventSystem : BaseService, IMixedRealityEventSystem
...
+                        Unregister(obj.Item2);
+                    }
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public virtual void RegisterHandler<T>(IEventSystemHandler handler) where T : IEventSystemHandler
+        {
+            #if WINDOWS_UWP && !ENABLE_IL2CPP
```
luis-valverde-ms: Document why we have two code paths here

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            yield return null;
+
+            // Event listener collection is filled for backward compatibility
+            CollectionAssert.AreEquivalent(
+                new List<GameObject> { objectBasedListener.gameObject },
+                inputSystem.EventListeners,
+                "Event listener for old event system API hasn't been registered correctly.");
+
+            CollectionAssert.AreEquivalent(
```
wiwei: Looks good!

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
googlan: Let's do this as a part of #4847. Right now the majority of Mrtk components still uses old API (I can't enable Obsolete warning because of that). Once those are cleaned up, we can mark methods as obsolete and add stricter validation on in `Register` / `Unregister` implementation, which will only remain in client code.

wiwei: sounds good to me!

PR #4721, opened 2019-06-05
Update the Spatial Awareness docs to provide a guide on how to access meshes

wiwei:
We had some repeated questions about how to do this, so I figured it would be nice just to have this in docs.

Un-obsoletes the GetObserver* functions, because the spatial awareness system doesn't actually have the data provider interface that it references (i.e. it tell you to check to see if it has it, and then you do, and you realize it doesn't have it)


Assets/MixedRealityToolkit/Interfaces/SpatialAwareness/IMixedRealitySpatialAwarenessSystem.cs
```diff
@@ -45,7 +45,6 @@ public interface IMixedRealitySpatialAwarenessSystem : IMixedRealityEventSystem
         /// <returns>
         /// Read only copy of the list of registered observers.
         /// </returns>
-        [Obsolete("GetObservers will be removed in a future release. Check to see if the instance implements IMixedRealityDataProviderAccess and call GetDataProviders.")]
```
davidkline-ms: The obsolete attribute should not be removed. you need to cast the spatial awareness instance to IMixedRealityDataProviderAccess to call GetProviders().

davidkline-ms: (SpatialAwarenessSystem as IMixedRealityDataProviderAccess)?.GetProviders()

davidkline-ms: i am happy to work with you on the documentation, etc :)

keveleigh:
It'd also be good to add a section on suspending and resuming the observers at runtime, which is I question I've seen asked a few times as well.

Assets/MixedRealityToolkit/Interfaces/SpatialAwareness/IMixedRealitySpatialAwarenessSystem.cs
davidkline-ms: #4737 fixes the error in spatial awareness where it was not specifying that it implements IMixedRealityDataProviderAccess

wiwei: At the time when I wrote this, IMixedRealityDataProviderAccess didn't exist on this object. We probably shouldn't have obsoleted these functions unless the alternative actually existed.

wiwei: What I'll do is merge the stabilization branch down to dev and then get this change updated

wiwei: Okay, merged in. I have to say that now that I've had some experience playing with this, the new API design isn't intuitive and doesn't really make sense to me - it feels like this is designed to be extendable and generic but errs so far on that side that usabillity suffers. We can probably chat about this on an issue afterward and isn't worth going into on this PR.

wiwei: To be clear, the new API surface with respect to how it applies in the SA system. I don't have an opinion yet on the others.

davidkline-ms: The reason for the new API design was to reduce breaking changes and add flexibility for implementations to add providers in the future.

We can go into more detail if desired :)

Documentation/SpatialAwareness/UsageGuide.md
```diff
@@ -0,0 +1,55 @@
+# Spatial Awareness - Usage Guide
+
+This document contains some guides for common tasks that programmatically access the spatial awareness system.
+
+## Enumerating and accessing the meshes
+
+Accessing the meshes that are currently known to the spatial awareness system involves first querying
+for the spatial awareness system, then getting a hold of the IMixedRealitySpatialAwarenessMeshObserver
+(note that the base spatial awareness observer has no notion of meshes), and then enumerating those
+meshes known that observer.
```
davidkline-ms:
```suggestion
meshes known to that observer.
```

wiwei: Thanks, for future reference I don't think we need to put ourselves as waiting for a single grammar fix (of which the solution was provided here).

PR #4722, opened 2019-06-05
fix for truncating pointer ray length before doing ui raycast

naostranMS:
## Overview
Calling UpdatePointerRayOnHit immediately after the Physics Raycast was truncating the ray length before the UI raycast. This can result in selecting a physics hit from a lower-priority layer over a UI hit that is further away but in a higher-priority layer.

## Changes
- don't truncate pointer ray until after both physics and ui raycasts
- removed UpdatePointerRayOnHit as it's only being done in a single place now

## Verification
One side-effect of this change is that it now does the RayStep truncation for SceneQueryType.SphereOverlap where it previously didn't - I don't really have any experience with that pointer type, so I'm not sure how that should be handled. It would be easy enough to check for that case and not do the truncation if that difference in behavior is intentional/significant.

>
>


luis-valverde-ms:
Hey @naostranMS , sorry for the delay. I'll have a look and get back to you between today and tomorrow.

Railboy:
I'm not seeing any side effects on the visuals. @julenka can you verify that pointer visuals look correct?

naostranMS:
> It still makes sense to shorten the ray if there are less than two priorities. That is the default case so I think it is worth adding a code path for that.

Yeah, that makes sense and is easy enough to add.

> Can we have a unit or playmode test for the bug we are fixing?

I suppose that seems reasonable, but I'm not sure what is required for that - do you have examples of how those tests are implemented in MRTK?

luis-valverde-ms:
> > It still makes sense to shorten the ray if there are less than two priorities. That is the default case so I think it is worth adding a code path for that.
> 
> Yeah, that makes sense and is easy enough to add.
> 
> > Can we have a unit or playmode test for the bug we are fixing?
> 
> I suppose that seems reasonable, but I'm not sure what is required for that - do you have examples of how those tests are implemented in MRTK?

Have a look at the tests under `Assets\MixedRealityToolkit.Tests` for examples and [here](https://docs.unity3d.com/Manual/PlaymodeTestFramework.html) for a reference on how to write and execute Unity Tests.

naostranMS:
@luis-valverde-ms

I added the truncation before UI raycast in the case of the physics raycasting returning a result at the highest priority. That said, one thing I noticed is the truncation doesn't do anything to stop subsequent ray steps from being checked in the graphics raycast - which seems pretty weird. I don't know why it's doing the truncate - if it's to optimize raycast performance, then maybe it would be better to just pass the max length into the graphics raycast (and not change the ray step) and check against that. If the intent is to actually update the pointer for other subsequent users - that's already being handled in other ways (e.g. LineBase.GetNormalizedLengthFromWorldLength in LinePointer.OnPostSceneQuery).

I also added some tests and verified that they caught the bug that I'm fixing.

Let me know what you think!

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -869,6 +869,15 @@ private void UpdatePointer(PointerData pointer)
                     QueryScene(pointer.Pointer, raycastProvider, prioritizedLayerMasks, hitResult3d);
                     PointerHitResult hit = hitResult3d;
 
+                    int hitResult3dLayer = hitResult3d.hitObject?.layer ?? -1;
+                    if (hitResult3dLayer == 0)
+                    {
+                        // if we have a hit in the highest priority layer, we can go ahead and truncate the pointer before doing the UI raycast
```
luis-valverde-ms: >if [](start = 27, length = 2)

Nit: start comments with capital for consistency.

Assets/MixedRealityToolkit.Tests/PlayModeTests/Components/FocusRaycastTestProxy.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UnityEngine;
+
+public class FocusRaycastTestProxy : MonoBehaviour
```
luis-valverde-ms: Documentation here to make life easier for people maintaining this test, please

Assets/MixedRealityToolkit.Tests/PlayModeTests/FocusProviderRaycastTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [UnityTest]
+        public IEnumerator TestRaycastProxies()
+        {
+            Assert.IsNotNull(pointer, "FocusProvider is null!");
+
+            foreach (var raycastTestProxy in raycastTestPrefabInstance.GetComponentsInChildren<FocusRaycastTestProxy>())
+            {
+                pointer.SetFromTestProxy(raycastTestProxy);
+                yield return pointer.WaitForFocusUpdate();
```
luis-valverde-ms: Why do we need a custom wait instruction instead of just 'yield return null' to wait for one frame?

Assets/MixedRealityToolkit.Tests/PlayModeTests/Components/TestPointer.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Input;
+using Microsoft.MixedReality.Toolkit.Physics;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Tests
+{
+    public class TestPointer : GenericPointer
```
luis-valverde-ms: Document at least the intent of the class

Assets/MixedRealityToolkit.Tests/PlayModeTests/FocusProviderRaycastTests.cs
```diff
@@ -0,0 +1,74 @@
...
+namespace Microsoft.MixedReality.Toolkit.Tests
+{
+    public class FocusProviderRaycastTests
+    {
+        private GameObject raycastTestPrefabInstance = null;
+        private TestPointer pointer = null;
+        IMixedRealityFocusProvider focusProvider = null;
+
+        [UnityTest]
+        public IEnumerator TestRaycastProxies()
```
luis-valverde-ms: Document the general test setup and workflow, e.g. test proxies are used to setup test cases and expected results, test below runs through them.

luis-valverde-ms:
@naostranMS , thanks for those changes. It is looking pretty good except for those few comments I left above, mostly to do with documenting things a bit to make the like of the person maintaining code base easier.

naostranMS:
@luis-valverde-ms 

err, well I tried to check the build again - I'm not sure why that unit test failed as it's not failing locally, so I was hoping to see if that actually repro'd.

I also added a fair bit of documentation - let me know if there are any more problems or if that will suffice.

luis-valverde-ms:
@naostranMS , everything looks good. I don't know why the test is failing, I can't repro locally either. @wiwei , who could have a look at that?

Issue #4702, opened 2019-06-04
Slate scroll with hand not working with GGV on Hololens 1 and Motion Controllers in VR


thalbern:
## Describe the bug

In the HandInteractionExample scene the Slate scroll with Hand is not working with GGV on Hololens 1 or the Motion Controllers in VR.
You can however scroll the slate by using the hands on Hololens 2 or using either of the Input Simulation modes in the editor (articulated or gesture)

## To reproduce

- Open the HandInteractionExample Scene in Unity
- Deploy on Hololens 1 and try to scroll the slate
- Start the Scene for Windows Mixed Reality in Unity Editor Playmode and try to scroll the slate

## Expected behavior

Slate should be scrollable with any type of input

## Screenshots

Offending Slate:
![image](https://user-images.githubusercontent.com/36998103/58896011-12113700-86ed-11e9-94c8-a25e314ffeff.png)

## Your Setup (please complete the following information)

- Unity Version 2018.4
- MRTK Version RC2

## Target Platform (please complete the following information)

- HoloLens
- WMR immersive
- OpenVR

PR #4735, opened 2019-06-05
Fix: Slate scroll with hand not working with GGV on Hololens 1 and Motion Controllers in VR #4702

johnppella:
## Changes
- Fixes: #4702

## Verification
hand held MR Controllers and GGV now scrolls slate control

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
@@ -646,15 +646,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
             if (data.currentPointer != null)
             {
                 Vector3 pt = data.currentPointer.Position;
-                //if (data.currentController.IsRotationAvailable)
-                //{
-                //   // Vector3 dir = controller.InputSource.Pointers[0].Rays[0].Direction * (controller.InputSource.Pointers[0].SphereCastRadius);
-                //    data.initialProjectedOffset = SnapFingerToQuad(pt);
-                //}
-                //else//pointer is GGVPOinter and has no SphereCastRadius
-                //{
-                    data.initialProjectedOffset = SnapFingerToQuad(pt);
-                //}
+                data.initialProjectedOffset = SnapFingerToQuad(pt);
             }
 
             data.touchingQuadCoord = GetUVFromPoint(data.touchingPoint);
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
@@ -839,8 +839,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
         #endregion IMixedRealitySourceStateHandler Methods
 
         #region Unused Methods
-        public void OnPositionInputChanged(InputEventData<Vector2> eventData) { }
-        public void OnInputPressed(InputEventData<float> eventData) { }
         public void OnSourceDetected(SourceStateEventData eventData) { }   
         public void OnPointerDragged(MixedRealityPointerEventData eventData) { }
         public void OnPointerClicked(MixedRealityPointerEventData eventData) { }
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
@@ -206,6 +206,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         #region Private Methods
         private bool TryGetMRControllerRayPoint(HandPanData data, out Vector3 rayPoint)
         {
+           
             if (data.currentPointer != null && data.currentController != null && data.currentController.IsPositionAvailable)
             {
                 rayPoint = data.touchingInitialPt + (SnapFingerToQuad(data.currentPointer.Position) - data.initialProjectedOffset);
@@ -226,17 +227,17 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
                 if (data.IsActive == true)
                 {
-                    if (TryGetMRControllerRayPoint(data, out tryHandPoint))
-                    {
-                        tryGetSucceeded = true;
-                    }
-                    else if (data.IsSourceNear == true)
+                    if (data.IsSourceNear == true)
                     {
                         tryGetSucceeded = TryGetHandPositionFromController(data.currentController, TrackedHandJoint.IndexTip, out tryHandPoint);
                     }
                     else
                     {
                         tryGetSucceeded = TryGetHandPositionFromController(data.currentController, TrackedHandJoint.Palm, out tryHandPoint);
+                    }
+                    if (!tryGetSucceeded)
+                    {
+                        tryGetSucceeded = TryGetMRControllerRayPoint(data, out tryHandPoint);
                     }
 
                     if (tryGetSucceeded == true)
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
@@ -31,6 +31,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
             public Vector2 touchingInitialUV = Vector2.zero;
             public Vector2 touchingUVOffset = Vector2.zero;
             public Vector2 touchingUVTotalOffset = Vector2.zero;
+            public Vector3 initialProjectedOffset = Vector3.zero;
             public IMixedRealityInputSource touchingSource = null;
             public IMixedRealityController currentController = null;
         }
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -200,30 +211,60 @@ private void Update()
...
+                    
                 }
             }
         }
         #endregion MonoBehaviour Handlers
 
 
         #region Private Methods
+        private bool TryGetMRControllerRayPoint(HandPanData data, out Vector3 rayPoint)
+        {
```
julenka: Any reason you are doing a string comparison here? We've been trying to avoid determining controller types by doing string comparison, since the name could easily change.

If you really want to check if it is a mixed reality controller, why not just do `if (currentController is WindowsMixedRealityController)`?

But, more generally, is this restricting you to only work with MR controllers? Would slate then break if you used an oculus, or a vive?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -200,30 +211,60 @@ private void Update()
...
+            if (data.currentController.InputSource.SourceName.Contains("Mixed Reality Controller"))
+            {
+                if (!(data.currentController.InputSource.Pointers[0] is GGVPointer))
```
julenka: What if the GGVPointer exists but is not the first pointer in the list?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -790,9 +861,13 @@ public void OnTouchCompleted(HandTrackingInputEventData eventData)
         /// </summary>
         public void OnInputDown(InputEventData eventData)
         {
-            EndTouch(eventData.SourceId);
-            SetHandDataFromController(eventData.InputSource.Pointers[0].Controller, false);
-            eventData.Use();
+            if (eventData.MixedRealityInputAction.Description != "None")
```
julenka: Could you explain why you are doing a string compare against "None" here? I am wondering if there's a different way to write this to avoid the string compare...

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -660,6 +721,16 @@ private bool TryGetHandPositionFromController(IMixedRealityController controller
             position = Vector3.zero;
             return false;
         }
+        private void GetControllerPoints(List<Vector3> points)
```
julenka: Is this method used anywhere?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -137,6 +138,21 @@ private bool scaleActive
...
+            {
+                if (inputSystem == null)
+                {
+                    MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem);
+                }
+                return inputSystem;
+            }
+        }
+
+        private IMixedRealityEyeGazeProvider EyeTrackingProvider => eyeTrackingProvider ?? (eyeTrackingProvider = InputSystem?.EyeGazeProvider);
```
julenka: Why do we need to get eye tracking data here?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: I can find a better test. This check is in there because a trigger pull on  a MC Controller fires several InputDown events- the description is 'None' on all but the last one which is the correct event to capture.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: isnt the first pointer in the list the one that is associated with the event?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: is there a more general generic controller type than WindowsMixedRealityController?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -623,6 +667,21 @@ private void SetHandDataFromController(IMixedRealityController controller, bool
                 }
             }
 
+            //store value in case of MRController
+            if (controller.InputSource.Pointers.Length > 0 )
```
julenka: Again it's dangerous to assume that GGVPointer is the first pointer in the list. Instead, you can check if the controller is a WindowsMixedRealityGGVHand

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: ill make this switch :)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: removing

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: old- ill remove it

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: will do

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: where are these types defined?

luis-valverde-ms:
From my ignorance of HandInteractionPanZoom,  this is how I would go about determining the pan delta: 
[PanTest.zip](https://github.com/microsoft/MixedRealityToolkit-Unity/files/3262528/PanTest.zip)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: got rid of string compares and Pointers[0] convention

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: oops missed one

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -788,29 +816,34 @@ public void OnTouchCompleted(HandTrackingInputEventData eventData)
...
-        public void OnSourceDetected(SourceStateEventData eventData) { }
         public void OnSourceLost(SourceStateEventData eventData)
         {
             EndTouch(eventData.SourceId);
             eventData.Use();
         }
         #endregion IMixedRealitySourceStateHandler Methods
+
+        #region Unused Methods
+        public void OnPositionInputChanged(InputEventData<Vector2> eventData) { }
```
julenka: I think you no longer need OnPositionInputChanged, looks like no interface implements this any more.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -788,29 +816,34 @@ public void OnTouchCompleted(HandTrackingInputEventData eventData)
...
+        public void OnInputPressed(InputEventData<float> eventData) { }
```
julenka: Same with OnInputPressed

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: oh! youre right!

johnppella: removing OnPositionInputChanged & OnInputPressed

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
johnppella: yup

PR #4752, opened 2019-06-06
Clean up the the MRTK architecture and packaging documents

wiwei:
I'm in the process of putting up some new MRTK overview docs (i.e. overall architecture, philosophy, that kinda stuff) and when I started that process, I was looking to see what already existed. A lot of this stuff was either not true, aspirational, or no longer accurate. I wanted to ensure that our documents actually reflect the current truth, not the past, not the future, only the current truth.

I bent this rule a little bit with the MRTK_Modularization file (which still contains references to future), though for those, they should be easy enough just to remove the "Upcoming Changes" section once those are done (and they will be done very soon)

Other things to mention:

1) We currently only have two packages (foundation, example). Our docs shouldn't refer to things that don't currently exist, as that only confuses people. And also to be clear, me writing this doesn't imply that this is how it will be forever, just that this is currently how things are.
2) The vNext doc actually did have some interesting stuff that probably would be worth carrying forward in the contribution guide (i.e. features must have examples, tests, be programmatically changable and not just at design time). I can carry this forward in a future change.


keveleigh:
image

Documentation/Packaging/MRTK_Modularization.md
```diff
@@ -48,62 +21,29 @@ through the Mixed Reality Toolkit core component.
...
-For projects where the required features and/or platforms are well known, unnecessary system services, providers 
-and features can be left unchecked to minimize the MRTK footprint.
-
-> NOTE: MRTK v2 **_requires_** the contents of the Assets\MixedRealityToolkit folder. 
-
-#### Delete Feature Folders
-
-After importing the Microsoft.MixedReality.Toolkit.Unity.Foundation-v2.0.0-RC1.unity package, the MRTK footprint can 
-be reduced by deleting unneeded folders from:
+It is possible to, during the import of the Foundation package, uncheck arbitrary items, though obviously this
```
thalbern: i'm not particularly fond of this paragraph. It reads a bit weird to me. Also we should avoid using "you" and "your" in the docs
My suggestion:
"It is possible to uncheck arbitrary items during the import of the Foundation package. However it's not recommended to do this at an early stage of development as it might break functionality. 
After having figured out the final feature set of the app, pruning unneeded providers and services can be done on the following folders:
"

thalbern: mhh the last sentence probably should be ... unneeded providers and services can be done in the following folders . sry not my native languag :D

wiwei: That's much better, changed to that!

PR #4756, opened 2019-06-06
Update projectversion.txt to 2018.4.1

davidkline-ms:
This change updates the project version for MRTK v2 to Unity 2018.4.1f1 to reduce the number of times customers see the "you may need to update scripts" message upon loading the project when cloning from GitHub.

It is a smaller version of #4468 that does not yet update the import time check that is part of MRTK (it still checks for 2018.3 in this change). Once the build automation machines are upated  to 2018.4, we will add the import time check back.


davidkline-ms:
PR opened to verify this change successfully passes CI. If so, will update the doc files.

Documentation/GettingStartedWithTheMRTK.md
```diff
@@ -9,7 +9,7 @@ The Mixed Reality Toolkit (MRTK) is a cross-platform toolkit for building Mixed
 To get started with the Mixed Reality Toolkit you will need:
 
 * [Visual Studio 2017](http://dev.windows.com/downloads)
```
keveleigh: VS 2019? I think that page defaults to 2019 now, and I've heard it can be difficult / not possible? to find 2017.

davidkline-ms: good point!

davidkline-ms: there does not appear to be a unique page for 2017... you have to click on "older versions"

davidkline-ms: updated download link

PR #4764, opened 2019-06-06
Consistency: make FocusProvider a core service and not a data provider

davidkline-ms:
Focus provider is considered a core system by the MixedRealityToolkit object. As such, it should implement IMixedRealitySystem, derive from BaseCoreSystem and be registered in the MixedRealityServiceRegistry.

This change correctly classifies and registers IMixedRealityFocusProvider instances.

Tested via the solvers scene on OpenVR and Windows Mixed Reality Immersive devices. Teleportation and object selection were the primary focus of interaction testing.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
@@ -16,12 +16,11 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// </summary>
     /// <remarks>There are convenience properties for getting only Gaze Pointer if needed.</remarks>
     [DocLink("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Input/Overview.html")]
-    public class FocusProvider : BaseDataProvider, IMixedRealityFocusProvider
+    public class FocusProvider : BaseCoreSystem, IMixedRealityFocusProvider
     {
         public FocusProvider(
             IMixedRealityServiceRegistrar registrar,
-            IMixedRealityInputSystem inputSystem,
-            MixedRealityInputSystemProfile profile) : base(registrar, inputSystem, null, DefaultPriority, profile)
+            MixedRealityInputSystemProfile profile) : base(registrar, profile)
         { }
 
         private readonly HashSet<PointerData> pointers = new HashSet<PointerData>();
@@ -75,9 +74,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
             {
                 MixedRealityInputSystemProfile profile = ConfigurationProfile as MixedRealityInputSystemProfile;
 
-                if ((Service != null) &&
-                    (profile != null) &&
-                    profile.PointerProfile != null)
+                if (profile != null && profile.PointerProfile != null)
                 {
                     return profile.PointerProfile.PointingExtent;
                 }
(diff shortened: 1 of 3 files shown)


julenka:
> Focus provider is considered a core system by the MixedRealityToolkit object. 

For my own education, why is Focus provider considered a core system? Can you provide evidence indicating it is a core system?

davidkline-ms:
You bet @julenka! From MixedRealityToolkit.cs

```
        private static bool IsCoreSystem(Type type)
        {
            if (type == null)
            {
                Debug.LogWarning("Null cannot be a core system.");
                return false;
            }

            return typeof(IMixedRealityInputSystem).IsAssignableFrom(type) ||
                   typeof(IMixedRealityFocusProvider).IsAssignableFrom(type) ||
                   typeof(IMixedRealityTeleportSystem).IsAssignableFrom(type) ||
                   typeof(IMixedRealityBoundarySystem).IsAssignableFrom(type) ||
                   typeof(IMixedRealitySpatialAwarenessSystem).IsAssignableFrom(type) ||
                   typeof(IMixedRealityDiagnosticsSystem).IsAssignableFrom(type);
        }
```

davidkline-ms:
There are plans to allow it to stand alone (separate from the input system) in a future version. Along with making it's type consistent, this is a first step to ensure there wont be breaking changes when changes are made.

wiwei:
> You bet @julenka! From MixedRealityToolkit.cs
> 
> ```
>         private static bool IsCoreSystem(Type type)
>         {
>             if (type == null)
>             {
>                 Debug.LogWarning("Null cannot be a core system.");
>                 return false;
>             }
> 
>             return typeof(IMixedRealityInputSystem).IsAssignableFrom(type) ||
>                    typeof(IMixedRealityFocusProvider).IsAssignableFrom(type) ||
>                    typeof(IMixedRealityTeleportSystem).IsAssignableFrom(type) ||
>                    typeof(IMixedRealityBoundarySystem).IsAssignableFrom(type) ||
>                    typeof(IMixedRealitySpatialAwarenessSystem).IsAssignableFrom(type) ||
>                    typeof(IMixedRealityDiagnosticsSystem).IsAssignableFrom(type);
>         }
> ```

I think this answer shows in code that there's a function that classifies a few different things as core services, but I don't think it necessarily answers the greater question of why this is a core service vs not a core service - i.e. it feels like an arbitrary distinction without a definition. I've also been confused on this terminology in the past and have read a few different parts of our docs that haven't necessarily helped my understanding - is there a definition/doc you can point us to that would help with that confusion?

davidkline-ms:
> I think this answer shows in code that there's a function that classifies a few different things as core services, but I don't think it necessarily answers the greater question of why this is a core service vs not a core service 

Excellent point. Historically, "core service" translates to "there shall only be one instance". (This is a tangential part of the service registration work that will ultimately enable standalone services).

The issue that this PR is attempting to remedy is that FocusProvider is treated throughout the code as a system, yet is implemented, instantiated and registered as a data provider (of which, the presumption is that there can be many).

The work that is forthcoming (in a future release) is to separate registration and configuration of the FocusProvider from the InputSystem as it has value to applications beyond just being used with input (ex: a cinematic experience that needs to know if the user's attention is on a particular scene element).

This change also standardizes how applications request the FocusProvider instance (previously, due to how it was defined and how data provider access is controlled by the input system was inconsistent from other components).

Hope this helps clear up questions.

Issue #4750, opened 2019-06-06
[RC2 Build Call] Slider, scroll bar and text selection don't react to hand movements on HL1


thalbern:
## Describe the bug

Slider, scroll bar and text selection in HandInteractionExample don't react to hand movements on Hololens 1

MRTK RC2 Build Call
Unity 2018.4

julenka:
This is for Unity UI, by the way.

julenka:
I'll take a look

PR #4765, opened 2019-06-07
[RC2] Add ability to scroll, drag UnityUI canvas on HoloLens 1

julenka:
## Overview
Add support for drag interactions in Unity UI for HoloLens 1 GGV interactions. This is done by using the pointer position instead of the pointer ray origin in cases when the pointer ray is decoupled from the actual pointer, like with hands (pointer ray is the head, but pointer position is hands).

## Changes
- Fixes: #4750 

## Verification
All tests done in editor, not on device.
- Tested articulated hands
- Tested gesture hands

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
@@ -191,8 +191,22 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
             // Reset the RaycastCamera for projecting (used in calculating deltas)
             Debug.Assert(pointer.Rays != null && pointer.Rays.Length > 0);
-            RaycastCamera.transform.position = pointer.Rays[0].Origin;
-            RaycastCamera.transform.rotation = Quaternion.LookRotation(pointer.Rays[0].Direction);
+
+            if (pointer.Controller != null && pointer.Controller.IsRotationAvailable)
+            {
+                RaycastCamera.transform.position = pointer.Rays[0].Origin;
+                RaycastCamera.transform.rotation = Quaternion.LookRotation(pointer.Rays[0].Direction);
+            }
+            else
+            {
+                // The pointer.Controller does not provide rotation, for example on HoloLens 1 hands.
+                // In this case pointer.Rays[0].Origin will be the head position, but we want the 
+                // hand to do drag operations, not the head.
+                // pointer.Position gives the position of the hand, use that to compute drag deltas.
+                RaycastCamera.transform.position = pointer.Position;
+                RaycastCamera.transform.rotation = Quaternion.LookRotation(pointer.Rays[0].Direction);
+            }
+
 
             // Populate eventDataLeft
             pointerData.eventDataLeft.Reset();


Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -191,8 +191,22 @@ protected MouseState GetMousePointerEventDataForMrtkPointer(PointerData pointerD
...
+            {
+                RaycastCamera.transform.position = pointer.Rays[0].Origin;
+                RaycastCamera.transform.rotation = Quaternion.LookRotation(pointer.Rays[0].Direction);
+            }
+            else
+            {
+                // The pointer.Controller does not provide rotation, for example on HoloLens 1 hands.
+                // In this case pointer.Rays[0].Origin will be the head position, but we want the 
+                // hand to do drag operations, not the head.
+                // pointer.Position gives the position of the hand, use that to compute drag deltas.
```
keveleigh: Is this assuming the GGVPointer is being used?

julenka: I don't think so. But it does assume that if a pointer does not provide rotation, then we should be using the pointer position instead of pointer.Rays[0].Origin. I couldn't think of any other way to support GGV hands without explicitly saying "if controller is a ggvhand" or "if input source is a ggvhand". I didn't want to do that check because then we would be introducing a dependency between .Services and .Providers, which I don't think exists right now. 

If you or anybody else has a suggestion for how to check "if we are on HoloLens 1 or with a GGV hand, use the position of the hand" I'm very interested! I myself am not happy with this, but couldn't think of a better way.

julenka:
> @julenka This code snippet would be very helpful for the developers who are looking for ways to distinguish HL1 hand vs HL2 hand, especially for the companies that maintain both HL1 and HL2 projects. It would be great if we can put this info somewhere in our input documentation.

Hi Yoon, I would recommend for companies USING mrtk (not for the core MRTK components) do a check like the following instead:

```
if (controller is WindowsMixedRealityArticulatedHand) {
// this is a HoloLens 2 articulated hand;
} else if (controller is WindowsMixedRealityGGVHand) {
// this is a HoloLens 1 hand;
}
```

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -191,8 +191,22 @@ protected MouseState GetMousePointerEventDataForMrtkPointer(PointerData pointerD
 
             // Reset the RaycastCamera for projecting (used in calculating deltas)
             Debug.Assert(pointer.Rays != null && pointer.Rays.Length > 0);
-            RaycastCamera.transform.position = pointer.Rays[0].Origin;
-            RaycastCamera.transform.rotation = Quaternion.LookRotation(pointer.Rays[0].Direction);
+
+            if (pointer.Controller != null && pointer.Controller.IsRotationAvailable)
```
luis-valverde-ms: Ideally, we should be able to do something like this all the time (no special casing):
```
RaycastCamera.transform.position = pointer.Position; 
RaycastCamera.transform.rotation = pointer.Rotation;
```

luis-valverde-ms: The pointer abstraction is most useful if we make the position and rotation in each implementation mean something similar so it can can be used in situations like this interchangeably.

julenka: I actually tried that first, but then touch interactions broke. It's really helpful to try it yourself and see how it breaks, since it's a bit complicated. The Poke Pointer will set its ray cast to be in front of the touchable surface even when the finger tip is behind the surface. However, the position and rotation always follows the position/rotation of the finger tip.

How would we solve this problem? 

A thing I thought up was to change position + rotation of the pointer to match the ray that poke pointer is using, but I was wary to do that because I didn't want to break existing behavior.

wiwei:
Following up here, from chatting with folks it sounds like there's some pending feedback, and while there is some brittle-ness to the approach here, changing it to use the suggestion above would require additional work that would make this change riskier. 

I'm pushing this in and then doing another build call - we can file another issue to track the feedback above, or look to address it if comes back to us again in the future.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
luis-valverde-ms: Very interesting point. It seems like it could make sense, in the future, differentiating between where the pointer _is_ and where the pointer _wants to be_. In the case of the poke pointer, the pointer would be considered to be on the surface of the UI element it is interacting with and would want to be where the finger tip is. We need to know where the pointer wants to be to drive things like push buttons.

Issue #4569, opened 2019-05-25
Create Hand Coach Prefabs to teach gestures / interactions


julenka:
## Overview
In OOBE and Learn Gestures app we have some nice 3D hand animations to teach people things like the system gesture, and suggest things like "press this button".

These hand coach prefabs could be helpful in teaching first time users how to interact with the mixed reality UI. The hand coach could also be useful for demonstrating hand movements when training workers, for example.

Here is an example of a hand coach module from the [MRDL Lunar Module project](https://github.com/Microsoft/MRDesignLabs_Unity_LunarModule):

![image](https://user-images.githubusercontent.com/168492/58362653-4c0d4e00-7e4e-11e9-9966-4bbe51630e7c.png)

As a first step we could provide pre-recorded 3D animations of hand movements for common motions like "press the button" or "grab the object". However, we could also create a way to record and serialize hand gestures, and then use these gestures in the hand coach.

julenka:
I added the Epic label because I think it would be awesome to have this :-)

julenka:
I would really like to have this for the next month if we can squeeze it in. Realistically though we have a pretty big number of bugs regarding bounding box and manipulation  handler that we need to fix so core dev team might not get to it. This is poachable / please feel free to help.

polar-kev:
Hand coach has graduated from experimental to a full feature in MRTK 2.6
https://docs.microsoft.com/windows/mixed-reality/mrtk-unity/features/ux-building-blocks/hand-coach

PR #4775, opened 2019-06-07
Input recording and playback feature

lukastoenneMS:
## Overview

Adds serialization, services, and tooling for recording hand and camera motion and animating simulated input.

The immediate goal is to allow automated testing of interactions and input systems. Another possible use case is animating objects for educational purposes (see #4569)

## Changes

* InputRecordingService to record input in the background. The recording buffer is time-limited by default and can be exported at any time, creating a "rolling buffer" that can be used for capturing input during interactions.
* InputPlaybackService can load input animation and re-apply it at runtime using the InputSimulationService.
* Serialization utilities for exporting recorded animation into a custom binary format.

**Note:** These are additional editor features that have been separated out to make code review easier:
* https://github.com/lukastoenneMS/MixedRealityToolkit-Unity/pull/2
* https://github.com/lukastoenneMS/MixedRealityToolkit-Unity/pull/1


lukastoenneMS:
TODO:
* Update the demo test to work with the new services.
* Write documentation and how-to for workflow.
* Make sure the InputAnimationTarget component works for clips in the animation editor.

chrisfromwork:
Do we need this commit history? Some of our tooling for viewing code reviews only allows for 64 commits. If we could get these commits squashed it may improve our review process.

lukastoenneMS:
> 
> 
> Do we need this commit history? Some of our tooling for viewing code reviews only allows for 64 commits. If we could get these commits squashed it may improve our review process.

No, we don't need all the commits. There has been a lot of back and forth over time, with me trying out different implementations and then rolling back. There isn't much value in keeping all that.

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/InputRecordingControls.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        public void Update()
+        {
+            bool isRecording = RecordingService.IsRecording;
+            if (isRecording != wasRecording)
+            {
+                if (isRecording)
+                {
+                    OnRecordingStarted.Invoke();
```
Troy-Ferrell: Instead of polling every update frame, why not add event/delegate to the RecordingService that notifies when recording state has changed? Then if need be, then funnel the C# event to the UnityEvent etc

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+            handPinchCurveLeft = new AnimationCurve();
+            handPinchCurveRight = new AnimationCurve();
+            handJointCurvesLeft = new Dictionary<TrackedHandJoint, PoseCurves>();
+            handJointCurvesRight = new Dictionary<TrackedHandJoint, PoseCurves>();
+            cameraCurves = new PoseCurves();
+
+            markers = new List<InputAnimationMarker>();
+        }
+
+        public bool TryGetHandJointCurves(Handedness handedness, TrackedHandJoint joint, out PoseCurves curves)
```
Troy-Ferrell: nit: docs like other public methods

Assets/MixedRealityToolkit.Services/InputAnimation/InputRecordingService.cs
```diff
@@ -0,0 +1,317 @@
...
+            BaseMixedRealityProfile profile = null) : base(registrar, inputSystem, name, priority, profile)
+        {}
+
+        /// <summary>
+        /// Return the service profile and ensure that the type is correct.
+        /// </summary>
+        public MixedRealityInputRecordingProfile InputRecordingProfile
+        {
+            get
+                {
```
Troy-Ferrell: nit: align brackets

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractionTests_Press.cs
```diff
@@ -0,0 +1,127 @@
...
+using UnityEngine;
+using UnityEngine.Playables;
+using UnityEngine.TestTools;
+using System.Collections;
+
+namespace Microsoft.MixedReality.Toolkit.Tests
+{
+    public class InteractionTests_Press
+    {
+        // /// Test InteractionTest_Press_Far
```
Troy-Ferrell: needed?

Packages/manifest.json
```diff
@@ -3,10 +3,10 @@
     "com.unity.ads": "2.3.1",
     "com.unity.analytics": "3.2.2",
     "com.unity.collab-proxy": "1.2.15",
-    "com.unity.package-manager-ui": "2.0.3",
+    "com.unity.package-manager-ui": "2.0.7",
```
Troy-Ferrell: remove files

SimonDarksideJ:
> Do we need this commit history? Some of our tooling for viewing code reviews only allows for 64 commits. If we could get these commits squashed it may improve our review process.

This can be solved by doing a squash and merge when the PR is accepted.

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputPlaybackService.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// </summary>
+        InputAnimation Animation { get; set; }
+
+        /// <summary>
+        /// True if the animation is currently playing.
+        /// </summary>
+        bool IsPlaying { get; }
+
+        /// <summary>
+        /// The local time relative to the start of the animation.
```
wiwei: Time is always a tricky thing - it would be helpful to know what the units are - i.e. are these milliseconds since epoch? Milliseconds since player start?

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputRecordingService.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        /// <remarks>
+        /// If recording is limited any input older than the RecordingBufferTimeLimit will be discarded.
+        /// </remarks>
+        bool UseBufferTimeLimit { get; set; }
+
+        /// <summary>
+        /// Size of the input recording buffer.
+        /// </summary>
+        float RecordingBufferTimeLimit { get; set; }
```
wiwei: It would also be helpful to add some clarification here about the units - i.e. the comments say "size of input recording buffer" but the variable name is "time limit", which is a bit add odds with each other (i.e. maybe it's size in seconds? or really, duration, in seconds? Or minutes, or milliseconds, etc)

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputRecordingService.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        /// <param name="filename">Name of the file to create.</param>
+        /// <param name="directory">Directory in which to create the file. If null the persistent data path of the app is used.</param>
+        /// <returns>File path where input has been recorded.</returns>
+        string SaveInputAnimation(string filename, string directory = null);
+
+        /// <summary>
+        /// Generate a file name for export.
+        /// </summary>
+        string GenerateOutputFilename();
```
wiwei: I haven't read on quite yet, but looking at this function and the ones above it, it almost seems like this is an implementation detail of how SaveInputAnimation(directory) works. Am I reading this correctly in that, in there were only a single overload (SaveInputAnimatino(filename, directory)), GenerateOutputFilename wouldn't be needed?

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputRecordingService.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Provides input recording into an internal buffer and exporting to files.
+    /// </summary>
+    public interface IMixedRealityInputRecordingService : IMixedRealityInputDeviceManager
+    {
+        /// <summary>
+        /// Input is being recorded.
```
wiwei:
```suggestion
        /// True if input is being recorded.
```

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputRecordingService.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// </summary>
+        bool IsRecording { get; }
+
+        /// <summary>
+        /// Limit the size of the recording buffer.
+        /// </summary>
+        /// <remarks>
+        /// If recording is limited any input older than the RecordingBufferTimeLimit will be discarded.
+        /// </remarks>
+        bool UseBufferTimeLimit { get; set; }
```
wiwei: Does this variable and the one below it have any effect after StartRecording has been called? Or is the contract that these only affect the NEXT invocation of StartRecording?

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+        {
+            if (handedness == Handedness.Left)
+            {
+                return handJointCurvesLeft.TryGetValue(joint, out curves);
+            }
+            else if (handedness == Handedness.Right)
+            {
+                return handJointCurvesRight.TryGetValue(joint, out curves);
+            }
+            curves = new PoseCurves();
```
wiwei: Nit: What happens here if we just null out curves (i.e. no need to allocate for what is an error case right?)

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+            else if (handedness == Handedness.Right)
+            {
+                AddHandStateKey(time, isTracked, isPinching, handTrackedCurveRight, handPinchCurveRight);
+            }
+        }
+
+        /// <summary>
+        /// Add a keyframe for one hand joint.
+        /// </summary>
+        public void AddHandJointKey(float time, Handedness handedness, TrackedHandJoint joint, MixedRealityPose jointPose, float positionThreshold, float rotationThreshold)
```
wiwei: Nit: Add a comment about what the threshold is (it's a squared magnitude thing)

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+            AddFloatKey(curveX, time, rotation.x);
+            AddFloatKey(curveY, time, rotation.y);
+            AddFloatKey(curveZ, time, rotation.z);
+            AddFloatKey(curveW, time, rotation.w);
+        }
+
+        /// Add a float value to an animation curve
+        private static int AddFloatKey(AnimationCurve curve, float time, float value)
+        {
+            // return curve.AddKey(time, value);
```
wiwei: Delete commented line of code

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+                if (angle0 <= sqrThreshold && angle1 <= sqrThreshold)
+                {
+                    curveX.RemoveKey(iX);
+                    curveY.RemoveKey(iY);
+                    curveZ.RemoveKey(iZ);
+                    curveW.RemoveKey(iW);
+                }
+            }
+
+            // TODO make use of Bezier interpolation to allow more aggressive compression
```
wiwei: If you have a pointer to the how-to aspect of this (i.e. a link or any other description) that would be helpful - otherwise for future generations this might just be an immediate dead end (i.e. no way of knowing how to take action on this TODO)

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+            AddFloatKey(curveZ, time, rotation.z);
+            AddFloatKey(curveW, time, rotation.w);
+        }
+
+        /// Add a float value to an animation curve
+        private static int AddFloatKey(AnimationCurve curve, float time, float value)
+        {
+            // return curve.AddKey(time, value);
+
+            // Use linear interpolation to avoid overshooting from bezier tangents
```
wiwei: I'm new to this but can you show me where the lerping is occuring here? I would have expected to see some sort of lerp call (or some non-zero keyframe params or something that might be like a lerp)

wiwei: Or is it just that, the lerp occurs over time to the specified value

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+            // TODO make use of Bezier interpolation to allow more aggressive compression
+            // and use tangents and weights to accurately merge adjacent splines.
+            AddFloatKey(curveX, time, rotation.x);
+            AddFloatKey(curveY, time, rotation.y);
+            AddFloatKey(curveZ, time, rotation.z);
+            AddFloatKey(curveW, time, rotation.w);
+        }
+
+        /// Add a float value to an animation curve
+        private static int AddFloatKey(AnimationCurve curve, float time, float value)
```
wiwei: Nit: Might be worth adding what this returns (I think the index into the animation curve)

Same for the other things down below.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+        /// Uses binary search.
+        /// </remarks>
+        private static int FindKeyframeInterval(AnimationCurve curve, float time)
+        {
+            var keys = curve.keys;
+            int lowIdx = -1;
+            int highIdx = keys.Length;
+            while (lowIdx < highIdx - 1)
+            {
+                int midIdx = (lowIdx + highIdx) >> 1;
```
wiwei: Is there a measureable difference in the perf of a shift vs simple divide by 2?

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+                else
+                {
+                    highIdx = midIdx;
+                }
+            }
+            return lowIdx;
+        }
+
+        /// Sort marker array by time values.
+        private void SortMarkers()
```
wiwei: Is this function still used? Doing a Ctrl+F on this page doesn't find anything

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
```diff
@@ -0,0 +1,710 @@
...
+
+        /// <summary>
+        /// Change the time of the marker at the given index.
+        /// </summary>
+        public void SetMarkerTime(int index, float time)
+        {
+            InputAnimationMarker marker = markers[index];
+            markers.RemoveAt(index);
+
+            int newIndex = FindMarkerInterval(time) + 1;
```
wiwei: This function is basically just a Remove and then Add.
In order to reduce code duplication, consider just doing:

InputAnimationMarker marker = markers[index];
RemoveMarker(index);
AddMarker(marker);

wiwei: Plus that way if you ever get an off by one error (especially dealing with that new index + 1 business) assuming Add/Remove works, then SetMarkerTime will always work.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimationSerializationUtils.cs
```diff
@@ -0,0 +1,184 @@
...
+
+            Keyframe[] keys = new Keyframe[keyframeCount];
+            for (int i = 0; i < keyframeCount; ++i)
+            {
+                keys[i].time = reader.ReadSingle();
+                keys[i].value = reader.ReadSingle();
+                keys[i].inTangent = 0.0f;
+                keys[i].outTangent = 0.0f;
+                keys[i].inWeight = 0.0f;
+                keys[i].outWeight = 1.0e6f;
```
wiwei: Nit: It might be worth having 1.0e6f as a constant (since it's being used elsewhere where binary data is involved). Otherwise, it's a somewhat magic number that would possibly be easy to miss by someone else making a change in the future.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimationSerializationUtils.cs
```diff
@@ -0,0 +1,184 @@
...
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Functions for serializing input animation data to and from binary files.
+    /// </summary>
+    public static class InputAnimationSerializationUtils
+    {
+        private static readonly int jointCount = Enum.GetNames(typeof(TrackedHandJoint)).Length;
+
+        public const string Extension = "bin";
```
wiwei: Is there any way that we can write a version or something to this as well, so that when making changes to this in the future, we know that when we read things, we shouldn't try to read an incompatible file?

I'm a bit concerned about future changes where, say we actually do want to start writing non zero values for the following fields for bool values:

keys[i].inTangent = 0.0f;
keys[i].outTangent = 0.0f;
keys[i].inWeight = 0.0f;

If someone had their own recorded files, our new reader would try to read in those values but since they were never written out with the older format, things would probably get corrupted pretty quickly.

Assets/MixedRealityToolkit.Services/InputAnimation/InputPlaybackService.cs
```diff
@@ -0,0 +1,190 @@
...
+        public InputPlaybackService(
+            IMixedRealityServiceRegistrar registrar,
+            IMixedRealityInputSystem inputSystem,
+            string name = null,
+            uint priority = DefaultPriority,
+            BaseMixedRealityProfile profile = null) : base(registrar, inputSystem, name, priority, profile)
+        {}
+
+        private IInputSimulationService inputSimService = null;
+        private IInputSimulationService InputSimService => inputSimService ?? (inputSimService = MixedRealityToolkit.Instance.GetService<IInputSimulationService>());
```
wiwei: Similar question that @Troy-Ferrell asked @davidkline-ms earlier

Assets/MixedRealityToolkit.Services/InputAnimation/InputRecordingService.cs
```diff
@@ -0,0 +1,317 @@
...
+            if (hand == null)
+            {
+                recordingBuffer.AddHandStateKey(time, handedness, false, false);
+                return false;
+            }
+
+            bool isTracked = (hand.TrackingState == TrackingState.Tracked);
+
+            // Extract extra information from current interactions
+            bool isPinching = false;
```
wiwei: Do you know if it's actually possible to be pinching if not tracked?

julenka:
Wow this looks really promising! Can't believe I misssed it. Can't wait to try it out :-)

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/InputRecordingControls.cs
```diff
@@ -0,0 +1,90 @@
...
+    public class InputRecordingControls : MonoBehaviour
+    {
+        private IMixedRealityInputRecordingService recordingService = null;
+        private IMixedRealityInputRecordingService RecordingService
+        {
+            get
+            {
+                if (recordingService == null)
+                {
+                    recordingService = MixedRealityToolkit.Instance.GetService<IMixedRealityInputRecordingService>();
```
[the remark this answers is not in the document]
lukastoenneMS: Done

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractionTests_Press.cs
lukastoenneMS: Must have missed this, removed now. I'm going to look into writing proper tests for the recording feature.

Packages/manifest.json
lukastoenneMS: done

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputRecordingService.cs
lukastoenneMS: These values will affect the recording buffer immediately. If you enable or shorten the limit while recording the excess keyframes are pruned right away. Disabling or extending the limit just allows the buffer to grow over time.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: The Unity Keyframe ctor takes values for tangents and weights for Bezier interpolation. I'm only creating linear interpolation keyframes though, by setting tangents and weights to zero.

I thought about utilizing additional interpolation features of Bezier splines to accurately record movement with fewer keyframes, but i don't have time to figure out the math for that and linear interpolation does the job for us. For reducing file size it's more important to have a sane binary format (e.g. glTF).

I've updated comments to explain a bit better.

If you feel like diving into Bezier curve math i can recommend this site: [https://pomax.github.io/bezierinfo/](https://pomax.github.io/bezierinfo/)

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: That's my C coder instinct shining through 😉 
Probably doesn't matter, but doesn't hurt either.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: Not used any more. Markers are only inserted one-by-one now and don't need global sorting. Removing.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: done

Assets/MixedRealityToolkit.Services/InputAnimation/InputRecordingService.cs
lukastoenneMS: done

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputPlaybackService.cs
lukastoenneMS: Improved comments.

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputRecordingService.cs
lukastoenneMS: Added explanation in comments.

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputRecordingService.cs
lukastoenneMS: done

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: Improved comments.

The squaring is just a minor internal performance thing to avoid square root in vector magnitude.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: done

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: Remove this TODO comment. Initially i considered using Bezier curves for optimization but that's a bit involved and linear interpolation is enough for the time being.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: done

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: Not quite, the marker can get inserted in the middle of the list, so that the time values remain sorted. If i just use markers.Add(marker) it can jumble the time values.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimationSerializationUtils.cs
lukastoenneMS: True.

Note that this value will not end up in exported data, it's just a way to create quasi-boolean animation curves represented by float values. The weight just needs to be arbitrarily large so that interpolation will effectively return zero or one over the whole interval until the value changes with the next keyframe.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimationSerializationUtils.cs
lukastoenneMS: Yeah, i will add a header for this so we can tell versions apart.

Note that i would rather get rid of this custom format ASAP and use glTF by default. I've got a branch with a proof-of-concept glTF exporter implementation, but can't say when this can be ready for use: https://github.com/lukastoenneMS/MixedRealityToolkit-Unity/tree/input_recording_gltf

Assets/MixedRealityToolkit.Services/InputAnimation/InputPlaybackService.cs
lukastoenneMS: Fixed.

Assets/MixedRealityToolkit.Services/InputAnimation/InputRecordingService.cs
lukastoenneMS: On the side of the input simulation this will never happen, because the hand device is destroyed when it stops tracking.

For the recorded input data itself, however, these are just separate animation tracks, there is no logic that prevents keyframes in the "IsPinching" curves when the "IsTracking" curve has a True value. It's not worth adding that kind of complication.

Nothing bad will happen when if you set the hand data to (isTracked=False, isPinching=True), the sim service just won't create a hand device and isPinching is ignored.

Assets/MixedRealityToolkit.Services/InputAnimation/IMixedRealityInputRecordingService.cs
lukastoenneMS: I've moved this into the static serialization utils now, and just pass a base name and appendTimestamp flag parameters instead of using the profile.

In this PR the function seems a bit pointless, but in [the follow-up PR to add editor recording tools](https://github.com/lukastoenneMS/MixedRealityToolkit-Unity/pull/2/files) it's used to initialize the suggested filename for export file selection dialogs.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimation.cs
lukastoenneMS: Yes, returning null now. I was using a struct for PoseCurves before which means null wouldn't work (and iirc would be stack-allocated), but for various reasons it's easier to use a class for the PoseCurves collection.

Assets/MixedRealityToolkit.Services/InputAnimation/InputAnimationSerializationUtils.cs
lukastoenneMS: Input animation format now has a simple header with a magic number (for identifying invalid files up front), a major and a minor version number. The rest of the file could change and we can decide to do versioning for backwards compatibility.

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/InputRecordingControls.cs
lukastoenneMS: done!

Issue #4783, opened 2019-06-07
Far clip plane not settable in MRTK profiles


Troy-Ferrell:
## Describe the bug

Cannot set far clip plane in MRTK camera profile. 

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.11f1]
- MRTK Version [e.g. v2.0]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

PR #4784, opened 2019-06-07
Initial addition of far plane property

Troy-Ferrell:
## Overview
This changes adds the far clip plane property for both the opaque and transparent system in the camera profile. This is important for MRTK developers to set different far clip plane values for opaque vs transparent. Default in Unity is 1000m which for Hololens is well outside the typical rendering/user volume. However, this large value (1000m) for the far clip plane reduces depth fidelity, especially for 16-bit. 

This change also updates the default camera profiles to set Transparent far to 50m and Opaque far to 1000m. 

## Changes
- Fixes: #4783 
Far clip plane not settable in MRTK profiles 

## Verification

Tested profiles actually setting correct values on Camera in Unity editor

Tested camera profiles without serialized far field in text file. MRTK takes default set in the script. 

Tested far clip plane (transparent) actually set on HL device with objects not rendered

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityCameraProfile.asset
+++ b/Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityCameraProfile.asset
@@ -14,10 +14,12 @@ MonoBehaviour:
   isCustomProfile: 0
   isCameraPersistent: 0
   nearClipPlaneOpaqueDisplay: 0.1
+  farClipPlaneOpaqueDisplay: 1000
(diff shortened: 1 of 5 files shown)


Troy-Ferrell:
> Nice change.
> 
> One question before approving: If the far clip plane properties are not in the profile, what is the behavior?

If the far clip plane properties are not in the asset file as serialized fields, then the system uses the defaults provided in the c# code (which are 1000 and 50)

wiwei:
@davidkline-ms changes look good to me, I'll push this in in 24 hours unless you have additional concerns/feedback.

Issue #4451, opened 2019-05-17
Teleportation for the HoloLens experience


cre8ivepark:
## Describe the problem
In HoloLens experience, there are situations that require movement inside a large object. (e.g. immersive view of architecture model). Moving camera is not recommended which can cause motion sickness. Need a way to support MRTK's existing teleportation(point destination and commit, fade transition) in the HoloLens apps too. 

## Describe the solution you'd like
We could use hand ray with the parabolic pointer when pointed the teleportable floor area. 

## Additional context
Partner requests

polar-kev:
As of MRTK 2.6, gesture teleport is supported on all platforms.

MRTK 2.6 release notes:
https://docs.microsoft.com/en-us/windows/mixed-reality/mrtk-unity/release-notes/mrtk-26-release-notes#teleportation-with-the-teleport-gesture-now-supported-on-all-platforms

PR #4785, opened 2019-06-07
Fix teleport pointer to support mrtk object or standalone service

davidkline-ms:
While working on an experimental standalone teleport service, it was discovered that there were hard coded references to the MixedRealityToolkit object that was blocking the standalone service.

This change updates how the teleport pointer behaves and has been tested in the following configurations:

* MRTK input and teleport enabled
* MRTK input enabled and teleport disabled
* MRTK input enebled, teleport enabled with standalone teleport service prototype

Each of the above configurations behaved as expected (teleport supported, not supported and supported, respectively).

This change is needed to ensure support for a standalone teleportation service is possible without requiring a new version of the mrtk core assembly.


cre8ivepark:
David, would this help to support this request #4451 from partners?

davidkline-ms:
> David, would this help to support this request #4451 from partners?

This is a separate issuefrom #4451. That request still needs investigation.

Issue #4802, opened 2019-06-10
Scene System Overview link is broken in Mixed Reality Configuration Guide


Railboy:
## Describe the Issue

Scene System Overview link in Mixed Reality Configuration Guide points to a missing page.

## Existing Doc Link

[Mixed Reality Configuration Guide - Scene System](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/MixedRealityConfigurationGuide.html#scenesystem)

Currently resolves to: [Scene System Getting Started](https://microsoft.github.io/SceneSystem/SceneSystemGettingStarted.md)

Should resolve to: [Scene System Getting Started](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/SceneSystem/SceneSystemGettingStarted.html)

PR #4803, opened 2019-06-10
Fix Scene System Overview link in Mixed Reality Configuration Guide

Railboy:
## Overview
Removed leading slash from scene system overview link.

## Changes
- Fixes: #4802

## Verification
@keveleigh Can you verify that this fixes the issue?


keveleigh:
Tested locally, and the link is fixed.

PR #3755, opened 2019-04-01
Merge the HL2 payload into mrtk_development

wiwei:
Brings the latest HL2 payload up into mrtk_development.

Note that there is still an open issue where the articulated hands visualization (NOT the simulator, the actual hands) doesn't work, which is being looked into.

PR #4804, opened 2019-06-10
Fix controller enum values in EditorWindowOptions.json

keveleigh:
## Overview
The SupportedControllerType enum was changed to flags in #3755, but I missed the fact that this enum is used in the json filed used to position the mapping positions. The only thing changed is the `"Controller":` values.

Broken:
MicrosoftTeams-image (3)

Fixed:
MicrosoftTeams-image (4)


wiwei:
I'll have to see if I can find an autoformatter that we can just apply to our json files - as you say it's extremely difficult to read these sort of changes when they're all on the same line

PR #4805, opened 2019-06-10
Add get services by registrar to the registry

davidkline-ms:
This change adds the ability to query the service registry to retrieve all services registered by a specific registrar. This change reduces the amount of code custom registrars are required to write to find their services.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
+++ b/Assets/MixedRealityToolkit/Utilities/MixedRealityServiceRegistry.cs
@@ -247,5 +247,35 @@ namespace Microsoft.MixedReality.Toolkit
                 out _,                  // The registrar out param is not used, it can be discarded.
                 name);
         }
+
+        /// <summary>
+        /// Returns the collection of all services registered by the specified registrar.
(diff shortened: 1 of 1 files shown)


wiwei:
> This change reduces the amount of code custom registrars are required to write to find their services.

Something that I'm trying to understand a bit more - if a registrar is responsible for registering its services, why does a registrar need something else (i.e. this thing) to remember the work that it did itself (and is also tracking itself)? 

i.e. the interface surface here:
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs

Seems to suggest that you can GetService on the registrar itself - in theory this means that any implementation should be storing a list for the services that it Registered/Unregistered.

davidkline-ms:
Good questions @wiwei.

I agree that registrar's can (and perhaps should) track the services that they register. The initial intent was to use the registry itself as the store, rather than requiring a duplicate instance in memory.

I am fine not taking this change. It was more a convenience method than an requirement.

PR #4810, opened 2019-06-10
U/mattjo/refactor spatial localization

matthejo:
## Changes
- Fixes: # .


Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/ISpatialCoordinate.cs
```diff
@@ -85,4 +81,9 @@ public interface ISpatialCoordinate
         /// </summary>
         Quaternion CoordinateToWorldSpace(Quaternion quaternion);
     }
+
+    public interface ISpatialCoordinate<TIdentifier> : ISpatialCoordinate
+    {
+        TIdentifier Id { get; }
```
andreiborodin: Is there a super strong preference not to have everything string? I realize this means conversion back and forth, but shouldn't be too bad?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialLocalizer.cs
```diff
@@ -32,53 +29,68 @@ public abstract class SpatialLocalizer : MonoBehaviour
...
+            }
 
-        /// <summary>
-        /// Deinitializes the session.
-        /// </summary>
-        /// <param name="actAsHost">If true, this localizer will operate under the assumption that it is hosting the spatial coordinate system.</param>
-        /// <param name="token">The token representing the session.</param>
-        internal abstract void Uninitialize(bool actAsHost, Guid token);
+            SpatialCoordinateSystemManager.Instance.RegisterSpatialLocalizer(this);
+        }
```
chrisfromwork: Does this require an active game object/enabled script for registration? Is our thought that each type of SpatialLocalizer will have a component enabled in the scene in order for registration to work?

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/MarkerDetectorSpatialLocalizer.cs
```diff
@@ -30,47 +37,106 @@ private void OnValidate()
...
+                : base(settings)
+            {
+                this.localizer = localizer;
+            }
+
+            public override async Task<ISpatialCoordinate> LocalizeAsync(CancellationToken cancellationToken)
+            {
+                localizer.DebugLog("Getting host coordinate");
+                localizer.markerDetector.SetMarkerSize(Settings.MarkerSize);
+
```
chrisfromwork: it feels like this should be calling the SpatialCoordinateService compared to directly using the marker detector.

chrisfromwork: But this organization does raise the question of whether a SpatialLocalizer can support multiple sessions at the same time. Having multiple sessions seems beyond the scope of what we should test for getting things to initially work.

---
In reply to: [292199660](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4810#discussion_r292199660) [](ancestors = 292199660)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/MarkerDetectorSpatialLocalizer.cs
```diff
@@ -30,47 +37,106 @@ private void OnValidate()
...
+        public override bool TryDeserializeSettings(BinaryReader reader, out MarkerDetectorLocalizationSettings settings)
+        {
+            return MarkerDetectorLocalizationSettings.TryDeserialize(reader, out settings);
+        }
 
-            spatialCoordinateService = new MarkerDetectorCoordinateService(markerDetector, debugLogging);
+        public override ISpatialLocalizationSession CreateLocalizationSession(MarkerDetectorLocalizationSettings settings)
+        {
+            return new LocalizationSession(this, settings);
         }
```
chrisfromwork: I think we need to check whether a session already exists/is localizing compared to blindly creating them. With how this exists today, you may end up having multiple sessions using the same marker detector, which is going to fail to find all markers with how marker detectors are started/stopped currently.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialCoordinateSystemManager.cs
```diff
@@ -54,143 +36,249 @@ public class SpatialCoordinateSystemManager : Singleton<SpatialCoordinateSystemM
...
                 }
-                if (StateSynchronizationBroadcaster.IsInitialized)
+
+                participant.IsLocatingSpatialCoordinate = true;
+                var coordinate = await currentLocalizationSession.LocalizeAsync(CancellationToken.None);
+                if (participant.PersistentCoordinateId != null)
                 {
-                    StateSynchronizationBroadcaster.Instance.UnregisterCommandHandler(command, this);
+                    participant.Coordinate = new WorldAnchorSpatialCoordinate(participant.PersistentCoordinateId, coordinate.CoordinateToWorldSpace(Vector3.zero), coordinate.CoordinateToWorldSpace(Quaternion.identity));
                 }
```
chrisfromwork: do we feel coordinate persistence should be at the SpatialCoordinateSystemManager level compared to the participant/localizer/coordinate service? I can see some scenarios where things like ASA may have their own persistence story.

Also, with the way this is set up today, I don't know that it will work with android/ios. There is no WorldAnchorStore on iOS/Android if I understand correctly.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialLocalizer.cs
matthejo: Yes, some active GameObject is necessary for this script to be present on to register the localizer. I'm open to other ideas for how this could work, but this matches the pattern used for ComponentBroadcasterService registration

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/MarkerDetectorSpatialLocalizer.cs
matthejo: Yes, this should be going through the SpatialCoordinateService, I had slashed and burned to get things working. I'll be putting this back the way it was, with refactors necessary to actually configure the marker-specific version before localizing

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/MarkerDetectorSpatialLocalizer.cs
matthejo: That can't happen today as long as localization is controlled through the SpatialCoordinateSystemManager, which allows only one session to happen at a time

matthejo: And I think some central service/authority has to be the gatekeeper for how sessions are managed, since a specific SpatialLocalizer won't know about other localizers.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialCoordinateSystemManager.cs
matthejo: I think anchor-persistence is very specific to HoloLens, and probably won't work as-is with mobile. I think that the decision to create/use a WorldAnchor or not should be controlled by the system that initiates localization in the first place. The DSLR compositor can request the connection to use the anchor, but the same app could also have a mobile participant join and that connection should not use an anchor. That's why I added a separate type of command to request using this anchor so that it can be made very specific to the DSLR scenario.

Assets/MixedRealityToolkit.Extensions/Sharing/SpatialAlignment/Common/ISpatialCoordinate.cs
matthejo: The way I thought about this:
1. From within the implementation of a SpatialCoordianteSystem, the identifier has some concrete meaning and concrete type. That type is probably not a string in most cases, and can be strongly-typed.
2. From outside the implementation of a SpatialCoordinateSystem, the identifier is pretty meaningless and probably shouldn't be used for a purpose, unless I'm missing some purpose. It may or may not uniquely identify that coordinate within the same SpatialCoordinateSystem, and it certainly doesn't globally-uniquely identify a coordinate.

Given those two things are true, it made more sense to me to make the identifier a concrete value, and make it an optional thing that doesn't necessarily need to be implemented to have an ISpatialCoordinate.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialCoordinateSystemManager.cs
chrisfromwork: This seems like good reasoning for keeping this persistence logic in the SpatialCoordinateSystemManager.

---
In reply to: [292215769](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4810#discussion_r292215769) [](ancestors = 292215769)

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/MarkerDetectorLocalizationSettings.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.IO;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
+{
+    public class MarkerDetectorLocalizationSettings : ISpatialLocalizationSettings
+    {
+        public int MarkerID { get; set; }
+        public float MarkerSize { get; set; }
+
```
chrisfromwork: Could this take a reference to the desired marker detector/marker detector type? Then we may be able to use these settings to declare the marker detector compared to just the size and marker id

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialCoordinateSystemParticipant.cs
```diff
@@ -6,267 +6,359 @@
...
+                this.transform.position = Coordinate.CoordinateToWorldSpace(Vector3.zero);
+                this.transform.rotation = Coordinate.CoordinateToWorldSpace(Quaternion.identity);
             }
 
-            if (isHost)
+            if (SocketEndpoint != null && SocketEndpoint.IsConnected)
             {
-                hostWorldRotation = localWorldRotation;
+                SendCoordinateStateMessage();
             }
```
chrisfromwork: do we need to send this every frame? For keynote monitoring, i get that you may want to know the state of things. but for mobile experiences, this may be added noise on the main user's end when having multiple viewers.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialCoordinateSystemParticipant.cs
```diff
@@ -6,267 +6,359 @@
...
 using UnityEngine;
+using UnityEngine.XR.WSA;
 
 namespace Microsoft.MixedReality.Toolkit.Extensions.Experimental.SpectatorView
 {
     /// <summary>
     /// The SpectatorView helper class for managing a participant in the spatial coordinate system
     /// </summary>
-    internal class SpatialCoordinateSystemParticipant : DisposableBase
+    internal class SpatialCoordinateSystemParticipant : MonoBehaviour
```
chrisfromwork: >MonoBehaviour [](start = 56, length = 13)

Does this really need to be a monobehaviour? Settings transforms for this component every update doesn't really seem needed. Or at least, that wasn't the mechanism for using this transform before and it feels a bit unclear why its needed here now.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialCoordinateSystemParticipant.cs
blazingpotsie: Just a key-frame animator's two cents.  Would be nice if eventually it was sent every frame but for most functionality i think once every second or even every other second would be sufficient.  There will be various forms of Lag anyway.  This issue has been on my mind for weeks and I would hate to see the need for perfect getting in the way of it functioning correctly.  I am assuming very subtle delays will be necessary.

matthejo: The state message is checked every frame, but a message is only sent if the status ends up being different from the previous frame. This should only happen when:
1. Tracking is lost or gained
2. The WorldAnchor adjusts the position of a WorldAnchor-based coordinate
3. A new marker detection session starts or ends

If there's noise from this message type, we should figure out what's changing on a per-frame basis and try to alleviate it.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/SpatialCoordinateSystemParticipant.cs
matthejo: The reason I made this a MonoBehaviour was to get an update loop so that the state message can be checked every frame. For a WorldAnchor-based coordinate, there's unfortunately no event to indicate that the WorldAnchor needed to modify the transform's position and rotation, so the only way to detect those changes is to poll each frame.

As I mentioned in the other comment, the change message is only sent if it's different from the previous frame's message, so that should minimize network traffic from these status updates.

Assets/MixedRealityToolkit.Extensions/SpectatorView/Scripts/Sharing/MarkerDetectorLocalizationSettings.cs
matthejo: I think this would require some registry of available marker detectors, similar to the registry of available SpatialLocalizers. My thought here was that there can be multiple SpatialLocalizers, one per type of marker detector, but the SpatialCoordinateService they use can all be the MarkerDetectorCoordinateService with a different IMarkerDetector in its prefab.

luis-valverde-ms:
@matthejo , @chrisfromwork , is this still alive?

matthejo:
Work on this has moved, so I'm going to close this PR.

Issue #4411, opened 2019-05-16
[MRTK Shell Parity] [Fingertip visualization] Match the final shell design spec. (distance, ring size, alignment behavior)


cre8ivepark:
Part of #4200 UX Controls Parity with HoloLens 2 Shell

cre8ivepark:
@Cameron-Micka Could you please take a look at the latest shell and tweak the values to match?

Cameron-Micka:
Certainly! I might be low on time until late June. But, I would love to "fix" this.

ADP-JoNeff:
Note: Still need to tune the sizing of the ring relative to distance

cre8ivepark:
Completed with #4836 
@ADP-JoNeff Please re-open & submit PR if you see any additional tunning is needed.

Issue #4413, opened 2019-05-16
[MRTK Shell Parity] [Slate] Fingertip ring shadow on surface needed


cre8ivepark:
Part of #4200 UX Controls Parity with HoloLens 2 Shell

Cameron-Micka:
Fixed in #4836.

PR #4836, opened 2019-06-11
Finger Tip Cursor Pairity

Cameron-Micka:
## Overview
Part of #4200 UX Controls Parity with HoloLens 2 Shell
This pull request introduces a finger tip "shadow" using the ProximityLight system. Proximity lights can now have per material color overrides and a "subtractive light" option. The finger tip cursor has also been updated to mimic the behavior found in the shell (size, shape, color, movement, etc.).

ShellCursor
[Better quality video link.](https://microsoft-my.sharepoint.com/:v:/p/thmicka/EVsRccmGTuJHt4I_EL_y42YBDXbbEJ-gbVGKWx3Xfsb20Q?e=iT9srx)

OverrideColor

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4411 & https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4413

## Verification
To test these changes the slate found in the HandInteractionExamples scene best demonstrates the new changes.
>
> git fetch origin pull/4836/head:name_of_local_branch
>


Assets/MixedRealityToolkit.SDK/StandardAssets/Shaders/FingerTipCursor.shader
```diff
@@ -0,0 +1,262 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+/// <summary>
+/// Note, this shader is generated from a tool and is not formated for user readability.
```
wiwei: Thank you for having the tool write out this note :)

One point of feedback on this is that for folks who are wanting to change this in the future, having some pointer/guidance on what that tool is (i.e. check out http://www.github.io/whatevertool) or otherwise some other docs/info about how it gets generated.

Otherwise the immediate action for someone who is doing this is to reverse engineer the process by doing random searches until the find the tool.

wiwei:
Non-shader code looks good to me (I don't have enough background in shader code to have an opinion there)

Assets/MixedRealityToolkit.SDK/StandardAssets/Shaders/FingerTipCursor.shader
Cameron-Micka: Sadly the tool is a internal, custom one written by a member of the Shell team. I could reach out to them about open sourcing it, but that would probably take some time and effort. I could also spend a bit of time in the future re-formatting and commenting this shader for user readability (and for the hand triangle shader).

wiwei: Ah, yeah that would probably be more time than just now. 

I think that time is probably better spent here making the tool better (i.e. better formatting, maybe automatically adding comments where appropriate), than trying to fix up the output each time you regenerate it... it's a losing battle.

wiwei:
Would still probably want to get @cre8ivepark's take on this to comment on the parity-ness, also an area which I am not an expert.

Cameron-Micka:
> It seems like the color of the ring is a little bit darker than the shell one. (gray)
> Could you please update it to white? (or is it because of the lighting?)

It does look a little off now that you mention it. It's using the same shader, material, and model from the Shell (and the shader doesn't have any lighting calculations). I'll look to see if some shader input or color space issue is causing the inconsistency.

Cameron-Micka:
Thank you @cre8ivepark!

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Prefabs/Slate.prefab
```diff
@@ -550,8 +552,8 @@ MonoBehaviour:
   panZoomSmoothing: 80
   panEventReceivers: []
   centerPoint: {fileID: 0}
-  leftPoint: {fileID: 4531382727374914902}
-  rightPoint: {fileID: 5073403802680592322}
+  leftPoint: {fileID: 0}
```
keveleigh: Are these fields unneeded now?

Assets/MixedRealityToolkit.SDK/StandardAssets/Shaders/HandTriangles.shader
```diff
@@ -611,9 +611,6 @@ SubShader {
         UNITY_SETUP_STEREO_EYE_INDEX_POST_VERTEX(vxIn[0]);
         //huxEye = _WorldSpaceCameraPos;
         //workaround for Unity's auto updater in 5.6
```
keveleigh: Looks like the code was removed below, so should the comment be removed as well?

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Prefabs/Slate.prefab
Cameron-Micka: I left them in so this wasn't a breaking change. And, developers may want to attach gameObjects to the touch points? (Rather than just a shadow.)

Assets/MixedRealityToolkit.SDK/StandardAssets/Shaders/HandTriangles.shader
Cameron-Micka: I left the code in sense this shader was generated from a tool, and it may be easier to resolve changes if we take an update from that tool.

Issue #4121, opened 2019-04-30
Play mode not starting, highlighting MixedRealityToolkit gameobject instead


Alexees:
## Describe the bug
It sometimes happens that I want to go into play mode, which I can't enter (exits immediately) and MixedRealityToolkit gets a ping.
I dug into where that comes from and it seems that the Active Profile is null then. I have to restart Unity for this to be fixed. Since this is a SerializedField with a proper value in it, I'm puzzled how this can happen.

## To reproduce

I do not have steps to reproduce it yet

## Expected behavior

The Active Profile never goes null

## Screenshots

![image](https://user-images.githubusercontent.com/32508825/56967891-68c1a900-6b62-11e9-8aea-07360b2f60d1.png)

## Your Setup (please complete the following information)

- Unity Version 2018.3.13f1
- MRTK Version v2.0

## Target Platform (please complete the following information)

- Editor

Railboy:
I was able to repro this once but I'm not quite sure how I did it.

>I dug into where that comes from and it seems that the Active Profile is null then.

Yeah this seems to be the immediate culprit:
```
if (playModeState == UnityEditor.PlayModeStateChange.ExitingEditMode && activeProfile == null)
{
    UnityEditor.EditorApplication.isPlaying = false;
    UnityEditor.Selection.activeObject = Instance;
    UnityEditor.EditorGUIUtility.PingObject(Instance);
}
```
I have a suspicion ResetConfiguration is getting called by a dead / dying instance of the toolkit.

Railboy:
I'm able to reproduce this consistently when using git to switch between branches while Unity is open.

- Create a new scene
- Use git to switch to / from  prerelease/2019_rc1 / mrtk_development/other major branch
- Let Unity re-import assets and recompile
- Add a MixedRealityToolkit instance using _MixedRealityToolkit/Add to Scene and Configure_
- Select a configuration profile
- Try to enter playmode

The bug happens regardless of which branch I'm in.

Restarting the project or triggering recompilation makes the bug stop.

Avalin:
Just a headsup, this issue is still relevant today.
I'm using Unity 2019.1.6f and MRTK RC1 Refresh.

I reproduced it like this:

- Add scene with MRTK integration (Playspace + Toolkit)
- Add new scene without MRTK integration (No camera).
- Start the new scene without MRTK integration. 
- Go to scene with the MRTK camera implemented 
- Try to enter playmode

[...]

Issue #4832, opened 2019-06-11
Non-core services have their Enable() function called twice


wiwei:
## Describe the bug

Non-core services (such as the WindowsMixedRealityDeviceManager) have their Enable() functions invoked twice.

From what I can see, this was a regression from this PR, which started invoking the registrar:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/3738

And the registrar would internally also end up calling Enable() on the service as well as part of registration.

## To reproduce

1) Open the hand interaction example scene.
2) Set a breakpoint on WindowsMixedRealityDeviceManager::Enable()
3) See how it hits twice.

We shouldn't be calling enable twice.

Railboy:
@davidkline-ms If we remove this Enable call in RegisterServiceInternal, Enable is only called once (in MRTK's OnEnable).

```
private bool RegisterServiceInternal(Type interfaceType, IMixedRealityService serviceInstance)
{
    ...

    if (!isInitializing)
    {
        serviceInstance.Initialize();
=====>  serviceInstance.Enable(); 
    }

    return true;
}
```

It looks like these aren't meant to be called while the MRTK is initializing. But the instance's OnEnable is called after the instance has been initialized. Is this step included to cover situations where you register a service in the middle of a session, well after the instance has been initialized?

davidkline-ms:
@Railboy, that is exactly where i was going to focus my investigation :)

davidkline-ms:
> Is this step included to cover situations where you register a service in the middle of a session, well after the instance has been initialized?

I believe this was a side-effect of my attempting to reuse code with the least amount of modifications,

Railboy:
@davidkline-ms I ask because I'm cleaning up instance registration in #4838 and **IF** there's a straightforward fix, this could easily be folded into that PR as a relevant issue.

davidkline-ms:
@Railboy, feel free to make the change and add me to the PR. I can help validate.

Railboy:
Committed.

davidkline-ms:
fixed by @Railboy

PR #4838, opened 2019-06-11
MRTK Instance Application Event Handling / Deterministic Instance Selection

Railboy:
## Overview
Moves all editor / application event listening to static handlers. Previously MixedRealityToolkit instances were responsible for monitoring application state changes individually. In practice this led to dead instances stopping play mode when they couldn't find a config profile.

Registered instances are now sorted by instance ID so that automatic active instance selection is deterministic.

## Changes
- Fixes: #4121

## Verification
- All EditMode tests pass.
- ~~All PlayMode tests pass if executed one at a time. When executed all at once, PressableButtonTests fail due to a GestureRecognizer error. @julenka is this a regression?~~
- All PlayMode tests pass after merging latest changes from dev branch.


wiwei:
@Railboy regarding the tests, can you check that your fork has the commits from https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4498?

He mentioned that there was some weird physics update bug, though fair warning if the failures you're seeing are completely deterministic in that they happen every single time, it could be a different issue.

Railboy:
@wiwei Tests still fail individually with latest mrtk_development merged in. One-by-one they all pass.

Tried this with the mrtk_development branch and I get the same result, so maybe it's expected behavior?

googlan:
Play mode test error was a bug and should be fixed by #4875 .

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -848,7 +820,7 @@ private bool RegisterServiceInternal(Type interfaceType, IMixedRealityService se
             if (!isInitializing)
             {
                 serviceInstance.Initialize();
-                serviceInstance.Enable();
+                //serviceInstance.Enable();
```
Ziugy: Is this supposed to be re-enabled or removed?

Railboy: This addresses #4832. I'll remove the comment.

julenka:
@railboy could you also please confirm that you verified #4121 no longer repros, and also that you tested HandInteractionExample scene? I didn't see that in the PR comments but want to make sure that both cases are tested before we complete the merge.

Railboy:
@julenka Correct, with this change in place I can no longer repro #4121 in HandInteractionExample or any other scene.

Issue #4453, opened 2019-05-17
[Proposal] Reduce the number of core assemblies


davidkline-ms:
## Describe the problem
We have received customer feedback that the number of MRTK core assemblies is overwhelming (in VS, etc). They have requested that we combine them into a single core assembly and a single core-editor assembly.

This _may_ also have the desired effect of reducing the number of instances of Unity reporting a failed dll copy on project (re)load.

## Describe the solution you'd like

Ultimate goal is to reduce the core assembies to the following:
- Microsoft.MixedReality.Toolkit
- Microsoft.MixedReality.Toolkit.Editor

This would eliminate:
- Microsoft.MixedReality.Toolkit.Core.Build
- Microsoft.MixedReality.Toolkit.Core.Definitions.Utilities.Editor
- Microsoft.MixedReality.Toolkit.Core.Extensions.EditorClassExtensions
- Microsoft.MixedReality.Toolkit.Core.Inspectors.ServiceInspectors
- Microsoft.MixedReality.Toolkit.Core.Inspectors
- Microsoft.MixedReality.Toolkit.Core.Utilities.Async
- Microsoft.MixedReality.Toolkit.Core.Utilities.Editor 
- Microsoft.MixedReality.Toolkit.Player
- Microsoft.MixedReality.Toolkit.Utilities.Gltf
- Microsoft.MixedReality.Toolkit.Utilities.Gltf.Importers

provencher:
I think Examples should be in a separate assembly, but all be grouped together.

Specific supported platforms should also be their own assemblies, so that when the package manager solution goes live, you can only bring in the examples you want.

provencher:
Personally, the way that @StephenHodgson setup the XRTK makes the most sense to me.
https://github.com/XRTK

davidkline-ms:
@provencher, there are currently no plans to bundle the services or platforms. Examples are something that needs to be further thought out

Railboy:
I like the simplicity but this might be case where devs don't realize what they'd be giving up.

Assemblies keep editor compile times down when editing scripts in large projects. Maybe we can measure the impact? I recall it being pretty miserable before they were introduced.

They're also free best-practice enforcement. They make it impossible to write the kind of spaghetti that made the HTK hard to consume.

provencher:
I remember speaking to @andreiborodin about how all these assemblies would eventually simply be imported via nuget or some other package management system. 

[...]

PR #4841, opened 2019-06-11
Reduce count and improve assembly naming consistency

davidkline-ms:
This change combines core and some editor assemblies to reduce the number of dlls generated and visual studio projects in the solution.

It also moves most editor related code into a single Editor folder.

Lastly, it updates all assembly names to follow the pattern of Micosoft.MixedReality.Toolkit.<name> and renames the asmdef files to match.

Gets as close as is currently possible (without significant code changes) to #4453.

Testing performed:
- Solvers scene
    - Verify input, teleport, diagnostics and boundary systems function as expected
- Edit mode tests pass


wiwei:
I'll probably hold off reviewing this until the build comes back green (looks like it fails pretty early on in CI).

I do have some general questions however about this change, since it looks like there was some discussion on the related issue, but it seems like we didn't fully land on all of open issues before sending out this PR:

1) This optimizes heavily for consumers of the MRTK who aren't making day to day changes to the MRTK, which is great for consumers. For folks who are working on the MRTK itself, it means that build times for making changes to things like inspectors (which is super common) will now go up. There are a few possibilities that could exist here:

- Consumer focus is so high that from an MRTK contributor perspective, any amount of pain we have is okay because we should be willing to take really large build times (i.e. 30 minutes) if it marginally increases consumer benefit it will be worth it regardless of the cost. This example is explicitly chosen to be arbitrarily high to emphasize that this path involves no cost/benefit weighing.
- There is some tradeoff where we can weigh benefit to consumers (i.e. reduced assembly count) to cost to contributors (i.e. 30 seconds additional build time per change)

It feels like our philosophy right now is #1, even though it feels like we should have done the due diligence to measure the impact of this change (which AFAICT wasn't done). This could be as informal as say, having one repo with changes one without, making some changes to some editor code and then using a stopwatch to time difference.

2) It seems like this only solves a subset of the problems - i.e. how many assemblies still remain after this? It feels like there may be a more general solution that involves CI-pipeline work that would post-process the asmdefs to collapse them into a smaller count (similar to how UnityEngine.dll has its own set of smaller DLLs but in the end you link against UnityEngine.dll)

I just want us to be very explicit about the degree to which we're solving the problem and the due diligence being done here (and to write it down)

davidkline-ms:
> l probably hold off reviewing this until the build comes back green (looks like it fails pretty early on in CI).

Should be fixed. Thanks to @keveleigh for spotting the breaking change.

> I just want us to be very explicit about the degree to which we're solving the problem and the due diligence being done here (and to write it down).

Agreed. 

The goal of this PR is to optimize for the largest set of customers (i.e. non-contributors). Based on what was discovered working on this, it was not possible to remove as many assemblies as was originally desired (were able to merge three editor assemblies down to one).

The largest benefit of this change is the improved consistency in assembly naming (some were not following the pattern of "Microsoft.MixedReality.Toolkit.<name>".

davidkline-ms:
Not encountering errors building the .net backend version on my local PC. retrying the CI

davidkline-ms:
closing to revisit in the next major release as the code layout and dependencies need to be thoroughly reviewed analyzed

Issue #4557, opened 2019-05-24
BoundingBox doesn't recompute bounds when boundsOverride is modified


lukastoenneMS:
## Describe the bug

BoundingBox reacts to transform changes of the Target object to recompute bounds. However, if a BoxCollider is used for boundsOverride to take the bounds values from, changes to the BoxCollider will not trigger a bounds update.

There is no equivalent "hasChanges" flag on BoxCollider, as there is on Transform (aside from the questionable practice of resetting the Target.transform.hasChanged flag from BoundingBox). Quick hack solution was to add a custom "HasDirtyBounds" flag that can be set externally to force a bounds update.

Problematic code:
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/47663d2a35becea2fd6db8b8cd4ec238507c3ad8/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs#L485

## To reproduce

1. Open the HandInteractionExamples scene
1. Select one of the bounding box examples, e.g. "ActivateByProximityAndPointer"
1. Go into play mode
1. Press T to show the left hand, hold shift, move the hand over the object so the bounding box is shown
1. Select the scene tab so you can see the bounds
1. Change the scale of the target object, observe how the bounds adapt
1. Change the size of the box collider, observe how the bounds _don't_ adapt

## Expected behavior

BoundingBox also works with a boundsOverride collider set.

## Screenshots

Initial bounds:
![image](https://user-images.githubusercontent.com/47210016/58322551-8f9d9300-7e21-11e9-90b3-a347b3e6ef6f.png)

Changing Target transform works:
![image](https://user-images.githubusercontent.com/47210016/58322613-c1165e80-7e21-11e9-9937-5243e63eb669.png)

Changing BoxCollider size does not work:
![image](https://user-images.githubusercontent.com/47210016/58322649-e0ad8700-7e21-11e9-9b8f-07ab30093d0c.png)

julenka:
How about the following solution approach: expose a method in bounding box "recalculate bounds" that would recompute the bounds? Whoever is changing the bounding box would need to call that method.

hdaUwe:
> 
> 
> How about the following solution approach: expose a method in bounding box "recalculate bounds" that would recompute the bounds? Whoever is changing the bounding box would need to call that method.

[...]

PR #4842, opened 2019-06-12
Bounding Box update when boundsOverride changes

julenka:
## Overview
- Fixes: #4557 
- Add test case in BoundingBoxRuntimeExample


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -788,6 +799,22 @@ private void Update()
...
             }
+            else if (HasBoundsOverrideChanged())
+            {
+                UpdateBounds();
+                UpdateRigHandles();
+            }
+        }
+        private bool HasBoundsOverrideChanged()
+        {
+            if (boundsOverride == null)
```
wiwei: Is it not supported to be able to set the boundsOverride to null? That is, it seems like it's possible to go like:

No bounds override -> bounds override A -> bounds override B -> bounds override A

But it looks like you can't be like:

No bounds override -> bounds override A -> No bounds override

This is fine if it's documented and there's no scenario in which that's useful. It's just generally good to be explicit about supported state changes (in the doccomments)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -788,6 +799,22 @@ private void Update()
...
+            {
+                return false;
+            }
+            Bounds curBounds = boundsOverride.bounds;
+            bool result = curBounds != prevBounds;
```
wiwei: Do you know if boundsOverride can really only be updated by changing BoundsOverride, or if there are also pretty common cases where the property itself is changed directly (i.e. play-mode live edits of properties?) I assume that the play mode live edits is a thing but just wanted to confirm.

The reason I ask is, otherwise it seems like the only place where the "bounds is dirty" state could be flipped is in BoundsOverride itself - as such it would seem to be more performant (though by how much, I have no idea) to just set a dirty bool in the setter, and then reset it here (so that you don't need to do a multi compare here on each update tick).

Even if the answer to the first is no, I don't think there's actually any action to be taken here (I don't have enough background on the significant of perf implications, and don't generally believe in making premature 'optimizations')

But just wanted to go through the thought process.

wiwei:
Minor comments, generally seems fine but not hitting approve because I have those open Q's, but not hitting block because I don't believe those should actually block.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
julenka: boundsOverride is usually modified directly, not through the accessor, for example like this:

```
var newObject = new GameObject();
var bc = newObject.AddComponent<BoxCollider>();
bc.center = new Vector3(.25f, 0, 0);
bc.size = new Vector3(0.162f, 0.1f, 1);
bbox.BoundsOverride = bc;
// This line of code directly modifies bc.size, no way I can think of to set a 'dirty big'
bc.size = new Vector3(0.5f, 0.1f, 1);
```

I don't think this adds much overhead because we already do quite a lot of comparing previous / current state in the Update() loop.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
julenka: Why can't you set boundsOverride to null? I have it in my test cases and it works fine. 

```
SetStatus("Remove target bounds override");
bbox.BoundsOverride = null;
Destroy(newObject);
newObject = null;
```

The resulting behavior is that the bounding box will snap to the bounds of the children.

I can add a comment that says "if you set boundsOverride to null, the bounding box will match the bounds of the game object or children.

wiwei: That's what I was trying to understand - if the bounds override is currently NOT null, then setting it to null will make HasBoundsOverrideChanged() report false (meaning, nothing has changed). However, in this example, you have changed something (you went from an override to no override)

I'm guessing that I'm missing something else (i.e. CreateRig gets called through some other path, or CreateRig doesn't need to be called when going from override to no override?)

wiwei: I'm not going to block on this one - I just want to understand the disconnect here between the supporting transitions to null and also making sure that we reevaluate state when transitioning to null.

wiwei: I'm not going to block on this one - I just want to understand the disconnect here between the supporting transitions to null and also making sure that we reevaluate state when transitioning to null.

thalbern: I see the same issue as will here - if I understand this correctly you should return true in this method in null case as well when the override was previously set.

julenka: Thanks for the comments. I don't want to return true if boundsOverride is null because that would mean we update rig and handles every frame that boundsOverride == null. I think the name of the method "HasBoundsOverrideChanged" is confusing so I added a comment (I could also call it HasBoundsOverrideBoundsChanged but that seems really long). 

Can you have a look and see if this new method makes more sense to you? I verified the new method works with my test scene.

Issue #4827, opened 2019-06-10
Editor Hand Simulation Mesh Collapsing to a single point


anuitz:
## Describe the bug

A few of us are hitting an issue where the joints in the editor hand simulation appears all on the exact same point. We are consuming the v2.0.0-20190521.3 MRTK nuget package via MSBuild (not nuget for unity). We can't debug the issue because there are no debug symbols, which makes this rather challenging to solve on our own. Does anyone have any ideas about what is causing this and how to fix it?

## To reproduce

1. Make any scene and add MRTK to the scene and configure
2. Press Play
3. Press whatever key shows the hand sim

## Expected behavior

A full hand mesh. Instead there is the 3d gizmo and a bunch of small cubes collapsed into a single point

## Screenshots

![image](https://user-images.githubusercontent.com/51683688/59233981-9f4a0500-8b9f-11e9-9b65-2c40d9dc0985.png)

![image](https://user-images.githubusercontent.com/51683688/59233988-a53fe600-8b9f-11e9-81a1-5198f9131f50.png)

![image](https://user-images.githubusercontent.com/51683688/59233997-ab35c700-8b9f-11e9-952b-0062fa2b12c1.png)

## Your Setup (please complete the following information)

- Unity Version 2018.4.1f1
- MRTK Version v2.0.0-20190521.3 

## Target Platform (please complete the following information)

This is during the editor step, by the target will be Hololens 2

lukastoenneMS:
I have a pretty good hunch what is happening here:

The `MixedRealityToolkitFiles` utility class is used to look for resource files based on relative paths: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/bffa2f2345c1bf531c3b09f79c49bcfa53b22d08/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs#L87

This expects either a "MixedRealityToolkit.Services" or "MRTK.Service" base path to find the simulated hand poses in `<basePath>/InputSimulation/ArticulatedHandPoses/` (any parent directory will work as long as the path ends with this combination). That works alright for the git repository and the Unity package.

However, the directory structure of the nuget package is such that the pose data files end up in `Microsoft.MixedReality.Toolkit.Services.<version>/MRTK/InputSimulation/ArticulatedHandPoses` which does not match the expected pattern.

[...]

PR #4844, opened 2019-06-12
Fix MixedRealityToolkitFiles utility to handle the nuget resource path format

lukastoenneMS:
## Overview

The files utility was using a simple "endsWith" string comparison to find resource folders for MRTK. This does not work for the path format used by nuget packages, e.g.:

"MixedRealityToolkit.Services" (Unity package)
vs.
"Microsoft.MixedReality.Toolkit.Services.2.0.0-20190611.2/MRTK" (nuget package)

The fix changes the MixedRealityToolkitFiles utility class to search for both patterns and find the matching module name.

## Changes
- Fixes: #4827 

## Verification
* Create a new Unity project
* Install latest NuGet for Unity from here: [https://github.com/GlitchEnzo/NuGetForUnity/releases](https://github.com/GlitchEnzo/NuGetForUnity/releases)
* Add the source URL https://pkgs.dev.azure.com/aipmr/_packaging/MixedRealityToolkit/nuget/v2/ through _Edit > Preferences > NuGet_ for Unity. Uncheck "Read-Only Packaging Files"
* Open the _NuGet > Manage NuGet Packages_ menu
* Check _Show Prerelease_
* Search for Microsoft.MixedReality.Toolkit and install
* Setup the scene with _Mixed Reality Toolkit > Add to Scene and Configure_
* Enter playmode and hold space: Hand joint visualizers are in the same spot (bug)

* Open `Assets/packages.config` and change MRTK package versions to **2.0.0-20190613.4** (a nuget package with the fix applied)
* Delete all "Microsoft.MixedReality.Toolkit***" folders under _Assets/Packages_
* Unity will automatically download the new package versions
* Enter playmode again and hold space: Hand visualizers should be working

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_02_MixedRealityToolkitFilesTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_02_MixedRealityToolkitFilesTests.cs
@@ -1,0 +1,121 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities.Editor;
+using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+using UnityEngine.TestTools;
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.MixedReality.Toolkit.Tests.Core
+{
+    // Tests for the MixedRealityToolkitFiles utility class
+    public class TestFixture_02_MixedRealityToolkitFilesTests
+    {
+        [Test]
+        public void Test_01_FindMatchingModule()
+        {
+            TestInvalidPath("");
+
+            // Test invalid base name
+            TestInvalidPath("aaa");
+            TestInvalidPath(".SDK");
+            TestInvalidPath("aaa.SDK");
+
+            // Test missing chars
+            TestInvalidPath("MixedRealityToolki.SDK");
+            TestInvalidPath("MixedRealityToolkit.SD");
+            TestInvalidPath("ixedRealityToolkit.SDK");
+            TestInvalidPath("MixedRealityToolkit.DK");
+
+            // Test missing dots
+            TestInvalidPath("SDK");
+            TestInvalidPath("MixedRealityToolkitSDK");
+
+            // Test valid paths
+            TestValidPath("MixedRealityToolkit", MixedRealityToolkitModuleType.Core);
+            TestValidPath("MixedRealityToolkit.SDK", MixedRealityToolkitModuleType.SDK);
+
+            var modules = Enum.GetValues(typeof(MixedRealityToolkitModuleType));
+            var moduleNames = Enum.GetNames(typeof(MixedRealityToolkitModuleType));
+            for (int i = 0; i < modules.Length; ++i)
+            {
+                var module = (MixedRealityToolkitModuleType)modules.GetValue(i);
+                if (module != MixedRealityToolkitModuleType.Core)
+                {
+                    TestValidPath($"MixedRealityToolkit.{moduleNames[i]}", module);
+                }
+            }
+        }
+
+        [Test]
+        public void Test_02_FindMatchingModuleNuget()
+        {
+            // Test invalid base name
+            TestInvalidPath("aaa.1.23-45/MRTK");
+            TestInvalidPath(".1.23-45.SDK/MRTK");
+            TestInvalidPath("aaa.1.23-45.SDK/MRTK");
+
+            // Test missing chars
+            TestInvalidPath("Microsoft.MixedReality.Toolki.SDK.1.23-45/MRTK");
+            TestInvalidPath("Microsoft.MixedReality.Toolkit.SD.1.23-45/MRTK");
+            TestInvalidPath("Microsoft.MixedReality.Toolkit.SDK.1.23-45/MRT");
+
+            TestInvalidPath("icrosoft.MixedReality.Toolkit.SDK.1.23-45/MRTK");
+            TestInvalidPath("Microsoft.MixedReality.Toolkit.DK.1.23-45/MRTK");
+            TestInvalidPath("Microsoft.MixedReality.Toolkit.SDK.1.23-45/RTK");
+
+            // Test missing dots
+            TestInvalidPath("Microsoft.MixedReality.Toolkit.SDK1.23-45/MRTK");
+            TestInvalidPath("Microsoft.MixedReality.ToolkitSDK.1.23-45/MRTK");
+            TestInvalidPath("Microsoft.MixedReality.ToolkitSDK1.23-45/MRTK");
+
+            // Test missing version
+            TestInvalidPath("Microsoft.MixedReality.Toolkit.SDK/MRTK");
+            // Test missing MRTK suffix
+            TestInvalidPath("Microsoft.MixedReality.Toolkit.SDK.1.23-45");
+
+            // Test valid paths
+            TestValidPath("Microsoft.MixedReality.Toolkit.1.23-45/MRTK", MixedRealityToolkitModuleType.Core);
+            TestValidPath("Microsoft.MixedReality.Toolkit.SDK.1.23-45/MRTK", MixedRealityToolkitModuleType.SDK);
+
+            var modules = Enum.GetValues(typeof(MixedRealityToolkitModuleType));
+            var moduleNames = Enum.GetNames(typeof(MixedRealityToolkitModuleType));
+            for (int i = 0; i < modules.Length; ++i)
+            {
+                var module = (MixedRealityToolkitModuleType)modules.GetValue(i);
+                if (module != MixedRealityToolkitModuleType.Core)
+                {
+                    TestValidPath($"Microsoft.MixedReality.Toolkit.{moduleNames[i]}.1.23-45/MRTK", (MixedRealityToolkitModuleType)modules.GetValue(i));
+                }
+            }
+        }
+
+        public void TestInvalidPath(string path)
+        {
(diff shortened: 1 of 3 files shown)


MenelvagorMilsom:
CI for this is queued for this here:
https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build%2Fresults?buildId=2321&_a=summary

It looks like the queue is a bit backed up.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -246,5 +206,70 @@ public static string MapRelativeFilePath(MixedRealityToolkitModuleType module, s
...
+            {
+                var moduleName = folderMatches[0].Groups["module"].Value;
+                if (moduleNameMap.TryGetValue(moduleName, out result))
+                {
+                    Debug.Log($"Found module path for {result}: {name}");
+                    return true;
+                }
+                else
+                {
+                    Debug.LogWarning($"Found MRTK folder with unknown module extension \"{moduleName}\"");
```
wiwei: We should probably have some other build/CI time test that validates that our modules/folders adhere to this - this warning is helpful but it's a bit late (i.e. imagine we add a new folder without updating the Dictionary above, we wouldn't find out until way after the build has completed (in fact, things would probably fail in ways that aren't obvious until we open the console and see this warning, and then go back from that).

This is a larger ask than what can happen in this PR and we should absolutely keep this, just wanted to raise that there's a fragility here that can and should be addressed earlier in the cycle.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -246,5 +206,70 @@ public static string MapRelativeFilePath(MixedRealityToolkitModuleType module, s
             }
             return null;
         }
+
+        private static readonly Dictionary<string, MixedRealityToolkitModuleType> moduleNameMap = new Dictionary<string, MixedRealityToolkitModuleType>()
+        {
+            { "", MixedRealityToolkitModuleType.Core },
```
wiwei: Is it possible to have a unit test that validates that all of the MixedRealityToolkitModuleType values are covered by this dictionary? Otherwise it would be easy to add a new type and forget to add a value here.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -246,5 +206,70 @@ public static string MapRelativeFilePath(MixedRealityToolkitModuleType module, s
...
+            { "Generated", MixedRealityToolkitModuleType.Generated },
+            { "Providers", MixedRealityToolkitModuleType.Providers },
+            { "Services", MixedRealityToolkitModuleType.Services },
+            { "SDK", MixedRealityToolkitModuleType.SDK },
+            { "Examples", MixedRealityToolkitModuleType.Examples },
+            { "Tests", MixedRealityToolkitModuleType.Tests }
+        };
+
+        private static bool FindMatchingModule(string folder, out MixedRealityToolkitModuleType result)
```
wiwei: Can we have a few basic unit tests for this function? i.e. something like:

test 1) pass in a unity package version that's at the root.
test 2) pass in a unity package version that's in the Packages/ folder
test 3) pass in a unity package version that's in some other arbitrary location
test 4) pass in the nuget version
test 5) totally unrelated path that shouldn't find a match.

Really this would also help to canon-ize (is that a word?) the types of paths we expect and show how this handles them

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
wiwei: Maybe my other comment about dictionary completeness will actually handle this, soooo maybe nevermind :)

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -144,25 +112,17 @@ static MixedRealityToolkitFiles()
...
-                if (directories.Count() == 0)
-                {
-                    directories = Directory.GetDirectories(rootPath, AlternateMixedRealityToolkitDirectory(module), SearchOption.AllDirectories);
-                }
-
-                directories = directories.Select(NormalizeSeparators);
-
-                foreach (string s in directories)
+                string normalizedFolder = NormalizeSeparators(folder);
+                if (FindMatchingModule(normalizedFolder, out MixedRealityToolkitModuleType module))
```
wiwei: I *think* that the code in this line up to 125 is shared with the 47-55 - could be worth sharing them.

wiwei:
Overall seems fine, mostly looking for some unit tests (for regex stuff because I suck at reading regex) and also for dictionary completeness

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
lukastoenneMS: Correct, i added register/unregister functions for the common code here.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
lukastoenneMS: Writing exhaustive tests for this utility class is a bit tricky, at least in its current state. It's a static class that searches all the folders at construction time, i.e. can't really test it on a clean slate with no existing folders. Can still test adding/removing folders in the existing project, although sandboxing changes in the file system for tests may require some work.

I'll try to get something started at least.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
lukastoenneMS: I've added loops to the unit test which check that each of the module enum values is found.

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_02_MixedRealityToolkitFilesTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            Assert.AreEqual(module, expectedModule);
+        }
+
+        [TearDown]
+        public void CleanupTests()
+        {
+            TestUtilities.TearDownScenes();
+        }
+
+        private void WaitUntilFoldersAvailable()
```
wiwei: I think this may be dead right? I assume this was from maybe some earlier iteration that tried to do real file IO stuff

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_02_MixedRealityToolkitFilesTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            TestInvalidPath("MixedRealityToolkit.SD");
+            TestInvalidPath("ixedRealityToolkit.SDK");
+            TestInvalidPath("MixedRealityToolkit.DK");
+
+            // Test missing dots
+            TestInvalidPath("SDK");
+            TestInvalidPath("MixedRealityToolkitSDK");
+
+            // Test valid paths
+            TestValidPath("MixedRealityToolkit", MixedRealityToolkitModuleType.Core);
```
wiwei: In the real code, is it always getting relative paths, or is it getting absolute paths i.e.:

c:\path\to\project\and\mrtk
/path/to/project/and/mrtk

If it's relative then this is good, if there are absolute paths involved it would be good to get those cases in as well, as those are more like the "real" stuff

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_02_MixedRealityToolkitFilesTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            TestValidPath("MixedRealityToolkit.SDK", MixedRealityToolkitModuleType.SDK);
+
+            var modules = Enum.GetValues(typeof(MixedRealityToolkitModuleType));
```
wiwei: Nit: Add a comment here saying that the intent is to ensure that every single enum value is properly handled by the Dictionary in MRTKFiles - in other words, if this fails, the likely cause is that you haven't updated that dictionary. This provides direct action when someone adds an enum later and tells the why the thing below is broken and how to fix it.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -144,25 +93,37 @@ static MixedRealityToolkitFiles()
...
-            foreach (MixedRealityToolkitModuleType module in Enum.GetValues(typeof(MixedRealityToolkitModuleType)))
+            foreach (string folder in Directory.EnumerateDirectories(rootPath))
             {
-                IEnumerable<string> directories = Directory.GetDirectories(rootPath, MixedRealityToolkitDirectory(module), SearchOption.AllDirectories);
+                RegisterModuleFolder(folder);
+            }
+        }
 
-                if (directories.Count() == 0)
+        private static void RegisterModuleFolder(string folder)
```
andreiborodin: Consider renaming as TryRegisterModuleFolder (and similiarly as TryUnregisterModuleFolder)

lukastoenneMS: Done.

PR #4848, opened 2019-06-12
Pull in the remaining RC2 stabilization changes into mrtk_development

wiwei:


Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs
```diff
@@ -191,8 +191,22 @@ protected MouseState GetMousePointerEventDataForMrtkPointer(PointerData pointerD
...
+                RaycastCamera.transform.rotation = Quaternion.LookRotation(pointer.Rays[0].Direction);
+            }
+            else
+            {
+                // The pointer.Controller does not provide rotation, for example on HoloLens 1 hands.
+                // In this case pointer.Rays[0].Origin will be the head position, but we want the 
+                // hand to do drag operations, not the head.
+                // pointer.Position gives the position of the hand, use that to compute drag deltas.
+                RaycastCamera.transform.position = pointer.Position;
+                RaycastCamera.transform.rotation = Quaternion.LookRotation(pointer.Rays[0].Direction);
```
SimonDarksideJ: This line seems duplicated, surely if check is only on the position and not both properties?  Seems overly onerous.  Unless the rotation is ALSO support to be calculated differently.

wiwei: @julenka I missed this when looking at https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4765

We can patch this up in mrtk_development - this PR is just about ensuring consistency between our release and dev branch

julenka: I thought about moving the rotation out of the else case but decided to keep it in the else case because keeping position and rotation together I think makes it easier to read. It's just a preference though and if you guys feel strongly you can move it out.

Issue #4849, opened 2019-06-12
EyeTrackingTarget does not properly override methods from InputSystemGlobalListener


Alexees:
## Describe the bug

The class EyeTrackingTarget does not override Start and OnDisable properly (Throwing warnings when compiled). The correct implementation of this would mean that the base class InputSystemGlobalListener changes its Start methods return type to Task as it's async resulting in futher code changes in other classes.
OnDisable can be corrected just fine.

wiwei:
Thanks for filing this! I think you pulled mrtk_development within the hour or two where this existed - this should be fixed by:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4845

wiwei:
Ah, wait, I missed the async part. Lemme fix that.

Alexees:
@wiwei just wanted to inform you, that now I get a compilation warning:

> Assets\MixedRealityToolkit\MixedRealityToolkit.SDK\Features\Input\Handlers\EyeTrackingTarget.cs(145,39): warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.

What will happen if you override InputSystemGlobalListener and you depend on the InputSystem? I would say code fails then. You'd need to await the base call, but then it needs to return Task, which does change it from being a Unity engine callback.
What about sealing the base Start and do a Task.Run on a seperate method that can be overridden properly?

PR #4850, opened 2019-06-12
Make EyeTrackingTarget's Start() function match the definition of its base class

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4849

I had missed this in https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4845


wiwei:
Actually, this was the wrong thing - this has its own warning but I don't know why I didn't see it in the console until now. I'm gonna revert this.

I don't feel comfortable changing the base definition of InputSystemGlobalListener at this point, so I think the best I can do is go with a sync Start()

Issue #4406, opened 2019-05-16
Getting started docs don't mention recommended build settings.


Samvid95:
I am trying to build a simple scene for Hololens v1 using MRTK v2. After solving a few problems I am facing the deployment issue on the side of Visual Studio. I am getting the error saying, "DEP3321: To deploy this application, your deployment target should be running Windows Universal Runtime version 10.0.18362.0 or higher. You currently are running version 10.0.17763.502. Please update your OS, or change your deployment target to a device with the appropriate version."

I am not familiar on how to update the OS on the Hololens to make it to the version required by the MRTKv2.

wiwei:
It's possible that the minimum platform version you have in your build settings within Unity were set to 18362 instead of 17763:

![image](https://user-images.githubusercontent.com/5840182/57871047-562fac80-77bd-11e9-859f-bc86c45911dc.png)

Usually we'll have the lowest possible selected (so that it's possible to deploy the app to almost anything)

Looking through our getting started guide, we don't have any mention of this recommendation, so I'll keep this issue open to track adding that note.

Samvid95:
Hahaha. Yess it was the minimum platform version that was confusing me. Thank you.

PR #4851, opened 2019-06-12
Update the getting started docs to include UWP build setting recommendations.

wiwei:
This is such a common question that I wanted to add some comments in the getting started doc about it. People are trying to configure their target/platform in some cases to the highest number, but in reality the defaults are actually pretty good. The only key part is the requirement to check that 18362 exists in the list, and if not, provides the solution for how to fix it.

Also changes the requirements section up a little to list the requirement link up front, and THEN providing clarification. The reality is that if you're targetting cross-platform with MRTK (which you should be) we shouldn't tell people that "oh actually it's not REALLY needed" and have them run into weird stuff down the line - let's get people to have the same prereqs that we have (which is just the latest set of SDKs) and lead them down the path of success.

- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4406


Documentation/GettingStartedWithTheMRTK.md
```diff
@@ -129,6 +132,19 @@ If you need to move the players start point, then **move the scene content and N
 
 You are now ready to start building your project!
 
+## Recommended build settings (UWP)
+
+When building for WMR, HoloLens 1, or HoloLens 2, it is recommended that your build settings "Target SDK Version"
+and "Minimum Platform Version" look like they do in the picture below:
+
+![](../Documentation/Images/getting_started/BuildWindow.png)
```
MenelvagorMilsom: I can't see this image [here](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/82eb6cb6754dd8f027ba5116ccc15afde52f99f3/Documentation/GettingStartedWithTheMRTK.md#recommended-build-settings-uwp). Is that definitely the right path?

wiwei: Ahhhhhhhh case sensitive paths.

PR #4858, opened 2019-06-12
Consistency in assembly and asmdef names

davidkline-ms:
The change updates all assembly names and asmdef file names to match the following format:

Microsoft.MixedReality.Toolkit.<name[.subname]>

Due to file name and path lengths, the examples\demos assemblies were renamed to drop "Examples" from their names.

Testing performed
- Run solver demo scene in the editor, ensure
    - Object selection
    - Teleport
    - Boundary
    - Diagnostics
- Build with .net backend and compile in Visual Studio 2017
- Editor tests pass


wiwei:
This I think is a pretty breaking change for folks who are going to upgrade from any of the existing RC builds.

Before this goes through, we need a story for how people can take this without significant pain (i.e. we write a migration script that will highlight changes required?)

davidkline-ms:
> Before this goes through, we need a story for how people can take this without significant pain (i.e. we write a migration script that will highlight changes required?)

If the customer updates over an existing build, there _may_ be duplicate asmdef files. The old versions will need to be deleted. I will write a mapping document (batch file?) to assist. Deleting the library and build folders would also likely be a good step (again will add to the document).

Without this change, there is currently a mix of different assembly names (Microsoft.MixedReality.Toolkit.<name> and MixedRealityToolkit.<name>) that lack consistency and branding.

Would the above referenced doc/script (as well as a confirmation test pass) help with concerns?

wiwei:
> > Before this goes through, we need a story for how people can take this without significant pain (i.e. we write a migration script that will highlight changes required?)
> 
> If the customer updates over an existing build, there _may_ be duplicate asmdef files. The old versions will need to be deleted. I will write a mapping document (batch file?) to assist. Deleting the library and build folders would also likely be a good step (again will add to the document).
> 
> Without this change, there is currently a mix of different assembly names (Microsoft.MixedReality.Toolkit. and MixedRealityToolkit.) that lack consistency and branding.
> 
> Would the above referenced doc/script (as well as a confirmation test pass) help with concerns?

I understand the lack of consistency, and it sucks. 

The issue that people have already gotten onto RC1/Refresh/RC2, and doing this would break them in ways that are frustrating.

They also get little value out of it, aside from no longer getting confused about why it's inconsistent.

My point is that, breaking changes **suck** and there needs to be a value/cost tradeoff. If the value isn't there, then we need SOME story to make the cost less (i.e. migration tool)

If we aren't willing to invest in making the entire thing good, we shouldn't invest in it at all.

davidkline-ms:
> My point is that, breaking changes suck and there needs to be a value/cost tradeoff. If the value isn't there, then we need SOME story to make the cost less (i.e. migration tool)

Fully agree.

Some of the value is in customer trust that the assemblies are from the same product and that we are doing our diligence that even the little details are done right :)

I'll put a migration story together and double check the validation.

SimonDarksideJ:
I would recommend NOT changing the namespaces in the files and ONLY move the folder structure around.  Either is a breaking change but moving folders is a simple solution.
The rule is only a guide for class discovery but it need not be hard rule.

Changing folders only requires removing and reimporting. no code change.
Changing namespaces requires changing all client code that relates to that namespace and there is no easy way to update an entire project.

keveleigh:
Are there any namespace changes in this PR? I didn't see any, but it's possible I missed one.
If we do ever need to change namespaces, we may be able to use Unity's `MovedFrom` attribute to hook into the API updater. I haven't tried it out yet, but it might be worth looking into if needed.

MicrosoftTeams-image (5)

davidkline-ms:
class namespaces. no, only asmdef and assembly names were changed. all else was untouched

davidkline-ms:
Merged development branch, add file documenting steps to update from rc2 to ga (currently, just the asmdef rename tables)

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/Microsoft.MixedReality.Toolkit.Providers.WindowsVoiceInput.asmdef
```diff
@@ -12,7 +12,7 @@
         "WindowsStandalone64"
     ],
     "excludePlatforms": [],
-    "allowUnsafeCode": false,
+    "allowUnsafeCode": true,
```
wiwei: Was this change to allow unsafe code intentional (and if so, what was the reason?) The rest of our asmdefs today don't have this, so I'm curious what motivated this change.

Assets/MixedRealityToolkit/StandardAssets/FontsSDFTextures/segoeui SDF.asset
```diff
@@ -15,7 +15,1820 @@ MonoBehaviour:
   hashCode: -1937467786
   material: {fileID: 21340371490990018}
   materialHashCode: 1402074941
-  fontAssetType: 1
+  m_Version: 1.1.0
```
wiwei: These asset files seem to have a bunch of changes that don't seem to be that related to these asmdef renames - might be worth seeing if they're needed, and if so, putting them in separately to keep each PR scoped.

davidkline-ms: fixed by restoring from the dev branch

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/Microsoft.MixedReality.Toolkit.Providers.WindowsVoiceInput.asmdef
davidkline-ms: accidental. should not have been changed

davidkline-ms: fixed

Issue #4250, opened 2019-05-09
No way to identify loaded holographic device


provencher:
## Describe the bug

HoloLens 1 and HoloLens 2 are both considered windows Holographic, and there is no way to differentiate them within a C# codebase, beyond manually testing if certain capabilities are available. 

Beyond that, testing capabilities can fail if permissions are not accepted, and so it is not a reliable way to differentiate devices.

## Expected behavior

A clear enum or string identifying the hardware loaded.
An easy to access capabilities profile identifying what the hardware is capable of doing, irrespective of the accepted permissions.

## Your Setup (please complete the following information)

- Unity Version 2018.3.8f1]
- MRTK Version v2.0 latest dev branch

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive

provencher:
It seems that you can at least get the OS architecture information from interop services
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.osarchitecture?view=netstandard-1.3

I created an active device checker like this. Its not perfect for distinguishing WMR from HoloLens however

     public static MixedRealityDevice CurrentlyActiveDevice
        {
            get
            {
#if UNITY_EDITOR
                return MixedRealityDevice.HoloLens2;
#elif UNITY_WSA
                switch (RuntimeInformation.OSArchitecture)
                {
                    case Architecture.Arm:
                    case Architecture.Arm64:
                        return MixedRealityDevice.HoloLens2;
                    case Architecture.X86:
                        return MixedRealityDevice.HoloLens;
                    default:
                    case Architecture.X64:
                        return MixedRealityDevice.WMR;
                }
#endif
            }
        }

provencher:
Ok this doesn't work

`Unsupported internal call for IL2CPP:"Is64BitOs" - L"It is not possible to check if the OS is a 64bit OS on the current platform."`

chrisfromwork:
Do you generate different appx's for hololens 1 vs hololens 2? It may not be an ideal long term solution, but you could create a custom preprocessor directive that you change in your player settings based on what you are building: https://docs.unity3d.com/Manual/PlatformDependentCompilation.html

[...]

PR #4882, opened 2019-06-14
Add checkcapability api, add demo scene

davidkline-ms:
This change introduces a new API to check if the platform on which an app is running supports specific capabilities:

- Articulated hands
- GGV hands
- Motion controllers
- Eye tracking
- Voice command (app defined keywords)
- Voice dictation (speech to text)
- Spatial awareness meshes
- Spatial awareness planes (future proofing, no provider currently exists)
- Spatial awareness points (future proofing, no provider currently exists)

The change adds the MixedRealityCapability enum and IMixedRealityCapabilityCheck interface, which defines the CheckCapability() method. 

The input system, spatial awareness system and relevant data providers now implement the IMixedRealityCapabilityCheck interface and report the status of the capabilities using a calling pattern similar to the following:

``` 
            IMixedRealityCapabilityCheck capabilityChecker = InputSystem as IMixedRealityCapabilityCheck;
            if (capabilityChecker != null)
            {
                bool isSupported = capabilityChecker.CheckCapability(MixedRealityCapability.ArticulatedHand);
            }            
```

If an application wishes to know WHICH manager is providing the support, it can request the data providers from the system and call CheckCapability on each until finding the match.

Also added is a demo scene illustrating how to use the new interface and display the status of each capability on the running platform. The following image was captured running in the editor with the build type set to UWP and all default systems enabled.

image

Testing performed:
- Editor play mode
- Compile x64 and run on Windows Mixed Reality Immersive headset
- Compile x86 and run on HoloLens
- Build using IL2CPP and .net backends
- Run edit mode tests

Fixes https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4250


Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputCapabilities.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    /// <summary>
+    /// Platform input capabilities.
+    /// </summary>
+    public enum MixedRealityInputCapabilities
+    {
+        ArticulatedHand = 0,
+        GGVHand,
+        MotionController,
+        EyeTracking
```
provencher: It would be nice to add speech capabilities, as well as spatial awareness.

provencher:
Very nice @davidkline-ms!

This is much needed and appreciated.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -14,6 +15,10 @@
 using WsaGestureSettings = UnityEngine.XR.WSA.Input.GestureSettings;
 #endif // UNITY_WSA
 
+#if WINDOWS_UWP
+using WindowsInputSpatial = global::Windows.UI.Input.Spatial;
+#endif // WINDOWS_WSA
```
wiwei:
```suggestion
#endif // WINDOWS_UWP
```

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -37,15 +42,51 @@ public class WindowsMixedRealityDeviceManager : BaseInputDeviceManager
...
+                        return WindowsInputSpatial.SpatialInteractionManager.IsSourceKindSupported(WindowsInputSpatial.SpatialInteractionSourceKind.Hand);
+
+                    case MixedRealityInputCapabilities.MotionController:
+                        return WindowsInputSpatial.SpatialInteractionManager.IsSourceKindSupported(WindowsInputSpatial.SpatialInteractionSourceKind.Controller);
+                }
+#endif // WINDOWS_UWP
+            }
+            else
+            {
+                // Pre-Windows 10 1903.
```
wiwei: Nit: Maybe move this to the same line as the else, if only because I'm not sure if this statement modifies the entire else block, or just the if statement right below. If this were at the else location (or if there were even an extra line below this) I think it would less ambiguous

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -37,15 +42,51 @@ public class WindowsMixedRealityDeviceManager : BaseInputDeviceManager
...
+                    return (capability == MixedRealityInputCapabilities.GGVHand);
+                }
+                else
+                {
+                    // Windows Mixed Reality Immersive devices support motion controllers
+                    return (capability == MixedRealityInputCapabilities.MotionController);
+                }
+            }
+
+                return false;
```
wiwei: Nit: Fix indent.

I would do a suggested change but honestly the suggestion think makes me delete arbitrary spaces and I can't count past 2 so here I am leaving this long comment where I could have probably actually made the suggested change by now.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -37,15 +42,51 @@ public class WindowsMixedRealityDeviceManager : BaseInputDeviceManager
...
 #if UNITY_WSA
 
-        /// <summary>
-        /// The initial size of interactionmanagerStates.
-        /// </summary>
-        /// <remarks>
-        /// This value is arbitrary but chosen to be a number larger than the typical expected number (to avoid
-        /// having to do further allocations).
-        /// </remarks>
+            /// <summary>
```
wiwei: I think extra tabs were added here by mistake (this should have the same indent as MaxInteractionSourceStates)

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityEyeGazeDataProvider.cs
```diff
@@ -51,14 +51,27 @@ public class WindowsMixedRealityEyeGazeDataProvider : BaseInputDeviceManager, IM
...
+#if WINDOWS_UWP
+
+        private static bool askedForETAccessAlready = false; // To make sure that this is only triggered once.
+
+        /// <inheritdoc/ >
+        public override bool CheckCapability(MixedRealityInputCapabilities capability)
+        {
+            if (WindowsApiChecker.UniversalApiContractV8_IsAvailable)
+            {
+                return (EyesPose.IsSupported() && (capability == MixedRealityInputCapabilities.EyeTracking));
```
wiwei: Nit: Consider doing the cheap check first (i.e. capability == EyeTracking) followed by the possibly not cheap check.

Ideally you'd want to short circuit in the order of having faster things check first.

In reality I doubt this has any real impact here, but it's generally a good practice/habit to get into.

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
```diff
@@ -53,6 +54,25 @@ public class InputSimulationService : BaseInputDeviceManager, IInputSimulationSe
...
+        public override bool CheckCapability(MixedRealityInputCapabilities capability)
+        {
+            switch (capability)
+            {
+                case MixedRealityInputCapabilities.ArticulatedHand:
+                    return (InputSimulationProfile.HandSimulationMode == HandSimulationMode.Articulated);
+
+                case MixedRealityInputCapabilities.GGVHand:
+                    // If any hand simulation is enabled, GGV interactions are supported.
+                    return (InputSimulationProfile.HandSimulationMode != HandSimulationMode.Disabled);
```
wiwei:
```suggestion
                    return (InputSimulationProfile.HandSimulationMode == HandSimulationMode.Gestures);
```

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -6,6 +6,7 @@
 using System.Collections.Generic;
 using UnityEngine;
 using UnityEngine.EventSystems;
+using UnityEngine.Jobs;
```
wiwei: How is this used down below? It doesn't look like there's any usage of Unity job stuff but then again I guess I don't have a photographic memory of everything in the Unity namespace.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputCapabilities.cs
wiwei: FWIW I'm also fine with going with this change first to establish the framework, and then doing the speech in its own PR (smaller PRs are easier to review, from a selfish perspective)

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputCapabilities.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Platform input capabilities.
+    /// </summary>
+    public enum MixedRealityInputCapabilities
```
wiwei: This is gonna be a pain (sorry) - can you rename this to MixedRealityInputCapability?

The plural here is confusing because it seems to imply that this enum can have more than one, but it's an identifier for a single capability.

Also see MS naming conventions:

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations

It's the first bullet point listed there.

wiwei: FWIW, this is also the case in the rest of our code (aside from a few exceptions which I wish we could change but wouldn't be worth breaking people for)

wiwei: Also, consider naming it not input specific, based on @provencher's request below for having spatial awareness (i.e. they are device/system capabilities, not just input potentially)

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
```diff
@@ -62,6 +62,14 @@ public interface IMixedRealityInputSystem : IMixedRealityEventSystem
         /// </summary>
         bool IsInputEnabled { get; }
 
+        /// <summary>
```
wiwei: If you want to do this in a way that isn't breaking, consider spawning a different interface:

IMixedRealityCapabilityChecker

Which would be implementable by other systems like spatial awareness (and I guess then you'd probably need a static helper/other thing that would federate calling out to the other systems since this might be beyond just input)

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputCapabilities.cs
wiwei: Looking at the other naming, since we should support those other things, we probably need to rename the capability enum to be a little more generic (i.e. not just input). Or I guess we could have two enums but that would be confusing.

wiwei:
Overall looks good! Had some comments about being able to extend this to handling other capabilities which might require some changes, but the general structure is sane to me.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
provencher: Having the jobs namespace adds a dependency on the experimental package as well. If it's necessary, that's fine, but if it isn't, it's not ideal go force users into using it.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputCapabilities.cs
provencher: Yeah this should be a general capabilities Api, mot just related to input.

provencher: There may well be some unestablished features that have yet to be released that may not work on hl1. It's good to be prepared

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
davidkline-ms: > and I can't count past 2 so 

Hahaha!

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
davidkline-ms: i didnt notice adding that.... might have been auto-added by Unity 2018.4.2

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityInputSystem.cs
davidkline-ms: >  consider spawning a different interface

That would be consistent with the IMixedRealityDataProviderAccess interface. I'll make the change

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
davidkline-ms: removed. was unused and was blocking .net backend builds

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputCapabilities.cs
davidkline-ms: > This is gonna be a pain (sorry) 

Not a pain at all. :)

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputCapabilities.cs
davidkline-ms: I'll add enums for the following (and implement the checks in a future PR)

- VoiceCommand (aka keywords)
- VoiceDictation
- SpatialAwareness (question: should there be three of these? mesh, plane, point)

davidkline-ms:
Note to reviewers: The work to add spatial and voice capability support was minimal. They have both been added to this PR.

davidkline-ms:
Additional note: this change was reworked to no longer be a breaking change

provencher:
Looks great! Nice job @davidkline-ms

Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
```diff
@@ -57,6 +57,16 @@ public class WindowsSpeechInputProvider : BaseInputDeviceManager, IMixedRealityS
 #endif
             false;
 
+        #region IMixedRealityCapabilityCheck Implementation
+
+        /// <inheritdoc/ >
+        public bool CheckCapability(MixedRealityCapability capability)
+        {
+            return (capability == MixedRealityCapability.VoiceCommand);
```
keveleigh: Do we want to check something with the platform, like microphone support?

jbienzms:
@davidkline-ms @keveleigh 

Sorry I didn't get to reviewing this until after it was merged, but I wanted to know why the explicit cast is required in the code snippet.

`IMixedRealityCapabilityCheck capabilityChecker = InputSystem as IMixedRealityCapabilityCheck;`

Every time we use [explicit interface implementation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation) we effectively hide those interface methods from the class definition. This requires the class to be cast to the interface before the method can be called which really breaks discoverability.

Looking at [WindowsMixedRealityDeviceManager](Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs) it appears we did _not_ use explicit interface implementation. Which is good for discoverability. But why then are we doing an explicit cast in the sample demo code?

davidkline-ms:
> Every time we use explicit interface implementation we effectively hide those interface methods from the class definition.

There are two reasons we have this approach:

1. Not all classes require all interface methods and should not be forced to implement stub implementations.
2. Every time we add to an interface, it is a breaking change (all implementations must implement all methods). With MRTK v2 GA, we will enter a 'no breaking changes' mode. This pattern allows for easy extension of functionality as needed.

> Looking at WindowsMixedRealityDeviceManager it appears we did not use explicit interface implementation.

If this is the case, it is a bug. I will address that ASAP.

davidkline-ms:
I  just confirmed that the WindowsMixedRealityDeviceManager does implement IMixedRealityCapabilityCheck

```
public class WindowsMixedRealityDeviceManager : BaseInputDeviceManager, IMixedRealityCapabilityCheck
```

jbienzms:
Yes, [WindowsMixedRealityDeviceManager](Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs) does implement **IMixedRealityCapabilityCheck** but it does so _implicitly_ rather than explicitly. 

```
#region IMixedRealityCapabilityCheck Implementation 
**public bool** CheckCapability(MixedRealityCapability capability) 
{ 
	...
}
```

rather than

```
#region IMixedRealityCapabilityCheck Implementation 
**bool IMixedRealityCapabilityCheck.**CheckCapability(MixedRealityCapability capability) 
{ 
	...
}
```


To be clear, it is my opinion that the way it's currently implemented is the preferred way. The way it's currently implemented, developers will be able to call the **CheckCapability** method on **WindowsMixedRealityDeviceManager** without having to cast it to an interface. This is discoverable while having to cast it to the interface is not discoverable.

> Every time we add to an interface, it is a breaking change (all implementations must implement all methods)

Let's look at this purely within the context of **WindowsMixedRealityDeviceManager**. It either implements the interface or it doesn't. _Implicit_ implementation means the methods are public and discoverable to developers using **WindowsMixedRealityDeviceManager**. _Explicit_ implementation means that developers will not know **WindowsMixedRealityDeviceManager** has those capabilities unless we provide documentation stating that it does (and the developer happens to search for and find said documentation).

Implicit vs explicit does not truly help avoid breaking changes. This is because even if the implementation is _explicit_, changing the interface (or deleting it all together) is still a breaking change to **WindowsMixedRealityDeviceManager** and those who use it. Even if the interface is explicit, the class will still have to be updated. And even though the methods aren't exposed directly on the **WindowsMixedRealityDeviceManager** class definition, any developer using those capabilities through a cast will still be broken when it changes. The only real difference is discoverability.

Also, explicit implementation wasn't really designed to hide public methods (though it can have the side effect of doing so). The primary design goal for explicit implementation is to allow the implementor to selectively change behavior when the class is called via a specific interface. This _may_ be important if the class implements more than one interface that has the same method signature.

From the [Explicit Implementation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation) docs:

> If a class implements two interfaces that contain a member with the same signature, then implementing that member on the class will cause both interfaces to use that member as their implementation. If the two interface members do not perform the same function, however, this can lead to an incorrect implementation of one or both of the interfaces. It is possible to implement an interface member explicitly—creating a class member that is only called through the interface, and is specific to that interface.

Since explicit implementation doesn't truly help avoid breaking changes, and since it makes these methods undiscoverable, I feel strongly that these interfaces should be implemented implicitly. It will greatly improve discoverability and also greatly reduce the teams need to document features that are only available through casting.

jbienzms:
To put it another way, if a developer doesn't use:

WindowsMixedRealityDeviceManager.CheckCapability

keveleigh:
The reason the cast is happening is because we're working based on the interface type, not the concrete `MixedRealityInputSystem` type, whether it's implemented explicitly or not.

image

Issue #4886, opened 2019-06-14
Font size of ToolTip does not change


deibich:
## Describe the bug

Font size of a ToolTip does not change when _Font Size_ in inspector is changed. _fontSize_ can't be edited from other scripts.

## To reproduce

1. Open Scene **TooltipExamples** / Drop a ToolTip in any scene
2. Select any ToolTip in scene
3. Change _Font Size_ in inspector
4. **ToolTip does not change**
5. Change _Tool Tip Text_ in inspector
6. Content **and** size of font changes

## Expected behavior

Changes on _Font Size_ should change the size immediately. _fontSize_ should be accessible from scripts.

## Screenshots

![tooltip_font_not_change](https://user-images.githubusercontent.com/17428923/59494495-8b4b1f80-8e8d-11e9-985f-d0076dfa0b77.gif)

## Your Setup (please complete the following information)

- Unity Version 2018.4.1f1, 2019.1.6f1
- MRTK Versionv 2.0.0 RC2.1
## Target Platform (please complete the following information)

- HoloLens
- Unity Editor

wiwei:
Thank you for filing this issue and fixing it!

I wanted to call out that this not only addresses the issue of it not updating live in the editor, but also exposes it programatically as well which is awesome.

PR #4887, opened 2019-06-14
Change font size of ToolTip

deibich:
## Overview
Font size of ToolTip changes ToolTip now without editing the text first. FontSize is available in scripts now. 

## Changes
- Fixes: #4886

tooltip_font_does_change


wiwei:
This looks great! Thanks for contributing :)

PR #4889, opened 2019-06-14
Jganser ci runner patch multicore appx build

jganser:
## Overview

commandline appx build with the functionality of the mrtk utillities
multicore option for appx build

## Changes
- Fixes: 
     issues with appx build in batchmode
- Adds Feature:
     Multi core appx build through msbuild flags /m /nr:false

## BuildTool apporach
```[c#]
public static async void BuildAndBundle()
{
           SetUpHololensBuildSettings(); // setup your build preferences in case they differ from developer machine to CI runners intention
           
            // First build SLN            
            UnityPlayerBuildTools.StartCommandLineBuild(true);
            var buildInfo = new UwpBuildInfo(true);
            buildInfo.Configuration = "Master"; // or what ever you desire
            buildInfo.BuildPlatform = EditorUserBuildSettings.wsaArchitecture; // still targeting Hololens 1
            buildInfo.Multicore = true;
            

            // Next build the APPX from Solution
            //EditorAssemblyReloadManager.LockReloadAssemblies = true; //was already set in StartCommandLineBuild
            UwpAppxBuildTools.BuildAppx(buildInfo);
            EditorAssemblyReloadManager.LockReloadAssemblies = false;
}
```


jganser:
Linked to [issue](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4890)

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -28,8 +28,9 @@ public static class UwpAppxBuildTools
         /// </summary>
         /// <param name="buildInfo"></param>
         /// <param name="cancellationToken"></param>
+        /// <param name="buildOnMultipleCores"></param>
         /// <returns>True, if the appx build was successful.</returns>
-        public static async Task<bool> BuildAppxAsync(UwpBuildInfo buildInfo, CancellationToken cancellationToken = default)
+        public static async Task<bool> BuildAppxAsync(UwpBuildInfo buildInfo, CancellationToken cancellationToken = default, bool buildOnMultipleCores = false)
```
MenelvagorMilsom: Is there any reason why we shouldn't enable this for the UWP build in UwpPlayerBuildTools.BuildPlayer?

wiwei:
Thanks for contributing! I'll take a look at this, though advance warning I'm not going to be able to really dig into this until Tuesday. I want to do a little bit of investigation around this + our pipelines to see how things run.

In the meantime if anyone else wants to jump in that's fine, just wanted to give a lower bound on the time to response on my part.

luis-valverde-ms:
Status update: 
- Waiting on @wiwei review.
- Wating on @jganser response to @MenelvagorMilsom question.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
jganser: As far as I can tell there is no reason not to enable it.

I tried to designe it with a minimal amount of possible conflicts in mind.


OR 
If you mean rather you would like to always use multicore builds, I'd say that it might be that in some cases, people would like to build in background on only one core.


To be honest I'm not sure what your question really is about, maybe you could elaborate it a little more. 😅

MenelvagorMilsom: Hey, sorry. I guess it would be nice if we can enable this from the command line build (see `UwpPlayerBuildTools.ParseBuildCommandLine`) and also change our yaml templates so that we provide this flag during ci builds (see `pipelines\`). We can try it out during a PR build then.

MenelvagorMilsom: If you need a hand with the yaml files in `pipelines\` I can give it a look instead.

jganser: I made a new commit, with a merge from current dev.
There I changed the multicore switch to a commandline argument and a corresponding UWPBuildInfo attribute.
It should work as UnityArgs, with the pipeline I guess.

wiwei:
First off, thanks for sending this out - sorry it's take a while for us to get back to this. There are a few overall questions that I think would be worth addressing first, at which point re-opening a new PR (i.e. without the other files like these included:

Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/BoomBox/glTF/BoomBox_Mat.mat
Assets/MixedRealityToolkit.Examples/Demos/Gltf/Models/Lantern/glTF/LanternPost_Mat.mat
Packages/manifest.json
ProjectSettings/ProjectVersion.txt

Would probably be good)

1) I think it would be good to understand why the sync version of these things is needed, vs just calling the async function with an await. I see a mention that this doesn't play well in non-batch mode, but it feels like this sounds like some issue in in the process launching code itself - it feels like the right thing to do would be to dig into that (or, if that was already investigated, we can post that information in the issue about what it is, and maybe we can fix it or the consciously make the workaround.
2) If we need to do the sync version, it might be good to still try to share as much code as possible - I think right now the PR still does some copy paste for some parts (i.e. the sync and async versions have shared code that could be made common)
3) Since the function StartCommandLineBuild is designed to be called from the command line (and thus, no parameters), if we want to kick off a build outside of the command line (i.e. through the snippet above) it might make more sense to have another function that's called "StartBuild" or something that is designed to be invoked not from the command line (where we could then maybe make both of those functions call the same thing. I guess the main comment here is, the function is called StartCommandLineBuild, so it should probably be something that is called from the command line - we could rename it as well (but this would need to be staged with other build changes, and probably might break others who are dependent on this.

If it would help, I'm also down to hop on a call to discuss this to iron these out and then get a new PR opened.

PR #4894, opened 2019-06-14
Update the UWP build tools to support adding the Gaze Input capability automatically

wiwei:
This change is intended to help make the eye tracking work a little easier to consume. Eye tracking requires the gazeInput capability setup on the Visual Studio project, but Unity currently doesn't support having this functionality in Unity itself. As a result, when you build the Unity project, you have to modify your AppX manifest and add the gazeInput capability. This is a huge pain for people who want to use eye tracking.

This change gets around this by adding an optional build flag that, when set, will add the capability to the manifest, and this will happen either at:

1) After the Unity build.
2) Before running the appx build.

The function itself is idempotent anyway so it's safe to invoke in both cases. This will also handle cases where the appx solution was built a while back (or was done with the unity tooling instead of our tooling).

image

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4245


wiwei:
Also, tests, yay! So good!

sostel:
Yay, thanks for adding this, @wiwei! A few comments/questions:

> every single time you rebuild the Unity project, you have to modify your AppX manifest and re-add the gazeInput capability

Only if you build into a new app folder. 

<br>
> This change gets around this by adding an optional build flag that, when set, will add the capability to the manifest

Optional build flag? Where can I find it? How do I set it? I synced to your branch and it didn't do it automatically. I looked at the Build Window. No luck.

wiwei:
> Yay, thanks for adding this, @wiwei! A few comments/questions:
> 
> > every single time you rebuild the Unity project, you have to modify your AppX manifest and re-add the gazeInput capability
> 
> Only if you build into a new app folder.
> 
> 
> > This change gets around this by adding an optional build flag that, when set, will add the capability to the manifest
> 
> Optional build flag? Where can I find it? How do I set it? I synced to your branch and it didn't do it automatically. I looked at the Build Window. No luck.

@sostel, updated the PR description to show the picture (as well as fix that incorrect statement)

luis-valverde-ms:
Status update: waiting on @keveleigh and @MenelvagorMilsom reviews.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
```diff
@@ -518,6 +522,18 @@ private void AppxBuildGUI()
                 EditorUserBuildSettings.wsaArchitecture = newBuildArchitectureString;
             }
 
+            // The 'Gaze Input' capability is newly added for the Windows SDK 18362, but existing versions
```
keveleigh: Nit: The capability was added earlier for preview APIs for desktop eye trackers. Support on HoloLens was added in 18362.

```suggestion
            // The 'Gaze Input' capability is newly added for the Windows SDK 17134, but existing versions
```

sostel:
> > Yay, thanks for adding this, @wiwei! A few comments/questions:
> > > every single time you rebuild the Unity project, you have to modify your AppX manifest and re-add the gazeInput capability
> > 
> > 
> > Only if you build into a new app folder.
> > > This change gets around this by adding an optional build flag that, when set, will add the capability to the manifest
> > 
> > 
> > Optional build flag? Where can I find it? How do I set it? I synced to your branch and it didn't do it automatically. I looked at the Build Window. No luck.
> 
> @sostel, updated the PR description to show the picture (as well as fix that incorrect statement)

I've tried it again, but I still don't get it to work. Does this assume that I *have to* build through the MRTK build menu? It won't automatically check the capability if I build via "File -> Build Settings -> Build" (the way I usually do it)? I at least didn't know that it's a strict requirement and I'm suspecting that others might not either. Could we just add an option to the MRTK profiler itself to run a post-processing script (e.g., if EyeGazeProvider is used, enable Gaze Input as capability)?

wiwei:
> > > Yay, thanks for adding this, @wiwei! A few comments/questions:
> > > > every single time you rebuild the Unity project, you have to modify your AppX manifest and re-add the gazeInput capability
> > > 
> > > 
> > > Only if you build into a new app folder.
> > > > This change gets around this by adding an optional build flag that, when set, will add the capability to the manifest
> > > 
> > > 
> > > Optional build flag? Where can I find it? How do I set it? I synced to your branch and it didn't do it automatically. I looked at the Build Window. No luck.
> > 
> > 
> > @sostel, updated the PR description to show the picture (as well as fix that incorrect statement)
> 
> I've tried it again, but I still don't get it to work. Does this assume that I _have to_ build through the MRTK build menu? It won't automatically check the capability if I build via "File -> Build Settings -> Build" (the way I usually do it)? I at least didn't know that it's a strict requirement and I'm suspecting that others might not either. Could we just add an option to the MRTK profiler itself to run a post-processing script (e.g., if EyeGazeProvider is used, enable Gaze Input as capability)?

It only works if you go through the MRTK build menu (i.e. the thing that is in the picture above). I'm not super excited about doing this in general, if only because I don't believe that modifying (silently) the behavior of Unity's built in build behaviors is something that's good to do (i.e. automagically making it so that Unity's build behaves slightly differently depending on something that's configured in your current scene - what if actually it's some other scene you have that isn't the main active profile? What if you're switching profiles at runtime?)

PR #4895, opened 2019-06-14
[Shell parity] Bounding Box Proximity Scaling handles

### LINKED ISSUE #4415: [MRTK Shell Parity] [Bounding Box] Handles should be scaled up on hand proximity

johnppella:
## Overview
Updated PR replacing #4646
Merged with the latest mrtk_development branch.

Fixes #4415


cre8ivepark:
Verified everything works well on the device. Also tested BoundingBoxRuntime Test
I have optimized the default handle size for the smaller object. The handles might look small in the test scene since the object is very large.

2019-06-14 14_31_55-Unity 2018 4 0f1 Personal - BoundingBoxRuntimeExample unity - MRTK-Public-Micros
2019-06-14 14_33_41-Unity 2018 4 0f1 Personal - BoundingBoxRuntimeExample unity - MRTK-Public-Micros

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -790,21 +947,36 @@ private void Start()
             {
                 Active = true;
             }
+            else if (activation == BoundingBoxActivationType.ActivateManually)
+            {
+                //activate to create handles etc. then deactivate.
```
julenka: Couldn't you just call CreateRig()?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -790,21 +947,36 @@ private void Start()
...
-            {
-                UpdateBounds();
-                UpdateRigHandles();
-                Target.transform.hasChanged = false;
+                if (currentPointer != null)
+                {
+                    TransformTarget();
+                    UpdateBounds();
+                    UpdateRigHandles();
+
```
julenka: nit:extra space

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1107,16 +1296,34 @@ private void AddLinks()
...
+                    }
+
                     if (rotationHandlePrefabColliderType == RotationHandlePrefabCollider.Sphere)
                     {
-                        SphereCollider collider = ball.AddComponent<SphereCollider>();
-                        collider.radius = 0.5f * rotationHandleDiameter;
+                        SphereCollider sphereCollider = ball.AddComponent<SphereCollider>();
+                        sphereCollider.radius = 0.5f * rotationHandleDiameter;
+                        float localScale = ball.transform.localScale.x;
+                        localScale = localScale == 0.0f ? 0.0001f : localScale;
```
julenka: What is this for? Can you explain via  a comment? I am not immediately clear why if localScale is zero we force it to be non-zero

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1107,16 +1296,34 @@ private void AddLinks()
...
+                        sphereCollider.radius = 0.04f / localScale;
                     }
-                    else
+                    else if( rotationHandlePrefabColliderType == RotationHandlePrefabCollider.Box)
                     {
                         Debug.Assert(rotationHandlePrefabColliderType == RotationHandlePrefabCollider.Box);
                         BoxCollider collider = ball.AddComponent<BoxCollider>();
-                        collider.size = rotationHandleDiameter * Vector3.one;
+                        float localScale = ball.transform.localScale.x;
+                        localScale = localScale == 0.0f ? 0.01f : localScale;
```
julenka: Looks like this is duplicate from above, consider pulling out of the if statement.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1196,6 +1389,30 @@ private void AddLinks()
             }
         }
 
+        private void HandleIgnoreCollider()
```
julenka: Can you please explain why you are doing this?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1734,6 +1954,126 @@ private void UpdateRigHandles()
...
+            handle.transform.localScale = new Vector3(newLocalScale, newLocalScale, newLocalScale);
+
+            Collider collider = handle.gameObject.GetComponent<Collider>();
+
+            float colliderScalar = type == HandleType.Rotation ? rotationHandleColliderDiameter : cornerHandleColliderScale;
+            if (collider && collider is BoxCollider)
+            {
+                newLocalScale = Mathf.Abs(newLocalScale);
+                newLocalScale = newLocalScale == 0.0f ? 0.01f : newLocalScale;
+                float colliderScale = colliderScalar * 0.02f;
```
julenka: It would be nice to avoid magic numbers like this, and in the line above. Can you explain why you have this multiplier here?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1107,16 +1296,34 @@ private void AddLinks()
...
                     }
-                    else
+                    else if( rotationHandlePrefabColliderType == RotationHandlePrefabCollider.Box)
                     {
                         Debug.Assert(rotationHandlePrefabColliderType == RotationHandlePrefabCollider.Box);
                         BoxCollider collider = ball.AddComponent<BoxCollider>();
-                        collider.size = rotationHandleDiameter * Vector3.one;
+                        float localScale = ball.transform.localScale.x;
+                        localScale = localScale == 0.0f ? 0.01f : localScale;
+                        float finalScale = rotationHandleColliderDiameter * 0.02f;
```
julenka: It woul be nice to avoid magic numbers like this. Can you explain why you have this multiplier here? Especially since you re-use 0.02f down below...

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1754,6 +2094,46 @@ private HandleType GetHandleType(Transform handle)
...
+        {
+            if (EyeTrackingProvider != null && EyeTrackingProvider.GazePointer != null && eyeTrackingProvider.GazePointer.Result != null)
+            {
+                point = eyeTrackingProvider.GazePointer.Result.Details.Point;
+                return true;
+            }
+            point = new Vector3(float.NaN, float.NaN, float.NaN);
+            return false;
+        }
+        private void GetControllerPoints(List<Vector3> points)
```
julenka: nit: please add a line between methods.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: when localScale was zero there was a divide error that caused code to be skipped over when the exception threw. I guess this current code could be replaced with :
localScale = Mathf.Max(0.0001f, localScale);

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: sure.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: its just to keep the numbers from getting really small for the user. These could be removed.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: yes

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: yes

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
cre8ivepark: This part is needed when BB is used with an object that has a rigid body. Because of the BB handle's collider, the object is slightly lifted up from the floor(or any colliderable plane) by the 1/2 amount of the collider size. This is an optional field that developers can specify the plane to ignore collision. (might be important for the situation when the object needs to be placed perfectly on the ground + uses BB handles.

cre8ivepark:
- Applied PR feedback
- Merged mrtk_development

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
julenka: Could you please provide an explanation? @cre8ivepark I was hoping that this magic number would be explained and moved out into a variable, but it's not something that I would block on. It would just be nice to reduce magic numbers in our code.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1752,6 +2094,49 @@ private HandleType GetHandleType(Transform handle)
             return HandleType.None;
         }
 
+        private bool TryGetPointerPoint(Handedness handed, out Vector3 point)
+        {
+            IMixedRealityHand hand = HandJointUtils.FindHand(handed);
+            if (hand != null && hand.InputSource.Pointers[0].Result != null)
+            {
+                point = hand.InputSource.Pointers[0].Result.Details.Point;
+                hand = null;
```
thalbern: we don't have to set this here - we're returning anyways

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -429,6 +451,24 @@ public float RotationHandleDiameter
             }
         }
 
+        [SerializeField]
+        [FormerlySerializedAs("ballRadius")]
```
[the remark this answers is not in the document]
johnppella: youre right- this looks like a copy paste error. removing the formerly line and changing tooltip

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -532,13 +572,97 @@ public bool ShowRotationHandleForZ
...
-
         public bool DrawTetherWhenManipulating
         {
             get { return drawTetherWhenManipulating; }
             set { drawTetherWhenManipulating = value; }
         }
 
+        [Header("Proximity")]
+        [SerializeField]
+        private bool proximityEffectActive = true;
```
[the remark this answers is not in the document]
johnppella: yes I would say its definitely time for that!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -532,13 +572,97 @@ public bool ShowRotationHandleForZ
...
+                return closeScale;
+            }
+            set
+            {
+                closeScale = value;
+            }
+        }
+        [SerializeField]
+        [Range(0.0f, 1.0f)]
+        private float farGrowRate = 0.3f;
```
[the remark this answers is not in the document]
johnppella: yes

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -532,13 +572,97 @@ public bool ShowRotationHandleForZ
...
+        [SerializeField]
+        [Tooltip("By a factor of how much do the handles grow when a hand gets close")]
+        [Range(0.005f, 0.1f)]
+        private float handleMediumProximity = 0.005f;
+        [SerializeField]
+        [Tooltip("How far away should the hand be before it starts scaling a handle")]
+        [Range(0.001f, 0.01f)]
+        private float handleCloseProximity = 0.0015f;
+        [SerializeField]
+        private float farScale = 0.01f;
```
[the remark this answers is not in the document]
johnppella: oops the tooltips were mixed up. fixed.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1125,25 +1334,11 @@ private void AddLinks()
                     ApplyMaterialToAllRenderers(ball, handleMaterial);
 
                     balls.Add(ball.transform);
+                    ballsProximate.Add(HandleProximityState.FullsizeNoProximity);
+                    ballRenderers.Add(ball.GetComponent<Renderer>());
```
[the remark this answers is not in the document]
johnppella: good point. if rotation handle is a primitive then it would always have a Renderer- BUT! if its a user-created prefab it might not. Adding check.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: yes

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
johnppella: removed

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1732,6 +1954,126 @@ private void UpdateRigHandles()
...
+            float newLocalScale = (handle.transform.localScale.x * (1.0f - weight)) + (scaleHandleSize * targetScale * weight) * defaultScale;
+            handle.transform.localScale = new Vector3(newLocalScale, newLocalScale, newLocalScale);
+
+            Collider collider = handle.gameObject.GetComponent<Collider>();
+
+            float colliderScalar = type == HandleType.Rotation ? rotationHandleColliderDiameter : cornerHandleColliderScale;
+            if (collider && collider is BoxCollider)
+            {
+                newLocalScale = Mathf.Abs(newLocalScale);
+                newLocalScale = newLocalScale == 0.0f ? 0.01f : newLocalScale;
```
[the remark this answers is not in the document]
johnppella: removed- divide by zero problem no longer exists.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1732,6 +1954,126 @@ private void UpdateRigHandles()
...
+                float colliderScale = colliderScalar * 0.02f;
+                (collider as BoxCollider).size = new Vector3(colliderScale / newLocalScale, colliderScale / newLocalScale, colliderScale / newLocalScale);
+            }
+            else if (collider && collider is SphereCollider)
+            {
+                newLocalScale = Mathf.Abs(newLocalScale);
+                float colliderScale = colliderScalar * 0.02f;
+                newLocalScale = newLocalScale == 0.0f ? 0.01f : newLocalScale;
```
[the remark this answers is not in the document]
johnppella: removed

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -538,13 +579,103 @@ public bool ShowRotationHandleForZ
...
+        [Tooltip("How far away should the hand be from a handle before it starts scaling the handle?")]
+        [Range(0.005f, 0.1f)]
+        private float handleMediumProximity = 0.005f;
+        [SerializeField]
+        [Tooltip("How far away should the hand be from a handle before it activates the close-proximity scaling effect?")]
+        [Range(0.001f, 0.01f)]
+        private float handleCloseProximity = 0.0015f;
+        [SerializeField]
+        [Tooltip("A Proximity-enabled Handle scales by this amount when a hand moves out of range")]
+        private float farScale = 0.0f;//0.01f;
```
thalbern: suggestion: use 1.0f as default value? we don't want users to set this up as 0.0f - so we shouldn't use it as default

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -806,9 +959,20 @@ public void CreateRig()
...
 
+
         #region MonoBehaviour Methods
+
         private void Start()
         {
+            //override Shell Parity settings if no prefab is specified for corner handles;
+            if (scaleHandlePrefab == null)
+            {
+                closeScale = 1.8f;
```
thalbern: the defaults for these fields are different than the values here - I would prefer we stick with one default setup

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -806,9 +959,20 @@ public void CreateRig()
...
+                mediumScale = 1.5f;
+                farScale = 1.0f;
+            }
+            farScale = Mathf.Max(farScale, 0.01f);
```
thalbern: Questions:
do we want to support users modifying these values during runtime? if so we wouldn't catch if someone puts a 0 there (moving it to the setter would fix that i suppose)

will we have a similar problem with the medium and close scale?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
cre8ivepark: @thalbern In this update, the handle is invisible(scale = 0) in default. The size increases based on the hand proximity.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
cre8ivepark: These are optimized handle size values for the default blue (HL1 style) handles when no handle prefabs are assigned. When custom handle prefabs including HL2 shell-parity handles are assigned, BB uses the default or assigned values through the inspector or code. With single default values, we couldn't achieve a proper scale.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1956,18 +1957,17 @@ private void UpdateRigHandles()
...
-                for (int i = 0; i < cornerVisuals.Count; i++)
-                {
-                   cornerVisuals[i].localScale = cornerDimensions;
-                }
-
-                for (int i = 0; i < balls.Count; i++)
-                {
-                    balls[i].localScale = ballDimenions;
-                }
+                //Vector3 cornerDimensions = Vector3.Scale(cornerVisualScaleToMatchHandleSize * Vector3.one, invRootScale);
```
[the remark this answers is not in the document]
cre8ivepark: This was intentional delete. I'll remove the commented out code lines. 
Reference: #5055

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
julenka: Instead of magic numbers could we move these to fields or constants in the file?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
cre8ivepark: - Removed hard-coded values. Optimized default values for the HL1 style (for the script used without prefabs)
- Removed Mathf.Max(farScale, 0.01f) line since it should allow 0 to make handle invisible
Example of FarScale = 0 / MediumScale = 1.0 / CloseScale = 1.5
MRTK_BBHandleSteps

cre8ivepark:
Removed hard-coded values. Optimized default values for the HL1 style (for the script used without prefabs)
For using HL2 handle prefabs, developers need to put proper scaling values. I'll update the docs for this.

cre8ivepark:
@julenka 7/15: Merged the latest mrtk_development, resolved merge conflicts in the Slate.prefab, HandInteractionExamples.unity and BoundingBox.cs
Verified the behavior in the Unity editor. 
*Slate.prefab and HandInteractionExamples are updated since they needed to have new 1meter-based handle assets and proper scaling values.

cre8ivepark:
@julenka Merged your latest updates. Tested in HandInteractionExamples and BoundingBoxExamples scene. Everything looks great. Now it is much easier to grab. Thank you!

One small update: The default Y value of the Rotation Handle Collider Padding was 0.0f. Was it intentional? I have updated it to 0.016f for the consistency.  (screenshots below)

```
        [SerializeField]
        [Tooltip("Additional padding to apply to the collider on rotate handle to make handle easier to hit")]
        private Vector3 rotateHandleColliderPadding = new Vector3(0.016f, 0.016f, 0.016f);
```

Before & After
2019-07-16 20_18_17-Untitled-1 @ 100% (Layer 2, RGB_8#) _

julenka:
> One small update: The default Y value of the Rotation Handle Collider Padding was 0.0f. Was it intentional? I have updated it to 0.016f for the consistency. (screenshots below)

I was worried that the 0.016 on the y would make rotate handles too big, but if you are okay with it, let's just go with 0.016f.

Issue #4202, opened 2019-05-07
MRTK uses OnValidate to initialize when it shouldn't be loaded


chbecker-ms:
OnValidate is being called, when a blank scene is loaded without MRTK even enabled.

OnValidate should not be using to initialize objects:

NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.Core.Definitions.MixedRealityCameraProfile.ApplySettingsForOpaqueDisplay () (at Assets/Libraries/Core/Assets/Core/MRTK/MixedRealityToolkit/Definitions/MixedRealityCameraProfile.cs:89)​
Microsoft.MixedReality.Toolkit.Core.Services.MixedRealityToolkit.InitializeServiceLocator () (at Assets/Libraries/Core/Assets/Core/MRTK/MixedRealityToolkit/Services/MixedRealityToolkit.cs:181)​
Microsoft.MixedReality.Toolkit.Core.Services.MixedRealityToolkit.InitializeInstance () (at Assets/Libraries/Core/Assets/Core/MRTK/MixedRealityToolkit/Services/MixedRealityToolkit.cs:458)​
Microsoft.MixedReality.Toolkit.Core.Services.MixedRealityToolkit.OnValidate () (at Assets/Libraries/Core/Assets/Core/MRTK/MixedRealityToolkit/Services/MixedRealityToolkit.cs:548)​

Railboy:
What version / branch produces these errors?

keveleigh:
@Railboy This one is a migration from our internal bug tracker to GitHub. It was originally filed in February, so it's possible much of this has been mitigated by your work since then.

If I'm remembering correctly, this issue only occurred in a specific project where, after a build, Unity reloaded the scene in the editor and a Toolkit script was trying to reference the camera before it had been reloaded, causing Unity to crash. I've never been able to repro it though.

Railboy:
I don't believe this will be a problem moving forward but for good measure we ought to remove the OnValidate call altogether. I'll do so in PR #4035

davidkline-ms:
Sounds good @Railboy! Will assign this issue to you

carmines:
Just to clarify, the issue is that OnValidate is being used incorrectly for all of these scripts. Remove them and use Awake() or OnEnable().

If the intent is to validate an object inspector properties, you should be using editor scripts. Additionally, if you are checking Ranges for properties, ensure the appropriate tags are being used.

[...]

PR #4899, opened 2019-06-14
 Removes OnValidate in cases where it was being used inappropriately

Railboy:
## Overview

There were several cases where OnValidate was being used inappropriately. This PR either moves those checks into Awake / OnEnable or removes them entirely.

OnValidate should be used to validate settings when standard field and RequireComponent attributes aren't up to the task, and where the same validation in an inspector would be unreasonably complex.

Some cases use OnValidate correctly. They have been marked with a comment.

ExecuteInEditMode has been replaced with [ExecuteAlways](https://docs.unity3d.com/ScriptReference/ExecuteAlways.html) for consistency. Behavior is the same.

## Changes
- Fixes: #4202


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointerVisual.cs
```diff
@@ -82,6 +77,11 @@ private void CheckAsset(object asset, string fieldname)
 
         public void Update()
         {
+            if (!Application.isPlaying)
```
keveleigh: Can update be called when the app isn't playing?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointerVisual.cs
```diff
@@ -82,6 +77,11 @@ private void CheckAsset(object asset, string fieldname)
 
         public void Update()
```
keveleigh: Do you know if there's a reason all the MonoBehaviour methods are `public` in this class?

Assets/MixedRealityToolkit.SDK/Features/Audio/Influencers/AudioInfluencerController.cs
```diff
@@ -151,7 +151,9 @@ public float NativeHighPassCutoffFrequency
         }
 
         private void Awake()
-        {           
+        {
+            audioSource = GetComponent<AudioSource>();
```
keveleigh: For many of these, the field will likely have

```
[SerializeField]
[HideInInspector]
```

which could be removed as well if we're moving to this pattern.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointerVisual.cs
Railboy: I don't. I'd prefer if they were private.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointerVisual.cs
Railboy: When a class uses the ExecuteAlways attribute Update can get called.

keveleigh: This class doesn't appear to have that attribute though

Railboy: My mistake, thanks.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/CursorModifier.cs
```diff
@@ -243,7 +243,7 @@ void IMixedRealityFocusChangedHandler.OnBeforeFocusChange(FocusEventData eventDa
         #region MonoBehaviour Implementation
 
         private void OnValidate()
-        {
+        {   // This is an appropriate use of OnValidate
```
keveleigh: I'm not super convinced there's a difference between this case and any of the others. In both cases, we're calling `GetComponent`. One just serializes the data while the other checks that it exists.

Serializing data at edit time is a valid thing to do. What's inappropriate about the other uses?

Railboy: Yeah it's pretty subjective, I could go either way on this one. I think what tipped the scales here was:
- you can't use RequireComponent because you need a _generic_ collider
- a custom inspector seems like total overkill for such a simple check
- this feels like an issue we should raise in-editor? This is the vaguest criteria.

Railboy:
Merged latest changes.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/ParabolicTeleportPointer.cs
```diff
@@ -26,36 +27,33 @@ public class ParabolicTeleportPointer : TeleportPointer
...
-                parabolicLineData.LineTransform = pointerHelper.transform;
-            }
         }
 
         private void EnsureSetup()
         {
             if (parabolicLineData == null)
             {
-                parabolicLineData = GetComponent<ParabolaPhysicalLineDataProvider>();
+                parabolicLineData = gameObject.GetComponent<ParabolaPhysicalLineDataProvider>();
```
julenka: Noticed this one was bigger than the others -- just making sure since we don't have test coverage here, could you check to make sure teleport pointers behave okay after this change?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -369,5 +369,13 @@ public override void OnInputUp(InputEventData eventData)
             // Poke pointer should not respond when a button is released or hand is un-pinched
             // It should only dispatch events based on collision with touchables.
         }
+
+        protected override void OnEnable()
+        {
+            base.OnEnable();
+
+            Debug.Assert(line != null, "No line renderer found in PokePointer.");
```
julenka: Why does pokepointer need a line renderer? Not sure we want to assert here.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointerVisual.cs
```diff
@@ -6,6 +6,7 @@
 
 namespace Microsoft.MixedReality.Toolkit.Input
 {
+    [ExecuteAlways]
```
julenka: This didnt' have execute in edit mode, why did you add ExecuteAlways here?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointerVisual.cs
```diff
@@ -82,6 +78,11 @@ private void CheckAsset(object asset, string fieldname)
 
         public void Update()
         {
+            if (!Application.isPlaying)
```
julenka: Why do you need this check here? I don't see why sphere pointer needs to have executealwaysattribute, and therefore why it needs this check.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
```diff
@@ -76,6 +76,11 @@ public void SetLocalUp(Vector3 newLocalUp)
 
         protected new void OnValidate()
         {
+            if (Application.isPlaying)
+            {   // Don't validate during play mode
```
julenka: Why don't we want to validate this one during play mode, but it's okay to do this in the components?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
Railboy: No idea. My priority was just to move code out of OnValidate.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointerVisual.cs
Railboy: To ensure OnEnable / CheckInitialization is called in edit mode. This is a substitute for OnValidate.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointerVisual.cs
Railboy: This class now uses ExecuteAlways so OnEnable is called in edit mode. Update also gets called, but because it executes more than initialization logic we exit early if we aren't in play mode.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
Railboy: In play mode the values being validated here are automatically checked by the methods you use to set them (eg SetLocalForward). So OnValidate is only necessary if you're tweaking the values by hand in the editor.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/ParabolicTeleportPointer.cs
Railboy: Teleportation and tooltips work - if you know of other components using parabolas let me know and I'll test them as well.

Issue #3767, opened 2019-04-02
Cleanup the EyeTracking Demo code and other HL2 artifacts


wiwei:
There's significant technical debt in the eye tracking demo code which we should clean up.

This has been filed to make sure we don't lose some of the interesting TODOs or comments in the following PR:

https://github.com/Microsoft/MixedRealityToolkit-Unity/pull/3755

wiwei:
InputPointVisualizer.cs had a large TODO:

  /// ------------
    /// *** TODO ***
    /// ------------
    /// - Load log files
    /// - Replace GameObjects with particle systems for higher perf on HoloLens
    /// - Question: How to visualize data from multiple ppl? Color lookup table? Limit number of scanpaths visible at a time?
    /// - Question: How can users load files from HoloLens? Built-in file explore? Load by session?
    /// - Allow for tweaking parameters while viewing the scene, such as "transparency" etc.
    /// ------------

Also VisModes needs summaries.

wiwei:
LogStructure_EyeGaze had a lot of TODO code (around raw eye gaze tracking) removed.

wiwei:
OnSelect_VisualizerInputController.cs needs tooltips for its serialized fields

wiwei:
ParticleHeatmap_ParticleData.cs needs a summary

wiwei:
UserInput_Playback.cs needs significant cleanup (see the original PR)

wiwei:
WindowsMixedRealityEyeGazeDataProvider.cs needs work around its confidenceOfSaccadeThreshold

wiwei:
Need to also followup on Interactable.cs's


            if (!CanInteract() || true)
            {
                return;
            }

            if (StateManager != null)
            {
                if (eventData == null && (HasFocus || IsGlobal)) // handle brute force
                {
                    StateManager.SetStateValue(InteractableStates.InteractableStateEnum.Visited, 1);
                    IncreaseDimensionIndex();
                    OnClick.Invoke();
                }
                else if (eventData != null && ShouldListen(eventData.MixedRealityInputAction))
                {
                    StateManager.SetStateValue(InteractableStates.InteractableStateEnum.Visited, 1);
                    IncreaseDimensionIndex();
                    OnClick.Invoke();
                    eventData.Use();
                }
            }
            */

Yoyozilla:
I think this is done?

keveleigh:
Some is in progress with #4032

[...]

PR #4919, opened 2019-06-17
Updates Eye Tracking Demo to use Scene System

Railboy:
## Overview

The Eye Tracking Demo uses a manager scene and additive loading, which is an ideal case for a Scene System demonstration. This PR updates the demo to use the scene system for loading and unloading of content.

- SceneSystem is now enabled in eye tracking configuration profile
- SceneSystem profile includes manager scene and all content scenes. (Lighting scene is disabled.)
- `LoadAdditiveScene.cs` and `OnLoadStartScene.cs` now use `MixedRealityToolkit.SceneSystem` to load content instead of `SceneManager`.

## Changes
- Related to #3767

## Verification
@sostel Can you verify that everything still works as intended?

>
>


sostel:
I've synced to the 'Railboy:eye_tracking_demo_scene_system' branch, but it's not working well for me :( 

**Initial setup:**
1. I've loaded the EyeTrackingDemo-00-RootScene.unity. 
2. When hitting "play", it shows the following errors:
image

3. I checked the 'Mixed Reality Toolkit -> Scene System':
image

4. Turns out: Users still need to add the scenes _again_ and check the "Editor Manage Build Settings" to get the ET demos to work. Is that intentional? 
---

A few additional remarks:
- Slow loading: I noticed that it takes quite a while after hitting "Play" that something happens. Sometimes up to half a minute.

- Once it goes into play mode, the view always jumps back from "Scene System" to "Camera".

- The "add logic" is not clear: I selected "Editor Manage Build Settings" and added the links to the scenes. Some scenes were added to the Build menu already (00 and 01) and some were new (02,03,04,05). When I hit "play", it threw the below error. I had to start it three times until it finally loaded the scene (EyeTrackingDemo-02...) correctly. It seems it only adds one scene at a time when I leave the play mode or when I select the 'Scene System' tab. So, the "Editor Manage Build Settings" isn't reliable. 

image

Also, I'm confused about the difference between "Editor Manage Build Settings" and "Editor Manage Loaded Scenes".

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Scripts/TargetBehaviors/LoadAdditiveScene.cs
```diff
@@ -26,45 +26,23 @@ public class LoadAdditiveScene : MonoBehaviour
...
         private float waitTimeInSecBeforeLoading = 0.25f;
 
-
         public void LoadScene()
         {
             LoadScene(SceneToBeLoaded);
         }
 
-        public void LoadScene(string sceneName)
+        public async void LoadScene(string sceneName)
```
sostel: Given that the script name is "LoadAdditiveScene", it might be nice to either use a completely new script and leave this one as it is. Or to also rename the script itself.

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Scripts/TargetBehaviors/OnLoadStartScene.cs
```diff
@@ -22,10 +22,7 @@ public void Start()
 
         private void LoadNewScene()
         {
-            if (SceneToBeLoaded != "")
-            {
-                SceneManager.LoadSceneAsync(SceneToBeLoaded, LoadSceneMode.Additive);
-            }
+            MixedRealityToolkit.SceneSystem.LoadContent(SceneToBeLoaded, LoadSceneMode.Additive);
```
sostel: What if the "SceneToBeLoaded" has not been assigned?

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Profiles/EyeTrackingDemoSceneSystemProfile.asset
```diff
@@ -0,0 +1,72 @@
...
+  m_EditorClassIdentifier: 
+  isCustomProfile: 1
+  useManagerScene: 1
+  managerScene:
+    Name: EyeTrackingDemo-00-RootScene
+    Path: Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Scenes/EyeTrackingDemo-00-RootScene.unity
+    Included: 1
+    BuildIndex: 0
+    Tag: Untagged
+    Asset: {fileID: 3671201678493226958, guid: 27c3fbe0128089f48994cdc6af8668fc, type: 3}
```
sostel: When first trying out this branch, the 'Scene System' showed me that the linked scenes were missing. After some testing, I found out that this was due to the fileID. When I replaced '3671201678493226958' with '102900000' for all entries, without changing anything else, it would correctly detect all the scenes in the Editor again.

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Scripts/TargetBehaviors/OnLoadStartScene.cs
Railboy: The scene system will give you a warning, much like the script used to.

Railboy:
>Also, I'm confused about the difference between "Editor Manage Build Settings" and "Editor Manage Loaded Scenes".

Editor Manage Build Settings is about which scenes are included in your build settings. Editor Manage Loaded Scenes is about which scenes are loaded in the editor.

Maybe these need better names? What about:
- Editor Auto-Include Scenes in Build
- Editor Auto-Load Scenes in Editor

Railboy:
>Turns out: Users still need to add the scenes again and check the "Editor Manage Build Settings" to get the ET demos to work. Is that intentional?

Partly intentional, partly a bug. I've enabled editor manage build settings by default. And the profile will update to reflect changes more aggressively now.

The missing scene / faulty file ID is a different issue, I'll try to repro that.

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/General/Profiles/EyeTrackingDemoSceneSystemProfile.asset
Railboy: Looks like this is a [Unity bug.](https://issuetracker.unity3d.com/issues/sceneasset-references-go-missing-when-upgrading-a-project-from-2018-dot-3-and-earlier-to-2019-dot-1-or-2019-dot-2) I've updated the scene info drawer to automatically by path if the asset is missing. Hopefully this won't be necessary after we all move to 2019.

julenka:
@Railboy and @sostel, is this PR still relevant? I know you've been doing work on the example scene hub is this part of that?

wiwei:
@julenka I'm going to close this one out as it's a couple of months old - there have been some changes since then so it would probably be good to open a new PR against most recent.

Issue #4831, opened 2019-06-11
onSceneGUIDelegate warning in Unity 2019


wiwei:
## Describe the bug

Usage of onSceneGUIDelegate logs warnings in latest Unity 2019 because it's been deprecated and replaced by 

## To reproduce

1) Open latest mrtk_development on the latest Unity 2019.
2) Notice warnings.

## Expected behavior

No warnings

## Your Setup (please complete the following information)

- Unity Version [e.g. 2019.1.ff1]
- mrtk_development

## Additional context

We should replace this with duringSceneGui.

https://github.com/Unity-Technologies/UnityCsReference/blob/master/Editor/Mono/SceneView/SceneView.cs

Conditional on 2018/2019 version checks.

Issue #4545, opened 2019-05-23
Namespace EventSystem not found


MitchelKerckhaert:
## Describe the bug

I have been trying to get started with the MixedRealityToolkit-Unity and am noticing a bug in the setup process. After importing the MixedRealityToolkit 90+ compiler errors pop up. Most of them say "type or namespace 'Evensystems' could not be found", but it appears for other namespaces aswell like: BaseEventData, IEventSystemHandler, RaycastResult and a couple others. I am using Unity 2019.3.0a3 and have setup a project with buildsettings for the Hololens (v1)

## To reproduce

1. Create a new Unity Project.
2. Click on Assets -> Import Package
3. Import the MixedRealityToolkit.
4. See the errors pop up in the console.

## Expected behavior

The MixedRealityToolkit is loaded in Assets. A dropdown menu for the toolkit should appear (among other setting changes).

## Your Setup (please complete the following information)

- Unity Version [2019.3.0a3]
- MRTK Version [ v2.0.0-RC1-Refresh ]

## Target Platform (please complete the following information)

- HoloLens

keveleigh:
>Unity Version [2019.3.0a3]

Our officially supported version is currently Unity 2018.3/.4. It's possible there are some breaking changes in Unity we'll need to account for when we add support for 2019.3, but we can't guarantee the MRTK will work in an alpha release of Unity.

MitchelKerckhaert:
Ah, Thank you! The package is indeed not compatible with Unity 2019.3.x (yet), but after Installing the 2018.3.x version of Unity all problems were resolved.

Kulanei:
Thanks for the answer ! I'll be using Unity 2018.3.x for the project, thanks for your time !

jernkuan:
Does not work for 2019.2 as well. Would be good to keep this issue open, and maybe update the README.MD to indicate the range of Unity version that it would work.

wiwei:
Closing this one out - we're clear in our getting started about using 2018.4.x, and we can update this when we've done the 2019.3 reconciliation

PR #4925, opened 2019-06-17
Remove warnings in Unity 2019

keveleigh:
## Overview
MixedRealityToolkitFacadeHandler and ServiceFacadeInspector:
image

Platform Utility:
image
image

MixedRealityEditorSettings:
image

Note: This doesn't remove all issues with 2019.3, as that version adds a requirement for the `Unity.ugui` package to be added to all our asmdefs (#4545). Unfortunately, that fix is backwards compatible with previous versions of Unity 2019 but causes errors in Unity 2018.

## Fixes
Fixes: #4831


SimonDarksideJ:
Suggest you also look to resolving #4931 as part of this PR.  Simply add a reference to "Unity UI" to Foundation and SDK (possibly Examples) asmdefs

keveleigh:
@SimonDarksideJ Unfortunately, adding that reference is backwards compatible with previous versions of Unity 2019 but causes errors in Unity 2018.

Issue #3264, opened 2018-12-12
Significant perf hit in Unity when MRTK Build Window is opened and visible


whitetigerhelix:
## Overview
I started experiencing extreme chop with Unity: the mouse jumping all over the place, and I couldn't click on anything or interact with the UI reliably.  Playing in the editor was impossible as there would be significant hitching.  I was freaking out because I couldn't make it go away, even after all the usual suspects and rebooting.
 
A colleague pointed out that having the MRTK Build Window visible seems to be causing this.  Sure enough, I switched to my Scene tab and the hitching and perf issues went away.

## Expected Behavior
No impact from having Build Window open

## Actual Behavior
Extreme blocking perf impact when Build Window is open

## Steps to reproduce
Open the MRTK Build Window and have it open and visible in your editor
Observe perf impact when using the editor, and when playing in editor

Note that this may not be as pronounced initially, and may be different for different projects or repos.
Note that rebooting, deleting library, etc do not improve the problem -- the perf hit persists.

## Unity Editor Version
Unity 2018.1.9f2

## Mixed Reality Toolkit Release Version
Latest mrtk_development

StephenHodgson:
Need a bit more information from you:

1. Which version are you using?
2. Can you grab a screenshot of the profiler for the editor?
3. In the profiler, specifically what was causing the perf impacts?

StephenHodgson:
I'm seeing well above 60 fps
![image](https://user-images.githubusercontent.com/13334553/49906129-6b2a9d00-fe25-11e8-83ff-971144459c5c.png)

whitetigerhelix:
Version: latest mrtk_development

The profiler yields nothing interesting.  In fact, this issue causes Unity itself to hitch.  So when watching the profiler, you see it happily profiling, then it pauses for a few fractions of a second, then runs smoothly for maybe a second, then hitches again (repeat endlessly).  In other words, the hitching affects the entire app, including the profiler, and the profiler does not pick up any spikes or anything.  

When I tab away from the Build Window, now Unity stops hitching and is usable.

I have multiple clones of my repo, and strangely the performance impact is not identical between them (some are worse than others, I haven't been able to figure out a pattern).  But the hitching is present for all of them.

Thanks

StephenHodgson:
Nothing coming up on the deep profiler either.

Pretty sure _something_ should be showing up on the profiler.

Do you have anything else that might help figure out what you're seeing, and how we can address it?

[...]

PR #4937, opened 2019-06-18
Build window performance fix

ossala:
## Overview
Fixing issue where the whole editor is locked by periodically checking for new appx directories.
#3264

## Changes
- Removed periodic build checking
- Added a refresh button to check for builds
- Added checking when switching to the deploy tab.


wiwei:
@ossala, thanks for doing this! Our team was just talking about this a little while ago. In order to merge this in, we'll need to wait for you to handle the contributor license agreement step above.

ossala:
Thanks, I've signed it now so should all be sorted 👍

Issue #4218, opened 2019-05-07
Poke pointer shouldn't iterate over all NearTouchInteractables every frame and use interface rather than static list


thalbern:
## Describe the bug

Currently we're iterating over all NearTouchInteractables that are registered to a static instance list every frame in the poke pointer which is a performance issue and will miss any interactable that isn't registering itself to that list.

Proposal:

Can we do a raycast to figure out which objects we're actually pointing to and then check for an interface instead of keeping that list around?

## To reproduce

see code

## Your Setup (please complete the following information)

- MRTK Version v2 RC1

## Target Platform (please complete the following information)

- HoloLens 2

wiwei:
@julenka Do you know why we didn't do a sphere raycast when this was originally written?

It seems like at the very least we should make it so that we don't re-allocate another read only array on each iteration (i.e. the AsReadOnly)

julenka:
Robert Ferrese wrote the original code, we should ask him. Here is a reason I think we might have wanted to keep a list of items instead of doing a ray cass in the scene. Let's say you have the following setup:

![image](https://user-images.githubusercontent.com/168492/57426382-d5c4e680-71d3-11e9-9fb6-3f4d3558c301.png)

In this case, if I cast a ray from the finger and take the closest hit, the ray would hit the grabbable box first, not the touchable collider. A similar thing would happen if you did sphere cast and just picked closest.

Note that we could instead use [Physics.RaycastAll](https://docs.unity3d.com/ScriptReference/Physics.RaycastAll.html)  to get all hits along a ray, then find the closest object that has a NearInteractionTouchable. Note that while this works for surfaces, it may not work if you approach a touchable volume from the side (something I believe Niall is working on ). Instead, using [Physics.SphereCastAll](https://docs.unity3d.com/ScriptReference/Physics.SphereCastAll.html) and finding the closest NearInteractionTouchable could work.

I think it's a reasonable issue to fix. Note also another issue with touch dispatch that we have, #4174.

lukastoenneMS:
A major problem seems to be that NearInteractionTouchable doesn't necessarily use a collider. Easiest/Cleanest solution would be to require collidables for touchables now, but that is a significant breaking change.

[...]

PR #4893, opened 2019-06-14
Playmode tests for NearInteractionTouchable classes

lukastoenneMS:
## Overview

Add playmode tests for basic NearInteractionTouchable functionality.

The tests count touchable events to make sure the 3 touchable types behave as expected.

## Verification
Open the Unity test runner, switch to PlayMode tests and run the NearInteractionTouchableTests section.

PR #4942, opened 2019-06-18
Poke pointer overlap test to replace static touchables list

lukastoenneMS:
## Overview
Replace the touchable search in a static instance list with a physics overlap test.

The `distBack` property has been renamed to `pokeThreshold`. This is the counterpart to the `debounceThreshold` value.

The `distFront` property has been moved to the PokePointer and renamed to `touchableDistance`. This is the distance at which interaction with the pointer will start.

## Changes
- Fixes: #4218 .

_Note: Contains touchables tests from #4893, this should be merged soon._


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -13,7 +13,25 @@ public class PokePointer : BaseControllerPointer, IMixedRealityNearPointer
...
-        
+
+        /// <summary>
+        /// Maximum distance a which a touchable surface can be interacted with.
+        /// </summary>
+        [SerializeField]
+        protected float touchableDistance = 0.2f;
+        public float TouchableDistance => touchableDistance;
+
+        /// <inheritdoc />
```
luis-valverde-ms: >/// <inheritdoc /> [](start = 8, length = 18)

Is this inherited?

lukastoenneMS: No, will fill out the comment.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -93,6 +115,29 @@ public override void OnPreSceneQuery()
             visuals.SetActive(IsActive);
         }
 
+        private bool FindClosestTouchableForLayerMask(LayerMask layerMask, out ColliderNearInteractionTouchable closest, out float distance, out Vector3 normal)
+        {
+            closest = null;
+            distance = float.PositiveInfinity;
+            normal = Vector3.zero;
+
+            Collider[] colliders = UnityEngine.Physics.OverlapSphere(Position, touchableDistance, layerMask, triggerInteraction);
```
Railboy: You can do this without allocations: https://docs.unity3d.com/ScriptReference/Physics.OverlapSphereNonAlloc.html

Have we profiled this stuff for speed etc?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -13,7 +13,25 @@ public class PokePointer : BaseControllerPointer, IMixedRealityNearPointer
...
+        /// <summary>
+        /// Maximum distance a which a touchable surface can be interacted with.
+        /// </summary>
+        [SerializeField]
+        protected float touchableDistance = 0.2f;
+        public float TouchableDistance => touchableDistance;
+
+        /// <inheritdoc />
+        private LayerMask[] pokeLayerMasks = null;
+        public LayerMask[] PokeLayerMasks => pokeLayerMasks;
```
luis-valverde-ms: If this is not serialized, can we do this instead?
public PokeLayerMasks LayerMask[] { get; private set; }?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
lukastoenneMS: I have read conflicting explanations of the OverlapSphereNonAlloc variant regarding resizing of the colliders buffer. There is not a great way to know the number of potential touchables we could find, so we can't allocate the buffer reliably beforehand. Some commenters said that it gets resized if the buffer is too small, but most say that you will miss potential colliders if the buffer is too small. We could of course just make this a configurable profile setting with a reasonable default (say 64) and print a warning if the buffer ever gets full.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
lukastoenneMS: Yes, good point.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
Railboy: >Some commenters said that it gets resized if the buffer is too small, but most say that you will miss potential colliders if the buffer is too small. 

I can test this to confirm but my understanding was that you'll miss colliders. The amount of garbage we'll generate with the non-alloc version is brutal so the overhead of a huge buffer is worth it. Making it a profile setting sounds ideal.

lukastoenneMS: Fixed, i'm using the NonAlloc variant now.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -46,31 +83,23 @@ public override void OnPreSceneQuery()
...
-            BaseNearInteractionTouchable newClosestTouchable = null;
+            var layerMasks = PrioritizedLayerMasksOverride ?? PokeLayerMasks;
+
+            // Find closest touchable
+            ColliderNearInteractionTouchable newClosestTouchable = null;
+            foreach (var layerMask in layerMasks)
             {
-                closestDistance = float.PositiveInfinity;
-                foreach (var prox in BaseNearInteractionTouchable.Instances)
+                if (FindClosestTouchableForLayerMask(layerMask, out newClosestTouchable, out closestDistance, out closestNormal))
```
julenka: Something to be careful of about touch interactions on HoloLens is people tend to move their hands quite far behind UI. For example, here is a capture of hand motion as somebody presses a button on a 2D slate:

image

Notice how far behind the slate the person's finger moves.

If we go with this approach it will be important to keep touchableDistance quite high (~ .5,[ average arm length of male is 65 cm](https://www.reference.com/science/long-average-human-arm-62c7536c5e56f385)).

Reason is it would surprise people if they are for example scrolling and get a touch up event as they are scrolling the surface. The reason this could happen is people could move their hand quite far behind the slate because there's no physical surface stopping them.

I'm not sure a value of 0.2 will be high enough to cover most of the ways that people scroll content -- people will likely disengage by moving their hand too far back behind the surface.

Note that if the same variable for determining maximum touchable distance ALSO turns on the poke pointer visual and turns on input focus, then you will start getting focus events from poke pointers even when the pointer is very far away.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnboundedPlane.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// in order to get PointerDown and PointerUp events whenever a PokePointer touches the
     /// attached collider through this surface.
     /// </summary>
-    public class NearInteractionTouchableUnboundedPlane : BaseNearInteractionTouchable
+    public class NearInteractionTouchableUnboundedPlane : ColliderNearInteractionTouchable
```
julenka: I'm not sure this class makes sense any more. It's original intention was to allow for a plane of infinite size that would be touchable. However, it seems like the poke pointer will always dispatch touchUp events once a pointer has passed some distance threshold. Additionally, I don't think the current behavior actually allows a touchable to be infinitely large. Finally, it makes the definition of "NearInteractionTouchable" a bit more muddled since NearInteractionTouchable is really a touchable surface of bounded size, which is essentially the same thing that this class is. 

How do you feel about removing it? I don't think anything except the touchable sample uses it.

julenka: It also was not released in RC2 so is not a breaking change.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityPointerProfile.cs
```diff
@@ -30,6 +30,23 @@ public class MixedRealityPointerProfile : BaseMixedRealityProfile
         /// </summary>
         public LayerMask[] PointingRaycastLayerMasks => pointingRaycastLayerMasks;
 
+        [SerializeField]
+        [Tooltip("The LayerMasks, in prioritized order, that are used to determine the touchable objects.")]
```
julenka: Can you clarify that to make an object touchable it must have both nearinteractiontouchable AND the layer mask?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -46,31 +83,23 @@ public override void OnPreSceneQuery()
...
                 }
             }
 
             if (newClosestTouchable != null)
             {
                 // Build ray (poke from in front to the back of the pointer position)
-                Vector3 start = Position - newClosestTouchable.DistBack * -closestNormal;
-                Vector3 end = Position + newClosestTouchable.DistFront * -closestNormal;
+                Vector3 start = Position - newClosestTouchable.PokeThreshold * -closestNormal;
+                Vector3 end = Position + touchableDistance * -closestNormal;
```
luis-valverde-ms: Speaking from total ignorance, making the touchable distance a pointer parameter instead of a setting per touchable seems like a breaking change. What's the motivation?

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionTouchable.cs
```diff
@@ -29,57 +26,47 @@ public abstract class BaseNearInteractionTouchable : MonoBehaviour
...
-        }
+        public abstract float DistanceToTouchable(Vector3 samplePoint, out Vector3 normal);
+    }
 
-        protected void OnValidate()
+    /// <summary>
+    /// Base class for all touchables using colliders
+    /// </summary>
+    [RequireComponent(typeof(Collider))]
+    public abstract class ColliderNearInteractionTouchable : BaseNearInteractionTouchable
```
luis-valverde-ms: It seems that with these changes all near interaction touchables must have a collider. Does it make sense to keep a separate ColliderNearInteractionTouchable class?

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnboundedPlane.cs
MenelvagorMilsom: I too am happy with the removal of this class. I was unsure how it should be distinct when adding it. As long as touchable are tied to colliders, functionally this is going to behave like `NearInteractionTouchable`. 
Currently, the only distinguishing features of this class is the distance function and cursor behaviour, but neither align with expectations as the collider is not infinite.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
```diff
@@ -0,0 +1,45 @@
...
+using UnityEngine;
+using UnityEngine.Serialization;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Use a Unity UI RectTransform as touchable surface.
+    /// </summary>
+    [RequireComponent(typeof(RectTransform))]
+    public class NearInteractionTouchableUnityUI : BaseNearInteractionTouchable
```
luis-valverde-ms: Doesn't this require a collider component too? If we require Unity UI to be wrapped in a collider for near interaction, we need to make sure that doesn't affect far interactions (i.e. interactions via raycast-based pointers).

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
MenelvagorMilsom: The touchable distance was made per touchable in #4263. These changes are moving it back.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -13,7 +13,28 @@ public class PokePointer : BaseControllerPointer, IMixedRealityNearPointer
 
         [SerializeField]
         protected GameObject visuals;
-        
+
+        /// <summary>
+        /// Maximum distance a which a touchable surface can be interacted with.
+        /// </summary>
+        [SerializeField]
+        protected float touchableDistance = 0.2f;
```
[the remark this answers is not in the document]
lukastoenneMS: I'm not clear what the sentence means either ... The touch system is convoluted and inconsistent TBH. I'll try to explain what happens:

_Reachable Objects_ are objects with a both a `BaseNearInteractionTouchable` and a collider within `TouchableDistance` from the poke pointer (based on [OverlapSphere](https://docs.unity3d.com/ScriptReference/Physics.OverlapSphere.html)).

If a poke pointer has no _active_ touchable (`PokePointer.currentTouchableObjectDown`), then it will try to select one from the Reachable Objects based on:
1. Layer mask priority: Lower-priority layer masks will only be considered if higher-priority layers don't contain any Reachable Objects.
1. Distance: the closest object in the highest priority layers is selected based on `BaseNearInteractionTouchable.DistanceToTouchable`
1. Poke Threshold: The object becomes active once it crosses the `PokeThreshold` distance (behind the surface). At this point the `OnTouchStarted` or `PointerDown` event is raised.

If  a poke pointer _does_  have an _active_ touchable it will not consider any other object, until the `DistanceToTouchable` exceeds the `DebounceThreshold` (in front of the surface). At this point the active object is cleared and the `OnTouchCompleted` or `PointerUp` event is raised.

I really don't know how to squeeze this complexity into a single comment, and rewriting the touch system is out of scope of this PR imho.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -13,7 +13,28 @@ public class PokePointer : BaseControllerPointer, IMixedRealityNearPointer
...
+        public float TouchableDistance => touchableDistance;
+
+        /// <summary>
+        /// The LayerMasks, in prioritized order, that are used to determine the touchable objects.
+        /// </summary>
+        public LayerMask[] PokeLayerMasks { get; private set; }
```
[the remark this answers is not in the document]
lukastoenneMS: Correct, only objects in one of the layers here will be considered. Any object in a higher-priority layer will trump any object in a lower priority layer, regardless of distance.

Note that these are layer masks, so a layer can have multiple priorities, but only the highest priority will matter. This system is copied from the PointingRaycastLayerMasks, although i'm not sure what the reasoning behind this complexity is - i just try to keep things consistent somewhat.

@luval-microsoft also would like to move this to the PokePointer itself.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -13,7 +13,28 @@ public class PokePointer : BaseControllerPointer, IMixedRealityNearPointer
...
+
+        /// <summary>
+        /// Specify whether queries for touchable surfaces hit triggers.
```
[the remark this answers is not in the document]
lukastoenneMS: It's a Unity setting for raycast/overlap/etc. that determines if trigger volumes (as opposed to physics colliders) are included. The default option is [UseGlobal](https://docs.unity3d.com/ScriptReference/QueryTriggerInteraction.html) which just means that it uses the global project setting for physics.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
lukastoenneMS: Background: Previously the BaseNearInteractionTouchable class was handling both the collider case and the collider-less case. The latter was only used for UnityUI case, so i've split the collider code off into a separate base class `ColliderNearInteractionTouchable` for simpler code.

So, no, the UnityUI case didn't and doesn't require a collider. There is a breaking change here: with the static list used previously the PokePointer would still have considered objects with UnityUI, but now they won't be included. However, even before this change the UnityUI case was quite pointless because it would never have generated events, since these would be raised through the FocusProvider with an additional raycast.

I can only assume that the UnityUI stuff is left over from HTK days and not functional. AFAIK we don't have any examples for it either.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionTouchable.cs
lukastoenneMS: See comment below about NearInteractionTouchableUnityUI. I would love to kick out the UnityUI class, at which point we could indeed merge the base classes too.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
MenelvagorMilsom: There are examples in HandInteractionExamples and HandInteractionUnityUIExample. They are functional and should be tested.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -13,7 +13,28 @@ public class PokePointer : BaseControllerPointer, IMixedRealityNearPointer
 
         [SerializeField]
         protected GameObject visuals;
-        
+
+        /// <summary>
+        /// Maximum distance a which a touchable surface can be interacted with.
```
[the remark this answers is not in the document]
lukastoenneMS: What's the difference?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
lukastoenneMS: This is the same default value as before, i.e. distFront on the BaseNearInteractionTouchable.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityPointerProfile.cs
```diff
@@ -30,6 +30,23 @@ public class MixedRealityPointerProfile : BaseMixedRealityProfile
...
+        /// </summary>
+        public LayerMask[] PokeRaycastLayerMasks => pokeRaycastLayerMasks;
+
+        [SerializeField]
+        [Tooltip("Maximum number of colliders that can be detected in a scene query.")]
+        private int sceneQueryBufferSize = 64;
+        /// <summary>
+        /// Maximum number of colliders that can be detected in a scene query.
+        /// </summary>
+        public int SceneQueryBufferSize => sceneQueryBufferSize;
```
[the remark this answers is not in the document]
lukastoenneMS: done.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
lukastoenneMS: I have no clue how the UnityUI thing works. The examples seem to be broken with this change, would be great to have an actual test indeed.

MenelvagorMilsom: Maybe check out `PokePointer.IsObjectPartOfTouchable`. It may shed some light on how the Unity UI stuff works.

luis-valverde-ms: I don't know the details but I know that `MixedRealityInputModule` translates our pointer events into the mouse pointer events that Unity UI understands (see `GetMousePointerEventDataForMrtkPointer()`).

lukastoenneMS: So it looks like what is happening right now is that the PokePointer in `OnPreSceneQuery` [sets up a ray step](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/e18e7078bde38454816a0d3d2bd7c5e54bc2ccc5/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs#L74), and the FocusProvider then [uses this to perform a graphics raycast](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/e18e7078bde38454816a0d3d2bd7c5e54bc2ccc5/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs#L1078). This goes into the internal Unity system for handling UI.

The problem now is that there is no equivalent to the Physics.OverlapSphere function for the graphics/event system in Unity. We would need this to filter UI elements for the poke pointer, otherwise we'd have to go back to a static list for the UnityUI touchable variants ...

lukastoenneMS: I can't find a solution to this. If nobody has any idea we might as well drop this PR.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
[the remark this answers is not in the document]
julenka: Thanks for that explanation. It was actually really helpful. Any reason we wouldn't want to have such a long comment in the code? I actually think it would be quite helpful for people trying to understand the touch input system.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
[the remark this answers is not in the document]
julenka: You could have for example a touchable volume instead of a touchable surface.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
julenka: I think we can do it, there's so many good things in this PR.

julenka: I think that dispatching of input in UnityUI is implemented in [MixedRealityInputModule (MRIM)](https://github.com/julenka/MixedRealityToolkit-Unity/blob/e5dac4afce15cc6719b6cd11b965b1dcb0919a6a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs#L302).

Here's what I think happens (based on my experience with [this PR](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4765/files). 

1. MRIM tracks all active pointers by listening to [sourcedetected](https://github.com/julenka/MixedRealityToolkit-Unity/blob/e5dac4afce15cc6719b6cd11b965b1dcb0919a6a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs#L302), sourcelost
2. MRIM implements [IMRPointerHandler](https://github.com/julenka/MixedRealityToolkit-Unity/blob/e5dac4afce15cc6719b6cd11b965b1dcb0919a6a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs#L11)
3. OnPointerDown, MRIM updates state of its tracked pointers.
4. In [Process()](https://github.com/julenka/MixedRealityToolkit-Unity/blob/e5dac4afce15cc6719b6cd11b965b1dcb0919a6a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs#L115), MRIM actually translates the pointer states it is tracking (pressed, not pressed, etc) into instructions for Unity to ProcessMouseEvent.
5. As part of ProcessMouseEvent, Unity calls GetMousePointerEventData, which calls [GetMousePointerEventDataForMRTKPointer](https://github.com/julenka/MixedRealityToolkit-Unity/blob/e5dac4afce15cc6719b6cd11b965b1dcb0919a6a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs#L188). 
6. That method is the thing that has the meat -- it takes the position and rotation of the pointer (actually, the rotations of Rays[0]) and converts that into a ray. This ray then gets used to figure out which UI element is hit. We don't need to worry about this luckily -- Unity does its own casting to figure out which sub element was hit.

In short, I think as long as you can [set up that ray step](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/e18e7078bde38454816a0d3d2bd7c5e54bc2ccc5/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs#L74) at some point so that the position of ray step matches some place in front of the surface, and the rotation points toward the surface (so that the ray will hit the Unity UI), you will be fine.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -46,31 +83,23 @@ public override void OnPreSceneQuery()
...
+                    break;
                 }
             }
 
             if (newClosestTouchable != null)
             {
                 // Build ray (poke from in front to the back of the pointer position)
-                Vector3 start = Position - newClosestTouchable.DistBack * -closestNormal;
-                Vector3 end = Position + newClosestTouchable.DistFront * -closestNormal;
+                Vector3 start = Position - newClosestTouchable.PokeThreshold * -closestNormal;
```
julenka: I actually think since you are not changing the RayStep here, that UnityUI is likely to work. You probably just need to add a UnityUITouchable component and set it up.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
julenka: Hm now reading Luis's comment above I feel like you and Luis probably already figured out what I said above...

julenka: Here's an idea: In FindClosestTouchableForLayerMask, could you just do a graphics raycast for UnityUI? Don't worry about a sphere cast, just do a raycast. Reasoning is this: Unity UI is always going to be a flat rectangle, so you will always touch them from the front. You could set the position of the ray to be finger tip, and the direction of the ray the normal of UnityUI surface.

So, first find the colliders from volumes using the sphere cast, then also look for any UnityUI by doing a graphics raycast, ignoring all surfaces farther than touchableDistance.

lukastoenneMS: The problem is in the first link of this chain: finding all potential touchables to examine (i.e. the "broad phase"). Without this no RayStep will be created and no graphics raycast happens.

Doing a graphics raycast isn't a good solution, because i don't know at that point in which direction to cast the ray. Just casting to "forward" (what does that mean for a poke pointer?) will probably miss ui elements and cause problems down the line.

I'm considering just adding back the static list for just the Unity canvas objects, so we can at least get a reliable behavior. Then perhaps we can try to get Unity to supply a proper overlap test for their UI elements.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
lukastoenneMS: I've tried that, and it doesn't work. As explained above, the problem is that we don't get here in the first place because the overlap test doesn't catch UI objects.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionTouchable.cs
lukastoenneMS: Never mind, UnityUI case seems to be too important to drop.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
julenka: In partner apps I've scene they usually just have one Unity UI canvas, maybe 2.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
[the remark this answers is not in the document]
lukastoenneMS: Fair enough, i'll add it somewhere prominent for people to find.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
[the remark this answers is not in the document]
lukastoenneMS: It's based on the virtual distance function of the touchable, to the PokePointer it makes no difference how its implemented. The explanation belongs into the implementation class to explain how that distance is computed.

Semantic nitpick: i would call it "planar surface" vs. "collider surface" instead of "volume".

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
lukastoenneMS: Doesn't really matter if it's 2 or 1000 though, we'd need some global refs to the canvases to examine. Could be automatic registry in a static list as before, or a simple input system pointer to the canvas to use, but the latter is even more error-prone.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -93,6 +122,34 @@ public override void OnPreSceneQuery()
...
+        private bool FindClosestTouchableForLayerMask(LayerMask layerMask, out ColliderNearInteractionTouchable closest, out float distance, out Vector3 normal)
+        {
+            closest = null;
+            distance = float.PositiveInfinity;
+            normal = Vector3.zero;
+
+            int numColliders = UnityEngine.Physics.OverlapSphereNonAlloc(Position, touchableDistance, queryBuffer, layerMask, triggerInteraction);
+            if (numColliders == queryBuffer.Length)
+            {
+                Debug.LogWarning("Maximum number of colliders found in PokePointer overlap query. Consider increasing the query buffer size in the pointer profile.");
```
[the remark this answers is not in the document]
lukastoenneMS: done

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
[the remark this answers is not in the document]
lukastoenneMS: I've added this as inline documentation on the PokePointer class now, with full xrefs to the mentioned properties and functions.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionTouchable.cs
```diff
@@ -29,57 +26,47 @@ public abstract class BaseNearInteractionTouchable : MonoBehaviour
...
-        /// The collider used by this touchable.
-        /// </summary>
-        [SerializeField]
-        [FormerlySerializedAs("collider")]
-        protected Collider touchableCollider;
-
-        protected float distFront = 0.2f;
-
-        [Tooltip("Distance behind the surface at which you will receive a touch up event")]
+        [Tooltip("Distance behind the surface at which you will receive a touch started event")]
```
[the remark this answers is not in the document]
lukastoenneMS: Added a bit more inline doc to explain these:

```csharp
        [Tooltip("Distance behind the surface at which you will receive a touch started event")]
        [SerializeField]
        [FormerlySerializedAs("distBack")]
        protected float pokeThreshold = 0.25f;
        /// <summary>
        /// Distance behind the surface at which the touchable becomes active.
        /// </summary>
        /// <remarks>
        /// When the pointer distance to the touchable object becomes less than -PokeThreshold (i.e. behind the surface),
        /// then the Touch Started event is raised and the touchable object becomes tracked by the pointer.
        /// </remarks>
        public float PokeThreshold => pokeThreshold;

        [Tooltip("Distance in front of the surface at which you will receive a touch completed event")]
        [SerializeField]
        protected float debounceThreshold = 0.01f;
        /// <summary>
        /// Distance in front of the surface at which you will receive a touch completed event.
        /// </summary>
        /// <remarks>
        /// When the touchable is active and the pointer distance becomes greater than +DebounceThreshold (i.e. in front of the surface),
        /// then the Touch Completed event is raised and the touchable object is released by the pointer.
        /// </remarks>
        public float DebounceThreshold => debounceThreshold;
```

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnboundedPlane.cs
lukastoenneMS: I have removed this class now.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -6,30 +6,98 @@
 
 namespace Microsoft.MixedReality.Toolkit.Input
 {
+    /// <summary>
```
julenka: Very helpful, thank you

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
julenka: I think static list for Unity UI is OK. It's definitely a case we want to still support (several partners depend on it).

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+using UnityEngine.Serialization;
+
+namespace Microsoft.MixedReality.Toolkit.Input
```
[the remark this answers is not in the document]
lukastoenneMS: Done. The Demos/HandTracking/Prefabs/TestCanvas.prefab is actually the only place where we have a Unity UI touchable.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityPointerProfile.cs
lukastoenneMS: Added \<remarks\> to both the PokePointer and BaseNearInteractionTouchable.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
lukastoenneMS: I've added back the static instances list for UnityUI touchables. It works fine in the examples, but would be great to have an automated test for this case too. I've tried creating a Canvas object and test components from script, but that didn't work for some reason (i'm not an expert on the UI system). Loading the prefab would be easier, but we haven't solved the question of test resources yet.

julenka: Have a look at #5033 for an example of how to load prefabs. It isn't ideal because if the prefab names change the tests will break. But it will at least unblock you. Maybe a dictionary of common test resources would be good, or a single class that would load test prefabs. This way there's just one place where the paths to prefabs used for tests exist. Something like :

Issue #4946, opened 2019-06-18
Normals Are Incorrect on Back Faces


Cameron-Micka:
## Describe the bug

When rendering the back faces of triangles, the normals are flipped the opposite way which can lead to artifacts in lighting calculations and other techniques which utilize normals.

## To reproduce

1. Create a MRTK/Standard shader material.
2. Disable back face culling.
3. Examine the lighting on a back face.

## Expected behavior

Normals are correct on front and back faces.

## Screenshots

<img width="1226" alt="Before" src="https://user-images.githubusercontent.com/13305729/59703059-f78b9300-91ad-11e9-8112-cca5d6b269be.PNG">

## Your Setup (please complete the following information)

- Unity Version 2018.4.1f1
- MRTK Version RC2.1

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

PR #4947, opened 2019-06-18
Fixing Normal Direction on Back Face Triangles

Cameron-Micka:
## Overview
When rendering the back faces of triangles, the normals are flipped the opposite way which can lead to artifacts in lighting calculations and other techniques which utilize normals. This change flips the normals based on which side of the triangle face is being rendered.

Compare

## Changes

- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4946

## Verification
To verify the fix look at lit surfaces with back face culling disabled in the Material Gallery.

>
>
>


ActiveNick:
@wiwei How much performance testing was done before approving this? Since this affects the standard shader used by virtually every single MRTK project, the implications are far reaching. I'd love to know which testbed we use to make sure we're not introducing performance issues.

Cameron-Micka:
@ActiveNick we have two test scenes we run though, the MaterialGallery and the StandardMaterialComparison scene. I also keep track of the average instruction count, branches, and texture fetches commonly used permutations generate. Due to the nature of it being an "uber shader" test coverage becomes difficult for all permutations due to the billions (maybe trillions now?) of variants possible.

PR #4965, opened 2019-06-19
Porting modified StablizationPlaneModifier from HoloToolkit

blgrossMS:
## Overview
HoloLens 1 and HoloLens 2 support SetFocusPoint which uses a plane to do late stage reprojection. For HoloLens 1 this is the best solution to get hologram stabilization, for HoloLens 2 this is a good solution if you:
- Are using Unity UI (which does not write depth)
- Do not want to use depth buffer sharing (potentially due to performance concerns)
- An interim solution while working to make your application properly write to the depth buffer

There are five modes that are supported in this solution:
1) Fixed- Sets a fixed plane at a user specified distance along the users gaze vector. 2m is the actual focus of the holograms which is why it is chosen as a default value.
2) Gaze Hit- Focus plane is set based on the hit position of the gaze ray and is oriented along the users gaze vector. 
3) Plane Override- Focus plane is based on where an override plane specifies it should be. This can be programmatically updated everyone frame. This can be especially useful for planar UI elements, and will make them appear more stable
4) Target Override- Focus plane position is based on a transform and oriented towards the user. This is useful for if the user is looking at an object, or an object that is moving. 
5) Off- This should be set if you want to use the default system behavior, or you want to switch at runtime to depth LSR 

There is one more mode that could be added on HoloLens 2, which is an eye gaze based approach. Fundamentally, Planar LSR is about trying to determine user dwell to stabilize at that location. By implementing a eye gaze hit it will increase the likelihood that the plane is at the correct location. 

## Changes
- From v1 implementation the plane override and target override modes have been added
- Instead of a series of Booleans, the logic has been made clearer with an enum
- Gizmos logic has been adjusted to render plane correctly
- Added better lerp in order to decrease jitter
- Removed Singleton

## Verification
Reviewers should ensure that I am following the proper coding style for MRTK, verify behavior, and ensure that file is in the correct location


Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+
+using Microsoft.MixedReality.Toolkit.Input;
+using System;
+using UnityEngine;
+using UnityEngine.XR.WSA;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities
+{
+    [Serializable]
+    public struct StabilizationPlaneOverride
```
wiwei: >StabilizationPlaneOverride [](start = 18, length = 26)

Nit: Please add a doccomment for this class and its fields.

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+    public class StabilizationPlaneModifier : MonoBehaviour
+    {
+        [System.Serializable]
+        public enum StabilizationPlaneMode
+        {
+            Fixed,
+            TargetOverride,
+            PlaneOverride,
+            GazeHit,
+            Off
```
wiwei: Nit: It would be good to have comments here for what they mean, so that for code readers/programmatic access folks, they'll have it in their autocomplete/IDE. I think that you have already written these things out in the tooltip below and also in the PR description, so copying those here should suffice

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+        public StabilizationPlaneMode Mode;
+
+        [Tooltip("When lerping, use unscaled time. This is useful for apps that have a pause mechanism or otherwise adjust the game timescale.")]
+        public bool UseUnscaledTime = true;
+
+        [Tooltip("Lerp speed when moving focus point closer.")]
+        public float LerpStabilizationPlanePowerCloser = 4.0f;
+
+        [Tooltip("Lerp speed when moving focus point farther away.")]
+        public float LerpStabilizationPlanePowerFarther = 7.0f;
```
wiwei: Nit: Here's a weird thing that I don't necessarily agree with but.... in order to be consistent with our other code, I think that we've moved away from having public fields, in favor of having private fields (like targetOverride) and public getters and setters.

I don't necessarily agree with this overall, but I would like for us to be consistent so absent us taking a large uturn on that policy I think it would be good to remain consistent with it.

So specifically what I'm asking for here is to make these private, add a public getter/setter, and then add the SerializeField property on the newly privated fields (so that they can keep getting serialized)

Sorry about this

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+    /// StabilizationPlaneModifier handles the setting of the stabilization plane in several ways.
+    /// Using StabilizationPlaneModifier with override DepthLSR which is automatically enabled via depth buffer sharing
+    /// StabilizationPlaneModifier is recommended for HoloLens 1, can be used for HoloLens 2, and does a no op for WMR
+    /// </summary>
+    public class StabilizationPlaneModifier : MonoBehaviour
+    {
+        [System.Serializable]
+        public enum StabilizationPlaneMode
+        {
+            Fixed,
```
wiwei: >Fixed [](start = 12, length = 5)

[https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum)

(In particular: DO provide a value of zero on simple enums.)

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+        private StabilizationPlaneOverride debugPlane;
+
+        /// <summary>
+        /// Used for representing the debug mesh
+        /// </summary>
+        private GameObject debugMesh;
+
+        private void Awake()
+        {
+#if UNITY_EDITOR
```
wiwei: Nit: Any chance this could wrap the entire function (i.e. so at non-editor case Awake() never gets invoked)?

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+        }
+
+        /// <summary>
+        /// Gets the origin of the gaze for purposes of placing the stabilization plane
+        /// </summary>
+        private Vector3 GazeOrigin
+        {
+            get
+            {
+                if (InputSystem.GazeProvider.Enabled)
```
wiwei: >InputSystem [](start = 20, length = 11)

Note that it's possible the InputSystem could be null if it hasn't initialized yet - this is a separate monobehavior that initializes separate from the input system itself, so depending on timing it could work out poorly in the first frame or work out just fine.

The simplest thing to avoid this would be to add a InputSystem?.GazeProvider...

wiwei: Same comments for the stuff down below.

---
In reply to: [295401779](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4965#discussion_r295401779) [](ancestors = 295401779)

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+            currentPlaneDistance = Mathf.Lerp(currentPlaneDistance, focusPointDistance, lerpPower * deltaTime);
+            gazeToPlane.Normalize();
+            planePosition = gazeOrigin + (gazeToPlane * currentPlaneDistance);
+
+#if UNITY_EDITOR
+            debugPlane.Center = planePosition;
+            debugPlane.Normal = OverridePlane.Normal;
+#elif UNITY_WSA
+            // Place the plane at the desired depth in front of the user and billboard it to the gaze origin.
+            HolographicSettings.SetFocusPointForFrame(planePosition, OverridePlane.Normal, velocity);
```
wiwei: Nit: Consider moving this code to a common helper (i.e. this code + the code in ConfigureTransformOverridePlane have a bunch of shared stuff that can probably be moved into a private helper)

wiwei:
Overall looks good to me. Also lemme know if you want me to just make these changes (i.e. give me write to the fork's branch and I can also patch things up).

wiwei:
One more thing to mention here, which is that we were wanting to move these to an experimental namespace within the MRTK itself, until we have proper test/documentation coverage. It's fine to go in without making this change and I can handle moving it around to the right place afterward

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+        }
+
+        [Tooltip("Choose mode for stabilization plane.\n1) Fixed- Submits plane at a fixed distance based on DefaultPlaneDistance field along the users gaze.\n2) Gaze Hit- Submits plane along the users gaze at the position of gaze hit.\n3) Target Override- Submits the plane at a fixed position along the users gaze based on the TargetOverride property.\n4) Plane Override- Submits the plane based on the OverridePlane property.\n5) Off- Does not call SetFocusPoint")]
+        public StabilizationPlaneMode Mode;
+
+        [Tooltip("When lerping, use unscaled time. This is useful for apps that have a pause mechanism or otherwise adjust the game timescale.")]
+        public bool UseUnscaledTime = true;
+
+        [Tooltip("Lerp speed when moving focus point closer.")]
+        public float LerpStabilizationPlanePowerCloser = 4.0f;
```
Troy-Ferrell: Maybe simpler variable name like LerpPowerPlaneCloser, still a mouthful but a bit shorter. Same with farther below

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+                    MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem);
+                }
+                return inputSystem;
+            }
+        }
+
+        /// <summary>
+        /// Position of the plane in world space.
+        /// </summary>
+        private Vector3 planePosition;
```
Troy-Ferrell: nit: Worth maybe adding get to this? Like what is the current focus point?

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+
+        /// <summary>
+        /// Used for representing the debug mesh
+        /// </summary>
+        private GameObject debugMesh;
+
+        private void Awake()
+        {
+#if UNITY_EDITOR
+            debugMesh = GameObject.CreatePrimitive(PrimitiveType.Quad);
```
Troy-Ferrell: Since this is not active whole time, add hideflag to hide in hierarchy?
https://docs.unity3d.com/ScriptReference/HideFlags.html

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+                default:
+                    break;
+            }
+        }
+
+        /// <summary>
+        /// Called by Unity when this script is loaded or a value is changed in the inspector.
+        /// Only called in editor, ensures that the property values always match the corresponding member variables.
+        /// </summary>
+        private void OnValidate()
```
carmines: remove OnValidate, for any initialization/validation of values for a Monobehavior should be done in the Awake, OnEnable, or Start functions.

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+        /// </summary>
+        private Vector3 GazeOrigin
+        {
+            get
+            {
+                if (InputSystem.GazeProvider.Enabled)
+                {
+                    return InputSystem.GazeProvider.GazeOrigin;
+                }
+                return Camera.main.transform.position;
```
Troy-Ferrell: CameraCache?

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+            // Place the plane at the desired depth in front of the user and billboard it to the gaze origin.
+            HolographicSettings.SetFocusPointForFrame(planePosition, -gazeDirection, velocity);
+#endif
+        }
+
+        private void ConfigureOverridePlane(float deltaTime)
+        {
+            planePosition = OverridePlane.Center;
+
+            Vector3 velocity = Vector3.zero;
```
Troy-Ferrell: nit: simpler velocity = TrackVelocity ? UpdateVelocity(deltaTime) : Vector3.zero;?

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+    [Serializable]
+    public struct StabilizationPlaneOverride
+    {
+        public Vector3 Center;
+        public Vector3 Normal;
+    }
+
+    /// <summary>
+    /// StabilizationPlaneModifier handles the setting of the stabilization plane in several ways.
+    /// Using StabilizationPlaneModifier with override DepthLSR which is automatically enabled via depth buffer sharing
```
Troy-Ferrell: Maybe call out SetFocusPointForFrame() api since that is heart of reason for this component. Also re-write this line to Stabilization PlaneModifer can be used in conjunction with DepthLSR which is enabled by Depth Buffer Sharing etc 

 Although on HL1 I think the stabilization plane can be set in RS5 via the depth buffer sharing and I'm not sure which one takes precedence...

Troy-Ferrell:
Awesome stuff! Thanks for getting this in

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
Troy-Ferrell: We have had so many problems with onvalidate in past

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
davidkline-ms: We use these comments to generate API documentation. Agree they should be added.

Thanks!

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
```diff
@@ -0,0 +1,390 @@
...
+
+        /// <summary>
+        /// When in editor, draws a magenta quad that visually represents the stabilization plane.
+        /// </summary>
+        private void OnDrawGizmos()
+        {
+            if (Application.isPlaying && DrawGizmos && Mode != StabilizationPlaneMode.Off)
+            {
+                Gizmos.color = Color.green;
+                Gizmos.DrawWireMesh(debugMesh.GetComponent<MeshFilter>().sharedMesh, debugPlane.Center, Quaternion.LookRotation(debugPlane.Normal));
```
Troy-Ferrell: Cache GetComponent<MeshFilter> instead of accessing every draw

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
blgrossMS: SetFocusPoint takes precedence over depth buffer sharing.

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
blgrossMS: good feedback

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
blgrossMS: I think I won't add this for now since ideally applications wouldn't be trying to see the stabilization plane position at runtime

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
blgrossMS: @wiwei Can you clarify the ask here? I'm confused by what you mean

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
blgrossMS: I think the code here is more readable

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
blgrossMS: done

Assets/MixedRealityToolkit.SDK/Features/Utilities/StabilizationPlaneModifier.cs
wiwei: Ah I think that my comments don't make sense anymore - the version I looked at had every line of code in Awake #IF-ed behind UNITY_EDITOR (which made me think that we could just cover the entire function around it). With the latest, I see that there's stuff happening outside the editor as well, so this isn't relevant.

wiwei:
I'll handle getting this in - the queue is a bit backed up (and I'm waiting for a particularly critical one to go before any other ones go in)

wiwei:
Added a commit that fixed some build warnings (a recent change made it so that we treat warnings as errors).

wiwei:
I think at this point all of the warnings should be resolved - just waiting on the latest CI run to validate

Issue #4847, opened 2019-06-12
Clean up usages of old IMixedRealityEventSystem API (Register/Unregister)


googlan:
## Overview
This task is a follow up for PR #4716  and related to task #4187 .
After API improvement done in PR #4716 , Register/Unregister methods of IMixedRealityEventSystem should be marked as obsolete and, eventually, removed. To do this, references to these methods in Mrtk code base should be replaced with new RegisterHandler/UnregisterHandler.
`InputSystemGlobalListener` should remain in-place for backward compatibility and replaced in Mrtk code base with similar component, which uses registerHandler/unregisterHandler (or not replaced, but removed completely).

Need to cache check of handler's parent object being in EventListeners collection (in `HandleEvent`).
Need to measure performance of `HandleEvent`.

PR #4970, opened 2019-06-19
Cleanup usages of old input system API (Register/Unregister)

googlan:
## Overview
Follow-up to change of `IMixedRealityEventSystem` API in PR #4716.  Corresponds to task #4847.

## Changes
* Removes usage of `Register(gameObject)` and `Unregister(gameObject)` from mrtk code base. Marks both as Obsolete.
* Removes inheritance on `InputSystemGlobalListener` and marks it as obsolete.
* Adds alternative helper component `InputSystemGlobalHandlerListener`, which behaves same way as the old one, but through the new API. Use this new class as a base class, where applicable.
* Moves check for using old and new API in a conflicting way to moment of registration. Emit error, if components on the same object use old and new API at the same time. This is a follow-up to comments on #4716 .
* Adds new tests for input system and new input handler base component.


Assets/MixedRealityToolkit/Interfaces/EventSystem/IMixedRealityEventSystem.cs
```diff
@@ -16,9 +16,9 @@ public interface IMixedRealityEventSystem : IMixedRealityService
         /// <summary>
         /// List of event listeners that are registered to this Event System.
         /// </summary>
-        // Marking EventListeners as obsolete triggers a lot of warnings in MRTK code base.
-        // This should be commented out once follow up task #4847 is complete.
-        //[Obsolete("EventListeners is replaced by EventHandlersByType and will be removed in a future release.")]*/
+        /// <remarks>
+        /// This collection is obsolete is replaced by handler-based storage and will be removed in a future release.
```
wiwei: > handler-based storag [](start = 54, length = 21)

Nit: It would be helpful to point to the specific version that is recommended instead (i.e, RegisterHandler). I think that the autocomplete tooltip should show that, so hopefully it'll be one less click/step to find the replacement.

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -104,25 +126,37 @@ private enum Action
         /// <inheritdoc />
         public virtual void RegisterHandler<T>(IEventSystemHandler handler) where T : IEventSystemHandler
         {
+            if(handler == null)
```
wiwei:
```suggestion
            if (handler == null)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -104,25 +126,37 @@ private enum Action
...
             #endif
+            Debug.Assert(typeof(T).IsAssignableFrom(handler.GetType()), "Handler passed to RegisterHandler doesn't implement a type given as generic parameter.");
 
             TraverseEventSystemHandlerHierarchy<T>(handler, RegisterHandler);
         }
 
         /// <inheritdoc />
         public virtual void UnregisterHandler<T>(IEventSystemHandler handler) where T : IEventSystemHandler
         {
+            if(handler == null)
```
wiwei:
```suggestion
            if (handler == null)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -137,6 +171,35 @@ public virtual void Register(GameObject listener)
...
+                    bool report = false;
+                    foreach (var typeEntry in EventHandlersByType)
+                    {
+                        for (int index = 0; index < typeEntry.Value.Count; index++)
+                        {
+                            var handlerEntry = typeEntry.Value[index];
+
+                            var comp = handlerEntry.handler as Component;
+
+                            if (comp && comp.gameObject == listener)
```
wiwei:
```suggestion
                            if (comp != null && comp.gameObject == listener)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -137,6 +171,35 @@ public virtual void Register(GameObject listener)
...
+                            {
+                                handlerEntry.parentObjectIsInObjectCollection = true;
+                                typeEntry.Value[index] = handlerEntry;
+                                report = true;
+                            }
+                        }
+                    }
+
+                    if(report)
```
wiwei:
```suggestion
                    if (report)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -153,6 +216,25 @@ public virtual void Unregister(GameObject listener)
             {
                 if (EventListeners.Contains(listener))
                 {
+                    // Reset cached flags in handler collection as object will not intercept the events anymore.
+                    // This is a slow loop, which is here to maintain backward compatibility and enable co-existing of
+                    // new and old API.
+                    foreach(var typeEntry in EventHandlersByType)
```
wiwei:
```suggestion
                    foreach (var typeEntry in EventHandlersByType)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -153,6 +216,25 @@ public virtual void Unregister(GameObject listener)
...
+                    {
+                        for(int index = 0; index < typeEntry.Value.Count; index++)
```
wiwei:
```suggestion
                        for (int index = 0; index < typeEntry.Value.Count; index++)
```

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -192,34 +274,42 @@ private void RegisterHandler(Type handlerType, IEventSystemHandler handler)
 
         private void AddHandlerToMap(Type handlerType, IEventSystemHandler handler)
         {
-            List<IEventSystemHandler> handlers;
+            List<EventHandlerEntry> handlers;
+            bool isParentObjectRegistered = false;
+
+            var componentHandler = handler as Component;
+            if (componentHandler && EventListeners.Contains(componentHandler.gameObject))
```
wiwei:
```suggestion
            if (componentHandler != null && EventListeners.Contains(componentHandler.gameObject))
```

Assets/MixedRealityToolkit.Examples/Demos/UX/BoundingBox/Scripts/BoundingBoxTest.cs
```diff
@@ -24,10 +24,36 @@ public class BoundingBoxTest : InputSystemGlobalListener, IMixedRealitySpeechHan
...
+        }
+
+        protected virtual void OnDisable()
+        {
+            InputSystem.UnregisterHandler<IMixedRealitySpeechHandler>(this);
+        }
+
         // Start is called before the first frame update
-        protected override void Start()
+        protected virtual void Start()
```
wiwei: >Start [](start = 31, length = 5)

One thing to be aware of is that the InputSystemGlobalListener.Start() also did this:

[https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs#L45](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalListener.cs#L45)

Without this it's possible that input system could be not valid that this this point.

BUTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTT

If you notice something interesting, this override was actually not marked async, so TECHNICALLY there was always a bug here where it wouldn't wait for the base class's await, so actually probably this is just fine.

Just wanted to call this out this as FYI, not because I think it's actually a real thing.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        protected virtual void OnEnable()
+        {
+            if (InputSystem != null && !lateInitialize)
+            {
+                RegisterHandlers();
+            }
+        }
+
+        protected virtual async void Start()
```
wiwei: >void [](start = 32, length = 4)

As long as we're adding a new file, it might be worth considering the definition here (or the structure via which sub-classes end up consuming this).

In particular, because this is async void, subclasses actually won't be able to await base.Start() because Start is void.

Then again, it's not really possible to mark this as async Task, because then it wouldn't match Unity's Start() definition (I don't think it will get called at that point - I'm not 100% on this part).

The only REAL thing woudl be to have like this other virtual AwaitingStart() function that looks like:

protected virtual async void Start()
{
   ... start stuff...
   await AwaitingStart();
}

protected virtual async Task AwaitingStart()
{
}

The other possibility is to actually make this sealed or not extendable (but leave a comment saying it would be possible by doing what is mentioned above)

I guess the main point here is that there was a deficiency in the old thing which, as long as we're adding new thing, we may not want to choose to carry over.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        /// <summary>
+        /// Utility method, which unregisters all handlers implemented by 'this' using Reflection.
+        /// T should be a derived class type
+        /// Avoid calling this method unless class implements so many interfaces that enumerating them manually becomes unclear for the reader
+        /// </summary>
+        protected void UnregisterAllHandlers<T>() where T: class, IEventSystemHandler
+        {
+            if(InputSystem == null)
```
wiwei:
```suggestion
            if (InputSystem == null)
```

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// <summary>
     /// Waits for a controller to be initialized, then synchronizes its transform position to a specified handedness.
     /// </summary>
-    public class ControllerPoseSynchronizer : InputSystemGlobalListener, IMixedRealityControllerPoseSynchronizer
+    public class ControllerPoseSynchronizer : InputSystemGlobalHandlerListener, IMixedRealityControllerPoseSynchronizer
```
wiwei: >InputSystemGlobalHandlerListener [](start = 46, length = 32)

For my own understanding, is the main value in going with InputSystemGlobalHandlerListener here (instead of just being a MonoBehavior) is how it will wait for the input system to be valid before starting up? Or is there some other aspect here of non-focus/gaze related stuff?

wiwei:
Change looks good to me, I think that I'd want someone else (i.e. @keveleigh or @davidkline-ms) to take a look to

Assets/MixedRealityToolkit.Examples/Demos/UX/BoundingBox/Scripts/BoundingBoxTest.cs
googlan: We also had a huge number of other components which didn't do registration in 'Start'. I think it is important in core components, which are added by mrtk itself, but shouldn't matter at all for user components.

---
In reply to: [295369799](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4970#discussion_r295369799) [](ancestors = 295369799)

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
googlan: Do we actually have a scenario where we'd want a delayed registration in `Start` and not just fail fast in `enable`, if input system is not present? 

---
In reply to: [295375494](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4970#discussion_r295375494) [](ancestors = 295375494)

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs
googlan: 3 things:
* hides waiting for input system in `Start`.
* InputSystem property was used in most of derived classes, and I didn't want to copy-paste it everywhere.
* additional utility method to register all handlers for a component via reflection (used in only one place for the moment, but it will not look nice without it)

---
In reply to: [295381267](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4970#discussion_r295381267) [](ancestors = 295381267)

googlan: Actually I misread the class what class the question was for..
Just first 2 reasons for this one.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
davidkline-ms: >Do we actually have a scenario where we'd want a delayed registration in Start and not just fail fast in enable, if input system is not present?

By that argument, should the base class Start need to be async at all? :)

googlan: My question was more into "do we actually need to wait until `Start`, do we have a case when input system is not available in `OnEnable`? Do we want such cases to be supported at all if they happen?". I haven't seen all parts of the system yet, but it looks like we don't. CC @davidkline-ms @wiwei

wiwei: I don't have enough background to say authoritatively.

@davidkline-ms, it has to be async because there is an await there.

I think my thought here is, if we just make this private, we can sidestep this question until someone needs it, at which point we can do the right thing.

wiwei: Right thing being, do some virtual task-returning thingy

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Threading.Tasks;
+using UnityEngine;
+using UnityEngine.EventSystems;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// This component ensures that input events are forwarded to this component when focus or gaze is not required.
+    /// </summary>
+    public abstract class InputSystemGlobalHandlerListener : MonoBehaviour
```
luis-valverde-ms: I'd rather not have this class and let users do like we do in DefaultPrimaryPointerSelector:
```
        public void Initialize()
        {
            IMixedRealityInputSystem inputSystem;
            if (MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem))
            {
                inputSystem.RegisterHandler<IMixedRealityPointerHandler>(this);
            }
        }
```

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// Use UnregisterHandler API of InputSystem
+        /// </summary>
+        protected abstract void UnregisterHandlers();
+
+        /// <summary>
+        /// Utility method, which registers all handlers implemented by 'this' using Reflection.
+        /// T should be a derived class type
+        /// Avoid using this method unless class implements so many interfaces that enumerating them manually becomes unclear for the reader
+        /// </summary>
+        protected void RegisterAllHandlers<T>() where T: class, IEventSystemHandler
```
luis-valverde-ms: I don't think it's worth having this utility. For me what the user gains is not worth the added maintenance burden.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
googlan: I wanted an easy way to migrate users to new API, if they used the old one -- just change inheritance to the new one, and it works. 
Also it hides async logic (which might go away) -- but it might go away.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
googlan: Maybe you're right. It's only used in GGVPointer implementation now. One place is not worth it.

googlan: Removed both. Don't know why I thought adding them was a good idea in the first place. I guess I just wanted to call a generic method via reflection.

luis-valverde-ms: I totally get that

Assets/MixedRealityToolkit.Examples/Demos/UX/BoundingBox/Scripts/BoundingBoxTest.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.MixedReality.Toolkit.UI
 {
-    public class BoundingBoxTest : InputSystemGlobalListener, IMixedRealitySpeechHandler
```
keveleigh: Might be interesting to make `InputSystemGlobalListener<T>` to continue helping devs make global listeners, instead of needing to manually register and unregister.

Issue #4857, opened 2019-06-12
Multiple click events raised from a single in-Editor click


PJBowron:
## Describe the bug

Using default MRTK configuration, in at least some situations a single click within the Unity editor (performed by holding left-shift and clicking the mouse on a click-responsive control) causes three click events to be raised within the input system, in my case on the _Left_PokePointer_, the _Left_DefaultControllerPointer_ and the _Left_GrabPointer_.

## To reproduce

Set up a clickable surface (e.g. a Unity button) and add a script that implements IMixedRealityPointerHandler. In play mode, hold left-shift, point the controller ray at the button and click the left mouse button.
Break execution in the click handler and trace through to see repeated events.

## Expected behavior

A single click on the part of the user should only raise a single Click event from the pointer deemed active.

## Your Setup

- Unity 2018.4.1f1
- MRTK RC1 Refresh

## Target Platform

- HoloLens 1/2 (n/a, as behaviour is in-Editor)

## Additional context

I'm guessing that reproduction may be inconsistent due to the interactable components used (e.g. graphic raycaster vs physics) and their structuring, though once you have it, repro is 100%. In my case the behaviour was highlighted because the clicked control activates new UI including a fallback listener that would immediately fire and close the UI up again, all in the same frame. My first control (a simple IMixedRealityPointerHandler implementer) fired on the DefaultControllerPointer's click. My fallback listener fired on the GrabPointer's click.

May relate to #3652

radicalad:
Looked into this bug while testing #3652 - I'm not reproing with a standard `IMixedRealityPointerHandler` on a Unity Button - can you share the script or project you're referencing in Additional context?

PJBowron:
Can't share that project, unfortunately, but I'll try to put together a minimum viable repro in a fresh one.

PJBowron:
Here's a unitypackage that demonstrates the issue.
Note that I didn't include the MRTK in order to keep the filesize manageable, therefore to put this together:
1. Start a fresh Unity project
2. Import MRTK RC1 Refresh
3. Import the attached package

You'll see a modification to MixedRealityInputSystem.cs will come in on step 3; this is simply a debug message that counts the raised pointer clicks - feel free to exclude it from the import.

[MrtkRC1RMultipleClicks.zip](https://github.com/microsoft/MixedRealityToolkit-Unity/files/3296386/MrtkRC1RMultipleClicks.zip)

radicalad:
[...]

PR #4974, opened 2019-06-19
Don't fire pointer events if IsInteractionEnabled == false

radicalad:
## Overview
The bug listed in #4857 was caused by a missed check `BaseControllerPointer`:
- Pointer Up/down events were fired without checking `IsInteractionEnabled`
- This caused Pointer Up event to be fired by the GrabPointer, even when it was disabled. The repro in #4857 had a modal dialog that would show when a button B was clicked. The modal would then disappear whenever it saw a click form any pointer. The bug was that after user pressed select on controller, Bwould receive click from line pointer, which would show modal M. Then the grab pointer (from that same select gesture) would fire a second click, which would hide the modal M.

## Changes
- Fixes: #4857 
- Also remove extra 0.05 m buffer when doing test for is sphere collider near grabbable. This was causing confusing UX feedback where the line pointer would disappear (therefore communicating that object is close enough to grab), but object was not actually grabbable.

## Note
This has been successfully tested both on device and in editor, doesn't break existing components such as `ManipulationHandler`.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
@@ -32,7 +32,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 Vector3 position;
                 if (TryGetNearGraspPoint(out position))
                 {
-                    return UnityEngine.Physics.CheckSphere(position, SphereCastRadius + 0.05f, ~UnityEngine.Physics.IgnoreRaycastLayer);
+                    return UnityEngine.Physics.CheckSphere(position, SphereCastRadius, ~UnityEngine.Physics.IgnoreRaycastLayer);
                 }
 
                 return false;
--- a/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
@@ -30,6 +30,8 @@ namespace Microsoft.MixedReality.Toolkit.Input
         private readonly Dictionary<uint, IMixedRealityPointerMediator> pointerMediators = new Dictionary<uint, IMixedRealityPointerMediator>();
         private PointerHitResult hitResult3d = new PointerHitResult();
         private PointerHitResult hitResultUi = new PointerHitResult();
+        private Collider[] colliderQueryBuffer;
+        private int sceneQueryBufferSize = 64;
 
         public IReadOnlyDictionary<uint, IMixedRealityPointerMediator> PointerMediators
         {
@@ -484,6 +486,8 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 FindOrCreateUiRaycastCamera();
             }
 
+            colliderQueryBuffer = new Collider[sceneQueryBufferSize];
+
             var primaryPointerSelectorType = InputSystem?.InputSystemProfile.PointerProfile.PrimaryPointerSelector.Type;
             if (primaryPointerSelectorType != null)
             {
@@ -866,7 +870,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
                     // Perform raycast to determine focused object
                     var raycastProvider = InputSystem.RaycastProvider;
                     hitResult3d.Clear();
-                    QueryScene(pointer.Pointer, raycastProvider, prioritizedLayerMasks, hitResult3d);
+                    QueryScene(pointer.Pointer, raycastProvider, prioritizedLayerMasks, hitResult3d, colliderQueryBuffer);
                     PointerHitResult hit = hitResult3d;
 
                     // If we have a unity event system, perform graphics raycasts as well to support Unity UI interactions
@@ -978,7 +982,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// </summary>
         /// <param name="pointerData"></param>
         /// <param name="prioritizedLayerMasks"></param>
-        private static void QueryScene(IMixedRealityPointer pointer, IMixedRealityRaycastProvider raycastProvider, LayerMask[] prioritizedLayerMasks, PointerHitResult hit)
+        private static void QueryScene(IMixedRealityPointer pointer, IMixedRealityRaycastProvider raycastProvider, LayerMask[] prioritizedLayerMasks, PointerHitResult hit, Collider[] colliderBuffer = null)
         {
             float rayStartDistance = 0;
             MixedRealityRaycastHit hitInfo;
@@ -1019,17 +1023,30 @@ namespace Microsoft.MixedReality.Toolkit.Input
                         }
                         break;
                     case SceneQueryType.SphereOverlap:
-                        Collider[] colliders = UnityEngine.Physics.OverlapSphere(pointer.Rays[i].Origin, pointer.SphereCastRadius, ~UnityEngine.Physics.IgnoreRaycastLayer);
 
-                        if (colliders.Length > 0)
+                        int numColliders;
+
+                        if (colliderBuffer != null)
+                        {
+                            numColliders = UnityEngine.Physics.OverlapSphereNonAlloc(pointer.Rays[i].Origin, pointer.SphereCastRadius, colliderBuffer, ~UnityEngine.Physics.IgnoreRaycastLayer);
+                        }
+                        else
+                        {
+                            colliderBuffer = UnityEngine.Physics.OverlapSphere(pointer.Rays[i].Origin, pointer.SphereCastRadius, ~UnityEngine.Physics.IgnoreRaycastLayer);
+                            numColliders = colliderBuffer.Length;
+                        }                        
+
+                        if (colliderBuffer.Length > 0)
                         {
                             Vector3 testPoint = pointer.Rays[i].Origin;
                             GameObject closest = null;
                             float closestDistance = Mathf.Infinity;
                             Vector3 objectHitPoint = testPoint;
 
-                            foreach (Collider collider in colliders)
+                            for (int c = 0; c < numColliders; c++)
                             {
+                                Collider collider = colliderBuffer[c];
+
                                 // Policy: in order for an collider to be near interactable it must have
                                 // a NearInteractionGrabbable component on it.
                                 // FIXME: This is assuming only the grab pointer is using SceneQueryType.SphereOverlap,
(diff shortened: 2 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
```diff
@@ -49,9 +44,28 @@ public override void OnPreSceneQuery()
...
+
             if (TryGetNearGraspPoint(out pointerPosition))
             {
+                closestDistance = float.PositiveInfinity;
+                foreach (BaseNearInteractionGrabbable prox in BaseNearInteractionGrabbable.Instances)
+                {
+                    if (prox.ColliderEnabled)
+                    {
+                        float dist = prox.DistanceToGrabbable(pointerPosition);
+                        if (dist < SphereCastRadius + 0.05f && dist < closestDistance)
```
lukastoenneMS: I guess it's some kind of margin? Does it make sense to make it configurable? (i know you didn't add it, but any idea what this is for?)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
```diff
@@ -141,5 +155,82 @@ private void OnDestroy()
                 Destroy(debugSphere.gameObject);
             }
         }
+
+        private static bool IsObjectPartOfGrabbable(GameObject targetObject, BaseNearInteractionGrabbable grabbable)
+        {
+            return targetObject != null && grabbable != null &&
+                (targetObject == grabbable.gameObject ||
+                // Descendant game objects are grabbable as well.
+                (targetObject.transform != null && grabbable.gameObject.transform != null &&
```
lukastoenneMS: Can GameObject.transform ever become null? We don't expect this to happen anywhere else at least. Removing those would make this check a little less verbose.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
```diff
@@ -141,5 +155,82 @@ private void OnDestroy()
...
+            {
+                if (currentGrabbableObjectDown == null)
+                {
+                    // In order to get reliable up/down event behavior, only allow the closest grabbable to be grabbed.
+                    if (IsObjectPartOfGrabbable(Result.CurrentPointerTarget, closestProximityGrabbable))
+                    {
+                        currentGrabbableObjectDown = Result.CurrentPointerTarget;
+
+                        //For now we're only generating pointer events
+                        if (Application.isEditor)
```
lukastoenneMS: This is not a reliable solution, we should fix the input simulation instead if there is a discrepancy with hand tracking. I'm trying to understand how these actions work, but don't have a good understanding yet.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionGrabbable.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// Base class for all NearInteractionGrabbables
+    ///
+    /// Technical details:
+    /// Provides a listing of near field grab proximity bounds.
+    /// This is used to detect if a contact point is near an object to turn on near field interactions
+    /// </summary>
+
+    public abstract class BaseNearInteractionGrabbable : MonoBehaviour
+    {
+        public static IReadOnlyCollection<BaseNearInteractionGrabbable> Instances => instances.AsReadOnly();
```
lukastoenneMS: This should use an overlap test similar to the one implemented in #4942. We're still trying to figure out how to make that work with UnityUI objects though.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
radicalad: I'm assuming it exists to bump up the inaccuracy of someone targeting an object to pick it up - i.e. they may mis-grab slightly. That said, SphereCastRadius could just be 0.05f larger 😄

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
radicalad: I grabbed this from PokePointer - IMO its redundant. I can clean it up.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
radicalad: This is a weird one, the GrabPointer is set to "Grip Press" - which on device works great. In editor, the "pinch" gesture only fires "Select". I don't know enough about input simulation to make that fix so I just stuck this in here. PokePointer doesn't even use the base.OnInputDown call so I figured it was okay.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionGrabbable.cs
radicalad: Thankfully we don't have to worry about UnityUI here since this is for grabbing objects. I'll take a look.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
radicalad: I removed `gameObject` since `grabbable.transform` is the same thing. Depending on garbage collection, a unity object can be null, but still exist so the component could still exist as well.

radicalad:
Updated to match #4942 's overlap test pattern. Tested - works as expected.

julenka:
@lukastoenneMS FYI, @radicalad  and I just had a great chat about this PR and we discovered a much simpler solution. Expect some changes from @radicalad soon. The root cause of the bug we discovered is that grab pointer is dispatching pointer events to a modal input handler, even when the InteractionEnabled == false on the grab pointer. The fix is to not dispatch pointer events when InteractionEnabled == false. 

Radicalad is also correct that IsNearObject will return true for ALL collidables, which means the grab pointer turns on / becomes active whenever it is near ANY collidable. The expected behavior is for grabpointer to only turn on when grabpointer is near a colldable that has a NearInteractionGrabbable on it.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
radicalad: Now that @julenka showed me the Physics Raycasting bits in the `FocusProvider`, I better understand why this 0.05f margin exists. `SpherePointer` flags itself as "near" a collider to let `FocusProvider` do the more expensive overlap check.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
radicalad: force reverted to go down a simpler route.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -30,6 +30,8 @@ public class FocusProvider : BaseCoreSystem, IMixedRealityFocusProvider
         private readonly Dictionary<uint, IMixedRealityPointerMediator> pointerMediators = new Dictionary<uint, IMixedRealityPointerMediator>();
         private PointerHitResult hitResult3d = new PointerHitResult();
         private PointerHitResult hitResultUi = new PointerHitResult();
+        private Collider[] colliderQueryBuffer;
+        private int sceneQueryBufferSize = 64;
```
lukastoenneMS: #4942 adds a configurable PokePointer property for this. The FocusProvider should probably use the PointerProfile like it does for other settings.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -1019,17 +1023,30 @@ private static void QueryScene(IMixedRealityPointer pointer, IMixedRealityRaycas
...
                         break;
                     case SceneQueryType.SphereOverlap:
-                        Collider[] colliders = UnityEngine.Physics.OverlapSphere(pointer.Rays[i].Origin, pointer.SphereCastRadius, ~UnityEngine.Physics.IgnoreRaycastLayer);
 
-                        if (colliders.Length > 0)
+                        int numColliders;
+
+                        if (colliderBuffer != null)
+                        {
+                            numColliders = UnityEngine.Physics.OverlapSphereNonAlloc(pointer.Rays[i].Origin, pointer.SphereCastRadius, colliderBuffer, ~UnityEngine.Physics.IgnoreRaycastLayer);
```
lukastoenneMS: I would suggest a simple buffer saturation check here like in #4942, so that users are notified in the unlikely case that the buffer is full and they may be missing interactions.

```
            if (numColliders == colliderBuffer.Length)
            {
                Debug.LogWarning($"Maximum number of {numColliders} colliders found in FocusProvider overlap query. Consider increasing the scene query buffer size in the pointer profile.");
            }
```

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -1019,17 +1023,30 @@ private static void QueryScene(IMixedRealityPointer pointer, IMixedRealityRaycas
...
+                        }
+                        else
+                        {
+                            colliderBuffer = UnityEngine.Physics.OverlapSphere(pointer.Rays[i].Origin, pointer.SphereCastRadius, ~UnityEngine.Physics.IgnoreRaycastLayer);
```
lukastoenneMS: Doesn't look like the unbuffered code path would ever be used now.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -1019,17 +1023,30 @@ private static void QueryScene(IMixedRealityPointer pointer, IMixedRealityRaycas
...
+
+                        if (colliderBuffer.Length > 0)
                         {
                             Vector3 testPoint = pointer.Rays[i].Origin;
                             GameObject closest = null;
                             float closestDistance = Mathf.Infinity;
                             Vector3 objectHitPoint = testPoint;
 
-                            foreach (Collider collider in colliders)
+                            for (int c = 0; c < numColliders; c++)
```
lukastoenneMS: Why not foreach?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: I think all the changes in this file are no longer necessary, actually. Addison and I discussed a much simpler solution. Addison is on pat leave also. I'll take over the PR and just provide the two fixes in basecontrollerpointer and sphere pointer.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: Removed the changes here since they aren'trelevant to the bugfix. Let's re-open the discussion in separate issue if needed.

lukastoenneMS:
@julenka Why did you revert the OverlapSphereNonAlloc change? I thought it was a rather good idea to not allocate on every frame.

Issue #4966, opened 2019-06-19
MixedRealityRaycastHit Throws Exceptions on Every Frame


jbienzms:
## Describe the bug
[MixedRealityRaycastHit](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityRaycastHit.cs) throws exceptions on every frame.

A [commit](https://github.com/microsoft/MixedRealityToolkit-Unity/commit/f896a975d23edeec8eb5e71afd8a27c01d7fb83e) was made 7 days ago to wrap access to `hitInfo.lightmapCoord` in a Try / Catch block. Unfortunately this does not correct the actual condition that's causing the exception to be thrown in the first place. Even though the exception is being handled, it's still causing an exception on _every frame_. This introduces a significant performance hit. Exceptions are very expensive and should be avoided at all costs in the render thread.

There is a comment in the recent comment:

> Accessing lightmap coord appears to throw a NullReferenceException in some cases, probably when lightmaps are not used.

Rather than catching the exception, effort should be made to determine if this is in fact the root cause and determine a way to avoid accessing the property under this condition. If we determine the condition is unavoidable, we should open a bug with Unity.

## To reproduce

1. Create a new Unity project
2. Import MRTK
3. Configure MRTK for HoloLens 2
4. Press play in Unity
5. Under Exception Settings in Visual Studio, set it to stop on all exceptions (even handled ones).
6. Attach Visual Studio to Unity

You will see the Visual Studio Debugger break on every frame.

## Expected behavior

An exception should not be thrown on every frame.

## Your Setup (please complete the following information)

- Unity Version: 2018.4.2
- MRTK Version: mrtk_develop as of 2019-06-19

## Target Platform (please complete the following information)

- HoloLens 2

wiwei:
This is bad, I'm fixing this now

PR #4975, opened 2019-06-19
Make it so that MixedRealityRaycastHit doesn't throw an exception every frame.

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4966

When the default raycaster goes through a physics hit and doesn't hit anything, the generated RaycastHit isn't usable. The raycaster returns true when a hit is successful (the only time when the out parameter should be accessed) but otherwise the out parameter shouldn't be used (and it should be null instead)

 lightmapCoord itself is kinda funky when the physics hit doesn't hit anything (if you see the stuff here https://github.com/Unity-Technologies/UnityCsReference/blob/master/Modules/Physics/ScriptBindings/Dynamics.bindings.cs you'll find that it actually tries to do stuff, but the object basically isn't valid to do stuff.) Since RaycastHit is a struct it's not a nullable type, but those objects really aren't safe to use in the case where the physics hit failed.

Also adds a test to show how this all works end to end (that it doesn't throw, and that it's null in that case).


wiwei:
Test failure I think is related to the way I didn't comment out the tests for UWP, will do that shortly.

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
```diff
@@ -19,12 +19,18 @@ public interface IMixedRealityRaycastProvider : IMixedRealityDataProvider
         /// <summary>
         /// Performs a raycast using the specified <see cref="Microsoft.MixedReality.Toolkit.Physics.RayStep"/>.
         /// </summary>
+        /// <remarks>
+        /// If nothing was hit, hitInfo will be null.
```
Railboy: We're already using the bool / out pattern that Unity uses for raycast hits. Shouldn't we just return false if we hit nothing, instead of making hitInfo null?

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityRaycastHit.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// <summary>
     /// The resulting hit information from an IMixedRealityRaycastProvider.
     /// </summary>
-    public struct MixedRealityRaycastHit
+    public class MixedRealityRaycastHit
```
Railboy: We were using a struct so that people could use MixedRealityRaycastHit in the same way as RaycastHit, and to make avoiding garbage easier.

If it's necessary to tell when a MixedRealityRaycastHit is 'valid' we could add a HitSomething field or a read-only Empty property that returns true if certain values haven't been set. Or we could just rely on the bool / out pattern that's already in place.

jbienzms: Actually, @Railboy makes a very good point here @wiwei. These raycasts happen _very_ frequently and on the UI thread. If a Class is used instead of a Struct it's going to add _significant_ pressure on the garbage collector. If there is any way to use a struct instead here, it would be highly desirable.

jbienzms:
Awesome work. Thanks @wiwei !

Issue #4644, opened 2019-05-30
CI Pipeline debuggability improvements


wiwei:
The latest set of CI breakages were somewhat challenging to debug because the set of logs that are published aren't fully comprehensive - you can't see the following logs:

%localappdata%/Unity/Editor/editor.log
$ArtifactLocationPlaceThing/build/build.log

There's some really useful information in both, which have so far required us to remote into the build machines in order to debug further. Ideally these should be published in the build step so that we can at least get access to them (so that we don't need to remote into build machines to look at logs)

In addition, we need to update the build YAML to indicate that SDK_18362 == True - some of the machines were set up without the latest SDK, which didn't become apparent until we remoted in to look at the logs.

wiwei:
A couple of notes on what happened here:

1) Some of the new build machines only had SDK 10.0.17763.0 installed.
2) One of the build machines had a busted Unity install.

PR #4982, opened 2019-06-19
CI Debuggability improvements

wiwei:
## Overview
Currently whenever the AppX build fails (or nuget restore fails), the build will fail "silently" in that an error will be reported, but you have no idea what the error actually is. We currently have to remote into the build machines to find the source of the error, or otherwise do a local repro (i.e. build .NET locally using C# projects).

Ideally, what should actually happen is, our CI pipeline should actually output the logs from msbuild.exe so that we don't need to do either of those things - all of the logs should just be visible from the CI output itself.

What this change does is a few things:
1) Adds support for a "logs directory" argument for our build tools, which will be the base location for any logs that are generated from sub-processes of our build tools. So while this doesn't affect the Unity log location, any processes we spin up from our build code should log there as a base.
2) Adds a NuGet restore and AppX build log (outputted to the logs directory)
3) Updates the YAML to provide the logs directory (and then also print out the results of the build) BEFORE the APPX publish step (which is typically the failing step).

## Changes
- Fixes: #4644


wiwei:
1000 attempts later, we got it.

Issue #4944, opened 2019-06-18
Error message spam when hitting play and switching tabs in the inspector of the MRTK gameobject


thalbern:
## Describe the bug

There's a constant error spam in the console whenever you hit play in unity and then switch the tabs in the property inspector of the MRTK gameobject

![image](https://user-images.githubusercontent.com/36998103/59694288-ab117900-91e8-11e9-9ba5-b7d1029f92a2.png)

## To reproduce

Create a new project
add the MRTK gameobject / playspace through the menu
hit play
select the mrtk gameobject
in the inspector click eg on the Teleport tab
![image](https://user-images.githubusercontent.com/36998103/59694367-cf6d5580-91e8-11e9-9ae8-8c108e587474.png)

## Your Setup (please complete the following information)

- Unity Version 2018.4
- MRTK Version v2.0 RC2.1

## Target Platform (please complete the following information)

- Unity

Railboy:
This is caused by tab switching triggering a 'changed' event in the toolkit inspector, which tears down services. (The idea is if any profile has changed in any way we can't guarantee that service states will reflect the new profile data, so we tear everything down and start over.)

So we can break this down into two issues:
- Changing an inspector tab shouldn't trigger 'changed' in the inspector
- Tearing down services in play mode shouldn't cause errors

Related: #3945

Troy-Ferrell:
@Railboy I agree, let's say issue #1 will be solved by PR = #4991

And leave issue #2 as tracked by #3945

Issue #3945, opened 2019-04-11
Activating / deactivating core services causes errors


Railboy:
## Overview
Activating and deactivating core MRTK services causes errors.

## Expected Behavior
Services should be able to handle when they or other services are activated / deactivated.

## Actual Behavior
- Starting with all services activated and deactivating _any_ service causes the following errors:

```
There's already a UIRaycastCamera in the scene. It will be ignored, so please delete it to avoid confusion.
UnityEngine.Debug:LogError(Object, Object)
Microsoft.MixedReality.Toolkit.Input.FocusProvider:Initialize() (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:427)
```

```
ArgumentNullException: Value cannot be null.
Parameter name: _unity_self
Microsoft.MixedReality.Toolkit.Windows.Input.GestureRecognizerExtensions.UpdateAndResetGestures (UnityEngine.XR.WSA.Input.GestureRecognizer recognizer, UnityEngine.XR.WSA.Input.GestureSettings gestureSettings) (at Assets/MixedRealityToolkit/Extensions/GestureRecognizerExtensions.cs:11)
```

```
MissingReferenceException: The object of type 'Camera' has been destroyed but you are still trying to access it.
Your script should either check if it is null or you should not destroy the object.
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputModule.GetMousePointerEventDataForMrtkPointer (Microsoft.MixedReality.Toolkit.Input.MixedRealityInputModule+PointerData pointerData) (at Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputModule.cs:176)
```

- Starting with all services activated and deactivating InputSystem in particular causes the following errors:

```
Unable to find IMixedRealityInputSystem service.
UnityEngine.Debug:LogError(Object)
Microsoft.MixedReality.Toolkit.MixedRealityToolkit:GetService(String, Boolean) (at Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs:258)
```

```
NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.Input.BaseCursor.UpdateCursorTransform () (at Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs:360)
[...]

Issue #4335, opened 2019-05-14
PROPOSAL: Add service profile accessor property to core system interfaces.


davidkline-ms:
By strict definition, this is a breaking change (any custom system implementation would be required to update). For application code, this provides a new option for gaining access to the profiles.

## Describe the problem

Currently, application code is required to call MixedRealityToolkit.Instance.ActiveProfile.<system>Profile to access profile data at runtime. The work related to #3545 enables developer customers to use a custom registration path for systems which does not gurantee that the MixedRealityToolkit object will be instantiated.

## Describe the solution you'd like

Add an accessor property to each core system that exposes the profile to application code, similar to the following:

```
IMixedRealityInputSystemProfile Profile { get; }
```

## Describe alternatives you've considered

Considering many (most?) systems provide individual profile settings as properties of their interfaces, one option is to disallow application code from directly accessing the profiles.

Railboy:
If we do this then we may want to standardize how profiles are used by services. At the moment some services continue to use profile data at runtime, so changing the profile changes their behavior. Other services consume the profile data on initialize and ignore it from that point forward.

So if all core services 'officially' expose their profile data, there may be an incorrect expectation that either a) the profile can be modified at runtime to affect the service, or b) that the data can be modified _without_ affecting the service.

davidkline-ms:
Great points, @Railboy!

davidkline-ms:
Step 1 should be to review what gets exposed by the services and what can only be accessed via the profiles.

If memory serves, very few (any?) services read profile settings on demand.

davidkline-ms:
Scanning for references to "ActiveProfile" uncovered two example/demo scripts that access a system profile:

- MixedRealityToolkit.Examples\Demos\HandTracking\Script\GestureTester.cs
- MixedRealityToolkit.Examples\Demos\HandTracking\Script\ToggleHandVisualisation.cs

davidkline-ms:
Other key locations:

- MixedRealityToolkit\Inspectors (profile inspectors)
- MixedRealityToolkit\Providers (controllers and visualizers)
- MixedRealityToolkit.Providers (Windows mixed reality and windows voice)
- MixedRealityTookit,SDK (ux features and speech input handlers)
- MixedRealityToolkit.Services (input system and simulation)

PR #4991, opened 2019-06-20
Stop MRTK resets with inspector tab selection & disable MRTK profile inspectors in play mode

Troy-Ferrell:
## Overview

MRTK services are not harden to be reliable enough to be destroyed & recreated at runtime atm. There is work to make this better. This change though does two things

1) Ensure changing tabs in the MRTK inspector is not considered as a profile "change" which will call MRTK to reset configuration. 

2) If in play mode, all profile inspector GUI's are disabled from editing

Number 2 was added because
a) Some MRTK profile changes will cause a reset which will spew errors & problems as in #4944 and #3945
b) Some MRTK profiles changes will not be consumed when changed at runtime. For example, modifying the MRTK near clip plane value in the camera profile will not modify the actual camera near clip plane at runtime. 

Drawback: if a developer wants to make a profile change even though they are aware it will not take effect to next build/play, they will have to exit play mode to make any edits. 

When MRTK services are more stable, it might be worth re-considering number 2. But then it would still have to be communicated per profile and possibly even per property, that some settings will not take effect immediately. 

## Changes
- Fixes: #4944 
Related: #3945

## Verification
Enter play mode and cannot edit both MRTK inspector when selecting game object and cannot edit profiles when drilldown on their own inspectors

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityToolkitConfigurationProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityToolkitConfigurationProfileInspector.cs
@@ -184,6 +184,8 @@ namespace Microsoft.MixedReality.Toolkit.Editor
             var profile = target as BaseMixedRealityProfile;
             if (!RenderAsSubProfile)
             {
+                CheckEditorPlayMode();
+
                 if (!profile.IsCustomProfile)
                 {
                     EditorGUILayout.HelpBox("Default MRTK profiles cannot be edited. Create a clone of this profile to modify settings.", MessageType.Warning);
@@ -230,6 +232,22 @@ namespace Microsoft.MixedReality.Toolkit.Editor
             EditorGUILayout.EndHorizontal();
 
             EditorGUILayout.LabelField(string.Empty, GUI.skin.horizontalSlider);
+        }
+
+        /// <summary>
+        /// If application is playing, then show warning to the user and disable inspector GUI
+        /// </summary>
+        /// <returns>true if application is playing, false otherwise</returns>
+        protected bool CheckEditorPlayMode()
+        {
+            if (Application.isPlaying)
+            {
+                EditorGUILayout.HelpBox("Mixed Reality Toolkit settings cannot be edited while in play mode.", MessageType.Warning);
+                GUI.enabled = false;
+                return true;
+            }
+
+            return false;
         }
 
         /// <summary>
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs
@@ -174,6 +174,8 @@ namespace Microsoft.MixedReality.Toolkit.Editor
 
             RenderMRTKLogo();
 
+            CheckEditorPlayMode();
+
             if (!MixedRealityToolkit.IsInitialized)
             {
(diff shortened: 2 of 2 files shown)


Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityToolkitConfigurationProfileInspector.cs
```diff
@@ -232,6 +234,22 @@ protected void RenderProfileHeader(string title, string description, Object sele
...
+        /// <summary>
+        /// If application is playing, then show warning to the user and disable inspector GUI
+        /// </summary>
+        /// <returns>true if application is playing, false otherwise</returns>
+        protected bool CheckEditorPlayMode()
+        {
+            if (Application.isPlaying)
+            {
+                EditorGUILayout.HelpBox("Mixed Reality Toolkit settings cannot be edited while in play mode.", MessageType.Warning);
+                GUI.enabled = false;
```
Railboy: I don't want to over-complicate this, but have you considered using [DisabledScope](https://docs.unity3d.com/ScriptReference/EditorGUI.BeginDisabledGroup.html) instead of `GUI.enabled`? `GUI.enabled` can sometimes have ugly side effects.

Troy-Ferrell: Interesting, didn't even know that was a thing. The only thing I'll have to test is if DisableScope also sets GUI.enabled. Because all of the sub-profiles read GUI.enabled if it was already disabled. So lots of code may break if I try to disable via this was but GUI.enabled is still true?

Railboy: Yeah, it may not be worth holding up this PR. My impulse is to merge this as-is and clean up GUI.enabled all in one go in a separate PR. But I'll defer to @wiwei and @davidkline-ms.

wiwei: Sounds good to me.

wiwei:
> This change helps to harden the policy that profile data should not be modified at runtime. Which is an ambiguity we've called out before, see #4335. I'm down for this if everyone else is.

I think that a longer term goal here is that we should be able to swap out profiles at runtime (i.e. not like, willy-nilly, but just having something do a device capability check and make sure they have the right configuration for the current device should be something we support). I know that the scenarios are a bit different but I wanted to call this out explicitly so that future readers don't get contrasting thoughts here.

Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityToolkitConfigurationProfileInspector.cs
Troy-Ferrell: Great, agreed. Leave that investigation to switch everything over for disablegroup to another PR

Troy-Ferrell:
> > This change helps to harden the policy that profile data should not be modified at runtime. Which is an ambiguity we've called out before, see #4335. I'm down for this if everyone else is.
> 
> I think that a longer term goal here is that we should be able to swap out profiles at runtime (i.e. not like, willy-nilly, but just having something do a device capability check and make sure they have the right configuration for the current device should be something we support). I know that the scenarios are a bit different but I wanted to call this out explicitly so that future readers don't get contrasting thoughts here.

Good callout! Yes, I agree the MRTK services should be able to be destroyed/created and dynamically configured at runtime. Once we get to that state, then it'll be good to consider removing this play mode lock

Issue #4992, opened 2019-06-20
Wireframe shader does not support transparency


Troy-Ferrell:
## Describe the bug

It is not possible to change the renderingmode and related options in the wireframe shader as these are not piped through like the MRTK Standard shader. Thus, one cannot create transparency for the wireframe faces. 

## To reproduce

Assign MRTK wireframe shader to a sphere and ensure baseColor is set to (0,0,0,0) but see that the faces are not actually transparent

## Expected behavior

Setting rendering mode to transparent should make the faces transparent or at least based off of the base color

## Your Setup (please complete the following information)

- Unity Version 2018.4
- MRTK Version Latest

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

PR #4993, opened 2019-06-20
Update Wireframe Shader & Inspector for transparency support

Troy-Ferrell:
## Overview
This change updates the MRTK wireframe shader to support multiple rendering modes identical to the functionality in the MRTK standard shader. This is primarily to support transparency & other similar settings. 

The change creates a new base shaderGUI inspector class to be shared by both wireframe & standard shader. The related functionality is removed from StandardShaderGUI and WireframeShaderGUI is given the necessary code to render it's base properties plus the shared rendering mode-type properties. 

The shader is also updated to take the output alpha from both the base & wire color values instead of just wire intensity which will always be transparent on the faces. This gives developers more flexibility and control

## Changes
- Fixes: #4992 

## Verification
image

image

image

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
@@ -71,9 +71,122 @@ namespace Microsoft.MixedReality.Toolkit.Editor
             public static GUIContent renderQueueOverride = new GUIContent("Render Queue Override", "Manually Override the Render Queue");
         }
 
+        protected bool initialized;
+
+        protected MaterialProperty renderingMode;
+        protected MaterialProperty customRenderingMode;
+        protected MaterialProperty sourceBlend;
+        protected MaterialProperty destinationBlend;
+        protected MaterialProperty blendOperation;
+        protected MaterialProperty depthTest;
+        protected MaterialProperty depthWrite;
+        protected MaterialProperty depthOffsetFactor;
+        protected MaterialProperty depthOffsetUnits;
+        protected MaterialProperty colorWriteMask;
+        protected MaterialProperty cullMode;
+        protected MaterialProperty renderQueueOverride;
+
         protected const string LegacyShadersPath = "Legacy Shaders/";
         protected const string TransparentShadersPath = "/Transparent/";
         protected const string TransparentCutoutShadersPath = "/Transparent/Cutout/";
+
+        public override void OnGUI(MaterialEditor materialEditor, MaterialProperty[] props)
+        {
+            Material material = (Material)materialEditor.target;
+
+            FindProperties(props);
+            Initialize(material);
+
+            RenderingModeOptions(materialEditor);
+        }
+
+        protected virtual void FindProperties(MaterialProperty[] props)
+        {
+            renderingMode = FindProperty(BaseStyles.renderingModeName, props);
+            customRenderingMode = FindProperty(BaseStyles.customRenderingModeName, props);
+            sourceBlend = FindProperty(BaseStyles.sourceBlendName, props);
+            destinationBlend = FindProperty(BaseStyles.destinationBlendName, props);
+            blendOperation = FindProperty(BaseStyles.blendOperationName, props);
+            depthTest = FindProperty(BaseStyles.depthTestName, props);
+            depthWrite = FindProperty(BaseStyles.depthWriteName, props);
+            depthOffsetFactor = FindProperty(BaseStyles.depthOffsetFactorName, props);
+            depthOffsetUnits = FindProperty(BaseStyles.depthOffsetUnitsName, props);
+            colorWriteMask = FindProperty(BaseStyles.colorWriteMaskName, props);
+
+            cullMode = FindProperty("_CullMode", props);
+            renderQueueOverride = FindProperty("_RenderQueueOverride", props);
+        }
+
+        protected void Initialize(Material material)
+        {
+            if (!initialized)
+            {
+                MaterialChanged(material);
+                initialized = true;
+            }
+        }
+
+        protected virtual void MaterialChanged(Material material)
+        {
+            SetupMaterialWithRenderingMode(material,
+                (RenderingMode)renderingMode.floatValue,
+                (CustomRenderingMode)customRenderingMode.floatValue,
+                (int)renderQueueOverride.floatValue);
+        }
+
+        protected void RenderingModeOptions(MaterialEditor materialEditor)
+        {
+            EditorGUI.BeginChangeCheck();
+
+            EditorGUI.showMixedValue = renderingMode.hasMixedValue;
+            RenderingMode mode = (RenderingMode)renderingMode.floatValue;
+            EditorGUI.BeginChangeCheck();
+            mode = (RenderingMode)EditorGUILayout.Popup(renderingMode.displayName, (int)mode, BaseStyles.renderingModeNames);
+
+            if (EditorGUI.EndChangeCheck())
+            {
+                materialEditor.RegisterPropertyChangeUndo(renderingMode.displayName);
+                renderingMode.floatValue = (float)mode;
+            }
+
+            EditorGUI.showMixedValue = false;
+
+            if (EditorGUI.EndChangeCheck())
+            {
+                Object[] targets = renderingMode.targets;
+
+                foreach (Object target in targets)
+                {
+                    MaterialChanged((Material)target);
+                }
+            }
+
+            if ((RenderingMode)renderingMode.floatValue == RenderingMode.Custom)
+            {
+                EditorGUI.indentLevel += 2;
+                customRenderingMode.floatValue = EditorGUILayout.Popup(customRenderingMode.displayName, (int)customRenderingMode.floatValue, BaseStyles.customRenderingModeNames);
+                materialEditor.ShaderProperty(sourceBlend, BaseStyles.sourceBlend);
+                materialEditor.ShaderProperty(destinationBlend, BaseStyles.destinationBlend);
+                materialEditor.ShaderProperty(blendOperation, BaseStyles.blendOperation);
+                materialEditor.ShaderProperty(depthTest, BaseStyles.depthTest);
+                depthWrite.floatValue = EditorGUILayout.Popup(depthWrite.displayName, (int)depthWrite.floatValue, BaseStyles.depthWriteNames);
+                materialEditor.ShaderProperty(depthOffsetFactor, BaseStyles.depthOffsetFactor);
+                materialEditor.ShaderProperty(depthOffsetUnits, BaseStyles.depthOffsetUnits);
+                materialEditor.ShaderProperty(colorWriteMask, BaseStyles.colorWriteMask);
+                EditorGUI.indentLevel -= 2;
+            }
+
+            if (!PropertyEnabled(depthWrite))
+            {
+                if (MixedRealityToolkitShaderGUIUtilities.DisplayDepthWriteWarning(materialEditor))
+                {
+                    renderingMode.floatValue = (float)RenderingMode.Custom;
+                    depthWrite.floatValue = (float)DepthWrite.On;
+                }
+            }
+
+            materialEditor.ShaderProperty(cullMode, BaseStyles.cullMode);
+        }
 
         protected static void SetupMaterialWithRenderingMode(Material material, RenderingMode mode, CustomRenderingMode customMode, int renderQueueOverride)
         {
--- a/Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
@@ -366,7 +366,7 @@ namespace Microsoft.MixedReality.Toolkit.Editor
         /// <param name="material">Material to modify</param>
         /// <param name="keywordName">Keyword of shader feature</param>
         /// <param name="propertyName">Associated property name for shader feature</param>
-        /// <param name="propertyValue">boolean toggle of active/inactive wrapped through float</param>
+        /// <param name="propertyValue">float to be treated as a boolean flag for setting shader feature active or inactive</param>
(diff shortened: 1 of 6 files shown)


Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
```diff
@@ -0,0 +1,274 @@
...
+                case RenderingMode.Opaque:
+                    {
+                        material.SetOverrideTag(BaseStyles.renderTypeName, BaseStyles.renderingModeNames[(int)RenderingMode.Opaque]);
+                        material.SetInt(BaseStyles.customRenderingModeName, (int)CustomRenderingMode.Opaque);
+                        material.SetInt(BaseStyles.sourceBlendName, (int)BlendMode.One);
+                        material.SetInt(BaseStyles.destinationBlendName, (int)BlendMode.Zero);
+                        material.SetInt(BaseStyles.blendOperationName, (int)BlendOp.Add);
+                        material.SetInt(BaseStyles.depthTestName, (int)CompareFunction.LessEqual);
+                        material.SetInt(BaseStyles.depthWriteName, (int)DepthWrite.On);
+                        material.SetFloat(BaseStyles.depthOffsetFactorName, 0.0f);
```
wiwei: Aside from the Custom rendering mode, several of these material BaseStyles end up having the same line of code/configuration in all of these options (i.e. 0.0f for this and depthOffsetUnitsName, depthTestName for LessEqual, and so on.

I can see value in going either way (i.e. having it duplicated explicitly shows for each configuration what is expected, having the common stuff taken out helps to highlight the deltas between each configuration)

I just wanted to ask about the intent and check that that was the goal, cause if so that's all good.

wiwei: I only ask because usually i see copy+paste = bad, but in some cases it's fine if the goal is being explicit about configuration.

Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            if (material.HasProperty(propertyName))
+            {
+                return material.GetColor(propertyName);
+            }
+
+            return null;
+        }
+
+        protected static void SetFloatProperty(Material material, string keywordName, string propertyName, float? propertyValue)
```
wiwei: >SetFloatProperty [](start = 30, length = 16)

Nit: It might be worth adding a doccomment on this function, if only because it seems to be doing more than just setting a float property (i.e. also optionally enabling the keyword based on the property value itself)

wiwei: I see that this was taken from the existing standard shader... I guess it's up to you here :)

---
In reply to: [296323173](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4993#discussion_r296323173) [](ancestors = 296323173)

Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
```diff
@@ -0,0 +1,274 @@
...
+using System;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.Rendering;
+using Object = UnityEngine.Object;
+
+namespace Microsoft.MixedReality.Toolkit.Editor
+{
+    /// <summary>
+    /// A custom shader inspector for the "Mixed Reality Toolkit/Wireframe" shader.
```
wiwei: >A custom shader inspector for the "Mixed Reality Toolkit/Wireframe" shader. [](start = 8, length = 75)

I think this was some copy pasta from the wireframe shader -maybe something like "A base shader inspector for other MRTK shaders"?

Assets/MixedRealityToolkit/Inspectors/MixedRealityWireframeShaderGUI.cs
```diff
@@ -0,0 +1,210 @@
...
+            RenderingModeOptions(materialEditor);
+
+            GUILayout.Label(Styles.mainPropertiesTitle, EditorStyles.boldLabel);
+            materialEditor.ShaderProperty(baseColor, Styles.baseColor);
+            materialEditor.ShaderProperty(wireColor, Styles.wireColor);
+            materialEditor.ShaderProperty(wireThickness, Styles.wireThickness);
+
+            AdvancedOptions(materialEditor, material);
+
+            //base.OnGUI(materialEditor, props);
```
wiwei: Nit: Delete commented out line

Assets/MixedRealityToolkit/Inspectors/MixedRealityWireframeShaderGUI.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        protected void Initialize(Material material)
+        {
+            if (!initialized)
+            {
+                MaterialChanged(material);
+                initialized = true;
+            }
+        }
```
wiwei: It looks like this + the initialized state are also common between the two current shaders - if the goal is to make the base one a common set of stuff, maybe this would be worth moving in there as well. If your thoughts are that this might not be something interesting to all subclasses, then it's fine to also keep out. Really just trying to understand the standards/thoughts that go into what is being made common.

Assets/MixedRealityToolkit/Inspectors/MixedRealityWireframeShaderGUI.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        protected void MaterialChanged(Material material)
+        {
+            SetupMaterialWithRenderingMode(material, 
+                (RenderingMode)renderingMode.floatValue, 
+                (CustomRenderingMode)customRenderingMode.floatValue, 
+                (int)renderQueueOverride.floatValue);
+        }
+
+        // TODO: Merge this out? with StandardShaderGUI
```
wiwei: > // TODO: Merge this out? with StandardShaderGUI [](start = 7, length = 48)

Nit: Avoid having dangling TODOs that don't have more background/value/stuff.

If you don't see any value in merging this and don't think it will ever matter just get rid of the TODO. If there is value it might be worth just doing now.

Assets/MixedRealityToolkit/Inspectors/MixedRealityWireframeShaderGUI.cs
```diff
@@ -0,0 +1,210 @@
...
+        public override void AssignNewShaderToMaterial(Material material, Shader oldShader, Shader newShader)
+        {
+            float? cullMode = GetFloatProperty(material, "_Cull");
+
+            base.AssignNewShaderToMaterial(material, oldShader, newShader);
+   
+            SetFloatProperty(material, null, "_CullMode", cullMode);
+
+            // Setup the rendering mode based on the old shader.
+            if (oldShader == null || !oldShader.name.Contains("Legacy Shaders/"))
```
wiwei: >"Legacy Shaders/ [](start = 62, length = 16)

Nit: Consider moving some of the shared string definitions to the base class, so that if we ever need to change this (however unlikely) we can reduce the amount of copy paste/chance for error then.

Assets/MixedRealityToolkit/Inspectors/MixedRealityWireframeShaderGUI.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (EditorGUI.EndChangeCheck())
+            {
+                materialEditor.RegisterPropertyChangeUndo(renderingMode.displayName);
+                renderingMode.floatValue = (float)mode;
+            }
+
+            EditorGUI.showMixedValue = false;
+
+            if (EditorGUI.EndChangeCheck())
+            {
```
wiwei: More general question but, is is there anything weird with having unapired EndChangeCheck()s (i.e. one begin, multiple ends?)

wiwei:
I don't really have enough background in shader work to really comment on the actual content, so I really just have code nits here.

To that point, the code itself seems fine to me, though I haven't done any usage testing. If there's anything specific that you'd recommend I do, let me know.

@Cameron-Micka the only thing here that might be interesting for my own knowledge is that in other issues we've recommended folks away from transparency due to perf things, so I'm trying to maybe understand if you think there's any worry that this might lead folks down a path where they might add a bunch of transparency and tank perf.

Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
Troy-Ferrell: @wiwei this is actually the original implementation that I moved to the base class. @Cameron-Micka , assuming you made this, is there any reason not to simplify?

Assets/MixedRealityToolkit/Inspectors/MixedRealityWireframeShaderGUI.cs
Troy-Ferrell: @wiwei yea I am on the fence with this one. I didn't put it in base because it relies on the materialproperty for the given shader (protected MaterialProperty renderingMode;) So I would have to move all of those to the base class for both WireframeGUI and StandardGUI, which is doable....thinking about it more, I think it's worth it. For any shaders in the future, we should be able to define rendering modes, transparency, etc. These kind of properties are fairly universal whether it's a lit or unlit shader or whatever. If for whatever reason we need to create a new shaderGUI that doesn't have these, then we can easily insert that

Assets/MixedRealityToolkit/Inspectors/MixedRealityWireframeShaderGUI.cs
Troy-Ferrell: yes, since I am just going to move the materialproperties per my other comment, I'll just move this out as well. Both classes should be able to share this.

Assets/MixedRealityToolkit/Inspectors/MixedRealityWireframeShaderGUI.cs
Troy-Ferrell: I didn't even catch this. It is original code. But I don't like the double end with one beginning either, even if it works. I'll fix

Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
Cameron-Micka: The original inspiration came from [Unity's implementation](https://github.com/Unity-Technologies/UnityCsReference/blob/053d3ce1c53ee3dd4ff1aa08522c1284453f0796/Editor/Mono/Inspector/StandardShaderGUI.cs#L335) but since I've added features it has ballooned a little bit. I somewhat like the explicitness so I wouldn't mind keeping it as is, but feel free to factor out the common properties as long as when the user selects "Custom" it retains their previous settings.

Assets/MixedRealityToolkit/StandardAssets/Shaders/MRTK_Wireframe.shader
```diff
@@ -27,7 +48,7 @@ Shader "Mixed Reality Toolkit/Wireframe"
 
             // We only target the HoloLens (and the Unity editor), so take advantage of shader model 5.
             #pragma target 5.0
-            #pragma only_renderers d3d11
+            #pragma only_renderers d3d11 //D3D 11 & 12
```
Cameron-Micka: I know this bug existed before your PR. But we should remove the "only_renderers d3d11" pragma and wrap "target 5.0" in this preprocessor:

#if defined(SHADER_API_D3D11)
            #pragma target 5.0
#endif

So we support other platforms (iOS, Android, etc.). Feel free to do this in your PR or I can open a bug.

Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
Cameron-Micka: TL;DR It's your call. ;)

Assets/MixedRealityToolkit/StandardAssets/Shaders/MRTK_Wireframe.shader
Troy-Ferrell: No, will do in this PR! I just was following the comment and assuming this shader was only officially supported for HL

Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
Troy-Ferrell: I agree with removing out the duplicated properties. Slims down the total file size and again keeps editing if need be in the future to one location instead of 5

Troy-Ferrell:
@wiwei or @Cameron-Micka  can you look at merging this in? thanks!

Cameron-Micka:
> @wiwei or @Cameron-Micka  can you look at merging this in? thanks!

I wish I could, but I don't have the "power." But, Will should be able to!

Assets/MixedRealityToolkit/StandardAssets/Shaders/MRTK_Wireframe.shader
```diff
@@ -111,12 +132,12 @@ Shader "Mixed Reality Toolkit/Wireframe"
 
                 // Fade out the alpha but not the color so we don't get any weird halo effects from
                 // a fade to a different color.
-                float4 color = I * _WireColor + (1 - I) * _BaseColor;
-                color.a = I;
-                return color;
+                return I * _WireColor + (1 - I) * _BaseColor;
```
keveleigh: The comment still mentions doing something special with the alpha vs the color, but it looks like the new code doesn't explicitly do anything with `.a` anymore? Will this now cause the "weird halo effects"?

Troy-Ferrell: Ah yes, I need to remove that comment. I don't believe that comment was valid to begin with and was written in error. I have not experienced any halo effects with my changes.

Troy-Ferrell:
@wiwei or @keveleigh , when you get a chance, can you merge this? It's approaching 3 weeks stale. Thanks!

PR #4995, opened 2019-06-20
[Shell Parity] Contextual cursors

MenelvagorMilsom:
Animate cursor based on context.
    
As far as I can see, the only cases where we need to show cursor context is when we're doing two hand manipulation and when we're using a bounding box. These changes update the cursor context state depending on ManipulationHandler and BoundingBox.

If anyone has any additional cases where we need to display a different cursor context, let me know.


MenelvagorMilsom:
I haven't tested this on device yet. I need to do some Berlin hack prep stuff this evening. I'll test on HL1 and  HL2 tomorrow.

Assets/MixedRealityToolkit.SDK/Features/UX/Meshes/Cursors/Materials/Light.mat
```diff
@@ -9,8 +9,7 @@ Material:
   m_PrefabAsset: {fileID: 0}
   m_Name: Light
   m_Shader: {fileID: 4800000, guid: 5bdea20278144b11916d77503ba1467a, type: 3}
-  m_ShaderKeywords: _DIRECTIONAL_LIGHT _DISABLE_ALBEDO_MAP _HOVER_LIGHT _REFLECTIONS
```
MenelvagorMilsom: I don't know where these changes came from, must have accidentally committed. I'll revert.

Assets/MixedRealityToolkit.SDK/Features/UX/Meshes/Cursors/Materials/Shadow.mat
```diff
@@ -9,8 +9,7 @@ Material:
   m_PrefabAsset: {fileID: 0}
   m_Name: Shadow
   m_Shader: {fileID: 4800000, guid: 5bdea20278144b11916d77503ba1467a, type: 3}
-  m_ShaderKeywords: _DIRECTIONAL_LIGHT _DISABLE_ALBEDO_MAP _HOVER_LIGHT _REFLECTIONS
-    _SPECULAR_HIGHLIGHTS
+  m_ShaderKeywords: _DISABLE_ALBEDO_MAP _SPECULAR_HIGHLIGHTS
```
MenelvagorMilsom: I don't know where these changes came from, must have accidentally committed. I'll revert.

Assets/MixedRealityToolkit.SDK/Features/UX/Meshes/Cursors/Materials/Light.mat
MenelvagorMilsom: Oh, these changes came from @cre8ivepark's fork. Yoon, do we need these changes and the changes to Shadow.mat?

julenka:
Interesting approach. What if instead of looking for specific components we created a 'contextualcursorinfo' component that could specify the type of cursor to show when the pointer hits that game object. It would look something like this:

```
public class ContextualCursorInfo : MonoBehaviour
{
  enum ContextualCursorType
  {
    Cross,
    Ring,
    Dot,
    NorthwestSouthEast,
    NorthEastSouthwest
  }
  [Tooltip("If the cursor hovers over this gameobject, set the cursor to this cursor visual")]
  public ContextualCursorType CursorType;
}

Then, in animated cursor, you could look on hit for something like this:

```
public void Update()
{
  hitObject = figureOutWhichColliderWeHit().gameObject;
  if(hitObject != null)
  {
      ContextualCursorInfo  ctxt = hitObject.GetComponent<ContextualCursorInfo>();
      if(ctxt != null)
      {
         cursorVisualToShow = GetVisualForCursorType();
         animateToNewCursorState(cursorVisualToShow);
      }
}
```
If we look for manipulation handlers or bounding boxes, then if somebody wants to show a custom cursor for some other reason, it may not be so easy for them to do that.

File: Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs [hunk 3]
```diff
@@ -476,6 +498,102 @@ public virtual CursorStateEnum CheckCursorState()
...
+            {
+                if ((CurrentCursorActions & CursorAction.Move) != 0)
+                {
+                    return CursorContextEnum.MoveCross;
+                }
+                else if ((CurrentCursorActions & CursorAction.Scale) != 0)
+                {
+                    if (handleRigTransform != null)
+                    {
+                        Vector3 adjustedHandlePos = handleRigTransform.InverseTransformVector(Position - handleRigTransform.position);
```
julenka: It looks like you're trying to figure out what type of resize handle to use here, but I'm not fully following how because I don't understand what handleRigTransform is. Could you explain in a comment what you are doing here?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1807,21 +1807,53 @@ private bool DoesActivationMatchFocus(FocusEventData eventData)
...
+                if (grabbedHandle)
+                {
+                    currentHandleType = GetHandleType(grabbedHandle.transform);
 
-            bool handInProximity = eventData.NewFocusedObject != null && eventData.NewFocusedObject.transform.IsChildOf(transform);
-            if (handInProximity == wireframeOnly)
+                    var bc = eventData.Pointer.BaseCursor as BaseCursor;
+                    if (bc != null)
+                    {
+                        bc.handleRigTransform = rigRoot.transform;
```
julenka: Setting handleRig from manipulationhandler feels like it's creating a bit of a coupling between BaseCursor and ManipulationHandler. You need to know that in order for the resize Rotate or Scale cursor actions to work, you also need to set handleRigTransform, and that specifically it needs to match the bounding box rigRoot transform (or at least have the same shape).

Is there a more general way we could do this?


Imagine you were an external developer trying to figure out how to make the cursor show the diagonal resize cursor on your own custom asset on hover. It would require some sleuthing into our code to realize that you need to set handleRigTransform, and further that it needs to match the bounding box.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
```diff
@@ -81,6 +81,8 @@ public string PointerName
 
         public IMixedRealityCursor BaseCursor { get; set; }
 
+        public IMixedRealityCursor GazeCursor { get { return gazeProvider.GazePointer.BaseCursor; } }
```
julenka: Why not just have the BaseCursor getter return gazeProvider.GazePointer.BaseCursor?

MenelvagorMilsom:
> What do you think? Did you already try this and end up not doing it? Please let me know if you have questions.

This all sounds perfectly reasonable. I can make these changes. I was unhappy with the coupling myself but wasn't entirely sure how best to resolve it in this context.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
MenelvagorMilsom: I wasn't 100% certain on this. Is it always true that a GGVPointer will not have BaseCursor set?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
MenelvagorMilsom: Hopefully the new logic makes a bit more sense. I can still add a comment if you feel like it's necessary.

MenelvagorMilsom: `adjustedHandlePos` is still a bad name. That needs changing.

MenelvagorMilsom:
I just tested on device. Works great on HL2 but the cursor seems to be gone on HL1. Investigating further.

Assets/MixedRealityToolkit.SDK/Features/UX/Meshes/Cursors/Materials/Light.mat
cre8ivepark: Yes, updated both Light.mat and Shadow.mat to make the cursor always visible by ignoring the lighting

Assets/MixedRealityToolkit.SDK/Features/UX/Meshes/Cursors/Materials/Shadow.mat
cre8ivepark: Yes, updated both Light.mat and Shadow.mat to make the cursor always visible by ignoring the lighting

MenelvagorMilsom:
> Currently, I don't see the move arrows on the movable objects

@cre8ivepark this is supported by attaching a `CursorContextManipulationHandler` to the object you wish to see the cursor context on. We can ensure that this component exists if an object has a `ManipulationHandler` if it's preferable.

MenelvagorMilsom:
>Vertical rotation handle shows horizontal rotation arrows on volumetric objects. (it shows vertical rotation arrows on flattened objects)

Yeah, this is a problem. We could do something similar to what is being done for scale handles. The problem is that this only works because the scale handles on the corners are cubes, whereas the rotate handles are spheres. This means that we won't get normals from the focusDetails that align with the face normals. 

Maybe in bounding box we could keep the sphere for visuals, but add a box collider rather than the sphere collider. Other than that we'd need a different solution.

MenelvagorMilsom:
>I just tested on device. Works great on HL2 but the cursor seems to be gone on HL1. Investigating further.

In regards to the issues I was seeing on HL1, I couldn't confirm during which commit the problem first arose. I'm at the Berlin Hack this week. Can someone else confirm that the cursor does not show on HL1? I'm not sure if this is an actual issue, or just the way I was building and deploying to device.

If it is indeed missing, this PR obviously cannot be completed until it is fixed.

cre8ivepark:
> > I just tested on device. Works great on HL2 but the cursor seems to be gone on HL1. Investigating further.
> 
> In regards to the issues I was seeing on HL1, I couldn't confirm during which commit the problem first arose. I'm at the Berlin Hack this week. Can someone else confirm that the cursor does not show on HL1? I'm not sure if this is an actual issue, or just the way I was building and deploying to device.
> 
> If it is indeed missing, this PR obviously cannot be completed until it is fixed.

The screenshots above - I took them from HoloLens 1 :) It is working in HL1.

cre8ivepark:
> > Vertical rotation handle shows horizontal rotation arrows on volumetric objects. (it shows vertical rotation arrows on flattened objects)
> 
> Yeah, this is a problem. We could do something similar to what is being done for scale handles. The problem is that this only works because the scale handles on the corners are cubes, whereas the rotate handles are spheres. This means that we won't get normals from the focusDetails that align with the face normals.
> 
> Maybe in bounding box we could keep the sphere for visuals, but add a box collider rather than the sphere collider. Other than that we'd need a different solution.

BoundingBox.cs already has an option for the Rotation Handle collider type. Shall we make the box collider as default? I am not clear when we need a sphere collider type.
2019-06-24 08_56_45-Unity 2018 4 0f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof

julenka:
> > > Vertical rotation handle shows horizontal rotation arrows on volumetric objects. (it shows vertical rotation arrows on flattened objects)
> > 
> > 
> > Yeah, this is a problem. We could do something similar to what is being done for scale handles. The problem is that this only works because the scale handles on the corners are cubes, whereas the rotate handles are spheres. This means that we won't get normals from the focusDetails that align with the face normals.
> > Maybe in bounding box we could keep the sphere for visuals, but add a box collider rather than the sphere collider. Other than that we'd need a different solution.
> 
> BoundingBox.cs already has an option for the Rotation Handle collider type. Shall we make the box collider as default and remove sphere type? I am not clear when we need a sphere collider type.
> 2019-06-24 08_56_45-Unity 2018 4 0f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof

I think we only have sphere collider type because that's what we started with in HTK. Will added the box collider type to improve performance (ostensible sphere colliders were less efficient?) but he didn't want to remove the sphere collider type since he didn't want to break people. To prevent breaking folks who really want to use spheres, how about we just change the default to use box colliders. The visuals should also be updated to use boxes (longer boxes like we have in the shell) instead of spheres.

MenelvagorMilsom:
>how about we just change the default to use box colliders.
This is probably a good idea anyway, but if we're continuing to allow different types of colliders for handles, they shouldn't break cursor context. I have some ideas of how we can get this to work.

cre8ivepark:
Could you please check:
- I don't see the ring/dot cursors on the hand ray
- I see the contextual icons are still visible even the cursor is not on the BB handles. 
MRTK_Cursor0627

MenelvagorMilsom:
> Could you please check:
> 
> * Ring/dot cursors are not displayed
> * Fingertip ring cursor is always visible
> * I see the contextual icons are still visible even the cursor is not on the BB handles.

Sorry, uncommited changes. I'll add back now.

cre8ivepark:
Thanks Niall, Could you please check:
- Ring/dot cursors are not displayed on the object (ray penetrating through the objects)
- Contextual icons are still displayed while moving inside the body of the objects
2019-06-28 16_44_54-Unity 2018 4 2f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof
2019-06-28 16_45_02-Unity 2018 4 2f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof
2019-06-28 16_44_00-Unity 2018 4 2f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof

julenka:
It would be very useful to write tests to automate all of these breaking cases that Yoon is finding -- I think it could be easy to break in the future but hard to catch.

MenelvagorMilsom:
> Thanks Niall, Could you please check:
> - Ring/dot cursors are not displayed on the object (ray penetrating through the objects)
> - Contextual icons are still displayed while moving inside the body of the objects

Apologies, I was pretty busy last week with the Berlin hack. This was another small error I should have caught while testing.

I'll add some automated tests now to make sure that this can be caught.

cre8ivepark:
As a separate issue for HoloLens 1 (GGV), I have opened a new bug #5183 which is currently not supported by MRTK.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
```diff
@@ -0,0 +1,339 @@
...
+        public void SetUp()
+        {
+            TestUtilities.InitializeMixedRealityToolkit(true);
+            TestUtilities.PlayspaceToOriginLookingForward();
+
+            // Target frame rate is set to 50 to match the physics
+            // tick rate. The rest of the test code needs to wait on a frame to have
+            // passed, and this is a rough way of ensuring that each WaitForFixedUpdate()
+            // will roughly wait for frame to also pass.
+            Application.targetFrameRate = 50;
```
googlan: Is it still necessary?

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
```diff
@@ -0,0 +1,339 @@
...
+            yield return hand.MoveTo(onObjectPos, numFramesPerMove);
+            VerifyCursorStateFromPointers(inputSystem.FocusProvider.GetPointers<ShellHandRayPointer>(), CursorStateEnum.InteractHover);
+
+            // Trigger pinch
+            yield return hand.SetGesture(ArticulatedHandPose.GestureId.Pinch);
+            VerifyCursorStateFromPointers(inputSystem.FocusProvider.GetPointers<ShellHandRayPointer>(), CursorStateEnum.Select);
+
+            // Release pinch
+            yield return hand.SetGesture(ArticulatedHandPose.GestureId.Open);
+            VerifyCursorStateFromPointers(inputSystem.FocusProvider.GetPointers<ShellHandRayPointer>(), CursorStateEnum.Release);
```
googlan: Do I understand correctly, that Release and InteractHover _must_ come on different frames? If that's the case, yield WaitForFixedUpdate should probably be just `yield return null`.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
```diff
@@ -0,0 +1,339 @@
...
+            cube.transform.localPosition = underPointerPos;
+            yield return new WaitForFixedUpdate();
+            yield return null;
+            VerifyCursorState(inputSystem.GazeProvider.GazeCursor, CursorStateEnum.ObserveHover);
+
+            // Right hand, pointer not on cube, hand open
+            cube.transform.localPosition = abovePointerPos;
+            yield return new WaitForFixedUpdate();
+            yield return null;
+            yield return rightHand.Show(Vector3.zero);
```
googlan: Don't need `yield return null` here, because `Show` will skip a frame.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
```diff
@@ -0,0 +1,339 @@
...
+            // This messes with the CursorContextManipulationHandler hoverCount, as it decrements without
+            // ever having incremented. To avoid this, we move the cube out of focus before we add the
+            // ManipulationHandler and CursorContextManipulationHandler.
+            cube.transform.localPosition = new Vector3(0, -2, 2);
+            yield return new WaitForFixedUpdate();
+            yield return null;
+
+            cube.AddComponent<ManipulationHandler>();
+            var temp = cube.AddComponent<CursorContextManipulationHandler>();
+            yield return new WaitForFixedUpdate();
```
googlan: Don't have to wait for physics here.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
```diff
@@ -0,0 +1,339 @@
...
+            // Move cube back to original postion (described above)
+            cube.transform.localPosition = new Vector3(0, 0, 2);
+            yield return new WaitForFixedUpdate();
+            yield return null;
+
+            // Show right hand on object
+            var rightHand = new TestHand(Handedness.Right);
+            Vector3 rightPos = new Vector3(0.05f, 0, 1.5f);
+            yield return rightHand.Show(rightPos);
+            yield return new WaitForFixedUpdate();
```
googlan: I believe we don't need Wait for fixed update and null here, because hand.Show should wait for a frame for hand to show,

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
```diff
@@ -0,0 +1,339 @@
...
+            yield return null;
+            VerifyCursorContextFromPointers(inputSystem.FocusProvider.GetPointers<ShellHandRayPointer>(), CursorContextEnum.None);
+
+            // Pinch right hand
+            yield return rightHand.SetGesture(ArticulatedHandPose.GestureId.Pinch);
+            yield return new WaitForFixedUpdate();
```
googlan: Same here and below

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
```diff
@@ -0,0 +1,339 @@
...
+
+            // Show both hands on object
+            yield return rightHand.SetGesture(ArticulatedHandPose.GestureId.Open);
+            yield return rightHand.Show(rightPos);
+            yield return leftHand.SetGesture(ArticulatedHandPose.GestureId.Open);
+            yield return new WaitForFixedUpdate();
+            yield return null;
+            VerifyCursorContextFromPointers(inputSystem.FocusProvider.GetPointers<ShellHandRayPointer>(), CursorContextEnum.MoveCross);
+
+            Object.Destroy(cube.GetComponent<ManipulationHandler>());
```
googlan: Probably not necessary, as cube is Destroyed as a whole in TearDown

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
```diff
@@ -0,0 +1,339 @@
...
+
+            var empty = new GameObject();
+            var info = cube.AddComponent<CursorContextInfo>();
+            info.ObjectCenter = empty.transform;
+
+            // Show hand on object
+            var hand = new TestHand(Handedness.Right);
+            Vector3 handPos = new Vector3(0.05f, 0, 1.5f);
+            yield return hand.Show(handPos);
+            yield return new WaitForFixedUpdate();
```
googlan: Don't need 2nd and 3rd yields.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
MenelvagorMilsom: In order to get the test to pass, I need `yield return null` for _Pinch right hand_, _Show left hand on object_ and _Pinch left hand_. I'm not sure why, but it won't work without them.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BaseCursorTests.cs
MenelvagorMilsom: I'm not so sure on the exact behaviour, but when I change this to `yield return null` it fails 50% of the time.

googlan: Interesting, ok, thanks for clarification.

Issue #4274, opened 2019-05-10
Manipulation Gesture is not triggering when expected, only triggers at the same time as the Navigation gesture


Kjakubzak:
## Describe the bug

We realized recently that Gestures are no longer working properly for our app; the Manipulate Gesture and Navigation gestures seem to be going off at the same time and the expected behavior of "tap to start Manipulate" is no longer working. Now, tapping and then moving your hand while holding the tap seems to trigger both Navigation and Manipulation.  

Making sure that this wasn't a project-specific bug, I then repro'd with the MRTK HandInteractionGesturesExample scene. GIF attached. 

## To reproduce

1. Create a bare sample project and bring the latest source into it (as well as the examples).
2. Open the example scene, HandInteractionGesturesExample. 
3. Set the input simulation profile to use the Gesture simulated hand rather than the articulated simulated hand
4. Try to manipulate the object in the example scene; it only triggers the manipulate orb when you begin to navigate (and they do so at the same time).

## Expected behavior

The Manipulate gesture starts when the simulated gesture hand taps, rather than when moving.

## Screenshots

![gif](https://i.imgur.com/M0PHSIS.gif)

## Your Setup (please complete the following information)

- Unity Version 2018.3.10f1
- MRTK Version 2.0.0-20190506.1 and 2.0.0-20190510.2

## Target Platform (please complete the following information)

- HoloLens/HoloLens 2

wiwei:
@Kjakubzak just to check, is this primarily happening with input simulation, or also something that happens on a real device? The way input simulation and input works is slightly different.

Yoyozilla:
We should validate our new gesture system

Yoyozilla:
@ritijain could you please follow up on urgency

lukastoenneMS:
So just to be absolutely clear, we basically expect these two cases and order of events:

"Hold":
1. Gaze focus on object (nothing happens yet)
1. Start tap:
  - Hold gesture **started**
  - Manipulation gesture **started** [NEW]
1. Release tap:
  - Hold gesture **completed**
  - Manipulation gesture **canceled** (or should it be **completed** here?)

"Move":
1. Gaze focus on object (nothing happens yet)
1. Start tap:
  - Hold gesture **started**
  - Manipulation gesture **started** [NEW]
1. Move hand:
  - Hold gesture **canceled**
  - Navigation gesture **started**
1. Release tap:
  - Manipulation gesture **completed**
[...]

PR #5002, opened 2019-06-21
Change gesture events from simulated hands to start manipulation when tapping

lukastoenneMS:
## Overview

Tweak the order of gesture events being raised, as described in #4274.

## Changes
- Fixes: #4274 .


lukastoenneMS:
I've tried writing tests for gesture events but it's not possible to change profiles for tests atm, for switching the hand simulation mode to GGV.

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/CustomProfiles/HandInteractionGestireEventsMixedRealityInputSimulationProfile.asset
```diff
@@ -0,0 +1,54 @@
+%YAML 1.1
```
wiwei: Nit Assets/MixedRealityToolkit.Examples/Demos/HandTracking/CustomProfiles/HandInteractionGestireEventsMixedRealityInputSimulationProfile.asset

I think that Gesture is mispelled in all of these example file names within this folder. (It's spelled Gestire here, but should be Gesture)

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/GestureTester.cs
```diff
@@ -70,21 +70,21 @@ public void OnGestureUpdated(InputEventData eventData)
...
-    {
-        Debug.Log($"OnGestureUpdated [{Time.frameCount}]: {eventData.MixedRealityInputAction.Description}");
-
-        var action = eventData.MixedRealityInputAction.Description;
-        if (action == "Manipulate Action")
-        {
-            SetIndicator(ManipulationIndicator, $"Manipulation: updated {eventData.InputData}", ManipulationMaterial, eventData.InputData);
-        }
-        else if (action == "Navigation Action")
+        public void OnGestureUpdated(InputEventData<Vector3> eventData)
```
wiwei: Man, github's whitespace review detection is really bad.

There's actually no delta here aside from whitespace, but what gets rendered here, it looks like so much could have changed.

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedGestureHand.cs
```diff
@@ -148,40 +149,47 @@ protected override void UpdateInteractions(SimulatedHandData handData)
...
+                                UpdateManipulation();
+                            }
+
+                            if (navigationInProgress)
+                            {
+                                UpdateNavigation();
+                            }
+                            else
                             {
                                 if (cumulativeDelta.magnitude > manipulationStartThreshold)
```
wiwei: > if (cumulativeDelta.magnitude > manipulationStartThreshold) [](start = 31, length = 60)

Do we expect that there should ever be a case where you don't get manipulated started events, just hold start -> hold end -> navigation start -> navigation update -> navigation complete?

I'm curious because assuming there's only a hold going on and no navigation, it looks like it's possible that this would start navigation, and then on the next interaction update because a navigation is in progress, it will never start the manipulation. Or is it the case that, cumulativeDelta is only ever updated AFTER a hold has already been processed?

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/GestureTester.cs
lukastoenneMS: Yeah, i just fixed indentation of these functions, sorry for the noise.

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/CustomProfiles/HandInteractionGestireEventsMixedRealityInputSimulationProfile.asset
lukastoenneMS: Well spotted, thanks.

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedGestureHand.cs
```diff
@@ -148,40 +149,47 @@ protected override void UpdateInteractions(SimulatedHandData handData)
...
-                            if (!manipulationInProgress)
+                            if (manipulationInProgress)
+                            {
+                                UpdateManipulation();
+                            }
+
+                            if (navigationInProgress)
+                            {
+                                UpdateNavigation();
+                            }
```
wiwei: Also when we're done with this PR, it might be worth updating the input docs to specify the expected order (basically copy paste what you had in the issue into the docs)

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedGestureHand.cs
wiwei: Yeah, trying this out (pulling this down locally) I have to hold for a fairly significant amount of time to trigger the manipulation event - if I don't wait a long time, it only ever goes into navigation mode.

---
In reply to: [296315026](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5002#discussion_r296315026) [](ancestors = 296315026)

wiwei:
Overall change looks fine to me, just wanted to see the resolution of a couple of comments is all

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedGestureHand.cs
lukastoenneMS: I've changed it now so that the manipulation event starts as soon as you pinch.

Note that there will never be "canceled" events for `Manipulation` and `Navigation`, only for `Hold` when you start moving after `Hold` has already started. If i cancel the manipulation on `Hold` start then it will only work by moving immediately, which is probably not desirable either. 

With the current behavior the events are largely independent from each other, with the exception of canceling `Hold` when moving, and i think this is the least complicated way to do it.

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedGestureHand.cs
lukastoenneMS: I've added a section to the documentation about the expected order of gesture events.

I've also linked the Gesture Tester example script there, which has proven useful to devs in the past for correct usage of the events, particularly for distinguishing the 3 event types based on generic InputEventData.

wiwei: Looks great, thanks!

Assets/MixedRealityToolkit.Services/InputSimulation/SimulatedGestureHand.cs
```diff
@@ -143,126 +144,178 @@ protected override void UpdateInteractions(SimulatedHandData handData)
...
-                                }
+                                UpdateManipulation();
                             }
-                            else
+                            if (navigationInProgress)
+                            {
+                                UpdateNavigation();
+                            }
+
+                            if (cumulativeDelta.magnitude > navigationStartThreshold)
```
wiwei: So here's a really dumb question from me, but looking at how cumulativeDelta is updated:

cumulativeDelta += currentPosition - lastPosition;

It seems like it would be possible to actually move your hand back into a position such that the cumulative delta ends up being 0,0,0 again, which would in theory start the hold here.

Is this not the case (i.e. is there some abs() going on elsewhere?)

lukastoenneMS: Yes, this can happen in theory (have never seen it happen in practice though, not even with the simulated hands). Note that i've copied this from the HTK implementation.

The cumulativeDelta is also used for determining the navigation rails when they are enabled, which requires keeing a vector. The function doesn't look very robust though, we may want to revisit this at some point - if anyone actually uses it.

Issue #5003, opened 2019-06-21
MRTK Build window causes performance drop in simulation


thalbern:
## Describe the bug

If you have your app running in the unity editor and you open the MRTK Build window you will notice a sudden performance drop that persists until you close the build window.

## To reproduce
Open any mrtk scene with the profiler visible
Run the scene
Open the MRTK Build window
Observe the performance drop

## Expected behavior
Unity inspectors and UIs shouldn't decrease the apps performance considerably.

## Your Setup (please complete the following information)

- Unity Version 2018.4
- MRTK Version 2 RC2.1

PR #5007, opened 2019-06-21
Remove update repaint method in build window

Troy-Ferrell:
## Overview
With build window visible and project running, developers notice massive performance drop. Even without play mode, this window is repainting itself more than necessary due to update() method.

This change just removes the excessive repaint

## Changes
- Fixes: #5003 

## Verification
Opened window with game mode and noticed perf drop in half. Now fixed
Close & open window fine
Select buttons to switch tabs 
Build worked


wiwei:
I was initially concerned about what would happen if we removed this, but from testing a few of the common scenarios, it looks like this is fine - I thought that this was causing some bug where Open APPX Packages Location wasn't un-greying out after building the appx, but it turns out, it's just an existing bug.

Troy-Ferrell:
Same sentiment Will, I think it was just left over and isn't needed for proper unity editor window behavior

Issue #4724, opened 2019-06-05
Feature: Scrollable Object Collection


cre8ivepark:
## Describe the bug
Part of #4200 UX Controls Parity with HoloLens 2 Shell

## Note
@radicalad is bringing OOBE's scrollable list to MRTK (work-in-progress)

For now, bringing this into experimental

radicalad:
@julenka FYI

julenka:
Experimental feature completed. Let's consider graduating it to prod in 2.3 or 2.4 once we've had some feedback from folks in the 2.2 release.

PR #5014, opened 2019-06-21
Experimental UX Feature: Scrolling object collection

radicalad:
## Overview
The long wait is over for #4724, this builds on the basic concept of `ObjectCollection` to allow developers to build scrollable lists of content! This control was originally built for the HoloLens 2 Out of Box Experience (OOBE).

## Features
- Works like any other object collection, add children, press "update."
- Scrolls both up/down, and left/right.
- Works with touch, as well as controllers / far / pointers.
- Can by dynamically instantiated.
- N items in list, any configuration of viewable and nonviewable content.
- Uses MRTK Standard shader and clipping primitive for scroll clipping effect.
- Multiple velocity options - None, None snap to item,  standard falloff per frame, falloff per item (always magically lands on a full viewable set)

## What still needs work (why its experimental)
The way the Input system works, there isn't way for children that handle input (i.e. buttons) to pass the event up to the parent (in this case the scrolling collection). I've stubbed (and commented out) a way I'd like to see input handled where an interactive object has a `PassThroughMode` that when toggled passes the event along up the hierarchy. That way a control like `ScrollingObjectCollection` can determine when events can/can't sent to that child object.
_This type of change is obviously huge for the input system and would require a separate PR(s), which is why I just commented it in for visibility._

The way this was handled in HL2 OOBE was by using an interface `IScrollingChild` which was identical to `IMixedRealityTouchHandler`. Buttons were duplicated ('ForcePressable') which allowed the `ScrollingObjectCollection` to handle all eventing and pass it down to its children as appropriate. Unfortunately this pattern is hacky and would require a duplicate UX control.

The short term solution is by using UnityEvents - this will eventually cause perf issues compared to the two options/solutions above.
_Example: 300+ objects (which SCO handles no problem with no perf) will be subscribed to events, broadcasting those will be the perf issue..._

## Changes
- New Scrolling Object Collection control
- Fixed a few namespace mismatches
- Added an example scene

Additive - should be no breaking changes here:
- Formatted / strongly typed a few things that I saw when I was in a few files
- Stubbed out (then commented out) proposed input solution
- Added "sender" to base input (part of the proposed input solution)
- Some material upgrades got in here somehow - I'm assuming it was from various merges with `mrtk_development` - I didn't touch anything directly nor did I intentionally add them to my commits.


cre8ivepark:
Examples are working well but I couldn't make it work with the buttons. 
MRTK_ScrollingCollection

luis-valverde-ms:
> The short term solution is by using UnityEvents - this will eventually cause perf issues compared to the two options/solutions above.

I don't understand how the UnityEvents are used to workaround the fact that the object collection does not receive events handled by child objects.

radicalad:
Child objects can’t implement any input handlers. Yoon’s comment is spot on
- this won’t work with pressable buttons / interactables / etc. until we
solve for the input system.

luis-valverde-ms:
@radicalad , I see. Ideally, I would rather the objects included in a scrollable container wouldn't have to be aware of it at all: you don't have to mark them as pass-through and instead they only receive the event if the container dispatches it to them.

A way to do that would be having a collider in the container wrapping up everything and the ability to define layers in the input system that don't stop raycasts/queries but capture events. The container will belong to one of this layers, so the actual object hit by the pointer is still the contained one but the input system raises the event on container.

luis-valverde-ms:
@radicalad , an option you could try without changes to the input system is registering the collection as a global listener for the events that you want to intercept. Did you try that?

Cameron-Micka:
@AMollis You should check out this PR. I liked the way you were able to bubble events to items in your scrollable items prototype. Maybe that could be applied here? Please see the comment above:

> The way the Input system works, there isn't way for children that handle input (i.e. buttons) to pass the event up to the parent (in this case the scrolling collection)...

julenka:
> Examples are working well but I couldn't make it work with the buttons. Shall we remove the PressableButtons from the example scene if it does not work currently?

Let's instead fix the code so that the scrollable object collection works with the pressable buttons.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -104,6 +104,16 @@ protected static IMixedRealityInputSystem InputSystem
         protected bool forceUpdate = false;
 
         // basic button states
+
+        //TODO: Enable PassThroughMode to enable events to continue up the hierarchy
```
julenka: could you please share the whiteboard photos we took together when we discussed architecture? I don't recall where we landed, but not sure if passing through only interactables is what we wanted. I thought we wanted to have something more general.

Cameron-Micka: We built a scrolling object collection with our latest partner and were able to circumvent the "pass though mode" idea. In the end we didn't have to make any changes to the MRTK input layer.

wiwei:
Closing this change out, as it's a pretty old PR. We should reopen when synced up against latest.

PR #5017, opened 2019-06-21
Fix the broken build.

wiwei:
There were two changes that went in, that each built successfully on their own, but combined led to a break. Basically, there were these new tests which assumed there were no default listeners of the new RegisterHandler input system event system. However, a checkin also went in that started consuming that by default (this is the primary pointer work). Both changes individually were correct, but together there was a new combination that hadn't existed.


wiwei:
https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=2554

This is the actual build that's running for this. Note that one of our build machines is currently busted,

Issue #5021, opened 2019-06-22
MRTK Utility for common access to services


Troy-Ferrell:
## Describe the problem

With the addition of the new **MixedRealityServiceRegistry.TryGetService** and related APIs (thanks @davidkline-ms !) every class now has to define their own local properties and setup for accessing the MRTK service such as is common below. This get property is copy/pasted continuously across both MRTK internal code & developer partner code. 

`        
private IMixedRealityInputSystem inputSystem = null;

        /// <summary>
        /// The active instance of the input system
        /// </summary>
        public IMixedRealityInputSystem InputSystem
        {
            get
            {
                if (inputSystem == null)
                {
                    MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem);
                }
                return inputSystem;
            }
        }
`

## Describe the solution you'd like
Allow static (i.e singleton-like which was previous way) access to the MRTK services currently active at runtime. Developers would still be able to call the MixedREalityServiceRegistry directly for special requests or designs etc but this will simplify the code and process for 95% of cases.

## Describe alternatives you've considered

If we teardown MRTK services and switch out for a new MRTK input system, I'm not sure on the impact...but then the existing copy/paste paradigm would be broken regardless. If we do cross this point, it would still be better to be able to change functionality/code in one place.

PR #5022, opened 2019-06-22
Adding mrtk utility for services

Troy-Ferrell:
## Overview

Creates utility for developers to easily access MRTK services in simpler code fashion by calling the static utility class **MRTKServicesUtils** (happy to change name to something better too)

@davidkline-ms since you are driving the work around **MixedRealityServiceRegistry**, I would really love your input on this.

## Changes
- Fixes: #5021 

## Verification
This is not a breaking change and I have modified no internal code to leverage this component. It is only an addition of a code path used continuously across the project both externally and internally. 

If approved, then we can consider another PR to change internal usage to this.


Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                if (spatialAwarenessSystem == null)
+                {
+                    MixedRealityServiceRegistry.TryGetService<IMixedRealitySpatialAwarenessSystem>(out spatialAwarenessSystem);
+                }
+                return spatialAwarenessSystem;
+            }
+        }
+
+        private static IMixedRealityTeleportSystem teleportSystem = null;
```
wiwei: Super nit: Add a newline after this line (to be consistent with the ones above).

I think same thing for the thingies down below.

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
```diff
@@ -0,0 +1,110 @@
...
+{
+    /// <summary>
+    /// Utility class to easily access references to runtime Mixed Reality Toolkit Services
+    /// </summary>
+    public static class MRTKServicesUtils
+    {
+        private static IMixedRealityInputSystem inputSystem = null;
+
+        /// <summary>
+        /// The active instance of the input system
```
wiwei: As long as we're adding this new stuff, it might be worth calling out explicitly that this will return a cached value, and that if the input system is ever torn down or destroyed, this would still technically return that object (until the InvalidateReferences() thing is called).

It might be worth either having:

1) Some doccomment remark to the class saying that this is the general behavior of the class.
2) Having a per-function remark (this increases copy paste, but at the same time increases the chances that folks will actually read it when they're doing autocomplete stuff)

wiwei:
@davidkline-ms I think that you will need to look at this before it goes in.

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
```diff
@@ -0,0 +1,110 @@
...
+using Microsoft.MixedReality.Toolkit.Input;
+using Microsoft.MixedReality.Toolkit.SpatialAwareness;
+using Microsoft.MixedReality.Toolkit.Teleport;
+
+namespace Microsoft.MixedReality.Toolkit
+{
+    /// <summary>
+    /// Utility class to easily access references to runtime Mixed Reality Toolkit Services
+    /// </summary>
+    public static class MRTKServicesUtils
```
davidkline-ms: nit: prefer expanded names (not acronyms)

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
davidkline-ms: Registrars could possibly trigger an invalidate or we could add an event to signal the cache to clear. Would love to hear thoughts on these options.

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                if (diagnosticsSystem == null)
+                {
+                    MixedRealityServiceRegistry.TryGetService<IMixedRealityDiagnosticsSystem>(out diagnosticsSystem);
+                }
+                return diagnosticsSystem;
+            }
+        }
+
+        public static void InvalidateReferences()
```
davidkline-ms: consider adding a flags param to allow invalidating individual services (or all)

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
Railboy: Naming nit: Consider calling this utility `Services`?

`Toolkit.MRTKServiceUtils.InputSystem` reads like some kind of hack or workaround.

`Toolkit.Services.InputSystem` reads more like an officially sanctioned approach.

davidkline-ms: Or, ServiceCache?

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            diagnosticsSystem = null;
+            teleportSystem = null;
+            spatialAwarenessSystem = null;
+            boundarySystem = null;
+            inputSystem = null;
```
Railboy: We're missing references to the scene system and camera system core services.

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
Troy-Ferrell: Yes, I thought of this too. The original approach people have been taking in code has the same problem. I'm not sure of the best way to approach on how to notify listeners of an invalidated service

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
Troy-Ferrell: ok I'll rename to ServiceCache and update comments to note that referenced services may be swapped out and no longer available if user modified at runtime

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
Troy-Ferrell: oh didn't even know these were available now, I just searched MRTK for the registry usage. Will add!

Assets/MixedRealityToolkit/Utilities/MRTKServicesUtils.cs
Troy-Ferrell: I am not going to add a flag for now. I want to keep things simple and resetting all local references does not have much of an impact. They will just be lazy-loaded again

Troy-Ferrell: I am going to keep it simple for now. It doesn't really impact much to just do another lazy load on next iteration after full reset

Troy-Ferrell:
@davidkline-ms @wiwei @Railboy hey all, btw what is the purpose of the service references in MixedRealityToolkit.InputSystem etc? So you can call MixedRealityToolkit.InputSystem and MixedRealityToolkit.INstance.InputSystem? and then this ServiceCache and the registry?

```
private static IMixedRealityTeleportSystem teleportSystem = null;

        /// <summary>
        /// Returns true if the MixedRealityToolkit exists and has an active profile that has Teleport system enabled.
        /// </summary>
        public static bool IsTeleportSystemEnabled => IsInitialized && Instance.HasActiveProfile && Instance.ActiveProfile.IsTeleportSystemEnabled;

        /// <summary>
        /// The current Teleport System registered with the Mixed Reality Toolkit.
        /// </summary>
        public static IMixedRealityTeleportSystem TeleportSystem
        {
            get
            {
                if (isApplicationQuitting)
                {
                    return null;
                }

                if (teleportSystem != null)
                {
                    return teleportSystem;
                }

                // Quiet warnings as we check for the service. If it's not available, it has probably been
                // disabled. We'll notify about that, in case it's an accident, but otherwise remain calm about it.
                teleportSystem = Instance.GetService<IMixedRealityTeleportSystem>(showLogs: false);
                if (logTeleportSystem && (teleportSystem == null))
                {
                    Debug.LogWarning("IMixedRealityTeleportSystem service is disabled. Teleport will not be available.\nCheck MRTK Configuration Profile settings if this is unexpected.");
                }
                // If we found a valid system, then we turn logging back on for the next time we need to search.
                // If we didn't find a valid system, then we stop logging so we don't spam the debug window.
                logTeleportSystem = teleportSystem != null;
                return teleportSystem;
            }
        }
```

Railboy:
>hey all, btw what is the purpose of the service references in MixedRealityToolkit.InputSystem etc?

@Troy-Ferrell My understanding is that it cuts down on boilerplate code for core services. In cases where there's a lot of cross-talk the code would bloat very quickly.

Troy-Ferrell:
@davidkline-ms , @Railboy , and @wiwei , I wanted to follow up on this thread. 

I created this MRTK services utility to simplify the grabbing process of a MRTK service from the registry. However, again there are static instances in the MixedRealityToolkit.cs

`MixedRealityToolkit.InputSystem.blah`

I have seen other people reference this code path (ex: #5395) and it's very confusing the difference for these duplicated code paths. 

If the preferred code path is to use the MixedRealityServiceRegistry class, then I think important next steps before GA would be:

1) Deprecate the methods for MixedRealityToolkit.InputSystem, etc. 
2) Add documentation for using this proper path of MixedRealityServiceRegistry

There is also still the potential issue of references being invalidated if MRTK resets etc.

Thoughts?

Troy-Ferrell:
> IsTeleportSystemEnabled

@Railboy Regarding this, do you mean there should be a way for people to access IsTeleportSystemEnabled etc at runtime and possibly this class utility could be used for that?

wiwei:
FWIW I'm all for this

Troy-Ferrell:
Going to re-think this a bit. This needs to be simpler for customers but I don't like the idea of a MRTK object, a service registry object, and a ServiceCache object. It just seems like more redundancy

davidkline-ms:
Sorry about the delay on this @Troy-Ferrell. I will work with you to land the implementation.

Issue #4843, opened 2019-06-12
Porting guide incorrectly says to use radialsolver for billboard 


julenka:
The porting guide says to replace Billboard with Solver but the solver doesn't implement billboard behavior. MRTK alread contains a billboard script so there's no need to have this row.

keveleigh:
The porting intent was to use radial view with `Reference Direction` set to `Object Oriented` and the `Tracked Object To Reference` set to `Head`, which should billboard it towards you (these settings should probably be added to the guide as well).
Migrating over the actual Billboard script wasn't originally intended (we wanted to replace as many one-off utilities with solvers as possible), but we decided not to remove it once we realized it had been added to the repo.

Troy-Ferrell:
@keveleigh that doesn't actually work for billboard though. Billboard generally is just rotation to face user, radialview will translate the model. It also doesn't seem to possible to lock what axis of rotation are possible.

keveleigh:
@troy-ferrell If you turn off the clamps at the bottom, the object should stay in place and only billboard:
![image](https://user-images.githubusercontent.com/3580640/60045131-ebee0f80-9678-11e9-89ca-b3d518b9e42e.png)

You're right that it doesn't seem to be possible to lock the axes of rotation though.

PR #5030, opened 2019-06-24
Updating porting guide to point to billboard class

Troy-Ferrell:
## Overview
Billboard class is still needed as RadialView does not satisfy necessary functionality. 

## Changes
- Fixes: #4843

### DIFF
--- a/Documentation/HTKToMRTKPortingGuide.md
+++ b/Documentation/HTKToMRTKPortingGuide.md
@@ -1,4 +1,4 @@
-# Porting Guide
+﻿# Porting Guide
 
 ## Controller and hand input
 
@@ -7,10 +7,8 @@
 |                           | HTK 2017 |  MRTK v2  |
 |---------------------------|----------|-----------|
 | Type                      | Specific events for buttons, with input type info when relevant. | Action / Gesture based input, passed along via events. |
-| Setup                     | Place the InputManager in the scene. | Enable the input system in the Configuration Profile and specify a concrete input system type. |
+| Setup                     | Place the InputManager in the scene. | Enable the input system in the [Configuration Profile](MixedRealityConfigurationGuide.md) and specify a concrete input system type. |
 | Configuration             | Configured in the Inspector, on each individual script in the scene. | Configured via the Mixed Reality Input System Profile and its related profile, listed below. |
-
-Learn more about the [new configuration system here](MixedRealityConfigurationGuide.md).
 
 Related profiles:
 * Mixed Reality Controller Mapping Profile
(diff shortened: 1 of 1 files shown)


Documentation/HTKToMRTKPortingGuide.md
```diff
@@ -125,7 +129,7 @@ Some Utilities have been reconciled as duplicates with the Solver system. Please
 
 | HTK 2017 |  MRTK v2  |
 |----------|-----------|
-| Billboard | [`RadialView`](xref:Microsoft.MixedReality.Toolkit.Utilities.Solvers.RadialView) [Solver](README_Solver.md) |
+| Billboard | [`Billboard`](xref:Microsoft.MixedReality.Toolkit.UI.Billboard) (Need to add new component with new namespace.) |
```
wiwei: (Need to add new component with new namespace.)

Is that a TODO for us, or is that an instruction to the user? I'm not really sure what to do with that text.

Documentation/HTKToMRTKPortingGuide.md
```diff
@@ -10,6 +10,8 @@
 | Setup                     | Place the InputManager in the scene. | Enable the input system in the Configuration Profile and specify a concrete input system type. |
 | Configuration             | Configured in the Inspector, on each individual script in the scene. | Configured via the Mixed Reality Input System Profile and its related profile, listed below. |
 
+Learn more about the [new configuration system here](MixedRealityConfigurationGuide.md).
```
thalbern: usually we're trying to avoid the "here" links (doc style guide) . It would be better if we would move those links into the text . you could link to the configuration guide in the first column on the word "Configuration" ?

Documentation/HTKToMRTKPortingGuide.md
```diff
@@ -26,6 +28,8 @@ Platform support components (e.g., Windows Mixed Reality Device Manager) must be
 
 Some events no longer have unique events and now contain a MixedRealityInputAction. These actions are specified in the Input Actions profile and mapped to specific controllers and platforms in the Controller Mapping profile. Events like `OnInputDown` should now check the MixedRealityInputAction type.
 
+Learn more about the [new input system here](/Input/Overview.md).
```
thalbern: same as above - would be great if you can link to this from already existing text so we avoid the "here" links

Documentation/HTKToMRTKPortingGuide.md
keveleigh: I think, at least, `RadialView` should be left here (in addition to `Billboard`) in case people want billboarding in the context of the solver system.
Billboarding without translation is possible with `RadialView` if angle and distance clamping are ignored, though it's currently not as configurable as Billboard wrt axis locking:
image

Documentation/HTKToMRTKPortingGuide.md
Troy-Ferrell: will do!

Documentation/HTKToMRTKPortingGuide.md
Troy-Ferrell: Good feedback, I was trying to communicate that developers still need to take action during port. Even though it's Billboard and Billboard, they are not the same.

Troy-Ferrell: @keveleigh  Interesting, that is good to know for movement configuration. I saw your other comments in #4843 to configure RadialView like billboard. The problem still is two-fold though:
1) they can't configure axis-locking for billboarding which is fairly common
2) this kind of parameter setting is non-intuitive

Even if I list RadialView as an option, developers won't know how to properly change it and then again still they will be blocked by axis locking for those interested in that. 

I feel like there needs to be a separate task to either a) upgrade billboard.cs to be part of the solver system or b) create good documentation and missing functionality to extend radialview for billboarding

julenka: I couldn't figure out how to make the radial solver billboard very easily, neither has any of the other 2 people I talked to about this. I think it's a sign that the component is not necessarily obvious to use. I would appreciate having billboard be part of solvers, or to provide an easy way to make a component billboard using radial solver. For now to unblock people it would be good to just point to billboard, since currently people have no great way to knowing how to add billboarding. At least docs pointing to something is better than nothing, no?

Issue #5023, opened 2019-06-23
Pressable button cannot be created in a root hierarchy


cre8ivepark:
## Describe the bug

- Creating a new button with PressableButton.cs in the scene root generates error messages and the button does not move.
- Moving this button into an empty game object as a child component resolves this symptom.
- Verified both on mrtk_release and mrtk_development

## Repro step
Create a button from scratch (needs only 1st step to verify the symptom)
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Documentation/README_Button.md#how-to-make-a-button-from-scratch

## Error message:
NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.UI.PressableButton.get_InitialPosition () (at Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs:192)
Microsoft.MixedReality.Toolkit.UI.PressableButton.GetWorldPositionAlongPushDirection (System.Single localDistance) (at Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs:338)
Microsoft.MixedReality.Toolkit.Editor.PressableButtonInspector.DrawPlaneAndHandle (UnityEngine.Vector3[] vertices, UnityEngine.Vector3 halfExtents, System.Single distance, Microsoft.MixedReality.Toolkit.Editor.PressableButtonInspector+ButtonInfo info, System.String label, System.Boolean editingEnabled) (at Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs:201)
Microsoft.MixedReality.Toolkit.Editor.PressableButtonInspector.DrawButtonInfo (Microsoft.MixedReality.Toolkit.Editor.PressableButtonInspector+ButtonInfo info, System.Boolean editingEnabled) (at Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs:138)
Microsoft.MixedReality.Toolkit.Editor.PressableButtonInspector.OnSceneGUI () (at Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs:106)
[...]

PR #5031, opened 2019-06-24
fixed initial offset for buttons that are created in root of prefab

thalbern:
## Overview
Fixed access of parent transform - if there's no parent the initial offset is vector3.zero

## Changes
- Fixes: #5023


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -206,14 +206,19 @@ private void OnEnable()
             currentPushDistance = startPushDistance;    
         }
 
+        Vector3 getPushSpaceSourceParentPosition()
```
keveleigh: Could this be turned into a property? Something like

`private Vector3 PushSpaceSourceParentPosition => (PushSpaceSourceTransform.parent != null) ? PushSpaceSourceTransform.parent.position : Vector3.zero;`

Either way, this is missing `private` and (nit) should be capitalized.

thalbern: thanks, my head was in C++ land yesterday :D

Issue #4723, opened 2019-06-05
Tests for Interactable


julenka:
## Overview
Write five tests to give confidence that this UX control will not break when somebody makes a change. The tests should most likely be PlayMode tests to ensure that end to end behavior works

## Test List
- [ ] Interactable can be instantiated from code
- [ ] Interactable responds to OnClick event when it gets clicked with hand
- [ ] Interactable can be configured to respond to "Menu" button, and responds when the menu button is pressed (moves to press state)
- [ ] Interactable can be configured to respond to voice and responds when a voice command is issued
- [ ] Interactable can be configured to respond to global select and responds when select is pressed even when it is not focused.
- [ ] Interactable works with GGV, articulated hands, gamepad, mouse, motion controller.

julenka:
merged.

Issue #4703, opened 2019-06-04
Global listener properties on Interactable do not respond to changes at runtime.


killerantz:
## Describe the bug

The property IsGlobal will work if checked before Runtime to register it as a global listener on enable, but setting it to false after runtime does not remove it as a global listener.

The property Requires Focus to false during runtime does not set it as a global listener.

## Describe the desired behavior

These public properties can be set to true or false at runtime to add or remove the Interactable as a global listener during runtime.

julenka:
When fixing this bug, update the Interactable tests from #5033 to not require `MixedRealityToolkit.InputSystem.PushModalInputHandler(interactableObject);` after setting Interactable.IsGlobal = true.

killerantz:
I snuck this fix in with PR #4695 which was merged yesterday. We had to updated the code to register to the IMixedRealitySpeechInputhandler and the IMixedRealityInputHandler in order to merge with MRTK_development, so I added the runtime checks while making those updates.

julenka:
Thank you! Closing as fixed by #4695

PR #5033, opened 2019-06-24
Adds PlayMode Tests for Interactables + Some Test Cleanup

Railboy:
## Overview
PlayMode tests for interactables.

## Changes
- Fixes: #4723
- I've updated `BaseNearInteractionTouchable` so that `EventsToReceive` is modifiable at runtime.

## Verification

We should make sure the simulated input events being raised correctly.

>
>


Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+
+            InstantiateDefaultInteractablePrefab(
+                new Vector3(0.0f, 0.0f, 0.5f),
+                new Vector3(-90f, 0f, 0f),
+                out interactableObject,
+                out interactable,
+                out translateTargetObject);
+
+            // Set up its voice command
+            interactable.VoiceCommand = "Select";
```
julenka: I think after #4695 you will need to update this to actually make sure that the button is being focused on, otherwise the test may fail. There's a new checkbox "requires focus" that might be checked in the model_pushbutton prefab. It will ensure that the object is actually being looked at before allowing interactable to activate from voice command.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+
+            // Move the camera to origin looking at +z to more easily see the button.
```
julenka: seems like common code that you could factor out, it gets used in several test methods.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+            Vector3 targetStartPosition = translateTargetObject.localPosition;
+
+            yield return null;
+
+            // Find an input source to associate with the input event (doesn't matter which one)
+            IMixedRealityInputSource defaultInputSource = MixedRealityToolkit.InputSystem.DetectedInputSources.FirstOrDefault();
+            Assert.NotNull(defaultInputSource, "At least one input source must be present for this test to work.");
+
+            // Add interactable as a global listener
+            MixedRealityToolkit.InputSystem.PushModalInputHandler(interactableObject);
```
julenka: This line shouldn't be here because we want to actually ensure that Interactable registers itself for global input properly.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+
+            // Raise a select down input event, then wait for transition to take place
+            MixedRealityToolkit.InputSystem.RaiseOnInputDown(defaultInputSource, Handedness.None, interactable.InputAction);
```
julenka: Can you also make sure that no pointer is focusing on the interactable? To make sure that the interactable isn't accidentally receiving input from a focused object.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+            yield return new WaitForSeconds(1f);
+
+            Assert.AreNotEqual(targetStartPosition, translateTargetObject.localPosition, "Transform target object was not translated by action.");
```
julenka: Seems like instead of making sure that the button moves, it would be better to register a delegate for the OnClick handler, like we do in pressablebuttontests:

```
bool buttonPressed = false;
buttonComponent.ButtonPressed.AddListener(() =>
{
    buttonPressed = true;
});
// press the button...
// blah
// blah
Assert.IsTrue(buttonPressed, "Button did not get pressed when hand moved to press it.");

```

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+
+            // Raise a select up input event, then wait for transition to take place
+            MixedRealityToolkit.InputSystem.RaiseOnInputUp(defaultInputSource, Handedness.Right, interactable.InputAction);
+            yield return new WaitForSeconds(1f);
+
+            Assert.AreEqual(targetStartPosition, translateTargetObject.localPosition, "Transform target object was not translated back by action.");
+
+            // Remove as global listener
+            MixedRealityToolkit.InputSystem.PopModalInputHandler();
```
julenka: Again, we shouldn't need this line. If Interactable.IsGlobal is true, it should register itself

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+            Object.Destroy(interactableObject);
+            // Wait for a frame to give Unity a change to actually destroy the object
+            yield return null;
+        }
+
+        /// <summary>
+        /// Assembles a push button from primitives and uses simulated hand input to press it.
+        /// </summary>
+        /// <returns></returns>
+        //[UnityTest] Temporarily ignoring this test until we can get animation working
```
julenka: What do you mean by "until we can get animation working"? What animation?

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+            // Move the hand forward to intersect the interactable
+            var inputSimulationService = PlayModeTestUtilities.GetInputSimulationService();
+            int numSteps = 32;
+            Vector3 p1 = new Vector3(0.0f, 0f, 0f);
+            Vector3 p2 = new Vector3(0.05f, 0f, 0.51f);
+            Vector3 p3 = new Vector3(0.0f, 0f, 0.0f);
+
+            yield return PlayModeTestUtilities.ShowHand(Handedness.Right, inputSimulationService);
+            yield return PlayModeTestUtilities.MoveHandFromTo(p1, p2, numSteps, ArticulatedHandPose.GestureId.Poke, Handedness.Right, inputSimulationService);
+            yield return new WaitForSeconds(1f);
```
julenka: Why do we need to wait a second here?

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+
+            Assert.AreNotEqual(targetStartPosition, translateTargetObject.transform.localPosition, "Translate target object was not translated by action.");
```
julenka: Again, would be better to use a delegate to ensure OnClick event was triggered.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,11 +39,470 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+
+            InteractableProfileItem profileItem = new InteractableProfileItem();
+            profileItem.Themes = new System.Collections.Generic.List<Theme>() { interactableTheme };
+            profileItem.Target = childObject;
+
+            interactable.Profiles = new System.Collections.Generic.List<InteractableProfileItem>() { profileItem };
+
+            #endregion
+
+            // Set the interactable to respond to the requested inut action
```
julenka: typo: inut => input

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
julenka: Note that this may require modifying Interactable.cs so that OnClick can actually be specified from code.

Railboy:
>Instead of checking that a press event happened by seeing if the movable component moved, it would be much better to actually add a handler in Interactable.OnClick that sets some variable like onclicked to true.

Adding a listener to OnClick is a good idea. We should keep the translate check too, though. There have been a few cases where input gets passed through the system without actually triggering a visible change. If that happens it could mean something has broken.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
Railboy: Mainly to make the test easier to eyeball in the editor, but also to give the interactable theme time to translate its target. I can cut it down if we want to keep the tests brief.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
Railboy: The interactable procedurally animates a transform. In every test but this one, the animation works.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
Railboy: Interactables register themselves in OnEnable is IsGlobal is set to true. By the time I've set IsGlobal to true, OnEnable has already fired. We should probably change this but these lines are necessary until we do.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionTouchable.cs
```diff
@@ -63,6 +67,13 @@ public abstract class BaseNearInteractionTouchable : MonoBehaviour
         protected void OnEnable()
         {
             instances.Add(this);
+
+            if (touchableCollider == null)
```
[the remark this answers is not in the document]
Railboy: That PR looks very close to being merged. I think I'll roll back my changes to BaseNearInteractionTouchable and work off your changes instead.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -763,6 +763,11 @@ protected void IncreaseDimensionIndex()
             }
         }
 
+        public void ForceUpdateThemes()
```
wiwei: >  public void ForceUpdateThemes() [](start = 6, length = 33)

Nit: Add a doccomment (really about when it should be called and guidance around it)

More generally, would it make sense to have it so that setting interactable.Profiles would trigger this automatically?

wiwei: Second question is less important.

---
In reply to: [297268770](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5033#discussion_r297268770) [](ancestors = 297268770)

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
julenka: I see. Instead of looking at transform movement, can we just add a handler to OnClick method?

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
julenka: Instead of looking at transform movement, can we just add a handler to OnClick method?

julenka: when validating the tests you can use `PlayModeTestUtilities.WaitForEnterKey` to actually wait until a key is pressed. I'd prefer to keep the test as short as possible (max 0.5 seconds) since if every test takes a second they will add up.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
Railboy: I think we should open a feature request issue for interactable runtime support in general. Changing themes and changing IsGlobal would both fit there, along with a static utility class for assembling themes.

Railboy:
After reviewing with @killerantz and @julenka we're breaking out runtime assembly of an Interactable theme to a separate issue: #5050

For now our runtime-assembled button will use a theme loaded from the asset database. This is still a worthwhile test.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionTouchable.cs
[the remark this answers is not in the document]
julenka: @lukastoenneMS  and @Railboy  please coordinate in case Railboy's PR goes in first. Seems like touchableCollider could go away entirely.

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
```diff
@@ -16,7 +17,7 @@ public class TestFixture_01_MixedRealityToolkitTests
         [Test]
         public void Test_01_InitializeMixedRealityToolkit()
         {
-            TestUtilities.CreateScenes();
+            TestUtilities.EditorCreateScenes();
```
julenka: This looks unrelated to the original change (Interactable tests). Did you intend to make the change here?

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -26,6 +28,55 @@ namespace Microsoft.MixedReality.Toolkit.Tests
 {
     public class PlayModeTestUtilities
     {
+        const string playModeTestSceneName = "PlayModeTestScene";
```
julenka: Is it possible that this name could already be loaded by SceneManager? Maybe something a bit more unique like "Microsoft.MixedReality.Toolkit.Tests.PlayModeTestScene", or "MixedRealityToolkit.PlayModeTestScene"? or, suffix it with the current date?

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -26,6 +28,55 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+                }
+            }
+            else
+            {
+                // At the very start of testing, playmode may not be active. In this case, create a fresh scene
+                Scene playModeTestScene = EditorSceneManager.NewScene(NewSceneSetup.DefaultGameObjects, NewSceneMode.Additive);
+                EditorSceneManager.SetActiveScene(playModeTestScene);
+            }
+
+            // Change our render settings so we can see what's going on more clearly
```
julenka: I've actually removed RnderSettings.skybox = null from all my recent tests because it's actually *harder* to see the black boxes we instantiateover a black background. Can you remove it?

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -26,6 +28,55 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+            TestUtilities.ShutdownMixedRealityToolkit();
+
+            Scene playModeTestScene = SceneManager.GetSceneByName(playModeTestSceneName);
+            if (playModeTestScene.isLoaded)
+            {
+                foreach (GameObject gameObject in playModeTestScene.GetRootGameObjects())
+                {
+                    GameObject.Destroy(gameObject);
+                }
+            }
```
julenka: Should you also unload the scene here and delete the scene file that you created?

julenka: Currently I notice quite a few extra scenes showing up after I run playmode tests.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -26,6 +28,55 @@ namespace Microsoft.MixedReality.Toolkit.Tests
 {
     public class PlayModeTestUtilities
     {
+        const string playModeTestSceneName = "PlayModeTestScene";
+
+        /// <summary>
+        /// Creates a play mode test scene, creates an MRTK instance, initializes playspace.
+        /// </summary>
+        public static void Setup()
```
julenka: I like all this cleanup, thanks 💯

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -26,45 +26,43 @@ public static class TestUtilities
...
-            // If any of our scenes were saved, tear down the assets
-            SceneAsset primaryTestSceneAsset = AssetDatabase.LoadAssetAtPath<SceneAsset>(primaryTestSceneTemporarySavePath);
-            if (primaryTestSceneAsset != null)
+            if (!EditorApplication.isPlaying)
             {
-                AssetDatabase.DeleteAsset(primaryTestSceneTemporarySavePath);
-            }
+                // If any of our scenes were saved, tear down the assets
+                SceneAsset primaryTestSceneAsset = AssetDatabase.LoadAssetAtPath<SceneAsset>(primaryTestSceneTemporarySavePath);
+                if (primaryTestSceneAsset != null)
```
julenka: I'm curious, why do we need to delete scenes we created in edit mode, but not onces created during playmode tests?

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -113,6 +112,20 @@ public static void InitializePlayspace()
             });
         }
 
+        /// <summary>
+        /// Forces the playspace camera to face forward.
+        /// </summary>
+        public static void FaceCameraForward()
```
julenka: Hm...this method does something really similar to the one above, but has a very different name. Perhaps this method should be called "PlayspaceToOriginLookingForward"  and the method above shoudl be called "PlayspaceTCornerLookingAtOrigin" or something like that. Trying to brainstorm more descriptive method names.

Assets/MixedRealityToolkit/Providers/BaseGenericInputSource.cs
```diff
@@ -22,7 +22,7 @@ public BaseGenericInputSource(string name, IMixedRealityPointer[] pointers = nul
         {
             SourceId = (InputSystem != null) ? InputSystem.GenerateNewSourceId() : 0;
             SourceName = name;
-            Pointers = pointers ?? new[] { InputSystem?.GazeProvider.GazePointer };
+            Pointers = pointers ?? new[] { InputSystem?.GazeProvider?.GazePointer };
```
julenka: seems unrelated, did this change sneak in? or are you fixing a bug you found? Sometimes I do that, too. I normally make a note of small bugs like this that I fix in the PR to help folks in the future.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,10 +55,422 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+            InstantiateDefaultInteractablePrefab(
+                new Vector3(0.025f, 0.05f, 0.5f),
+                new Vector3(-90f, 0f, 0f),
+                out interactableObject, 
+                out interactable,
+                out translateTargetObject);
+
+            // Subscribe to interactable's on click so we know the click went through
+            bool wasClicked = false;
+            interactable.OnClick.AddListener(() => { wasClicked = true; });
```
julenka: 🙏

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
julenka: I've actually got a GA bug to fix that bug! #4703. I'll update this test to remove the push / pop modals.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,10 +55,422 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+        /// </summary>
+        /// <param name="position"></param>
+        /// <param name="rotation"></param>
+        /// <param name="interactableObject"></param>
+        /// <param name="interactable"></param>
+        /// <param name="translateTargetObject"></param>
+        private void InstantiateDefaultInteractablePrefab(Vector3 position, Vector3 rotation, out GameObject interactableObject, out Interactable interactable, out Transform translateTargetObject)
+        {
+            // Load interactable prefab
+            Object interactablePrefab = AssetDatabase.LoadAssetAtPath("Assets/MixedRealityToolkit.Examples/Demos/UX/Interactables/Prefabs/Model_PushButton.prefab", typeof(Object));
```
julenka: suggest moving that string into a variable defined above so people don't need to hunt for it. I feel like I've done this myself and regretted already not making it a variable (in pressablebuttontests).

Assets/MixedRealityToolkit/Providers/BaseGenericInputSource.cs
Railboy: This is a bug I noticed in the editor tests.

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/TestFixture_01_MixedRealityToolkitTests.cs
Railboy: I wanted to differentiate clearly between play mode setup and TestUtilities.CreateScenes, which is strictly for editor tests.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
Railboy: Unloading is async-only now, and we can't guarantee that it'll be fully unloaded before the next test. It's simpler to just destroy everything in the scene.

Railboy:
TestSimulatedMenuInputOnPrefab and TestSimulatedVoiceInputOnPrefab are failing on build check. Can anyone get these to fail locally?

julenka:
Hey Lars, I've got a script that you can run which will run the tests in headless aka batch mode, I've found that some tests fail only in batch mode. You can run it from powershell.
[ci_run_tests.zip](https://github.com/microsoft/MixedRealityToolkit-Unity/files/3331871/ci_run_tests.zip)

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -37,10 +56,457 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+#endif
+
+            // Set the interactable to respond to the requested input action
+            MixedRealityInputAction selectAction = MixedRealityToolkit.InputSystem.InputSystemProfile.InputActionsProfile.InputActions.Where(m => m.Description == selectActionDescription).FirstOrDefault();
+            Assert.NotNull(selectAction.Description, "Couldn't find " + selectActionDescription + " input action in input system profile.");
+            interactable.InputAction = selectAction;
+        }
+
+        /// <summary>
+        /// Instantiates the default interactable buttom.
```
googlan: Nit: button

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -26,6 +28,88 @@ namespace Microsoft.MixedReality.Toolkit.Tests
...
+
+                if (!sceneExists)
+                {
+                    Scene playModeTestScene = SceneManager.CreateScene(playModeTestSceneName);
+                    SceneManager.SetActiveScene(playModeTestScene);
+                }
+            }
+            else
+            {
+                // At the very start of testing, playmode may not be active.
```
googlan: Did this part appear due to calling this method in `IPrebuildSetup.Setup` ?  I think this method shouldn't be called outside of play mode at all.

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -29,45 +29,43 @@ public static class TestUtilities
...
+        /// Used only in editor tests.
         /// </summary>
-        public static void TearDownScenes()
+        public static void EditorTearDownScenes()
         {
 #if UNITY_EDITOR
-            // If any of our scenes were saved, tear down the assets
-            SceneAsset primaryTestSceneAsset = AssetDatabase.LoadAssetAtPath<SceneAsset>(primaryTestSceneTemporarySavePath);
-            if (primaryTestSceneAsset != null)
+            if (!EditorApplication.isPlaying)
```
googlan: Maybe we should assert if this method is called from play mode?

Railboy:
@googlan Thanks I'll see if that helps.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InputEventSystemTests.cs
```diff
@@ -25,10 +23,16 @@ namespace Microsoft.MixedReality.Toolkit.Tests
 
     class InputEventSystemTests
     {
+        [SetUp]
+        public void Setup()
+        {
+            PlayModeTestUtilities.Setup();
```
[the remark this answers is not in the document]
Railboy: SetupMrtkWithoutGlobalInputHandlers was using editor-only scene creation so it still needs to be modified. Since it starts by initializing an MRTK with a default profile, the 'standard' setup will still work fine. I'll just have it modify the instance created by the standard setup.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
Railboy: YES, and fixing that dramatically reduces setup complexity, thanks for calling this out.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -10,16 +10,38 @@
...
+using System.Linq;
+using UnityEditor;
 using UnityEngine;
 using UnityEngine.TestTools;
 
 namespace Microsoft.MixedReality.Toolkit.Tests
 {
     class InteractableTests
     {
+        const float buttonPressAnimationDelay = 0.75f;
```
julenka:
```suggestion
        const float buttonPressAnimationDelay = 0.25f;
        const float buttonReleaseAnimationDelay = 0.25f;
```

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -10,16 +10,38 @@
...
+        const float buttonReleaseAnimationDelay = 0.75f;
```
julenka:
```suggestion
        const float buttonReleaseAnimationDelay = 0.25f;
```

julenka:
A few tests are still not passing so we files #5153 to track fixing those, since debugging this is currently blocking progress on other GA issues.

Railboy:
The new raycast provider tests were using the 'wrong' utilities class (as of this PR) for setup and teardown.

PR #5034, opened 2019-06-24
[HTK Parity] Progress Indicator UX Controls

Railboy:
## Overview
Adds progress indicator controls to SDK. Broken out of [Scene Transition Service PR.](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/3961)

## Changes
- Fixes: #3899


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/IProgressIndicator.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Threading.Tasks;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.SceneTransitions
```
wiwei: >Extensions.SceneTransitions [](start = 41, length = 27)

Since these are general UX controls, do you think they should remain in the SceneTransistions/Extensions? Or where the other thingies should go.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
```diff
@@ -0,0 +1,97 @@
...
+using TMPro;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Extensions.SceneTransitions
+{
+    /// <summary>
+    /// Displays progress bar and (optionally) percentage text.
+    /// </summary>
+    [ExecuteAlways]
+    public class ProgressIndicatorLoadingBar : MonoBehaviour, IProgressIndicator
```
wiwei: @julenka question for you on what you think about experimental features as they relate to new UI/UX controls.

In particular, I'm kinda curious what your thoughts on if new features NEED tests and documentation before they graduate out of experimental.

Given how little time we have before we snap and how busy everyone else is, I want to avoid us adding new features that don't have tests and documentation collateral, where we need to support that API surface forever going forward. If we had tests/docs going in at the same time (or if this is literally the next thing that's going to be worked on, then I guess it's not a huge deal)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
```diff
@@ -0,0 +1,97 @@
...
+                smoothProgress = 1f;
+
+            barScale.x = Mathf.Clamp(smoothProgress, 0.01f, 1f);
+            progressBar.localScale = barScale;
+
+            progressText.gameObject.SetActive(displayPercentage);
+            progressText.text = string.Format(progressStringFormat, smoothProgress);
+
+#if UNITY_EDITOR
+            UnityEditor.EditorUtility.SetDirty(this);
```
wiwei: If you're not doing inspector stuff, should this actually be like:

https://docs.unity3d.com/ScriptReference/SceneManagement.EditorSceneManager.MarkSceneDirty.html

Looking at the docs for SetDirty it seems to suggest not using it (and instead use RecordUndo, which is not relevant for this, or MarkSceneDirty, depending on what got dirty)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// Target for smooth progress, 0-1. Serialized to test in editor.
+        /// </summary>
+        [SerializeField]
+        [Range(0f, 1f)]
+        private float targetProgress = 0f;
+        
+        public async Task OpenAsync()
+        {
+            gameObject.SetActive(true);
```
wiwei: It might be worth checking that the state is currently closed before running this right? 

No need to do any await/running if it's already open (or transition from opening to open)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
```diff
@@ -0,0 +1,97 @@
...
+            await Task.Yield();
+
+            state = ProgressIndicatorState.Open;
+        }
+
+        public async Task CloseAsync()
+        {
+            state = ProgressIndicatorState.Closing;
+
+            await Task.Yield();
```
wiwei: It might be worth documenting that OpenAsync and CloseAsync are not thread-safe and one should be fully-awaited before the other one is called. Otherwise it would be too easy to get into an inconsistent state (i.e. if you call open and then close but don't await either, it's possible that you'll end up in an Open state even when the gameObject is set to not active).

This does put the burden on the caller, which maybe is fine, but if we do that it should be documented.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            state = ProgressIndicatorState.Closed;
+
+            gameObject.SetActive(false);
+        }
+
+        private void Update()
+        {
+            if (state != ProgressIndicatorState.Open)
```
wiwei: Nit: Missing {} here, and also 83-84

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/IProgressIndicator.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        ProgressIndicatorState State { get; }
+
+        /// <summary>
+        /// The message to display during loading.
+        /// </summary>
+        string Message { set; }
+
+        /// <summary>
+        /// Loading progress value from 0-1
```
wiwei:
```suggestion
        /// Loading progress value from 0 (just started) to 1 (complete)
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
```diff
@@ -0,0 +1,97 @@
...
+{
+    /// <summary>
+    /// Displays progress bar and (optionally) percentage text.
+    /// </summary>
+    [ExecuteAlways]
+    public class ProgressIndicatorLoadingBar : MonoBehaviour, IProgressIndicator
+    {
+        const float SmoothProgressSpeed = 0.25f;
+
+        public Transform MainTransform { get { return transform; } }
```
wiwei: Nit: inheritdoc here and below on the public members that implement IProgressIndicator

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
```diff
@@ -0,0 +1,97 @@
...
+            smoothProgress = Mathf.Lerp(smoothProgress, targetProgress, SmoothProgressSpeed);
+
+            if (smoothProgress > 0.99f)
+                smoothProgress = 1f;
+
+            barScale.x = Mathf.Clamp(smoothProgress, 0.01f, 1f);
+            progressBar.localScale = barScale;
+
+            progressText.gameObject.SetActive(displayPercentage);
+            progressText.text = string.Format(progressStringFormat, smoothProgress);
```
wiwei: I think that this might allocate a new string on every single Update() - it might be worth caching the last value IF it's still the same, or otherwise doing some work to pre-allocate the strings that could render.

@Cameron-Micka had similar things with the visual profiler.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
Railboy: This just ensures that the default inspector refreshes every frame. Without it, the visual state of the inspector can sometimes drift out of sync with the object.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
Railboy: Yikes good point. These were designed for use by a system that wouldn't have made this mistake. Good catch.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
wiwei: Good to know, thanks!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorObjectDisplay.cs
```diff
@@ -0,0 +1,136 @@
...
+        private bool xAxisRotation = false;
+        [SerializeField]
+        private bool yAxisRotation = true;
+        [SerializeField]
+        private bool zAxisRotation = false;
+
+        [Header("Grow / Shrink curve on open / close")]
+        [SerializeField]
+        [Tooltip("Curve for opening animation. MUST end with value >= 1 or transition will not complete.")]
+        private AnimationCurve openCurve = AnimationCurve.EaseInOut(0, 0, 1, 1);
```
wiwei: Nit: Is there any way to validate that these things are correct (i.e. in OpenAsync or CloseAsync, could they Debug.Warn or something if they're being misused?)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorObjectDisplay.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                openScale = openCurve.Evaluate(Time.unscaledTime - startTime);
+                scaleTargetObject.transform.localScale = Vector3.one * currentScale * openScale;
+                await Task.Yield();
+            }
+
+            state = ProgressIndicatorState.Open;
+        }
+
+        public async Task CloseAsync()
```
wiwei: Nit: Inheritdoc for all the interface functions/properties here as well.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorObjectDisplay.cs
wiwei: Otherwise the infinite loop is gonna be a bad time (and possibly hard to debug for clients)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorObjectDisplay.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        private void Reset()
+        {
+            elapsedTime = 0.0f;
+            currentScale = minScale;
+        }
+
+        private void Update()
+        {
+            elapsedTime += Time.unscaledDeltaTime;
```
wiwei: Is elapsedTime still being used for something in this class? It looks like it gets set, and then updated, but never used for any other interesting part later in this file.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorObjectDisplay.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            if (state == ProgressIndicatorState.Open)
+            {
+                // Only scale while we're not opening or closing
+                currentScale = Mathf.Lerp(minScale, maxScale, progress);
+                scaleTargetObject.localScale = Vector3.one * currentScale;
+            }
+
+            if (rotationActive)
```
wiwei: Is rotation still interesting even when the indicator is closed? Just curious about the usage of this thing (esp. because the code above only does things when open)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+        [SerializeField]
+        private GameObject[] orbs = null;
+        
+        [SerializeField]
+        private TextMeshPro messageText = null;
+
+        [SerializeField]
+        private ProgressIndicatorState state = ProgressIndicatorState.Closed;
+        
+        public float RotationSpeedRawDegrees;
```
wiwei: Nit: This is a bit inconsistent with the rest of the code that we have, which has private serialized fields, and the public setters and getters.

Why? I don't know, it doesn't necessarily make sense to me, but I'd still like for us to be consistent and stick to a given style without our codebase before consciously moving our entire coding style over.

@davidkline-ms please correct me if I'm wrong here.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+        public float SpacingDegrees;
```
wiwei: For public fields, please add doccomments to explain what these are doing, even if they are reasonably named if they are here (i.e. spacing degrees - I assume these are degrees between the orbs, but maybe it's something else? - the fact that I have to make an assumption and can only validate it via looking at the implementation means having a doccomment would be helpful

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+        public bool TestStop = false;
+        public bool HasAnimationFinished = false;
+
+        private float progress;
+        private float timeElapsed;
+        private int deployedCount;
+        private bool timeUpdated;
+        private float[] angles;
+        private float timeSlice;
+        private float deg2rad = Mathf.PI / 180.0f;
```
wiwei: Seems like a useful thing that could exist in like, a MathUtils file... if not, maybe a private const? I don't think that deg2rad should change (it's just a conversion constant)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        private void StartOrbs()
+        {
+            rotationWhenStopped = 0.0f;
+            stopRequested = false;
+            timeSlice = 0;
+            timeUpdated = false;
+            deployedCount = 1;
+            timeElapsed = 0.0f;
+            angles = new float[orbs.Length];
```
wiwei: Nit: Consider only doing this if null or if angles.length != orbs.Length. Otherwise there's no reason to blow away a chunk of memory that's identical to what is needed here.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+            for (int i = 0; i < angles.Length; ++i)
+            {
+                angles[i] = 0;
+            }
+
+            dots = new Renderer[5];
```
wiwei: What's the significance of the number 5? It might be worth leaving a comment or using a const variable instead of using a magic number.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+            propertyBlocks = new MaterialPropertyBlock[dots.Length];
+            
+            for (int i = 0; i < orbs.Length; ++i)
```
wiwei: It looks like this iterates over orbs.Length, but also indexes into dots - if orbs has length longer than 5 (imagine, 10), then it will null ref when it tries to hit dots[9] which doesn't exist.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        public void StopOrbs()
+        {
+            stopRequested = true;
+            rotationWhenStopped = angles[0];
+        }
+
+        private void Update()
+        {
+            if (state == ProgressIndicatorState.Closed)
```
wiwei: Super nit: Missing braces here

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+                return;
+
+            if (HasAnimationFinished == false)
```
wiwei:
```suggestion
            if (!HasAnimationFinished)
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+            {
+                UpdateTime();
+                ControlDotStarts();
+                IncrementOrbs();
+                HandleTestStop();
```
wiwei: Consider making HandleTestStop an editor-only thing (I assume it's not something that will be exercised in player mode)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+                HandleStopping();
+            }
+        }
+        
+        private void UpdateTime()
+        {
+            if (timeUpdated == false)
```
wiwei:
```suggestion
            if (!timeUpdated)
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            HandleFade(index);
+        }
+
+        private void HandleFade(int index)
+        {
+            Color adjustedColor = propertyBlocks[index].GetColor("_Color");
+
+            //fade in
+            if (stopRequested == false && adjustedColor.a < 1.0f)
```
wiwei:
```suggestion
            if (!stopRequested && adjustedColor.a < 1.0f)
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+                adjustedColor.a -= (1.0f * timeSlice);
+                adjustedColor.a = Mathf.Max(0.0f, adjustedColor.a);
+                propertyBlocks[index].SetColor("_Color", adjustedColor);
+                dots[index].SetPropertyBlock(propertyBlocks[index]);
+            }
+        }
+
+        private void HandleTestStop()
+        {
+            if (TestStop == true && stopRequested == false)
```
wiwei:
```suggestion
            if (TestStop && !stopRequested)
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
```diff
@@ -0,0 +1,207 @@
...
+            {
+                StopOrbs();
+            }
+        }
+
+        private void HandleStopping()
+        {
+            Color adjustedColor = propertyBlocks[orbs.Length - 1].GetColor("_Color");
+            if (stopRequested == true && adjustedColor.a <= 0.01f)
```
wiwei:
```suggestion
            if (stopRequested && adjustedColor.a <= 0.01f)
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
wiwei: FWIW, pre-allocate could be overkill, at least caching would be great.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorOrbsRotator.cs
Railboy: This is due to this being ported from a legacy control. I'll make these fields consistent.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/IProgressIndicator.cs
Railboy: I'll change it to an SDK-appropriate namespace.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorObjectDisplay.cs
Railboy: It can be interesting during opening / closing transitions.

Railboy:
>Do you plan on also adding docs for this within the next couple of weeks?

Probably sooner. And I'll add some tests tomorrow.

wiwei:
@Railboy ping me when you want me to take another look here

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/ProgressIndicators/ProgressIndicatorLoadingBar.cs
```diff
@@ -0,0 +1,134 @@
...
+        public async Task OpenAsync()
+        {
+            switch (state)
+            {
+                case ProgressIndicatorState.Closed:
+                    break;
+
+                default:
+                    throw new System.Exception("Can't open in state " + state);
+            }
```
wiwei: Nit: It would be a little more concise to just do:

if (state != ProgressIndicatorState.Closed)
{
    throw new System.Exception("Can't open in state " + state);
}

Assets/MixedRealityToolkit.Tests/PlayModeTests/LoadingIndicatorTests.cs
```diff
@@ -0,0 +1,144 @@
...
+            await progressIndicator.CloseAsync();
+
+            // Make sure it's actually closed
+            Assert.True(progressIndicator.State == ProgressIndicatorState.Closed, "Progress indicator was not closed after close async call: " + progressIndicator.State);
+        }
+
+        [TearDown]
+        public void ShutdownMrtk()
+        {
+            //TestUtilities.ShutdownMixedRealityToolkit();
```
wiwei: >TestUtilities [](start = 14, length = 13)

Delete this line (and probably the TearDown function as well because it'll be empty)

Issue #5029, opened 2019-06-24
Simulated hand ray does not show until you move the hand


luis-valverde-ms:
## To reproduce

1. Play any MRTK scene with the default configuration profiles.
2. Bring up the right simulated hand by pressing 'Y'.
3. Observe how no hand ray is displayed.
4. Move hand by moving the mouse while pressing space.
5. Observe how hand ray is displayed.

![SimulatedHandRay](https://user-images.githubusercontent.com/47451126/60026425-73e11300-9693-11e9-82ab-322683d90d12.gif)

## Expected behavior

The hand ray shows from the beginning.

## Your Setup (please complete the following information)

- Unity Version: 2018.4.1.f1
- MRTK Version: development post RC 2.1 (79044680e42ea7)

## Target Platform (please complete the following information)

- Editor

## Additional context

The ray pointer is missing the input changed event raised by the ` SimulatedArticulatedHand` to notify of the initial index finger pose. This is caused by `InputSystemGlobalListener.lateInitialize` being set to true and thus delaying the registration of the pointer as a global listener to a later point when the event has already been raised.

Alexees:
I would like to add that the same thing happens in the portal with the emulated motion controllers, with the addition that the transform gizmos are also only loaded when you move at least once.

julenka:
Thanks for the great analysis, Luis. You pretty much figured it all out. Wish all bugs were written like that haha. 

I actually hit this bug when writing my bounding box tests. I worked around it by setting `        private bool lateInitialize = false;` in InputSystemGlobalHandlerListener. Not sure why lateInitialize needs to be true. I'm still working on my bounding box tests, might take another day. If nobody else fixes this in an earlier PR I can check it in.

julenka:
This also fixes a bug where the hand ray shows at the origin until you move it (which I thought was an open issue but I cannot find it...)

julenka:
I'm assigning myself because I have a fix but this is POACHABLE -- I will fix it as part of my PR for bounding box tests which might go out for PR in about a day.

luis-valverde-ms:
@julenka , thanks for taking over this. A few things to bear in mind:

1. After #4970 , global listeners distributed with MRTK inherit from `InputSystemGlobalHandlerListener` instead.  `InputSystemGlobalListener` is still around for backwards compatibility.
[...]

Issue #4730, opened 2019-06-05
Tests for Bounding Box


julenka:
## Overview
Write five tests to give confidence that bounding box will not break when somebody makes a change. The tests should most likely be PlayMode tests to ensure that end to end behavior works.

## Test List
- [X] Bounding box can be instantiated from code
- [x] Bounding box scale works with near 
- [x] Bounding box scale works with far
- [x] Changing the transform of bounding box target updates the bounding box bounds
- [x] Changing the boundsoverride of bounding box updates the bounding box

julenka:
merged.

PR #5036, opened 2019-06-25
Bounding Box Tests

julenka:
## Overview
Add bounding box tests, a couple more small test utilities, and simulated hand ray not showing up until you move hand a bit.

## Changes
- Fixes: #4730 by implementing all tests outlined in the bug. The tests actually highlighted bug #5029 which I then fixed thanks to @luis-valverde-ms excellent but description
- Fixes: #5029 
- Add couple utility methods to tests
- Make a couple members in input simulation service set-table to make them easier to test.

Thanks to Lukas for making the input simulation system easy to modify at run time so I can do near and far HoloLens 1 style tests :-)

## Verification
- PlayMode tests pass individually
- When I run all the tests, I get a stale scene left over, need to fix still.
- Getting the following error currently when I run any test (whether it is the tests I added, or not). Possibly existing bug:

```
NullReferenceException: Object reference not set to an instance of an object
UnityEditor.TestTools.TestRunner.PlaymodeLauncher+BackgroundWatcher.OnPlayModeStateChanged
```


Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
```diff
@@ -0,0 +1,282 @@
...
+
+            bbox.transform.localScale *= 0.5f;
+            bbox.Active = true;
+
+            return bbox;
+        }
+        #endregion
+
+        [UnityTest]
+        public IEnumerator BBoxInstantiate()
```
julenka: add comment

Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
```diff
@@ -0,0 +1,282 @@
...
+        #region Utilities
+        [TearDown]
+        public void ShutdownMrtk()
+        {
+            TestUtilities.ShutdownMixedRealityToolkit();
+        }
+
+        /// <summary>
+        /// Instantiates a bounding box at 0, 0, -1.5f
+        /// box is at scale 1,1,1
```
julenka: scale is 0.5, fix comment

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// </summary>
     public abstract class InputSystemGlobalHandlerListener : MonoBehaviour
     {
-        private bool lateInitialize = true;
+        private bool lateInitialize = false;
```
lukastoenneMS: @googlan Is this a bugfix? Doesn't look related to the rest of the changes here.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
```diff
@@ -0,0 +1,282 @@
...
+
+            var newObject = new GameObject();
+            var bc = newObject.AddComponent<BoxCollider>();
+            bc.center = new Vector3(.25f, 0, 0);
+            bc.size = new Vector3(0.162f, 0.1f, 1);
+            bbox.BoundsOverride = bc;
+            yield return null;
+
+            // Get the dimensions of the corners
+            List<GameObject> corners = FindDescendantsContainingName(bbox.gameObject, "corner_");
```
lukastoenneMS: Can we add a function to BoundingBox to get the GameObject for all or one specific corner(s)? Could just be a wrapper around the FindDescendants, but would keep the logic inside BoundingBox.cs.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
```diff
@@ -0,0 +1,282 @@
...
+            TestUtilities.AssertAboutEqual(bounds.center, startCenter, "bbox incorrect center at start");
+            TestUtilities.AssertAboutEqual(bounds.size, startSize, "bbox incorrect size at start");
+
+            var iss = PlayModeTestUtilities.GetInputSimulationService();
+            var oldIsp = iss.InputSimulationProfile;
+            var isp = new MixedRealityInputSimulationProfile
+            {
+                HandSimulationMode = HandSimulationMode.Gestures
+            };
+            iss.InputSimulationProfile = isp;
```
lukastoenneMS: So swapping out just the input sim profile works like this? I tried something similar for writing gesture event tests, but was setting the entire MRTK.Instance.ActiveProfile because i didn't want to risk changing assets (the default input system profile) just by running tests. If the tests fail they would not reset the profile and inadvertently change the asset, which might get saved automatically by Unity. But changing the ActiveProfile at runtime causes a lot of broken behavior. Just changing the input sim profile probably is easier to keep working.

Assets/MixedRealityToolkit.Tests/TestUtilities.cs
```diff
@@ -163,5 +165,11 @@ public static void ShutdownMixedRealityToolkit()
             return ScriptableObject.CreateInstance<T>();
 #endif
         }
+
+        public static void AssertAboutEqual(Vector3 actual, Vector3 expected, string message)
```
luis-valverde-ms: Unity has something for this, have a look at Vector3EqualityComparer here: https://docs.unity3d.com/2017.4/Documentation/Manual/PlaymodeTestFramework.html.

thalbern: thanks! I should use that for manipulationhandler as well - doing the manual magnitude compare there as well

Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
julenka: Yep, this was working for me! The only thing I needed to make sure was to change the profile back to original at the end of the test.

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
julenka: Yeah it's a bugfix, for #5029. I need it for the tests to pass. I can make it as a separate PR, but didn't want to add extra overhead of getting approve for such a small change.

julenka: I found the commit that introduced the lateInitialize variable:

```
1c17f7847e1d3a3a8e125b876a9c57ad1fb42a6f
Fixed input handler race condition if input system and teleport system not found before gameobjects wake up
```

I think with my change, the handler will never register itself if the input system is not available when the component gets enabled. I think we need one more change. I'll go ahead and pull that as separate pr.

Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
julenka: ah great suggestion, thanks.

julenka:
This PR blocked by #5044 need to complete that first or tests won't pass

Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
julenka: Moved to separate PR.

julenka:
Error from CI is:

  F:\aipmragent_work\1\s\Assets\MixedRealityToolkit.Tests\TestUtilities.cs(8,19): **error CS0234: The type or namespace name 'TestTools' does not exist in the namespace 'UnityEngine' (are you missing an assembly reference?)** [f:\aipmragent_work\1\a\build\UWP_x86_.NET\GeneratedProjects\UWP\Microsoft.MixedReality.Toolkit.Tests\Microsoft.MixedReality.Toolkit.Tests.csproj]

Probably need to modify .asmdef files. Investigating.

julenka:
WaitForEndOfFrame [is not invoked when running tests in batch mode](https://docs.unity3d.com/ScriptReference/WaitForEndOfFrame.html)

PR #5044, opened 2019-06-25
Fix simulated hand ray not correct until hand moves

julenka:
## Overview
Update InputSystemGlobalHandlerListener to try register as a global listener in OnEnable. Previously it was trying to register in Start, causing it to miss some events.

## Changes
- Fixes: #5029

## Background
The `lateInitialize` variable was introduced in the following commit:

```
1c17f7847e1d3a3a8e125b876a9c57ad1fb42a6f
Author: Stephen Hodgson <hodgson.designs@gmail.com>
Date:   Sun Sep 9 14:13:25 2018 -0700

Fixed input handler race condition if input system and teleport system not found before gameobjects wake up
```

So it sounds like there were cases when a component was enabled before input system was valid. However, we do want to register some components in OnEnable to ensure they get all input events. Therefore, modify the code s.t. we try to register both in OnEnable and in Start.

## Verification
Verified that simulated hand ray is valid when hand appears.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/InputSystemGlobalHandlerListener.cs
@@ -12,7 +12,8 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// </summary>
     public abstract class InputSystemGlobalHandlerListener : MonoBehaviour
     {
-        private bool lateInitialize = true;
+        // If true, we will try to register ourselves as a global input listener in Start
+        private bool lateInitialize = false;
         private IMixedRealityInputSystem inputSystem = null;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


wiwei:
It might also be interesting to note that, while this does help with this scenario, I would imagine that generally speaking as part of registering, consumers should also try to get the initial state - if that isn't happening, then going forward we might want to fix things that way.

Generally speaking if you want to get state you:
1) Register
2) Poll for state

or 
1) Register with immediate callback for starting state.

And that way you guarantee you don't miss data.

With this fix it definitely works, but it seems like we're just pushing #1 earlier (and in theory still could be problematic based on load order) but is fine for the time being.

wiwei:
Also, do you know if this is 100% repro, or intermittent? i.e. it seems like we shouldn't regress this in future changes, and if this something a test could be written for without too much trouble, this might be worth it.

julenka:
> Also, do you know if this is 100% repro, or intermittent? i.e. it seems like we shouldn't regress this in future changes, and if this something a test could be written for without too much trouble, this might be worth it.

It's 100% repro

julenka:
> > Also, do you know if this is 100% repro, or intermittent? i.e. it seems like we shouldn't regress this in future changes, and if this something a test could be written for without too much trouble, this might be worth it.
> 
> It's 100% repro. Additionally, the bounding box tests break if this bug regresses. When scaling bounding box using GGV, the box will jump during the first frame if we ever regress this. I'll see if I can add a test to verify this quickly.

Writing the test ended up being more than 20 minutes of work, so I'm going to file it as an issue for us to resolve post-GA. #5053

Issue #2934, opened 2018-10-17
Fature-Request: Display Browser Window in 3D space


Augmentaio:
Mixed Reality Toolkit

It would be fantastic to have an interactive browser window (2D canvas or 3D box) to open HTML pages from WWW. There is no satisfactory solution in asset store or forums on this.

ForrestTrepte:
@Augmentaio, we now have an example of launching a browser window in the HandInteractionsExamples Unity scene as of #5047. I have attached a (compressed) GIF showing what this looks like. Does that solve your problem? Note that this runs the browser as a separate application so it isn't really like a web control that can be tightly integrated within your application.
![Launch](https://user-images.githubusercontent.com/17389739/60473013-18c59800-9c20-11e9-8522-3398af6eae62.gif)

Yoyozilla:
closing as there's an example to show how to do this

PR #5047, opened 2019-06-25
launch browser example

ForrestTrepte:
## Overview

Add a button that launches the Edge browser. This demonstrates that the browser slate can be launched while your application continues running. ~~The web page that is launched contains controls modeled after the Unity GUI controls in the examples, showing a button, checkbox, slider, dropdown, and edit box. This demonstrates the option of integrating with existing web tools or of creating simple form UI using HTML and shows what this kind of interaction would feel like.~~

~~Note that I originally implemented this by calling the UWP LaunchUriAsync API, but later found out that is unnecessary because Unity's Application.OpenURL seems to work just as well.~~

## Changes
- Fixes: #2934.

## Verification

~~Note that this references a web page in order to demonstrate using web controls from within a browser. Because I didn't want to host a web site for this purpose, I am linking to a page on github. That page won't exist unless this PR is first merged to mrtk_development.~~

~~If you would like to try this change locally, you can change the URL to point to the HTML file in my fork as follows:~~
* ~~In HandInteractionExamples.unity, select the SceneContent/LaunchBrowserButton object.~~
* ~~In the inspector, scroll down to the Interactable OnClick event. Change the URL to the following: ```http://htmlpreview.github.io/?https://raw.githubusercontent.com/ForrestTrepte/MixedRealityToolkit-Unity/feature/launchBrowser/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/WebForm/SampleWebForm.htm```.~~

### DIFF
--- a/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/LaunchUri.cs
+++ b/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/LaunchUri.cs
@@ -24,7 +24,7 @@ namespace Microsoft.MixedReality.Toolkit.Examples.Demos
                 bool result = await global::Windows.System.Launcher.LaunchUriAsync(new System.Uri(uri));
                 if (!result)
                 {
-                    Debug.LogError("Browser failed to launch.");
+                    Debug.LogError("Launching URI failed to launch.");
                 }
             }, false);
 #else
--- a/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/LaunchUri.cs
+++ b/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/LaunchUri.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Examples.Demos
+{
+    public class LaunchUri : MonoBehaviour
+    {
+        /// <summary>
+        /// Launch a UWP slate app. In most cases, your experience can continue running while the
+        /// launched app renders on top.
+        /// </summary>
+        /// <param name="uri">Url of the web page or app to launch. See https://docs.microsoft.com/en-us/windows/uwp/launch-resume/launch-default-app
+        /// for more information about the protocols that can be used when launching apps.</param>
+        public void Launch(string uri)
+        {
+            Debug.Log($"LaunchUri: Launching {uri}");
+
+#if WINDOWS_UWP
+            UnityEngine.WSA.Application.InvokeOnUIThread(async () =>
+            {
+                bool result = await global::Windows.System.Launcher.LaunchUriAsync(new System.Uri(uri));
+                if (!result)
+                {
+                    Debug.LogError("Browser failed to launch.");
+                }
+            }, false);
+#else
+            Application.OpenURL(uri);
+#endif
+        }
+    }
+}
--- a/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Scenes/HandInteractionExamples.unity
+++ b/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Scenes/HandInteractionExamples.unity
@@ -753,6 +753,10 @@ PrefabInstance:
   m_Modification:
     m_TransformParent: {fileID: 16674827}
     m_Modifications:
+    - target: {fileID: 316800718, guid: 45fd0ad89d6d17b4fbe68eb48dbe9de9, type: 3}
+      propertyPath: m_Name
+      value: Pressable Button (4)
+      objectReference: {fileID: 0}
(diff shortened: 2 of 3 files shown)


julenka:
> Note that I originally implemented this by calling the UWP LaunchUriAsync API, but later found out that is unnecessary because Unity's Application.OpenURL seems to work just as well.

Could you please still include the LaunchUriAsync example as well (perhaps a button next to the webpage button that opens settings by launching ms-settings://)? That will teach people that you can actually launch any UWP as long as it implements a protocol.

julenka:
Could you also please include a screenshot of the new button you added to the scene as part of this PR, so we can find it and also visually review the code? Even better would be a gif of the new behavior, if you have time. For example, here's a really nice PR that demonstrates new features: https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4532

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/WebForm/SampleWebForm.htm
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
julenka: I wonder if rendering a custom page like this might confuse people. Have you tried instead just launching something like http://aka.ms/mrtk? or http://aka.ms/mrtkdocs?

ForrestTrepte: Thanks for reviewing, @julenka . I will delete the web page that demonstrates interacting with a form of buttons/sliders/etc. and instead launch aka.ms/mrdocs in order to demonstrate that you can launch an arbitrary web site. I will also add a button to launch the settings app and make a comment that any 2D UWP application can also be launched in this fashion as documented at https://docs.microsoft.com/en-us/windows/uwp/launch-resume/launch-default-app.

How does that sound?

julenka: That sounds great, thank you!!!

ForrestTrepte:
I'd be happy to capture a gif if there's an easy way to do that. Is there a tool that folks recommend for converting the .mp4 from a mixed reality capture into a gif? Would I need to make it super-short so that the file size isn't too large for github? @Cameron-Micka, please let me know if you have any tips for doing this.

Cameron-Micka:
@ForrestTrepte I've just been using the tool @jwittner recommends called [LICEcap](https://www.cockos.com/licecap/). GitHub only accepts files up to 10Mbs, so if I find I'm over that limit I resize and/or compress my gif from LICEcap with this online tool: https://ezgif.com/resize

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/WebForm/SampleWebForm.htm
ForrestTrepte: This is implemented in my latest push.

ForrestTrepte:
The latest implementation includes two buttons for launching aka.ms/mrdocs and settings.
Launch

ForrestTrepte:
@julenka: Ready for re-review.

Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/LaunchUri.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            Debug.Log($"LaunchUri: Launching {uri}");
+
+#if WINDOWS_UWP
+            UnityEngine.WSA.Application.InvokeOnUIThread(async () =>
+            {
+                bool result = await global::Windows.System.Launcher.LaunchUriAsync(new System.Uri(uri));
+                if (!result)
+                {
+                    Debug.LogError("Browser failed to launch.");
```
julenka: nit: suggest $"Launching URI failed to launch"

csxbwang:
How to locate the launched browser to the specified location？

jwittner:
@csxbwang - in my experience the window is spawned by the OS in a "follow" mode and stays in my view until I pin it.

csxbwang:
> @csxbwang - in my experience the window is spawned by the OS in a "follow" mode and stays in my view until I pin it.

Can I write codes to control the launched window, for example, automatically pin it and translate it to pre-defined location by attaching a world anchor?

Wolvie9:
...I'm not to sure, but I think so

**FirstLostLastFound°°

jwittner:
@csxbwang - not something I've tested, but looking at the APIs I don't believe so. You're effectively launching a URI directly into the system and then letting it do it's thing. I'm not certain there aren't APIs for controlling external applications, but due to the sandbox nature of UWP I'm not optimistic they exist.

paulparkinson:
In addition to not being able to position the launched program/browser. It seems you can only launch one (at least using  LaunchUri). Any thoughts on how to launch multiple?

ForrestTrepte:
Note that this is an old (merged) pull request. @paulparkinson, for posterity, if you don't receive an answer here you may want to open a [feature request issue](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/new/choose).

Issue #4998, opened 2019-06-20
pointers should not always raise PointerClicked on InputUp


naostranMS:
## Describe the bug

In HTK, we used to get InputClicked events in response to tap events from the gesture recognizer (for WMR/HL). Now in MRTK, it seems that the tap event isn't handled at all, and PointerClicked happens unconditionally with PointerUp (at least for BaseControllerPointer, GGVPointer, and PokePointer) whereas there used to be a distinction: tap would only happen when there's a short time between input down/up, whereas hold/manipulate happen after a long period of time (it seems like tap and hold should be mutually exclusive, but I'm not 100% certain if that is actually the case for WMR).

Often PointerUp and PointerClicked can be used interchangeably, but there are some cases where it is helpful to just respond to a click - particularly when hold and/or manipulate are also used for different functionality, in which case it is helpful to have the other gestures filtered automatically.

Note that UnityTouchController already works that way (although it's kind of weird that sometimes the event is coming from the pointer and other times from the controller) - it would be nice to see the same applied to other controller types (WMR in particular).
 
Here's the old version:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/98365a45ca0a8d99b857afb288a4b4dbc4311daf/Assets/HoloToolkit/Input/Scripts/InputSources/GesturesInput.cs#L38
And the new one:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/e5dac4afce15cc6719b6cd11b965b1dcb0919a6a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs#L403

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive

naostranMS:
@julenka You had mentioned also having an issue with this in the past - do you have any insight into whether using the tap gesture (for WMR at least) - as it used to - would work?
I'm currently in the process of adding support for the tap gesture to input sim and WMRDeviceManager, so it would be easy enough for me to switch this over to tap if that is a helpful change.

julenka:
I think the tap gesture  would only get raised on device if you release select without moving, is that what you were asking about?

julenka:
Hm...what do you mean by a tap gesture? Do you mean a press followed by a release, or just a press?

naostranMS:
[...]

PR #5051, opened 2019-06-25
Added tap gesture handling for WMR and input sim

naostranMS:
## Overview
The tap gesture was previously used (in HTK) to distinguish an input up from a quick click vs a hold or manipulate gesture. Although it is possible to get more or less the same event by handling the other gestures on the receiver side (e.g. check for hold/manipulate/navigate between input down and input up, and if none of the other gestures fired, raise a tap/click event) but it seems much simpler to handle it using the gesture recognizer.
I'm re-adding that functionality for WMR and input simulation, and have added the tap action to the default gesture profiles.

## Changes
- Added tap handling to WMRDeviceManager
- Added tap handling to SimulatedGestureHand (this is also undoing a recent change to how input sim fired the manipulation gesture - see: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4274)
- Added tap action/gesture to default configuration profiles
- Added test of tap action to GestureTester/HandInteractionGestureEventsExample
- Fixed a bug in settings rails vs non-rails navigation gesture settings (I think...)

## Other notes
As a follow-up, I'd like to use the tap gesture for the pointer click event, rather than pointer click effectively being identical to pointer up. Any feedback on that idea would be appreciated: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4998


Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
```diff
@@ -59,3 +59,6 @@ MonoBehaviour:
   - id: 15
     description: Toggle Profiler
     axisConstraint: 0
+  - id: 16
+    description: Tap
```
keveleigh: Do we need `Tap` to be distinct from `Select`? `Select` has always been the equivalent of an airtap, whether in HTK or MRTK.

keveleigh: I can see having both become confusing ("Do I want my buttons to respond to taps? Select? If I choose tap, do my buttons work on platforms that don't support 'tap'?" )

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityGesturesProfile.asset
```diff
@@ -36,3 +37,9 @@ MonoBehaviour:
       id: 9
       description: Manipulate Action
       axisConstraint: 0
+  - description: Tap
```
keveleigh: A few of the historical reasons we didn't add tap to the gesture recognizer in v2:
1. Not all platforms support a specifically defined "tap".
1. We could define "tap" the way we define "click" today for those platforms, while specially defining "click/tap" for WMR.
    a. This is interesting, because then we have a better defined click on WMR while still supporting click on other platforms
    b. but then we run into the potential of non-deterministic ordering of events, especially across platforms, since we'll be listening to one polling-based implementation for "down", "up", and (on some platforms) "click", and an event-based implementation for "click/tap", on only one platform.

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
naostranMS: This is a good point and just something I probably didn't think through enough. My immediate feeling is that using Select is more sensible, but I will look into that a bit more.

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityGesturesProfile.asset
naostranMS: Yeah, this makes sense. I think your point in 2.a. about having a "better defined click" is particularly compelling to me - as my work is focused on WMR, "native" behavior is preferable to "consistent cross-platform" behavior. I appreciate that the latter is important for MRTK and I think that has to be the first priority, but I'd like to get both if we can.
Regarding 2.b. - that seems totally plausible and I can imagine it causing problems. I assume if we are just making up the events ourselves (as in UnityTouchController) we can make it match, but I wouldn't be that surprised if other platforms have defined "clicks" that don't fully agree with WMR.

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityInputActionsProfile.asset
naostranMS: given that the Select action is already used for PointerUp/PointerDown/PointerClicked, I feel satisfied it's the right choice for this

naostranMS: I've updated to using select in almost all places (some I felt it still made sense to keep the "tap" name as that's what the gesture recognizer calls it, but it uses the "select" input action and I called the new gesture type "select" as well) - let me know what you think.

Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityGesturesProfile.asset
naostranMS: @keveleigh I guess in the end all I really care about is that there is some event (click/tap/select) on input up that is mutually exclusive with hold/manipulate. I don't think there's any particular reason it needs to come from the gesture recognizer - e.g. I'd also be ok with having the pointer decide what constitutes a click vs hold vs drag (as in UnityTouchController).

At the same time, I also kind of feel like maybe it's ok to use the platform implementations as long as it is understood that the exact ordering is not well-defined. I don't recall offhand which of tap or input up comes first (or even if it is necessarily consistent), and as it logically seems like they should happen at the same time I would try to avoid writing code that assumes a certain ordering. I also appreciate that such assumptions often happen unintentionally - but that seems unavoidable to some extent when dealing with heterogeneous controllers. E.g. does MRTK make any guarantees about the timing of the hold/manipulate gestures? Or just assume that they are "close enough"?

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs
```diff
@@ -398,52 +402,79 @@ private void RegisterGestureEvents()
...
+            if (selectAction != MixedRealityInputAction.None)
+            {
+                gestureRecognizer.Tapped -= GestureRecognizer_Tapped;
+            }
         }
 
         private void RegisterNavigationEvents()
         {
-            if (navigationGestureRecognizer == null)
+            if (navigationAction != MixedRealityInputAction.None)
```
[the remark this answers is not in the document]
naostranMS: I think that's reasonable. On a related note, it wasn't really clear to me why it's using separate gesture recognizers and switching between them vs having a single gesture recognizer and switching between manipulation/navigation settings using SetRecognizableGestures. IIRC, going into navigation mode also turns off hold (and now tap), which seems kind of weird - I understand why manipulate and navigate would be mutually exclusive, but it seems like hold/tap are useful in both cases.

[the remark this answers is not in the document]
keveleigh: >IIRC, going into navigation mode also turns off hold (and now tap)

Oh huh, I thought we had done some tests a while back and found that wasn't the case, but I could be misremembering.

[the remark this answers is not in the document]
naostranMS: Admittedly, I haven't tested it on device - it does work the way you describe in input sim, but input sim doesn't really work the same way as WMRDeviceManager.
But from looking at the code, WMRDeviceManager enforces that NavigationRecognizerEnabled and GestureRecognizerEnabled are mutually exclusive, and their corresponding gesture recognizers are started/canceled accordingly. In the default gesture profile, Manipulation Gestures includes tap and hold:
image
but Navigation Gestures doesn't:
image
That said, I don't seen anywhere that actually switches to navigation mode (and we don't use it for our app) so I don't know how important it is.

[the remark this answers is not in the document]
naostranMS: The navigation gesture recognizer also doesn't register handlers for the tap/hold events, so I don't really see how it could be sending those events (unless the gesture recognizers just ignore CancelGestures and keep firing the events).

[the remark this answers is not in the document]
naostranMS: I just tried enabling the navigation gesture recognizer (with default WMR gesture settings) using a WMR headset, and didn't get tap/hold events.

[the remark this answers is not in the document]
naostranMS: also, I went ahead and added the check you requested initially (and fixed the merge conflict) - so hopefully this change is ready to go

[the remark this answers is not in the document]
SimonDarksideJ: NIT: with regards to MixedRealityInputAction.None
You should also create a cached reference to this value instead of calling it each time, to save the cost of calling new.

Issue #4434, opened 2019-05-16
Bounding box scaling varies based on whether the box is directly on the target/collider or parented to it


djohnsomsft:
The scaling of the bounding box is inconsistent depending on whether it is applied directly to the target or to a parent of the target. While the scaling starts out the same, if the bounding box is on the target, the handles will shrink to be unusably small.

This example shows a set of two of the same model:
The left one has no target and the overridden BoxCollider on the same object as the BoundingBox
The right one has the BoundingBox on a parent with a disabled BoxCollider, and the overridden BoxCollider + Target referencing the child.

![boundingboxes0](https://user-images.githubusercontent.com/27969160/57886136-635e9280-77e1-11e9-90de-55f2f9acbbb8.PNG)

You can see they both start out the same:

![boundingboxes1](https://user-images.githubusercontent.com/27969160/57886237-9a34a880-77e1-11e9-8ad7-5323f19034fb.PNG)

However, upon scaling down, the handles for the unparented one also scale down to become tiny. The parented one maintains handle sizes consistent with the set handle size in the editor properties and remain usable.

![boundingboxes2](https://user-images.githubusercontent.com/27969160/57886306-c4866600-77e1-11e9-88d2-c66fa2a3099b.PNG)

Expected behavior is that either the developer controls this (fixed vs scaling size), potentially able to change the fixed size programmatically if the model gets too small.

julenka:
Okay, I understand why this is happening. 

In first case (no parent), the bounding box is scaling SampleLargeModelBounded, which is also scaling rigRoot. BoundingBox has no code to ensure that after scaling some parent transform, the affordances maintain size.

![image](https://user-images.githubusercontent.com/168492/60148208-fbf31580-9784-11e9-9614-98ebf0281648.png)


In the second case, the scale of SampleParentBounds always remains at Vector3.one, because the child object "Collection 2" is getting scaled.

![image](https://user-images.githubusercontent.com/168492/60148237-204ef200-9785-11e9-9b21-ca1dda41d532.png)

julenka:
merged

PR #5055, opened 2019-06-26
BoundingBox Fix Corners incorrect size when rigroot is under targetTransform

julenka:
## Overview
See issue description for explanation of bug. Fixed by recomputing scale of corners when targetTransform changes.

## Changes
- Fixes: #4434

## Verification
- Verified fixes issue
- verified that boundingboxruntime tests pass 
- Also verified that HandInteractionExamples scene works.


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1695,6 +1702,27 @@ private void UpdateRigHandles()
...
+                {
+                    if (scaleHandlePrefab == null)
+                    {
+                        corners[i].localScale = cornerDimensions;
+                    }
+                    else
+                    {
+                        // If we are modifying a corner that has a prefab, it has a different structure
+                        // Modify the visuals root instead.
+                        corners[i].GetChild(0).GetChild(0).localScale = cornerDimensions;
```
thalbern: this is not very clear to me. can there potentially be another prefab used that has a different hierarchy structure? can that break the behavior?

julenka: I tried corners[i], and then corners[i].GetChild(0), but neither of those worked. The prefab is constructed in that very particular way, if you look at CreateScaleCorners(). You need to scale those particular nodes. I could instead create a list of the nodes to scale and use those...

julenka: i was also quite bummed that i had to special case corners made via prefabs in this way. And then to look 2 levels deep in the hierarchy. Maybe @luis-valverde-ms  has some suggestions since he did bounding box prefabs?

luis-valverde-ms: As far as i can remember I ended up with that hierarchy as the simplest one that would allow me to use mirroring scale to instance the same prefab for each corner.

julenka: All right, I updated the approach to use an array of cornerVisual transforms, hopefully that approach is more understandable.

julenka:
John actually came up with a much nicer solution to this -- just set localscale = 1 at the very top of the updateRigHandlers function. He's going to update to use that approach in his PR #4895

Issue #4753, opened 2019-06-06
Tests for Input System


julenka:
## Overview
Write minimum five tests to give confidence that input system will not break when somebody makes a change. The tests should most likely be PlayMode tests to ensure that end to end behavior works.

## Test List

- Pointer events: #5064 

**TODO: asignee please write down cases to test. You can use [this issue](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4723) as an example**

PR #5064, opened 2019-06-26
Pointer events tests

luis-valverde-ms:
Tests to verify that pointer events are being raised correctly. This is part of the resolution for #4753 . 

It also includes a minor fix to `PlayModeTestUtilities` so they now also update the left simulated hand when required.


Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs
```diff
@@ -0,0 +1,313 @@
...
+            }
+
+            public State state = new State();
+            private int numFocusing;
+
+            void IMixedRealityFocusChangedHandler.OnBeforeFocusChange(FocusEventData eventData)
+            {
+                Assert.AreNotSame(eventData.OldFocusedObject, eventData.NewFocusedObject);
+                Assert.AreEqual(state.numFocusChanged, state.numBeforeFocusChange);
+
```
lukastoenneMS: Can we assert numFocusEnter == numFocusExit here? Or should it be numFocusEnter  == numFocusExit + 1, i.e. already entered?

Similarly i would expect numFocusEnter == numFocusExit + 1 in OnFocusChanged.

luis-valverde-ms: Focus enter/left events are received only for the first pointer entering/last leaving.

lukastoenneMS: I see, never mind then.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs
```diff
@@ -0,0 +1,313 @@
...
+            // Show right hand far enough from the test collider to not give it focus
+            var rightHand = new Hand(Handedness.Right);
+            Vector3 rightNoFocusPos = new Vector3(0.05f, 0, 0.5f);
+            yield return rightHand.Show(rightNoFocusPos);
+
+            // Make sure no events have been raised
+            var expected = new PointerHandler.State();
+            handler.state.AssertEqual(expected);
+
+            // Show right hand far enough from the test collider to not give it focus
```
wiwei:
```suggestion
            // Show left hand far enough from the test collider to not give it focus
```

Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs
```diff
@@ -0,0 +1,313 @@
...
+            {
+                gestureId = newGestureId;
+                return PlayModeTestUtilities.MoveHandFromTo(position, position, 1, gestureId, handedness, simulationService);
+            }
+        }
+
+        // Initializes MRTK, instantiates the test content prefab and adds a pointer handler to the test collider
+        private PointerHandler SetupScene()
+        {
+            TestUtilities.InitializeMixedRealityToolkit(true);
```
julenka: Have a look at how lars is setting up and destroying tests in #5033. I think Lars will soon get his changes in, you'll probably need to update setup/teardown to match that pattern.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs
```diff
@@ -0,0 +1,313 @@
...
+using Microsoft.MixedReality.Toolkit.Utilities;
+using Microsoft.MixedReality.Toolkit.Input;
+
+namespace Microsoft.MixedReality.Toolkit.Tests
+{
+    // Tests to verify that pointer events are being raised correctly
+    public class PointerEventsTests
+    {
+
+        // Helper script used to keep track of the number of pointer events raised
```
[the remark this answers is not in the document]
luis-valverde-ms: I don't think it is worth using `FocusedObjectEventCatcher<>` here as we're counting events separately and there is not that much more to it.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs
luis-valverde-ms: We should probably do a batch update of all tests when Lars's changes are in.

Issue #5061, opened 2019-06-26
Auto-generated camera is missing MixedRealityInputModule


lukastoenneMS:
## Describe the bug

When the CameraCache does not find a camera it will auto-generate one. This auto-generated camera, however, does not have a MixedRealityInputModule script component on it.

The MixedRealityInputModule script pipes MRTK input data into the Unity event system. It is necessary in particular for make Unity UI elements work.

## To reproduce

1. Create an new scene
1. _Mixed Reality Toolkit > Add to Scene and Configure_
1. Remove the main camera from the MixedRealityPlayspace
1. Enter PlayMode

A new camera is generated automatically with a warning:
"No main camera found. The Mixed Reality Toolkit requires at least one camera in the scene. One will be generated now."

## Expected behavior

The generated camera has a MixedRealityInputModule. Unity UI elements work.

## Additional context

Playmode tests are using auto-generated camera and do not work with Unity UI buttons for this reason.

Note that the MixedRealityInputModule type is in the InputSystem service assembly and is not accessible directly in the CameraCache.

wiwei:
Just tried this out on latest, looks like the input module is getting added. (I thought I saw @railboy make a change to address this some time back)

PR #5065, opened 2019-06-26
Unity ui test

lukastoenneMS:
## Overview

Basic tests for Unity UI elements and poke pointer interaction.

## Changes

Adds tests for NearInteractionTouchableUnityUI in combination with different UI components.

Only the Button and Toggle UI components are tested right now, other elements should be added later.


lukastoenneMS:
Consider this part of #4753

lukastoenneMS:
This PR contains a workaround for #5061 which should be removed when the generated cameras work correctly.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
```diff
@@ -29,6 +29,8 @@ public class NearInteractionTouchable : ColliderNearInteractionTouchable
         protected Vector3 localUp = Vector3.up;
         public Vector3 LocalUp { get => localUp; }
 
+        public bool AreLocalVectorsOrthogonal => Vector3.Dot(localForward, localUp) == 0;
+
```
wiwei: Nit: Add a doccomment (in particular, it would be good to clarify that it only reasons over LocalForward and LocalUp - there are other Local* vectors which aren't involved in this calculation and I might expect as a reader to see those somehow factored in)

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -339,6 +339,78 @@ public IEnumerator NearInteractionTouchableOverlapQuerySaturation()
...
+            yield return PlayModeTestUtilities.ShowHand(Handedness.Right, inputSim);
+
+            using (var catcher = new UnityToggleEventCatcher(toggle))
+            {
+                // Touch started and completed when entering and exiting
+                yield return PlayModeTestUtilities.MoveHandFromTo(initialHandPosition, objectPosition, numSteps, ArticulatedHandPose.GestureId.Open, Handedness.Right, inputSim);
+                Assert.IsFalse(catcher.Value);
+                Assert.AreEqual(0, catcher.Changed);
+                yield return PlayModeTestUtilities.MoveHandFromTo(objectPosition, initialHandPosition, numSteps, ArticulatedHandPose.GestureId.Pinch, Handedness.Right, inputSim);
+                Assert.IsTrue(catcher.Value);
```
wiwei: >Value [](start = 38, length = 5)

Nit: It might be a little more readable to have this named IsToggled (instead of just 'Value')

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -339,6 +339,78 @@ public IEnumerator NearInteractionTouchableOverlapQuerySaturation()
...
+                Assert.AreEqual(1, catcher.Changed);
```
wiwei: Would it be possible to also try toggling again (to show that the reverse transition also works?)

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -339,6 +339,78 @@ public IEnumerator NearInteractionTouchableOverlapQuerySaturation()
...
+        /// </summary>
+        /// <returns></returns>
+        [UnityTest]
+        public IEnumerator NearInteractionTouchableUnityUiButton()
+        {
+            UnityUiUtilities.EnsureInputModule();
+
+            var canvas = UnityUiUtilities.CreateCanvas(0.002f);
+
+            var button = UnityUiUtilities.CreateButton(Color.gray, Color.blue, Color.green);
```
julenka: Nice, this will make tests for Unity UI far interaction easy to write too!

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
lukastoenneMS: Added that now.

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
lukastoenneMS: Renamed to IsOn now (mirrors the Toggle.isOn property)

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
lukastoenneMS: done

keveleigh:
Looks like two of the new tests are failing

lukastoenneMS:
When i run this locally only the second test fails, and only when i run it in succession with the previous button test. The PokePointer correctly handles the touchable and the raycast is happening, but somehow the Unity Toggle is never clicked. I suspect more issues with the event system or raycast cameras (see #5061) combined with the fact that we recycle the scene instead of loading a new one. But that's just me poking around in the dark ...

lukastoenneMS:
So it turns out this is a wider problem with how services are torn down between tests, but the scene is re-used. The MRInputModule is created just once, but when the first test ends the MRTK event system service is destroyed and recreated, which unregisters the input module as a source state handler. In the next test it won't see any of the hand controllers being added.

I've put in a workaround to destroy and recreate the input module as well between tests, so it re-registers with the event system. This is a more fundamental problem with services and i've logged it here with a suggestion to at least detect potential issues in the future:  #5116

lukastoenneMS:
Still failing on CI, but now every test passes locally for me, both individually and in sequence :sad

PR #5068, opened 2019-06-26
Enforce compiler warnings as errors in the MRTK.

wiwei:
It is critically important to ensure that the MRTK builds without warnings and errors. We have consumers that have strict "warnings as errors."

Over the past month, we've seen a number of changes go in with compiler warnings - these never should have been checked in in the first place.

This change adds a C# compiler flag to will treat all compiler warnings as errors.

Note that this DOES have a ramification on a subset of consumers that currently pull directly from mrtk_development (and mrtk_release, once this change goes out as part of GA). There's significant benefit to ensuring that our code compiles without warnings or errors.


Assets/csc.rsp
```diff
@@ -0,0 +1 @@
+-warnaserror+
```
keveleigh: Is it possible to pass this via our msbuild scripts instead of adding this file?

wiwei: It is, though this means that people will catch this at PR CI time, rather than in their inner loop.

wiwei: https://www.reddit.com/r/Unity3D/comments/adfxna/unity_developer/

wiwei:
I ended up having to add CS1701 to a global ignore list, because:

error CS1701: Assuming assembly reference 'System.Numerics.Vectors, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' used by 'Windows.Foundation.UniversalApiContract' matches identity 'System.Numerics.Vectors, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' of 'System.Numerics.Vectors', you may need to supply runtime policy

It doesn't seem like you can reference a particular version for a given reference (i.e. referencing 4.0.0 of Vectors instead of 4.1.0)

wiwei:
Also ended up fixing new warnings that snuck since I started this PR, which highlights the ongoing need for this

wiwei:
The irony is that I checked in a change myself that caused a warnings as errors issue. (In between starting working on this and today)

Issue #4123, opened 2019-04-30
Toggle button has design and feature bugs


Troy-Ferrell:
## Describe the bug

The Interactable component can be utilized for Toggle button functionality and there is a toggle prefab available. However, there are two bugs: a design bug and a feature bug. 

Design:
To create a toggle button, one has to modify the dimensions to "2" and then they will get the "Can select" and "Can deselect" options available. Also it is not obvious that the additional theme field provided for the 2nd dimensions corresponds to the "selected" dimension state. 

Feature:
In order to validate whether the button is in the toggled state, one would think to use the "Interactable.IsToggled" property but this is only set by "Interactable.SetToggled" which is never called. Instead, the developer must call "Interactable.GetDimensionIndex == 1" which is not obvious.

## To reproduce

Look at Interactable.cs and the inspector

## Expected behavior

1) Configuring a button to be "toggleable" should be easier
2) SetToggled and IsToggled should be set as the user clicks on the button and selects/deselects

## Your Setup (please complete the following information)

- Unity Version 2018.3 or 2019.1
- MRTK Version RC1

## Target Platform (please complete the following information)

- HoloLens 2

## Additional context

Microsoft => trferrel

thedevleon:
Any updates to this? I spent a while trying to figure out why SetToggled(bool) would not change the visual state (dimension) before I found this. For now I created a workaround that simply sets the DimensionIndex to 0 or 1.

killerantz:
SetToggled(bool) only sets the toggled state, it does not change the dimensions, your workaround is the proper way at this point.

This is great feedback and similar to another issue (#3603) around easier access to toggle functionality and making sense of this system.

Thinking out loud for clarification, its valid to say when speaking about toggles, we would expect a two state system, so we would also expect that SetToggled(bool) would also control dimension index (setting to 0 or 1) based on the toggled state. Then we can look at Interactable.IsToggled to determine Toggle state value and dimension value (0 or 1) in this context.

On the issue of setting an initial checked or toggled state, having a Default Dimension field under Can Deselect would be fine?

Dimensions allow Interactable to go beyond two states, but toggles are a much more common use-case. I understand the confusion that comes with trying to be too flexible.

Proposal:
- Use Dimensions to enable Toggle functionality - Is a tooltip and documentation enough to discover this feature of someone is looking for toggle? Update field to say Dimensions (Toggle)
- When dimensions == 2, we will add "Toggled" to the second theme title
[...]

Issue #3603, opened 2019-03-11
No way to set Interactable / Checkbox to checked on start?


genereddick:
## Overview
I can't seem to find any way to set an Interactable / checkbox to start checked.

## Expected Behavior
Property that allows setting checkbox to checked

## Actual Behavior
Can't find property

## Steps to reproduce
_(Links to sample github project preferred)_

## Unity Editor Version
2018.3.8f1

## Mixed Reality Toolkit Release Version
vNext Beta 2

kolodi:
Achievable like this:
1. in `MixedRealityToolkit.SDK\Features\UX\Interactable\Scripts\Interactable.cs` script
add a field `public int DefaultDimension = 0;`
2. at the end of `Awake()` method set the dimension index like this: `SetDimensionIndex(DefaultDimension);`
3. to expose it in inspector add this 2 lines in `MixedRealityToolkit.SDK\Inspectors\UX\Interactable\InteractableInspector.cs` inside `RenderCustomInspector()` method:
```
SerializedProperty defaultDimension = serializedObject.FindProperty("DefaultDimension");
defaultDimension.intValue = EditorGUILayout.IntField(new GUIContent("Default Dimension", "Active dimension on start"), defaultDimension.intValue);
```

killerantz:
Created PR #5069 for potential fix. I know it's more than what we are asking here.
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5069

Yoyozilla:
fixed

ZAO29:
Finally, what is the property to know if a checkbox is checked ? or to check/uncheck it ?

killerantz:
I believe it's IsToggled.

PR #5069, opened 2019-06-26
Fix/toggle updates

killerantz:
## Overview
Addressing issues #3603 and #4123 
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/3603#event-2382237668

Most of this PR is outlined in issue #4123 
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4123#event-2412120239

Basically the Dimensions feature enables so advanced functionality, but there was not a way to set the startDimensionIndex. This is an easy fix as outlined in #3603.

But, the fact that many developers just want to use the toggle feature part of Dimensions, which seems more complicated than it should. So this PR attempts to also address this issue and a tiered approach to the complexity.

## Changes
Most of this is purely Inspector work, so the main functionality will still be the same and there should not be any breaking changes. The only potential issue is when Dimensions is set to 2 and someone calls Interactable.SetToggle(bool toggle). When Dimensions is set to 2 (toggle mode), then calling SetToggle will also call SetDimensionIndex(0/1). Calling SetDimensionIndex(0/1) to change the toggled state still works as it did before and developers can ignore the IsToggled state in the Interactable (they were never connected). The issue may be in the case where someone is calls both SetDiemsionIndex(0/1) and SetToggle(bool toggle), now SetDimensionIndex(0/1) will get called twice, but the outcome will be the same. When the Interactable is clicked, when in Toggle mode, IsToggled will be set now as well.

Now for the Inspector updates. We now have a drop down for Selection Mode. **These images are old and the new term is Selection Mode instead of Selection Type - but open to better suggestions**
image

The default value is Button which sets Dimensions to 1 behind the scenes.
Setting the drop down to Toggle will set Dimensions to 2.

image

There's also a number added to each themes with a "(Toggled)" appended to the second theme label.
The IsToggled checkbox will set the initial state at runtime. When the app is running, the IsToggled checkbox becomes read-only, but still reflects the toggled state. We don't want anyone overriding the toggle in the inspector when running, but it can be done through code at run-time by calling SetDiemensionIndex(0/1) or SetToggle(bool toggle).

Lastly, the Multi Dimensions option

image

In this mode we refer to Dimensions for everything, setting the start dimensions to at least 3 with the option to set the start Dimension index value. In this mode SetToggle(bool toggle) will not effect the Dimension index.

## Verification
- No breaking changes
- When in Multi Dimension mode, switching Dimensions to 1 or 2 will change the drop down to Button or Toggle respectively.
- Documentation should be clearer
- Should be easier to use and understand for basic functionality.
- Now validating Dimensions and StartDimensionIndex will not go out of range and break stuff when updated through the Inspector. Maybe we should check on start as well?


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -59,6 +59,11 @@ protected static IMixedRealityInputSystem InputSystem
         /// </summary>
         public HashSet<IMixedRealityInputSource> PressingInputSources => pressingInputSources;
 
+        /// <summary>
+        /// Values for user-friendly dimensions settings
+        /// </summary>
+        public enum SelectionTypes { Button, Toggle, MultiDimension };
```
julenka: Since the new name is SelectionMode, can you update the enum name also?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -93,6 +98,11 @@ protected static IMixedRealityInputSystem InputSystem
         /// </summary>
         public int Dimensions = 1;
 
+        /// <summary>
+        /// The Dimension value to set on start
+        /// </summary>
+        public int StartDimensionIndex = 0;
```
julenka: note: it's probably too much for this PR, but per the [Coding Guidelines](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/CodingGuidelines.html#access-modifiers) it would be nice to make these serialized and private, and then expose the fields via public properties.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -214,20 +214,91 @@ public virtual void RenderCustomInspector()
...
-            EditorGUILayout.PropertyField(dimensions, new GUIContent("Dimensions", "Toggle or sequence button levels"));
+            EditorGUI.BeginProperty(position, selectionTypeLabel, dimensions);
+            {
+                selectionType = (Interactable.SelectionTypes)EditorGUI.EnumPopup(position, selectionTypeLabel, (Interactable.SelectionTypes)(selectionTypeIndex));
+                selectionTypeIndex = (int)selectionType;
+
+                if (selectionTypeIndex >= minMultiDimensionValue - 1)
+                {
+                    // multi dimension mode
+                    if (dimensions.intValue < minMultiDimensionValue)
```
julenka: Why are you only updating dimensions.intValue if it's less than minMultiDimensionValue? Why not always update it?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: I've been thinking about this for a while now, I'll at least make the new stuff follow the guidelines.

killerantz:
Thanks @Troy-Ferrell for all the great feedback.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
killerantz: I moved the enum out to it's own file and added clarification on each value.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
killerantz: Great points, made updates as suggested.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -729,6 +740,12 @@ public virtual void SetToggled(bool toggled)
         {
             IsToggled = toggled;
             SetState(InteractableStates.InteractableStateEnum.Toggled, toggled);
+
+            // if in toggle mode
+            if (Dimensions == 2)
+            {
+                SetDimensionIndex(toggled ? 1 : 0);
+            }
```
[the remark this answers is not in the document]
killerantz: Added a log, hopefully this is clear enough. ```Debug.Log("SetToggled(bool) called, but SelectionMode is set to " + (SelectionModes)(selectedMode - 1) + ", so DimensionIndex was unchanged.");```

julenka:
> Basically the Dimensions feature enables so advanced functionality, but there was not a way to set the startDimensionIndex. This is an easy fix as outlined in #3603.

Is there a typo here? What does "so advanced functionality" mean here?

julenka:
@killerantz it looks like there are some compile errors in this PR, could you please fix? It actually looks like a warning, but we are now treating warnings as errors

-----CompilerOutput:-stdout--exitcode: 1--compilationhadfailure: True--outfile: Temp/Microsoft.MixedReality.Toolkit.SDK.dll
Microsoft (R) Visual C# Compiler version 2.9.1.65535 (9d34608e)
Copyright (C) Microsoft Corporation. All rights reserved.

Assets\MixedRealityToolkit.SDK\Features\UX\Interactable\Scripts\Interactable.cs(100,21): error CS0414: The field 'Interactable.StartDimensionIndex' is assigned but its value is never used
-----CompilerOutput:-stderr----------

killerantz:
Apologies for missing that and not testing properly.

julenka:
Here is how you can write tests and also run tests in MRTK [mrtk tests how-to](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/UnitTests.html)

HerrKatzen:
I created a temporary fix for checkboxes, only until you publish this fix.
Can be found here if anyone need it.
https://stackoverflow.com/questions/56972035/checkbox-not-working-as-intended-in-mixed-reality-toolkit-v2

Issue #4011, opened 2019-04-18
Intermittent GUILayout: Mismatched LayoutGroup.repaint when opening Interactable


junpark12:
## Describe the bug
After I import MRTK v2 RC1, created a scene. 
Simply drag PressableButton.prefab and hit run button.. i received the following Error.

"ArgumentException: GUILayout: Mismatched LayoutGroup.repaint"

## To reproduce

1. Import MRTK v2 RC1
2. create new scene and click MRTK> add to scene and config
3. Click play button in unity. 

## Screenshots

![image](https://user-images.githubusercontent.com/20918194/56331072-c0a4fb00-61c5-11e9-9846-c0448208ec93.png)

![image](https://user-images.githubusercontent.com/20918194/56331088-d1557100-61c5-11e9-8643-64f26c21cc0d.png)

## Your Setup (please complete the following information)
- MRTK V2 RC1
- Unity 2018.3.10f1

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

cre8ivepark:
@killerantz @Railboy Might be something related to the Interactable's inspector or Pressable Button's visual gizmo?

Railboy:
@cre8ivepark Possibly, but I can't repro this with either inspector active.

@junpark12 Can you try collapsing the inspector for each component on the button prefab and let us know when the warnings cease?

junpark12:
@Railboy  if i collapse all the item in inspector menu then no exception.. Thanks.

Railboy:
@junpark12 No problem. I was hoping to use that experiment to determine which inspector was the culprit. Sorry I wasn't clear. Could you try opening them one-by-one and let us know which inspector causes the errors?

killerantz:
These repaint errors are a pain to track down. I'll keep an eye out as I work on some press updates.

junpark12:
Acutally it randomly happens. sometime without collapsing the menu it worked and sometimes not..

julenka:
merged

PR #5070, opened 2019-06-26
updated InteractableInspector to fix mismatch repaint issue

killerantz:
## Overview
Fix for Issue #4011 

When initializing the Interactable the inspector does a check to see if the States property is not null. If it's null, then we try to fill it with the DefaultInteractableStates asset.

The issue came up when the States property was filled in the layout phase, but never rendered. During the repaint phase, now that the property is not null, we attempt to render the field in the inspector.

## Changes
We are now rendering the field in both phases, in the layout phase after the property is filled.

## Verification
I was able to consistently repro this by adding the Interactable component to a GameObject using the Add Component button in the inspector.

1 Create a new Scene
2 Add MRTK by selecting MRTK/Add to Scene and Configure... in the menu
3 Drag a pressable button into the scene
4 press play
5 observe no errors

or 

1 Create a game object in the scene
2 Select it and press Add Component in the inspector
3 Select Interactable as the component to add
4 observe no errors.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
+++ b/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
@@ -99,7 +99,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
                 drawerStarted = true;
                 if (showStates != prefsShowStates)
                 {
-                   EditorPrefs.SetBool(statesPrefKey, showStates);
+                    EditorPrefs.SetBool(statesPrefKey, showStates);
                 }
             }
             else
@@ -119,6 +119,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
                         }
                     }
 
+                    EditorGUI.indentLevel = indentOnSectionStart + 1;
                     showStates = InspectorUIUtility.DrawSectionStart(states.objectReferenceValue.name + " (Click to edit)", indentOnSectionStart + 2, prefsShowStates, FontStyle.Normal, false);
                     drawerStarted = true;
                 }
--- a/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
+++ b/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
@@ -89,16 +89,17 @@ namespace Microsoft.MixedReality.Toolkit.UI
             bool showStates = false;
             SerializedProperty states = serializedObject.FindProperty("States");
             bool drawerStarted = false;
+            string statesPrefKey = "Settings_States";
+            bool prefsShowStates = EditorPrefs.GetBool(statesPrefKey);
+
             if (states.objectReferenceValue != null)
             {
-                string statesPrefKey = "Settings_States";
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -135,7 +139,7 @@ public virtual void RenderCustomInspector()
             {
                 InspectorUIUtility.DrawSectionEnd(indentOnSectionStart);
             }
-
+
```
wiwei:
```suggestion

```

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -89,16 +89,17 @@ public virtual void RenderCustomInspector()
...
             {
-                string statesPrefKey = "Settings_States";
-                bool prefsShowStates = EditorPrefs.GetBool(statesPrefKey);
                 EditorGUI.indentLevel = indentOnSectionStart + 1;
                 showStates = InspectorUIUtility.DrawSectionStart(states.objectReferenceValue.name + " (Click to edit)", indentOnSectionStart + 2, prefsShowStates, FontStyle.Normal, false);
                 drawerStarted = true;
                 if (showStates != prefsShowStates)
                 {
-                    EditorPrefs.SetBool(statesPrefKey, showStates);
+                   EditorPrefs.SetBool(statesPrefKey, showStates);
```
wiwei:
```suggestion
                    EditorPrefs.SetBool(statesPrefKey, showStates);
```

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -89,16 +89,17 @@ public virtual void RenderCustomInspector()
             bool showStates = false;
             SerializedProperty states = serializedObject.FindProperty("States");
             bool drawerStarted = false;
+            string statesPrefKey = "Settings_States";
```
wiwei: Was it necessary to bring out these two statements into this scope? It looks like these are still only used inside the if clause of line 95

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
```diff
@@ -89,16 +89,17 @@ public virtual void RenderCustomInspector()
...
+            bool prefsShowStates = EditorPrefs.GetBool(statesPrefKey);
+
             if (states.objectReferenceValue != null)
             {
-                string statesPrefKey = "Settings_States";
-                bool prefsShowStates = EditorPrefs.GetBool(statesPrefKey);
                 EditorGUI.indentLevel = indentOnSectionStart + 1;
```
wiwei: Can you elaborate a bit more (on the PR, maybe not in comments in the code) about why an indent happens here but not down below?

killerantz: That's an oversight. I'll update it.

julenka:
uh oh, looks like boundingbox might have an unreliable test.

Issue #4546, opened 2019-05-23
BoundingBox needs to print error if maniphandler is above BoundingBox


julenka:
## Overview
This bug is a follow up issue #4393. That bug was closed by design and the requirement is devs always put the bounding box component before manipulation handler on a game object if they are on the same object.

To prevent people from accidentally putting manipulation handler above bounding box, we should have some validation to check that manipulation handler component comes after bounding box on a game object.

julenka:
Fixed by fixing original issue #4393

PR #5072, opened 2019-06-26
BoundingBox check that it is before maniphandler in components list

julenka:
## Changes
- Fixes: #4546 by adding a check in OnValidate

## Verification
- Checked that bug is fixed
- Verify that HandInteractionExampleScene works


julenka:
Closing in favor of approach in #5086 which doesn't require people put maniphandler and boundingbox in specific order, but instead has ManpulationHandler ignore colliders using InteractiveColliderContext component.

PR #5086, opened 2019-06-27
Add ability for some ManipulationHandler to ignore pointer events from gameobjects using InteractiveColliderContext

### LINKED ISSUE #4393: BoundingBox script doesn't work  when ManipulationHandler component is also added to the same gameobject

### LINKED ISSUE #4546: BoundingBox needs to print error if maniphandler is above BoundingBox

julenka:
This is an alternative approach for PR #5072 to fix #4393 (follow-up bugfix is #4546)

## Looking for feedback
I'm looking for feedback regarding the approach in this PR compares to PR #5072. Would love to hear if folks see issues with this approach. Also suggestions about naming the component.

## Overview
The root cause of #4393 is that when maniphandler is above boundingbox, maniphandler will swallow all pointer events originating from scale, rotate affordances before bounding box has a chance to handle it. This causes the scale, rotate affordances to be impossible to grab. The fix is to put bounding box before manip handler. This allows bounding box to handle the affordances first.

PR #5072 solves this by throwing error if bounding box is ever coming after manip handler.

An alternate approach is to somehow tell ManipulationHandler to ignore events from certain gameobjects. I considered using tags or layers but decided against it because those approaches did not feel extensible (layers only have 31 values so potentially customers could run out if we took one for layer "Ignore Manipulationhandler", and only one tag allowed per gameobject).

An alternative to tags or layers is to add a component I called `ManipulationHandlerContext` component that can provide configurable info to a manipulationhandler whenever that gameobject is grabbed.  On PointerDOwn, ManipulationHandler then checks if the object the pointer event came from has a ManipHandlerContext and ignores the event source if IgnoredByManipulationHandler == true. This allows us to have descendents that raise pointer events, but which can be ignored by manipulation handler (because they could be used by some other components).

BoundingBox then adds these context components to its affordances, and sets IgnoredByManipulationHandler = true. This then prevents manipulationhandler from stealing the events that come from the bounding box affordances.

## Verification
- Verified that this allows bounding box handles to be grabbed regardless of whether bounding box is above / below manipulationhandler


julenka:
Note: if I was to go with this approach I need to add docs in ManipulationHandler explaining ManipulationHandlerContext.

julenka:
Okay, I updated to use a more general "InteractiveColliderContext" component. Going to mark this as ready for review then and close the other PR, in favor of this one.

luis-valverde-ms:
What if instead we add a component to the rig root game object created by the bounding box that catches those events and dispatches them to the bounding box? I'd rather that to the context component approach because the latter forces event listeners (e..g manipulation handler) to check for the context while the former is completely transparent.

julenka:
> What if instead we add a component to the rig root game object created by the bounding box that catches those events and dispatches them to the bounding box? I'd rather that to the context component approach because the latter forces event listeners (e..g manipulation handler) to check for the context while the former is completely transparent.

That could work, let me try it out.

julenka:
> > What if instead we add a component to the rig root game object created by the bounding box that catches those events and dispatches them to the bounding box? I'd rather that to the context component approach because the latter forces event listeners (e..g manipulation handler) to check for the context while the former is completely transparent.
> 
> That could work, let me try it out and see.

Seems to be working! I like that solution a lot better, third times a charm :-) Will send that out shortly and closing this PR.

Issue #4801, opened 2019-06-10
Documentation links don't appear in profile inspector unless service is enabled


Railboy:
## Describe the bug

Doc links don't appear in a profile inspector until a service is enabled.

## To reproduce

- Select a MixedRealityToolkit instance and select the Input service.
- Toggle Enable Input System.
- Observe Documentation link appear / disappear.

## Expected behavior

Doc link should be visible whether a service is enabled or not.

## Screenshots

![Doc1](https://user-images.githubusercontent.com/9789716/59208888-e532a900-8b5e-11e9-9a86-2725d02bf0e3.PNG)

![Doc2](https://user-images.githubusercontent.com/9789716/59208891-e6fc6c80-8b5e-11e9-9a7c-04fe17d3ee51.PNG)

- Unity Version 2019.1
- MRTK Version 2.0 mrtk_development

PR #5093, opened 2019-06-27
Fix documentation links in inspectors

Troy-Ferrell:
## Overview
This change tries to simplify and fix some of the documentation link features in inspectors.

1) Update to optimize window to use documentation link button style & API
2) Fixed documentation link to be on profiles and not active systems in MRTK (which is problematic in editor mode)
3) Fixed create asset menu for MRTK profiles to be their own folder category
4) Remove duplicate utility inspector class and merged functionality to consolidate files

mrtk-profiles-create-menu

## Changes
- Fixes: #4801 

## Verification
Tested in MRTK inspector
Test in drill down profiles
tested in service facade inspector view
tested in optimize window view


Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityCameraProfileInspector.cs
```diff
@@ -1,99 +0,0 @@
...
-// Licensed under the MIT License. See LICENSE in the project root for license information. 
-
-using Microsoft.MixedReality.Toolkit.Utilities.Editor;
-using UnityEditor;
-using UnityEngine;
-
-namespace Microsoft.MixedReality.Toolkit.Editor
-{
-    [CustomEditor(typeof(MixedRealityCameraProfile))]
-    public class MixedRealityCameraProfileInspector : BaseMixedRealityToolkitConfigurationProfileInspector
```
wiwei: I didn't see this called out in the PR notes - was there some other camera profile inspector that already existed (so this was a duplicate?). Otherwise it looks like if you view the profile itself (outside of the MRTK object) you'll lose this functionality

PR #5094, opened 2019-06-27
Fixed pointer events tests to adapt to changes to PlayModeTestUtilities.ShowHand.

luis-valverde-ms:
`PlayModeTestUtilities.ShowHand` now steps an extra frame after showing hands, which changes the events the pointer handler receives so we need to adjust the expected values accordingly. Also, simplified interpolation math in `PlayModeTestUtilities.MoveHandFromTo`

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs
@@ -55,6 +55,10 @@ namespace Microsoft.MixedReality.Toolkit.Tests
                 Assert.AreNotSame(eventData.OldFocusedObject, eventData.NewFocusedObject);
                 Assert.AreEqual(state.numFocusChanged, state.numBeforeFocusChange);
 
+                // Uncomment to help debugging issues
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs
```diff
@@ -124,27 +133,45 @@ private PointerHandler SetupScene()
...
-        public IEnumerator TestOnePointer()
+        public IEnumerator Test()
         {
-            var handler = SetupScene();
-            var rightHand = new TestHand(Handedness.Right);
+            // Step once to receive focus events from gaze pointer
+            yield return null;
+
+            var expected = new PointerHandler.State();
+            expected.numBeforeFocusChange++;
```
lukastoenneMS: Use explicit numbers instead of increment to make clear what we are expecting here.

luis-valverde-ms: I think it is not too confusing to rely on the state being initialized to zero.

Issue #5095, opened 2019-06-27
Asset Dependency Tracking


Cameron-Micka:
## Describe the problem

When bringing in new assets, or examining old ones, it is difficult to understand which assets are in use or referenced from other assets. This leads to keeping assets around which are no longer in use.

## Describe the solution you'd like

I would like a tool that tells the user which assets are in use (and by what).

## Additional context

The shell team is about to bring in a large amount of UI assets that need to be combed over for which assets are still being used. This tool would prevent them from manually inspecting assets.

PR #5096, opened 2019-06-27
Dependency Window

Cameron-Micka:
## Overview
Often in Unity it is difficult to gleam which assets are being used, and what is referencing them. The "Find References in Scene" option works great when you are only concered with the current scene, but what about your entire Unity project? This is where the Dependency Window can be useful.

The  Dependency Window displays how assets reference and depend on each other. Dependencies are calculated by parsing guids within project YAML files (note, script to script dependencies are not considered).

DependencyWindow

The window displays a list of assets the currently select asset depends on, and a hierarchical list of assets that depend on it. If nothing depends on the currently selected asset you can consider deleting it from your project (note some assets are loaded programmatically via APIs like Shader.Find() and may not be caught by the dependency tracker).

The window can also display just a list of all assets which are not referenced by any other assets and could be considered for deletion:
List

Example Use:
DependencyWindow

The need for this tool was brought up when the Shell team mentioned they have many UI assets they wish to bring into the MRTK, but would like a non-manual way to determine which assets are being used.

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5095

## Verification
Please give this tool a try with your Unity project. I would be curious to know which dependencies it misses and how long it takes to build the dependency graph on your PC.

>
>


Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
```diff
@@ -0,0 +1,339 @@
...
+            var window = GetWindow<DependencyWindow>();
+            window.titleContent = new GUIContent(windowTitle, EditorGUIUtility.IconContent("d_EditCollider").image);
+            window.minSize = new Vector2(520.0f, 380.0f);
+            window.RefreshAssetGraph();
+            window.Show();
+        }
+
+        private void OnGUI()
+        {
+            MixedRealityEditorUtility.RenderMixedRealityToolkitLogo();
```
Troy-Ferrell: with PR #5093 , I am moving the RenderMixedRealityToolkit into MixedRealityInspectorUtility. We had duplicate files all doing editor utils functions

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
```diff
@@ -0,0 +1,339 @@
...
+
+            EditorGUILayout.LabelField("Mixed Reality Toolkit Dependency Window", new GUIStyle() {fontSize = 12, fontStyle = FontStyle.Bold });
```
Troy-Ferrell: Change this style from black to white text per your screenshot. Also maybe move GUIStyle to MixedRealityStyles.cs (something like that)

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
```diff
@@ -0,0 +1,339 @@
...
+                    RefreshAssetGraph();
+                }
+            }
+            EditorGUILayout.EndHorizontal();
+
+            EditorGUILayout.Space();
+
+            GUILayout.BeginVertical("Box");
+            {
+                GUILayout.Label("Dependency Graph", EditorStyles.boldLabel);
```
Troy-Ferrell: Can you to using EditorGUILayout instead of GUILayout? 1) to be consistent with other inspector/editor files and 2) EditoGUILayout generally has more features (with the very odd weird exception of GUILayout.Button)

https://docs.unity3d.com/ScriptReference/EditorGUILayout.LabelField.html

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
```diff
@@ -0,0 +1,339 @@
...
+        {
+            EditorGUILayout.BeginHorizontal();
+            {
+                GUILayout.Space(depth * 16);
+
+                if (depth != maxDepth)
+                {
+                    var path = AssetDatabase.GUIDToAssetPath(node.guid);
+
+                    if (GUILayout.Button("Select", GUILayout.ExpandWidth(false)))
```
Troy-Ferrell: Does EditorGUIlayout.ObjectField not work? Like you use earlier for AssetSelection. Or do you like showing the full path for the asset?
assetSelection = EditorGUILayout.ObjectField("Asset Selection", assetSelection, typeof(Object), false);

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
```diff
@@ -0,0 +1,339 @@
...
+            MixedRealityEditorUtility.RenderMixedRealityToolkitLogo();
+
+            EditorGUILayout.LabelField("Mixed Reality Toolkit Dependency Window", new GUIStyle() {fontSize = 12, fontStyle = FontStyle.Bold });
+            EditorGUILayout.LabelField("This tool displays how assets reference and depend on each other. Dependencies are calculated by parsing guids within project YAML files, code dependencies are not considered.", EditorStyles.wordWrappedLabel);
+
+            EditorGUILayout.Space();
+
+            if (EditorSettings.serializationMode != SerializationMode.ForceText)
+            {
+                EditorGUILayout.HelpBox("Dependencies can only be tracked with text assets. Please change the project serialization mode to \"Force Text\" via Edit->Project Settings->Editor->Asset Serialization", MessageType.Error);
```
Troy-Ferrell: Can you add a button to just launch the menu directly? I think this should work:
https://docs.unity3d.com/ScriptReference/EditorApplication.ExecuteMenuItem.html

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
```diff
@@ -0,0 +1,339 @@
...
+            }
+
+            assetSelection = EditorGUILayout.ObjectField("Asset Selection", assetSelection, typeof(Object), false);
```
Troy-Ferrell: Should we explicitly disable the gui (GUI.enabled = false) if the user is in play mode (EditorApplication.isPlaying || EditorApplication.isPauded)? Basically try to make sure we aren't doing any logic in playmode if this window happens to be visible so we don't kill the user's framerate. Sort of like #5003

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
```diff
@@ -0,0 +1,339 @@
...
+using UnityEditor;
+using UnityEngine;
+
+using Object = UnityEngine.Object;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
+{
+    public class DependencyWindow : EditorWindow
+    {
+        private Object assetSelection = null;
```
Troy-Ferrell: Have you tried showing and then close tab the window? Do all variables reset to defaults?

Troy-Ferrell:
Also expanding on my original comment, maybe add a tab that shows you assets with absolutely no dependencies in the project?

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
Cameron-Micka: Ah, thank you for the info. Updated!

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
Cameron-Micka: Good idea!

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
Cameron-Micka: I switched over as many as I could (sans Button, parametrized Space, and a few Labels that require special alignment). That's so weird their isn't a EditorGUILayout Button?

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
Cameron-Micka: I didn't even think of this. Great suggestion! Switched it over.

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
Cameron-Micka: I'm not sure I can with a project setting? But, I just added a "Fix Now" buttoon like you find in the material inspector.

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
Cameron-Micka: Didn't even think of this, added!

Assets/MixedRealityToolkit/Utilities/Editor/DependencyWindow.cs
Cameron-Micka: When you close the window it destroys the DependencyWindow and hence the dependencyGraph, so everything is restored to the defaults when you re-open the window.

Cameron-Micka:
> Also expanding on my original comment, maybe add a tab that shows you assets with absolutely no dependencies in the project?

Thank you for the geat suggestions! I like this idea. Let me see if I can add that tab real quick.

julenka:
> I would seriously love to use this on MRTK itself and remove any old assets not in use. Examples package in particular is really bloated

I Created #5296

PR #5101, opened 2019-06-27
Fix the broken .NET-C# projects build.

wiwei:
This is an oldie but goldie. There's an issue that arises only when building on the .NET backend with C# projects enabled (this is a deprecated scenario but still one that is really useful for fast iteration).

The problem is that even if an assembly is marked as a "test assembly" (as the play mode tests are), they will still be included in player builds for .NET-C# project builds. This is what the old comment was alluding to (but wasn't very specific about)

We also explored doing things like marking the play mode tests as editor only (i.e. matching what the edit mode test asmdef looks like) but this had the side effect of removing all tests from the test runner (bad, tests are gone). Also tried making things editor + Win32 (standalone) + Win64 (standalone) only as well, though this still had the same problem even when building UWP .NET

I updated the comment to be a little clearer about what's going on here and also updated the test guidance as well.

Lastly, got rid of some unused using statements while I was in there.

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
@@ -3,24 +3,21 @@
 
 #if !WINDOWS_UWP
 // When the .NET scripting backend is enabled and C# projects are built
-// Unity doesn't include the required assemblies (i.e. the ones below).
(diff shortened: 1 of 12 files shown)


wiwei:
Note that the actual fix is part of TestHand.cs

Everything else is about updating the comment for a little more clarity.

wiwei:
It looks like the build is still broken due a test that didn't work - I'll fix that next

PR #5102, opened 2019-06-27
Fix the broken NuGet build.

wiwei:
It's been a few days of us not building NuGet packages - here's what happened.

We hit a CI outage earlier in the week caused by an upstream Azure DevOps issue (a bad nuget push package was pushed out). Our build was broken for a couple of days, during which we still had changes going in (mostly we were judging by green-ness of mrtk_pr, which is identical to CI except it doesn't produce and publish NuGet packages).

The problem is, without coverage on the NuGet building, we didn't know that we broke the nuget packages until the upstream nuget push package got fixed.

The asset retargeter assumes that every single class/object in the MRTK is prefixed by the MRTK namespace. This is a fairly reasonable assumption to make, and it throws when it finds something that violates it (so that we can then fix it). The fix is thus to make sure that the class in question is put into an MRTK namespace (the example demos one)

I will follow up here and do work to make sure that the build breaks when this happens, instead of just showing warnings.

### DIFF
--- a/Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
+++ b/Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
@@ -1,49 +1,51 @@
-﻿using System;
-using Microsoft.MixedReality.Toolkit;
+﻿using Microsoft.MixedReality.Toolkit;
 using Microsoft.MixedReality.Toolkit.Input;
 using UnityEngine;
 
-// Simple example script that subscribes to primary pointer changes and applies a cursor highlight to the current one.
-public class PrimaryPointerHandlerExample : MonoBehaviour
+namespace Microsoft.MixedReality.Toolkit.Examples.Demos
 {
-    public GameObject CursorHighlight;
-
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/PrimaryPointer/PrimaryPointerHandlerExample.cs
```diff
@@ -1,49 +1,51 @@
-﻿using System;
-using Microsoft.MixedReality.Toolkit;
+﻿using Microsoft.MixedReality.Toolkit;
```
keveleigh: Missing copyright header too

wiwei: UwU

wiwei: ¯\_(ツ)_/¯

wiwei: (☞ﾟヮﾟ)☞

PR #5103, opened 2019-06-27
Ensure that asset retargeting fails on non-zero Unity exit codes.

wiwei:
We've already seen a number of cases where we have failed to push NuGet packages, because of classes that were introduced with no namespaces. These show up as an error in the asset retargeting step, but because we don't return the error code that Unity gives us, the rest of the build moves on (and then later shows a warning saying that "no nuget packages were pushed out")

This updates the build powershell script so that any break that would cause us to stop producing a nuget package would actually cause a build failure, rather than just a warning and slap on the wrist.


wiwei:
> Are we able to do the same thing for the Unity app build steps?

Yes, though I hadn't done that because it's a little more complicated - though I guess not too complicated. There are a few other changes that need to be made (i.e. we still want to dump out the msbuild logs in the case that it's an msbuild case so that it's still debuggable).

wiwei:
It turns out, more things were broken, I have to fix those before I submit this.

Issue #4997, opened 2019-06-20
MRTK inspector tab selection is not preserved when entering playmode


Railboy:
## Describe the bug

When entering play mode with an MRTK instance selected, the inspector's tab selection resets.

## To reproduce

- Create a new scene
- Add an MRTK instance and select it
- Select 'Spatial Mapping' tab
- Enter play mode
- Observe tab selection default back to 'Camera'

## Expected behavior

Tab selection should be preserved when entering play mode.

Railboy:
@Troy-Ferrell is this something you're already working on? If not I have some ideas.

Troy-Ferrell:
No I haven’t. We can add a saved state

PR #5118, opened 2019-06-28
Add MRTK inspector preferences 

Troy-Ferrell:
## Overview
This change adds MRTK editor preferences to save the view state of certain foldouts. This way when user enters play mode or inspector refreshes, it doesn't repaint to default state. 

Also fixed missing issue for input system that causes MRTK to reset when showing/hiding input foldouts which become considered as "changes"

## Changes
- Fixes: #4997 

## Verification
Enter/exit playmode
Drill down into input profile


keveleigh:
Are these per-project or global prefs?

Troy-Ferrell:
@keveleigh these should just be global preferences for a local user on a particular Unity version as they are saved in the Windows registry. I saw the EditorPreferences class which wraps around EditorPrefs Unity class. But I don't care for saving a unique key for every product name. This preference isn't shared between developers or in the git repo. If a developer is switching between projects and a particular fold out is still shown, that's fine. I'd rather not create a million keys and have it just be shared across the environment. It's a short term lived use (like between play mode & edit mode). Thanks for the callout though!

Issue #5119, opened 2019-06-28
OnFocus events are not routed through global listeners


Alexees:
## Describe the bug

The IMixedRealityFocusHandler interface can be implemented and registered with the InputSystem.RegisterHandler method. Yet nothing is received because the FocusProvider directly calls the receiver instead of passing through the regular eventchain.
(I mentioned it in #3607 and forgot about it)

## To reproduce

1. Register any IMixedRealityInputHandler implemeter with the InputSystem and gaze on something
2. Note that you do not receive anything

## Expected behavior

You can listen to all events you are allowed to register the corresponding interface for.

## Your Setup (please complete the following information)

- Unity Version 2018.4.2f1
- MRTK Version mrtk_development 27.06.

## Target Platform (please complete the following information)

- Editor

wiwei:
Closing this as the linked PR is in

PR #5120, opened 2019-06-28
Route all Focus Events through regular EventSystem

Alexees:
This implementation routes all 4 Focus events through the regular InputSystem logic, which means they can all be listened to from global listeners.

These event calls have been the only calls that had their own direct calls to ExecuteEvents.ExecuteHierarchy. Calls to ExecuteEvents.ExecuteHierarchy are now reduced to where they belong:

- DispatchEventToFallbackHandlers
- DispatchEventToObjectFocusedByPointer

This implementation adds one more field to the FocusEventData and distinguishes in the MixedRealityInputSystem between FocusEventData and BaseInputEventData, because they do not have a common way of telling who receives the Input.

This change is not breaking any interfaces besides

- Fixes: #5119

Cons: if eventdata gets used on the way, focusing breaks which was formerly not possible to do. Maybe in the case of FocusEventData, used should be ignored.


Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -468,10 +485,19 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
...
-        private bool DispatchEventToObjectFocusedByPointer<T>(IMixedRealityPointer mixedRealityPointer, BaseInputEventData baseInputEventData,
+        private bool DispatchEventToObjectFocusedByPointer<T>(IMixedRealityPointer mixedRealityPointer, BaseEventData baseEventData,
             bool modalEventHandled, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
         {
-            GameObject focusedObject = FocusProvider?.GetFocusedObject(mixedRealityPointer);
+            GameObject focusedObject;
+            var focusEventData = baseEventData as FocusEventData;
+            if (focusEventData != null)
+            {
+                focusedObject = focusEventData.NewFocusedObject == null ? focusEventData.OldFocusedObject : focusEventData.NewFocusedObject;
```
Alexees: @luis-valverde-ms @keveleigh I just realized that this is not distinct code. If you're going to review anyways it would be nice to know if this whole thing is even the right way to do this.

Alexees: I fixed that issue.

Alexees:
I found that instead of adding to the FocusEventData, I could utilize the selectedObject field of the BaseEventData. Thoughts?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -468,10 +485,12 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
...
         /// Returns true if the event was handled by a modal handler
         /// </summary>
-        private bool DispatchEventToObjectFocusedByPointer<T>(IMixedRealityPointer mixedRealityPointer, BaseInputEventData baseInputEventData,
+        private bool DispatchEventToObjectFocusedByPointer<T>(IMixedRealityPointer mixedRealityPointer, BaseEventData baseEventData,
             bool modalEventHandled, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
         {
-            GameObject focusedObject = FocusProvider?.GetFocusedObject(mixedRealityPointer);
+
+            var focusEventData = baseEventData as FocusEventData;
+            GameObject focusedObject = focusEventData != null ? focusEventData.FocusReceiverObject : FocusProvider?.GetFocusedObject(mixedRealityPointer);
```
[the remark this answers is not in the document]
Alexees: That's handled in the FocusReceiver. I did not come up with a better name, but focusEventData.FocusReceiverObject is basically the gameobject that was either entered or exited depending on which delegate is currently used. These originate from the FocusProvider which called the events directly on them.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -938,7 +961,7 @@ public void RaiseFocusExit(IMixedRealityPointer pointer, GameObject unfocusedObj
         {
             focusEventData.Initialize(pointer);
 
-            ExecuteEvents.ExecuteHierarchy(unfocusedObject, focusEventData, OnFocusExitEventHandler);
+            HandleEvent(focusEventData, OnFocusExitEventHandler);
```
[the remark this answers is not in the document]
Alexees: unfocusedObject is passed along with the FocusEventData.FocusReceiverObject , so yes.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -379,29 +379,47 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
...
                 return;
             }
 
-            Debug.Assert(baseInputEventData.InputSource.Pointers != null, $"InputSource {baseInputEventData.InputSource.SourceName} doesn't have any registered pointers! Input Sources without pointers should use the GazeProvider's pointer as a default fallback.");
-
             var modalEventHandled = false;
-            // Get the focused object for each pointer of the event source
-            for (int i = 0; i < baseInputEventData.InputSource.Pointers.Length && !baseInputEventData.used; i++)
+
+            var baseInputEventData = eventData as BaseInputEventData;
```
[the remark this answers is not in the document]
Alexees: I don't, but I would guess the difference is that BaseInputEventData can hold InputSources which have multiple pointers. The FocusProvider on the other hand only handles registered pointers and can not provide the InputSource for the BaseInputEventData, hence uses the FocusEventData.
Fun fact: IMixedRealityFocusProvider provides all its implementation methods by pointer, which the Toolkit does not use anywhere. The only implementer of it is the FocusProvider and it registeres pointers by InputSource.
If it wasn't for the interface, there wouldn't be the need for a FocusEventData at all.
Only other issue would be that FocusExit needs the OldFocusedObject somehow, but this could go into BaseEventData selectedObject, which is also not used at all at the moment (assigment wise #5139).

Alexees:
> Also, you may want to run the unity playmode tests in the latest mrtk_development, because we have tests that can help with verifying if this change might break existing focus behavior. You can also run the following command from powershell to run the playmode tests in batch mode, which is how then run on our CI server
I did, but I do not know what to expect from it. The resulting log is quite long.

julenka:
> I found that instead of adding to the FocusEventData, I could utilize the selectedObject field of the BaseEventData. Thoughts?

It would be nice to not introduce another variable if we don't need to, but a possible point of confusion is nothing else in MRTK uses the selectedObject field. My evidence: I've noticed that selectedObject value is null for all OnPointer* events.

Unity's docs about selectedObject say "The object currently considered selected by the EventSystem." I'm not sure that you're using it in the same way, since you say above:

> That's handled in the FocusReceiver. I did not come up with a better name, but focusEventData.FocusReceiverObject is basically the gameobject that was either entered or exited depending on which delegate is currently used. These originate from the FocusProvider which called the events directly on them.

I am not sure that you're using FocusReceiverObject in the same way selectedObject is intended to be used.

Therefore, I'd suggest to keep the new field and not use selectedObject. @luis-valverde-ms might have some thoughts as well.

julenka:
> > Also, you may want to run the unity playmode tests in the latest mrtk_development, because we have tests that can help with verifying if this change might break existing focus behavior. You can also run the following command from powershell to run the playmode tests in batch mode, which is how then run on our CI server
> 
> I did, but I do not know what to expect from it. The resulting log is quite long.

There should be a testresults.xml file that gets output. On the first line you will see number of passed and failed. You're right though it's not user friendly. The more user friendly way is to run the playmode tests directly from the editor. I think @thalbern wrote a guide for this.

thalbern:
> > > Also, you may want to run the unity playmode tests in the latest mrtk_development, because we have tests that can help with verifying if this change might break existing focus behavior. You can also run the following command from powershell to run the playmode tests in batch mode, which is how then run on our CI server
> > 
> > 
> > I did, but I do not know what to expect from it. The resulting log is quite long.
> 
> There should be a testresults.xml file that gets output. On the first line you will see number of passed and failed. You're right though it's not user friendly. The more user friendly way is to run the playmode tests directly from the editor. I think @thalbern Bernadette Thalhammer FTE wrote a guide for this.

test doc is here:
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/UnitTests.html

luis-valverde-ms:
Thanks giving this a try, @Alexees . +1 to what @googlan was saying. Also, I don't think we need the focus receiver in the event, it can be obtained from the pointer.Result.

Alexees:
> Thanks giving this a try, @Alexees . +1 to what @googlan was saying. Also, I don't think we need the focus receiver in the event, it can be obtained from the pointer.Result.

The reason for passing the focus receiver is that in the case of FocusExit it would be the pointer.Result.PreviousPointerTarget. It's the only event I think that does not use CurrentPointerTarget. Do you know a better way to handle this special case?

Alexees:
> If I understood correctly, the issue to solve was an ability to receive focus events globally.

That's correct
> My first concern is - do we want a global handler for focus events to "intercept" an event and prevent dispatching it to actual focused object? This will happen now after `DispatchEventToGlobalListeners` call, because event data will be used, and `HandleEvent` will early out. This is expected logic for usual input events (e.g. speech).

It must not be intercepted. This is not implemented yet, I mentioned this missing at the end of my explanation as a disadvantage of the current implementation

> they shouldn't benefit on traversal of `modalInput`.
I have no understanding of modal handlers yet, so I guess you're right that they should be ignored

> We're also guaranteed that `DispatchEventToFallbackHandlers` won't be called for focus events because focus receiver object is not null.

I don't understand this one. Do you mean it should be guaranteed, or it is and shouldn't be.
> 
> I think that it would be better to avoid using `HandleEvent` for focus events and instead have a separate, simpler `DispatchFocusEvent` (or similar) method, which will
> 
> * call `DispatchEventToGlobalListeners` for a given focus event -- this will work as expected.
> * without checking of `eventData.used` call `ExecuteHierarchy` on a given focused object.
> * `focusReceiver` object can become method parameter and doesn't have to be stored in `eventData`.
> 
> This way we will keep implementation of `HandleEvent` simpler (it is already quite complex, and `ifs` based on eventData type don't make it nicer) and will provide support for global focus handlers without adding performance cost or unwanted side effects.

Totally agree here. My implementation tries to not alter any interface 'cause... you know... you never know what companies or Microsoft folks allow or not.

luis-valverde-ms:
> The reason for passing the focus receiver is that in the case of FocusExit it would be the pointer.Result.PreviousPointerTarget. It's the only event I think that does not use CurrentPointerTarget. Do you know a better way to handle this special case?

I'd say that because you are receiving the event via `OnFocusExit` you know that the game object the event refers to is the one `PreviousPointerTarget`.

googlan:
> It must not be intercepted. This is not implemented yet, I mentioned this missing at the end of my explanation as a disadvantage of the current implementation
Ok. I mist have misread this part. In this case having separate `Dispatch` method should help.

> I don't understand this one. Do you mean it should be guaranteed, or it is and shouldn't be.
> 
Sorry, my bad. I meant that it is guaranteed in your implementation, meaning that some logic of `HandleEvent` is not required for focus events at all.

> Totally agree here. My implementation tries to not alter any interface 'cause... you know... you never know what companies or Microsoft folks allow or not.
You are correct about interfaces, but by 'interfaces' we mean the abstract ones, which can potentially be reimplemented by our users. Adding methods (in this case a non-public method) to the particular implementation is absolutely fine, as it doesn't break existing code.

Alexees:
> > The reason for passing the focus receiver is that in the case of FocusExit it would be the pointer.Result.PreviousPointerTarget. It's the only event I think that does not use CurrentPointerTarget. Do you know a better way to handle this special case?
> 
> I'd say that because you are receiving the event via `OnFocusExit` you know that the game object the event refers to is the one `PreviousPointerTarget`.

I was referring to DispatchEventToObjectFocusedByPointer, which needs to know about the actual receiver in the case of FocusExit  but cannot any other way than passing it along with the eventdata.
The previous implementation FocusProvider?.GetFocusedObject would return the CurrentPointerTarget only.

Alexees:
> correct about interfaces, but by 'interfaces' we mean the abstract ones, which can potentially be reimplemented by our users. Adding methods (in this case a non-public method) to the particular implementation is absolutely fine, as it doesn't break existing code.

If that's the case let me have another go :)

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -843,17 +868,18 @@ public void RaiseSourcePoseChanged(IMixedRealityInputSource source, IMixedRealit
         /// <inheritdoc />
         public void RaisePreFocusChanged(IMixedRealityPointer pointer, GameObject oldFocusedObject, GameObject newFocusedObject)
```
luis-valverde-ms: With the current implementation it seems that we can dispatch the same event up to three times to global event listeners. We should do that just once, at the beginning of the function. Same goes for `RaiseFocusChanged`.

googlan: It would depend on a usecase we want to support. These 3 events are invoked on different objects. Would we ever want information about focus receiver passed to a global listener? E.g. some debug inspector for focus events might want to receive all 3 copies of an event.

Alexees: I changed this to being called just once on the global listeners. Since the FocusEventData effectively contains all 3 possible receivers, you can read all three if you want.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -405,6 +405,57 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
             }
         }
 
+        /// <summary>
+        /// Handles a pre and post focus events
+        /// We send all focus events to all global listeners and the actual focus change receivers. the use flag is completely ignored to avoid any interception.
+        /// </summary>
+        public void HandleEvent<T>(FocusEventData focusEventData, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
```
luis-valverde-ms: Can this be made private? Maybe also name it HandleFocusChangedEvents for added clarity.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -405,6 +405,57 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
...
+                    // ignored.
+                }
+            }
+        }
+
+        /// <summary>
+        /// Handles focus enter and exit
+        /// We send the focus event to all global listeners and the actual focus change receiver. the use flag is completely ignored to avoid any interception.
+        /// </summary>
+        public void HandleEvent<T>(GameObject focusReceiver, FocusEventData focusEventData, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
```
luis-valverde-ms: Make it private and name it HandleFocusEvent?

luis-valverde-ms: I find `focusReceiver` a confusing name when this is used to handle a FocusExit event. What about `eventTarget`?

luis-valverde-ms:
Looking fine overall except for my few comments above!

Alexees:
@luis-valverde-ms applied all your suggestions.
Additionally changed the two new methods from being generic to what their only purpose interface is.

googlan:
@Alexees  it looks like last merged to your branch revision of mrtk_development had broken tests (I _think_ that's the related build break https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=2790). 
Can I please ask you to merge the latest, and I will schedule CI tests again?

Alexees:
> @Alexees it looks like last merged to your branch revision of mrtk_development had broken tests (I _think_ that's the related build break https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=2790).
> Can I please ask you to merge the latest, and I will schedule CI tests again?

@googlan I'm sorry, I'm completely at a loss here. I have never seen an Azure Pipeline log and it does not seem easy to understand it, at least I could not find where that failed. And I absolutely don't know what you mean by "merge the latest". What shall I merge and of what do I merge the latest?

googlan:
@Alexees Sorry, I'll clarify :)
Your change is based on 'mrtk_development' branch. You added this commit earlier, about 3 days ago: 
"Merge branch 'mrtk_development' of https://github.com/Microsoft/MixedRealityToolkit-Unity into mrtk_focusGlobalListener'".
At the moment of this merge, CI build on 'mrtk_dev' branch was broken (one test wasn't passing). This has been fixed shortly after, but this fix is not in your branch yet. What I'm asking is to merge 'mrtk_development' into your branch again, so it receives fix for a broken test. I called it 'merge the latest' and it didn't occur to me that it doesn't make sense without prior explanation. Does it make sense  now?

I posted link to Azure log for others' to reference to avoid further questions -- it's not related to your changes. I brought all this up because I've seen failure in the exact same test in previous CI attempt on your branch (it was scheduled by Julia). 

I'll explain a bit about Azure, if it's not interesting for you, then don't read any further.

CI logs are not easy to understand indeed, because this particular CI build is quite complex. However you don't have to read the log, if the failing (red) build step is "Publish Test Results". In this case you can see details on a separate tab:
https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=2790&view=ms.vss-test-web.build-test-results-tab
Here you can see that 103 tests passed and 1 has failed. You can click the test itself and it will show you on the right side which assert has failed.

Alexees:
@googlan I got the part with the dev branch (did update) and will check out the CI logs again for sure :)

Issue #5124, opened 2019-06-28
Scene system profile inspector doesn't reflect state of build settings


Railboy:
## Describe the bug

Scene system profile inspector doesn't reflect when scenes are in build settings unless you manually refresh it.

## To reproduce

 - Open a scene system profile.
 - Open the build settings window.
 - Add / remove a scene included in the profile from build settings window.
 - Observe no change in the profile.

## Expected behavior

Scene system profile should always reflect build settings.

Issue #4516, opened 2019-05-22
PressableButtonInspector marks the asset as "dirty" on every draw


lukastoenneMS:
## Describe the bug

The PressableButtonInspector flags the target asset as dirty in every OnInspectorGUI draw call:
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/9abf0a1eac906c7ffae688ffd5b5d125202f4ca7/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs#L299

## Expected behavior

Inspector should only mark the asset dirty when it actually changes settings. Apart from that it is also recommended by Unity to use Undo.RecordObject instead: https://docs.unity3d.com/ScriptReference/EditorUtility.SetDirty.html

lukastoenneMS:
This causes Unity to constantly save the scene while a button object is selected.

kmoverall:
I have also seen this bug on all classes that use derivatives of the BaseLineDataProviderInspector (Base, Ellipse, Simple, ParabolicConstrained, ParabolaPhysical, Rectangle, Bezier, and SplineLineDataProviderInspector)

Railboy:
@kmoverall @lukastoenneMS If you open a separate issue for the line classes I'll clean them up. They should be using callbacks.

PR #5125, opened 2019-06-28
Fixes scene system profile inspector to better reflect state of build settings

Railboy:
## Changes
- Fixes: #5124

### DIFF
--- a/Assets/MixedRealityToolkit.Services/SceneSystem/MixedRealitySceneSystemEditorOperations.cs
+++ b/Assets/MixedRealityToolkit.Services/SceneSystem/MixedRealitySceneSystemEditorOperations.cs
@@ -164,6 +164,8 @@ namespace Microsoft.MixedReality.Toolkit.SceneSystem
             EditorSceneManager.newSceneCreated += EditorSceneManagerNewSceneCreated;
             EditorSceneManager.sceneOpened += EditorSceneManagerSceneOpened;
             EditorSceneManager.sceneClosed += EditorSceneManagerSceneClosed;
+
+            EditorBuildSettings.sceneListChanged += EditorSceneListChanged;
         }
 
         private void EditorUnsubscribeFromEvents()
@@ -175,6 +177,8 @@ namespace Microsoft.MixedReality.Toolkit.SceneSystem
             EditorSceneManager.newSceneCreated += EditorSceneManagerNewSceneCreated;
             EditorSceneManager.sceneOpened -= EditorSceneManagerSceneOpened;
             EditorSceneManager.sceneClosed -= EditorSceneManagerSceneClosed;
+
+            EditorBuildSettings.sceneListChanged -= EditorSceneListChanged;
         }
 
         #endregion
@@ -202,6 +206,13 @@ namespace Microsoft.MixedReality.Toolkit.SceneSystem
         private void EditorApplicationProjectChanged()
         {
             buildSettingsDirty = true;
+        }
+
+        private void EditorSceneListChanged()
+        {
+            buildSettingsDirty = true;
+
+            EditorCheckForChanges();
         }
 
         private void EditorSceneManagerSceneClosed(Scene scene) { activeSceneDirty = true; }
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit/Inspectors/PropertyDrawers/SceneInfoDrawer.cs
```diff
@@ -1,10 +1,7 @@
-﻿using System;
-using Boo.Lang;
```
wiwei: >Boo.Lang; [](start = 6, length = 9)

What

Assets/MixedRealityToolkit.Services/SceneSystem/MixedRealitySceneSystemEditorOperations.cs
```diff
@@ -641,40 +655,42 @@ private void EditorUpdateBuildSettings()
...
             foreach (SceneInfo contentScene in profile.ContentScenes)
             {
-                changedScenes |= EditorSceneUtils.AddSceneToBuildSettings(
+                if (EditorSceneUtils.AddSceneToBuildSettings(
                     contentScene,
                     cachedBuildScenes, 
-                    EditorSceneUtils.BuildIndexTarget.None);
+                    EditorSceneUtils.BuildIndexTarget.None))
+                {
+                    cachedBuildScenes = EditorBuildSettings.scenes;
```
wiwei: >cachedBuildScenes = EditorBuildSettings.scenes; [](start = 20, length = 47)

The previous thing would set the cached build scenes at the very end (instead of setting it on each iteration) - curious to know about why we went from reconciling/snapping the state of the array at the end, vs incrementally building it (and thus, calling it multiple times as it gets increasingly larger)

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealitySceneSystemProfileInspector.cs
```diff
@@ -213,6 +213,9 @@ public override void OnInspectorGUI()
             });
 
             serializedObject.ApplyModifiedProperties();
+
+            // Keep this inspector perpetually refreshed
+            EditorUtility.SetDirty(target);
```
wiwei: >SetDirty [](start = 26, length = 8)

Just to check - this doesn't cause us to spike Unity CPU usage when this inspector is open right? I have flashbacks to the build window

Assets/MixedRealityToolkit/Inspectors/PropertyDrawers/SceneInfoDrawer.cs
wiwei: Boo.Lang!?

wiwei: image

Assets/MixedRealityToolkit.Services/SceneSystem/MixedRealitySceneSystemEditorOperations.cs
Railboy: This was working fine if only one scene was being added at a time, but it would get confused if multiple scenes were being added. Later parts of the loop would be working off old data.

Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealitySceneSystemProfileInspector.cs
Railboy: Not in this case but yes that can absolutely be a problem so this technique should be used sparingly.

keveleigh: There was a report that using this constantly, even when changes aren't being made, marks the scene as changed and needing to be saved: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4516, which seems annoying

Railboy: In this case it only marks the profile as dirty and not the scene. It must be because it's a ScriptableObject asset and not an actual prefab.

Assets/MixedRealityToolkit/Inspectors/PropertyDrawers/SceneInfoDrawer.cs
wiwei: I think there might be some money in a github app that finds usage of Boo.Lang and posts pictures of Boo

PR #5127, opened 2019-06-28
Homogenize Script Asset Icons Round 2

Railboy:
## Overview
Updates meta files to use consistent script icons.
Also adds a 'Test' script icon to mark our testing scripts.

Follow-up to #4059


wiwei:
One thought here, based on feedback we got from customers, is that the fact that we have custom CS script icons is actually super confusing for people who are used to stock Unity.

It might be MORE appropriate to do a pass where we don't do custom icons for .cs files, so that they can look like regular scripts that Unity devs are used to.

Railboy:
@wiwei I'm up for either option, but while we're deciding we should keep things consistent.

wiwei:
> @wiwei I'm up for either option, but while we're deciding we should keep things consistent.

Sounds good to me

wiwei:
> > @wiwei I'm up for either option, but while we're deciding we should keep things consistent.
> 
> Sounds good to me

The consistency argument will win me over 99% of the time.

Issue #4636, opened 2019-05-30
Spatial mapping simulation in Editor


danielescudero:
When developing/designing apps that use spatial mapping, it is required to deploy in order to test. So developers would like to do some easy spatial map simulation in editor.

Ideally dragging a prefab in the scene to test a sample spatial mesh and ensure their renderers are working correctly/right material, basic hit / raytest tests to make sure it works, etc. Without having to build to device.

Ideally we can simulate:
- Refresh rate
- Renderers
- Office room, small/medium/large areas
- Spatial understanding
- Use developer-generated meshes

keveleigh:
Are you able to use Unity's built-in room simulation to get a mesh?

https://docs.unity3d.com/Manual/wmr_testing.html#SimulateInEditor

keveleigh:
Also, related to #3535

PR #5128, opened 2019-06-28
[Customer request][HTK parity] Add 3D model spatial mesh observer for use in editor

### LINKED ISSUE #3535: Need ability to load (in Unity) saved spatial mesh data (HoloToolkit parity)
## Overview
HoloToolkit provided a mechanism to read previously scanned spatial mesh data (custom data fomat) from the disk while in the Unity editor. MRTK should have a similar feature and also support loading .obj files that can be created via Windows Device Portal.

davidkline-ms:
This PR introduces the SpatialObjectMeshObserver for the Spatial Awareness System. It supports sending meshes from 3D models to the spatial awareness system while running in the editor, allowing for rapid iteration of features such as programmatic object placement.

Also added is an asset importer for the HoloToolkit .room file format to enable re-use of scans created with the remote / file observer.

Model formats used while testing the observer.
- .room
- .obj
- .gltf
- .glb

Other related changes:
- Add optional object parent parameter to SpatialAwarenessMeshObject.Create
- Fix issue in WindowsMixedRealitySpatialMeshObserver where meshes were not properly parented (takes advantage of the new parameter)

Changes tested in editor (confirm new observer behavior) and on HoloLens (confirm existing observer behavior).

Fixes #3535, #4636


davidkline-ms:
looks like some ux tests may have been broken by the simulated spatial mesh being in front of some components. removing the new observer from the default profile

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserverProfile.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using Microsoft.MixedReality.Toolkit.SpatialAwareness;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SpatialObjectMeshObserver
+{
+    /// <summary>
+    /// Configuration profile for the spatial object mesh observer.
+    /// </summary>
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Object Mesh Observer Profile", fileName = "ObjectMeshObserverProfile", order = 1000)] // todo
```
keveleigh: >// todo

Still todo?

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/RoomFileFormat/RoomFileSerializer.cs
```diff
@@ -0,0 +1,218 @@
...
+        private static int HeaderSize = sizeof(int) * 2;
+
+        /// <summary>
+        /// Serializes a list of Mesh objects to the specified writer.
+        /// Optionally transforms the vertices into the supplied secondarySpace.
+        /// </summary>
+        /// <param name="writer">The writer to which to write the serialized objects.</param>
+        /// <param name="meshes">List of MeshFilter objects to be serialized.</param>
+        /// <param name="secondarySpace">New space to transform the vertices into.</param>
+        public static void Serialize(BinaryWriter writer, IList<MeshFilter> meshes, Transform secondarySpace = null)
```
keveleigh: Do we support / want to support serializing to the HTK's format? Or is this left over and unneeded now that we support additional "real" formats?

davidkline-ms: it's not really needed anymore. i can remove if you think it is the right thing to do

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserverProfile.cs
davidkline-ms: it was todo to pick the right number :\

davidkline-ms: removed

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/RoomFileFormat/RoomFileSerializer.cs
davidkline-ms: Serialize and Write* removed as they are not needed for the asset importer and end up being stale code.

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserverProfile.cs
keveleigh: Should this be added to the `ProfileMenuItemIndices` enum?

davidkline-ms: it feels like we need a followup PR to separate service and data provider menus.

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
```diff
@@ -0,0 +1,336 @@
...
+                // Send the mesh removed event
+                SpatialAwarenessSystem?.RaiseMeshRemoved(this, id);
+            }
+        }
+
+        #endregion IMixedRealitySpatialAwarenessObserver implementation
+
+        #region IMixedRealitySpatialAwarenessMeshObserver implementation
+
+        public SpatialAwarenessMeshDisplayOptions displayOption = SpatialAwarenessMeshDisplayOptions.Visible;
```
jbienzms: The field here should not be public. There is already a public property, which is part of the interface. Field should be marked serializable if necessary and should be private.

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
```diff
@@ -0,0 +1,336 @@
...
+        }
+
+        /// <inheritdoc />
+        public SpatialAwarenessMeshLevelOfDetail LevelOfDetail { get; set; } = SpatialAwarenessMeshLevelOfDetail.Coarse;
+
+        private Dictionary<int, SpatialAwarenessMeshObject> meshes = new Dictionary<int, SpatialAwarenessMeshObject>();
+
+        /// <inheritdoc />
+        public IReadOnlyDictionary<int, SpatialAwarenessMeshObject> Meshes =>
+            new Dictionary<int, SpatialAwarenessMeshObject>(meshes) as IReadOnlyDictionary<int, SpatialAwarenessMeshObject>;
```
jbienzms: This implementation will generate a new dictionary each time the **.Meshes** property is read. Since the contents of this dictionary will be unlikely to change over time, consider caching this instance and returning it for future requests. Of course, if a method like **RemoveMeshObject** is called, the cached dictionary would need to be nullified so that the property accessor would generate it again.

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
```diff
@@ -0,0 +1,336 @@
...
+            get => meshPhysicsLayer;
+
+            set
+            {
+                if ((value < 0) || (value > 31))
+                {
+                    Debug.LogError("Specified MeshPhysicsLayer is out of bounds. Please use a value between 0 and 31, inclusive.");
+                    return;
+                }
+                meshPhysicsLayer = value;
```
jbienzms: If this property is changed at runtime, at what point does an already loaded mesh get moved to the updated physics layer? It appears the code in **Resume** does not run continuously, and that appears to be the only time the layer is used.

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
```diff
@@ -0,0 +1,336 @@
...
+        public int MeshPhysicsLayerMask => (1 << MeshPhysicsLayer);
+
+        /// <inheritdoc />
+        public bool RecalculateNormals { get; set; } = true;
+
+        /// <inheritdoc />
+        public int TrianglesPerCubicMeter { get; set; } = 0;
+
+        /// <inheritdoc />
+        public Material OcclusionMaterial { get; set; } = null;
```
jbienzms: If the mesh is already displaying the occlusion material and this property is changed, how will the mesh get updated to the new material?

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
```diff
@@ -0,0 +1,336 @@
...
+
+        /// <inheritdoc />
+        public Material VisibleMaterial { get; set; } = null;
```
jbienzms: If the mesh is already displaying the visible material and this property is changed, how will the mesh get updated to the new material?

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserverProfileInspector.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        public override void OnInspectorGUI()
+        {
+            RenderProfileHeader(ProfileTitle, ProfileDescription, target, true, BackProfileType.SpatialAwareness);
+
+            using (new GUIEnabledWrapper(!IsProfileLock((BaseMixedRealityProfile)target)))
+            {
+                serializedObject.Update();
+
+                EditorGUILayout.PropertyField(spatialMeshObject);
```
jbienzms: The provider really doesn't do anything if spatialMeshObject doesn't have at least one MeshFilter. Therefore it probably makes sense for the inspector to warn the user if they select a GameObject here that doesn't have any MeshFilters using **GetComponentsInChildren**.

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
```diff
@@ -0,0 +1,336 @@
...
+        /// <param name="profile">The service's configuration profile.</param>
+        public SpatialObjectMeshObserver(
+            IMixedRealityServiceRegistrar registrar,
+            IMixedRealitySpatialAwarenessSystem spatialAwarenessSystem,
+            string name = null,
+            uint priority = DefaultPriority,
+            BaseMixedRealityProfile profile = null) : base(registrar, spatialAwarenessSystem, name, priority, profile)
+        {
+            int i = 0;
+            i++;
```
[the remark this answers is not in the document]
davidkline-ms: this was debugging code... thanks! :D

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
davidkline-ms: good catch... was supposed to be private

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
davidkline-ms: that is by design so that the collection does not change during enumeration. same behavior as the WMR Mesh observer

davidkline-ms: for this pr, this code will not change. we should open a discussion re: safety vs performance and guidance relating to the two

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
davidkline-ms: interesting.. i dont think even the live mesh handles that :)

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
davidkline-ms: you update the display option. we _can_ change the semantic.

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
davidkline-ms: same as the occlusion material question :)

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
davidkline-ms: ex:

observer.OcclusionMaterial = newMaterial;
observer.DisplayOption = Occlusion;

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserverProfileInspector.cs
davidkline-ms: i like that

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
davidkline-ms: added support and confirmed in editor.

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/RoomFileFormat/RoomFileImporter.cs
```diff
@@ -0,0 +1,66 @@
...
+            using (BinaryReader reader = OpenFileForRead(fileInfo.FullName))
+            {
+                meshes = RoomFileSerializer.Deserialize(reader);
+            }
+
+            GameObject model = new GameObject(name);
+            context.AddObjectToAsset(name, model);
+
+            int i = 0;
+            foreach (Mesh mesh in meshes)
```
keveleigh: Since we're keeping track of an `i` here anyway (and incrementing it each loop), does it make sense (is it possible) to just use a `for` loop here?

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserverProfile.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// Configuration profile for the spatial object mesh observer.
+    /// </summary>
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Object Mesh Observer Profile", fileName = "ObjectMeshObserverProfile", order = 1000)]
+    public class SpatialObjectMeshObserverProfile : MixedRealitySpatialAwarenessMeshObserverProfile
+    {
+        [SerializeField]
+        [Tooltip("The model containing the desired mesh data.")]
+        private GameObject spatialMeshObject = null;
+
+        public GameObject SpatialMeshObject => spatialMeshObject;
```
keveleigh: Summary tag?

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/RoomFileFormat/RoomFileImporter.cs
davidkline-ms: good call!

Assets/MixedRealityToolkit.Providers/ObjectMeshObserver/SpatialObjectMeshObserver.cs
```diff
@@ -331,6 +381,20 @@ private void ApplyUpdatedMeshDisplayOption(SpatialAwarenessMeshDisplayOptions op
             }
         }
 
+        /// <summary>
+        ///
```
jbienzms: I know it's private, but just pointing out the missing comment.

davidkline-ms: already fixed.

davidkline-ms:
> I don't see the warning in the inspector about the GameObject missing a MeshFilter. Other than that, these changes look great!

You should. Try creating an empty game object in your scene, Making a prefab and then adding to the profile

davidkline-ms:
```
	                MeshFilter[] filters = (spatialMeshObject.objectReferenceValue as GameObject)?.GetComponentsInChildren<MeshFilter>();
                if ((filters == null) ||
                    (filters.Length == 0))
                {
                    EditorGUILayout.HelpBox("The specified Spatial Mesh Object is not specified or does not appear to be a 3D model.", MessageType.Warning);
                }
```

julenka:
Hi @davidkline-ms, could you please help myself and others understand how to specify the room file to use to test spatial awareness in editor? It would be awesome if you could provide some instructions in the PR, and even better if you could please point us to documentation.

julenka:
Figured it out! You need to add a new SpatialMeshObserver of type SpatialObjectMeshObserver.

spatialmeshobserver example

PR #5133, opened 2019-06-28
Update the build player step to fail the build if the player fails to build.

wiwei:
It turns out right now, if the player build fails, the build can silently succeed. This has happened with the Standalone build, and we haven't caught this because we have depended on the BuildAppX stage in order to catch that the player never succeeded. Of course, with Standalone builds, there is no AppX stage, so failures there just don't get caught.

This change does a couple of things.
1) Makes it so that a failure code returned by the unity build step (i.e. build failure or exception) will actually fail the build, instead of silently succeeding.
2) Changes it so that we will display the unity log in the same step in the case of failure (if we didn't change the "print whatever logs" stages, the failure would have prevented the next steps from going, and we wouldn't see the logs.)

Note that this works with this PR:
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5103

To make sure that we'll fail the build and still get useful logging information.


wiwei:
Working out the kinks, hold off for a bit

wiwei:
Looks like it's good to go - https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=2860

I'm scheduling a new CI build just because I made a cosmetic text change (I noticed that the unity player log section was Begin -> Begin, instead of Begin -> End).

PR #5135, opened 2019-06-29
Changes to enable future stand alone support for the input system

davidkline-ms:
This PR contains changes being needed to enable future use of the input system as a standalone service.

Changes:
- IMixedRealityRaycastProvider implements IMixedRealityService instead of IMixedRealityDataProvider
- DefaultRaycastProvider class inherits from BaseCoreSystem instead of BaseDataProvider
- Raycast provider is now registered as a service instead of as a data provider
- Input system checks for and adds MixedRealityInputModule in Initialize regardless of mode (edit / play)
- Input system sets camera transform to Vector3.zero and Quaternion.identity in Initialize regardless of mode (edit / run)

More likely to come.


luis-valverde-ms:
At a glance everything looks sensible to me

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -145,26 +145,22 @@ public override void Initialize()
                 return;
             }
 
-            if (!Application.isPlaying)
```
lukastoenneMS: I assume this will fix #5061 ?

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -145,26 +145,22 @@ public override void Initialize()
...
-                    // Nothing. Input module is setup correctly.
-                }
-                else
-                {
-                    Debug.LogError($"For Mixed Reality Toolkit input to work properly, please remove your other input module(s) and add a {typeof(MixedRealityInputModule).Name} to your main camera.", inputModules[0]);
-                }
+            if (inputModules.Length == 0)
+            {
+                // There is no input module attached to the camera, add one.
+                CameraCache.Main.gameObject.AddComponent<MixedRealityInputModule>();
```
lukastoenneMS: Note: i'm currently adding an input module manually for Unity UI tests, and to make this work on CI i had to set the [forceModuleActive](https://docs.unity3d.com/ScriptReference/EventSystems.StandaloneInputModule-forceModuleActive.html) flag. Otherwise the module wasn't activating on CI agents, presumably because they don't have any input device attached (it works locally without the flag).

[https://github.com/microsoft/MixedRealityToolkit-Unity/blob/7b334842040124642f440c09f7f037903a64fa34/Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs#L135](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/7b334842040124642f440c09f7f037903a64fa34/Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs#L135)

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
davidkline-ms: It should as this is a run time addition. :)

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
davidkline-ms: I'll be sure to run the CI tests to make sure there nothing conflicts

wiwei: You'll want to change:

Assets\MixedRealityToolkit.Tests\PlayModeTests\PlayModeTestUtilities.cs

To have this function (overwrite the existing EnsureInputModule())

        public static void EnsureInputModule()
        {
            if (CameraCache.Main)
            {
                var inputModule = CameraCache.Main.gameObject.GetComponent<MixedRealityInputModule>();
                if (inputModule == null)
                {
                    inputModule = CameraCache.Main.gameObject.AddComponent<MixedRealityInputModule>();
                }
                inputModule.forceModuleActive = true;
            }
        }




---
In reply to: [299560293](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5135#discussion_r299560293) [](ancestors = 299560293)

wiwei: This should fix the test failures.

The cause here is that the test code utility is no longer setting forceModuleActive (since you're adding the input module here, it's no longer being done in the test code)

Issue #4792, opened 2019-06-10
Documentation


wiwei:
This epic tracks all of the documentation issues scheduled for GA

wiwei:
Closing out GA epics

PR #5137, opened 2019-06-29
[Docs] Adding disclaimer message on development branch features

cre8ivepark:
## Overview
Since the documentation portal pages are generated from mrtk_development branch, it has the pages about the features only available in the development branch. 
Adding a disclaimer block to the feature pages that are not available in the release branch yet.

## Changes
- Fixes: #4792


wiwei:
Just FYI, I've filed https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5155 to track reverting this change once we go up to mrtk_release.

Otherwise, this is gonna get forgotten.

More generally it feels like there should be automated tooling here so we don't have to do this, but that's for another day I guess

PR #5140, opened 2019-06-29
GridObjectCollection Distance field for Surface Type Plane

Alexees:
## Overview
GridObjectCollection has 4 Surface Types, of which 3 are radial and have inspector fields to define the radius from the origin and the radial range.
The Plane type neither makes any use of them in the code nor hides them when selected

This PR adds a Distance field for the PlaneType.
It creates custom inspectors that hide and show fields according to the drop down configurations.
It alters the BaseCollectionInspector and seals the OnInspectorGUI in favor of OnInspectorGUIInsertion, which when overridden will draw additional inspector fields before its own.


Assets/MixedRealityToolkit.SDK/Inspectors/UX/Collections/GridObjectCollectionInspector.cs
```diff
@@ -0,0 +1,66 @@
...
+            cellWidth = serializedObject.FindProperty("cellWidth");
+            cellHeight = serializedObject.FindProperty("cellHeight");
+        }
+
+        protected override void OnInspectorGUIInsertion()
+        {
+            EditorGUILayout.PropertyField(surfaceType);
+            EditorGUILayout.PropertyField(orientType);
+            EditorGUILayout.PropertyField(layout);
+            if (surfaceType.enumValueIndex == 1)
```
Railboy: Nit: For clarity, converting the int value to the actual enum type would be preferred. Same goes for layout type.

Alexees: done

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Collections/GridObjectCollection.cs
```diff
@@ -79,6 +79,20 @@ public float RadialRange
             set { radialRange = value; }
         }
 
+        [SerializeField]
+        [Tooltip("Distance for plane layout")]
```
davidkline-ms: nit: specify units (presuming meters)

Alexees: none of the other fields specify it's units. You sure that people wouldn't just assume its meters, like everything else is? Can't remember anything being something else actually.

wiwei: Default is meters, which I think is fairly common enough that not specifying here is fine.

I think the times where it does become value is time (i.e. is it seconds, milliseconds, etc)

Issue #5144, opened 2019-06-30
ControllerFinder does not need interface for source detection


Alexees:
## Describe the bug

Even though ControllerFinder implements IMixedRealitySourceStateHandler, it does not register to the InputSystem and therefore never receives events for the detection of controllers.

wiwei:
Closing because the linked PR is in

PR #5145, opened 2019-06-30
ControllerFinder now registers its interface with the InputSystem

Alexees:
## Changes
- Fixes: #5144

ControllerFinder is capable of searching controllers on it's own but should also be capable of receiving controller detection events in the case a handedness is requested where there is no controller for yet.
Registering the corresponding events completes the amount of necessary calls.

In other words, the interface implementation was already there, it was just not listening at all


wiwei:
Thanks! Change looks good.

Issue #4530, opened 2019-05-22
The MRTK does not support being unloaded and destroyed


provencher:
## Describe the bug

Currently, if I use SceneManager.LoadScene to change scenes, the MRTK does not get properly disposed, and many errors surface when entering the new scene. This further means that all the memory allocated by the mrtk is not cleaned up when destroyed, which means that it leaks.

If the new scene contains an MRTKInstance, then errors prevent the scene from running properly.
If it does not and I instantiate an instance, the same errors appear.

If I set the MRTK instance to DontDestroyOnLoad, then objects the MRTK Instance created will have been destroyed, and cause errors. Notably created pointers will have been destroyed, and code using them will throw missing references exceptions.

Without using DontDestroyOnLoad, which I understand to be unsupported, I get the following errors:

`ArgumentNullException: Value cannot be null.

Parameter name: _unity_self
Microsoft.MixedReality.Toolkit.Windows.Input.GestureRecognizerExtensions.UpdateAndResetGestures (UnityEngine.XR.WSA.Input.GestureRecognizer recognizer, UnityEngine.XR.WSA.Input.GestureSettings gestureSettings) `

in GestureRecognizerExtensions.cs on line 11
`bool reset = recognizer.IsCapturingGestures();`

This seems to be due to the fact that the gesture recognizers are static instances, and when they are disposed they are never set to null, in order to be cleaned up by the garbage collector.

Next we have 
`NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.Input.BaseGenericInputSource..ctor (System.String name, Microsoft.MixedReality.Toolkit.Input.IMixedRealityPointer[] pointers, Microsoft.MixedReality.Toolkit.Input.InputSourceType sourceType` on line 25

It seems the GazeProvider's Gazepointer instance is null, and is never properly recreated.

This error appears twice consecutively. When I did some debugging, it seems this is related to Speech and Dictation providers not being properly cleaned up.

## To reproduce

1. Open a scene with an MRTK instance
2. Load a new scene with a different MRTK instance (or even a copy of the same one)
2 ALT. Load a new scene and then at some later time instantiate an MRTK instance
3. Watch the errors poor in, which will mean nothing works.

## Expected behavior

It should be possible to destroy the MRTK instance, when the MRTK instance is destroyed by a scene change, it should be fully cleaned up and not leak anything into the new scene.

[...]

PR #5150, opened 2019-07-01
Input recording editor window

lukastoenneMS:
## Overview

Stage 2 of the input recording feature: Editor tools for recording and playing back input animation.

This provides editor-level GUI for the recording and playback services.

## Changes
InputRecordingWindow is a standalone window that can be opened from the Mixed Reality Toolkit menu.
It has two modes:
- _Record_ shows a recording button that toggles the recording service. When recording is stopped, a popup is shown to select a file destination for the recorded data.
image
- _Playback_ allows loading recorded input from a file. The data is then applied back through the input playback service.
image

The Input Recording Window is an alternative to the toggle button in the HandTracking examples. Both of these will work independently by toggling the input recording service.
image


Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/InputRecordingControls.cs
```diff
@@ -68,8 +68,11 @@ public void ToggleRecording()
         /// </summary>
```
wiwei: Nit: Add a remark saying that this will only work once recording has stopped.

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,330 @@
...
+        private InputAnimation animation
+        {
+            get { return PlaybackService?.Animation; }
+            set { if (PlaybackService != null) PlaybackService.Animation = value; }
+        }
+
+        private string loadedFilePath = "";
+
+        private IMixedRealityInputRecordingService recordingService = null;
+        private IMixedRealityInputRecordingService RecordingService => recordingService ?? (recordingService = MixedRealityToolkit.Instance.GetService<IMixedRealityInputRecordingService>());
```
wiwei: Super nit: Some of these lines are super long (consider taming them to be like 100 characters).

Normally I wouldn't have seen this issue but since this PR has over a certain number of commits, Codeflow isn't able to open it up, and github CR stuff has a limit on width.

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,330 @@
...
+            window.titleContent = new GUIContent("Input Recording");
+            window.minSize = new Vector2(380.0f, 680.0f);
+            window.Show();
+        }
+
+        private void OnGUI()
+        {
+            LoadIcons();
+
+            string[] modeStrings = { "Record", "Playback" };
```
wiwei: Super nit: Consider bringing this closer to the definition of the mode enum, just so that these two related and similarly-ordered entities are in the same rough area.

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,330 @@
...
+                    DrawRecordingGUI();
+                    break;
+                case ToolMode.Playback:
+                    DrawPlaybackGUI();
+                    break;
+            }
+
+            EditorGUILayout.Space();
+
+// XXX Reloading the scene is currently not supported,
```
wiwei: @Railboy FYI

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,330 @@
...
+                    float newTimeLimit = EditorGUILayout.FloatField(RecordingService.RecordingBufferTimeLimit);
+                    if (newTimeLimit != RecordingService.RecordingBufferTimeLimit)
+                    {
+                        RecordingService.RecordingBufferTimeLimit = newTimeLimit;
+                    }
+                }
+            }
+
+            bool wasRecording = RecordingService.IsRecording;
+            bool record = GUILayout.Toggle(wasRecording, wasRecording ? new GUIContent(iconRecordActive, "Stop recording input animation") : new GUIContent(iconRecord, "Record new input animation"), "Button");
```
wiwei: Super nit: Another fairly long line - could be helpful to bring the wasRecording ? thing down to the next line at least.

wiwei: bool record = GUILayout.Toggle(
                           wasRecording,
                           wasRecording ?
                               new GUIContent(iconRecordActive, "Stop recording input animation") :
                               new GUIContent(iconRecord, "Record new input animation"),
                           "Button");

wiwei: damnit it doesn't preserve my spaces nooooo

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            using (new GUILayout.VerticalScope(EditorStyles.helpBox))
+            {
+                GUILayout.Label("Animation Info:", EditorStyles.boldLabel);
+
+                if (animation != null)
+                {
+                    GUILayout.Label($"File Path: {loadedFilePath}");
+                    GUILayout.Label($"Duration: {animation.Duration}");
+                }
```
wiwei: Nit: Maybe if there's no animation, maybe worth showing a little label saying "no animation loaded" - otherwise I think you just have this trailing label that says "Animation Info:"

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,330 @@
...
+                    InputAnimationSerializationUtils.Extension,
+                    "Enter filename for exporting input animation");
+            }
+
+            if (outputPath.Length > 0)
+            {
+                string filename = Path.GetFileName(outputPath);
+                string directory = Path.GetDirectoryName(outputPath);
+
+                string result = RecordingService.SaveInputAnimation(filename, directory);
```
wiwei: Seeing as though there was a check added to the recording service (that a save would only occur if recording wasn't happening) - would it be worth adding something similar here or something?

More generally it does seem like an implementation detail/contract of the recording service (save should only be supported while recording has stopped), so maybe this should then just be an overload of the recording service itself

lukastoenneMS: The "isRecording" check was actually added in the InputRecordingControls script for the demo files. The recording service does not care about when the recorded data is serialized.

The reason for checking the recording state is that the SaveRecordedInput function is triggered by the button click - but it's supposed to work as a toggle button (for which we don't have a true concept in MRTK afaik) and so we get a click event both when starting and stopping the recording. Saving on Start does not make any sense (the buffer would be empty), so i added that check to only record after stopping.

lukastoenneMS:
> 
> 
> This is awesome!
> 
> I pulled the change and tried it out - I couldn't quite figure out the skip one frame button - I was kinda hoping I could use it to see the hand frame by frame, but when the play button isn't enabled, you don't see anything

Yeah, there is a bit of a problem here: In order to show and animate the hands the input simulation the `InputSimService.UserInputEnabled` option is disabled by the playback service _during continuous playback_. When scrubbing the timeline or stepping (without "play" button enabled) the user input is enabled and any hand data changes are immediately overridden by the simulation service.

I would like to keep the simulation and playback services entirely separate: Both could create simulated hand controllers, the sim service would use only mouse+keyboard, the playback service uses only animation data. That avoids the problem of having to negotiate "ownership" of the hands between the two services. They will still fight for the camera, but since the simulation service doesn't modify camera transform unless you actually use the mouse/keyboard it won't conflict with animation.

julenka:
Wow, very cool! Question: is there an option to enable a 30 second buffer that you can dump at any time to a file? So for example enable a mode where the last 30 seconds of data is always kept in a rolling buffer, and then a voice command or button or system call we can make that will output that recording to a file on disk somewhere that we can then pull and view.

Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Panels/ToggleFeaturesPanel.prefab
```diff
@@ -398,6 +387,11 @@ PrefabInstance:
   m_Modification:
     m_TransformParent: {fileID: 8503270331930508642}
     m_Modifications:
+    - target: {fileID: 538639403742340272, guid: 9215a7c858170d74fb2257375d5feaf1,
```
julenka: Make sure you've pulled the latest changes to this file. Will made some edits recently you might be squashing over.

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,351 @@
...
+            get { return PlaybackService?.Animation; }
+            set { if (PlaybackService != null) PlaybackService.Animation = value; }
+        }
+
+        private string loadedFilePath = "";
+
+        private IMixedRealityInputRecordingService recordingService = null;
+        private IMixedRealityInputRecordingService RecordingService =>
+            recordingService ??
+            (recordingService = MixedRealityToolkit.Instance.GetService<IMixedRealityInputRecordingService>());
```
julenka: @davidkline-ms is this the preferred way to access a service?

ixedRealityToolkit.Instance.GetService<IMixedRealityInputRecordingService>()

 

If not, what is the guidance for the right way to acquire services? I always look for code to copy/paste but would prefer a page I can refer to with "here are a bunch of exmaples of the preferred way to acquire services". For example, I often point people to https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/CodingGuidelines.html for instructions on the preferred way to have properties with public getters/setters.

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,351 @@
...
+                case ToolMode.Playback:
+                    DrawPlaybackGUI();
+                    break;
+            }
+
+            EditorGUILayout.Space();
+
+// XXX Reloading the scene is currently not supported,
+// due to the life cycle of the MRTK "instance" object.
+// Enable the button below once scene reloading is supported!
```
julenka: Could we please provide a bug reference here?

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -0,0 +1,351 @@
...
+                    string filepath = EditorUtility.OpenFilePanel(
+                        "Select input animation file",
+                        "",
+                        InputAnimationSerializationUtils.Extension);
+
+                    LoadAnimation(filepath);
+                }
+            }
+
+            using (new GUIEnabledWrapper(PlaybackService != null))
```
julenka: @killerantz  FYI -- this is a pretty cool / clean way to show GUI as disabled if condition isn't satisfied (in this case PlaybackService != null) -- might be a nice way to clean up interactable. I didn't know about it until reading this PR, figured I would share.

Documentation/InputSimulation/InputAnimationRecording.md
```diff
@@ -51,7 +51,42 @@ The current animation time can also be controlled directly with the [LocalTime](
 
 # Editor tools for recording and playing input animation
 
->TODO: added in separate PR
+A number of tools exist in the Unity editor for recording and examining input animation. These tools can be accessed in the _Input Recording_ window, which can be opened from the _"Mixed Reality Toolkit > Utilities > Input Recording"_ menu.
```
julenka: Thank you for writing docs!!!!!

Documentation/InputSimulation/InputAnimationRecording.md
```diff
@@ -51,7 +51,42 @@ The current animation time can also be controlled directly with the [LocalTime](
...
+
+<a target="_blank" href="../../Documentation/Images/InputSimulation/MRTK_InputAnimation_EditorToolsMenu.png">
+  <img src="../../Documentation/Images/InputSimulation/MRTK_InputAnimation_EditorToolsMenu.png" title="Editor tools menu" width="80%" class="center" />
+</a>
+
+> Note: Input recording and playback only works during play mode.
```
julenka: Like the input simulation videos which were a huge hit, I have a feeling that making + sharing videos of how to use recording and playback in MRTK will be a big hit. @cre8ivepark  as FYI.

julenka:
These icons look unexpected -- the arrows to the left to me indicate "reset to the beginning" whereas the arrow to the right indicates to me "step forward one frame". But reading the code it looks like you support both jump forward/back and step forward/back. Is the picture just out of date?
image

Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Panels/ToggleFeaturesPanel.prefab
lukastoenneMS: Whenever i get conflicts on asset files i usually avoid trying to merge and just checkout from mrtk_development, then redo my changes in the editor. There may have been some sneaky auto-merge here, i'll check again.

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
lukastoenneMS: Ah yes, i should update these to MixedRealityServiceRegistry as well. Things have changed a lot since i first wrote this code.

Documentation/InputSimulation/InputAnimationRecording.md
lukastoenneMS: Agreed, but we should have a stable feature set before making videos. Otherwise it's just more work if we have to redo it, and confusing when there are outdated versions.

Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Panels/ToggleFeaturesPanel.prefab
lukastoenneMS: This should be alright. The only actual change i did was removing the "SaveRecordedInput" target from the OnRecordingStopped event of the ToggleFeaturesPanel, and add it as a second action to the ToggleInputRecording button. The reason is that i want to avoid saving recorded input twice, via the ToggleFeaturesPanel and then the recording button in the editor tools.

The rest of the changes seem to be some kind of automated Unity fluff that i have no control over. In any case, it's based on latest mrtk_development, so should not affect any of Yoon's changes.

lukastoenneMS:
> 
> 
> These icons look unexpected -- the arrows to the left to me indicate "reset to the beginning" whereas the arrow to the right indicates to me "step forward one frame". But reading the code it looks like you support both jump forward/back and step forward/back. Is the picture just out of date?
> image

I'm not using some of these icons because they don't make sense in the context of animating input.
- Step Backward doesn't work very well with irreversible scene changes.
- Pause and Stop are not needed here, it's either playing the animation or not. There is the Jump Backward button for resetting the animation (although this doesn't include reloading the scene and can also leave irreversible changes).

I'm keeping the icon files around just to have a complete set, in case we want to change the buttons at some point.

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
lukastoenneMS: I have added a ref to #4530 now

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
lukastoenneMS: fixed.

lukastoenneMS:
#5173 fixes the problem of missing hands when stepping or scrubbing through input animation, instead of continuous playback.

Issue #5107, opened 2019-06-28
HandInteractionExample in HL2 scene drops to 30 fps when you bring hands up


julenka:
## Overview
When compiled in Master or Release, the hand interactionexamples scene on HoloLens2 will drop to 30 fps when you bring your hands into view, and stays that way even when hand mesh and joints are turned off.

## Expected Behavior
Framerate should stay at 60.

## Steps to reproduce
1. Build and deploy HandInteractionExample scene to HoloLens 2
2. Keep the profiler on.
2. Play around with the holograms for about 1 minute, especially bounding box
3. Bring hand up and target the lunarmodule with hand rays

You will notice that after being on for a while the framerate will drop to 30 fps.

## Mixed Reality Toolkit Release Version
mrtk_development latest

lukastoenneMS:
@Railboy Can we consider this fixed now?

wiwei:
Closing this one out.

Railboy:
@lukastoenneMS @wiwei There are still performance issues with hand interaction that should be looked into. I will write up a separate issue after some more research. They're not severe enough to kill the frame rate for the hand interaction examples scene, so we can consider this closed.

Issue #5012, opened 2019-06-21
Resolve ambiguity in profile consumption best practices


Railboy:
## Describe the problem

There's no official best practice regarding profile data consumption or modification of profile data at runtime.

This has led to different services consuming profile data in different ways. Some variants:
- Copy all profile data locally and then discard the profile entirely.
- Copy mutable profile data locally, but continue to reference read-only profile data for lifetime of the service.
- Use profile data exclusively and make no local copies.

This has also led to profile inspectors behaving in different ways. Some allow you to modify their contents in play mode and observe the effects of those changes on the active service. Others are completely locked down.

Lots of open questions here. Should services copy and forget profile data? If not, should runtime profile usage be limited to read-only data? If not, is it ever appropriate to modify profile data at runtime to observe changes in service behavior? If so, _when_ is it appropriate (eg in editor vs in build)?

## Describe the solution you'd like

We need to define some profile data consumption best practices, then update our existing services to follow those best practices.

## Additional context

Discussed previously here: #4335

davidkline-ms:
Agreed. Core systems and MRTK team provided extension services should behave the same with regards to data read from profiles.

I believe the most consistent behavior (shared by a majority of existing services) is the following:
- Read the profile into service properties
- Any profile settings that support runtime modification should have setters
- Read-only profile settings should have only getters
- All runtime changes should be made to the service properties, not the profile

Zee2:
Closing as stale; I don't think we'll be making major changes to the profile systems in the future.

PR #5156, opened 2019-07-01
Performance optimization with emphasis on HandInteractionExamples

Railboy:
## Overview

This reduces allocations from ~15k per frame to ~100 bytes per frame, depending on what's going on in the scene.
- Fixes: #5107

## Changes

- Debug.Assert has been replaced with if / Debug.LogError in cases where string formatting was used. String.Format was being invoked even when the asserts passed.
- BoundingBoxHelper no longer allocates new arrays for every operation.
- FocusProvider now uses SphereOverlapNonAlloc. A new FocusQueryBufferSize field has been added to the input profile. This follows the pattern established in PokePointer.
- NearInteractionTouchableUnityUI.Instances no longer returns a read-only collection. We simply iterate over the list.
- Removed Linq from BaseEventSystem.
- ~~We now omit lightmapCoords in MixedRealityRaycastHit by default via a preprocessor directive. Including them requires a null check against a Unity object at a minumum. Devs can enable them by defining MRTK_ENABLE_LIGHTMAPCOORD_IN_RAYCASTHIT.~~ Reverted in favor of @wiwei's fix.
- HandInteractionPanZoom now uses SetUvs and passes results from GetUvs to other methods.
- In ManipulationHandler moved Linq extension methods to a wrapper method to be dealt with later.
- Changed some materials to use instancing.

## Issues I didn't address
- UnityEvents are toxic. Eg, ManipulationHandler allocates about 1.5k when invoking any ManipulationEvent, 1.1k of which is UnityEvent overhead. ManipulationEvents are discrete so they're not a huge problem. But we should avoid using these for update events.
- We should further optimize handler-related code in BaseEventSystem. Looking up event handler by type ends up being expensive in aggregate. Maybe a custom equality comparer?

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
@@ -200,7 +200,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
         private TwoHandScaleLogic scaleLogic;
         private TwoHandRotateLogic rotateLogic;
         private Dictionary<uint, IMixedRealityPointer> pointerIdToPointerMap = new Dictionary<uint, IMixedRealityPointer>();
-
         private Quaternion objectToHandRotation;
         private Vector3 objectToHandTranslation;
         private bool isNearManipulation;
@@ -531,7 +530,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
         private void HandleOneHandMoveUpdated()
         {
             Debug.Assert(pointerIdToPointerMap.Count == 1);
-            IMixedRealityPointer pointer = pointerIdToPointerMap.Values.First();
+            IMixedRealityPointer pointer = GetFirstPointer();
 
             Quaternion targetRotation = Quaternion.identity;
             RotateInOneHandType rotateInOneHandType = isNearManipulation ? oneHandRotationModeNear : oneHandRotationModeFar;
@@ -622,7 +621,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
         private void HandleOneHandMoveStarted()
         {
             Assert.IsTrue(pointerIdToPointerMap.Count == 1);
-            IMixedRealityPointer pointer = pointerIdToPointerMap.Values.First();
+            IMixedRealityPointer pointer = GetFirstPointer();
 
             moveLogic.Setup(GetPointersCentroid(), hostTransform.position);
 
@@ -754,6 +753,13 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
                 rigidBody = null;
             }
+        }
+
+        private IMixedRealityPointer GetFirstPointer()
+        {
+            // We may be able to do this without allocating memory.
+            // Moving to a method for later investigation.
+            return pointerIdToPointerMap.Values.First();
         }
 
         #endregion
--- a/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
@@ -21,7 +21,9 @@ namespace Microsoft.MixedReality.Toolkit.Input
         public FocusProvider(
             IMixedRealityServiceRegistrar registrar,
             MixedRealityInputSystemProfile profile) : base(registrar, profile)
-        { }
+        {
+            maxQuerySceneResults = profile.FocusQueryBufferSize;
+        }
 
         private readonly HashSet<PointerData> pointers = new HashSet<PointerData>();
         private readonly HashSet<GameObject> pendingOverallFocusEnterSet = new HashSet<GameObject>();
@@ -30,6 +32,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         private readonly Dictionary<uint, IMixedRealityPointerMediator> pointerMediators = new Dictionary<uint, IMixedRealityPointerMediator>();
         private PointerHitResult hitResult3d = new PointerHitResult();
         private PointerHitResult hitResultUi = new PointerHitResult();
+        private int maxQuerySceneResults = 64;
 
         public IReadOnlyDictionary<uint, IMixedRealityPointerMediator> PointerMediators
         {
@@ -866,7 +869,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
                     // Perform raycast to determine focused object
                     var raycastProvider = InputSystem.RaycastProvider;
                     hitResult3d.Clear();
-                    QueryScene(pointer.Pointer, raycastProvider, prioritizedLayerMasks, hitResult3d);
+                    QueryScene(pointer.Pointer, raycastProvider, prioritizedLayerMasks, hitResult3d, maxQuerySceneResults);
                     PointerHitResult hit = hitResult3d;
 
                     // If we have a unity event system, perform graphics raycasts as well to support Unity UI interactions
@@ -973,12 +976,15 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
         #region Physics Raycasting
 
+        // Colliders used to store sphere overlap results
+        private static Collider[] colliders = null;
+
         /// <summary>
         /// Perform a scene query to determine which scene objects with a collider is currently being gazed at, if any.
         /// </summary>
         /// <param name="pointerData"></param>
         /// <param name="prioritizedLayerMasks"></param>
-        private static void QueryScene(IMixedRealityPointer pointer, IMixedRealityRaycastProvider raycastProvider, LayerMask[] prioritizedLayerMasks, PointerHitResult hit)
+        private static void QueryScene(IMixedRealityPointer pointer, IMixedRealityRaycastProvider raycastProvider, LayerMask[] prioritizedLayerMasks, PointerHitResult hit, int maxQuerySceneResults)
         {
             float rayStartDistance = 0;
             MixedRealityRaycastHit hitInfo;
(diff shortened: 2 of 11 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
```diff
@@ -17,7 +15,7 @@ public class NearInteractionTouchableUnityUI : BaseNearInteractionTouchable
     {
         private RectTransform rectTransform;
 
-        public static IReadOnlyCollection<NearInteractionTouchableUnityUI> Instances { get { return instances.AsReadOnly(); } }
+        public static List<NearInteractionTouchableUnityUI> Instances => instances;
```
keveleigh: I don't think this retains the original intent, which was to prevent the collection from being added to / removed from.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
```diff
@@ -390,7 +390,7 @@ public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.Event
                 return;
             }
 
-            Debug.Assert(baseInputEventData.InputSource.Pointers != null, $"InputSource {baseInputEventData.InputSource.SourceName} doesn't have any registered pointers! Input Sources without pointers should use the GazeProvider's pointer as a default fallback.");
+            if (baseInputEventData.InputSource.Pointers == null) { Debug.LogError($"InputSource {baseInputEventData.InputSource.SourceName} doesn't have any registered pointers! Input Sources without pointers should use the GazeProvider's pointer as a default fallback."); }
```
keveleigh: Debug.Assert gets compiled out in many cases. With this change, now we have an empty `if null` check in those cases, instead of just nothing. That feels weird, especially for something that won't exist when the app is actually built to be released.

wiwei: Yep - if you're profiling on anything other than the master build (on device) you will see allocations caused by string manipulations.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -756,6 +755,13 @@ private void ReleaseRigidBody()
             }
         }
 
+        private IMixedRealityPointer GetFirstPointer()
+        {
+            // We may be able to do this without allocating memory.
+            // Moving to a method for later investigation.
+            return pointerIdToPointerMap.Values.First();
```
wiwei: One minor improvement on this would be to cache this value anytime the pointer map is updated - so on pointer down/up, it will call .First() just once (which would then be reused for the duration of the move "updated" path).

More generally this would also lead to actual determinism in the algorithm as well, because right now it's only working thanks to luck - dictionary value enumeration is never advertised as being deterministic.

Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
Railboy: The impact in the editor is really bad, though. We can't count on devs profiling on device to iron out performance, so they'll see the core input system generating tons of garbage and not necessarily understand that it won't follow them onto the device.

An alternative is to pass an unformatted message to the assert.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -30,6 +32,7 @@ public class FocusProvider : BaseCoreSystem, IMixedRealityFocusProvider
         private readonly Dictionary<uint, IMixedRealityPointerMediator> pointerMediators = new Dictionary<uint, IMixedRealityPointerMediator>();
         private PointerHitResult hitResult3d = new PointerHitResult();
         private PointerHitResult hitResultUi = new PointerHitResult();
+        private int maxQuerySceneResults = 64;
```
wiwei: @julenka, do you have any data on some of the more complex scenes that people were building? 64 seems like it could be a small number potentially depending on how many near interaction things that exist.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
Railboy: Is there a way to deliver a read-only collection without garbage / that you can iterate over without an enumerator? I feel like there must be.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
wiwei: Ah, I guess it's part of the profile anyway, so it's changeable at least there.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -1019,17 +1025,33 @@ private static void QueryScene(IMixedRealityPointer pointer, IMixedRealityRaycas
...
-                        Collider[] colliders = UnityEngine.Physics.OverlapSphere(pointer.Rays[i].Origin, pointer.SphereCastRadius, ~UnityEngine.Physics.IgnoreRaycastLayer);
+                        // Set up our results array
+                        if (colliders == null)
+                        {
+                            colliders = new Collider[maxQuerySceneResults];
+                        }
+                        else if (colliders.Length != maxQuerySceneResults)
+                        {
+                            Array.Resize<Collider>(ref colliders, maxQuerySceneResults);
+                        }
```
wiwei: Is the plan here to make it so that maxQuerySceneResults is changeable at runtime as well? Otherwise this code will never be invoked/useful.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
Railboy: I should probably set this to 0 to make it clear that this value gets blown away by the profile value. I have it set to 128 in the profile by default.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
wiwei: I think that List itself also just implements IReadOnlyCollection:

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1?view=netframework-4.8

So you might be able to just return instances itself (and leave the IReadOnlyCollection part alone)

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputSystemProfile.cs
```diff
@@ -56,6 +56,16 @@ public SystemType RaycastProviderType
...
 
+        /// <summary>
+        /// Maximum number of colliders that can be detected in a SphereOverlap scene query.
+        /// </summary>
+        public int FocusQueryBufferSize => focusQueryBufferSize;
+
+        [SerializeField]
+        [Range(32, 2048)]
+        [Tooltip("Maximum number of colliders that can be detected in a SphereOverlap scene query.")]
+        private int focusQueryBufferSize = 128;
```
wiwei: Nit: Change the value in the FocusProvider.cs thing to 128 as well (or change this to 64) to be consistent. It doesn't make sense for us to have two different defaults.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputSystemProfile.cs
```diff
@@ -56,6 +56,16 @@ public SystemType RaycastProviderType
             internal set { raycastProviderType = value; }
         }
 
+        /// <summary>
+        /// Maximum number of colliders that can be detected in a SphereOverlap scene query.
+        /// </summary>
+        public int FocusQueryBufferSize => focusQueryBufferSize;
+
+        [SerializeField]
+        [Range(32, 2048)]
```
wiwei: Shouldn't this just be 1? i.e. why can't I try to detect just a single object?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
Railboy: My guess is it will never change but that's not totally settled: #5012 
So I figured better safe than sorry.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityRaycastHit.cs
```diff
@@ -1,6 +1,10 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+// Including lightmap coord has a performance cost so we've disabled it by default.
+// Uncomment the define to enable lightmap coords.
+// #define MRTK_ENABLE_LIGHTMAPCOORD_IN_RAYCASTHIT
```
wiwei: I know that I let the other change languish - I'm going to get that fixed today. Can you leave this file untouched so that my change can just do its part?

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityInputSystemProfile.cs
Railboy: Good point.

Assets/MixedRealityToolkit/Definitions/InputSystem/MixedRealityRaycastHit.cs
Railboy: Does your change avoid a Unity.Object reference null check?

wiwei: I'm going to change it to be a struct, so no more null checks.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchableUnityUI.cs
Railboy: Tried changing Instances to return an IReadOnlyList and it works great.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
Railboy: I think that would be ideal. I'm going to leave that for @julenka to work out since she's most familiar with the state machine.

julenka: Easiest may literally be to just have two fields: 

```
Pointer firstPointer;
Pointer secondPointer;

OnPointerDown(Pointer newPointer)
{
secondPointer = newPointer
}

OnPointerUp/OnSourceLost(Pointer lostPointer)
{
  if (lostPointer == secondPointer)
  {
     secondPointer = null;
  }
  if (lostPointer == firstPointer)
  {
    firstPointer = secondPointer;
    secondPointer = null;
  }
}
```

This of course assumed you never have more than two pointers, but I think this is fair and we can debug assert that.

Issue #5142, opened 2019-06-30
Orbital not capable of what Billboard can do (which is not a solver)


Alexees:
## Describe the bug

Orbital  does not have any option to follow one object while looking at another on the y-axis only. The Orbital script has all sorts SolverOrientationType modes instead, which completely define the rotation.
Orbital does not have a TransformTarget field even though it can for example look at the camera, with is an option of SolverOrientationType  and is hard coded.

## Expected behavior

Orbital should use the same PivotAxis enum like Billboard does.
In addition, it should be able to look at custom targets, like Billboard does
Basically, it should integrate the capability of being billboard, which it nearly is.

Or maybe creating a Billboard Solver and just ditch Orbital

Also, there are quite a lot of enums and a couple of them are pretty much either a portion of another or a combination of them:

OrientationType
PivotAxis
RadialViewReferenceDirection
RotationConstraintType
SolverOrientationType

And why doesn't the Pivot Axis enum make use of the FlagsAttribute?

julenka:
Closing in favor of larger refactor #5713

Issue #4095, opened 2019-04-26
Orbital needs a little redesign


Alexees:
Orbital provides two offsets in the inspector (LocalOffset and WorldOffset). The tooltip prohibits the use of both at the same time.
This should be reduced to just an offset and an enum distinguishing between local and world mode for that. Nothing really prevents you from screwing this up.

The difference between these two is just that worldOffset uses yawOnly. They have nothing to to with wether they follow a "LocalPosition" or a "WorldPosition" (static position) as I would have expected.

Futhermore, there is the orientationType yawOnly, which explicitly uses the TransformTarget. it cannot be combined with all the other orientations which could use this "keep it up right" option.

yaw Only should be a boolean used with all other options.

Yoyozilla:
This is a feature request. updating tag to reflect.

wiwei:
@cre8ivepark what do you think of this request?

julenka:
Closing in favor of #5713

PR #5159, opened 2019-07-01
Orbital2

Alexees:
## Overview
Based on my findings in #5142, I started the work on another Orbital so I could have a Menu that would circle a target, faces the user while showing it's contents. In doing so, I tried to create Orbital as flexible as possible so it could be used in all sorts of szenarios.

I did not like the fact that there were enums that declared concrete configurations like "look at camera only on the y axis. What if I wanted to have the same constraint but look at something else?

That's why I came up with this.

The current feature set is the following:

- It can follow the same Transform the SolverHandler follows, follow all the TrackedObjectTypes or any other gameobjct
- It uses the local Position and Rotation offset of the SolverHandler
- It can look at the same targets defined in SolverHandler
- The look rotation can be constraint on any axis (like billboard could)
- It can snap angles on any constraint axis

 - Special cases like solver and orbital use the same target handle this behaviour correctly
- The inspector reflects the decisions that can be made by hiding and showing fields

Example usage (described rotating menu above):
- One GameObject (billboardParent) would target the GameObject it wants to circle
- It looks at the head and constraints its look rotation on the y axis
- a second GameObject targets and looks at billboardParent
- it does not use smoothing, has a positional offset and a 180° rotation offset

The whole construct can be offset from the object itself and will always show its content to the user while maintaining an upright position.

Feel free to check it out.
In my opinion, SolverHandler could benefit from the choice of local or world offset too.


Alexees:
I'd like to reference #4095 as my starting point for doing this

Alexees:
Hey, thanks for looking into this, nearly forgot about this myself as it has been quite some time I created this (at least in mrtk_development terms 😆 )

While I was doing Orbital2 (which was intended to replace Orbital 1), it already felt bloated. Looking at some of the other solvers, their inspectors are quite large too.
When I saw #5393 I thought there must be a better way to stack those capabilities. Especially constraining a Vector3 should not be something every solver should be capable of but should be some kind of modifier that does this prior the the solver operating with it.
I don't know what a system like this could look like, it's just a guts feeling that there is a more elegant solution to all the solvers.

Then, a lot of values that are used by the solvers are accessed via SolverHandler property from within their SolverUpdateEntry. This is probably fine for some, but especially deltaTime and TransformTarget are used everywhere and could be passed along via parameters of SolverUpdateEntry.

And finally, teleporting a configured construction is not that easy. while the solvers support a SnapTo method, there is no such thing in the SolverHandler. I needed to do this to get my setup teleported:

```
        transform.position = targetPosition;
        transform.rotation = lookRotation;

        solverHandler.GoalPosition = targetPosition;
        solverHandler.GoalRotation = lookRotation;

        orbital2.SnapTo(targetPosition, lookRotation, transform.localScale);
```

I wouldn't say that Orbital2 is a good solution on its own, especially the workaround with ControllerFinderOrbital2.
I just think configuring a GameObject to do what you want to it do should somehow follow the way you would break it up in your head, like I did in my Example, should be possible.
Having things like "Follow Camera but on y-Axis only" is a IMHO out of the question bad idea.

SimonDarksideJ:
I see these changes as extensions to the existing orbital solver, rather than an entirely new solver.  Better these changes are incorporated in to the existing as options.

Assets/MixedRealityToolkit/Definitions/Utilities/TransformationSpaceType.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        /// <summary>
+        /// Use the define position, rotation or scale in local space
+        /// </summary>
+        LocalSpace = 0,
+
+        /// <summary>
+        /// Use the define position, rotation or scale in world space
+        /// </summary>
+        WorldSpace
```
julenka: Hi @Alexees , I'm curious....did you ever have a case when you wanted the offset to be in world space? It seems by default the offset is in local space alway.

julenka:
Hi @alexees, I just tried out orbital2 and it was really nice! I found it much easier to understand than orbital. Thanks for the great improvement! I'm currently fleshing out a ticket that combines this with #5611, so we can have a consolidated "follow" solver.

Assets/MixedRealityToolkit/Definitions/Utilities/TransformationSpaceType.cs
Alexees: I did not think about use cases really :)
I was just a little frustrated about Orbital that I thought, make it as flexible as it can be. If you can follow something, with offset, and at the same time look at something, with offset. Maybe someone might find this useful. 
> It seems by default the offset is in local space always

I don't understand, some value has to be the default. It certainly doesn't mean worldspace isn't used.

Issue #4447, opened 2019-05-17
Tooltips rendering grey plane in scene


davidraguilar:
## Describe the bug

When using any of the ToolTip prefabs, a grey plane is displayed in the scene rendering the tooltips useless. It works fine when playing the scene in the editor, but broken when testing on device. This happens when using tooltips in a custom scene as well as when testing the TooltipExamples demo scene in the MRTK examples.

## To reproduce

1. Load TooltipExamples demo scene found here: MixedRealityToolkit.Examples/Demos/UX/Tooltips/Scenes/TooltipExamples.unity
2. Build
3. Deploy to device and start app
4. See grey plane

## Expected behavior

Regular tooltip functionality without the grey plane

## Screenshots

![20190517_004015_HoloLens](https://user-images.githubusercontent.com/1686969/57912300-d5b19000-783e-11e9-9e1c-fb203b7cde15.jpg)

![20190517_003950_HoloLens](https://user-images.githubusercontent.com/1686969/57912328-e6fa9c80-783e-11e9-841c-650356aa8e52.jpg)

![20190517_003929_HoloLens](https://user-images.githubusercontent.com/1686969/57912346-ea8e2380-783e-11e9-9f00-051ae916d23a.jpg)

## Your Setup (please complete the following information)

- Unity Version 2018.4.0f1, 2018.3.14f1
- MRTK Version v2.0.0-RC1-Refresh

## Target Platform (please complete the following information)

- HoloLens

Railboy:
@wiwei I've been discussing this with @cre8ivepark and we should probably elevate to a GA blocker.

cre8ivepark:
@wiwei Added GA block label since the functionality is broken and not usable.

Issue #5126, opened 2019-06-28
Tooltip background responsiveness


derekfreed:
## Describe the bug

If a tooltip is dynamically instantiated and text is assigned, the background does not react and expand accordingly.

## To reproduce

1. Instantiate a tooltip prefab
2. In the same frame set the tooltiptext to a string value
3. Observe bug that background does not expand to fit the new text

## Expected behavior

I expect that setting the tooltiptext value will cause the background panel to expand and fit the text

## Screenshots
![image](https://user-images.githubusercontent.com/10147303/60358009-fd415f80-9989-11e9-9c14-7f4b61d16dc5.png)

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.13f1]
- MRTK V2

## Target Platform (please complete the following information)

- HoloLens

derekfreed:
Currently I can mitigate this by waiting a couple frames to set the text

PR #5161, opened 2019-07-01
Fixes ToolTip background plane size issues

Railboy:
## Overview
Fixes ToolTip background plane size issues.

## Changes
- Fixes: #4447, #5126

## Verification
Leaving this as a draft until it's been tested on multiple devices.

>
>

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Tooltips/ToolTip.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Tooltips/ToolTip.cs
@@ -176,7 +176,10 @@ namespace Microsoft.MixedReality.Toolkit.UI
             set
             {
                 contentScale = value;
-                RefreshLocalContent();
+                if (!Application.isPlaying)
+                {   // Only force refresh in edit mode
+                    RefreshLocalContent();
+                }
             }
         }
 
@@ -194,7 +197,10 @@ namespace Microsoft.MixedReality.Toolkit.UI
             set
             {
                 fontSize = value;
-                RefreshLocalContent();
+                if (!Application.isPlaying)
+                {   // Only force refresh in edit mode
+                    RefreshLocalContent();
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


cre8ivepark:
Thanks Lars! Verified in HoloLens 2. All tooltip text and backplates are properly displayed.
MRTK_Tooltip3

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Tooltips/ToolTip.cs
```diff
@@ -140,10 +140,10 @@ public string ToolTipText
         {
             set
             {
-                if (value != toolTipText)
```
wiwei: >if (value != toolTipText) [](start = 16, length = 25)

Given that this was removed - was it intended that setting the text to the same thing should also trigger a local content refresh (even if it's being set to the same thing?)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Tooltips/ToolTip.cs
```diff
@@ -140,10 +140,10 @@ public string ToolTipText
...
-                {
-                    toolTipText = value;
-                    RefreshLocalContent();                   
+                toolTipText = value;
+                if (!Application.isPlaying)
+                {   // Only force refresh in edit mode
```
wiwei: I also see that only ToolTipText - is there a specific reason why only this property has this guard (but things like ContentScale, FontSize, and others don't have it?)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Tooltips/ToolTip.cs
Railboy: In editor mode yes, because you can't count on Update to be called consistently.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Tooltips/ToolTip.cs
Railboy: Now that you point it out I guess I don't see the harm in changing them too.

PR #5162, opened 2019-07-01
Fix boundingbox doesn't work if maniphandler appears above it

julenka:
## Overview
Third attempt to fix issue #4546 which is caused by #4393. This bug fixes the root issue #4393.

The root cause of #4393 is that when maniphandler is above boundingbox, maniphandler will swallow all pointer events originating from scale, rotate affordances before bounding box has a chance to handle it. This causes the scale, rotate affordances to be impossible to grab. The fix is to put bounding box before manip handler. This change instead will make boundingbox work regardless of if it is before or after maniphandler.

The fix (suggested by @luis-valverde-ms, thanks!) is to put all the pointer event handlers on rigRoot instead of on BoundingBox. This means that all pointer events coming from scale, rotate handles will get handled by rigroot, before manip handler has a chance to swallow the events. A simple solution that doesn't require any changes to the toolkit.

## Changes
- Fixes: #4546 #4393
- Fix bug where PointerHandler was not configurable in code
- BoundingBox no longer needs to implement IMixedRealityPointerHandler, so remove that.

## Verification
- Verified RuntimeBoundingBoxExample scene
- Verified HandInteractionExample scene works
- Verified #4393 is fixed in my own test scene
- Verified playmode tests pass


Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/PointerHandler.cs
```diff
@@ -17,22 +17,22 @@ public class PointerHandler : BaseInputHandler, IMixedRealityPointerHandler
         /// <summary>
         /// Unity event raised on pointer down.
         /// </summary>
-        public PointerUnityEvent OnPointerDown;
+        public PointerUnityEvent OnPointerDown = new PointerUnityEvent();
```
luis-valverde-ms: Why do we need this?

luis-valverde-ms: Hmm, did you perhaps hit an issue when subscribing to the event from code?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1457,12 +1458,21 @@ private void SetMaterials()
...
+        private void InitializeRigRoot()
         {
-            rigRoot = new GameObject(rigRootName).transform;
+            var rigRootObj = new GameObject(rigRootName);
+            rigRoot = rigRootObj.transform;
             rigRoot.parent = transform;
 
+            var pH = rigRootObj.AddComponent<PointerHandler>();
+            pH.OnPointerDown.AddListener(OnPointerDown);
+            pH.OnPointerDragged.AddListener(OnPointerDragged);
```
julenka: @Railboy I believe adding a listener to onpointerdragged, which gets called every frame, may increase memory usage. This is based on reading this comment in #5156 

> UnityEvents are toxic. Eg, ManipulationHandler allocates about 1.5k when invoking any ManipulationEvent, 1.1k of which is UnityEvent overhead. ManipulationEvents are discrete so they're not a huge problem. But we should avoid using these for update events.

I am thinking to create a follow up PR to remove this line, since we don't even do anything in OnPointerDragged (my mistake that I added it). Are there any other changes that you see in this PR that could cause lots of memory usage? I was only worried about onpointerdragged because it gets called every frame.

luis-valverde-ms: _UnityEvents are toxic_ is a very strong statement that moves us away from the most Unity-like way of handling events. I'd say we need more hard data before accepting that as a general guideline. I'd like to see what's the allocation footprint of a invoking a UnityEvent in a Master build, that's what I'd use as a reference.

Railboy: @luis-valverde-ms Yes that was a bit hyperbolic, I was just stunned to see those events allocating more than 1k per invocation.

>I'd like to see what's the allocation footprint of a invoking a UnityEvent in a Master build, that's what I'd use as a reference.

Good idea, I'll give that a shot and we'll see what happens.

luis-valverde-ms: Thank you! Let me know how it goes. I'd be suprised if UnityEvent is doing something very unreasonable perf/allocation wise in Master builds.

PR #5163, opened 2019-07-01
Update the build pipeline to first build standalone, in order to avoid infinitely hanging play mode tests.

wiwei:
It turns out that if there's a build error, running play mode tests will hang indefinitely in batch mode (this is because it shows a dialog saying "hey, make sure your stuff builds"). Previously this wasn't an issue because the test step happened after everything else, but now that we moved the test step up to be the first thing, this is now problematic.

Building standalone is the fastest of all of the flavors (because it doesn't have to build an AppX), so we're moving this to the top.

### DIFF
--- a/pipelines/templates/common.yml
+++ b/pipelines/templates/common.yml
@@ -1,8 +1,17 @@
 # [Template] Common build tasks shared between CI builds and PR validation.
 
 steps:
-# Run tests first so, that engineers can get test failure notifications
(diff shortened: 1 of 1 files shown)


wiwei:
@andreiborodin FYI, in case the asset retargetting step MUST immediately follow the build standalone step. If this is the case I'll need to reconsider and move something else around instead.

Issue #4978, opened 2019-06-19
[MRTK Shell Parity] [Button] Front cage should be compressed


cre8ivepark:
## Describe the bug
Button's front cage should be compressed. Currently, it moves back and penetrates through the backplate.
Part of #4200 UX Controls Parity with HoloLens 2 Shell

## To reproduce
In the HandInteractionExamples scene, press button with a finger.

## Expected behavior
Front cage is compressed on the press interaction.

![2019-06-19 14_51_53-Unity 2018 4 0f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof](https://user-images.githubusercontent.com/13754172/59803921-cf7d5c00-92a1-11e9-89fd-86f7aa3b666c.png)

## Current behavior
![2019-06-19 14_49_50-Unity 2018 4 0f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof](https://user-images.githubusercontent.com/13754172/59803827-86c5a300-92a1-11e9-98ca-e70daa7d5ecf.png)

## Branch
mrtk_development

ADP-JoNeff:
#Close
Dupe of #4562

cre8ivepark:
Completed with #5168

Issue #4564, opened 2019-05-24
Button 'splash visual' needs to trigger not on touch but on button press


julenka:
## Overview
The splash visual currently triggers when the finger touches the button, but the button activates on press. The splash visual should trigger when the button actually gets pressed.

cre8ivepark:
Completed with #5168

Issue #4976, opened 2019-06-19
[MRTK Shell Parity] [Bounding Box] In Flattened mode, the border visualization is too thin and almost invisible


cre8ivepark:
## Describe the bug
In Flattened mode, the border visualization is too thin and almost invisible
Part of #4200 UX Controls Parity with HoloLens 2 Shell

## To reproduce
In HandInteractionExamples scene, hover flattened examples (slate, MRTK plane) with hand ray or finger to make the border visible.

## Expected behavior
Shows solid border similar to the Bounding Box in the 3D objects

## Screenshots
![2019-06-19 14_40_52-Unity 2018 4 0f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof](https://user-images.githubusercontent.com/13754172/59803527-9395c700-92a0-11e9-8233-601169f6b8d5.png)
![2019-06-19 14_41_19-Unity 2018 4 0f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof](https://user-images.githubusercontent.com/13754172/59803528-9395c700-92a0-11e9-9a37-8dd795509d16.png)
![2019-06-19 09_34_42-Unity 2018 4 0f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof](https://user-images.githubusercontent.com/13754172/59803529-9395c700-92a0-11e9-8fa3-bed78bd23597.png)

## Branch
mrtk_development

julenka:
Fixed

PR #5168, opened 2019-07-01
[Shell Parity] [Button] Compressible Button

Cameron-Micka:
## Overview
The HoloLens 2 pressable button's front plate now compresses like it does in the shell and the front plate splash visual now triggers when the button is pressed.

Compress

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4978 https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4564

## Verification
Please open a scene with the PressableButtonHoloLens2 prefab in it and try pressing a few buttons. The HandInteractionExamples scene is a good place to try out the buttons.

>
>
>


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -264,18 +267,21 @@ private void Update()
...
-            foreach (var pointer in eventData.InputSource.Pointers)
+            if (currentInputSources.Count != 0)
             {
-                ProximityLight[] proximityLights = pointer.BaseCursor?.GameObjectReference?.GetComponentsInChildren<ProximityLight>();
-
-                if (proximityLights != null)
+                foreach (var pointer in currentInputSources[currentInputSources.Count - 1].Pointers)
                 {
-                    foreach (var proximityLight in proximityLights)
+                    ProximityLight[] proximityLights = pointer.BaseCursor?.GameObjectReference?.GetComponentsInChildren<ProximityLight>();
```
julenka: For a future PR: I wish that we could decouple movement / algorithms from visuals like light pulsing. Would have preferrred to have a LightPulse component that listens to when the button is pressed to play a pulse. That way you could customize / change when the pulse plays without needing to modify pressablebutton.cs.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -312,9 +318,12 @@ void IMixedRealityTouchHandler.OnTouchStarted(HandTrackingInputEventData eventDa
             }
 
             touchPoints.Add(eventData.Controller, eventData.InputData);
-            IsTouching = true;
 
-            PulseProximityLight(eventData);
+            // Make sure only one instance of this input source exists and is at the "top of the stack."
```
julenka: Why does it need to be at the end of the list? Why not use a hashset?

Assets/MixedRealityToolkit/Utilities/MaintainBorderLightWidth.cs
```diff
@@ -0,0 +1,71 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities
+{
+    /// <summary>
+    /// Utility component to keep the border light width a constant size no mater the object scale.
```
julenka: Does this require a specific shader is used? Can you add that to the comment?

julenka: Could you please explain why we need this class? I also wonder if it's appropriate to stick in utilities. If it's specific to the button, should it go there?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
Cameron-Micka: I wish we could too! But, sadly the ButtonPressed event doesn't pass any state along to know which input source fired the event.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
Cameron-Micka: Using a set was my first reaction (since I wanted to avoid duplicates), but I also needed something sorted by "age." I thought about SortedSet too, but I didn't think the overhead warranted something this simple.

Assets/MixedRealityToolkit/Utilities/MaintainBorderLightWidth.cs
Cameron-Micka: Good idea! It does require the MRTK/Standard shader, or one which follows similar naming conventions. I'll add that comment.

Cameron-Micka: Right now the "_BORDER_LIGHT" feature of the MRTK/Standard shader maintains the border width based on the object's scale. For example if you have a border width of 10 pixels on an object of scale (1,1,1) then you scale that same object to (2,2,2) your border will now be 20 pixels wide. Sometimes developers want to maintain the border width no matter the object scale (in the case of the button front plate as it get's compressed, and a few other instances I've encounterted). Right now the only thing using this behavior is the button, but I wouldn't say this behavior is unique to the button. I'm happy to move it elsewhere if there is a more appropriate location?

cre8ivepark: Could we used this for the Bounding Box (flattened) too?
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4976

Cameron-Micka: I bet we could! We should give it a try.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButtonHoloLens2.cs
```diff
@@ -15,6 +15,30 @@ public class PressableButtonHoloLens2 : PressableButton
...
+                {
+                    initialCompressableButtonVisualsLocalScale = compressableButtonVisuals.transform.localScale;
+                }
+
+            }
+        }
+
+        [SerializeField]
+        [Range(0.0f, 1.0f)]
+        [Tooltip("The minimum percentage of the original scale the compressableButtonVisuals can be compressed to.")]
```
thalbern: MinCompressPercentage  is a public field: please add this tooltip as a /// comment as well so it shows up in the docs

Assets/MixedRealityToolkit/Utilities/MaintainBorderLightWidth.cs
```diff
@@ -0,0 +1,73 @@
...
+                // Multiply the initial border width by the amount need to maintain it's value at the new scale.
+                var scalePercentage = minScale / initialScale[minAxis];
+
+                if (scalePercentage.Equals(0.0f))
+                {
+                    scalePercentage = 1.0f;
+                }
+
+                targetRenderer.GetPropertyBlock(properties);
+                properties.SetFloat(borderWidthID, initialBorderWidth / scalePercentage);
```
thalbern: I think I'm missing some information here:
So we're multiplying that scalePercentage that we got from the axis with the smallest scale value with the border but the border could be potentially in all three dimensions? x, y and z axis? So if only the z axis got scaled down won't that blow up borders going along x and y axis?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButtonHoloLens2.cs
Cameron-Micka: Good catch, I'll fix that!

Assets/MixedRealityToolkit/Utilities/MaintainBorderLightWidth.cs
Cameron-Micka: I'm glad you brought this up since the logic does look misleading. In the shader (where the border width value is used) it also looks for the smallest axes scale and then uses that to calculate the border size. So this behavior almost does the inverse of that to maintain border width.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
julenka: Good point. I created #5201 to track fixing this.

PR #5173, opened 2019-07-02
Input playback hand devices

lukastoenneMS:
## Overview

This gives the InputPlaybackService its own set of hand controllers, independently from the InputSimulationService.

The reason is that playing back input animation will conflict with user input in the editor, and typically input simulation with mouse and keyboard overrides animation. User input currently has to be disabled (see [UserInputEnabled](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/53d6ac190089a4856811df49cdf91c59f8951120/Assets/MixedRealityToolkit.Services/InputSimulation/IInputSimulationService.cs#L25) to allow the animation to modify hand data. With separate hand controllers for simulation and for animation this is no longer necessary.

## Changes
- InputPlaybackService moved into the InputSimulation assembly, since it shares more features with InputSimulationService.
- BaseInputSimulationService shared by simulation and playback services: Common functionality for creating simulated hand devices, extracted from InputSimulationService.
- BaseHandVisualizer fix to make it act only on joint events of the actual controller, otherwise any hand controller would change all the visualizers in the scene.


Assets/MixedRealityToolkit.SDK/Profiles/HoloLens2/DefaultHoloLens2InputSystemProfile.asset
```diff
@@ -85,10 +85,10 @@ MonoBehaviour:
     deviceManagerProfile: {fileID: 11400000, guid: d0f5a7f6d1f9f0b4cb6eb35c797a0f04,
       type: 2}
   - componentType:
-      reference: Microsoft.MixedReality.Toolkit.Input.InputPlaybackService, Microsoft.MixedReality.Toolkit.Services.InputAnimation
+      reference: Microsoft.MixedReality.Toolkit.Input.InputPlaybackService, Microsoft.MixedReality.Toolkit.Services.InputSimulation.Editor
```
wiwei: >reference [](start = 6, length = 9)

It might also be necessary to update the DefaultHoloLens1 instance of the input profile as well

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/BaseInputSimulationService.cs
```diff
@@ -0,0 +1,155 @@
...
+                case HandSimulationMode.Gestures:
+                    controller = new SimulatedGestureHand(TrackingState.Tracked, handedness, inputSource);
+                    break;
+                default:
+                    controller = null;
+                    break;
+            }
+
+            System.Type controllerType = simulationMode == HandSimulationMode.Gestures ? typeof(SimulatedGestureHand) : typeof(SimulatedArticulatedHand);
+            if (controller == null)
```
wiwei: >controller [](start = 16, length = 10)

I think this should be controllerType (judging by how Debug.LogError is mentioning controllerType

wiwei: Ah, I guess maybe the line before was just for creating a more debug-friendly message. I guess in this case it was weird to read a set and then a null check for something that wasn't just set.

---
In reply to: [314510681](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5173#discussion_r314510681) [](ancestors = 314510681)

wiwei: Yeah nevermind :)

---
In reply to: [314511235](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5173#discussion_r314511235) [](ancestors = 314511235,314510681)

Issue #5174, opened 2019-07-02
Screenshot Tool


Cameron-Micka:
## Describe the problem

Taking screenshots in Unity for MRTK documentation and promotional imagery can be burdensome and the output often looks less than desirable. 

## Describe the solution you'd like

I would like a screenshot tool that outputs images at "print media" resolutions and can export the current camera view with a transparent background.

## Additional context

This tool can be used not only for documentation but helping developers showcase their application for the store or other media outlets.

Cameron-Micka:
Fixed in https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5175

PR #5175, opened 2019-07-02
Screenshot Utility

Cameron-Micka:
## Overview

Often taking screenshots in Unity for MRTK documentation and promotional imagery can be burdensome and the output often looks less than desirable. This is where the ScreenshotUtility class comes into play. 

This class aides in taking screenshots via menu items and public APIs. Screenshots can be capture at various resolutions and with the current camera's clear color or a transparent clear color for use in easy post compositing of images.

Menu Item:
ScreenshotMenuItem

Output Image:
Screenshot_2019-07-02_01-47-14-PM

Promotional Example:
CLO17_casualOffice_012

CLO17_casualOffice_012_mrtk

This tool can be used not only for documentation but helping developers showcase their application for the store or other media outlets.

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5174

## Verification
Please try the tool within the Unity editor for functionality and usability.

>
>
>


Assets/MixedRealityToolkit/Utilities/Editor/ScreenshotUtility.cs
```diff
@@ -0,0 +1,146 @@
+﻿using System;
```
julenka: Remember to add the copyright block

Cameron-Micka: Good catch! I'll fix this.

Cameron-Micka:
@julenka I can certainly write some tests and documentation. Do you know if the build machine's test run in Unity headless mode, or just normal Unity. My tests might fail if their isn't a valid device context on the build machine. Also, is it fine to be writing out a few MBs of screenshot images in the tests? I don't want to pollute the build machines if the workspaces don't get wiped.

Cameron-Micka:
@julenka added some tests and documentation (my first time adding MRTK tests). Please let me know if anything else should be tested!

Cameron-Micka:
Looks like the test failed because the test runner could not write to "C:\Windows\ServiceProfiles\NetworkService\Desktop\Screenshot_2019-07-03_01-59-37-PM.png" Is there another way to test this on the build machine? Note, the tests run successfully locally.

Cameron-Micka:
Any chance we can get this reviewed/merged in soon? Thank you!

Assets/MixedRealityToolkit/Utilities/Editor/ScreenshotUtility.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            // Create a camera clone with a transparent clear color.
+            var renderCamera = new GameObject().AddComponent<Camera>();
+            renderCamera.transform.position = camera.transform.position;
+            renderCamera.transform.rotation = camera.transform.rotation;
+            renderCamera.clearFlags = transparentClearColor ? CameraClearFlags.Color : camera.clearFlags;
+            renderCamera.backgroundColor = transparentClearColor ? new Color(0.0f, 0.0f, 0.0f, 0.0f) : camera.backgroundColor;
+            renderCamera.fieldOfView = camera.fieldOfView;
+            renderCamera.nearClipPlane = camera.nearClipPlane;
+            renderCamera.farClipPlane = camera.farClipPlane;
```
Railboy: Would be nice to capture orthographic cameras as well.

Assets/MixedRealityToolkit/Utilities/Editor/ScreenshotUtility.cs
```diff
@@ -0,0 +1,152 @@
...
+            Debug.LogFormat("Screenshot captured to: {0}", path);
+
+            return true;
+        }
+
+        /// <summary>
+        /// Gets a screenshot file name based on time and located on the current user's desktop.
+        /// </summary>
+        /// <returns>The filename of the screenshot.</returns>
+        public static string GetScreenshotFileName()
```
Railboy: It would be really nice to pull this from `MixedRealityPreferences` so we could set the output path and screenshot format in the project settings.

Cameron-Micka: Oh I like that idea, I can add that!

Assets/MixedRealityToolkit/Utilities/Editor/ScreenshotUtility.cs
Cameron-Micka: That's a good idea, let me try that out.

Assets/MixedRealityToolkit/Utilities/Editor/ScreenshotUtility.cs
Cameron-Micka: I just started going down this path but decided against it to keep this functionality independent of the MRTK (in case someone want's to just drop this script into older MRTK versions, or projects without the MRTK).

julenka:
> Do you know if the build machine's test run in Unity headless mode, or just normal Unity. My tests might fail if their isn't a valid device context on the build machine. 

The machines run in Unity headless mode. If you can't actually capture the screen output on the device, perhaps there are ways to verify that other parts of the system are working?

> Also, is it fine to be writing out a few MBs of screenshot images in the tests? I don't want to pollute the build machines if the workspaces don't get wiped.

As part of cleanup, the generated images should be deleted.

Cameron-Micka:
Awesome thank you for the info! It looks like the tests passed? So, something must have been captured. I also clean up any files generated in the tests.

julenka:
Hi @cameron-micka looks like the PR check failed, could you please have a look?

PR #4616, opened 2019-05-29
UPM compliant packages for MRTK (new PR)

ei2kpi-ptc:
## Overview
Creates UPM compliant packages for Unity 2018.3 / 2018.4 (LTS)
Simplified PR from #4404 which was abandoned.

## Changes
- Fixes: #4403 .

## Verification
Please specifically check out the changes in MixedRealityToolkitFiles.cs since it looks like there is a fair amount of code there that does string manipulation to find the MRTK folders and assets. I am pretty sure I fixed most of code paths but would like to get some extra coverage here.

PR #4404, opened 2019-05-16
UPM compliant packages for MRTK

ei2kpi-ptc:
## Overview
Creates UPM compliant packages for Unity 2018.3

## Changes
- Fixes: #4403 .

## Verification
- Please specifically check out the changes in `MixedRealityToolkitFiles.cs` since it looks like there is a fair amount of code there that does string manipulation to find the MRTK folders and assets. I am pretty sure I fixed most of code paths but would like to get some extra coverage here.
- I already have an Azure Pipeline build that zips up and creates the UPM compliant packages in my fork. See release here: https://github.com/ei2kpi-ptc/MixedRealityToolkit-Unity/releases/tag/v2.0.0-RC1-Refresh-UPM. It would be great if we could add this to the MRTK releases as well since it is just a simple archive of the folders.

Issue #4403, opened 2019-05-16
Support 2018.3 Unity Package Manager


ei2kpi-ptc:
## Describe the problem

Importing MRTK via .unitypackages is always painful since it pulls it straight into the Assets folder.

## Describe the solution you'd like

Unity 2018 supports Unity Packages through the UPM system. 2018.3 does not support repos outside of Unity's own package repo but it does support local packages through the UI.

## Describe alternatives you've considered

I know we are planning to support Nuget but it looks like that might still be a little ways away?

## Additional context

I already have a PR + Azure Pipeline working for this...

Yoyozilla:
This is not GA blocking. we can do this after GA.

wiwei:
Alrighty, let's use this issue to get feedback on UPM support + desire to get it going forward. I spent some time last month trying this out (based on the work that happened in this PR: https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4616) - for the most part things were good, but there were a few issues with UPM + MRTK in general where we didn't have time in the current iteration to release with quality.

There are a few issues when consuming MRTK in UPM:
1) Pickers (i.e. the very basic thing that shows up when you try to select a profile) don't show assets in the UPM package location. I assume this is because of a distinction where assets in the Packages location are intended to be immutable (or maybe this is unrelated, but a side effect). So to work around this we'd either have to reimplement a picker, or more probably have some sort of code import step like text mesh pro (and then figure out how to do versioning between things)
2) Certain types of assets (like audio) can't be referenced from your scene directly if the audio file is in the packages folder. Similar to 1), this would mean we'd probably have to do the text mesh pro import for the set of asset types that need this behavior.
3) This would bifurcate the development process for folks using MRTK using as sources or compiled binaries. so we'd need to figure out if there is a single flow that we could support, or how we support now the two different ways of consuming it (i.e. do people who grab .unitypackages or compiled binaries also need to do this update?)
4) Example scenes can't be directly opened in when in the packages/ folder, meaning again 1) is necessary in order to copy things over (similar to text mesh pro)

If none of those things 1-4 are interesting to consumers (i.e. they just want the code directly) then consuming as UPM does work fine. I ultimately opted to not do the half-way work here because I'd ideally be able to push out a story that worked well end to end (and 1-4 prevent it from working well end to end)

yacuzo:
My thoughts on this as a UPM user:

[...]

PR #5179, opened 2019-07-02
Added ability for MRTK to appear as a virtual Package Manager Package

jbienzms:
## Overview
Added the ability for MRTK, when pulled from GitHub to a local folder, to appear as a virtual package in Package Manager. Please see updates made to [DownloadingTheMRTK.md](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/def4d6e77ae83dcde237805fed684f0a1df06303/Documentation/DownloadingTheMRTK.md) for how this can be used and why it's valuable.

## Changes
- Added **Assets\package.json** which is the virtual package descriptor.
- Updated **DownloadingTheMRTK.md** with information about the 3 approaches and why developers may wish to use each.

## Verification
Other that documentation, the only real change is adding a **package.json** file to the **Assets** folder. I have tested that this file doesn't break opening the MRTK in Unity, and I have tested that this file does in fact allow the MRTK to appear as a virtual package in other projects. Please just verify this to be the case on your machine as well.

### DIFF
--- a/Assets/package.json
+++ b/Assets/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "com.microsoft.mixedrealitytoolkit",
+  "name": "com.microsoft.mixedreality.toolkit.all",
   "version": "2.0.0",
   "displayName": "Mixed Reality Toolkit",
   "description": "MRTK virtual package.",
@@ -19,7 +19,6 @@
   ],
   "author": {
     "name": "Microsoft",
-    "email": "opencode@microsoft.com",
     "url": "https://github.com/microsoft/MixedRealityToolkit-Unity"
   } 
 }
--- a/Assets/package.json
+++ b/Assets/package.json
@@ -2,7 +2,7 @@
   "name": "com.microsoft.mixedrealitytoolkit",
   "version": "2.0.0",
   "displayName": "Mixed Reality Toolkit",
-  "description": "MRTK virtal package.",
+  "description": "MRTK virtual package.",
   "unity": "2018.4",
   "unityRelease": "0f1",
   "dependencies": {
--- a/Assets/package.json
+++ b/Assets/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "com.microsoft.mixedrealitytoolkit",
+  "version": "2.0.0",
+  "displayName": "Mixed Reality Toolkit",
+  "description": "MRTK virtal package.",
+  "unity": "2018.4",
+  "unityRelease": "0f1",
+  "dependencies": {
+ },
+ "keywords": [
+    "holotoolkit",
+	"holotoolkit-unity",
+	"mixed-reality",
+	"hololens",
+	"mixedrealitytoolkit",
+	"mrtk",
+	"mixedrealitytookit-unity",
+	"openvr"
+  ],
+  "author": {
+    "name": "Microsoft",
+    "email": "opencode@microsoft.com",
+    "url": "https://github.com/microsoft/MixedRealityToolkit-Unity"
+  } 
+}
--- a/Assets/package.json
+++ b/Assets/package.json
@@ -14,7 +14,7 @@
 	"hololens",
 	"mixedrealitytoolkit",
 	"mrtk",
-	"mixedrealitytookit-unity",
+	"mixedrealitytoolkit-unity",
 	"openvr"
   ],
   "author": {
--- a/Assets/package.json.meta
+++ b/Assets/package.json.meta
@@ -1,0 +1,7 @@
+fileFormatVersion: 2
+guid: 30c70a5f6f83acc4e912f668dd2c2b24
+TextScriptImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
--- a/Documentation/DownloadingTheMRTK.md
+++ b/Documentation/DownloadingTheMRTK.md
@@ -2,22 +2,43 @@
 
 ![](../Documentation/Images/MRTK_Logo_Rev.png)
 
-The MRTK is available via the following methods. For feedback on these methods, or to request an additional distribution method, please [file an issue on Github](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/new/choose)
+Below you'll find several methods for importing the MRTK into your project. For feedback on these methods, or to request an additional distribution method, please [file an issue on GitHub](https://github.com/Microsoft/MixedRealityToolkit-Unity/issues/new/choose)
 
-## Unity asset downloadable from the MRTK GitHub site
+## Unity Asset Packages - Easy to get started
 
-You can download the packaged [Unity assets from here](https://github.com/Microsoft/MixedRealityToolkit-Unity/releases) for importing into your project.
+The easiest way to get started is to download MRTK as a set of Unity Asset Packages. These can be downloaded from the [Releases](https://github.com/Microsoft/MixedRealityToolkit-Unity/releases) page of the MRTK project. 
 
-There are two Unity packages available for each release:
+There are two asset packages for each release:
 
-### Foundation
-This package contains the MRTK itself and all of the code and materials needed to build Mixed Reality experiences.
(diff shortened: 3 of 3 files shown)


jbienzms:
P.S. I added the Hack Week tag to this because we found out how to do this from Jerome at Unity during the London Hack Week. Several partners at that event indicated a desire to import MRTK in this way.

Assets/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "com.microsoft.mixedrealitytoolkit",
+  "version": "2.0.0",
+  "displayName": "Mixed Reality Toolkit",
+  "description": "MRTK virtal package.",
```
keveleigh:
```suggestion
  "description": "MRTK virtual package.",
```

Assets/package.json
```diff
@@ -0,0 +1,25 @@
...
+	"mixed-reality",
+	"hololens",
+	"mixedrealitytoolkit",
+	"mrtk",
+	"mixedrealitytookit-unity",
+	"openvr"
+  ],
+  "author": {
+    "name": "Microsoft",
+    "email": "opencode@microsoft.com",
```
keveleigh: I wonder if this should be an MRTK team email address?

Assets/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "dependencies": {
+ },
+ "keywords": [
+    "holotoolkit",
+	"holotoolkit-unity",
+	"mixed-reality",
+	"hololens",
+	"mixedrealitytoolkit",
+	"mrtk",
+	"mixedrealitytookit-unity",
```
keveleigh:
```suggestion
	"mixedrealitytoolkit-unity",
```

Assets/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "com.microsoft.mixedrealitytoolkit",
```
keveleigh: Is this / should this PR be related to #4616?
Commenting specifically on the name, since there was some discussion either on that or an earlier version on naming conventions

keveleigh: I can't find that discussion at the moment though. Maybe @andreiborodin remembers?

Assets/package.json
jbienzms: Thanks for catching this @keveleigh. This looks like a great fix and I'm trying to accept it, but I'm getting this error:

image

I'm wondering if it's because of where I created the branch? When I went to commit this branch to source control I got an error about needing to do a pull request instead, but all of the files were actually pushed up into the branch. I'm wondering if this error about not being able to apply the suggestion is somehow due to a DevOps rule?

Assets/package.json
jbienzms: Yes, I agree. I reached out to @wiwei on this yesterday and asked what e-mail to use. He said there wasn't a public one for MRTK but he was going to bring it up in an internal thread. I'd be happy to change this to whatever the team feels it should be.

Assets/package.json
jbienzms: Same issue as above. I cannot commit the change.

Assets/package.json
jbienzms: Confirmed. This is because it's a "protected branch".

image

@keveleigh or @davidkline-ms any idea why I was able to create this branch and push to it, but now I can't update it? I see lots of feature branches. Do I just not have the right permissions as a contributor or how should we move forward?

Assets/package.json
jbienzms: @keveleigh Good point. This is very very similar to #4616. This PR is intended to allow you to easily bring MRTK pulled down from GitHub into a Unity project. It appears #4616 is about packaging MRTK up as a UPM for distribution, but I _think_ it could work the same way. I will start a question over there and see what comes of it.

Assets/package.json
jbienzms: Turns out e-mail address is optional. I missed that in the schema definition. I just removed the address altogether.

Assets/package.json
jbienzms: I have updated the package name based on my comments on #4616 [here](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4616#issuecomment-508151668). Assuming you guys agee with that approach, I believe this is resolved.

jbienzms:
Yikes. It looks like this may not fully work as intended. At first glance, everything was working great in the editor. But it appears the inspectors don't work correctly if MRTK is not actually in a folder.

`Failed to locate MixedRealityToolkit folders in the project.
UnityEngine.Debug:LogError(Object)
Microsoft.MixedReality.Toolkit.Utilities.Editor.MixedRealityToolkitFiles:MapRelativeFilePath(MixedRealityToolkitModuleType, String) (at C:/Users/jbienz/Code/Microsoft/MixedRealityToolkit-Unity/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs:234)
Microsoft.MixedReality.Toolkit.Utilities.Editor.MixedRealityToolkitFiles:MapRelativeFilePath(String) (at C:/Users/jbienz/Code/Microsoft/MixedRealityToolkit-Unity/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs:222)
Microsoft.MixedReality.Toolkit.Utilities.Editor.MixedRealityInspectorUtility:.cctor() (at C:/Users/jbienz/Code/Microsoft/MixedRealityToolkit-Unity/Assets/MixedRealityToolkit/Inspectors/Utilities/MixedRealityInspectorUtility.cs:41)
Microsoft.MixedReality.Toolkit.Editor.BaseMixedRealityToolkitConfigurationProfileInspector:RenderMRTKLogo() (at C:/Users/jbienz/Code/Microsoft/MixedRealityToolkit-Unity/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityToolkitConfigurationProfileInspector.cs:69)
Microsoft.MixedReality.Toolkit.Editor.MixedRealityToolkitConfigurationProfileInspector:OnInspectorGUI() (at C:/Users/jbienz/Code/Microsoft/MixedRealityToolkit-Unity/Assets/MixedRealityToolkit/Inspectors/Profiles/MixedRealityToolkitConfigurationProfileInspector.cs:178)
Microsoft.MixedReality.Toolkit.Editor.MixedRealityToolkitInspector:OnInspectorGUI() (at C:/Users/jbienz/Code/Microsoft/MixedRealityToolkit-Unity/Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs:117)
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr)`

I wonder if this is solved by #4616, since that PR does appear to update the code in **MixedRealityToolkitFiles.cs**.

yacuzo:
The UPM terminology for a package referenced by path is a 'local package'. Please stick to the same term to not confuse anyone. 

#4616 does exactly the same thing, but is more complete it seems. You have to import the 4 folders as different packages though. I'm using #4616 as a local package and it works fine, so this PR seems redundant to me.

wiwei:
Hold off on this for a bit, I wanted to circulate some thoughts on what we're gonna do for UPM.

jbienzms:
Just wanted to note that this PR can probably be abandoned in favor of either #4616 or #4404 if one of those are accepted. I believe they do a better job of handling file paths than my attempt an also represents the MRTK as individual packages.

julenka:
Hi Jared, it turns out that adding UPM support has a long tail of issues and bugs that will has been working on for a while:
From #4403:

> I spent some time last month trying this out (based on the work that happened in this PR: #4616) - for the most part things were good, but there were a few issues with UPM + MRTK in general where we didn't have time in the current iteration to release with quality. 

Will outlined the 4 issues he had in #4403. 

From Will:

> I ultimately opted to not do the half-way work here because I'd ideally be able to push out a story that worked well end to end (and 1-4 prevent it from working well end to end)

This PR would have the same issues as those outlined in #4403. So for now our plan is to close this PR and other UPM PRs, and instead track updating MRTK to work with UPM in #4403. The goal is to fix the issues 1 - 4 will mentions that prevent UPM from working well end to end.

PR #5182, opened 2019-07-03
Add guidance on how to find valid xrefs to reference.

wiwei:
It turns out that the syntax for xrefs is super hard and not actually easy to find. However, running docfx locally will generate a convenient map, which takes away the memorization process (you can just Ctrl+F for thing you want to link to)

Adding this nugget of guidance to our dev docs.

### DIFF
--- a/Documentation/Contributing/DevDocGuide.md
+++ b/Documentation/Contributing/DevDocGuide.md
@@ -63,6 +63,33 @@ This links to the [BoundarySystem API](xref:Microsoft.MixedReality.Toolkit.Bound
 as well as this short version: @Microsoft.MixedReality.Toolkit.Boundary
 ```
 
+## Enumerating available xrefs
+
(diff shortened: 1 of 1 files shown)


Documentation/Contributing/DevDocGuide.md
```diff
@@ -63,6 +63,33 @@ This links to the [BoundarySystem API](xref:Microsoft.MixedReality.Toolkit.Bound
 as well as this short version: @Microsoft.MixedReality.Toolkit.Boundary
 ```
 
+## Enumerating available xrefs
+
+xref syntax can be difficult to remember - it's possible to enumerate all of the available xref IDs by first running
```
thalbern:
```suggestion
Xref syntax can be difficult to remember - it's possible to enumerate all of the available xref IDs by first running
```

Xref looks weird but it's the beginning of a sentence so it should be uppercase I think?

PR #5187, opened 2019-07-03
Create a copy of CI definition using internal agent pool

macborow:
## Overview
We need a duplicate of the pipeline definition for now because it is not possible to re-assign the Agent Pool using the UI. This is temporary until we can shut down the feed in AIPMR.


macborow:
BTW, there's no need to run PR check for this change.

PR #5194, opened 2019-07-03
Fix regression caused by #5167

### LINKED ISSUE #5192: Speech Commands are not working in any scenes, in Editor
## Overview
Speech commands are not working in any scenes. To repro, grab latest mrtk_development, play any scene, and try to invoke "select" or any other speech command by pressing on the keyboard. You will see that no actual speech commands are invoked.

You can test this by opening the BoundingBoxRuntimeExample scene. The scene will update its state when you press the '1' key. The actual behavior is no state changes happen when 1 key is pressed.

In WindowsSpeechInputProvider it looks like the provider is exiting early, failing to set keywordRecognizer. If keywordRecognizer is null, then the speech system does not work. The return is failing early because Commands is nul.

``
        public override void Enable()
        {
            if (!Application.isPlaying || 
                (Commands == null) ||
                (Commands.Length == 0)) { return; }
```

davidkline-ms:
#5167 introduced a regression in accessing the registered speech commands.

This change fixes it by referencing the correct profile in the property.

Fixes #5192

Confirmed issue and verified fix using keyboard shortcuts in editor and stepping through the speech provider startup code in the debugger.

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsVoiceInput/WindowsSpeechInputProvider.cs
@@ -38,7 +38,7 @@ namespace Microsoft.MixedReality.Toolkit.Windows.Input
         /// <summary>
         /// The keywords to be recognized and optional keyboard shortcuts.
         /// </summary>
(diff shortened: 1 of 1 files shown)


wiwei:
Is it possible to have a test cover this really basic scenario (i.e. "select") and validate that it works?

Issue #5184, opened 2019-07-03
Logging of EyeGaze data to csv file is culture-specific


ChrisK284:
## Describe the bug
Logs of EyeGaze data are written culture-specific into .csv files. This means for cultures which use a comma instead of a dot as decimal separator (e.g. german) that the csv file gets currupted and can't be read for playback since it assumes commas as separators.
This happens both when using the recording on a PC and on HoloLens 2.

## To reproduce

**Variant a**
1. Start the EyeTrackingDemo-05-Visualizer Demo on a PC in Unity with culture settings applied that uses commas as decimal separators.
2. Record eye gaze data for heatmap.
3. Play heatmap data.
4. Nothing gets displayed.

**Variant b**
1. Deploy the EyeTrackingDemo-05-Visualizer Demo to a HoloLens 2 which was set up with a language that uses commas as decimal separators.
2. Record eye gaze data for heatmap.
3. Play heatmap data.
4. Nothing gets displayed

## Expected behavior

The logging should happen with dot decimal separator to be able to play back the records no matter which culture setting is used.

## Setup

- Unity 2018.4.2
- MRTK v2.0.0 RC2.1

## Target Platform

- PC
or
- HoloLens 2

PR #5195, opened 2019-07-03
Logging of EyeGaze data to csv file is culture-specific

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5184

On certain locales, the CSV delimiter isn't the comma, but the semicolon instead.

Previously, we assumed that the delimiter would always be the comma, but this would cause problems in cases like de-DE where numbers are written with commas (instead of decimals) to represent the decimal digits.

So for example, en_US:

123.123, ABCABC, XYZXYZ

Without this change, it would look like this on de_DE
123,123, ABCABC, XYZXYZ

This is obviously totally wrong - now we have four cells, it doesn't make sense!

The solution is to use the real delimiter for the current locale:
123,123; ABCABC; XYZXYZ

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/EditModeTests/Examples/Demos/EyeTracking/DemoVisualizer/Scripts/UserInputRecorderTest.cs
+++ b/Assets/MixedRealityToolkit.Tests/EditModeTests/Examples/Demos/EyeTracking/DemoVisualizer/Scripts/UserInputRecorderTest.cs
@@ -1,0 +1,36 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using NUnit.Framework;
+using System.Globalization;
+
+namespace Microsoft.MixedReality.Toolkit.Examples.Demos.EyeTracking.Logging
+{
+    class UserInputRecorderTests
+    {
+        [Test]
+        public void TestGetStringFormat()
+        {
+            // This test array is passed into UserInputRecorder.GetStringFormat and
+            // its content doesn't matter, only its length. The length is arbitrarily
+            // chosen to be relatively short.
+            object[] testArray = new object[5];
+
+            // en-US uses comma separated delimiters.
+            CultureInfo.CurrentCulture = new CultureInfo("en-US", false);
+
+            // In en-US, the generated format should be comma-delimited.
+            Assert.AreEqual(
+                "{0}, {1}, {2}, {3}, {4}",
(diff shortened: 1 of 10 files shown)


sostel:
Ehm, doesn't CSV stand for Comma Separated Values? 🤨 I'm not sure what you mean with "Without this change, it would look like this on de_DE: 123,123, ABCABC, XYZXYZ". On what do you mean?

wiwei:
Yeah it's actually SCSV.

The leading s is silent though

wiwei:
But actually:

https://stackoverflow.com/questions/8092718/how-do-i-detect-the-users-locale-to-get-the-correct-csv-separator
https://superuser.com/questions/606272/how-to-get-excel-to-interpret-the-comma-as-a-default-delimiter-in-csv-files

sostel:
> But actually:
> 
> https://stackoverflow.com/questions/8092718/how-do-i-detect-the-users-locale-to-get-the-correct-csv-separator
> https://superuser.com/questions/606272/how-to-get-excel-to-interpret-the-comma-as-a-default-delimiter-in-csv-files

Interesting! Thanks for addressing it, Will!

keveleigh:
>to avoid this issue is double-quote the relevant decimal numbers within the CSV file as such: `"10,20", "1,50", "This is another column"`

Was there a reason you went with semicolon over wrapping in `"`? I wonder if all CSV parsers handle semicolon correctly

Assets/MixedRealityToolkit.Tests/EditModeTests/Examples/Demos/EyeTracking/DemoVisualizer/Scripts/UserInputRecorderTest.cs
```diff
@@ -0,0 +1,36 @@
...
+            // In en-US, the generated format should be comma-delimited.
+            Assert.AreEqual(
+                "{0}, {1}, {2}, {3}, {4}",
+                UserInputRecorder.GetStringFormat(testArray));
+
+            // de-DE uses semicolon separated delimiters.
+            CultureInfo.CurrentCulture = new CultureInfo("de-DE", false);
+
+            // In de-DE, the generated format should be semicolon-delimited.
+            Assert.AreEqual(
```
keveleigh: In addition to checking for the correct delimiter, did you check that eye tracking playback from the original issue works with this new format?

wiwei: I did not!

PR #5215, opened 2019-07-05
MSBuild Project Generation

andreiborodin:
This change contains the necessary tooling to automatically generate an SDK style MSBuild C# projects and solution for the Unity project. This, after it's checked in, will be used in to build appropriate DLLs for NuGet packages being built.

The process goes as follows:
- Starting with MSBuildTools.GenerateSDKProjects
- First copy the Packages folder and clean up the output directories.
- After that , parse all the available platforms of the current editor and filter to the ones we currently focus on supporting.
- Next create a platforms props file for each of those platforms using the templates embedded next to the code. First try to find the concrete template for the platform/configuration combination (ie. UWP.Player.Template.Props), then a more general just platform, and then fall back to the common one.
- Now, parse the Unity project by instantiating "new UnityProjectInfo(platforms)". This fetches all assemblies embedded, as well as all the Dll files found in the project and parses all that data into CSProjectInfo and PluginAssemblyInfo. This includes the Packages folder.
- Then, begin the Export process with the solution adding a reference to each CSProj to be generated
- The above process calls into generate all the CSProject files as well adding all source, plugins, reference and configuring when each gets build.
- Finally, a CSMap is added for each assembly (that will be used in the future).
- All the output goes to a single root directory next to Assets: MSBuild.


Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/SolutionTemplate.sln.meta
```diff
@@ -0,0 +1,7 @@
+fileFormatVersion: 2
```
MenelvagorMilsom: I notice you have a `.meta` file for `SolutionTemplate.sln` and `SDKProjectTemplate.csproj`. These files don't seem to be checked in on this PR's branch. Also, when I synced your fork and tried to run the _Generate C# SDK Projects_ command from Unity, I got the following errors:
image
Did you forget to commit these files?

andreiborodin: Dang, yeah, these files got git ignored. Will re-add, thanks!

ryantrem:
Probably a future improvement, but instead of doing text replacement in the MSBuild template projects, you could use the [Microsoft.Build](https://www.nuget.org/packages/Microsoft.Build/15.9.20) package to edit the MSBuild. The advantage of this would be that you could edit projects non-destructively, which would allow users to modify the generated projects, check them in, and use them for other purposes (like referencing other MSBuild projects, etc.).

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/SDKProjectTemplate.csproj
```diff
@@ -0,0 +1,119 @@
+<Project>
+  <PropertyGroup>
+    <!--anborod: This is a weird thing, it is a required property (even if commented)-->
+    <!--<TargetFrameworks>netstandard2.0;uap10.0;net46</TargetFrameworks> -->
```
wiwei: >!--<TargetFrameworks>netstandard2.0;uap10.0;net46</TargetFrameworks> --> [](start = 5, length = 72)

If this is omitted, is it pretty obvious about the failures that start happening, or it is pretty obscure? This comment isn't asking for any changes here I'm just curious to know the story behind it.

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/SDKProjectTemplate.csproj
```diff
@@ -0,0 +1,119 @@
...
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <Configurations>InEditor;Player</Configurations>
+    <Platforms><!--SUPPORTED_PLATFORMS_TOKEN--></Platforms>
+    <UnityPlatform>$(Platform)</UnityPlatform>
+    <UnityConfiguration>$(Configuration)</UnityConfiguration>
+  </PropertyGroup>
+
+  <!-- This is special cased here for WSA player only-->
```
wiwei: > [](start = 2, length = 54)

Is this comment supposed to be for this line for the one below that looks specifically for WSA player?

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/SDKProjectTemplate.csproj
```diff
@@ -0,0 +1,119 @@
...
+  
+  <PropertyGroup Condition="'$(UnityConfiguration)'=='Player'">
+    <AssemblySearchPaths>$(AssemblySearchPaths);<!--PLAYER_ASSEMBLY_SEARCH_PATHS_TOKEN--></AssemblySearchPaths>
+  </PropertyGroup>
+  
+  <PropertyGroup>
+    <DefaultItemExcludes>$(DefaultItemExcludes)obj;bin;*.asmdef;*.asmdef.meta;*.csmap;</DefaultItemExcludes>
+  </PropertyGroup>
+
+  <!--PROJECT_REFERENCE_SET_TEMPLATE_START-->
```
wiwei: > [](start = 2, length = 43)

Super nit: Consistency with spacing for these temlpate start and ends (i.e. move this two spaces left, or move the end two to the right)

wiwei:
To make sure I understand a little better here, is this more like creating the build files via APIs instead of "by hand" (so if there were changes to the msbuild schema our code could still in theory continue to work even after schema changes, but the templates might get broken?)

---
In reply to: [509762183](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#issuecomment-509762183) [](ancestors = 509762183)

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/WSA.Player.Template.props
```diff
@@ -0,0 +1,32 @@
...
+    </Reference>
+    <Reference Include="WinRTLegacy">
+      <HintPath>$(UnityEditorInstallFolder)Data\PlaybackEngines\MetroSupport\Managed\UAP\WinRTLegacy.dll</HintPath>
+    </Reference>
+
+    <!--PLATFORM_COMMON_REFERENCE_TEMPLATE_START-->
+    <Reference Include="##REFERENCE_TOKEN##">
+      <HintPath><!--HINT_PATH_TOKEN--></HintPath>
+    </Reference>
+<!--PLATFORM_COMMON_REFERENCE_TEMPLATE_END-->
```
wiwei: Super nit: indent four spaces

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/WSA.Player.Template.props
```diff
@@ -0,0 +1,32 @@
+<Project>
+  <PropertyGroup>
+    <AssemblySearchPaths>$(AssemblySearchPaths);<!--PLATFORM_COMMON_ASSEMBLY_SEARCH_PATHS_TOKEN-->;C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETCore\v4.5;$(UnityEditorInstallFolder)Data\PlaybackEngines\MetroSupport\Managed\UAP\</AssemblySearchPaths>
```
wiwei: >C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETCore\v4.5 [](start = 99, length = 77)

These are things that are always installed to C: right (or the windows install drive?) Is there any risk that these might actually be on another drive letter?

Assets/MixedRealityToolkit/MSBuild/Scripts/AssemblyDefinitionInfo.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        /// <summary>
+        /// After it's parsed from JSON, this method should be invoked to validate some of the values and set additional properties.
+        /// </summary>
+        public void Validate()
+        {
+            if (excludePlatforms.Length > 0 && includePlatforms.Length > 0)
+            {
+                Debug.LogError($"Assembly definition file '{name}' contains both excluded and included platform list, will refer to only included.");
+                excludePlatforms = Array.Empty<string>();
```
wiwei: If we just threw here instead of logging an error, would that halt our build process? I'm curious because this is also something that Unity itself says is not supported, though maybe as is done here it just defaults to one of them:

[https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html](https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html)

In our case if we're hitting this we should probably just fail the build and make people fix it, rather than silently (because the build will keep going). If this isn't invoked from the build then this is fine.

Assets/MixedRealityToolkit/MSBuild/Scripts/CompilationPlatformInfo.cs
```diff
@@ -0,0 +1,176 @@
...
+                flags = AssemblyBuilderFlags.None
+            };
+
+            HashSet<string> platformCommonDefines = new HashSet<string>(builder.defaultDefines);
+            HashSet<string> playerDefines = new HashSet<string>(builder.defaultDefines);
+
+            HashSet<string> platformCommonReferences = new HashSet<string>(FilterOutProjectReferences(builder.defaultReferences));
+            HashSet<string> playerReferences = new HashSet<string>(FilterOutProjectReferences(builder.defaultReferences));
+
+            builder.flags = AssemblyBuilderFlags.EditorAssembly;
```
wiwei: >builder.flags = AssemblyBuilderFlags.EditorAssembly; [](start = 12, length = 52)

Can you add a comment for why this has to be set after the creation of the builder? A naive reading of this would maybe have a person just move this to the constructor.

wiwei: But clearly that would be bad, otherwise it wouldn't have been done this way.

---
In reply to: [301785013](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301785013) [](ancestors = 301785013)

wiwei: Ah, I guess what's weird to me is that setting the property actually changes the hashsets of other properties, I guess that was what was throwing me off. Nevermind!

---
In reply to: [301785122](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301785122) [](ancestors = 301785122,301785013)

Assets/MixedRealityToolkit/MSBuild/Scripts/CompilationPlatformInfo.cs
```diff
@@ -0,0 +1,176 @@
...
+            playerReferences.RemoveWhere(t => platformCommonReferences.Contains(t));
+            inEditorReferences.RemoveWhere(t => platformCommonReferences.Contains(t));
+
+            // Get specialized
+            playerDefines.RemoveWhere(t => platformCommonDefines.Contains(t));
+            inEditorDefines.RemoveWhere(t => platformCommonDefines.Contains(t));
+
+            playerReferences.RemoveWhere(t => platformCommonReferences.Contains(t));
+            inEditorReferences.RemoveWhere(t => platformCommonReferences.Contains(t));
+
```
wiwei: Nit: For future cases, consider using things like Except, Intersect, Union and the corresponding HashSet operating functions:

[https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.exceptwith?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.exceptwith?view=netframework-4.8)

If only because it can be at times more readable to read the fundamental set operation words (i.e. this thing  = A intersect B) instead of going A but remove everything from one and then do the same to the other and then use the stuff that remains.

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectDependency.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#if UNITY_EDITOR
+using System.Collections.Generic;
+using UnityEditor;
+
+namespace Microsoft.MixedReality.Toolkit.MSBuild
+{
+    public class CSProjectDependency<T>
+    {
```
wiwei: Nit: Please add some doccomments.

Assets/MixedRealityToolkit/MSBuild/Scripts/ReferenceItemInfo.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// In the editor, we can support all patforms if it's a pre-defined assembly, or an asmdef with Editor platform checked. 
+        /// Otherwise we fallback to just the platforms specified in the editor.
+        /// </remarks>
+        public IReadOnlyDictionary<BuildTarget, CompilationPlatformInfo> InEditorPlatforms { get; protected set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <remarks>
+        /// In the player, we support any platform if pre-defined assembly, or the ones explicitly specified in the AsmDef player.
```
wiwei: Nit: It's weird seeing the summaries empty here, but having remarks (i.e. clarifying comments, but no main comments to clarify)

Assets/MixedRealityToolkit/MSBuild/Scripts/ReferenceItemInfo.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <summary>
+        /// Gets name of the reference item.
+        /// </summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// </summary>
+        /// <remarks>
+        /// In the editor, we can support all patforms if it's a pre-defined assembly, or an asmdef with Editor platform checked.
```
wiwei: >patforms [](start = 46, length = 8)

Nit: platforms

wiwei: Thought also patforms I can get behind

---
In reply to: [301789756](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301789756) [](ancestors = 301789756)

wiwei: >In the editor, we can support all patforms if it's a pre-defined assembly, or an asmdef with Editor platform checked.  [](start = 12, length = 118)

I'm trying to understand this comment a little more - is this saying that if if the overall thing is being built for editor + assembly, or editor + asmdef, this gets ignored? i.e. so the usage of this is done somewhat externally of this class?

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.MSBuild
+{
+    /// <summary>
+    /// From which source was the project created.
+    /// </summary>
+    public enum ProjectType
+    {
+        /// <summary>
+        /// The project is backed by an Assembly-Definition file.
```
wiwei: >Assembly-Definition [](start = 40, length = 19)

Nit: assembly definition (capitilization here and down below) or make every usage of this term have this capitalization (just to be consistent at least within this file)

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+            switch (assembly.name)
+            {
+                case "Assembly-CSharp":
+                case "Assembly-CSharp-firstpass":
+                    return ProjectType.PredefinedAssembly;
+                case "Assembly-CSharp-editor":
+                case "Assembly-CSharp-firstpass-editor":
+                    return ProjectType.PredefinedEditorAssembly;
+                default:
+                    Debug.LogError($"Predefined assembly '{assembly.name}' was not recognized, this generally means it should be added to the switch statement in CSProjectInfo:GetProjectType. Treating is as a PredefinedAssembly instead of PredefinedEditorAssembly.");
```
wiwei: >Debug.LogError($"Predefined assembly '{assembly.name}' was not recognized, this generally means it should be added to the switch statement in CSProjectInfo:GetProjectType. Treating is as a PredefinedAssembly instead of PredefinedEditorAssembly."); [](start = 20, length = 247)

Another question about how this surfaces in the build - it would be useful if we could throw/abort the build entirely (if we aren't going to invest in build logging changes to grab from STDERR or parse errors). If the plan, however, is to update the build steps to look for error logs (of this specific type) then this is fine - my main concern is that usually we haven't done this, and these sorts of things can fail silently (because they only get exercised in batch mode and on CI usually we don't look at the logged errors, only the overall failure/success).

The downside to throwing, of course, is that we'll only see a single error (and we would have to re-run things possibly multiple times to get all errors, so really there are pros and cons to both ways)

Assets/MixedRealityToolkit/MSBuild/Scripts/ReferenceItemInfo.cs
wiwei: Actually, this might actually be a sin of documentation, but it might be helpful just to point to:

private ReadOnlyDictionary<BuildTarget, CompilationPlatformInfo> GetCompilationPlatforms(bool inEditor)

Which is what defines the rules here.

---
In reply to: [301791189](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301791189) [](ancestors = 301791189)

Assets/MixedRealityToolkit/MSBuild/Scripts/MSBuildTools.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            BuildTarget.StandaloneWindows,
+            BuildTarget.StandaloneWindows64,
+            BuildTarget.iOS,
+            BuildTarget.Android,
+            BuildTarget.WSAPlayer
+        };
+
+        public const string CSharpVersion = "7.3";
+        //public const string UWPTargetPlatformVersion = "10.0.18362.0";
```
wiwei: >//public const string UWPTargetPlatformVersion = "10.0.18362.0"; [](start = 8, length = 64)

Nit: Delete commented out code

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+            bool returnNoPlatforms = (!inEditor && ProjectType == ProjectType.PredefinedEditorAssembly)
+                || (!inEditor && ProjectType == ProjectType.EditorAsmDef)
+                || (!inEditor && ProjectType == ProjectType.AsmDef && AssemblyDefinitionInfo.TestAssembly);
+
+            if (returnNoPlatforms)
+            {
+                return new ReadOnlyDictionary<BuildTarget, CompilationPlatformInfo>(new Dictionary<BuildTarget, CompilationPlatformInfo>());
+            }
+
+            // We only are an asmdef at this point, as we handle all predefined assembly at line 130, inEditor && predefined editor assembly at line 130, and !inEditor && predefined editor assembly at line 139
```
wiwei: >line 130, inEditor && predefined editor assembly at line 130, and !inEditor && predefined editor assembly at line 139 [](start = 92, length = 117)

Nit: I would probably add some other named comment sections:

i.e. on line 130, have a comment saying "predefined assembly reconciliaion section", and then here refer to the named section.

Line numbers can and will change arbitrarily and this comment will go out of date pretty easily (and it's really easy to forget updating this as you change any of the above lines)

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+        internal void AddDependency(PluginAssemblyInfo pluginAssemblyInfo)
+        {
+            AddDependency(pluginAssemblyDependencies, pluginAssemblyInfo);
+        }
+
+        private void AddDependency<T>(List<CSProjectDependency<T>> items, T referenceInfo) where T : ReferenceItemInfo
+        {
+            items.Add(new CSProjectDependency<T>(referenceInfo,
+                new HashSet<BuildTarget>(InEditorPlatforms.Keys.Intersect(referenceInfo.InEditorPlatforms.Keys)),
+                new HashSet<BuildTarget>(PlayerPlatforms.Keys.Intersect(referenceInfo.PlayerPlatforms.Keys))));
```
wiwei: This takes an intersection here because our new dependency can only support the subset that both of these references support (i.e. if we support A, B, this other thing supports B, C, then we can only really support B between the two of us) right?

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+                    { projectReferenceSetTemplate, projectReferences },
+                    { sourceIncludeTemplate, string.Join("\r\n", sourceIncludes) },
+                    { suportedPlatformBuildConditionTemplate, string.Join("\r\n", supportedPlatformBuildConditions) }
+                };
+
+                projectFileTemplateText = Utilities.ReplaceTokens(projectFileTemplateText, tokens);
+            }
+            else
+            {
+                Debug.LogError("Failed to find ProjectReferenceSet and/or Source_Include templates in the project template file.");
```
wiwei: > Debug.LogError("Failed to find ProjectReferenceSet and/or Source_Include templates in the project template file."); [](start = 15, length = 116)

Nit: It might be helpful to log the missing files.

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+                    { "<!--SUPPORTED_PLATFORMS_TOKEN-->", string.Join(";", availablePlatforms.Select(t=>t.Name)) },
+
+                    { "<!--INEDITOR_ASSEMBLY_SEARCH_PATHS_TOKEN-->", string.Join(";", inEditorSearchPaths) },
+                    { "<!--PLAYER_ASSEMBLY_SEARCH_PATHS_TOKEN-->", string.Join(";", playerSearchPaths) },
+
+                    { "##PLATFORM_PROPS_FOLDER_PATH_TOKEN##", propsOutputFolder },
+
+                    { projectReferenceSetTemplate, projectReferences },
+                    { sourceIncludeTemplate, string.Join("\r\n", sourceIncludes) },
+                    { suportedPlatformBuildConditionTemplate, string.Join("\r\n", supportedPlatformBuildConditions) }
```
wiwei: Are any of these optional? What I'm curious about is if these templates end up getting validated in making sure that they have the required tokens. i.e. it's pretty easy for me to make a mispelling in a token, and if so, substitution here would silently fail (even though I would expect that tokens should be validated as being valid and present)

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+                Dictionary<string, string> sourceGuidToClassName = new Dictionary<string, string>();
+                foreach (string source in Assembly.sourceFiles)
+                {
+                    ProcessSourceFile(source, sourceIncludeTemplate, sourceIncludes, sourceGuidToClassName);
+                }
+
+                File.WriteAllLines(Path.Combine(propsOutputFolder, $"{Guid.ToString()}.csmap"), sourceGuidToClassName.Select(t => $"{t.Key}:{t.Value}"));
+
+                bool bothConfigurations = PlayerPlatforms.Count > 0;
+
```
wiwei: I think this variable is unused

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+                    {
+                        //TODO Native plugins aren't yet supported
+                        continue;
+                    }
+                    List<string> platformConditions = GetPlatformConditions(inEditor ? InEditorPlatforms : PlayerPlatforms, inEditor ? dependency.InEditorSupportedPlatforms : dependency.PlayerSupportedPlatforms);
+
+                    pluginReferences.Add(Utilities.ReplaceTokens(pluginReferenceTemplate, new Dictionary<string, string>()
+                    {
+                        { "##REFERENCE_TOKEN##", dependency.Dependency.Name },
+                        { "##CONDITION_TOKEN##", platformConditions.Count == 0 ? "'false'" : string.Join(" OR ", platformConditions)},
```
wiwei: >"'false'" [](start = 81, length = 9)

Do you know why this condition token needs to be double-encapsulated by " and ', but the one on line 287 doesn't?

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+                        { "##CONDITION_TOKEN##", platformConditions.Count == 0 ? "false" : string.Join(" OR ", platformConditions)}
+                    }));
+                }
+
+                List<string> pluginReferences = new List<string>();
+                foreach (CSProjectDependency<PluginAssemblyInfo> dependency in pluginAssemblyDependencies)
+                {
+                    if (dependency.Dependency.Type == PluginType.Native)
+                    {
+                        //TODO Native plugins aren't yet supported
```
wiwei: >//TODO Native plugins aren't yet supported [](start = 24, length = 42)

We don't like to have blank TODOs checked in here - if you can add more details or have an issue filed (and then link to that issue) that would be best. Otherwise this will probably live on forever with little background

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
```diff
@@ -0,0 +1,363 @@
...
+
+        private string GetDefaultPlatform(IReadOnlyDictionary<BuildTarget, CompilationPlatformInfo> platforms)
+        {
+
+            if (platforms.TryGetValue(BuildTarget.StandaloneWindows, out CompilationPlatformInfo defaultPlatform)
+                || platforms.TryGetValue(BuildTarget.StandaloneWindows64, out defaultPlatform)
+                || platforms.TryGetValue(BuildTarget.WSAPlayer, out defaultPlatform)
+                || platforms.TryGetValue(BuildTarget.Android, out defaultPlatform)
+                || platforms.TryGetValue(BuildTarget.iOS, out defaultPlatform)
+                || platforms.TryGetValue(BuildTarget.NoTarget, out defaultPlatform))
```
wiwei: Should this be kept in sync with the define:

supportedBuildTargets

In MSBuildTools.cs?

Assets/MixedRealityToolkit/MSBuild/Scripts/MSBuildTools.cs
```diff
@@ -0,0 +1,199 @@
...
+            Utilities.EnsureCleanDirectory(Application.dataPath.Replace("Assets", "MRTKBuild"));
+
+            string path = Application.dataPath.Replace("Assets", "MSBuild");
+            try
+            {
+                Utilities.EnsureCleanDirectory(path);
+            }
+            catch (IOException ex)
+            {
+                if (ex.Message.Contains(@".vs\MRTK\v15\Server\sqlite3\db.lock"))
```
wiwei: >.vs\MRTK\v15\Server\sqlite3\db.lock [](start = 42, length = 35)

Thoughts on just looking for db.lock specifically, - if only because seeing a version number makes me think this logic can get broken on future releases.

Assets/MixedRealityToolkit/MSBuild/Scripts/MSBuildTools.cs
```diff
@@ -0,0 +1,199 @@
...
+                .Select(CompilationPlatformInfo.GetCompilationPlatform)
+                .ToList();
+
+            CompilationPlatformInfo editorPlatform = CompilationPlatformInfo.GetEditorPlatform();
+
+            string propsOutputFolder = Application.dataPath.Replace("Assets", "MSBuild");
+            CreateCommonPropsFile(platforms, editorPlatform, propsOutputFolder);
+            UnityProjectInfo unityProjectInfo = new UnityProjectInfo(platforms);
+
+            //// Read the solution template
```
wiwei: >//// [](start = 12, length = 4)

Tooooooooo many //'s here and below

Assets/MixedRealityToolkit/MSBuild/Scripts/PluginAssemblyInfo.cs
```diff
@@ -0,0 +1,287 @@
...
+        ///// Creates a new instance of the <see cref="PluginAssemblyInfo"/>.
+        ///// </summary>
+        ///// <param name="availablePlatforms"></param>
+        ///// <param name="guid"></param>
+        ///// <param name="assetsRelativePath"></param>
+        //public PluginAssemblyInfo(IEnumerable<CompilationPlatformInfo> availablePlatforms, Guid guid, string assetsRelativePath)
+        //    : this(availablePlatforms, guid, assetsRelativePath, Path.GetFullPath(Utilities.GetFullPathFromKnownRelative(assetsRelativePath)))
+        //{
+
+        //}
```
wiwei: Nit: Delete this commented out code

Assets/MixedRealityToolkit/MSBuild/Scripts/TargetFramework.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Returns the configured <see cref="TargetFramework"/> for the <see cref="BuildTargetGroup"/>.
+        /// </summary>
+        /// <param name="this">The <see cref="BuildTargetGroup"/> to get <see cref="TargetFramework"/> for.</param>
+        /// <returns>The <see cref="TargetFramework"/> configured for given <see cref="BuildTargetGroup"/>.</returns>
+        public static TargetFramework GetTargetFramework(this BuildTargetGroup @this)
+        {
+            if (@this == BuildTargetGroup.Unknown)
+            {
+                //TODO this may be different on older unity versions
```
wiwei: >//TODO this may be different on older unity versions [](start = 16, length = 52)

Is the plan to only support 2018 and 2019? Or is this just a general comment?

Assets/MixedRealityToolkit/MSBuild/Scripts/TargetFramework.cs
```diff
@@ -0,0 +1,76 @@
+﻿#if UNITY_EDITOR
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
```
wiwei: Nit: Move the license to be the first thing

Assets/MixedRealityToolkit/MSBuild/Scripts/UnityProjectInfo.cs
```diff
@@ -0,0 +1,214 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.MSBuild
+{
+    /// <summary>
+    /// A helper class to parse the state of the current Unity project.
+    /// </summary>
+    public class UnityProjectInfo
+    {
+        private const string SDKProjectTypeGuid = "FAE04EC0-301F-11D3-BF4B-00C04F79EFBC";
+
```
wiwei: I think this is unused

Assets/MixedRealityToolkit/MSBuild/Scripts/Utilities.cs
```diff
@@ -0,0 +1,345 @@
...
+                case BuildTarget.NoTarget:
+                    return BuildTargetGroup.Unknown;
+                default:
+                    throw new PlatformNotSupportedException($"Don't currently support {buildTarget}");
+            }
+        }
+
+        public static IEnumerable<Version> GetUWPSDKs()
+        {
+            // May also be of interest to GetReferences
```
wiwei: >GetReferences [](start = 42, length = 13)

GetReferences isn't defined in this class, and is fairly generically named - this comment makes me ask more questions than it answers, so it might be better to just delete this, or provide additional context to make it more clear

wiwei:
@keveleigh Also I think it might be worth moving this msbuild stuff into its own top level folder in Assets - I don't think we necessarily want to include this for people who are consuming this as an assert bundle (yes, we can deselect manually, but this is independent enough that this should be its own top level thing)

Assets/MixedRealityToolkit/MSBuild/Scripts/PluginAssemblyInfo.cs
```diff
@@ -0,0 +1,287 @@
...
+        Native
+    }
+
+    /// <summary>
+    /// This is the information for the plugins in the Unity project.
+    /// </summary>
+    public class PluginAssemblyInfo : ReferenceItemInfo
+    {
+        /// <summary>
+        /// Ges the type of Plugin.
```
MenelvagorMilsom: Nit: typo
```suggestion
        /// Gets the type of Plugin.
```

Assets/MixedRealityToolkit/MSBuild/Scripts/UnityProjectInfo.cs
```diff
@@ -0,0 +1,214 @@
...
+        private Dictionary<string, PluginAssemblyInfo> ScanForPluginDLLs()
+        {
+            return Directory.GetFiles(Application.dataPath, "*.dll", SearchOption.AllDirectories)
+                .Select(t => new { FullPath = t, AssetsRelativePath = Utilities.GetAssetsRelativePathFrom(t) })
+                .Select(t => new PluginAssemblyInfo(availablePlatforms, Guid.Parse(AssetDatabase.AssetPathToGUID(t.AssetsRelativePath)), t.AssetsRelativePath, t.FullPath))
+                .ToDictionary(t => t.AssetsRelativePath);
+        }
+
+        /// <summary>
+        /// Exports the projec tinfo into a solution file, and the CSProject files.
```
MenelvagorMilsom: Nit: type
```suggestion
        /// Exports the project info into a solution file, and the CSProject files.
```

Assets/MixedRealityToolkit/MSBuild/Scripts/Utilities.cs
```diff
@@ -0,0 +1,345 @@
...
+        public static string GetRelativePath(string thisAbsolute, string thatAbsolute)
+        {
+            thisAbsolute = Path.GetFullPath(thisAbsolute);
+            thatAbsolute = Path.GetFullPath(thatAbsolute);
+
+            return new Uri(thisAbsolute).MakeRelativeUri(new Uri(thatAbsolute)).OriginalString;
+        }
+
+        /// <summary>
+        /// Gets a relative path between two knwon relative paths (inside Assets or Pacakges)
```
MenelvagorMilsom: Nit: type
```suggestion
        /// Gets a relative path between two known relative paths (inside Assets or Packages)
```

Assets/MixedRealityToolkit/MSBuild/Scripts/MSBuildTools.cs
```diff
@@ -0,0 +1,205 @@
...
+            }
+
+            List<CompilationPlatformInfo> platforms = CompilationPipeline.GetAssemblyDefinitionPlatforms()
+                .Where(t => supportedBuildTargets.Contains(t.BuildTarget))
+                .Select(CompilationPlatformInfo.GetCompilationPlatform)
+                .ToList();
+
+            CompilationPlatformInfo editorPlatform = CompilationPlatformInfo.GetEditorPlatform();
+
+            string propsOutputFolder = Application.dataPath.Replace("Assets", "MSBuild");
```
wiwei: >Application.dataPath.Replace("Assets", "MSBuild"); [](start = 39, length = 50)

This should be the same as "path" above right? I don't think path gets edited or anything

Assets/MixedRealityToolkit/MSBuild/Scripts/MSBuildTools.cs
```diff
@@ -0,0 +1,205 @@
...
+            string projectTemplateText = File.ReadAllText(Utilities.GetAssetsRelativePathFrom(TemplateFiles.Instance.SDKProjectFileTemplatePath));
+
+            unityProjectInfo.ExportSolution(solutionTemplateText, projectTemplateText, propsOutputFolder);
+
+            foreach (string otherFile in TemplateFiles.Instance.OtherFiles)
+            {
+                File.Copy(Utilities.GetFullPathFromKnownRelative(otherFile), Path.Combine(propsOutputFolder, Path.GetFileName(otherFile)));
+            }
+
+            Debug.Log("Completed.");
```
wiwei: >Completed." [](start = 23, length = 11)

Nit: "GenerateSDKProjects completed" (otherwise it's super generic in the logs)

Assets/MixedRealityToolkit/MSBuild/Scripts/PluginAssemblyInfo.cs
```diff
@@ -0,0 +1,287 @@
...
+        /// </summary>
+        public PluginType Type { get; }
+
+        /// <summary>
+        /// Gets the path relative to the assets folder.
+        /// </summary>
+        public string AssetsRelativePath { get; }
+
+        /// <summary>
+        /// Gets whether this plugin is auto referenced.
```
wiwei: >auto referenced. [](start = 40, length = 16)

What does this mean?

wiwei: Ah, is this unity "auto reference" concept?

---
In reply to: [302781511](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r302781511) [](ancestors = 302781511)

Assets/MixedRealityToolkit/MSBuild/Scripts/PluginAssemblyInfo.cs
```diff
@@ -0,0 +1,287 @@
...
+            }
+            // else fall through to use -first method 
+
+            string line;
+            while ((line = ReadUntil(reader, "first:", "userData:")).Contains("first:") && !reader.EndOfStream)
+            {
+                string[] platformLineParts = reader.ReadLine().Split(':');
+                string platform = platformLineParts[1].Trim();
+
+                if (platformLineParts[0].Contains("Facebook"))
```
wiwei: >Facebook [](start = 51, length = 8)

Wait whattttttttttt

Assets/MixedRealityToolkit/MSBuild/Scripts/TemplateFiles.cs
```diff
@@ -0,0 +1,138 @@
...
+
+        private static TemplateFiles instance;
+
+        /// <summary>
+        /// Gets the singleton instance (created on demand) of this class.
+        /// </summary>
+        public static TemplateFiles Instance => instance ?? (instance = new TemplateFiles());
+
+        private string generatedCSMetaTemplateText;
+
```
wiwei: I think dead variable (would probably be worth just going through and cleaning whatever build warnings exist, as they should also end up being build errors in CI)

Assets/MixedRealityToolkit/MSBuild/Scripts/TemplateFiles.cs
```diff
@@ -0,0 +1,138 @@
...
+            Utilities.GetPathsFromGuidsInPlace(templateFolders);
+
+            if (templateFolders.Length == 0)
+            {
+                Debug.LogError($"Templates folder '{TemplateFilesFolderName}' not found.");
+            }
+            else if (templateFolders.Length > 1)
+            {
+                Debug.LogWarning($"Stange, more than one directory exists for template files:\n {string.Join("\n", templateFolders)}");
+            }
```
wiwei: Should either of these end up throwing as well (i.e. if we can't get any template folders, is that actually super bad for this, or is this... an acceptable situation to be in?)

Assets/MixedRealityToolkit/MSBuild/Scripts/UnityProjectInfo.cs
```diff
@@ -0,0 +1,214 @@
...
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.MSBuild
+{
+    /// <summary>
+    /// A helper class to parse the state of the current Unity project.
+    /// </summary>
+    public class UnityProjectInfo
+    {
+        private const string SDKProjectTypeGuid = "FAE04EC0-301F-11D3-BF4B-00C04F79EFBC";
```
wiwei: >FAE04EC0 [](start = 51, length = 8)

Is this just an arbitrarily generated thing, or is there some secret or guidance for how we might want to rev or change it going forward?

Assets/MixedRealityToolkit/MSBuild/Scripts/UnityProjectInfo.cs
```diff
@@ -0,0 +1,214 @@
...
+                        project.AddDependency(dependency);
+                    }
+                    else
+                    {
+                        Debug.LogError($"Failed to get dependency '{reference}' for project '{project.Name}'.");
+                    }
+                }
+
+
+                // TODO Find a better way to filter which plugins should be included; project.Assembly.allReferences isn't correct
```
wiwei: What are the other ramifications of this TODO? It feels kinda scary to read this (i.e. this isn't really correct) - is it more like, it's not exactly optimal?

andreiborodin:
Yes, that's correct!

---
In reply to: [509795975](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#issuecomment-509795975) [](ancestors = 509795975,509762183)

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/SolutionTemplate.sln.meta
andreiborodin: Yep, added them

---
In reply to: [301149984](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301149984) [](ancestors = 301149984)

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/SDKProjectTemplate.csproj
andreiborodin: Yeah, project won't load, it's strange.

---
In reply to: [301778093](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301778093) [](ancestors = 301778093)

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/SDKProjectTemplate.csproj
andreiborodin: Will update.

---
In reply to: [301778494](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301778494) [](ancestors = 301778494)

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/WSA.Player.Template.props
andreiborodin: Will remove indent on start for consistency,

---
In reply to: [301780377](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301780377) [](ancestors = 301780377)

Assets/MixedRealityToolkit/MSBuild/MSBuildTemplates/WSA.Player.Template.props
andreiborodin: There is, and I am sure this won't work on Mac either. That, however, is P3 I would say, where P1 is get build working, P2 is local dev box.

Having said that, I may need to update it as we get build stuff.

---
In reply to: [301780755](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301780755) [](ancestors = 301780755)

Assets/MixedRealityToolkit/MSBuild/Scripts/CompilationPlatformInfo.cs
andreiborodin: Will add comment.

---
In reply to: [301785751](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301785751) [](ancestors = 301785751,301785122,301785013)

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
andreiborodin: Yes

---
In reply to: [301798052](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301798052) [](ancestors = 301798052)

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
andreiborodin: It doesn't, fixed :)

---
In reply to: [301800758](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301800758) [](ancestors = 301800758)

Assets/MixedRealityToolkit/MSBuild/Scripts/CSProjectInfo.cs
andreiborodin: Yeah, but no longer used :)

---
In reply to: [301802009](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301802009) [](ancestors = 301802009)

Assets/MixedRealityToolkit/MSBuild/Scripts/TargetFramework.cs
andreiborodin: Currently yes, and I don't know whether to leave this comment, or remove the "TODO" aspect of it


---
In reply to: [301803266](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301803266) [](ancestors = 301803266)

Assets/MixedRealityToolkit/MSBuild/Scripts/PluginAssemblyInfo.cs
andreiborodin: Yep, will add comment.

---
In reply to: [302781656](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r302781656) [](ancestors = 302781656,302781511)

Assets/MixedRealityToolkit/MSBuild/Scripts/TemplateFiles.cs
andreiborodin: Yeah, i don't know. I am relying on the consumer throwing when it doesn't find the template.

---
In reply to: [302782608](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r302782608) [](ancestors = 302782608)

Assets/MixedRealityToolkit/MSBuild/Scripts/UnityProjectInfo.cs
andreiborodin: Yeah, the auto-referenced.

---
In reply to: [302783219](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r302783219) [](ancestors = 302783219)

andreiborodin:
I will leave it for now, may probably move it outside of the repo in the near future.

---
In reply to: [509835489](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#issuecomment-509835489) [](ancestors = 509835489)

Assets/MixedRealityToolkit/MSBuild/Scripts/MSBuildTools.cs
andreiborodin: Good idea.

---
In reply to: [301802469](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5215#discussion_r301802469) [](ancestors = 301802469)

wiwei:
It looks like the docs build failed:

[19-07-26 12:53:07.041]Warning:[MetadataCommand.ExtractMetadata](D:/a/1/s/Assets/MixedRealityToolkit.Tools/MSBuild/MSBuildTemplates/SolutionTemplate.sln)Error opening solution D:/a/1/s/Assets/MixedRealityToolkit.Tools/MSBuild/MSBuildTemplates/SolutionTemplate.sln: Error parsing the solution configuration section in solution file. The entry "#CONFIGURATION_PLATFORM_TEMPLATE 		<Configuration>|<Platform> = <Configuration>|<Platform>" is invalid.  D:\a\1\s\Assets\MixedRealityToolkit.Tools\MSBuild\MSBuildTemplates\SolutionTemplate.sln. Ignored.
	1 Warning(s)

We might want to just save the template as SolutionTemplate.sln.tmpl or something, so that docfx doesn't attempt to parse it (and others don't try to open it up thinking it's a valid thing to open by itself)

PR #5227, opened 2019-07-08
Support focus requested property change in runtime

googlan:
## Overview

Allow `BaseInputHandler.IsFocusRequested` to be changed during play mode (and propagate a change to handler registration).

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/BaseInputHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/BaseInputHandler.cs
@@ -14,7 +14,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         [Tooltip("Is Focus required to receive input events on this GameObject?")]
         private bool isFocusRequired = true;
 
-        private bool initializationDone = false;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/BaseInputHandler.cs
```diff
@@ -14,6 +14,8 @@ public abstract class BaseInputHandler : InputSystemGlobalHandlerListener
         [Tooltip("Is Focus required to receive input events on this GameObject?")]
         private bool isFocusRequired = true;
 
+        private bool activeFocusValue = true;
```
luis-valverde-ms: For clarity, I'd call this `areHandlersRegistered`

luis-valverde-ms: Also, you should probably check this value in `OnDisable()` instead of checking `isFocusRequired`.

googlan: Good catch.

Issue #5214, opened 2019-07-05
MRTK's ReadOnlyAttribute conflicts with Unity Entities' Jobs and Hybrid Renderer packages


Ziugy:
## Describe the bug

If either the Jobs or Hybrid Renderer package are installed on a project that includes MRTK then you will receive a CS0592 warning in both packages.

```
Library\PackageCache\com.unity.rendering.hybrid@0.0.1-preview.4\Unity.Rendering.Hybrid\LODGroupSystem.cs(31,41): error CS0592: Attribute 'ReadOnly' is not valid on this declaration type. It is only valid on 'field' declarations.
```

## To reproduce

1. Start with a blank Unity project with MRTK.
2. Click on Window > Package Manager to open the package manager window.
3. In the package manager window click on the Advanced combo box and select or make sure the Show Preview Packages option is enabled.
4. Search and install either, or both, Hybrid Renderer and Jobs package(s).
5. See error in Console after package(s) finish installing.

## Expected behavior

ReadOnlyAttribute namespace should not conflict. I would assume the MRTK ReadOnlyAttribute would be wrapped in a namespace. Any reason it's not?

I would say the error also falls on Unity's side as well. They should really move their `using ReadOnlyAttribute = Unity.Collections.ReadOnlyAttribute;` inside the namespace. Best way to provide this feedback to them?

## Screenshots

N/A

## Your Setup (please complete the following information)

- Unity Version [2018.4.2f1]
- MRTK Version [RC1]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2

keveleigh:
Thanks for the heads up! Looks like the namespace was missed for that file.

The namespace was added in #5228.

PR #5228, opened 2019-07-08
Add namespace to ReadOnlyAttribute

keveleigh:
## Overview
A handful of attributes were left outside the MRTK namespace, leading to conflicts with external packages.

## Changes
- Fixes: #5214


wiwei:
We should probably have some tooling around to prevent this from happening in the future as well.

wiwei:
Please also update the release docs to mention this namespace change:

https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html

wiwei:
Still lgtm

PR #5229, opened 2019-07-08
[Shell Parity] [Slate] Slate Flattened Border Fix

Cameron-Micka:
## Overview

Fixing an issue where flattened bounding boxes would be completely flattened on an axis results in a very thin bounding box border. Bounding boxes now have a display scale (tweakable) for the flattened axis.

Capture

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4976

## Verification
Please see the slates contained in the HandInteractionExamples scene.

>
>

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
@@ -187,20 +187,20 @@ namespace Microsoft.MixedReality.Toolkit.UI
         }
 
         [SerializeField]
-        [Tooltip("When an axis is flattened what value to set that axes scale to.")]
-        private float flattenAxisScale = 0.0f;
+        [Tooltip("When an axis is flattened what value to set that axis's scale to for display.")]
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -185,6 +185,27 @@ public FlattenModeType FlattenAxis
                 }
             }
         }
+
+        [SerializeField]
+        [Tooltip("When an axis is flattened what value to set that axes scale to.")]
```
wiwei:
```suggestion
        [Tooltip("When an axis is flattened, what value to set that axis's scale to.")]
```

wiwei: Here and below.

Cameron-Micka: Nice catch, fixed!

PR #5230, opened 2019-07-08
add playmode test that toggles the profiler using speech kbd shortcuts

### LINKED ISSUE #5216: Write a test to ensure voice commands work in MRTK
## Overview
This is a follow-up task for bug #5176, which was caused by a regression in PR  #5132. 

We should write a test to ensure that changes to MRTK do not break the ability to use voice commands.

The test would do the following:

- Run a scene that has MRTK, with default voice commands.
- Trigger a voice command that is registered in the voice commands list, such as "menu" or "select" (keyboard keys '1' and '2').
- Ensure that the voice command event gets raised.

Here is an example test:

- Create a gameobject, add a component that listens for voice commands, and simulate pressing the '1' key (in editor, pressing '1' simulates a voice command 'select'). Ensure that the component received the 'select' keyword.

In case simulating keyboard buttons is not possible in Unity, here is a second idea:

- In the voice data provider, throw an exception if voice system is not enabled, then in the test load up a scene and ensure no exceptions get thrown.

davidkline-ms:
This change adds a test to toggle the diagnostics system's profiler off using the speech provider. In MRTK v2, the profiler toggle command is associated with the "9" key on the keyboard. This key stroke is used to simulate a user speaking "Toggle Profiler"

Resolves #5216

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
@@ -10,17 +10,15 @@
 // issue will likely persist for 2018, this issue is worked around by wrapping all
 // play mode tests in this check.
 
-using NUnit.Framework;
-using System.Collections;
-using UnityEngine;
-using UnityEngine.TestTools;
+using Microsoft.MixedReality.Toolkit.Diagnostics;
 using Microsoft.MixedReality.Toolkit.Input;
 using Microsoft.MixedReality.Toolkit.Utilities;
-using System.Linq;
-using Microsoft.MixedReality.Toolkit.Diagnostics;
-using System.Runtime.InteropServices;
+using NUnit.Framework;
 using System;
-using Microsoft.MixedReality.Toolkit.UI;
+using System.Collections;
+using System.Linq;
+using UnityEngine;
+using UnityEngine.TestTools;
 
 namespace Microsoft.MixedReality.Toolkit.Tests
 {
@@ -38,16 +36,21 @@ namespace Microsoft.MixedReality.Toolkit.Tests
             PlayModeTestUtilities.TearDown();
         }
 
-        /// <summary>
-        /// </summary>
-        /// <returns></returns>
         [UnityTest]
         public IEnumerator TestToggleProfilerCommand()
         {
-            // Confirm that the Diagnostics system is enabled and the VisualProfiler is in the scene.
+            int frameDelay = 10;
+
+            // Confirm that the diagnostics system is enabled.
             IMixedRealityDiagnosticsSystem diagnosticsSystem = null;
             MixedRealityServiceRegistry.TryGetService<IMixedRealityDiagnosticsSystem>(out diagnosticsSystem);
             Assert.IsNotNull(diagnosticsSystem, "The diagnostics system is not enabled in the scene.");
+            yield return null;
+
+            // This test uses the input system to simulate speech commands.
+            IMixedRealityInputSystem inputSystem = null;
+            MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem);
+            Assert.IsNotNull(inputSystem, "The input system is not enabled in the scene.");
             yield return null;
 
             // Verfiy that the VisualProfiler is enabled.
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            // Verfiy that the VisualProfiler is enabled.
+            Assert.IsTrue(diagnosticsSystem.ShowProfiler, "The VisualProfiler is not active.");
+            yield return null;
+
+            // Toggle the profiler visualization off.
+            KeyboardEvent(VirtualKey_9, ScanCode_9, 0, IntPtr.Zero); // key down
+            yield return null;
+            KeyboardEvent(VirtualKey_9, ScanCode_9, KeyboardFlag_KeyUp, IntPtr.Zero); // key up
+            for (int i = 0; i < 10; i++) { yield return null; }
```
wiwei: What's the significance of these 10 yields?

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        /// <summary>
+        /// Virutal Key Code and Scan Code values used in this test.
+        /// </summary>
+        private const Byte VirtualKey_9 = 0x39;
+        private const Byte ScanCode_9 = 0x0A;
+
+        /// <summary>
+        /// Flags, for the keybd_event function, used in this test.
```
wiwei:
```suggestion
        /// Flags, for the KeyboardEvent function, used in this test.
```

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        /// <summary>
+        /// P/Invoke signature for the Windows keybd_event function.
+        /// </summary>
+        /// <param name="keyCode">Virtual Key Code for the desired keyboard event.</param>
+        /// <param name="scanCode">Scan Code for the desired keybaord event.</param>
+        /// <param name="flags">Flags used to interpret the keyboard event.</param>
+        /// <param name="extraInfo">Additional information associated with the keyboard event.</param>
+        [DllImport("user32.dll", EntryPoint = "keybd_event")]
+        private static extern void KeyboardEvent(
```
wiwei: Because this is a windows specific binary, we might want to wrap this entire code on a platform specific check so that when people are doing other platforms (i.e. running this test on mac) it doesn't fail expectedly (i.e. user32.dll doesn't exist there)

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            PlayModeTestUtilities.TearDown();
+        }
+
+        /// <summary>
+        /// </summary>
```
wiwei: Nit: Add summary information or delete the empty doccomments

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
davidkline-ms: timing to allow the system state to be updated by the speech event handler.

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
davidkline-ms: good call

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            Assert.IsNotNull(diagnosticsSystem, "The diagnostics system is not enabled in the scene.");
+            yield return null;
+
+            // This test uses the input system to simulate speech commands.
+            IMixedRealityInputSystem inputSystem = null;
+            MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem);
+            Assert.IsNotNull(inputSystem, "The input system is not enabled in the scene.");
+            yield return null;
+
+            // Verfiy that the VisualProfiler is enabled.
```
wiwei:
```suggestion
            // Verify that the VisualProfiler is enabled.
```

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        [TearDown]
+        public void TearDown()
+        {
+            PlayModeTestUtilities.TearDown();
+        }
+
+        [UnityTest]
+        public IEnumerator TestToggleProfilerCommand()
+        {
+            int frameDelay = 10;
```
wiwei: Nit: Bring this closer to where it's used.

wiwei: A lot of coding guidelines I've worked with recommend doing this because you can keep related concepts close together, instead of having a block of a ton of variables up front that aren't used to way later in the function.

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpeechTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            // Toggle the profiler visualization off.
+            var gazeInputSource = inputSystem.DetectedInputSources.Where(x => x.SourceName.Equals("Gaze")).First();
+            inputSystem.RaiseSpeechCommandRecognized(
+                gazeInputSource, 
+                RecognitionConfidenceLevel.High, 
+                new TimeSpan(), 
+                DateTime.Now, 
+                new SpeechCommands("toggle profiler", KeyCode.Alpha9, MixedRealityInputAction.None));
+            // It may take a few frames before the event is handled and the system responds to the state change.
+            for (int i = 0; i < frameDelay; i++) { yield return null; }
```
wiwei: Does this fail every single time when there's only a single yield return null;?

I suppose in general there's not really any gaurantee that everything happens in a frame, so maybe having some general awaiting assert would be more helpful (i.e. it waits for the condition to be true and waits for up to one second).

The second part of the comment

davidkline-ms: in an earlier incarnation, it failed each time unless under the debugger (timing changes)... i felt this was a more robust way ensure a good buffer of time :)

julenka:
Shoot, it looks like this test still passes when speech is broken. I've updated #5216 with details.

PR #5233, opened 2019-07-08
Add experimental standalone service managers

davidkline-ms:
This change adds the experimental standalone service managers from the 2.x/serviceContainers feature branch into the development branch's SDK/Experimental folder.

These service managers may change in the coming releases and should not be presumed to be resistant to breaking changes. In fact, the expectation should be that these will change based in customer feedback.

Please note that these managers use the same profile system as the MixedRealityToolkit object. Inspector UI to display profiles has not yet been implemented. You can view the profile contents by double clicking the name of the profile file in the service manager inspector.


Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,391 @@
...
+
+using Microsoft.MixedReality.Toolkit.Utilities;
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit
+{
+    /// <summary>
+    /// Base class for service manager implementations.
```
thalbern: what does a base service manager do, what is it for? more doc?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,391 @@
...
+            T serviceInstance = FindService<T>();
+            if (serviceInstance != null)
+            {
+                services.Add(serviceInstance);
+            }
+            return services;
+        }
+
+        /// <inheritdoc />
+        public bool IsDataProviderRegistered<T>(string name = null) where T : IMixedRealityDataProvider
```
thalbern: would be good to have documentation for all the public methods and properties

davidkline-ms: These methods are documented as part of the IMixedRealityServiceRegistrar interface :)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: i'll write something a little more descriptive :)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Camera/Scripts/CameraSystemManager.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.CameraSystem
+{
+    /// <summary>
+    /// Stand alone camera system manager.
```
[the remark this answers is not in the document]
davidkline-ms: done :)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Input/Scripts/InputSystemManager.cs
```diff
@@ -0,0 +1,104 @@
...
+using UnityEngine.EventSystems;
+
+#if UNITY_EDITOR
+using Microsoft.MixedReality.Toolkit.Input.Editor;
+#endif
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Stand alone input system manager.
```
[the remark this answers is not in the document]
davidkline-ms: done :)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: done :)

Assets/MixedRealityToolkit.SDK/Experimental/Features/SpatialAwareness/Scripts/SpatialAwarenessSystemManager.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.SpatialAwareness
+{
+    /// <summary>
+    /// Stand alone spatial awareness system manager.
```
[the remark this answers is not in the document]
davidkline-ms: done :)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Diagnostics/Scripts/DiagnosticsSystemManager.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Diagnostics
+{
+    /// <summary>
+    /// Stand alone diagnostics system manager.
```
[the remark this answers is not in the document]
davidkline-ms: done :)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Boundary/Scripts/BoundarySystemManager.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Boundary
+{
+    /// <summary>
+    /// Service manager supporting running the boundary system, without requring the MixedRealityToolkit object.
```
wiwei: >requring [](start = 72, length = 8)

requiring

Assets/MixedRealityToolkit.SDK/Experimental/Features/Boundary/Scripts/BoundarySystemManager.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    public class BoundarySystemManager : BaseServiceManager
+    {
+        [SerializeField]
+        [Tooltip("The scale (room, world, etc) of the experience.")]
+        private ExperienceScale expereinceScale = ExperienceScale.Room;
```
wiwei: >expereinceScale [](start = 32, length = 15)

experienceScale

Assets/MixedRealityToolkit.SDK/Experimental/Features/Boundary/Scripts/BoundarySystemManager.cs
wiwei: I think this misspelling is copied into the rest of the system managers, so you may want to do a Ctrl+F across all files.

---
In reply to: [301753191](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5233#discussion_r301753191) [](ancestors = 301753191)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Boundary/Scripts/BoundarySystemManager.cs
wiwei: >ExperienceScale [](start = 16, length = 15)

Also, do you know why this is on this, and not the profile itself? I see that on the regular MRTK object it's also on the main thing, but it's really just used for the boundary stuff.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+                    serviceSnapshot[i]?.Update();
+                }
+
+                IMixedRealityDataProvider[] providers = dataProviders.ToArray();
+                for (int i = 0; i < providers.Length; i++)
+                {
+                    providers[i]?.Update();
+                }
+            }
+        }
```
wiwei: Nit: Add an extra line after this.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        {
+            if (Application.isPlaying)
+            {
+                IReadOnlyList<IMixedRealityService> serviceSnapshot = new List<IMixedRealityService>(registeredServices.Values);
+                for (int i = 0; i < serviceSnapshot.Count; i++)
+                {
+                    serviceSnapshot[i]?.Update();
+                }
+
+                IMixedRealityDataProvider[] providers = dataProviders.ToArray();
```
wiwei: It seems like this will new memory in every Update() call, which will lead to GC alloc pretty quickly - can we avoid doing this?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        #region MonoBehaviour implementation
+
+        protected virtual void Update()
+        {
+            if (Application.isPlaying)
+            {
+                IReadOnlyList<IMixedRealityService> serviceSnapshot = new List<IMixedRealityService>(registeredServices.Values);
+                for (int i = 0; i < serviceSnapshot.Count; i++)
+                {
+                    serviceSnapshot[i]?.Update();
```
wiwei: > serviceSnapshot[i]?.Update(); [](start = 19, length = 30)

Is there a good reason why we would continue to store null values in registeredServices?

It doesn't make that much sense to me that we do null checks on each thing - we should just not store null values as they aren't useful.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+                }
+
+                IMixedRealityDataProvider[] providers = dataProviders.ToArray();
+                for (int i = 0; i < providers.Length; i++)
+                {
+                    providers[i]?.Update();
```
wiwei: >providers [](start = 20, length = 9)

Same comment about null providers - we should avoid storing null providers

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+                    providers[i]?.Disable();
+                }
+
+                IReadOnlyList<IMixedRealityService> serviceSnapshot = new List<IMixedRealityService>(registeredServices.Values);
+                for (int i = 0; i < serviceSnapshot.Count; i++)
+                {
+                    serviceSnapshot[i]?.Disable();
+                }
+            }
+        }
```
wiwei: Same general comments about gc alloc and null checking

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        #region IMixedRealityServiceRegistrar implementation
+
+        /// <inheritdoc />
+        public T GetDataProvider<T>(string name = null) where T : IMixedRealityDataProvider
+        {
+            Type interfaceType = typeof(T);
+            T provider = default(T);
+
+            IMixedRealityDataProvider[] providers = dataProviders.ToArray();
+            for (int i = 0; i < providers.Length; i++)
```
wiwei: Just iterate over the list directly instead of allocating another array for this purpose (same comment everywhere else)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        }
+
+        /// <inheritdoc />
+        public IReadOnlyList<T> GetDataProviders<T>(string name = null) where T : IMixedRealityDataProvider
+        {
+            if (!string.IsNullOrWhiteSpace(name))
+            {
+                Debug.LogError("This registrar does not support requesting multiple data providers of the same interface type and name.");
+                return new List<T>(); ;
+            }
```
wiwei: Why is this one different from the defined functionality of the base [https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Interfaces/Registrars/IMixedRealityServiceRegistrar.cs)?

It seems like if we inheritdoc and implement the interface, we should honor the contract. Otherwise, we shouldn't implement the interface. Or, provide more guidance about why this one in particular doesn't adhere to the existing interface.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        }
+
+        /// <inheritdoc />
+        public T GetService<T>(string name = null, bool showLogs = true) where T : IMixedRealityService
+        {
+            T serviceInstance = FindService<T>(name);
+
+            if (showLogs && (serviceInstance == null))
+            {
+                Debug.Log("Failed to get the requested service.");
```
wiwei: >Failed to get the requested service. [](start = 27, length = 36)

It would probably be helpful to also log the requested service type to make this more debuggable. I think that lukas fixed this in another area already.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        /// <inheritdoc />
+        public IReadOnlyList<T> GetServices<T>(string name = null) where T : IMixedRealityService
+        {
+            List<T> services = new List<T>();
+
+            if (!string.IsNullOrWhiteSpace(name))
+            {
+                Debug.LogError("This registrar does not support requesting multiple services of the same interface type and name.");
+                return services;
+            }
```
wiwei: Same question on why this isn't supported (see above!)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
wiwei: Otherwise this needlessly allocates memory for real value (you're not modifying the provider list here, which I guess would be possible elsewhere)

---
In reply to: [301757810](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5233#discussion_r301757810) [](ancestors = 301757810)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        {
+            if ((dataProviderInstance == null) || (dataProviders.Contains(dataProviderInstance))) { return false; }
+
+            dataProviders.Add(dataProviderInstance);
+
+            return true;
+        }
+
+        /// <inheritdoc />
+        public bool RegisterDataProvider<T>(Type concreteType, SupportedPlatforms supportedPlatforms = (SupportedPlatforms)(-1), params object[] args) where T : IMixedRealityDataProvider
```
wiwei: >-1 [](start = 124, length = 2)

Is this supposed to be "all" or "none"?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        {
+            Type interfaceType = typeof(T);
+
+            if (registeredServices.ContainsKey(interfaceType))
+            {
+                Debug.LogError("This registrar does not support registering multiple services of the same interface type.");
+                return false;
+            }
+
+            bool registered = MixedRealityServiceRegistry.AddService<T>(serviceInstance, this);
```
wiwei: >bool registered = MixedRealityServiceRegistry.AddService<T>(serviceInstance, this); [](start = 12, length = 83)

It's interesting here that this delegates registration up, but queries later still are purely local - it's not really a traditional service registration chain, but rather this is just doing some extra work here to make it so that the registrar always knows about all services.

Can you help elaborate a bit on how the service related APIs are used on this level, vs. the top level? i.e. in what scenarios are people going to be calling into the camera monobehaviour directly and asking it for the camera service, vs just asking the MRTK registrar? It feels like there's some extra complexity here that I fully understand and would love to understand a little bit more.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        /// Activates an instance of the specified concrete type using the provided argument collection.
+        /// </summary>
+        /// <typeparam name="T">The interface which must be implemented by the concrete type.</typeparam>
+        /// <param name="concreteType">The type of object to be instantiated.</param>
+        /// <param name="supportedPlatforms">The platform(s) on which the concrete type is supported.</param>
+        /// <param name="args">Collection of arguments to provide to the concrete type's constructor.</param>
+        /// <returns>An instance of the concrete type. Returns a default value of T (typically null) in the event of a failure.</returns>
+        private T ActivateInstance<T>(Type concreteType, SupportedPlatforms supportedPlatforms = (SupportedPlatforms)(-1), params object[] args) where T : IMixedRealityService
+        {
+            if (concreteType == null) { return default(T); }
```
wiwei: >default(T) [](start = 47, length = 10)

Does this have to be default(T) because the compiler thinks we might be using structs or something else here? Otherwise it would be great to just be explicit and be like, we return null in this case.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
wiwei: I mean, I see that Initialize uses this, but it seems like this is more of an internal implementation detail than a useful public API.

---
In reply to: [301762816](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5233#discussion_r301762816) [](ancestors = 301762816)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        /// </summary>
+        /// <typeparam name="T">The interface type for the service to be intialized.</typeparam>
+        /// <param name="concreteType">The concrete type of the service to initialize.</param>
+        /// <param name="supportedPlatforms">The platoform(s) on which the service is supported.</param>
+        /// <param name="args">Arguments to provide to the service class constructor.</param>
+        protected virtual void Initialize<T>(Type concreteType, SupportedPlatforms supportedPlatforms = (SupportedPlatforms)(-1), params object[] args) where T : IMixedRealityService
+        {
+            RegisterService<T>(concreteType, supportedPlatforms, args);
+            T serviceInstance = FindService<T>();
+            serviceInstance?.Initialize();
```
wiwei: Should we look at the returned value of RegisterService? No need to FindService if it didn't succeed.

Also, should we log in the case that RegisterService fails?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
```diff
@@ -0,0 +1,395 @@
...
+        /// <typeparam name="T">The interface type of the service to locate.</typeparam>
+        /// <param name="name">The name of the desired service.</param>
+        /// <returns>Instance of the interface type, or null if not found.</returns>
+        private T FindService<T>(string name = null) where T : IMixedRealityService
+        {
+            Type interfaceType = typeof(T);
+
+            if (!registeredServices.ContainsKey(interfaceType)) { return default(T); }
+
+            return (T)registeredServices[interfaceType];
```
wiwei: Use [https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.8) instead of doing two lookups.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Input/Scripts/InputSystemManager.cs
```diff
@@ -0,0 +1,104 @@
...
+            base.OnDestroy();
+        }
+
+        /// <inheritdoc />
+        public override bool RegisterDataProvider<T>(T dataProviderInstance)
+        {
+            bool registered = base.RegisterDataProvider<T>(dataProviderInstance);
+            if (registered)
+            {
+                dataProviderInstance.Initialize();
```
wiwei: >dataProviderInstance.Initialize(); [](start = 16, length = 34)

Do you have any guidance on why the input system has to call this as part of register data provider, but this isn't part of the base register data provider implementation?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Input/Scripts/InputSystemManager.cs
```diff
@@ -0,0 +1,104 @@
...
+            args = new object[] { this, profile };
+            Initialize<IMixedRealityFocusProvider>(profile.FocusProviderType.Type, args: args);
+
+            // The input system uses the raycast provider specified in the profile.
+            // The args for the focus provider are:
+            // * The registrar
+            // * The input system profile
+            args = new object[] { this, profile };
+            Initialize<IMixedRealityRaycastProvider>(profile.RaycastProviderType, args: args);
+
```
wiwei: It might be worth adding a comment stating that these lines are order dependent (I assume that the focus provider/raycast provider require the input system already having been registered?)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Boundary/Scripts/BoundarySystemManager.cs
davidkline-ms: yeah copy/paste typo :)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Boundary/Scripts/BoundarySystemManager.cs
davidkline-ms: It was originally intended to span multiple systems. Currently, it only impacts boundary, though it can (should) be used to determine the location of the origin (head or floor)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: I'll take a look

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: was a safely check. there shouldnt ever be nulls

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: The interface was designed to support the MRTK object's public methods. we should consider removing this one as it is of limited use.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: All. This is the same semantic used elsewhere (ex: MRTK object)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: > Does this have to be default(T) because the compiler thinks we might be using structs or something else here?

Exactly

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: FindService gives us the instance that was created by this overload of RegisterService,

Assets/MixedRealityToolkit.SDK/Experimental/Features/Input/Scripts/InputSystemManager.cs
davidkline-ms: you know, that should likely be pushed into the base class... good catch!

Assets/MixedRealityToolkit.SDK/Experimental/Features/Boundary/Scripts/BoundarySystemManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Boundary/Scripts/BoundarySystemManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Input/Scripts/InputSystemManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: while retaining the MRTK object's behavior

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: AddService is what updates the registry. If we do not add the service, application code cannot locate the service without having foreknowledge of the registrar(s) in the scene.

The local storage within the registrar implementation is a performance optimization so that it does not need to query the registry each update, enable, etc.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: fixed

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: removed the short circuit and added a full implementation

Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
davidkline-ms: removed the short circuit and added a full implementation

Issue #5027, opened 2019-06-24
MouseController stutters when used on BoundingBox


Alexees:
## Describe the bug

The 3D mouse cursor of the MouseController stutters really bad when handles of the BoundingBox are being manipulated.
This is especially noticeable in the editor. During normal operation, the rendered debug lines green and brown match up all the time. While dragging handles of the box around, they diverge pretty bad. While the green line indicates that input is being processed at a smooth rate, the brown one, along with the cursor jumps around. The faster you go, the worse it gets.

## To reproduce

1. Enable the MouseController for Windows Editor in the MRTK Input Provider list
2. Run the BoundingBoxRuntimeExample
3. Modify the boundingbox and watch the 3D mouse cursor

## Expected behavior

Smooth movement with the mouse

## Screenshots

![image](https://user-images.githubusercontent.com/32508825/60025168-5fa01480-9699-11e9-84d2-e93e500b14b5.png)

## Your Setup (please complete the following information)

- Unity Version 2018.4.2f1
- MRTK Version mrtk_development

## Target Platform (please complete the following information)

- Unity Editor

PR #5241, opened 2019-07-09
Fixing 3D Mouse Cursor BoundingBox manipulation stuttering

Alexees:
## Changes
- Fixes: #5027 

I can not completely say why this fixes the issue, but I think it has something to do with the mouse pointer Transform being positioned and rotated more than once during different steps and then used for calculations that are off at that time.


thalbern:
@Alexees can I close this one? This should be covered by your 2d mouse cursor PR?

Alexees:
> @Alexees can I close this one? This should be covered by your 2d mouse cursor PR?

Yes sure

Issue #5206, opened 2019-07-04
Normal Mouse Input


Alexees:
## Describe the problem

I've tried to use the current MouseController but even though it's fancy and all, it's not as easy to use as the regular 2D mouse cursor. It's way too large when raycasting closeobjects, it's way too small in the distance, it's movement is angular based and different from how you configured your mouse to behave on your desktop.

## Describe the solution you'd like

An additional MouseController that is just the desktop mouse.

## Additional context

The current MouseDeviceManager implementation handles hiding the real cursor for the MouseController instead of letting it do so itself. The controller should know if he has its own appearance on screen or not.

wiwei:
Closing this out, as the linked merged PR has gone in

PR #5242, opened 2019-07-09
2D Mouse Pointer

Alexees:
#Overview

I implemented the logic for a regular 2D mouse cursor that behaves like the already existing 3D mouse cursor. I felt this was needed as this makes testing a little faster and smoother than with the 3D mouse cursor.
The code structure needs some review though, cause I think the inheritance structure is not ideal and the names could be better (like MousePointer2D and MousePointer3D or something like that).

#Verification
Simply choose the DefaultMousePointer prefab over the MousePointer prefab in the InputSystem pointer profile, add the "Unity Mouse Device Manager" as an InputDataProvider and you're done. DefaultMousePointer ignores the profile though, because all it included was a mouse speed, which now comes from the OS.

## Changes
- Fixes: #5206


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
```diff
@@ -207,67 +72,44 @@ public override void OnInputChanged(InputEventData<MixedRealityPose> eventData)
...
+                UnityEditor.EditorWindow.focusedWindow.ShowNotification(new GUIContent("Press \"ESC\" to regain mouse control"));
             }
+#endif
 
-            transform.Rotate(mouseDeltaRotation, Space.World);
+            Cursor.visible = false;
+            Cursor.lockState = CursorLockMode.Locked;
+        }
+
+        protected override void SetVisibility(bool visible)
```
thalbern: do you need this override? isn't the behavior the same as in the base class?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultMousePointer.cs
```diff
@@ -0,0 +1,222 @@
...
+            set
+            {
+                controller = value;
+                TrackingState = TrackingState.NotApplicable;
+
+                if (controller != null && gameObject != null)
+                {
+                    InputSourceParent = controller.InputSource;
+                    Handedness = controller.ControllerHandedness;
+                    gameObject.name = "Spatial Mouse Pointer";
```
thalbern: this should probably be Screen Space Mouse Pointer in case of the 2d one?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultMousePointer.cs
```diff
@@ -0,0 +1,222 @@
...
+        private bool isInteractionEnabled = false;
+
+        private bool cursorWasDisabledOnDown = false;
+
+        protected bool isDisabled = true;
+
+        #region IMixedRealityMousePointer Implementation
+
+        [SerializeField]
+        [Tooltip("Should the mouse cursor be hidden when no active input is received?")]
```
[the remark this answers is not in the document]
Alexees: Added them

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/MousePointer.cs
Alexees: was partially different, now it's more clear

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/DefaultMousePointer.cs
Alexees: each subclass now defines the name via overriden property

Alexees:
> Can we have a BaseMousePointer that has the base behavior and a ScreenSpaceMousePointer deriving from that, which will implement the PreSceneQuery for the 2d pointer (afais that's the only method specific to the 2d version) and another SpatialMousePointer that derives from BaseMouseCursor that has all the 3d mouse cursor specific implementations?
> 
> so it would be: ScreenSpaceMousePointer and SpatialMousePointer, both deriving from BaseMousePointer.
> 
> I don't like having this very specific 2d behavior in that base class - would be good to have that in a 2d specific class :)

I created the BaseMousePointer class and derived both versions from it. I renamed everything Mouse related to what it belongs to with the new names

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseMousePointer.cs
```diff
@@ -26,23 +25,29 @@ public class DefaultMousePointer : BaseControllerPointer, IMixedRealityMousePoin
...
         public bool HideCursorWhenInactive => hideCursorWhenInactive;
 
         [SerializeField]
         [Range(0.01f, 1f)]
         [Tooltip("What is the movement threshold to reach before un-hiding mouse cursor?")]
         private float movementThresholdToUnHide = 0.1f;
 
-        /// <inheritdoc />
+        /// <summary>
+        /// What is the movement threshold to reach before un-hiding mouse cursor?
```
thalbern: nit: it's actually a bit weird that we put the comments in question format - this should be a description: "Movement threshold to reach before un-hiding the mouse cursor"

same applies for the other comments and tooltips

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/ScreenSpaceMousePointer.cs
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using Microsoft.MixedReality.Toolkit.Utilities;
+using UInput = UnityEngine.Input;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Screen Space Mouse Pointer Implementation.
```
thalbern: add description what the class does (eg keeps mouse pointer within screen boundaries)

Alexees:
> LGTM - As we're renaming the Mouse Pointer we should add this to the breaking changes list : https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html - can you update that article with your review please?

Would love to, but I can not find a Breaking Changes list, only a documentation about what is considered one.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseMousePointer.cs
Alexees: speaking of this portion of the code. What about removing hideCursorWhenInactive and instead check for hideTimeout > 0?

thalbern:
> > LGTM - As we're renaming the Mouse Pointer we should add this to the breaking changes list : https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html - can you update that article with your review please?
> 
> Would love to, but I can not find a Breaking Changes list, only a documentation about what is considered one.

It's here: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Documentation/UpdatingToGA.md :)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseMousePointer.cs
thalbern: if we do that we loose the option to hide the cursor immediately after movement stops. you'd have to set some value in there to trigger the hide. I wouldn't change the behavior for now, also this would mean another breaking change

thalbern:
> > > LGTM - As we're renaming the Mouse Pointer we should add this to the breaking changes list : https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html - can you update that article with your review please?
> > 
> > 
> > Would love to, but I can not find a Breaking Changes list, only a documentation about what is considered one.
> 
> It's here: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Documentation/UpdatingToGA.md :)

just thinking, if you would leave the spatial mouse pointer just as mouse pointer for now there wouldn't be a need of doing a breaking change

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseMousePointer.cs
Alexees: Did not thing anyone would want that, but true. reverted it

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/ScreenSpaceMousePointer.cs
Alexees: done

Alexees:
> > > > LGTM - As we're renaming the Mouse Pointer we should add this to the breaking changes list : https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html - can you update that article with your review please?
> > > 
> > > 
> > > Would love to, but I can not find a Breaking Changes list, only a documentation about what is considered one.
> > 
> > 
> > It's here: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Documentation/UpdatingToGA.md :)
> 
> just thinking, if you would leave the spatial mouse pointer just as mouse pointer for now there wouldn't be a need of doing a breaking change

that's right, I reverted the files.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseMousePointer.cs
thalbern: yeah you're right it's unlikely someone wants it, but i wouldn't remove it for now :) thanks!

thalbern:
> > > > > LGTM - As we're renaming the Mouse Pointer we should add this to the breaking changes list : https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html - can you update that article with your review please?
> > > > 
> > > > 
> > > > Would love to, but I can not find a Breaking Changes list, only a documentation about what is considered one.
> > > 
> > > 
> > > It's here: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Documentation/UpdatingToGA.md :)
> > 
> > 
> > just thinking, if you would leave the spatial mouse pointer just as mouse pointer for now there wouldn't be a need of doing a breaking change
> 
> that's right, I reverted the files.



> > > > > LGTM - As we're renaming the Mouse Pointer we should add this to the breaking changes list : https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html - can you update that article with your review please?
> > > > 
> > > > 
> > > > Would love to, but I can not find a Breaking Changes list, only a documentation about what is considered one.
> > > 
> > > 
> > > It's here: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Documentation/UpdatingToGA.md :)
> > 
> > 
> > just thinking, if you would leave the spatial mouse pointer just as mouse pointer for now there wouldn't be a need of doing a breaking change
> 
> that's right, I reverted the files.

sry i think i wasn't clear - i meant to discuss if we should rename the spatialmousepointer back to mousepointer so we won't have a breaking change bc of that rename. what do you think?

btw thanks a lot for fixing all those things!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/ScreenSpaceMousePointer.cs
thalbern: i can't find the description. wondering if you didn't push the last commit or if github is just not showing me the change?

also if you can add one or two sentences to the spatialmousepointer class description , just what it does (mouse pointer moves around camera in 3d space , not bound to screenspace)

Alexees:
> sry i think i wasn't clear - i meant to discuss if we should rename the spatialmousepointer back to mousepointer so we won't have a breaking change bc of that rename. what do you think?
> 
> btw thanks a lot for fixing all those things!

It would make more sense. But you said it would be a breaking change, which I would like to understand. Prefabs do not loose preferences, scripts are not derived from...
If it is, and thinking nobody is using the 3D mouse, it's not worth even talking about it :P
if it is not... I'd do the rename again

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/ScreenSpaceMousePointer.cs
Alexees: Ok, I added something in. I just did not think that I should write any more than there was before I even touched the class.

thalbern:
> > sry i think i wasn't clear - i meant to discuss if we should rename the spatialmousepointer back to mousepointer so we won't have a breaking change bc of that rename. what do you think?
> > btw thanks a lot for fixing all those things!
> 
> It would make more sense. But you said it would be a breaking change, which I would like to understand. Prefabs do not loose preferences, scripts are not derived from...
> If it is, and thinking nobody is using the 3D mouse, it's not worth even talking about it :P
> if it is not... I'd do the rename again

I just thought if there are some ppl out there that are using the mouse pointer and have them configured in their profile they would loose their existing settings by the renaming from MousePointer to SpatialMousePointer - that was my point. (and sry about the back and forth renaming, I'm just trying to be careful with unnecessary breaking changes).

Alexees:
> I just thought if there are some ppl out there that are using the mouse pointer and have them configured in their profile they would loose their existing settings by the renaming from MousePointer to SpatialMousePointer - that was my point. (and sry about the back and forth renaming, I'm just trying to be careful with unnecessary breaking changes).

I did that and profiles listed the prefab under the new name. Serialization does not use names.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/ScreenSpaceMousePointer.cs
Alexees: Somehow I screwed up. I have the same repository twice with just the s in upper and lower case. But I can't open both at the same time for a merge, so I have to see what I can do about that.

Alexees: Managed to merge. Now the description is there

thalbern:
> > I just thought if there are some ppl out there that are using the mouse pointer and have them configured in their profile they would loose their existing settings by the renaming from MousePointer to SpatialMousePointer - that was my point. (and sry about the back and forth renaming, I'm just trying to be careful with unnecessary breaking changes).
> 
> I did that and profiles listed the prefab under the new name. Serialization does not use names.

ha! i didn't know that would work - sry for the confusion. :)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/ScreenSpaceMousePointer.cs
thalbern: I know the documentation in most classes is not great - but we're trying to improve that. So whenever a new feature is added or an existing feature gets fixed / refactored we should update / improve documentation as well :)

Issue #4559, opened 2019-05-24
Poke pointer position does not match shell


julenka:
# Overview
Placement of near interaction point should follow the same position as Shell fingertip cursor: (fingertip joint pos) + (tip radius) * (dir from knuckle to fingertip).

The fingertip “pose” was modified in both WindowsMixedRealityArticulatedHand and SimulatedArticulatedHand to return this pose, refactored to a single utility function.

PR #5244, opened 2019-07-09
Change finger cursor position to match shell

MenelvagorMilsom:
## Overview
Update finger cursor position calculations to match shell.

## Changes
- Fixes: #4559.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/FingerCursor.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/FingerCursor.cs
@@ -66,6 +66,12 @@ namespace Microsoft.MixedReality.Toolkit.Input
                     indexFingerRotation = transform.rotation;
                 }
 
+                Vector3 indexKnucklePosition;
+                if (!TryGetJoint(TrackedHandJoint.IndexKnuckle, out indexKnucklePosition, out _)) // knuckle rotation not used
+                {
+                    indexKnucklePosition = transform.position;
+                }
+
                 Vector3 thumbPosition;
                 Quaternion thumbRotation;
                 if (!TryGetJoint(TrackedHandJoint.ThumbTip, out thumbPosition, out thumbRotation))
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/FingerCursor.cs
```diff
@@ -118,14 +124,21 @@ protected override void UpdateCursorTransform()
 
                     if (indexFingerRingRenderer != null)
                     {
-                        TranslateToFinger(indexFingerRingRenderer.transform, deltaTime, indexFingerPosition, indexFingerRotation, nearGrabbable);
+                        if (nearGrabbable)
```
MenelvagorMilsom: Does anyone know if this unique behaviour for grab is desired? Also, is the thumb cursor needed?

julenka: @Cameron-Micka  I believe this was for an experiment, and can be removed (no need for thumb cursor)

MenelvagorMilsom:
## Before
pokepointerbefore

## After
pokepointerafter

MenelvagorMilsom:
The difference is very subtle. It is most pronounced when pinching:
## Before
The pointer is below the fingertip joint, following from where the finger is pointing.
image
## After
The pointer is further up than before. Here, the cursor centre, the fingertip and the knuckle are colinear.
image

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/FingerCursor.cs
Cameron-Micka: Yes, this was for a experiment if we wanted a cursor on the thumb when grabbing. I noticed the shell still has (dead) code paths for this. So, maybe it will come back at some point?

Cameron-Micka:
Looks great! As mentioned above the thumb cursor was an experiment that the shell ended up not doing either (but still has a code path for). Deleting the thumb cursor logic if desired should be safe since nothing is using it at the moment.

MenelvagorMilsom:
@Cameron-Micka do you know if the unique behaviour for grab should be removed as well?

Cameron-Micka:
@MenelvagorMilsom I think it would be safe to remove that as well. But, I would recommend checking out the latest Shell to make sure grabs don't have an specific cursor logic.

julenka:
> @MenelvagorMilsom I think it would be safe to remove that as well. But, I would recommend checking out the latest Shell to make sure grabs don't have an specific cursor logic.

I can confirm that as of now there's no specific grab cursor logic.

Issue #4295, opened 2019-05-11
Rotations marked as 'maintain rotation to user" rotate as I roll my head.


chbecker-ms:
(Ported to GitHub, reference 20941268)

chbecker-ms:
Comment from Luis:

A bit more info here, this seems to refer to the One Hand Rotation Modes in the Manipulation Handler script.

luis-valverde-ms:
This sounds like a job for @julenka . Feel free to re-assign otherwise.

thalbern:
according to documentation this is the expected behavior https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/README_ManipulationHandler.html

julenka:
While it's true the docs say this, the end experience feels odd, don't you think? Perhaps we should update the documentation as part of this fix. My thinking is that 'maintain rotation to user' should really be 'maintain rotation to user about the Y axis'. In other words, the object should not rotate about the X or Z axes if you rotate your head about the X or Z axis.

thalbern:
I'll try how it feels with Y only

thalbern:
I think it would be fine to have pitch and yaw  (x / y) , just roll feels very awkward.

PR #5245, opened 2019-07-09
removed roll from "maintain rotation to user" one hand manipulation mode

thalbern:
manipulation handler one hand manipulation set to maintain rotation to user will now not take roll into account anymore.

## Changes
- Fixes: #4295 

## Verification
tested with WMR headset


wiwei:
Change looks good - if there's a way to add a simple test for this (i.e. play mode test to ManipulationHandlerTests.cs) that would also be great to get coverage for this behavior.

thalbern:
> Change looks good - if there's a way to add a simple test for this (i.e. play mode test to ManipulationHandlerTests.cs) that would also be great to get coverage for this behavior.

done!

Issue #5183, opened 2019-07-03
[HL1] Cursor should follow the movement for manipulation 


cre8ivepark:
## Describe the bug
For GGV input in HoloLens 1, currently, MRTK's pointer/cursor does not follow on tap-and-drag gesture movement (grab point for manipulation/navigation and BB handle) which is confusing.

### Current MRTK behavior:
![MRTK_ContextualCursor0702b](https://user-images.githubusercontent.com/13754172/60564894-4856cc00-9d16-11e9-9145-d0bb85d785c3.gif)
![MRTK_ContextualCursor0702c](https://user-images.githubusercontent.com/13754172/60564896-4856cc00-9d16-11e9-92e4-9d9971450fb2.gif)

### Shell behavior:
Cursor follows the grab point 
![MRTK_ContextualCursorShell](https://user-images.githubusercontent.com/13754172/60565033-c5824100-9d16-11e9-9553-b7fe658f3e21.gif)
movement 

## Target Platform (please complete the following information)
- HoloLens

julenka:
Would be nice to have this for GA but don't think it's required. However, it is a shell parity issue.

MenelvagorMilsom:
It appears as if the reason this behaviour is no longer working for GGV is because `InternalGazePointer.RaisePointerDown` and `InternalGazePointer.RaisePointerUp` never appear to be called. The reason for this can be seen in `GazeProvider.OnInputDown`:
```
public void OnInputDown(InputEventData eventData)
{
    for (int i = 0; i < eventData.InputSource.Pointers.Length; i++)
    {
        if (eventData.InputSource.Pointers[i].PointerId == GazePointer.PointerId)
        {
            gazePointer.RaisePointerDown(eventData.MixedRealityInputAction, eventData.Handedness, eventData.InputSource);
            return;
        }
    }
}
```
Because the hand input sources do not have the `InternalGazePointer` in their pointer list, this condition does not pass when pinching.

As a result, `IsFocusLocked` is never being set to true, and the cursors position is never being updated as the focus object moves. 

I've hacked a proof of concept fix for this using reflection, and it seems to work well. I'll see if there's a proper way to ensure that these functions are being called.

PR #5247, opened 2019-07-09
Cursor track manipulation

MenelvagorMilsom:
## Overview

- Allows the addition of the internal gaze pointer into the pointer list of some input sources
- Sets the default for `setCursorInvisibleWhenFocusLocked` to false
- Sets `setCursorInvisibleWhenFocusLocked` to false for all example scenes

I'm not certain this hasn't broken something, I'm still testing and investigating. I just wanted to open this PR and gather some feedback.

## Changes
- Fixes: #5183.


MenelvagorMilsom:
There's definitely some weird behaviour with these changes on HL1. 

- Moving an object towards/away from you scales the object.
- Object sometimes stuck in manipulate mode after hand release.
- Blooming while manipulating breaks cursor state.

I'll continue to investigate this tomorrow.

wiwei:
Took a quick peek, though based on the last comment left by @MenelvagorMilsom, I'll probably hold off until another update. The code change itself seems fine, I think that what I'm lacking a bit is the "how and why" parts of the description (i.e. the PR description is great at saying what is being changed, but is missing how the problem exists and how this fixes it).

I'm assuming that currently the gaze pointer is only added to the speech/dictation (or otherwise non-target sources) because they still need some sort of target in order to act on, so I'm curious how this change affects the philosophy of what the gaze pointer is for (i.e. should it also be used for things that already have pointers that target things?)

Should this have just been an additional config on the base MRTK profile (i.e. add the gaze pointer or something to these input types?)

That sorta info

MenelvagorMilsom:
> Took a quick peek, though based on the last comment left by @MenelvagorMilsom, I'll probably hold off until another update. The code change itself seems fine, I think that what I'm lacking a bit is the "how and why" parts of the description (i.e. the PR description is great at saying what is being changed, but is missing how the problem exists and how this fixes it).

There's a better description on [here](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5183#issuecomment-508794041). If you think it would be useful I can add this to description to this PR or to the commit messages.

MenelvagorMilsom:
> I'm assuming that currently the gaze pointer is only added to the speech/dictation (or otherwise non-target sources) because they still need some sort of target in order to act on, so I'm curious how this change affects the philosophy of what the gaze pointer is for (i.e. should it also be used for things that already have pointers that target things?)

For HL1 interaction, hands cannot target, and so I see no need to change our philosophy of what the gaze pointer is for. The idea was to make HL1 hands work in a similar way to dictation/speech.

However there is room to make this change more targeted for HL1, i.e. this pointer should not be added to hands on HL2. That way the change would be more consistent with our previous conception of the gaze pointer.

MenelvagorMilsom:
> Should this have just been an additional config on the base MRTK profile (i.e. add the gaze pointer or something to these input types?)

This would mean exposing the InternalGazePointer, which I'm not certain is something we want to do.

julenka:
> However there is room to make this change more targeted for HL1, i.e. this pointer should not be added to hands on HL2. That way the change would be more consistent with our previous conception of the gaze pointer.

Are you saying that with this change we wouldn't be able to use GGV style interaction on HL2? Many people still want the option to use GGV HoloLens 1 style interactions, even on HoloLens 2. Don't think we want to go down that route.

julenka:
why are we setting setcursortoinvisiblewhenfocuslocked to false from true?

julenka:
> Took a quick peek, though based on the last comment left by @MenelvagorMilsom, I'll probably hold off until another update. The code change itself seems fine, I think that what I'm lacking a bit is the "how and why" parts of the description (i.e. the PR description is great at saying what is being changed, but is missing how the problem exists and how this fixes it).
> 
> I'm assuming that currently the gaze pointer is only added to the speech/dictation (or otherwise non-target sources) because they still need some sort of target in order to act on, so I'm curious how this change affects the philosophy of what the gaze pointer is for (i.e. should it also be used for things that already have pointers that target things?)
> 
> Should this have just been an additional config on the base MRTK profile (i.e. add the gaze pointer or something to these input types?)
> 
> That sorta info

+1, would love to get some more explanation of the why behind this change before looking too deeply.

MenelvagorMilsom:
> Are you saying that with this change we wouldn't be able to use GGV style interaction on HL2? Many people still want the option to use GGV HoloLens 1 style interactions, even on HoloLens 2. Don't think we want to go down that route.

That's a good point. I guess I more meant that we should only enable this change for GGV and not Articulated Hands. At the moment it is enabled for both.

MenelvagorMilsom:
> Would be great to get a bit more of explanation "why" behind the changes, more details about how the code changes enable "cursor track manipulation", and what "cursor track manipulation" is.

As I pointed out to Will, there's a better description on [the issue linked to this PR](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5183). I'll add some of this description into the description of this PR for clarity.

MenelvagorMilsom:
> why are we setting setcursortoinvisiblewhenfocuslocked to false from true?

These changes actually change it from true to false. I'm not sure why the default was true before. We couldn't see that it was true before because its effect is dependent on the `IsFocusLocked` value I mentioned before (see [`GazeProvider.Update`](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/ab6bfc0b97a59d2ba76b2f1695c43115aba3bb63/Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs#L372)):
```csharp
// If flagged to do so (setCursorInvisibleWhenFocusLocked) and active (IsInteractionEnabled), set the visibility to !IsFocusLocked,
// but don't touch the visibility when not active or not flagged.
if (setCursorInvisibleWhenFocusLocked && gazePointer != null &&
    gazePointer.IsInteractionEnabled && gazePointer.IsFocusLocked == GazeCursor.IsVisible)
{
    GazeCursor.SetVisibility(!gazePointer.IsFocusLocked);
}
```
Alongside these changes we need to default this to false and change the value in the sample scenes in order to see the cursor.

MenelvagorMilsom:
The newest changes here skip `InternalGazePointer.RaisePointerDown` and directly sets IsFocusLocked in the GGVPointer.

julenka:
> The newest changes here skip `InternalGazePointer.RaisePointerDown` and directly sets IsFocusLocked in the GGVPointer.

Does locking the GGVPointer actually lock the cursor to the object? I don't think so, unless you changed some code to enable this.

> I'm not certain this is the best thing to do (seems pretty hacky), let me know what you all think.
What was wrong with using OnPointerDown?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
```diff
@@ -240,6 +242,8 @@ public void OnInputDown(InputEventData eventData)
                             c.SourceDownIds.Add(eventData.SourceId);
                         }
                         InputSystem.RaisePointerDown(this, selectAction, Controller.ControllerHandedness);
+
+                        gazeProvider.GazePointer.IsFocusLocked = (gazeProvider.GazePointer.Result?.Details.Object != null);
```
julenka: Doesn't the input system already lock / unlock pointers? Or, is the problem that the GGVPointer gets locked, but not the gaze pointer?

I'd recomment adding a comment explaining we we are locking / unlocking here.

MenelvagorMilsom:
> Does locking the GGVPointer actually lock the cursor to the object? I don't think so, unless you changed some code to enable this.

Sorry, I should revise that to:

The newest changes here skip `InternalGazePointer.RaisePointerDown` and directly sets `InternalGazePointer.IsFocusLocked` in the GGVPointer.

MenelvagorMilsom:
> It seems reasonable to me, and I'm guessing the other approaches didn't work.

Luis advised against adding ways to distinguish pointers with the ability to target and pointers without that ability. Instead he advised that for GGV, you shouldn't think of them as two pointers with distinct capabilities; the `GGVPointer` can target, but it uses the `InternalGazePointer` to do so.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
MenelvagorMilsom: > Or, is the problem that the GGVPointer gets locked, but not the gaze pointer?

Yeah, exactly this. I'll add a comment.

MenelvagorMilsom:
I just tested this on HL1 and HL2 with GGVPointer. Everything is working as expected.

PR #5248, opened 2019-07-09
Input recording gltf

lukastoenneMS:
## Overview

Add export and import of input animation data to and from glTF files.
This supplements (and should ultimately replace) the custom binary format, and provides a standardized way to exchange recorded input data with other tools.

A new set of utility classes has been added to serialize `InputAnimation` curves into a binary buffer suitable for glTF accessors. The existing GltfObject in MRTK is constructed with simple nodes that represent the camera and hand controllers. A matching importer finds matching nodes in any glTF file to try and interpret the animation data as MRTK input again.

The Unity JsonUtility class is not flexible enough to handle details of the glTF schema, such as minimum array length, or whether enums should be exported by name or by int value. For this reason an alternative JSON builder and parser pair has been added. These can handle JsonAttributes, which specify schema details as part of the C# reflection, so that the GltfObject can still be exported as "structured JSON".

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
+++ b/Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
@@ -1,15 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using Microsoft.MixedReality.Toolkit;
 using Microsoft.MixedReality.Toolkit.Utilities.Editor;
 using UnityEngine;
-using UnityEngine.Playables;
-using UnityEngine.SceneManagement;
 using UnityEditor;
 using System;
 using System.IO;
-using System.Text.RegularExpressions;
 
 namespace Microsoft.MixedReality.Toolkit.Input
 {
@@ -190,6 +186,13 @@ namespace Microsoft.MixedReality.Toolkit.Input
             DrawAnimationInfo();
         }
 
+        static readonly string[] FileTypeFilters = new string[]
+            {
+                "glTF Binary File", InputAnimationGltfUtils.Extension,
+                "Input Animation Binary File", InputAnimationBinaryUtils.Extension,
+                "All files", "*"
+            };
+
         private void DrawPlaybackGUI()
         {
             if (!Application.isPlaying)
@@ -204,12 +207,12 @@ namespace Microsoft.MixedReality.Toolkit.Input
             {
                 if (GUILayout.Button("Load ..."))
                 {
-                    string filepath = EditorUtility.OpenFilePanel(
+                    string filepath = EditorUtility.OpenFilePanelWithFilters(
                         "Select input animation file",
                         "",
-                        InputAnimationSerializationUtils.Extension);
+                        FileTypeFilters);
 
-                    LoadAnimation(filepath);
+                    TryLoadAnimation(filepath);
                 }
             }
 
@@ -295,23 +298,12 @@ namespace Microsoft.MixedReality.Toolkit.Input
         private void SaveAnimation(bool loadAfterExport)
         {
             string outputPath;
-            if (loadedFilePath.Length > 0)
-            {
-                string loadedDirectory = Path.GetDirectoryName(loadedFilePath);
-                outputPath = EditorUtility.SaveFilePanel(
-                    "Select output path",
-                    loadedDirectory,
-                    InputAnimationSerializationUtils.GetOutputFilename(),
-                    InputAnimationSerializationUtils.Extension);
-            }
-            else
-            {
-                outputPath = EditorUtility.SaveFilePanelInProject(
-                    "Select output path",
-                    InputAnimationSerializationUtils.GetOutputFilename(),
-                    InputAnimationSerializationUtils.Extension,
-                    "Enter filename for exporting input animation");
-            }
+            string initialDirectory = (loadedFilePath.Length > 0 ? Path.GetDirectoryName(loadedFilePath) : Application.dataPath);
+            outputPath = EditorUtility.SaveFilePanel(
+                "Select output path",
+                initialDirectory,
+                InputAnimationGltfUtils.GetOutputFilename(),
+                InputAnimationGltfUtils.Extension);
 
             if (outputPath.Length > 0)
             {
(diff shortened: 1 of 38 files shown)


lukastoenneMS:
I'm going to try and unify the glTF and the binary format, such that the .bin files can be used as-is as a binary buffer for glTF animations. That way we can either export/import just the .bin binary data, or a .glTF text file with uri's referencing those .bin files, or a .glb file with the binary data integrated as a buffer.

Documentation/InputSimulation/InputAnimationGltfStructure.md
```diff
@@ -0,0 +1,58 @@
...
+* "RingDistalJoint"
+* "RingTip"
+* "PinkyMetacarpal"
+* "PinkyKnuckle"
+* "PinkyMiddleJoint"
+* "PinkyDistalJoint"
+* "PinkyTip"
+* "None"
+
+The position of the nodes in the scene hierarchy does not matter. The exporter will group hand nodes under a common parent node for convenience, but any node with a matching name will be recognized by the importer.
```
thalbern: this sentence is already in the above paragraph? is it meant to be in this one as well?

lukastoenneMS: Yes, this is intentional, it applies to both the hand tracking/pinching nodes and the joint nodes.

Assets/MixedRealityToolkit.Services/InputAnimation/InputRecordingService.cs
```diff
@@ -270,7 +270,7 @@ private bool RecordInputHandData(Handedness handedness)
         /// <inheritdoc />
```
Troy-Ferrell: I tried pulling this branch and cannot seem to get a record working. In particular I get a nullreferenceexception for OnRecordingStarted & OnRecordingStopped. These should both be accessed via OnRecordingStart?.Invoke();

Also when I hit record/stop in the HandInteractionRecordArticulatedHands example scene, I don't see any file posted? 

Also the materials in the HandInteractionRecordArticulatedHands example scene use the Unity standard shaders

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -204,12 +207,12 @@ private void DrawPlaybackGUI()
             {
                 if (GUILayout.Button("Load ..."))
```
Troy-Ferrell: When I open the branch, it looks like the PlaybackService is null when I enter playmode and select the Playback tab which throws error in console 

This is in the HandInteractionRecordArticulatedHands example scene. This of course also blocks from me from trying to load a file

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -204,12 +207,12 @@ private void DrawPlaybackGUI()
...
                 {
-                    string filepath = EditorUtility.OpenFilePanel(
+                    string filepath = EditorUtility.OpenFilePanelWithFilters(
                         "Select input animation file",
                         "",
-                        InputAnimationSerializationUtils.Extension);
+                        FileTypeFilters);
 
-                    LoadAnimation(filepath);
+                    TryLoadAnimation(filepath);
```
Troy-Ferrell: Not sure how long the loading takes because of bug above. But may want to disable this button if it is truly async. The user could try to hit load again and re-select the file if it looks like nothing happened. 

Another consideration is to make the async call/task a one with progress and utilize EditorUtility.DisplayCancelableProgressBar()

Troy-Ferrell:
Excited about this tool. When I switched to the branch, I couldn't get it working though. I looked through the docs and nothing stood out to me. I could see the input recording service was added to the MRTK profile input data providers. I also couldn't get the example scene to work for me. 

Whatever the issue is, the docs should be updated to include any additional necessary steps etc to setup.

Assets/MixedRealityToolkit.Services/InputAnimation/Editor/InputRecordingWindow.cs
```diff
@@ -323,21 +315,31 @@ private void SaveAnimation(bool loadAfterExport)
...
                 if (loadAfterExport)
                 {
-                    LoadAnimation(result);
+                    TryLoadAnimation(result);
                 }
             }
         }
 
-        private void LoadAnimation(string filepath)
+        private async void TryLoadAnimation(string filepath, int attempts = 3, int waitMs = 1000)
```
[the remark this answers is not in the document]
lukastoenneMS: This is a bit awkward: when saving the file in the UnityEditor and then immediately reopening it there will be a file lock and an IOException. Some other process is accessing the file for a split second (maybe Unity for creating the meta file or something), which makes our attempt to open it fail. Could perhaps be solved by having a shared write-read handle and only releasing the writer after the reader is created, but that requires some changes to the underlying service.

This workaround is just for the editor tool.

Assets/MixedRealityToolkit.Services/InputAnimation/InputRecordingService.cs
lukastoenneMS: Probably the input service data providers got messed up again, it happens all the time because MRTK doesn't have proper versioning ...

lukastoenneMS: You should be able to fix it manually by setting the data provider types in the input system profile for the recording and playback services. I'll have a look after merging mrtk_dev.

lukastoenneMS:
I will postpone this until after GA, too much going on.

StephenHodgson:
> The Unity JsonUtility class is not flexible enough to handle details of the glTF schema, such as minimum array length, or whether enums should be exported by name or by int value. For this reason an alternative JSON builder and parser pair has been added. These can handle JsonAttributes, which specify schema details as part of the C# reflection, so that the GltfObject can still be exported as "structured JSON".

There was a very specific requirement to use the `JsonUtility` might as well replace it with the KhronosGroups implementation if you're going to go this far.

lukastoenneMS:
I don't know of anyone who'd have a concrete use for the feature at this time, and i'm working in a different area now. Wouldn't mind closing this PR. I will keep the branch around in case someone wants to pick it up (preferably using UnityGltf instead).

PR #5249, opened 2019-07-09
Fix VisualProfilerControl.ToggleProfiler implementation

davidkline-ms:
The VisualProfilerControl script's ToggleProfiler method was toggling all diagnostics, rather than just the profiler. While today there is only a single diagnostic that is controlled by ToggleDiagnostics, in the future, it is anticipated that additional displays (ex: debug panel, individual control of the profiler's settings, etc) will be added.


wiwei:
I kinda wanna ask for a test here because this should be super testable (just create a test case that creates an object with this monobehavior, and then calls this function) but also the test doesn't have THAT much value, so probably don't do it in this case.

PR #5250, opened 2019-07-09
Fix motion controller not moving with teleportation on-device

keveleigh:
## Overview
When using the platform motion controller models, they showed up in the wrong place on-device after teleportation.

This was because we transformed the poses for the explicit pointer and grip poses, but not for the source pose.

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs
@@ -131,6 +131,11 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input
 
                 IsRotationAvailable = interactionSourceState.sourcePose.TryGetRotation(out currentSourceRotation);
 
+                // We want the source to follow the Playspace, so fold in the playspace transform here to 
+                // put the source pose into world space.
+                currentSourcePosition = MixedRealityPlayspace.TransformPoint(currentSourcePosition);
+                currentSourceRotation = MixedRealityPlayspace.Rotation * currentSourceRotation;
+
                 // Devices are considered tracked if we receive position OR rotation data from the sensors.
                 TrackingState = (IsPositionAvailable || IsRotationAvailable) ? TrackingState.Tracked : TrackingState.NotTracked;
             }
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs
```diff
@@ -131,6 +131,11 @@ private void UpdateSourceData(InteractionSourceState interactionSourceState)
 
                 IsRotationAvailable = interactionSourceState.sourcePose.TryGetRotation(out currentSourceRotation);
 
+                // We want the source to follow the Playspace, so fold in the playspace transform here to 
+                // put the source pose into world space.
+                currentSourcePosition = MixedRealityPlayspace.TransformPoint(currentSourcePosition);
+                currentSourceRotation = MixedRealityPlayspace.Rotation * currentSourceRotation;
```
wiwei: Does this end up no-oping when hands are involved?

wiwei: It feels like it's fixing up a controller scenario but also is involved in a relatively fundamental hand scenario, so I'm curious if that stuff is being fixed up some other way right now

---
In reply to: [301810770](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5250#discussion_r301810770) [](ancestors = 301810770)

keveleigh: `MixedRealityPlayspace` will be at the origin, unless teleportation is used. So, yeah, in most cases on HoloLens this will no-op, unless a dev decides to use teleportation on a HoloLens.

keveleigh: These lines are currently being used on our grip data as well:
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/415582cd3256b633e862c8db8c829844bd13d603/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs#L217
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/415582cd3256b633e862c8db8c829844bd13d603/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs#L218

wiwei: Yep, I saw the other lines, I was just more concerned about modifications to some other path.

Issue #4210, opened 2019-05-07
Throwing objects is not working in ManipulationHandler, object just drops to ground


Tom-Wells-NW:
## Describe the problem

Please add **DragAndDropHandler** (or something similar) back into the MRTK v2. I think throwing and tossing interactable objects will be useful in MR games. ManipulationHandler does not seem to allow for throwing or tossing objects.

## Additional context

For and example of what I mean, see this video: https://www.youtube.com/watch?v=CcoX6g67S6o

Shizola:
@cre8ivepark Hi Yoon, I tested this the other day and can confirm this behaviour doesn't currently work. 
If you try the HandInteractionExamples scene and try to throw one of the cubes, it will just drop directly downwards. I had little look and it seems the method that gets the velocity of the controllers always returns zero.

I'm switching over from using VRTK and I would definitely appreciate more grabbing / throwing options.

Tom-Wells-NW:
Thanks for the feedback. The DragAndDropHandler also had a distance scale property (that might not be useful in non-gaming systems). Would it be possible to add this property to the ManipulationHandler? The distance scale property is a Z-axis scale movement multiplier.

julenka:
Hi @Tom-Wells-NW, yes it  is possible to add, however it would be great to move this into a separate issue / feature request., since this issue is specifically about throwing objects.

googlan:
Should be fixed with #5259 . This fixes specifically Release behavior for MR controllers and articulated hands.

PR #5259, opened 2019-07-10
Add better velocity calculation for MR controllers and simulated hand

googlan:
## Overview

Fix is triggered by #4210 , which was, as it turned out, unrelated to `ManipulationHandler`.
MR controllers were completely missing calculation of controller velocity and angular velocity on a per-frame basis.
This change adds velocity calculation to both MR articulated hands and MR controller. Velocity is calculated using several of last registered controller positions -- this doesn't increase CPU cost, but protects from jitter.  Old approach to velocity calculation was missing actual velocity change in most cases.


Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityArticulatedHand.cs
```diff
@@ -348,6 +349,56 @@ private void UpdateIndexFingerData(InteractionSourceState interactionSourceState
 #endif // WINDOWS_UWP
         }
 
+#if WINDOWS_UWP
+        // Velocity internal states
+        private float deltaTimeStart;
+        private readonly int velocityUpdateInterval = 6;
+        private int frameOn = 0;
+        private Vector3[] velocityPositionsCache = new Vector3[6];
```
wiwei: Nit: Use the same value velocityUpdateInterval instead of hardcording the number 6

Assets/MixedRealityToolkit/Providers/Hands/BaseHand.cs
```diff
@@ -54,30 +56,39 @@ public override void SetupDefaultInteractions(Handedness controllerHandedness)
...
+                Velocity = (newPositionsSum - velocityPositionsSum) / deltaTime / velocityUpdateInterval;
+
+                Quaternion rotation = Quaternion.FromToRotation(velocityNormalsSum / velocityUpdateInterval, newNormalsSum / velocityUpdateInterval);
                 Vector3 rotationRate = rotation.eulerAngles * Mathf.Deg2Rad;
                 AngularVelocity = rotationRate / deltaTime;
+
+                velocityPositionsCache[frameIndex] = newPosition;
+                velocityNormalsCache[frameIndex] = newNormal;
+                velocityPositionsSum = newPositionsSum;
+                velocityNormalsSum = newNormalsSum;
```
wiwei: This velocity calculation code looks to be decently common between a few different files - it might be worth abstracting that out into a helper class (i.e. VelocityCalculator.cs), where each frame you're responsible for calling UpdateVelocity(NewPosition, NewNormal), and then the caller can then just call GetVelocity and GetAngularVelocity.

This would also make it possible to unit test this functionality (i.e. you could give it 10 different points and normal and then show the expected numerical amount)

googlan: I wasn't sure if we would want the same algorithm for all controllers, because robustness of positions varies from controller to controller. So I chose not to abstract it yet. I can do that if we think this proposed implementation is okay for the foreseeable future.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
```diff
@@ -237,6 +238,40 @@ private void UpdateMenuData(InteractionSourceState interactionSourceState, Mixed
...
+        // Velocity internal states
+        private float deltaTimeStart;
+        private readonly int velocityUpdateInterval = 6;
+        private int frameOn = 0;
+        private Vector3[] velocityPositionsCache = new Vector3[6];
+        private Vector3 velocityPositionsSum = Vector3.zero;
+
+        protected void UpdateVelocity()
+        {
+            if (frameOn < velocityUpdateInterval)
```
keveleigh: I'm not sure if we provide this data for hands, but for motion controllers [velocity](https://docs.unity3d.com/ScriptReference/XR.WSA.Input.InteractionSourcePose.TryGetVelocity.html) and [angular velocity](https://docs.unity3d.com/ScriptReference/XR.WSA.Input.InteractionSourcePose.TryGetAngularVelocity.html) is provided by the platform. Do we want to use those values instead of calculating our own?

luis-valverde-ms: +1 one to that

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityArticulatedHand.cs
```diff
@@ -348,6 +349,56 @@ private void UpdateIndexFingerData(InteractionSourceState interactionSourceState
...
+
+                float deltaTime = Time.unscaledTime - deltaTimeStart;
+
+                Vector3 newPosition = unityJointPositions[(int)HandJointKind.Palm];
+                Vector3 newNormal = unityJointOrientations[(int)HandJointKind.Palm] * Vector3.up;
+
+                Vector3 newPositionsSum = velocityPositionsSum - velocityPositionsCache[frameIndex] + newPosition;
+                Vector3 newNormalsSum = velocityNormalsSum - velocityNormalsCache[frameIndex] + newNormal;
+
+                Velocity = (newPositionsSum - velocityPositionsSum) / deltaTime / velocityUpdateInterval;
```
lukastoenneMS: Have you considered using an exponential moving average instead of the uniform weight for each frame? It might give better results for sharply changing velocities. Not sure how noticeable that would be in practice but i think it's worth trying out and shouldn't be too hard to implement:
[https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average)

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
lukastoenneMS: +1 to using the platform-supplied velocity where available, i would assume it's more accurate.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityArticulatedHand.cs
googlan: I did consider it. It will become more accurate if we will be taking more frames into account (15+ probably). For small sequences like 6 I'm using (it's taken semi-arbitrarily, will get an average velocity over 100ms time when running at 60 fps, which seemed reasonable), perceived difference will be small. 
Also on a small measurement size, uniform weights should be less susceptible to jitters in hand positions, which we still have.

thalbern:
> +1 on what will said, if possible move the velocity calculations into a util method

Apart form the example scene lukas mentioned, is is possible to add a unit test for checking that the throw behavior doesn't break again?

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityController.cs
googlan: That's cool, I didn't know about this one. It works for both MR controllers and for hands.

Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityArticulatedHand.cs
googlan: Doesn't matter now as I'll change it to platform-supplied values :)

Assets/MixedRealityToolkit/Providers/Hands/BaseHand.cs
googlan: I changed the other 2 to platform values, so this code is only used for simulated hands now. Seems okay to leave it as is.

googlan:
Technically unit test is possible, but it will not be testing everything, as the velocity logic is implemented in controllers, so it will only test that velocity is applied correctly with simulated hands. I'll try to add one.

Issue #5238, opened 2019-07-09
Write tests that combine multiple instances and types of touchables


lukastoenneMS:
#5217 was a regression that could have been detected by testing multiple touchables in the same scene.

Test the following scenarios:
1. Multiple touchables facing the same direction. Make sure the closest one is activated.
1. Multiple touchables facing different directions (both concave and convex arrangement). Make sure normals are returned correctly and closest touchable is detected correctly.
1. Multiple _types_ of touchables (rect, volume, UnityUI) in the same scene. Make sure they work well together.

PR #5264, opened 2019-07-10
New test for NearInteractionTouchables to make sure the poke distance…

lukastoenneMS:
## Overview

New test for ensuring that touchable distance values behave as expected.

There is a lot of confusing behavior surrounding the touchables, see the <remarks> section above the test function. This behavior may be partly intentional or not, but i cannot tell because nothing is documented. I'm assuming here that the current state of the code is correct and made all the tests pass.

## Changes
- Fixes: #5238


Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -421,6 +421,150 @@ public IEnumerator NearInteractionTouchableUnityUiToggle()
...
+
+            Vector3 center = touchable.transform.position;
+
+            float margin = 0.001f;
+            Vector3 pStart = center + new Vector3(0, 0, -touchableDistance - 0.5f);
+            Vector3 pTouch = center + new Vector3(0, 0, -touchableDistance + margin);
+            // XXX the PokeThreshold ray start offset is not taken into account (see remarks above)
+            Vector3 pPoke = center + new Vector3(0, 0, -colliderThickness);
+            Vector3 pDebounce = center + new Vector3(0, 0, -colliderThickness - touchable.DebounceThreshold - margin);
+            //Vector3 pPoke = center + new Vector3(0, 0, -colliderThickness[i] - touchable.PokeThreshold + margin);
```
MenelvagorMilsom: Remove commented code

lukastoenneMS: I've left it here to show what behavior i would have expected: i thought the touch would start at `PokeDistance` - but it only starts when getting right up to the collider surface. I would like to get some confirmation (@julenka, @wiwei?) that this is the intended behavior and not a bug before removing this code.

MenelvagorMilsom: Ah I see, no worries

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -421,6 +421,150 @@ public IEnumerator NearInteractionTouchableUnityUiToggle()
...
+            UnityEngine.Object.Destroy(touchableVolume.gameObject);
+            UnityEngine.Object.Destroy(touchableUI.gameObject);
+        }
+
+        private IEnumerator TestTouchableDistances(BaseNearInteractionTouchable touchable, float colliderThickness, GameObject objectDownExpected)
+        {
+            Handedness handedness = Handedness.Right;
+            ArticulatedHandPose.GestureId gesture = ArticulatedHandPose.GestureId.Open;
+
+            yield return PlayModeTestUtilities.ShowHand(handedness, inputSim);
```
thalbern: you could use @luis-valverde-ms  hand utility class instead of this (and the other MoveHandFromTo calls)

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -421,6 +421,150 @@ public IEnumerator NearInteractionTouchableUnityUiToggle()
             UnityEngine.Object.Destroy(canvas.gameObject);
         }
 
+        /// <summary>
+        /// Test minimum touch distance.
+        /// </summary>
+        /// <remarks>
+        /// Near interaction uses two separate distance calculations:
+        /// 1. The PokePointer uses the DistanceToTouchable to determine the closest touchable. This distance value is also used compared to PokePointer.TouchableDistance to determine the set of eligible touchables in the first place.
```
thalbern: i'm a bit scared of "touching" touchable now - if there's no issue for this yet please create one on github referring to this description - we should do a refactor!

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
wiwei: @julenka would know more in this area than I would.

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -421,6 +421,150 @@ public IEnumerator NearInteractionTouchableUnityUiToggle()
...
+        ///     When the touchable is "down" and the ray length is _greater_ than PokeThreshold + DebounceThreshold (!) the TouchUp event is raised and the touchable is released.
+        /// 
+        /// The normal vector for performing the ray cast, however, is still the one returned by the first distance calculation.
+        /// 
+        /// NearInteractionTouchable also calculates distance from the object center, but the raycast ignores this and uses distance from the collider.
+        /// UnityUI OTOH does a raycast against a flat canvas, so the two distance values _should_ match (but it's not guaranteed).
+        /// 
+        /// Default PokeThreshold is larger than the TouchableDistance, so we have to reduce it in the tests below to avoid dropping the active touchable before exceeding the DebounceThreshold.
+        /// 
+        /// The ray in the second step starts at Position - PokeThreshold, but this is not taken into account when raising events. So effectively one has to move the pointer right up to the collider in order to trigger a touch, regardless of the PokeThreshold. This may be intentional or not, it's not documented at all.
```
julenka: The expected behavior is to not need to move right up to the collider unless PokeThreshold is zero. If PokeThreshold was 0.1, then the touch should start at distance 0.1 away

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
julenka: The expected behavior is to not need to move right up to the collider unless PokeThreshold is zero. If PokeThreshold was 0.1, then the touch should start at distance 0.1 away.

Let's file a new bug and try to get this fixed soon. Not a must have for GA, but perhaps an "on deck" bug feature for the next release.

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
lukastoenneMS: Shouldn't be hard to fix, just have to add the right offset in the distance calculation in PokePointer. There just wasn't any comment in the code that indicated the intended behavior.

lukastoenneMS: I have fixed the poke pointer distance calculation now and updated the tests. Seems to work fine in the test scene too.

lukastoenneMS:
You may have noticed that some other NearInteractionTouchable tests are failing right now after changing the PokeThreshold distance. The reason for this is complex and rooted in the interaction between different pointers:

1. The test is expecting exactly 1 touch event when moving towards the object, but it receives 2.
1. The 1st touch event should (ideally) not be raised, because it originates from the far-interaction pointer. However, `MixedRealityInputSystem.DispatchEventToObjectFocusedByPointer` dispatches events to all focused objects, regardless of the type of pointer doing the focusing.
1. Before i added PokeThreshold offset to the raycast distance calculation, the first event would not happen because the far interaction pointer was always disabled by the time the touch happened: The touch event would be raised only when hitting the actual collider, so there were always at least a few frames in between where the far interaction would be disabled.
1. Disabling far interaction is done by the DefaultPointerMediator when there are any `MixedRealityNearPointers` around that report `IsNearObject==true`. Before the distance changes this would happen because of the GrabPointer, which is otherwise independent from the touch interaction.
1. After fixing the PokePointer distance calculation the touch now happens much further out from the collider (0.2m by default). This is larger than the radius of the GrabPointer (which should be fine, but means that far interaction is not disabled by the GrabPointer before touch happens).
1. We would expect the PokePointer to also disable the far interaction like the GrabPointer. This was not happening, however, because the touch event was happening even _before_ PokePointer.IsNearObject became true. The reason is that IsInteractionEnabled was not overriden by PokePointer to include IsNearObject, instead it was using the base version from BaseControllerPointer. The FocusProvider would then consider the PokePointer enabled and do a raycast using the _default RayStep_ (zero origin, forward direction). As soon as this ray hits anything it would then cause a touch event through PokePointer when the ray length is below the PokeThreshold. Importantly the PokeThreshold default is actually larger than the TouchableDistance of the pointer, so the touch event would happen way before the PokePointer would consider itself IsNearObject, which would disable the far interaction.
1. **Even fixing the IsInteractionEnabled function does not solve the problem entirely.** That's because in order for the DefaultPointerMediator to do its job and disable far interaction it has to update at least once, which means there has to be at least 1 frame between the near pointer having IsNearObject (to disable far interaction) and raising the touch event.

julenka:
> The 1st touch event should (ideally) not be raised, because it originates from the far-interaction pointer. However, MixedRealityInputSystem.DispatchEventToObjectFocusedByPointer dispatches events to all focused objects, regardless of the type of pointer doing the focusing.

Are you sure about this? Reading the code, I see...

[RaisePointerDown](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/f3a97916629a965c9fc45e85290c453ebc9b45f3/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs#L931) calls [HandlePointerEvent](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/f3a97916629a965c9fc45e85290c453ebc9b45f3/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs#L379) which calls [DispatchEventToObjectFocusedByPointer](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/f3a97916629a965c9fc45e85290c453ebc9b45f3/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs#L437)

So, my understanding is the the flow would be:
* PokePointer p calls RaisePointerDown(p)
* RaisePointerDown(p) calls HandlePointerEvent(p)
* HandlePointerEvent(p) calls DispatchEventToObjectFocusedByPointer(p)
* DispatchEventToObjectFocusedByPointer(p) figures out that pointer p is focusing the touch slate and dispatches to that.

I'm not seeing how the far pointer would be dispatching the poke pointer unless the far pointer is receiving an OnInputDown event from something else

lukastoenneMS:
I just noticed that the tests seem to be passing now ... I think i had some debug instructions exit early, so the DefaultPointerMediator wasn't updating the far interaction.

In any case, regarding the double event handling: 
This is the callstack when the `TouchStarted` event is raised:

```
Microsoft.MixedReality.Toolkit.Tests.TouchEventCatcher:OnTouchStarted (HandTrackingInputEventData)+0x12 at E:\mrtk\Assets\MixedRealityToolkit.Tests\PlayModeTests\EventCatcherTestUtilities.cs:[86:13-86:42]	C#
<>c:<.cctor>b__194_46 (IMixedRealityTouchHandler, BaseEventData)+0xa at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:[1691:17-1691:48]	C#
UnityEngine.EventSystems.ExecuteEvents:Execute (GameObject, BaseEventData, EventFunction`1)+0x73 at C:\buildslave\unity\build\Extensions\guisystem\UnityEngine.UI\EventSystem\ExecuteEvents.cs:261	C#
UnityEngine.EventSystems.ExecuteEvents:ExecuteHierarchy (GameObject, BaseEventData, EventFunction`1)+0x28 at C:\buildslave\unity\build\Extensions\guisystem\UnityEngine.UI\EventSystem\ExecuteEvents.cs:286	C#
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:DispatchEventToObjectFocusedByPointer (IMixedRealityPointer, BaseInputEventData, Boolean, EventFunction`1)+0xe0 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:[589:17-589:97]	C#
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:HandleEvent (BaseEventData, EventFunction`1)+0xa5 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:[416:17-416:172]	C#
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:RaiseOnTouchStarted (IMixedRealityInputSource, IMixedRealityController, Handedness, Vector3)+0x1e at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:[1701:13-1701:81]	C#
Microsoft.MixedReality.Toolkit.Input.PokePointer:TryRaisePokeDown (GameObject, Vector3)+0x95 at E:\mrtk\Assets\MixedRealityToolkit.SDK\Features\UX\Scripts\Pointers\PokePointer.cs:[276:25-276:116]	C#
Microsoft.MixedReality.Toolkit.Input.PokePointer:OnPostSceneQuery ()+0xb0 at E:\mrtk\Assets\MixedRealityToolkit.SDK\Features\UX\Scripts\Pointers\PokePointer.cs:[222:21-222:77]	C#
Microsoft.MixedReality.Toolkit.Input.FocusProvider:UpdatePointer (PointerData)+0x1c7 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\FocusProvider.cs:[917:13-917:48]	C#
Microsoft.MixedReality.Toolkit.Input.FocusProvider:UpdatePointers ()+0x88 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\FocusProvider.cs:[812:17-812:40]	C#
Microsoft.MixedReality.Toolkit.Input.FocusProvider:Update ()+0x12 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\FocusProvider.cs:[519:13-519:30]	C#
```

`MixedRealityInputSystem:HandleEvent` loops over _all_ the pointers of the input source, the only filtering that is happening is by looking at `focusedObject` in `DispatchEventToObjectFocusedByPointer`:

```
            // If event was not handled by modal, pass it on to the current focused object
            if (focusedObject != null)
            {
                ExecuteEvents.ExecuteHierarchy(focusedObject, baseInputEventData, eventHandler);
            }
```

So if both the ShellHandRayPointer and the PokePointer have a focusedDetails.Object in the FocusProvider there is nothing stopping the event handler from being invoked twice.

It looks like it's working now, but you are entirely correct that this is super fishy ...

lukastoenneMS:
Regarding the double event handling: 
This is the callstack when the `TouchStarted` event is raised:

```
Microsoft.MixedReality.Toolkit.Tests.TouchEventCatcher:OnTouchStarted (HandTrackingInputEventData)+0x12 at E:\mrtk\Assets\MixedRealityToolkit.Tests\PlayModeTests\EventCatcherTestUtilities.cs:[86:13-86:42]	C#
<>c:<.cctor>b__194_46 (IMixedRealityTouchHandler, BaseEventData)+0xa at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:[1691:17-1691:48]	C#
UnityEngine.EventSystems.ExecuteEvents:Execute (GameObject, BaseEventData, EventFunction`1)+0x73 at C:\buildslave\unity\build\Extensions\guisystem\UnityEngine.UI\EventSystem\ExecuteEvents.cs:261	C#
UnityEngine.EventSystems.ExecuteEvents:ExecuteHierarchy (GameObject, BaseEventData, EventFunction`1)+0x28 at C:\buildslave\unity\build\Extensions\guisystem\UnityEngine.UI\EventSystem\ExecuteEvents.cs:286	C#
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:DispatchEventToObjectFocusedByPointer (IMixedRealityPointer, BaseInputEventData, Boolean, EventFunction`1)+0xe0 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:[589:17-589:97]	C#
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:HandleEvent (BaseEventData, EventFunction`1)+0xa5 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:[416:17-416:172]	C#
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:RaiseOnTouchStarted (IMixedRealityInputSource, IMixedRealityController, Handedness, Vector3)+0x1e at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\MixedRealityInputSystem.cs:[1701:13-1701:81]	C#
Microsoft.MixedReality.Toolkit.Input.PokePointer:TryRaisePokeDown (GameObject, Vector3)+0x95 at E:\mrtk\Assets\MixedRealityToolkit.SDK\Features\UX\Scripts\Pointers\PokePointer.cs:[276:25-276:116]	C#
Microsoft.MixedReality.Toolkit.Input.PokePointer:OnPostSceneQuery ()+0xb0 at E:\mrtk\Assets\MixedRealityToolkit.SDK\Features\UX\Scripts\Pointers\PokePointer.cs:[222:21-222:77]	C#
Microsoft.MixedReality.Toolkit.Input.FocusProvider:UpdatePointer (PointerData)+0x1c7 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\FocusProvider.cs:[917:13-917:48]	C#
Microsoft.MixedReality.Toolkit.Input.FocusProvider:UpdatePointers ()+0x88 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\FocusProvider.cs:[812:17-812:40]	C#
Microsoft.MixedReality.Toolkit.Input.FocusProvider:Update ()+0x12 at E:\mrtk\Assets\MixedRealityToolkit.Services\InputSystem\FocusProvider.cs:[519:13-519:30]	C#
```

`MixedRealityInputSystem:HandleEvent` loops over _all_ the pointers of the input source, the only filtering that is happening is by looking at `focusedObject` in `DispatchEventToObjectFocusedByPointer`:

```
            // If event was not handled by modal, pass it on to the current focused object
            if (focusedObject != null)
            {
                ExecuteEvents.ExecuteHierarchy(focusedObject, baseInputEventData, eventHandler);
            }
```

So if both the ShellHandRayPointer and the PokePointer have a focusedDetails.Object in the FocusProvider there is nothing stopping the event handler from being invoked twice.

julenka:
Oh! Okay, I understand now. This is happening from the **OnTouchStarted** event, not the pointer down event, is being dispatched to all pointers. I actually thing that for OnTouchStarted we should only be dispatching from the poke pointer, here's what I would propose:

* Have OnTouchStarted call HandlePointerEvent so that it only dispatches events for the current pointer. * Add comment explaining why. Something like "OnTouchDown is raised specifically from the poke pointer, and should only be dispatched to the object focused by the poke pointer, not all other pointers"

Great bug catch and thanks for helping to fix this!

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -421,6 +421,147 @@ public IEnumerator NearInteractionTouchableUnityUiToggle()
...
         }
 
+        /// <summary>
+        /// Test minimum touch distance.
+        /// </summary>
+        /// <remarks>
+        /// Near interaction uses two separate distance calculations:
+        /// 1. The PokePointer uses the DistanceToTouchable to determine the closest touchable. This distance value is also used compared to PokePointer.TouchableDistance to determine the set of eligible touchables in the first place.
+        ///     There is no explicit spec which space this distance is computed in, although it should be in world space. Due to the way touchables convert pointer position into local space first and then do a scale-only inverse transform there may be errors.
+        /// 1. After the PokePointer has selected the closest touchable, it then performs a raycast against either the collider or the UnityUI canvas, depending on type. The (world space) length of that ray is then used as the actual distance for triggering touch events.
```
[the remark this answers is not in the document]
lukastoenneMS: Oh, i just meant to say that it's not guaranteed that the distance values returned by touchables match up with actual euclidian distance. It's probably fine as long as you don't combine rotations and non-uniform local scale.

A more annoying aspect is that the touchable distance function has nothing to do with the distances actually used for touch events. As long as the collider is reasonably flat and centered it works ok, but with larger colliders and possibly object centers outside the actual collider i would expect very weird results, because what is considered the "closest" touchable is based purely on the object transform while the raycast for events may hit a completely different object. That's not easy to fix without rewriting the whole system though.

Assets/MixedRealityToolkit.Tests/PlayModeTests/NearInteractionTouchableTests.cs
```diff
@@ -421,6 +421,147 @@ public IEnumerator NearInteractionTouchableUnityUiToggle()
...
+        ///     When the ray length is _less_ than the PokeThreshold the TouchDown event is raised and the touchable is "down".
+        ///     When the touchable is "down" and the ray length is _greater_ than PokeThreshold + DebounceThreshold (!) the TouchUp event is raised and the touchable is released.
```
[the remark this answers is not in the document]
lukastoenneMS: That's ok, it just wasn't clear that the debounce threshold is actually added onto the PokeThreshold rather than used as a separate distance from the collider surface.

lukastoenneMS:
For some reason the tests for physics touchables now seem to be passing after merging in latest mrtk_development, but the UnityUI tests are broken 🤷‍♂

lukastoenneMS:
This is really fishy, but from what i'm seeing in the UnityUI tests it looks like this:

1. Before my conjunction of `PokePointer.IsInteractionEnabled` with `IsNearObject` the poke pointer would perform a raycast _all the time_, even outside the "near object" radius. This is crucial because the raycast is what triggers the UnityUI Click event when exiting the button.
1. Now with `base.IsInteractionEnabled && IsNearObject` the raycast would only happen when the object is in "near" range. This should be fine, except that the object becomes focus-locked after the first raycast. The FocusProvider will then no longer perform any raycast updates.
1. Previously the reason that UnityUI events still worked at all is because when _exiting_ the `IsNearObject` range the object would be focus-unlocked while `IsInteractionEnabled` would still be true, so the FocusProvider would start raycasting again. That would in turn cause a UnityUI raycast, which then sees the button as "exited" and trigger the Click.

lukastoenneMS:
To put it in a more high-level way:
1. UnityUI-in-MRTK relies on doing raycasts continuously, so that it can perform its own internal distance computations and detect when entering and exiting buttons.
1. MRTK's FocusProvider assumes that raycasts are only needed when the pointer does not have a focus-locked object. Once the object is focus-locked the raycast pseudo-result is just updated in `UpdateFocusLockedHit` without doing actual raycasts, thus starving UnityUI.

At least that's what i _think_ is happening. Would be great if someone involved in building this system could chime in, because i feel i'm just scratching the surface.

wiwei:
> To put it in a more high-level way:
> 
> 1. UnityUI-in-MRTK relies on doing raycasts continuously, so that it can perform its own internal distance computations and detect when entering and exiting buttons.
> 2. MRTK's FocusProvider assumes that raycasts are only needed when the pointer does not have a focus-locked object. Once the object is focus-locked the raycast pseudo-result is just updated in `UpdateFocusLockedHit` without doing actual raycasts, thus starving UnityUI.
> 
> At least that's what i _think_ is happening. Would be great if someone involved in building this system could chime in, because i feel i'm just scratching the surface.

I think that the person who worked on this last (Daniel) might have some background here, but otherwise I think you're probably the foremost expert on this topic at this point.

julenka:
> * UnityUI-in-MRTK relies on doing raycasts continuously, so that it can perform its own internal distance computations and detect when entering and exiting buttons.
> * MRTK's FocusProvider assumes that raycasts are only needed when the pointer does not have a focus-locked object. Once the object is focus-locked the raycast pseudo-result is just updated in `UpdateFocusLockedHit` without doing actual raycasts, thus starving UnityUI.

I'm fairly familiar with focus provider and input dispatch, though at this point not as much as you probably :-)

Did you try performing a UnityUI raycast regardless of the focus locked state?

lukastoenneMS:
After a few hours more of debug printing and trying to wrap my head around this, i think the issue with UnityUI events is this:
* UnityUI events are raised when the input module calls [ProcessMouseEvent](https://docs.unity3d.com/2018.4/Documentation/ScriptReference/EventSystems.StandaloneInputModule.ProcessMouseEvent.html). With the `MixedRealityInputModule` implementation this happens for all active pointers. (I'm not sure if Process() is guaranteed to be called after each pointer change or just once per frame, and what happens when multiple pointers change state if the latter is the case).
* [GetMousePointerEventData](https://docs.unity3d.com/2018.4/Documentation/ScriptReference/EventSystems.PointerInputModule.GetMousePointerEventData.html) is used to indicate clicks etc. in UnityUI. The mouse button state is set by the MRInputModule based on PointerUp/Down events (see GetMousePointerEventDataForMrtkPointer and pointerDataToUpdate[pointerId].nextPressState).
* With the previous (incorrect) PokePointer distances, the pointer up/down events would happen so far inside the pointer near interaction "bubble" that there would usually be one frame after "PointerUp" where the UI ray would still hit the object. This is enough to trigger a button click etc.
* With the new (correct) PokePointer distances the default poke threshold (0.25) is actually outside the default TouchableDistance (0.2, both are used as world-space distances). For non-UnityUI interactions this works alright, but for UI events its's necessary to have at least one frame after "mouse button released" where the ray hits the object.

lukastoenneMS:
I have now remove PokeThreshold altogether rather than trying to accommodate the distance calculation. It was breaking all kinds of established situations and making things generally more complicated by having an additional variable, so Julia and i decided to remove it instead.

Issue #5200, opened 2019-07-04
Write a test to ensure you can run HandInteractionExample scene


julenka:
## Overview
This is a follow up to #5176. To ensure we don't hit this again, let's write a test to check that the HandInteractionExample scene can run without errors. 

The test should
- Load the handinteractionexamples scene
- Run for a second or two (at least 60 frames) to ensure no errors are hit.
- Access the input system

Here is a test that could work, suggested by @wiwei

```
            SceneManager.LoadScene("MixedRealityToolkit.Examples/Demos/HandTracking/Scenes/HandInteractionExamples.unity");

             IMixedRealityInputSystem inputSystem = null;
            MixedRealityServiceRegistry.TryGetService(out inputSystem);
            Assert.NotNull(inputSystem);
            yield return null;
```

PR #5266, opened 2019-07-10
MRTK Demo Scene Loading Tests

Railboy:
## Overview
Adds a test to load and play HandInteractionExamples scene. Checks to ensure MRTK input system has been initialized.

Experimented with adding scenes to build settings prior to build via IPreprocessBuild but was unsuccessful. Instead I've added the test scene to the default build settings.

## Changes
- Fixes: #5200

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/PlayModeTests/DemoSceneTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/PlayModeTests/DemoSceneTests.cs
@@ -1,4 +1,16 @@
-﻿using System.Collections;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#if !WINDOWS_UWP
+// When the .NET scripting backend is enabled and C# projects are built
+// The assembly that this file is part of is still built for the player,
+// even though the assembly itself is marked as a test assembly (this is not
+// expected because test assemblies should not be included in player builds).
+// Because the .NET backend is deprecated in 2018 and removed in 2019 and this
+// issue will likely persist for 2018, this issue is worked around by wrapping all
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.Tests/PlayModeTests/DemoSceneTests.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System.Collections;
```
wiwei: Take a look at the other play mode tests - we end up building only in non-UWP because of weirdness with .NET. (So you'll need to add that #ifdef chunk here as well)

Assets/MixedRealityToolkit.Tests/PlayModeTests/DemoSceneTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        const string HandInteractionExamplesSceneName = "HandInteractionExamples";
+        const string HandInteractionExamplesScenePath = "MixedRealityToolkit.Examples/Demos/HandTracking/Scenes/HandInteractionExamples.unity";
+
+        const float ScenePlayDuration = 1f;
+
+        [UnityTest]
+        public IEnumerator LoadHandInteractionExamplesScene()
+        {
+            AsyncOperation loadOp = SceneManager.LoadSceneAsync(HandInteractionExamplesSceneName);
```
wiwei: It might be worth adding a comment here saying that if this has failed, the cause is probably because you have removed the HandInteractionScene from EditorBuildSettings.asset.

It's somewhat common for folks to modify this locally and sometimes even commit changes back (as we've seen due to the evolving things here)

wiwei:
Merging this in, as it has good value, I can do the comment update later.

PR #5267, opened 2019-07-10
Remove mesh specific event raising methods from spatial awareness system

davidkline-ms:
The spatial awareness system previously was exposing mesh specific methods to raise events to registered listeners. This is not a sustainable implementation in that future spatial observers will wish to raise events for other types of data (ex: planes, points).

To reduce the likelihood of future breaking changes / observer incompatibilities, this breaking change is being made,

- Observers now no longer call a Raise<data>Added, etc methods on the spatial awareness system interface
- Observers wishing to inform client code (should be all of them) will now call the spatial awareness system's HandleEvent method and provide the data and the delegate to call.
- NO changes are required of existing client code.

While working on this change, a race condition was discovered in the spatial awareness demo scene where, when using the model observer, events were not received. This has been resolved.

Tested using the spatial awareness demo scene running in editor and deployed to a HoloLens (viewing the event messages in the Visual Studio debugger)


Assets/MixedRealityToolkit.Services/SpatialAwarenessSystem/MixedRealitySpatialAwarenessSystem.cs
```diff
@@ -50,6 +54,12 @@ public bool CheckCapability(MixedRealityCapability capability)
 
         private MixedRealitySpatialAwarenessEventData<SpatialAwarenessMeshObject> meshEventData = null;
 
+        /// <inheritdoc />
+        public override void HandleEvent<T>(BaseEventData eventData, ExecuteEvents.EventFunction<T> eventHandler)
+        {
+            base.HandleEvent(eventData, eventHandler);
```
keveleigh: What's the purpose of an override that only calls the `base` function?

davidkline-ms: it helped with debugging :D 

i can remove, though we may wish to remove a similar method in other services.

davidkline-ms: removed

Assets/MixedRealityToolkit.Examples/Demos/SpatialAwareness/Scripts/DemoSpatialMeshHandler.cs
```diff
@@ -34,13 +34,12 @@ private IMixedRealitySpatialAwarenessSystem SpatialAwarenessSystem
...
-        private async void OnEnable()
+        private void Start()
         {
-            await new WaitUntil(() => SpatialAwarenessSystem != null);
-            SpatialAwarenessSystem.RegisterHandler<SpatialAwarenessHandler>(this);
+            SpatialAwarenessSystem?.RegisterHandler<SpatialAwarenessHandler>(this);
         }
 
-        private void OnDisable()
+        private void OnDestroy()
```
keveleigh: This may be more an open ended question, but does it make sense to stay registered and continue listening to / handling events even when disabled?

davidkline-ms: good question...

Issue #4825, opened 2019-06-10
ServiceInspectors assembly has dependencies on individual service binaries


davidkline-ms:
Core assemblies should not have dependencies on service, platform or feature assemblies.

The new service inspectors assembly directly references
- services.boundary
- services.input
- services,scenesystem
- services.spatialawarenss
- services.teleport

davidkline-ms:
Per offline design discussion, giving to @Railboy to address.

Thanks, sir!

julenka:
fixed.

PR #5268, opened 2019-07-10
Removes dependencies on individual service binaries from service inspectors

Railboy:
## Overview
Service inspectors now work with service interfaces instead of concrete implementations.

For most inspectors this change was trivial. For `IMixedRealitySceneService` an optional `IMixedRealitySceneServiceEditor` interface was created for editor-only functionality used by the inspector.

## Changes
- Fixes: #4825

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
@@ -168,9 +168,9 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Facades
             }
             return false;
         }
-               
+
         /// <summary>
-        /// Draws the custom inspector gui for service type.
+        /// Draws the custom inspector gui for all of the service's interfaces that have custom inspectors.
         /// </summary>
         /// <param name="facade"></param>
         /// <returns></returns>
@@ -190,7 +190,7 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Facades
         }
 
         /// <summary>
-        /// Draws the profile for the service type, if wanted by inspector and found.
+        /// Draws the profile for all of the service's interfaces that have custom inspectors, if wanted by inspector and found.
         /// </summary>
         /// <param name="serviceType"></param>
         /// <returns></returns>
@@ -342,7 +342,7 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Facades
             foreach (Type interfaceType in facade.ServiceType.GetInterfaces())
             {
                 IMixedRealityServiceInspector inspectorInstance;
-                if (GetServiceInspectorInstance(facade.ServiceType, out inspectorInstance))
+                if (GetServiceInspectorInstance(interfaceType, out inspectorInstance))
                 {
                     // If we've implemented a facade inspector, draw gizmos now
                     inspectorInstance.DrawGizmos(facade.Service);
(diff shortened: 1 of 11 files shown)


Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystemEditor.cs
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.MixedReality.Toolkit.SceneSystem;
+
+/// <summary>
+/// Optional editor-only interface for use with facade inspectors.
+/// If a scene system service does not implement this interface, the facade will not be rendered.
+/// </summary>
+public interface IMixedRealitySceneSystemEditor
```
wiwei: >IMixedRealitySceneSystemEditor [](start = 17, length = 30)

This interface is in the global namespace, please move to the non-global namespace (I think this is also probably what failed the build)

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystemEditor.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    SceneInfo[] ContentScenes { get; }
+
+    /// <summary>
+    ///  Returns the lighting scenes in the scene service profile.
+    /// </summary>
+    SceneInfo[] LightingScenes { get; }
+    
+    /// <summary>
+    /// Loads the next content scene in-editor. Use instead of LoadNextContent while not in play mode.
```
wiwei: >LoadNextContent [](start = 63, length = 15)

Where is LoadNextContent? i.e. maybe have a reference to it explicitly (I would think it would be on this interface, but it si not)

Assets/MixedRealityToolkit/Interfaces/SceneSystem/IMixedRealitySceneSystemEditor.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    /// <param name="wrap"></param>
```
wiwei: >/// <param name="wrap"></param> [](start = 4, length = 31)

Can you add a comment about what wrap does? i.e. I assume this means it will wrap around back to the beginning when already at the end, but it could also mean "wrap this scene in some other scene?" or something else non-sensical and we may as well get rid of any room for doubt

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -329,14 +339,15 @@ private static void DrawGizmos(ServiceFacade facade, GizmoType type)
             InitializeServiceInspectorLookup();
 
             // Find and draw the custom inspector
-            IMixedRealityServiceInspector inspectorInstance;
-            if (!GetServiceInspectorInstance(facade.ServiceType, out inspectorInstance))
+            foreach (Type interfaceType in facade.ServiceType.GetInterfaces())
             {
-                return;
+                IMixedRealityServiceInspector inspectorInstance;
+                if (GetServiceInspectorInstance(facade.ServiceType, out inspectorInstance))
```
wiwei: >facade.ServiceType [](start = 48, length = 18)

Should this be using interfaceType, not facade.ServiceType?

Assets/MixedRealityToolkit/Inspectors/ServiceInspectors/ServiceFacadeInspector.cs
```diff
@@ -362,15 +373,18 @@ private static void DrawSceneGUI(SceneView sceneView)
                     continue;
                 }
 
-                IMixedRealityServiceInspector inspectorInstance;
-                if (!GetServiceInspectorInstance(inspectorTypePair.Key, out inspectorInstance))
+                foreach (Type interfaceType in inspectorTypePair.Key.GetInterfaces())
                 {
```
wiwei: Similar comment here - how we loop over each interface type but we don't use it in this loop.

Railboy: Thank you, good catch.

davidkline-ms:
pulled this and it looks to be displaying the profiles correctly.

davidkline-ms:
looking into the change now

PR #5270, opened 2019-07-10
camera service manager: fix missing service type reference

davidkline-ms:
The camera service manager prefab lost the reference to the concrete service type. this change repairs the issue.


wiwei:
What did this manifest itself as? i.e. a failure to start the camera service entirely?

davidkline-ms:
> What did this manifest itself as? i.e. a failure to start the camera service entirely?

Failure to instantiate the service.

wiwei:
> > What did this manifest itself as? i.e. a failure to start the camera service entirely?
> 
> Failure to instantiate the service.

Thoughts on maybe expanding @railboy's tests here:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5266

To just make sure we can grab all of the services? It would prevent these types of super simple and also super critical regressions going forward.

davidkline-ms:
This was separate from the mrtk object. it was related to the experimentatl standalone service work. i'll write some simple tests shortly

Issue #5218, opened 2019-07-08
[Shell-Parity] Adjust fingertip proximity light size/color


cre8ivepark:
## Describe the bug
Adjust fingertip proximity light size/color to match the latest updates in the shell.

Part of Shell-Parity UX Items #4200

PR #5271, opened 2019-07-10
[Shell Parity] Proximity Light Shell Parity

Cameron-Micka:
Tweaking proximity light behavior and parameters to closer match the shell.

Below this change is the proximity light in the Shell (left) and the MRTK (right).
CaptureBefore

After this change is the proximity light in the Shell (left) and the MRTK (right).

Capture

Note, the colors were tuned as close as possible but the shell using a texture whereas the MRTK uses a 3 way color blend to allow for easy customization. The button border width was also adjusted to match the shell.  

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5218

## Verification
Try pressing buttons in the Hand Interaction Examples scene.

>
>
>
>

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/ProximityLight.cs
+++ b/Assets/MixedRealityToolkit/Utilities/ProximityLight.cs
@@ -96,7 +96,7 @@ namespace Microsoft.MixedReality.Toolkit.Utilities
             [Tooltip("The color of the ProximityLight gradient at the center (RGB) and (A) is gradient extent.")]
             [ColorUsageAttribute(true, true)]
             [SerializeField]
-            private Color centerColor = new Color(20.0f / 255.0f, 121.0f / 255.0f, 250.0f / 255.0f, 0.0f / 255.0f);
+            private Color centerColor = new Color(54.0f / 255.0f, 142.0f / 255.0f, 250.0f / 255.0f, 0.0f / 255.0f);
 
             /// <summary>
             /// The color of the ProximityLight gradient at the center (RGB) and (A) is gradient extent.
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit/Utilities/ProximityLight.cs
```diff
@@ -124,7 +124,7 @@ public Color OuterColor
             [Tooltip("The color of the ProximityLight gradient at the outer (RGB) and (A) is gradient extent.")]
             [SerializeField]
             [ColorUsageAttribute(true, true)]
-            private Color outerColor = new Color((70.0f * 5.0f) / 255.0f, (0.0f * 5.0f) / 255.0f, (191.0f * 5.0f) / 255.0f, 255.0f / 255.0f);
+            private Color outerColor = new Color((82.0f * 3.0f) / 255.0f, (31.0f * 3.0f) / 255.0f, (191.0f * 3.0f) / 255.0f, 255.0f / 255.0f);
```
Troy-Ferrell: Why times 3 instead of just putting like 246 / 255?

wiwei: There's some other weird stuff here like 255.0f/255.0f (could just be 1.0f). I guess maybe the goal was consistency with the rest of the line but it does feel like there's just noise here.

Cameron-Micka: These are great questions! I did that to match the values found in the default color picker for HDR colors. (In recent versions of Unity they have also exposed floating point color values.) But, by default an HDR color value is displayed in Unity as 3 unsigned byte values and a floating point scaler.

Issue #5273, opened 2019-07-10
link.xml generator


wiwei:
Something that we've found happen when consuming via NuGet (compiled binaries) or UPM (source) is that a lot of our components are getting optimized out in IL2CPP bytecode stripping:

https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4265

Basically, because of how we do psuedo-injection via the configuration in our profiles, and because the types contained within those profiles aren't actually directly GUID or asset references (they are instead SystemType references which basically just two strings, one of which is the type name, one of which is the assembly), when IL2CPP bytecode stripping comes into play, it optimize away some relatively critical parts of the system like:

1) Key parts of the the input system
2) Boundary/Spatial/Diagnostic (and other systems).

The exact things that get stripped away depend on the consumption mechanism and the agressiveness of the bytecode stripping setting.

In order to make these basic scenarios work, we're going to add a link.xml to the root Asset folder which will ensure that those core systems can get generated.

Note that this isn't EXACTLY optimal because certain customers might actually want some of these things to be stripped away (for example, if you don't ever plan on having the spatial system or boundary system turned on, optimizing them away is actually fine). To this, one of the things we can do is write up a Utility (in the MRTK tooling window) that will look at your current profile to generate the appropriate link.xml file that you can include to ensure a minimal set of lines in link.xml (and thus the minimal set of things that get included)

Until then, for people who want to build and don't ever plan on having the teleportation of boundary system enabled, for example, can modify the link.xml file to look like:

```
<linker>
  <assembly fullname="Microsoft.MixedReality.Toolkit" preserve="all"/>
  <assembly fullname="Microsoft.MixedReality.Toolkit.Providers.WindowsMixedReality" preserve="all"/>
  <assembly fullname="Microsoft.MixedReality.Toolkit.Providers.WindowsVoiceInput" preserve="all"/>
  <assembly fullname="Microsoft.MixedReality.Toolkit.Providers.OpenVR" preserve="all"/>
  <assembly fullname="Microsoft.MixedReality.Toolkit.SDK" preserve="all"/>
  <assembly fullname="Microsoft.MixedReality.Toolkit.Services.InputSystem" preserve="all"/>
  REMOVE THIS LINE: <assembly fullname="Microsoft.MixedReality.Toolkit.Services.BoundarySystem" preserve="all"/>
  <assembly fullname="Microsoft.MixedReality.Toolkit.Services.CameraSystem" preserve="all"/>
[...]

Issue #4265, opened 2019-05-10
NuGet + IL2CPP + UWP Doesn't work for Hands (SystemType reference is broken)


chbecker-ms:
Take latest NuGet packages, build IL2CPP for UWP, and the hands won't work.

The log has the following (it appears getting the type from the serialized reference is broken).

Failed to properly initialize the MixedRealityToolkit
UnityEngine.DebugLogHandler:Internal_Log(LogType, String, Object)​
UnityEngine.DebugLogHandler:LogFormat(LogType, Object, String, Object[])​
UnityEngine.Logger:Log(LogType, Object)​
UnityEngine.Debug:LogError(Object)​
Microsoft.MixedReality.Toolkit.MixedRealityToolkit:Awake()​
 ​
(Filename: C:\buildslave\unity\build\Runtime/Export/Debug.bindings.h Line: 45)​
​
Unable to register a service with a null concrete type.​
UnityEngine.DebugLogHandler:Internal_Log(LogType, String, Object)​
UnityEngine.DebugLogHandler:LogFormat(LogType, Object, String, Object[])​
UnityEngine.Logger:Log(LogType, Object)​
UnityEngine.Debug:LogError(Object)​
Microsoft.MixedReality.Toolkit.MixedRealityToolkit:RegisterService(Type, SupportedPlatforms, Object[])​
Microsoft.MixedReality.Toolkit.MixedRealityToolkit:RegisterDataProvider(Type, SupportedPlatforms, Object[])​
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:Enable()​
Microsoft.MixedReality.Toolkit.<>c:<EnableAllServices>b__63_0(IMixedRealityService)​
System.Action`1:Invoke(T)​
Microsoft.MixedReality.Toolkit.MixedRealityToolkit:ExecuteOnAllServices(Action`1)​
Microsoft.MixedReality.Toolkit.MixedRealityToolkit:EnableAllServices()​
Microsoft.MixedReality.Toolkit.MixedRealityToolkit:OnEnable()​
 ​
(Filename: C:\buildslave\unity\build\Runtime/Export/Debug.bindings.h Line: 45)​
​
Unable to register a service with a null concrete type.​
UnityEngine.DebugLogHandler:Internal_Log(LogType, String, Object)​
UnityEngine.DebugLogHandler:LogFormat(LogType, Object, String, Object[])​
[...]

PR #5274, opened 2019-07-10
Add a link.xml to Assets to prevent critical MRTK components from getting optimized away in IL2CPP builds

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4265

See this comment in that issue for some more details on what happened here:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4265#issuecomment-510246866

Also, in case you were wondering how this was generated, these are the assemblies that are referenced in the default MRTK profiles (i.e. look for things of type "SystemType" and look at their corresponding references). In the long run this has to be more automated, as this was obviously a highly manual process. See this issue (https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5273) for more details on that.

### DIFF
--- a/Assets/link.xml
+++ b/Assets/link.xml
@@ -4,7 +4,7 @@
     during IL2CPP builds. More details on when this is needed and why this is needed
     can be found here: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5273
 
-    If your application some specific services (for example, if teleportation system is
+    If your application doesn't use some specific services (for example, if teleportation system is
     disabled in the profile), it is possible to remove their corresponding lines down
     below (in the previous example, we would remove the TeleportSystem below).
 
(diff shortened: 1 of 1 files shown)


Assets/link.xml
```diff
@@ -0,0 +1,26 @@
+<linker>
+  <!--
+    This link.xml file is provided to prevent MRTK code from being optimized away
+    during IL2CPP builds. More details on when this is needed and why this is needed
+    can be found here: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5273
+
+    If your application some specific services (for example, if teleportation system is
```
keveleigh:
```suggestion
    If your application doesn't use some specific services (for example, if teleportation system is
```

PR #5275, opened 2019-07-10
Fix the broken NuGet packages.

wiwei:
Our NuGet packages have been broken due to this PR:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4858

Which updated assembly names, but didn't also update the corresponding nuspec files to ensure that they would continue to get copied over.

Thankfully, there was also this pretty sweet migration guide which makes it super clear where things moved: https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html


wiwei:
Thanks to @andreiborodin for fixing this - I'm just shepherding this in.

Issue #5279, opened 2019-07-11
Remove unused materials


Troy-Ferrell:
## Describe the bug

Remove un-used assets thanks to #5096 , focused mostly on eyetracking sample

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.11f1]
- MRTK Version [e.g. v2.0]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

wiwei:
Closing this one out, as the linked PR was merged. Feel free to reopen if there are other specific assets that you want to remove and such.

Issue #5251, opened 2019-07-09
PointerClickHandler and PointerHandler are almost identical, very confusing


julenka:
## Overview
There are two scripts located under Assets\MixedRealityToolkit.SDK\Features\Input\Handlers which are nearly identical, `PointerHandler` and `PointerClickHandler`. We should depracate/delete the PointerClickHandler class to reduce confusion regarding which class to use.

We should remove PointerClickHandler in favor of PointerHandler for three reasons:

* Both classes expose Unity events for listening to pointer events (down, up, clicked, etc.), though only PointerHandler can be configured via code (PointerClickHandler requires setup in the designer because the InputActionEventPair fields in PointerClickHandler are private.
* PointerClickHandler naming is misleading -- the class responds to all input events, but the naming leads devs to believe it is only for click events.
* PointerClickHandler requires devs to specify an input action for every event, the default action is None. Several devs did not notice this input action field for the event handlers, and did not change it from "None". Unless the input action field is configured properly, the events will not respond to most inputs.

julenka:
@luis-valverde-ms you know about pointer events -- do you have any context for why we have the duplicate? Or, suggestions on how to deprecate PointerClickHandler? My vote would be to just delete the file. @wiwei  is that appropriate to just delete the file PointerClickHandler?

luis-valverde-ms:
I think I just left `PointerClickHandler` in when I introduced `PointerHandler` to avoid breaking existing code, but the latter should supersede the former.

PR #5281, opened 2019-07-11
Deletion of unused assets

Troy-Ferrell:
## Overview
Removing mostly materials and assets found with #5096 that aren't used in the eye-tracking scenes or related

Thanks again @Cameron-Micka 

## Changes
- Fixes: #5279 

## Verification
Ran eye-tracking scenes on device


wiwei:
It might be worth updating the CI itself to catch these sorts of issues automatically and flag them, instead of having to be a manual process that is run every so often.

In general, we do have to be somewhat careful about this going forward, because even though something may not be used in the MRTK itself, they may still be used by consumers (i.e. in theory someone could have taken a dependency on coffeecup).

Just calling out that we have deleted "unused" stuff from the MRTK before and have caused issues because others have used it.

The particular things I was concerned about in this change (i.e. material_37 and 38 appear to have been ones that probably shouldn't have gone in in the first place) look to be okay.

For future readers, if you do stumble across this PR as something that got rid of a dependency you had, please file an issue and reference this PR.

Troy-Ferrell:
> It might be worth updating the CI itself to catch these sorts of issues automatically and flag them, instead of having to be a manual process that is run every so often.
> 
> In general, we do have to be somewhat careful about this going forward, because even though something may not be used in the MRTK itself, they may still be used by consumers (i.e. in theory someone could have taken a dependency on coffeecup).
> 
> Just calling out that we have deleted "unused" stuff from the MRTK before and have caused issues because others have used it.
> 
> The particular things I was concerned about in this change (i.e. material_37 and 38 appear to have been ones that probably shouldn't have gone in in the first place) look to be okay.
> 
> For future readers, if you do stumble across this PR as something that got rid of a dependency you had, please file an issue and reference this PR.

Thanks Will. All good points. My two key thoughts
1) I love the idea of this in CI but it may be annoying to filter correctly. For example, some assets such as prefabs, scenes, textures, shaders etc, may be "unused" but we would still want them there. Right now there are 150 assets excluding scenes marked as unused in the dependency window. That's a lot of warnings that we mostly want to ignore etc.

2) I think this re-iterates that we need a deprecation strategy. I'm not worried about removing these particular frivolous materials now, especially before GA. But if there is a script that may be used by existing people, such as #5251 that @julenka,  that is redundant or old, we need to start removing those and have a process to allow people to switch/migrate etc.

Cameron-Micka:
It would really cool to get all of our example scenes into the build settings (or a scriptable object) then we could track which scenes may have fallen into obscurity too. This could be wrapped into [this proposal](https://teams.microsoft.com/l/message/19:9fa7cf795ff34aed8cf5f4da15a99d0f@thread.skype/1562104603903?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=fcd54363-a331-4ab6-8b7c-5687877d77f8&parentMessageId=1562104603903&teamName=Mixed%20Reality%20Toolkit%20INTERNAL&channelName=Proposals&createdTime=1562104603903).

PR #5284, opened 2019-07-11
Update the MRTK NuGet packages to have only a Foundation and Examples package

wiwei:
The current NuGet package count is somewhat of a reflection of the folder structure of the MRTK, and ultimately in order to build interesting things you need to have all four MRTK packages (i.e. Core, Providers, SDK, Services) in order to build anything interesting (for example, things in Examples). Rather than ship four packages that all must be included, we'll avoid the overhead of having to juggle four different things and just ship those as a single package.

This is basically a followup to:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4387

Which the same thing (except there have been some asmdef changes since that was made, so the exact lines in the combined nuspec are a little different)

**Also** includes a fixup to the licenseUrl section (master isn't a branch that's present, but mrkt_release is)


andreiborodin:
Did you meant to make changes to the main foundation package to include other packages? Or is GitHub not showing the changes?

wiwei:
> Did you meant to make changes to the main foundation package to include other packages? Or is GitHub not showing the changes?

Dafuq. I'm scratching my head because I made those changes. Seeing what happened...

wiwei:
Okay, I have no idea how my change didn't get through for that one (I assume I staged an earlier change and that's what I committed, and the blew it away when I was swapping branches).

Issue #5239, opened 2019-07-09
Clipping plane not working on Hololens 1 in some cases


benjamincoven:
## Describe the bug

I am trying to add a clipping plane feature in my application using the MRTK standard shader. It works very well in the editor but once in the Hololens it doesn't.

I tried to build the project with the material gallery demo scene and here the clipping plane works. I surely missed something and maybe it's not a MRTK bug, but I can't figure what is the problem.

## Screenshots

in the editor

![image](https://user-images.githubusercontent.com/12698212/60872397-e544c800-a234-11e9-9fb4-7bda0afe0837.png)

![image](https://user-images.githubusercontent.com/12698212/60872814-a7946f00-a235-11e9-8bc2-3cce53b3e01c.png)

I am using the shader from  #4946

in the Hololens 1

![image](https://user-images.githubusercontent.com/12698212/60872647-5b492f00-a235-11e9-8ea0-8588a16ea39e.png)

## Expected behavior

Clipping plane works as in the editor

## Your Setup (please complete the following information)

- Unity Version 2019.1.8
- MRTK Version RC2.1

## Target Platform (please complete the following information)

- HoloLens 1

keveleigh:
@Cameron-Micka, any ideas?

Cameron-Micka:
@benjamincoven what I'm guessing is going on is that particular shader permutation isn't getting included with the build at build time. Can you check in the material's YAML text if the "_CLIPPING_PLANE" shader keyword is included? It might look something like this: `m_ShaderKeywords: _CLIPPING_PLANE, ...`

Since in your screenshot it looks like you have the "_CLIPPING_PLANE" keyword enabled I'm guessing the ClippingPlane script is turning off that keyword automatically for you (maybe on scene onload?).

I plan to included all the clipping permutations in builds to allow developers to turn these features on dynamically, I just need to verify it doesn't explode our shader permutations.

Cameron-Micka:
One thing to try is check the "Clipping Plane" checkbox to enabled on the material before opening your scene/prefab that has the ClippingPlane component.

benjamincoven:
Thanks for your help.

The material YAML has the _CLIPPING_PLANE keyword, I tried this on one of the MeshRenderer:

```
void Update()
    {
        Debug.Log(Renderer.material.IsKeywordEnabled("_CLIPPING_PLANE"));
    }
```

It's true.
[...]

Issue #5290, opened 2019-07-11
Additive Lit Materials Do Not Get Clipped Properly by Clipping Primitives


Cameron-Micka:
## Describe the bug

Additive materials which are lit do not get clipped properly by Clipping Primitives (plane, sphere, box). The lit pixels are still visible.

## To reproduce

1. Add a renderer to the scene that uses the MRTK/Standard material that is both additive and lit.
2. Add a clipping primitive to the same scene.
3. Add the renderer to the "Renderers" property of the clipping primitive
4. Notice the renderer does not get clipped completely on the lit portions.

## Expected behavior

All pixels are clipped by the clipping primitive.

Troy-Ferrell:
@Cameron-Micka fixed?

Cameron-Micka:
Yep fixed!

Issue #5289, opened 2019-07-11
Clipping Primitive Editor Support


Cameron-Micka:
## Describe the bug

When adding a renderer to a clipping primitive (plane, sphere, box) in the editor (not in play mode) the renderer does not get clipped until play mode is entered. This is a regression from https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4899 

## To reproduce

1. Add a renderer to the scene that uses the MRTK/Standard material.
2. Add a clipping primitive to the same scene.
3. Add the renderer to the "Renderers" property of the clipping primitive
4. Notice the renderer does not get clipped.

## Expected behavior

The renderer gets clipped as soon as it is added to the "Renderers" property of the clipping primitive.

PR #5291, opened 2019-07-11
Clipping Primitive Fixes

Cameron-Micka:
## Overview

The pull request address a few issues around clipping primitives.

1) Clipping primitives can now be toggled on/off at runtime without requiring shader permutations to be know at (player) build time. This adds some extra shader permutation memory overhead, but is still quite small when considered against traditional 3D applications (games).

2) Materials which are additive and lit are not clipped properly, before the lit portions of the material were not clipped. There is also a new property for non-opaque or cutout materials called **Blended Clipping Width** which allows developers to set the width of the clip blend region (before it was hardcoded to always be 1 meter in length).

Capture

3) Clipping primitives can now be easily visualized in the editor which was a regression from https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4899

4) Also removed some (lower) range constraints to aide with materials used on objects with tiny scales.

Clip

> From left to right opaque, additive, and transparent materials.

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5239 https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5289 https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5290

## Verification
Check out the MaterialGallery scene to determine if primitive clipping is still working.
>
>
>
>


Cameron-Micka:
Any chance we can get this reviewed/merged in soon? Thank you!

Issue #4289, opened 2019-05-10
Changing ActiveProfile at runtime does not work


chbecker-ms:
(Ported to GitHub, reference 21522491)

Issue encountered using MRTK V2.0.0 RC1 Refresh

There's a property on MRTK.Instance called ActiveProfile, which supposedly will
swap out the profile and reset the configuration, but when I attempted to call
this I ran into a bunch of the below errors, and fixing just the camera related
errors exposed a lot more.  I managed to "fix" these locally by actively by
hunting for objects that weren't cleaned up and fixing them up.  Below is most
of the set of changes I had to make to allow swapping active profiles.

**Changes I made to get this working in my project:**

- [x]  MixedRealityToolkit.cs:UpdateAllServices

    - [x]  Removal of activeSystems, and registeredMixedRealityServices in foreach
        loop was causing loop to fail due to the collection changing.  Added a
        ToList() on each to fix.

    [davidkline-ms] addressing with try/catch(InvalidOperationException) for better performance.

- [x]  UIRaycastCamera

    - [x]  Had to change destroy call to DestroyImmediate in FocusProvider.cs from
        Destroy, as we needed to clean up references in a single frame.

    - [ ]  Added IsNull check in MixedRealityInputModule.cs:
        GetMousePointerEventDataForMrtkPointer to check for a destroyed
        RaycastCamera, and find the new one in the scene if so.  Also since
        there is some delay I added a short circuit return for when
        RaycastCamera was null.

- [ ]  ShellHandRayPointer

    - [ ]  Any hand ray or controller pointer could become stale when swapping
        profiles, because we do not disable the linerenderer 100% of the time
        when setting a shell hand ray pointer inactive.  Had to add a foreach
        loop to OnPostSceneQuery in the !IsActive branch to manually disable all
        linerenderers.

- [x] Gaze Cursor

    - [x] Had to add setting the GazeCursor in GazeProvider.OnEnable, since it was not swapping gaze cursors when swapping profiles.

[davidkline-ms] Solving using DestroyImmediate to ensure correct re-creating.

- [ ]  WindowsMixedRealityDeviceManager: Disable()

    - [ ]  Had to set gestureRecognizer, and NavigationRecognizer to null after
        dispose.  We were trying to ref the disposed instances.

- [ ] GestureRecognizreExtensions

    - [ ]  Had to add null check for recognizer to avoid accessing disposed or
        destroyed instance.

[...]

PR #5292, opened 2019-07-11
Service locator service iterator stability

davidkline-ms:
It was pointed out in #4289 that the iterators used in the service locator are impacted by service collection changes during profile change at runtime.

While reviewing #5268, this issue was encountered when loading a new scene when the scene service is enabled in the main configuration profile.

This change iterates over a copy of the collection instead of the live collection which protects from such changes.

It also updates the LateUpdateAllServices to match the other methods (ex: UpdateAllServices) by calling ExecuteOnAllServices instead of using a near identical implementation.


Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -1022,12 +1006,12 @@ private bool ExecuteOnAllServices(Action<IMixedRealityService> execute)
             // If the Mixed Reality Toolkit is not configured, stop.
             if (!HasProfileAndIsInitialized) { return false; }
 
-            foreach (var system in activeSystems)
+            foreach (var system in activeSystems.ToList())
```
keveleigh: Is this now going to allocate multiple times a frame on each `Update`, `LateUpdate`, etc?

davidkline-ms: unfortunately, yes. i have not found a safer way to handle this scenario. ideas?

keveleigh: If so, I wonder if we should try to follow the pattern in `BaseEventSystem` instead, where it saves off actions (in that case, registering a listener) for a predetermined, known good time in the app cycle

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/0734ab93655113a974682137477cd095d63f87e2/Assets/MixedRealityToolkit/Services/BaseEventSystem.cs#L208

Any profile changes (which I think is what this is aiming to support) could be saved off and handled maybe at the end of update / lateupdate / some known good time, probably near the end of the frame

SimonDarksideJ: This would be preferable using a static list that is populated/cleared down.  ToList has too many performance overheads

Issue #5116, opened 2019-06-28
Add warnings when services with active listeners get destroyed


lukastoenneMS:
## Describe the problem

MRTK services can get destroyed at runtime, removing event listeners etc.
Similarly services can get created at runtime _after_ components that use them.
Even if components check for a service on Start/Awake/Enable, that service may be removed and recreated later, leaving a dysfunctional component.

### Example:
During tests we create a scene with a camera and a MixedRealityInputModule on it (for piping events to the Unity event system). Between tests the event system service is torn down and recreated for the next test, which leaves a broken input module that will not receive source state events any longer.

The fix for this is to destroy the input module at the end of each test and recreate it for the next. But this kind of problem is easy to produce and hard to detect.

## Describe the solution you'd like

There should be a debug more and/or test where services print warnings when they are destroyed while listeners are still registered. That way we at least know when components may be broken.

IssueSyncBot:
We appreciate your feedback and thank you for reporting this issue.

Microsoft Mixed Reality Toolkit version 2 (MRTK2) is currently in limited support. This means that Microsoft is only fixing high priority security issues. Unfortunately, this issue does not meet the necessary priority and will be closed. If you strongly feel that this issue deserves more attention, please open a new issue and explain why it is important.

Microsoft recommends that all new HoloLens 2 Unity applications use MRTK3 instead of MRTK2.

Please note that MRTK3 was released in August 2023. It features an all-new architecture for developing rich mixed reality experiences and has a minimum requirement of Unity 2021.3 LTS. For more information about MRTK3, please visit [https://www.mixedrealitytoolkit.org](https://www.mixedrealitytoolkit.org).

Thank you for your continued support of the Mixed Reality Toolkit!

PR #5300, opened 2019-07-12
Add error logs on destroying event system with dangling listeners

googlan:
## Overview

This change has originally been triggered by #5116 to first of all ease test debugging, when test runtime environment is not clean after running another test.
So, the change should only have been messages in `BaseEventSystem`. However, adding these messages highlighted few fundamental issues:
* Mrtk had incorrect order of deinitialization for core systems. They should be disabled and destroyed in a reverse order to how they are initialized. This PR fixes that.
* Unregistering event handlers in unity `OnDestroy` is semantically incorrect: Mrtk tear down happens in the middle of `Update` loop, which Deinitializes and Destroys all global services. When components are then destroyed _in the end_ of the `Update` loop, they can't rely on valid state of services (e.g. Input System). Ideally we would like global listeners to go away in `OnDisable`. However, this would require to move their initialization to `OnEnable`, which is much trickier to do. Solved this by modifying `IMixedRealityCursor` interface. I don't like it, but it solves an issue.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
@@ -182,6 +182,15 @@ namespace Microsoft.MixedReality.Toolkit.Input
         }
 
         /// <inheritdoc />
+        public virtual void Destroy()
+        {
+            // Cursor needs to unregister its input handlers explicitly, while input system is still active.
+            // If this would be done from OnDestroy, it will happen in the end of Update loop,
+            // when the input system itself is already destroyed.
+            UnregisterManagers();
+        }
+
+        /// <inheritdoc />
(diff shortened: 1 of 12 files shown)


wiwei:
@keveleigh to take a peek here as well.

The change to IMixedRealityCursor feels not great to me too, though the alternative (of having a separate interface just for the shutdown/destroy method) really feels like overkill. It feels like the lesser of two evils.

Technically this is a breaking change (though because we provide the base implementation as well in BaseCursor I think that it should be okay

googlan:
@wiwei Agree about everything said regarding `IMixedRealityCursor`. Just need to choose the lesser evil.

googlan:
Apparently when unloading a scene Unity disables/destroys game objects in a scene in an arbitrary order. `HandInteractionExamples` scene generates different sequence of error messages for every "play-stop" in the Editor (not tests, just running it from the Editor). 
I'm gonna disable these warnings in `EventSystem.Destroy` by default and enable them in our play mode tests, which use standard TestSetup. They can be enabled for normal debugging if necessary, but I think in most non-test cases they will cause more confusion than benefit.

PR #5302, opened 2019-07-12
Update the build scripts to produce Foundation and Examples .unitypackages.

wiwei:
build.ps1 used to be used to do both nuget packing and unity package generation. It hasn't been used since we moved NuGet pack to the YAML pipelines themselves, the way it worked before anyways was by looking at the directory that nuspec files were in, which is not valid for packages like the Foundation one (which contains a number of other folders).

This change removes the old build.ps1 script in favor of having a script that is explicitly about building MRTK .unitypackage files. Inside you will find the configuration itself (i.e. the set of packages along with their corresponding folders)


scripts/packaging/unitypackage.ps1
```diff
@@ -0,0 +1,245 @@
...
+
+    - Examples
+
+    This contains all of the content under MixedRealityToolkit.Examples
+
+    Note that these packages are intended to mirror the NuGet packages
+    described in Assets/MixedReality.Toolkit.Foundation.nuspec and
+    Assets/MixedRealityToolkit.Examples/MixedReality.Toolkit.Examples.nuspec.
+
+    Assumes that the current working directory of the script is in the root
```
wiwei:
```suggestion
    Defaults to assuming that the current working directory of the script is in the root
```

scripts/packaging/unitypackage.ps1
```diff
@@ -0,0 +1,245 @@
...
+
+Import-Module UnitySetup -MinimumVersion '4.0.97' -ErrorAction Stop
+
+# This hashtable contains mapping of the packages (by name) to the set
+# of top level folders that should be included in that package.
+# The keys of this hashtable will contribute to the final naming of the package
+# (for example, in Microsoft.MixedReality.Toolkit.Unity.Foundation, the Foundation
+# section comes from the key below).
+#
+# Note that capatilization below in the key itself is significant. Capitalization
```
keveleigh:
```suggestion
# Note that capitalization below in the key itself is significant. Capitalization
```

wiwei: yikes

wiwei:
To be totally honest, I think that there might be something weird with the Write-Error stuff in the end (I'm not sure that it will actually kill the CI, which it should). I'll have to play around with this on CI to see what happens however since it is something that really only becomes apparent there.

Issue #5301, opened 2019-07-12
Clipping Primitive Documentaton/Example Use is Sparse


Cameron-Micka:
## Describe the Issue
Documentation around how to use the clipping primitives in conjunction with the MRTK/Standard shader is pretty minimal and developers aren't always sure how to use them.

## Feature Area
An example scene using the clipping primitives (other than the MaterialGallery) would be very helpful for developers to understand their usage.

## Existing Doc Link
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Documentation/README_MRTKStandardShader.md#additional-feature-documentation

PR #5303, opened 2019-07-12
Clipping Examples Scene

Cameron-Micka:
## Overview

Many developers have expressed interest in clipping their meshes but are unaware of what is already possible with the MRTK/Standard shader in conjunction with clipping primitives. The scene in this pull request, called **ClippingExamples**, shows how the MRTK/Standard shader in conjunction with clipping primitives (ClippingPlane, ClippingSphere, & ClippingBox) can be used to dynamically clip pixels on a mesh.

Screenshot_2019-07-12_02-37-28-PM

Try grabbing the sphere or box which a near or far interaction. Then move the sphere or box until it intersects with the heart. You will notice that any portions of the heart that intersect with the sphere or box will not be rendered.

ClipExample2

The heart model comes from [Microsoft's gallery on Remix3D](https://www.remix3d.com/details/G009SVN05LMF?section=other-models).

@designnomad came up with the original idea to build a clipping example scene for developers. Thank you Hae Jin!

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5301

## Verification
Open the ClippingExamples scene and try moving the clipping primitives. 

>
>


wiwei:
So cool!

julenka:
This needs to be posted to Twitter. So cool!

Sent from phone, please excuse typos.

cre8ivepark:
Shared on Twitter and multiple FB HoloLens groups!
https://twitter.com/cre8ivepark/status/1149859635626532864

cre8ivepark:
Do you need any additional setup other than checking 'Clipping Border' in the target object's shader?

Cameron-Micka:
@cre8ivepark if you are working with late-ish mrtk_development all you need to do is drag and drop the renderers you want to clip into the “Renderers” property array on your ClippingPlane, Sphere or Box. It even works in edit mode if you want to test. No need to check “Clipping Border” on the material unless you want a border. Let me know if that works for you!

Assets/MixedRealityToolkit/Utilities/ClippingSphere.cs
```diff
@@ -12,14 +12,17 @@ namespace Microsoft.MixedReality.Toolkit.Utilities
...
-        [Tooltip("The radius of the clipping sphere.")]
-        [SerializeField]
-        protected float radius = 0.5f;
-
         /// <summary>
-        /// The radius of the clipping sphere.
+        /// The radius of the clipping sphere, which is determined by the largest axis of the transform's scale.
         /// </summary>
-        public float Radius => radius;
+        public float Radius
```
wiwei: Can you also add a note to:

https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html

This is technically a breaking change because we're changing the shape of our API to one that is explicit (i.e. passing a radius via config or runtime setting) to one that is implicit based on the transform of the tracked object. It seems like this one seems more useful especially for clipping sphere, but it would be good to call out the change there (i.e. before  you would have to track the transform and update the radius yourself, now it automatically tracks it, if you want to change it, just update the transform of the sphere as you normally would.)

Cameron-Micka: Thank you for reviewing, I’ll update this document tomorrow!

Cameron-Micka:
@cre8ivepark Awesome! It looks like @Troy-Ferrell has already done the work to convert the rover materials in this PR: https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5333

Issue #4755, opened 2019-06-06
Problematic BoundingBox bounds setting & calculations


Troy-Ferrell:
## Describe the bug

The BoundingBox components code path for calculating bounds has various issues depending on certain cases & uses. 

1) Bounds Overide property is private [SerializedField] and thus cannot be set at runtime. For example, one customer loads model file dynamically and AddComponent<BoundingBox>() but cannot set the bounds override. 

Issues with BoundingBox.GetTargetBounds()
2) The first phase of the method looks at children objects that are not part of the BB rig. However, the code only takes into account direct children of the Target object and not sub-children. (Also, code can be simplified to be foreach (Transform child in transform) {} )

3) GetSingleObjectBounds() is not accurate as a default BoxCollider added to an empty gameobject will be center <0,0,0> but size <1,1,1> which will pass the zero size check in the algorithm

4) The next stage of the algorithm can be simplified by leveraging the existing code in MRTK of Transform.GetColliderBounds() which does the exact same thing

5) There is also already an enum in GetTargetBounds() for what type of bounds calculation was utilized. It would actually be nice to have this setting visible to the user for them to choose which method of bounds calculation should be utilized. 

Related issues:
#4557 
BoundingBox doesn't recompute bounds when boundsOverride is modified

#4434 
Bounding box scaling varies based on whether the box is directly on the target/collider or parented to it

## To reproduce

Notice sub-children beyond L1 are not considered even if they have colliders attached
Notice empty game objects having impact on bounds calculation when using encapsulation 

## Expected behavior

1) Making BoundsOverride public settable property like other updated properties (ShowWireframe etc)

2) GetTargetBounds() when calculating bounds on first attempt, should look at all children in hierarchy of target, not just L1

3) Empty gameobjects should be ignored for calculating bounds. 

4) Re-use code of GetColliderBounds() already in MRTK to reduce redundancy 

## Your Setup (please complete the following information)

- Unity Version 2018.4
- MRTK Version RC2 (latest right before launch

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

Alexees:
#5307 Fixes 3.

Alexees:
The current code looks like as if BoundsOverride can be set via property

Troy-Ferrell:
@thalbern FYI in case any of this overlaps for the new UX design

keveleigh:
[...]

PR #5307, opened 2019-07-15
BoundingBox now correctly calculates bounds

### LINKED ISSUE #5111: BoundingBox bounds calculation issues
## Describe the bug

The BoundingBox has several issues related to calculating its bounds. To inverse the bug, the BoundingBox only calculates them correctly when one of these cases is true:

- The BoundingBox script is on the GameObject that has the collider and does not have any child colliders
- The BoundingBox stays Axis aligned while being initialized with colliders at child level 1 only

Here are the cases that I found:

- If the BoundingBox's GameObject, with collider, has another child with a collider, the calculation only happens on the child

image

- If the BoundingBox is calculating either a child or multiple colliders while rotated, the size is wrong and based on the axis aligned bounds of the calculated colliders

image

## To reproduce

1. Open the "BoundingBoxRuntimeExample" scene
2. Replace the first block of code in BoundingBoxTest.cs with this:
```
                var cube = GameObject.CreatePrimitive(PrimitiveType.Cube);
                Debug.Assert(darkGrayMaterial != null);
                cube.GetComponent<MeshRenderer>().material = darkGrayMaterial;
                cube.transform.position = cubePosition;
                var child2 = GameObject.CreatePrimitive(PrimitiveType.Cube);
                child2.transform.SetParent(cube.transform, false);
                child2.transform.localScale *= .2f;
                child2.transform.localPosition = Vector3.right;
                //cube.transform.rotation = new Quaternion(.5f, .5f, .5f, 1f);
```
2. Press play and see how the BoundingBox is being calculated only for child2
3. outcomment the last line
4. Play again and see how the bounds of child2 are too large

A couple of findings I made:

The following code does not take BoundingBox GameObject itself into account and it only interates one level of children (#4755 also mentions this):
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/ab6bfc0b97a59d2ba76b2f1695c43115aba3bb63/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs#L1183-L1190

GetSingleObjectBounds simply uses collider bounds, which are axis aligned world space bounds and therefore inappropriate for the calculation. 

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/ab6bfc0b97a59d2ba76b2f1695c43115aba3bb63/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs#L1293-L1310

## Expected behavior

Whatever hierarchical structure, I expect the BoundingBox to include all colliders, starting with the GameObject it is attached to. If it's GameObject does not have a Collider, bounds should be calculated for all children.
It should not matter how the colliders are oriented, the Box should be aligned with the Transform it belongs to.

- Unity Version 2018.4.2f1
- MRTK Version mrtk_development 27.06.

- Unity Editor

## Additional context

I remember the old HTK BoundingBox to be capable of dealing with this by adding all transformed corner points of all bounds together and then using this as local bounds volume for the box.

Alexees:
## Overview
The BoundingBox had a couple of issues with calculating its bounds that this pull request adresses:

1. It only considered its own, or the first descendant's bounds for the calculation

This led to the box never really spanning across all objects.
Now, the BoundingBox wraps the whole hierarchy.

2. Calculations were based on Renderer bounds

Those bounds are in world space and therefore not ideal for the calculation of the bounds of a rotated object. The pull request assumes the box stays aligned with the Gameobject the script is attached to and only calculates bounds from MeshRenderer.sharedMesh or Colliders, for which local space variants exist in the BoundsExtensions class.

3. Bounds were calculated by some predefined order of component appearance

The user now can choose between 4 options of how he wants to get the box calculated:

- RendererOverCollider
- ColliderOverRenderer
- ColliderOnly
- RendererOnly

Motivation for this change came from the old MRDL BoundingBox, which transferred objects corner points into the space of the boundingbox parent, resulting in a perfect box wrapping everything

Note, the previous implementation was broken, meaning, all other Renderer calculations were broken anyway, so if necessary, other Renderer bounds calculation need to be added if necessary

## Fixes
#5111 #5307


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1315,159 +1339,108 @@ private void SetBoundingBoxCollider()
...
+                cachedTargetCollider.center = bounds.center;
+                cachedTargetCollider.size = bounds.size;
             }
 
-            Vector3 scale = cachedTargetCollider.transform.lossyScale;
-            Vector3 invScale = new Vector3(1.0f / scale[0], 1.0f / scale[1], 1.0f / scale[2]);
-            cachedTargetCollider.size += Vector3.Scale(boxPadding, invScale);
+            if (boxPadding != Vector3.zero)
+            {
+                Vector3 scale = cachedTargetCollider.transform.lossyScale;
```
julenka: Is it possible to hit divide by zero here?

Alexees: Will check, the code was already there, I just wrapped it checking boxPadding

julenka:
Thanks for making this fix and sorry for the late review! I noticed in #5111 you provided a code block that you used to verify your fix -- could you please add these cases to BoundingBoxTest.cs?

The code I'm referring to is the following:
```
                var cube = GameObject.CreatePrimitive(PrimitiveType.Cube);
                Debug.Assert(darkGrayMaterial != null);
                cube.GetComponent<MeshRenderer>().material = darkGrayMaterial;
                cube.transform.position = cubePosition;
                var child2 = GameObject.CreatePrimitive(PrimitiveType.Cube);
                child2.transform.SetParent(cube.transform, false);
                child2.transform.localScale *= .2f;
                child2.transform.localPosition = Vector3.right;
                //cube.transform.rotation = new Quaternion(.5f, .5f, .5f, 1f);
```

julenka:
> could you please add these cases to BoundingBoxTest.cs?

In fact, it would be *even better* to add these tests directly to the boundingbox playmode tests, located at Assets\MixedRealityToolkit.Tests\PlayModeTests\BoundingBoxTests.cs 

We should work towards moving all the tests in BoundingBoxRuntimeExamples into the PlayMode tests so they can be run as part of PR validation and continuous integration (CI). That's out of scope for this PR, but your specific case would be a great place to start.

julenka:
Hi Alex, I pulled your changes and ran the BoundingBoxRuntimeExample test, and I found that one of the steps had incorrect bounds:

image


Could you please fix?

Alexees:
> > could you please add these cases to BoundingBoxTest.cs?
> 
> In fact, it would be _even better_ to add these tests directly to the boundingbox playmode tests, located at Assets\MixedRealityToolkit.Tests\PlayModeTests\BoundingBoxTests.cs
> 
> We should work towards moving all the tests in BoundingBoxRuntimeExamples into the PlayMode tests so they can be run as part of PR validation and continuous integration (CI). That's out of scope for this PR, but your specific case would be a great place to start.

I'll go and add it, but with the slight modification of the child being scewed. Scaling AND Rotating both parent and child would be the ultimate configuration. Ok, when the BoundingBox parent is scewed, we're screwed 👅

Alexees:
> Hi Alex, I pulled your changes and ran the BoundingBoxRuntimeExample test, and I found that one of the steps had incorrect bounds:
> 
> image
> 
> Could you please fix?

Fixed

Assets/MixedRealityToolkit.Examples/Demos/UX/BoundingBox/Scripts/BoundingBoxTest.cs
```diff
@@ -93,6 +93,7 @@ private IEnumerator Sequence()
 
                 SetStatus("Change target bounds override size");
                 bc.size = new Vector3(0.5f, 0.1f, 1);
```
julenka: The expected behavior is for the bbox component to actually detect the change in the bounds size automatically, calling createrig will actually hide the real bug. Can you remove this and I'll file a bug that bbox doesn't change when the bounds changes?

Assets/MixedRealityToolkit.Examples/Demos/UX/BoundingBox/Scripts/BoundingBoxTest.cs
```diff
@@ -192,20 +193,45 @@ private IEnumerator Sequence()
...
+                while (childTransform.childCount > 0)
+                {
+                    childTransform = childTransform.GetChild(0);
+                    float baseScale = lastParent == null ? 0.1f : 1f;
+                    childTransform.transform.localScale = new Vector3(baseScale * Random.Range(.5f, 2f), baseScale * Random.Range(.5f, 2f), baseScale * Random.Range(.5f, 2f));
+                }
+
+                bbox.WireframeEdgeRadius = 1f;
+                bbox.CreateRig();
+                SetStatus("Delete GameObject");
```
julenka: Thanks for more tests :-)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -79,16 +79,20 @@ private enum CardinalAxisType
         }
 
         /// <summary>
-        /// This enum is used internally to define how an object's bounds are calculated in order to fit the boundingbox
+        /// This enum defines what volume type the bound calculation depends on and its priority
         /// to it.
         /// </summary>
-        private enum BoundsCalculationMethod
+        public enum BoundsCalculationMethod
```
julenka: It's not immediately clear to me what these enums mean, which makes me think folks consuming this will be confused as well. Could you please explain what each value means in the summary comment? This way it will be captured in the documentation so folks can reference it.

Assets/MixedRealityToolkit.Examples/Demos/UX/BoundingBox/Scripts/BoundingBoxTest.cs
julenka: Filed #5414 to track fixing the bug

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1536,158 +1566,121 @@ private void SetBoundingBoxCollider()
...
-            BoxCollider boxCollider = Target.AddComponent<BoxCollider>();
-            bounds = boxCollider.bounds;
-            Destroy(boxCollider);
-            boundsMethod = BoundsCalculationMethod.Collider;
-            return bounds;
+            return finalBounds;
         }
 
-        private Bounds GetSingleObjectBounds(GameObject gameObject)
+        private bool AddRendererBoundsToTarget(KeyValuePair<Transform, Bounds> rendererBoundsByTarget)
```
julenka: In MRTK when we try to do something and then return bool if succeeded / failed, we normally use TryXYZ, could you please rename to TryAddRendererBoundsToTarget?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1536,158 +1566,121 @@ private void SetBoundingBoxCollider()
...
-            }
-            else
-            {
-                bounds = boxCollider.bounds;
-            }
+            if (rendererBoundsByTarget.Key == null) { return false; }
+
+            Vector3[] cornersToWorld = null;
+            rendererBoundsByTarget.Value.GetCornerPositions(rendererBoundsByTarget.Key, ref cornersToWorld);
+            colliderCorners.AddRange(cornersToWorld);
```
julenka: To help with understandability in future, could you please add comment to colliderCorners explaining what that variable represents?

Assets/MixedRealityToolkit/Extensions/BoundsExtensions.cs
```diff
@@ -354,54 +354,62 @@ public static void GetColliderBoundsPoints(GameObject target, List<Vector3> boun
...
-                    Bounds boxBounds = new Bounds(bc.center, bc.size);
-                    boxBounds.GetCornerPositions(bc.transform, ref corners);
-                    boundsPoints.AddRange(corners);
+        /// <summary>
+        /// Method to get bounds from a single Collider
+        /// </summary>
+        /// <param name="collider">Target collider</param>
+        /// <param name="boundsPoints">array reference that gets filled with points</param>
+        /// <param name="ignoreLayers">layerMask to simplify search</param>
+        public static void GetColliderBoundsPoints(Collider collider, List<Vector3> boundsPoints, LayerMask ignoreLayers)
```
julenka: This looks like a function that would be a great candidate for a unit test / edit mode test, since it has known inputs and outputs. It will be great to have to not only verify your changes but more importantly make sure nobody breaks you in the future :-)

Assets/MixedRealityToolkit/Extensions/BoundsExtensions.cs
```diff
@@ -354,54 +354,62 @@ public static void GetColliderBoundsPoints(GameObject target, List<Vector3> boun
...
+                sphereBounds.GetFacePositions(sc.transform, ref corners);
+                boundsPoints.AddRange(corners);
+            }
+            else if (collider is BoxCollider)
+            {
+                BoxCollider bc = collider as BoxCollider;
+                Bounds boxBounds = new Bounds(bc.center, bc.size);
+                boxBounds.GetCornerPositions(bc.transform, ref corners);
+                boundsPoints.AddRange(corners);
+
```
julenka: nit: extra line

Assets/MixedRealityToolkit/Extensions/BoundsExtensions.cs
julenka: Since this PR fixes existing functionality I would be okay to complete this PR without tests, but would strongly suggest to add if possible. If not, could you please file an issue to write a test for this function so we don't forget? You can either assign it to yourself if you have time to help, or we can find somebody later to help with this.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -835,6 +859,9 @@ private IMixedRealityInputSystem InputSystem
         private bool isChildOfTarget = false;
         private static readonly string rigRootName = "rigRoot";
 
+        // Cache for Collider corner points, returned as List from BoundsExtensions
+        private static List<Vector3> colliderCorners = new List<Vector3>();
```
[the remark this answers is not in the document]
Alexees: The list is only ever used during the bounds calculation process and cleared at the beginning of it. Unless you multi-thread bounds calculations, this list works for any number of BBs.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1283,15 +1310,18 @@ private void AddComponentsToAffordance(GameObject afford, Bounds bounds, Rotatio
...
-            foreach (Renderer r in g.GetComponentsInChildren<Renderer>())
+            Mesh currentMesh;
+            foreach (MeshFilter r in g.GetComponentsInChildren<MeshFilter>())
             {
+                if ((currentMesh = r.sharedMesh) == null) { continue; }
+
                 if (b.size == Vector3.zero)
                 {
-                    b = r.bounds;
+                    b = r.sharedMesh.bounds;
```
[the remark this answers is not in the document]
Alexees: you're right, it's currentMesh and it does not matter ;)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1521,7 +1551,7 @@ private Vector3 GetBoxDisplayScale()
         private void SetBoundingBoxCollider()
         {
             // Make sure that the bounds of all child objects are up to date before we compute bounds
-            UnityEngine.Physics.SyncTransforms();
+            UnityPhysics.SyncTransforms();
 
             //Collider.bounds is world space bounding volume.
```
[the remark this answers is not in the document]
Alexees: Not sure about the whole `SyncTransforms `and `cachedTargetCollider.transform.hasChanged = true;`
 part, cause I have no idea what could make these relevant, but the comments after that are not relevant anymore.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1536,158 +1566,121 @@ private void SetBoundingBoxCollider()
...
+
+            // Collect all Transforms except for the rigRoot(s) transform structure(s)
+            // Its possible we have two rigRoots here, the one about to be deleted and the new one
+            // Since those have the gizmo structure childed, be need to ommit them completely in the calculation of the bounds
+            // This can only happen by name unless there is a better idea of tracking the rigRoot that needs destruction
+
+            List<Transform> childTransforms = new List<Transform>();
+            childTransforms.Add(Target.transform);
+
+            foreach (Transform childTransform in Target.transform)
```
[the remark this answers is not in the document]
Alexees: yeah and as distinct from `foreach (Transform childTransform in Target.GetComponentsInChildren<Transform>())`  
 it excludes the target transform already.

https://docs.unity3d.com/ScriptReference/Transform.html

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1536,158 +1566,121 @@ private void SetBoundingBoxCollider()
...
+
+                if (boundsCalculationMethod != BoundsCalculationMethod.RendererOnly)
                 {
-                    Bounds colliderBounds = colliders[i].bounds;
-                    if (colliderBounds.size != Vector3.zero)
-                    {
-                        bounds.Encapsulate(colliderBounds);
-                    }
+                    Collider collider = childTransform.GetComponent<Collider>();
+                    colliderByTransform = collider != null ? new KeyValuePair<Transform, Collider>(childTransform, collider) : new KeyValuePair<Transform, Collider>();
```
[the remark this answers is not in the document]
Alexees: KeyValuePair is a struct and I needed something to check for null in the code. But since I'm only to continue the loop in that case I can already do that here and just not create a KeyValuePair at all.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1536,158 +1566,121 @@ private void SetBoundingBoxCollider()
...
 
-                boundsMethod = BoundsCalculationMethod.Renderers;
-                return bounds;
+                // Do the collider for the one case that we chose RendererOverCollider and did not find a renderer
+                AddColliderBoundsToTarget(colliderByTransform);
             }
 
-            MeshFilter[] meshFilters = Target.GetComponentsInChildren<MeshFilter>();
-            if (meshFilters.Length > 0)
+            if (colliderCorners.Count == 0) { return new Bounds(); }
```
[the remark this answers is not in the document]
Alexees: If we neither found a single renderer nor a single  collider, the list of corners is empty and we can just return no bounds at all. The name might be misleading though, colliderCorners also contains RendererBoundsCorners. Renamed it to totalBoundsCorners.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Alexees: not returning a bool anymore, therefore skipped

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Alexees: I rephrased it, but the variable is explained where it's declared.

Assets/MixedRealityToolkit/Extensions/BoundsExtensions.cs
Alexees: Sadly, I don't have the time to write tests too. Writing fixes is already a time consuming process. Btw. I'm not allowed or don't know how to assign issues to myself.

> could you please file an issue to write a test for this function so we don't forget

It's not new really. I only extracted this from the other existing method which was meant for collider arrays only and I needed it for one collider at a time. The other method hopefully does have a test already.

Assets/MixedRealityToolkit/Extensions/BoundsExtensions.cs
Alexees: what does nit mean?

Assets/MixedRealityToolkit/Extensions/BoundsExtensions.cs
julenka: Filed #5486

Assets/MixedRealityToolkit/Extensions/BoundsExtensions.cs
julenka: It just means a small comment on something that's not super important but might make things more readable.

julenka:
@Alexees sorry about the late reply again -- I'm actually on vacation right now (went on vacation last Friday). I've approved and kicked off the tests. I get back from vacation middle of next week and will be much more available then.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -79,16 +79,24 @@ private enum CardinalAxisType
...
+        /// ColliderOnly: Omits Colliders and uses Renderers for the bounds calculation exclusively
         /// </summary>
-        private enum BoundsCalculationMethod
+        public enum BoundsCalculationMethod
         {
-            Collider = 0,
-            Colliders,
-            Renderers,
-            MeshFilters
+            RendererOverCollider = 0,
```
Troy-Ferrell: Can you move the summary comments to each enum? i.e
/// <summary> blah </summary>
RendererOverCollider = 0,
/// <summary> blah </summary>
ColliderOverRenderer , etc

This way they should auto show up in the doc description. example although blank:
https://microsoft.github.io/MixedRealityToolkit-Unity/api/Microsoft.MixedReality.Toolkit.UI.BoundingBox.WireframeType.html

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -143,6 +151,22 @@ public BoxCollider BoundsOverride
             }
         }
 
+        [SerializeField]
+        [Tooltip("Defines the volume type and the priority for the bounds calculation")]
+        private BoundsCalculationMethod boundsCalculationMethod = BoundsCalculationMethod.RendererOverCollider;
+        public BoundsCalculationMethod CalculationMethod
```
Troy-Ferrell: Add /// Summary tag for docs please

Assets/MixedRealityToolkit/Extensions/BoundsExtensions.cs
```diff
@@ -354,54 +354,62 @@ public static void GetColliderBoundsPoints(GameObject target, List<Vector3> boun
...
+            {
+                CapsuleCollider cc = collider as CapsuleCollider;
+                Bounds capsuleBounds = new Bounds(cc.center, Vector3.zero);
+                switch (cc.direction)
                 {
-                    MeshCollider mc = colliders[i] as MeshCollider;
-                    Bounds meshBounds = mc.sharedMesh.bounds;
-                    meshBounds.GetCornerPositions(mc.transform, ref corners);
-                    boundsPoints.AddRange(corners);
+                    case 0:
```
Troy-Ferrell: can you make the numeric values private const ints at top or at the very least add a comment referring to which direction each value corresponds to?
https://docs.unity3d.com/ScriptReference/CapsuleCollider-direction.html

Really surprised Unity used regular ints instead of an enum

Assets/MixedRealityToolkit.Examples/Demos/UX/BoundingBox/Scripts/BoundingBoxTest.cs
```diff
@@ -192,20 +192,45 @@ private IEnumerator Sequence()
 
                 GameObject multiRoot = new GameObject();
                 multiRoot.name = "multiRoot";
-                multiRoot.transform.position = cubePosition;
+                Vector3 forwardOffset = Vector3.forward * .5f;
```
Troy-Ferrell: Can you actually rename this to BoundingBoxExampleTest or BoundingBoxExample? There is both a BoundingBoxTest.cs and a BoundingBoxTests.cs which is very confusing. The latter though I believe is the only one that goes through CI and our PR automation checking.

Also, I'm still going through all of this PR. But should there be any tests modified or added in BoundingBoxTests.cs for this change?

Troy-Ferrell:
Thanks for making the change! I'll did a quick pass and I'll look more critically later. FYI, I think there may be some overlap between this issue and this change:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4755

I haven't dug deep enough into this PR to look between the two but just FYI

Alexees:
> Thanks for making the change! I'll did a quick pass and I'll look more critically later. FYI, I think there may be some overlap between this issue and this change:
> #4755 

I know, I did the changes with exactly that issue in mind. Wouldn't have implemented the enum for choosing how to calculate bounds otherwise :)

Zee2:
Looks like these changes have caused a bug that destructively changes the proxy collider. #7997

Issue #4436, opened 2019-05-16
Add min/max scaling value field to the ManipulationHandler


cre8ivepark:
## Describe the problem
Currently, Bounding Box has min/max scaling values that developers can set. However, ManipulationHandler does not have these limits and the user can scale up/down without constraints. ManipulationHandler is frequently used together with the BoundingBox on the same object. 

For the content developer, there is no way to limit the scaling min/max for the two-handed manipulation.
There should be a way to limit min/max scale on ManipulationHandler(for the two handed-scaling) too.

## Describe the solution you'd like
Add min/max scaling value field to the ManipulationHandler

PR #5309, opened 2019-07-15
Add min/max scaling values to ManipulationHandler

MenelvagorMilsom:
## Overview
Repurpose min/max scaling in `BoundingBox` to use in the `ManipulationHandler`. Objects with a manipulation handler now have minimum and maximum scale values.

## Changes
- Fixes: #4436.
- Adds `TransformHelper`, intended to allow for sharing of utilities and transform functionality between `BoundingBox` and `ManipulationHandler`.
  - Implemented as a MonoBehaviour, ensured by both `BoundingBox` and `ManipulationHandler`.
  - Currently only implements shared scaling functionality.
- Change `BoundingBox` to use `TransformHelper`.
- Use min/max scaling and `TransformHelper` in `ManipulationHandler`
- Add tests for min/max scaling

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
@@ -148,46 +148,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
             set => constraintOnMovement = value;
         }
 
-        [SerializeField]
-        [Tooltip("Minimum scaling allowed relative to the initial size")]
-        private float scaleMinimum = 0.2f;
-
-        /// <summary>
-        /// Public property for the scale maximum, in the target's local scale.
-        /// Set this value with SetScaleLimits.
-        /// </summary>
-        public float ScaleMinimum
-        {
-            get
-            {
-                if (transformHelper != null)
-                {
-                    return transformHelper.ScaleMinimum;
-                }
-                return 0.0f;
-            }
-        }
-
-        [SerializeField]
-        [Tooltip("Maximum scaling allowed relative to the initial size")]
-        private float scaleMaximum = 2.0f;
-
-        /// <summary>
-        /// Public property for the scale maximum, in the target's local scale.
-        /// Set this value with SetScaleLimits.
-        /// </summary>
-        public float ScaleMaximum
-        {
-            get
-            {
-                if (transformHelper != null)
-                {
-                    return transformHelper.ScaleMaximum;
-                }
-                return 0.0f;
-            }
-        }
-
         [Header("Smoothing")]
         [SerializeField]
         [Tooltip("Check to enable frame-rate independent smoothing. ")]
@@ -281,7 +241,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
         private Quaternion startObjectRotationFlatCameraSpace;
         private Quaternion hostWorldRotationOnManipulationStart;
 
-        private TransformHelper transformHelper;
+        private TransformScaleHandler scaleHandler;
 
         #endregion
 
@@ -300,28 +260,9 @@ namespace Microsoft.MixedReality.Toolkit.UI
                 hostTransform = transform;
             }
 
-            transformHelper = this.EnsureComponent<TransformHelper>();
-            transformHelper.Initialize(hostTransform);
-            transformHelper.SetScaleLimits(scaleMinimum, scaleMaximum);
+            scaleHandler = this.GetComponent<TransformScaleHandler>();
         }
         #endregion MonoBehaviour Functions
-
-        #region Public Methods
-
-        /// <summary>
-        /// Sets the minimum/maximum scale for the manipulation handler at runtime.
-        /// </summary>
-        /// <param name="min">Minimum scale</param>
-        /// <param name="max">Maximum scale</param>
-        /// <param name="relativeToInitialState">If true the values will be multiplied by scale of target at startup. If false they will be in absolute local scale.</param>
-        public void SetScaleLimits(float min, float max, bool relativeToInitialState = true)
-        {
-            scaleMaximum = max;
-            scaleMinimum = min;
-            transformHelper.SetScaleLimits(min, max, relativeToInitialState);
-        }
-
-        #endregion Public Methods
 
         #region Private Methods
         private Vector3 GetPointersCentroid()
@@ -659,7 +600,10 @@ namespace Microsoft.MixedReality.Toolkit.UI
             // Currently the two hand rotation algorithm doesn't allow for lerping, but it should. Fix this.
             hostTransform.rotation = Quaternion.Lerp(hostTransform.rotation, targetRotationTwoHands, lerpAmount);
 
-            targetScale = transformHelper.ClampScale(targetScale);
+            if (scaleHandler != null)
+            {
+                targetScale = scaleHandler.ClampScale(targetScale);
+            }
             hostTransform.localScale = Vector3.Lerp(hostTransform.localScale, targetScale, lerpAmount);
         }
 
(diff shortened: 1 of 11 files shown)


Assets/MixedRealityToolkit/Utilities/Physics/TransformHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (targetTransform == null)
+            {
+                targetTransform = transform;
+                initialScale = transform.localScale;
+            }
+        }
+
+        #region Scale Utilities
+
+        public void SetScaleLimits(float min, float max, bool relativeToInitialState = true)
```
wiwei: >public void SetScaleLimits(float min, float max, bool relativeToInitialState = true) [](start = 8, length = 84)

Nit: Missing doccomments - probably just take the comments from:

/// <summary>
        /// Sets the minimum/maximum scale for the transform helper.
        /// </summary>
        /// <param name="min">Minimum scale</param>
        /// <param name="max">Maximum scale</param>
        /// <param name="relativeToInitialState">If true the values will be multiplied by scale of target at startup. If false they will be in absolute local scale.</param>

Assets/MixedRealityToolkit/Utilities/Physics/TransformHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+                minimumScale = initialScale * min;
+            }
+            else
+            {
+                maximumScale = new Vector3(max, max, max);
+                minimumScale = new Vector3(min, min, min);
+            }
+        }
+
+        public Vector3 ClampScale(Vector3 scale)
```
wiwei: >public Vector3 ClampScale(Vector3 scale) [](start = 8, length = 40)

Nit: Add a doccomment here as well.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1562,69 +1565,12 @@ private void CaptureInitialState()
             var target = Target;
             if (target != null)
             {
-                initialScaleAtStart = target.transform.localScale;
-
-                maximumScale = initialScaleAtStart * scaleMaximum;
-                minimumScale = initialScaleAtStart * scaleMinimum;
-                isChildOfTarget = transform.IsChildOf(target.transform);
```
wiwei: >isChildOfTarget [](start = 16, length = 15)

This line isn't getting set anymore, but still exists elsewhere - should this remain?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -150,24 +151,45 @@ public BoundingBoxActivationType BoundingBoxActivation
...
+        [Tooltip("Maximum scaling allowed relative to the initial size")]
+        private float scaleMaximum = 2.0f;
+
 
         /// <summary>
-        /// Public property for the scale maximum, in the target's local scale.
+        /// Public property for the scale minimum, in the target's local scale.
         /// Set this value with SetScaleLimits.
         /// </summary>
-        public float ScaleMaximum => maximumScale.x;
```
wiwei: >public float ScaleMaximum => maximumScale.x; [](start = 8, length = 44)

This change also makes it so it's no longer possible to set the scale min/maxes after the fact (is that the intention?)

Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
```diff
@@ -144,6 +144,54 @@ public IEnumerator ScaleViaNearInteration()
 
         }
 
+        /// <summary>
+        /// This tests the minimum and maximum scaling for the bounding box.
+        /// </summary>
+        /// <returns></returns>
```
wiwei: >/// <returns></returns> [](start = 8, length = 23)

Nit: Delete this line.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
MenelvagorMilsom: Setting the min/max values after the fact is still possible using `SetScaleLimits`.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
MenelvagorMilsom: Oops, I'll add that back.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -279,9 +321,30 @@ private void Start()
...
+        /// <summary>
+        /// Sets the minimum/maximum scale for the manipulation handler at runtime.
+        /// </summary>
+        /// <param name="min">Minimum scale</param>
+        /// <param name="max">Maximum scale</param>
+        /// <param name="relativeToInitialState">If true the values will be multiplied by scale of target at startup. If false they will be in absolute local scale.</param>
+        public void SetScaleLimits(float min, float max, bool relativeToInitialState = true)
+        {
+            scaleMaximum = max;
+            scaleMinimum = min;
```
thalbern: it feels a bit wrong to me that we set the scale min and max here but not retain the info about this being relative to the initial state. so if you would call start() again after setscalelimits has been called with relativeToInitialState is true , the scale would suddenly change because it's interpreting those values as relativeToInitialState the second time this is called.

Assets/MixedRealityToolkit/Utilities/Physics/TransformHelper.cs
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Physics
+{
+    /// <summary>
+    /// Helper class containing move/rotate/scale related utility functions.
```
thalbern: I don't like that we're storing those min max values in two different classes. Would it be an option to remove them here and instead change SetScaleLimits to AdjustScaleLimits(..) which does the initialscale multiplication but returns that value instead of storing it and ClampScale would then take the max scale values as second parameter ? that way you keep the data in one place and it doesn't diverge

MenelvagorMilsom: Ah, there is one issue worth discussing here. If we do things like this, if you have both a `ManipulationHandler` and a `BoundingBox` on an object, the object will have two different min/max scales.

Say for example someone had MH max scale as 3 and BB max scale as 5. If they scaled to max using the BB scale handles and used MH two hand manipulation, the scale would jump down to three.

MenelvagorMilsom: Previously there was conflict as to which min/max scale would be taken, but at least there was one authoritative min/max scale.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/TransformScaleHandler.cs
```diff
@@ -0,0 +1,175 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// Helper class containing move/rotate/scale related utility functions.
```
thalbern: Please adjust this comment. Its not a helper class anymore. We might want to add this component to the docs of manipulation handler or bounding box as well. And make sure its used in one of the example scenes so ppl see how to use it there as well

Issue #4941, opened 2019-06-18
Popup for enabling XR setting in new unity project doesn't check platform


thalbern:
## Describe the bug

Whenever someone creates a new project with MRTK we're displaying a popup telling the user to enable the XR settings, however we're not checking which platform is currently selected. So if the user is eg. on windows platform and switches over to UWP after that popup was confirmed, the setting won't be enabled in the UWP build.

## To reproduce

- create new unity project (make sure your platform is not set to UWP)
- import mrtk packages
- observe popup
- switch platform to UWP
- observe missing XR setting

# Platform
- Unity Version 2018.4
- MRTK Version v2 RC2.1

![image](https://user-images.githubusercontent.com/36998103/59691575-82d34b80-91e3-11e9-8482-55e68fae3ce0.png)

Troy-Ferrell:
My PR doesn't enable XR support for every build platform. But at least users will be notified if they do switch build target and the optimize window tool also notifies as well

thalbern:
I don't get a notification if I switch platform. Have been running into this several times now (ending up with a flat app on HL2)

Troy-Ferrell:
It's showing up for me. Can you check again latest?

![image](https://user-images.githubusercontent.com/25975362/63188680-d82aae00-c016-11e9-9d11-19957e3862ec.png)

thalbern:
Ah! I was waiting for a popup like the first time you import mrtk. I'll check if that console output is there.

thalbern:
checked with stabilization branch from friday and can confirm I have that popup now - did the original test with one of our prebuilt packages, so checking latest release package later today as well.

Troy-Ferrell:
Yea we switched to that to be a bit less intrusive to the developer.

PR #5319, opened 2019-07-16
Improve configuration checking by MRTK

Troy-Ferrell:
## Overview
This change improves notifying the user of non-recommended settings currently configured in the developer's unity project. In particular the focus is on Single Pass Instanced rendering which has been shown to have tremendous performance increases for XR projects and should be pushed as the default configuration.
https://blogs.unity3d.com/2017/11/21/how-to-maximize-ar-and-vr-performance-with-advanced-stereo-rendering/

- Added listener to MixedRealityEditorSettings to check for when play mode is entered. When this happens, the system will check if certain configurations have not been set and alert user via warnings (see screenshot). This warning & check logic is only in editor

- Added preference for user to decide whether to run configuration check or not (see screenshot)

- Updated the settings prompt to also check for single pass instanced to be default rendering path if user accepts. I did not add enable depth buffer sharing because this should be default for new unity projects. Furthermore, enabling depth buffer sharing can have problematic outcomes on Hololens if data is not written to depth compared to single pass instancing (which only affects custom shaders potentially). 

- Fixed MixedRealityEditorSettings to use correct ignore prompt preferences key. Item was duplicated and broken between the preferences window and if the user clicked "ignore" button the prompt. The keys were not matching. I have updated the preference window (MixedRealityPreferences.cs) to use "_MixedRealityToolkit_Editor_IgnoreSettingsPrompts" with the underscore since it is more likely users have hit the ignore button via the prompt then gone into preferences project settings window. 

- Also moved MixedRealityPreferences to utilities/editor package so it can be utilized across project and thus had to make class public

image

image

## Changes
- Fixes: #4941 , #2860
Related: #1939 

## Verification
Enter play mode with various settings off and notice warnings logged
Turn off preference in project settings, note warnings are not filed on re-run of above


jbienzms:
I guess the only issue I have with this approach is that currently there are so many warnings in MRTK projects I see developers turning warnings off and only focusing on errors. I almost feel like not having shared depth buffer turned on should be an error instead of a warning, but I do understand there are some situations where this is actually desired. I just wish that these warnings could be surfaced in a more visual way. For example, I wish they could be surfaced similar to your performance analysis tool rather than just warnings in the console window.

jbienzms:
Looks like we still need a check to warn if the audio spatializer isn't set to Microsoft HRTF?

jbienzms:
Looks like we still need a check to warn if the audio spatializer isn't set to Microsoft HRTF?

Troy-Ferrell:
@jbienzms , yes I struggled with deciding how best to surface these suggestions. It's kind of like the spatial mapping system default state. Some wanted it on by default while others hated it being on by default. Errors will be much more prominent but then a developer who wants to ignore this will have to go to the preferences and check "not run" effectively. I tried adding the bold rich text to make the warnings stand out. I also considered a prompt dialog...but again some developers will likely get extremely pissed. 

Depth buffer sharing is slightly lower than single pass instanced rendering. The problem with depth LSR is it requires the developer to submit a valid depth buffer which means work and understanding on their side. Whereas SPI only affects you if you have custom shaders which is more rare, and if you are writing shaders yourself, you likely know about SPI. 

I actually missed the HRTF spatializer setting. That may be harder. I need to check if I can set the scriptable object like I do lighting settings. Is there issues if people have the Oculus spatializer set basically?

jbienzms:
> @jbienzms , yes I struggled with deciding how best to surface these suggestions.

Yeah, this is a challenging one. Especially since some are recommended but optional. That's what I miss about the old HTK "Configure my project" wizard.

> I actually missed the HRTF spatializer setting. That may be harder. I need to check if I can set the scriptable object like I do lighting settings. Is there issues if people have the Oculus spatializer set basically?
> 
> **EDIT:**
> Added log warning check for MS HRTF audio spatializer only when build target platform is UWP since MRTK developers may be building for non-Microsoft platforms

I think this is a good approach. Technically we will pretty much always want to use Microsoft HRTF on HoloLens and ideally use it on WMR immersive. But if they're targeting another platform, or even if they're targeting Vive or Rift on PC they could be using a different spatializer. So ideally I would only warn when targeting HoloLens, but I don't think we can know that in Unity. So probably just warning for UWP is best for now.

Troy-Ferrell:
Updated PR to point to prerelease/2.0.0_stabilization

PR #5328, opened 2019-07-16
adding link to the dependency window doc

Yoyozilla:
## Changes
- Fixes: # .

### DIFF
--- a/Documentation/toc.yml
+++ b/Documentation/toc.yml
@@ -141,7 +141,7 @@
     - name: Build Window
       href: Utilities/BuildWindow.md
     - name: Dependency Window
-      href: DependencyWindow.md
+      href: Utilities/DependencyWindow.md
     - name: Optimize Window
       href: Utilities/OptimizeWindow.md
     - name: Extension Service Creation Wizard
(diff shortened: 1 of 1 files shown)


Documentation/toc.yml
```diff
@@ -140,6 +140,8 @@
     items:
     - name: Build Window
       href: Utilities/BuildWindow.md
+    - name: Build Window
+      href: DependencyWindow.md
```
Cameron-Micka: I'm not too familiar with the "toc" file. Does this line need to be in the utilities namespace?
  
`href: Utilities/DependencyWindow.md`

Documentation/toc.yml
```diff
@@ -140,6 +140,8 @@
     items:
     - name: Build Window
       href: Utilities/BuildWindow.md
+    - name: Build Window
```
Cameron-Micka: Should this be?

`- name: Dependency Window`

Documentation/toc.yml
Yoyozilla: Checking..

Documentation/toc.yml
Yoyozilla: fixed

Issue #4917, opened 2019-06-17
change UI raycast to allow customization similar to IMixedRealityRaycastProvider for physics


naostranMS:
## Describe the problem
Originally in HTK, the UI raycast used distance to pick which UI element to focus on, but that was changed to work better with coplanar UI in this change:
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/1101

In our case the UI is not coplanar, and it's awkward to have to reorder the UI to make raycasts work as expected, so prioritizing distance over depth makes more sense.

## Describe the solution you'd like

Reorder the UI raycast prioritization to put distance before depth, and change the distance check to bypass (and move on to depth test) if the delta is less than some epsilon. With the epsilon check, I think it would still work the same way for coplanar UI, while also allowing it to use the distance check first for non-coplanar UI. (I'm also open to suggestions for better ways to do the coplanar test, but the epsilon seems like it should work and is simple enough)

## Describe alternatives you've considered

Another solution would be to make the UI raycast prioritization overridable, and then we could just override the prioritization locally to do distance before depth.

Railboy:
Unless I've misunderstood the problem, this sounds like something a [custom raycast provider class](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4678) could fix.

naostranMS:
@Railboy The custom raycaster is really tailored toward physics raycast and not the UI raycast. I could theoretically make it work that way but there are a couple of things that would make it pretty awkward:
- the custom raycaster returns a MixedRealityRaycastHit which is based on the physics RaycastHit
- it still does RaycastGraphics after the configurable raycast, which might just end up overriding the result anyway

Admittedly, I had meant to add this feedback to that PR but took too long:
1. I wish the default physics raycaster allowed for overriding prioritization without having to override all of the raycast behavior (and that would apply to using the configurable raycaster here as well - I really just want to override prioritization).
2. I think it could be beneficial to have a more generalized version of the customized raycasters and raycast prioritization/comparators that would encompass UI raycasts as well.

naostranMS:
Another part of it is just that I think the default behavior would be more intuitive if it more closely mimicked draw behavior. That is, it seems like the expected behavior would be that the focus provider will pick whatever is drawn on top, but it doesn't work that way if the UI isn't coplanar.

Railboy:
>raycast prioritization/comparators that would encompass UI raycasts as well.

[...]

PR #5330, opened 2019-07-17
Update the IMixedRealityRaycastProvider to support custom graphics raycast functionality

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4917

This change adds an extension point to the IMixedRealityRaycastProvider to allow for customizing how the GraphicsRaycast works. This allows for clients to customize the priority order of raycast results.

The goal here was to make it possible for clients to change the order of RaycastResultComparer::Comparers, so the changes involved:

1) Adding the GraphicsRaycast function to IMixedRealityRaycastProvider.
2) Updating RaycastResultComparer's Comparers comparison functions so that they can be referenced in subclasses.
3) Updating the default implementation to do exactly what it does today.

Note that in theory it would be possible to accomplish this by just doing #1 (this would require callers to do a bunch of duplicated work, so I opted to do #2-#3 as well)


wiwei:
If this looks good for folks, I will also update the docs to explain how this could be used to accomplish the layer reordering

Kjakubzak:
Looks good so far. Will you also be changing what the FocusProvider calls for graphics raycasts (similar to the last change where direct calls to the MixedRealityRaycaster became calls to the RaycastProvider)?

wiwei:
> Looks good so far. Will you also be changing what the FocusProvider calls for graphics raycasts (similar to the last change where direct calls to the MixedRealityRaycaster became calls to the RaycastProvider)?

Doh, I missed staging that file in the last commit. I've updated it now, please take another look!

naostranMS:
Yeah, this will work for Layout. Thanks for doing the extra work to make it easier to just override the prioritization!

wiwei:
@lukastoenneMS @keveleigh can you guys take a look at this one?

Assets/MixedRealityToolkit/Interfaces/InputSystem/IMixedRealityRaycastProvider.cs
```diff
@@ -27,5 +28,11 @@ public interface IMixedRealityRaycastProvider : IMixedRealityService
         /// </summary>
         /// <returns>Whether or not the SphereCast hit something.</returns>
         bool SphereCast(RayStep step, float radius, LayerMask[] prioritizedLayerMasks, out MixedRealityRaycastHit hitInfo);
+
+        /// <summary>
+        /// Performs a graphics raycast against the specified layerMasks.
+        /// </summary>
+        /// <returns>The RaycastResult of the raycast.</returns>
+        RaycastResult GraphicsRaycast(EventSystem eventSystem, PointerEventData pointerEventData, LayerMask[] layerMasks);
```
keveleigh: Would we want to separate this out into an `IMixedRealityGraphicsRaycastProvider` instead?

wiwei: What are you thoughts on the pros/cons of this? I was thinking about this and I can see a world in which maybe someone only wants to implement one of these, though I was hoping to reduce interface concept overload by saying:

"If you want to have custom raycast behavior, you should implement the raycast behavior interface." For folks who want to implement just one of those, you can always take the base class itself.

keveleigh: That was mainly my thought, about only wanting to implement one. Worst case, they can copy over the existing implementation, which is just a single line anyway

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/d097f5b87365296b936b364da36630643e118fbd/Assets/MixedRealityToolkit.Services/InputSystem/DefaultRaycastProvider.cs#L39

Issue #5240, opened 2019-07-09
Hand needs to be way too close to grabbable before hand ray disappears


julenka:
## Overview
The hand ray currently only turns off when the hand is incredibly close to a grabbable, when the hand is ~1cm near a grabbable. The ray should turn off when the hand is ~10cm near a grabbable.

This should be implemented by having a separate radius for turning off the line ray, separate from the radius that is used for doing a query for determining what to grab.

Currently, SpherePointer.IsNearObject contains the following:

```
return UnityEngine.Physics.CheckSphere(position, SphereCastRadius, ~UnityEngine.Physics.IgnoreRaycastLayer);
```

Instead of `SphereCastRadius` (used when determining grab start/edn), we should use a different radius `HandRayOffRadius`.

## Expected Behavior
The ray should turn off when hand is ~10cm near a grabbable
![image](https://user-images.githubusercontent.com/168492/60873303-2064ec80-a271-11e9-8c6a-66f69d218f65.png)

## Actual Behavior
Ray is still on, even when hand is very close to grabbable
![image](https://user-images.githubusercontent.com/168492/60873140-d24fe900-a270-11e9-8cb6-0066a3992c7a.png)

## Steps to reproduce
_(Links to sample github project preferred)_

keveleigh:
There was originally a 0.05 buffer between the grab radius and the hand ray disappearing, but it looks like that was removed in https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4974 via https://github.com/microsoft/MixedRealityToolkit-Unity/commit/b8160ca9de48c104f474b9d9141cda71a90a65c6#diff-2be5244793f0a456321d940ebafd4e55.

In addition, the grab radius was reduced to a third of its previous size in https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4898.

Those two combined are likely contributing to this issue.

lukastoenneMS:
Should this behavior actually be controlled by grabbables? Because it sounds to me like disabling the far interaction pointer should be a feature of that pointer, instead of relying on a large grabbable radius to disable a different kind of pointer. So

1. When line pointer hit is closer than X, the pointer is hidden and interaction disabled. Pointer keeps raycasting though, we want to un-hide when moving further away.
1. Grabbable distance is a separate issue, should the previous radius of 0.15 be restored? It does sound a little bit excessive and if it was just to hide the line pointer then the 1st point would take care of that.

[...]

PR #5339, opened 2019-07-17
Add a secondary radius to the SpherePointer to disable far interaction earlier

lukastoenneMS:
## Overview

There is supposed to be a "dead zone" between disabling the far interaction and enabling (grabbable) near interaction with the SpherePointer. Previously there was a hardcoded margin in the IsNearObject function which has been removed, as well as a significantly larger base SphereCastRadius.

Introducing a configurable NearObjectMargin property adds back the dead zone, while keeping the SphereCastRadius at the desired smaller value (0.05m).

Also added a test to make sure the distance values work as expected.

## Changes
- Fixes: #5240 .


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
```diff
@@ -13,6 +13,26 @@ public class SpherePointer : BaseControllerPointer, IMixedRealityNearPointer
...
+        [Min(0.0f)]
+        [Tooltip("Additional distance between SphereCastRadius and NearObjectRadius")]
+        private float nearObjectMargin = 0.2f;
+        /// <summary>
+        /// Additional distance between <see cref="BaseControllerPointer.SphereCastRadius"/> and <see cref="NearObjectRadius"/>.
+        /// </summary>
+        /// <remarks>
+        /// This creates a dead zone in which far interaction is disabled before objects become grabbable.
+        /// </remarks>
+        public float NearObjectMargin => nearObjectMargin;
```
keveleigh: Does this (and `NearObjectRadius`) need to be public? They seem like values only used internally in this script.
Making things public for tests feels strange to me.

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpherePointerTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        // Keeping this low by default so the test runs fast. Increase it to be able to see hand movements in the editor.
+        private const int numFramesPerMove = 3;
+
+        private float colliderSurfaceZ;
+
+        // Initializes MRTK, instantiates the test content prefab and adds a pointer handler to the test collider
+        [SetUp]
+        public void SetUp()
+        {
+            TestUtilities.InitializeMixedRealityToolkit(true);
```
julenka: Recommend using `PlayModeTestUtilities.Setup`, see [SlateTests](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/a5359f5ed4dacfd812b7bc983285693f96bc0c55/Assets/MixedRealityToolkit.Tests/PlayModeTests/SlateTests.cs#L31)

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpherePointerTests.cs
```diff
@@ -0,0 +1,110 @@
...
+using Microsoft.MixedReality.Toolkit.Utilities;
+using Microsoft.MixedReality.Toolkit.Input;
+
+namespace Microsoft.MixedReality.Toolkit.Tests
+{
+    // Tests to verify sphere pointer distances
+    public class SpherePointerTests
+    {
+        // Keeping this low by default so the test runs fast. Increase it to be able to see hand movements in the editor.
+        private const int numFramesPerMove = 3;
```
julenka: We should do this for all tests, good idea.

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpherePointerTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            colliderSurfaceZ = centerZ - scale * 0.5f;
+
+            var collider = cube.GetComponentInChildren<Collider>();
+            Assert.IsNotNull(collider);
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            TestUtilities.ShutdownMixedRealityToolkit();
```
julenka: Recommend PlayModeTests.TearDown

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
julenka: I could thing of an example: what if you wanted to show an object *just out of reach*? So, at the point where you still have a hand ray but if you move just a bit closer the ray disappears. How would a consumer determine that distance without this var?

lukastoenneMS: I like to have at least a public getter for serializable fields. It seems wrong to be able to edit something in the inspector but having no way to at least _read_ that value at runtime.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
```diff
@@ -29,17 +49,31 @@ public bool IsNearObject
...
+        {
+            get
+            {
+                if (IsFocusLocked)
+                {
+                    return true;
+                }
+                else if (base.IsInteractionEnabled && TryGetNearGraspPoint(out Vector3 position))
+                {
+                    return UnityEngine.Physics.CheckSphere(position, SphereCastRadius, ~UnityEngine.Physics.IgnoreRaycastLayer);
```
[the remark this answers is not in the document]
lukastoenneMS: I'm just reproducing the previous behavior, which also did not include a check for specific components beyond the collider. I'd rather not make a potentially breaking change at this point.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
lukastoenneMS: It's an aspect of Unity i'm really not fond of: this fact that you have more access to settings in the editor by default than through the API. At least that's how our access policy with private fields plays out in practice.

Assets/MixedRealityToolkit.Tests/PlayModeTests/SpherePointerTests.cs
lukastoenneMS: Yes, and preferably all tests should use a shared value (right now there are still a few tests using 30 or so steps). I'll make a separate PR for that.

lukastoenneMS: #5358

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
keveleigh: Works for me! Thanks for the additional context.

Issue #5342, opened 2019-07-17
UnityTouchDeviceManager does not raise InputActionEvents


jbienzms:
## Describe the bug

In #4006 we introduced the ability for Input Actions to be handled in a source-agnostic way. This allows an application developer to handle Input Actions in a single handler regardless of whether the action was triggered by Voice, Controller, Hands, etc.

The TouchScreen controller can be configured to handle Input Actions. Those actions are never received by classes registered to receive **IMixedRealityInputActionHandler** events.

![image](https://user-images.githubusercontent.com/5230670/61393720-32551980-a887-11e9-8390-9725c8ee7298.png)

## To reproduce

1. Create a global handler that implements **IMixedRealityInputActionHandler**
2. Touch the screen

## Expected behavior

The **IMixedRealityInputActionHandler** should receive the Select input action but it does not.

## Your Setup (please complete the following information)

- Unity Version 2018.4.4
- MRTK Version mrtk_develop as of 7/15

## Target Platform (please complete the following information)

- Unity Editor
- Android

wiwei:
@davidkline-ms This I think goes under better support for mobile. See @keveleigh's comments in the PR https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5343

davidkline-ms:
Confirmed #5903 resolves this issue

PR #5343, opened 2019-07-17
Added ability for Pointers to raise IMixedRealityInputActionHandler events

jbienzms:
# Overview

This PR addresses #5342.

## Changes
- Created OnPointerDownWithActionEventHandler and OnPointerUpWithActionEventHandler
- Modified RaisePointerDown and RaisePointerUp to call the methods above if an input action is present

## Verification
I am unclear on the need for HandlePointerEvent vs just HandleEvent. HandlePointerEvent was specifically requiring IMixedRealityPointerHandler but did not do anything with that specific interface. The implementation of HandlePointerEvent looks nearly identical to HandleEvent. I did leave HandlePointerEvent in there but just changed it to accept IEventSystemHandler as this was required in order to be able to accept an IMixedRealityInputActionHandler. Everything seems to be functioning, but please validate this approach. It would be good to validate if we even need a separate HandlePointerEvent, and if not just collapse it into HandleEvent.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
@@ -114,10 +114,10 @@ namespace Microsoft.MixedReality.Toolkit.Input
             {
                 IMixedRealityCapabilityCheck capabilityChecker = deviceManagers[i] as IMixedRealityCapabilityCheck;
 
-                // If one of the running data providers supports the requested capability, 
+                // If one of the running data providers supports the requested capability,
                 // the application has the needed support to leverage the desired functionality.
                 if ((capabilityChecker != null) &&
-                    capabilityChecker.CheckCapability(capability))
+                        capabilityChecker.CheckCapability(capability))
                 {
                     return true;
                 }
@@ -241,9 +241,9 @@ namespace Microsoft.MixedReality.Toolkit.Input
                     object[] args = { Registrar, this, configuration.ComponentName, configuration.Priority, configuration.DeviceManagerProfile };
 
                     if (Registrar.RegisterDataProvider<IMixedRealityInputDeviceManager>(
-                        configuration.ComponentType.Type,
-                        configuration.RuntimePlatform,
-                        args))
+                                configuration.ComponentType.Type,
+                                configuration.RuntimePlatform,
+                                args))
                     {
                         deviceManagers.Add(Registrar.GetDataProvider<IMixedRealityInputDeviceManager>(configuration.ComponentName));
                     }
@@ -478,7 +478,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// Assumption: We only send pointer events to the objects that pointers are focusing, except for global event listeners (which listen to everything)
         /// In contract, all other events get sent to all other pointers attached to a given input source
         /// </summary>
-        private void HandlePointerEvent<T>(BaseEventData eventData, ExecuteEvents.EventFunction<T> eventHandler) where T : IMixedRealityPointerHandler
+        private void HandlePointerEvent<T>(BaseEventData eventData, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
         {
             if (disabledRefCount > 0)
             {
@@ -549,7 +549,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// Returns true if the event was handled by a modal handler
         /// </summary>
         private bool DispatchEventToObjectFocusedByPointer<T>(IMixedRealityPointer mixedRealityPointer, BaseInputEventData baseInputEventData,
-            bool modalEventHandled, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
+                bool modalEventHandled, ExecuteEvents.EventFunction<T> eventHandler) where T : IEventSystemHandler
         {
             GameObject focusedObject = FocusProvider?.GetFocusedObject(mixedRealityPointer);
 
@@ -636,7 +636,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         }
 
         /// <summary>
-        /// Pop disabled input state. When the last disabled state is 
+        /// Pop disabled input state. When the last disabled state is
         /// popped off the stack input will be re-enabled.
         /// </summary>
         public void PopInputDisable()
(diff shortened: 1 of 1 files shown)


keveleigh:
Will this not fire duplicate events for pointers that fire because of input events / actions?

Also, there seems to be a lot of whitespace changes, so it's possible I missed some of the actual changes. Would it be possible to revert those?
image

jbienzms:
Hey everyone. Sorry I haven't had a chance to update this. @keveleigh @wiwei or @davidkline-ms , can you help me resolve the current conflict at lines 241 - 254? What was changed here? It looks like the code was wrapped in an If block, and if successful the input system is added as a data provider? I don't want to make this change blindly without understanding what's going on.

Is the only other thing blocking this from merge right now the tabs in the file? I'd like to get this merged ASAP before other changes happen and get it out of sync again.

Thanks!

keveleigh:
@jbienzms Also my question above:

>Will this not fire duplicate events for pointers that fire because of input events / actions?

From my read, it looks like we'll fire the IMixedRealityInputActionHandler for both the input event and the pointer event, which likely have the same action and will appear as a duplicate.

jbienzms:
Hey @keveleigh , just wanted to follow up on this.

> Also, there seems to be a lot of whitespace changes, so it's possible I missed some of the actual changes. Would it be possible to revert those?

_Personally_ I prefer the whitespace changes because it indents the parameters to the child level of the method rather than having them at the same level as the method definition. However (and more importantly) this branch was created by the customer on their account. I could ask the customer to make this change if necessary, but in the meantime I was hoping you could maybe use this feature when reviewing the code to filter out whitespace changes:

image

> From my read, it looks like we'll fire the `IMixedRealityInputActionHandler` for both the input event and the pointer event, which likely have the same action and will appear as a duplicate.

I'm trying to better understand this issue. I did in fact copy the current pattern implemented by InputDown and InputUp. Namely, PointerDown and PointerUp now call `OnPointerDownWithActionEventHandler` and `OnPointerUpWithActionEventHandler` respectively.

Are all pointer events mirrored by input events? And wouldn't this also mean that Gesture events would currently be sending duplicates as well, since Gesture events are considered input events and also raise their own input actions? (`OnGestureStartedWithAction` and `OnGestureCompletedWithAction`)

We made this pull request because we noticed that on both Android and in the Unity editor, touch screen events never triggered the expected `IMixedRealityInputActionHandler`. If you're saying that touch screen events are expected to trigger InputEvents and those InputEvents should in turn trigger `IMixedRealityInputActionHandler`, then maybe we need to look at why that's not behaving as expected. It wasn't until we implemented the changes in this PR that we got `IMixedRealityInputActionHandler` events to trigger for touch screen presses.

keveleigh:
@jbienzms Thanks for the tip about hiding whitespace in reviews!

>Are all pointer events mirrored by input events? 

Right, the vast majority of pointers listen for OnInputUp and OnInputDown in order to fire their own pointer events for their specified input actions.

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/f3a97916629a965c9fc45e85290c453ebc9b45f3/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs#L456

It seems like certain pointers / controllers (like this case, the touch controller) are bypassing that, causing this asymmetry.

>If you're saying that touch screen events are expected to trigger InputEvents and those InputEvents should in turn trigger IMixedRealityInputActionHandler, then maybe we need to look at why that's not behaving as expected. 

Yep, that's what I'm thinking!

>And wouldn't this also mean that Gesture events would currently be sending duplicates as well, since Gesture events are considered input events and also raise their own input actions?

I don't believe so, since Gesture events are triggered directly on their own without a corresponding Input event.

jbienzms:
> I don't believe so, since Gesture events are triggered directly on their own without a corresponding Input event.

Interesting. Any idea why then we're not seeing **IMixedRealityInputActionHandler** triggered for touch events as mentioned in #5342?

julenka:
I agree with @keveleigh here:

> If you're saying that touch screen events are expected to trigger InputEvents and those InputEvents should in turn trigger IMixedRealityInputActionHandler, then maybe we need to look at why that's not behaving as expected.

> > Yep, that's what I'm thinking!

The end goal of this PR and of issue #5342 is to get the touchscreen to raise OnInputDown and OnInputUp events, and it looks like the UnityTouchDeviceManager is behaving not like expected.

We expect that  touch device manager (TDM) should be  raising OnInputDown and OnInputUp events, which the touch pointers should be listening to. Instead, TDM is raising OnPointer* events.

@jbienzms  would you be okay if we closed this PR in favor of a separate PR that fixes the root issue?

@jbienzms Would you also be okay if we changed the title of #5342 to reflect the root cause?

As @keveleigh points out (and I think @luis-valverde-ms would agree also), pointers railing InputDown events would actually create duplicate events that we do not want. Input sources raise input events (handled by IMRInputActionHandler), pointers raise pointer events (handled by IMRPointerHandler).

julenka:
> I am unclear on the need for HandlePointerEvent vs just HandleEvent.

Ah yes, it's not immediately obvious and in fact originally everything went through handleevent, which caused a very subtle bug. The bug was fixed when we were developing internally at MSFT so I can't reference the PR here. The issue was that pointers would dispatch events to all other pointers. 

It's a little subtle but this is what HandleEvent does:
1. Dispatch to global handlers
2. **For every pointer attached to the input source, dispatch the event to the object focused by the pointer**. 
3. Dispatch to fallback handler

And this is what HandlePointerEvent does:
1. Dispatch to global handlers
2. Dispatch the event to the object focused by the pointer. 
3. Dispatch to fallback handler

Without the difference, you would have things like the grab pointer dispatching OnPointerDown when you touched something. So then all of your handlers would get ~3 pointer down, updated, and up events every time a pointer did something.

Here's a copy / paste of the internal message with the bug fix:

## Explanation of the Bug
Currently MixedRealityInputSystem dispatches any event `E` to all pointers attached to `E.InputSource`. Therefore, when a Pointer wants to send a PointerDown event to the object it is focusing, instead the following happens when trying to dispatch pointer event `E` from pointer `P`:

1. Initialize `E` with information about pointer `P` so that `eventData.Pointer` returns P
1. Get the InputSource `S` attached to pointer `P`
2. For all pointers `P2` on `S`...
  a. If `P2` is focused on some object `O`
  b. Dispatch pointer event `E` to `O`, even if `P2` ~= `P`.

This causes two problems when you have multiple pointers on an input source, for example a `GrabPointer`, `PokePointer`, and `LinePointer` as we have on fully articulated hands:

Each `PointerDown` event will get sent N times, where N = # pointers attached to a source.
If a `LinePointer` is focused on something, but `GrabPointer` is not, then in step 1, `E.Pointer` is set to something that doesn't actually have object focus (`E.Pointer.Result.Object == null`). Then, if something tried to access `E.Pointer.Result.Object` it will get null pointer. This violates an assumption that for any PointerEvent `E2` being dispatched, `E2` is focusing some object.

## Explanation of the Fix
When dispatching a pointer event, this event should only be sent to global event listeners and the object being focused by the pointer, NOT all other pointers attached `pointer.InputSource`.

Create a new method `HandlePointerEvent` that should be used to dispatch all events for pointers (which are the only things that can provide object focus). This method only dispatched events to global handlers and the object being focused, not to other pointers on the input source.

jbienzms:
> jbienzms would you be okay if we closed this PR in favor of a separate PR that fixes the root issue?

Yes, I would be OK with that.

> jbienzms Would you also be okay if we changed the title of #5342 to reflect the root cause?

Yes, for sure. I wasn't clear on the chain of events here, so if the proper fix is a different approach that's perfectly OK.

> The bug was fixed when we were developing internally at MSFT so I can't reference the PR here.

That was very helpful information. Unfortunately I don't think I fully understood it all, but the key point is that there is a reason for it and we have a path to fixing this issue.

Thank you @julenka and @keveleigh for taking the time to research this issue and determine the proper fix.

julenka:
Closing this PR in favor of fixing root cause reflected in #5342 (UnityTouchDeviceManager does not raise input actions)

PR #5345, opened 2019-07-17
Update the unitypackage.ps1 to actually build the examples package.

wiwei:
I was looking at the output from the latest CI runs, and while there was the presence of the Foundation unity package, the examples package was missing:

It turns out, this exit statement was too aggressive, and was leading to us only building the Foundation package. Now we will only early exit in the case of failure.


wiwei:
Waiting for this run:

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=3412

Issue #5353, opened 2019-07-17
There is no easy way to run tests in batch mode from command line


julenka:
## Overview
Running the unity tests in batch mode from command line is possible, but running the command and parsing output is not straightforward.

There should be a script people can run that will run the tests and give them test result output that's easy to read.

julenka:
Merged.

Issue #5317, opened 2019-07-16
MRTK tests documentation is hard to find on bing and google


julenka:
## Overview
The following [bing query](https://www.bing.com/search?q=mrtk+tests&form=WNSGPH&qs=SW&cvid=89cb823109e44107a7a725be2d4091d6&pq=mrtk+tests&cc=US&setlang=en-US&nclid=749F70F130801DE116EE8D6C9A5B5051&ts=1563235959490&wsso=Moderate) does not point to the mrtk tests page on github and neither does this [google query ](https://www.google.com/search?q=mrtk+tests)

This task covers asking Google to re-index our site as well as adding common keywords to docs so that the page is easier to find.

wiwei:
After @thalbern re-requested indexing, this has gotten a lot better. (i.e. for both, mrtk tests is now the first thing)

We can open a new doc bug if there are additional keywords that should be leading users here - note that we should be wary of spending too much time doing SEO stuff, as that's not where we will give consumers the most value.

PR #5354, opened 2019-07-18
Add script to run playmode tests from powershell

julenka:
## Overview
Sometimes tests that pass in the Unity editor fail when run from console (in headless, batch mode). Finding the command to run this is possible, but not necessarily easy. Provide a script that runs the playmode tests locally, parses results, and provides a simple output (number passed, failed tests, and list failed tests).

## Changes
- Fixes: #5353
- Fixes: #5317 
- Add docs for how to run the test
- Change title of tests page to hopefully make the page easier to find from search engine.


wiwei:
Yasss

julenka:
> I wonder if there would be an easy way to get the exact same script runnable here as the script we run on the pipeline, so we don't potentially have the same / similar code in two places.

@MenelvagorMilsom is this script acceptable to run on CI? Not sure...it does stuff like open up a new window...It does print some basic test results which is nice but CI already does this.

julenka:
> Weird that the script is showing up as binary in the PR (clicking through show the script just fine). LGTM though!

@keveleigh argh I saved the file as utf-16 for some reason. Ignored it yesterday because I was tired. Fixed it now, saved at utf-8.

MenelvagorMilsom:
> @MenelvagorMilsom is this script acceptable to run on CI? Not sure...it does stuff like open up a new window...It does print some basic test results which is nice but CI already does this.

We're already doing something similar [here](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/pipelines/templates/tests.yml). On CI we write our results to the console continuously so that we can observe progress. Also we may be able to run this script on CI easier if we could specify a unity install folder and let the script find the .exe.

Otherwise these scripts are very similar, we should be able make a few changes and use this on CI.

Issue #5114, opened 2019-06-28
MSBuild not found during build scripts


MenelvagorMilsom:
## Describe the bug
A partner at the Berlin Hack tried to extend BuildAppx method to build for HL2. However, BuildAppxAsync throws an error when trying to find MSBuild.exe:
![msbuildfail](https://user-images.githubusercontent.com/47415945/60334198-346c3c80-999b-11e9-9df5-1c63363cb1e9.png) 

The problem is probably that the methods expects to find the MSBuild.exe in a directory that ends with _\MSBuild\15.0\Bin\MSBuild.exe_ but MSBuild cannot be found at this locaton for there VS Installation.

## Expected Behaviour
The build should find MSBuild independent of where it is located in your VS installation.

wiwei:
Resolved by https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5415, which was submitted on behalf of @julianloehr-kg

PR #5356, opened 2019-07-18
Let vswhere find the path for msbuild

julianloehr-kg:
## Overview
 
Use the `vswhere` to find the MSBuild path (as recommended by themselves: https://github.com/microsoft/vswhere/wiki/Find-MSBuild), instead of trying to build the path ourselves.

## Changes
- Fixes: #5114.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
@@ -171,7 +171,7 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
                     UseShellExecute = false,
                     RedirectStandardOutput = true,
                     RedirectStandardError = true,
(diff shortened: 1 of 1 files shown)


wiwei:
This is awesome!

wiwei:
Looks like this causes our CI to fail - I'll need to debug a bit to see what's going on here

wiwei:
Okay, figured this one out, it looks like this is an issue where if we have 2017 installed, the vswhere.exe that comes with it actually doesn't have this new -find feature (this -find feature was added in Jan-ish of 2019, which is obviously two years ahead of VS2017).

I think that we need to be able to support both at least for now - what we could do here instead if try once with the new thing, and if that fails, we can fall back to the current way - this way we can preserve support for folks who are using 2017 (many people right now) while still supporting the 2019 (increasingly more people)

wiwei:
Cool, I took this change and updated it to support both VS2017's version of vswhere and VS2019's version of vswhere.

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5415

Thanks so much for contributing! I'm going to close this one out and submit the other one on your behalf.

PR #5357, opened 2019-07-18
Add support for .msix package files

julianloehr-kg:
## Overview

The build pipeline sometimes produces `.msix` or `.msixbundle` package files instead of `.appx`. When trying to deploy those packages they're obviously not found by the MRTK deploy pipeline.


wiwei:
Thanks, looks great!

For others, some more context:
https://blogs.msdn.microsoft.com/sgern/2018/06/29/but-what-about-app-v/

PR #5358, opened 2019-07-18
Replace explicit hand animation step numbers with a central definition.

lukastoenneMS:
## Overview

Hand movement in tests is currently done with explicit frame step numbers in every test. Each test has its own definition for the number of tests, some running with lower numbers to decrease overall test run time.

This change adds a few central hand move step numbers in the PlayModeTestUtilities for fast/slow/default movement. The default is lower when run in batch mode to speed up test runs.

## Verification
Make sure tests pass both locally and on CI (use the -batchmode argument to run CI-like tests locally).

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
+++ b/Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
@@ -35,6 +35,15 @@ namespace Microsoft.MixedReality.Toolkit.Tests
         // Unity's default scene name for a recently created scene
         const string playModeTestSceneName = "MixedRealityToolkit.PlayModeTestScene";
 
+        public const int HandMoveStepsFast = 3;
+        public const int HandMoveStepsSlow = 30;
+
+        // Keep number of steps low so tests run fast in batch mode.
+        private const int handMoveStepsDefaultBatchMode = 5;
+        // Increased number of steps to be able to see hand movements in the editor test runner.
+        private const int handMoveStepsDefaultEditor = 20;
+        public static int HandMoveStepsDefault => Application.isBatchMode ? handMoveStepsDefaultBatchMode : handMoveStepsDefaultEditor;
+
         private static Stack<MixedRealityInputSimulationProfile> inputSimulationProfiles = new Stack<MixedRealityInputSimulationProfile>();
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


lukastoenneMS:
**Warning** A lot of tests are failing with slightly changed hand animation steps, this needs to be investigated first.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
```diff
@@ -262,6 +274,36 @@ internal static IEnumerator SetHandState(Vector3 handPos, ArticulatedHandPose.Ge
             }
```
luis-valverde-ms: Wouldn't it be better to return `WaitForFixedUpdate` here?

Assets/MixedRealityToolkit.Tests/PlayModeTests/TestHand.cs
```diff
@@ -50,30 +50,60 @@ public IEnumerator Hide()
             yield return new WaitForFixedUpdate();
         }
 
-        public IEnumerator MoveTo(Vector3 newPosition, int numSteps = 30)
+        public IEnumerator MoveToSteps(Vector3 newPosition, int numSteps)
```
luis-valverde-ms: Can we have a single MoveTo method defaulting numSteps to `PlayModeTestUtilities.HandMoveStepsDefault`?

lukastoenneMS: I was thinking that having an separate name for the explicit step version would be more clear, but no strong opinion on that, i will change it.

_Minor technical note_: i can't use the `PlayModeTestUtilities.HandMoveStepsDefault` directly as a default for the parameter, because it's not a runtime const. Instead i would just use two overloads for the same function, which amounts to the same thing.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PlayModeTestUtilities.cs
lukastoenneMS: Yeah, probably a good idea, to make sure the input simulation service sees the changed hand data and updates.

On a related note: I have #5173 which makes it takes away the monopoly on simulated hands from the input simulation service. In a similar vein we could actually create true hand controllers in the tests themselves, rather than changing hand data and then waiting for the service to pick up on the changes. That should hopefully reduce the error cases where "wait for one frame" fixes things magically.

wiwei:
This PR is super old, and I'm going to close this out - note that I've taken what was done and reopened a new one here:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/6915

I think that there's good value in this change, so I didn't want to just close without bringing forward some goodness.

Issue #5305, opened 2019-07-13
Rotation constraints are ignored during one handed movement


julenka:
## Overview
Manipulationhandler has section "constraints on rotation". These are applied when doing two handed rotation, but are ignored when doing one handed movement.

Note: after this bug fix, the rotation mode "gravity aligned face user" is no longer needed, so I would recommend removing it (or marking it as deprecated somehow).

## Steps to reproduce
Configure handler for one handed movement - face user, and then set "rotation constraint" to "y only".

## Expected Behavior
Object will face user, but rotate only about Y axis

## Actual Behavior
Object faces the user, including rotation about the x axis.

wiwei:
@julenka Marking as deprecated/obsolete would be best.

thalbern:
@julenka the current implementation of constraints is that they're only applied to one hand manipulations that don't perform any auto orientations on the object. I suppose that was done because some of the rotation modes wouldn't make sense with the constraints in place. 

So constraints currently work for:
- rotateaboutgrabpoint
- rotateaboutobjectcenter

We could move that code that scales the rotations out of that specific section and apply it to all rotation modes instead

julenka:
> We could move that code that scales the rotations out of that specific section and apply it to all rotation modes instead

That's what I was thinking. I had written the original code, but when I revisited my own design last week I found it confusing / not intuitive that the rotation constraints did not apply to the one hand modes.

PR #5359, opened 2019-07-18
Fixes on constraints of one hand rotation + tests

thalbern:
## Overview
- moved constraint calculation so it can be applied to all rotation modes
- fixed constraint calculation so original rotation is maintained (and not set to 0 like it was before)
- added caching of original rotation for constraints so objects wouldn't slowly drift of hand
- added tests for checking one hand rotation with and without constraints
- added rotate method for test hand / test utilities

## Changes
- Fixes: #5305 

## Verification
- editor hand simulation
- WMR (far and near)
- HL2
- unit test


julenka:
Lol you said DO NOT REVIEW in big letters so of course I had to look :-)

wiwei:
> Lol you said DO NOT REVIEW in big letters so of course I had to look :-)

Yeah this is the like the do not upvote sorta thing.

thalbern:
> Lol you said DO NOT REVIEW in big letters so of course I had to look :-)

haha that was just a sneak peak for lukas! - now it's ready, added you both to the review 😸

julenka:
> added caching of original rotation for constraints so objects wouldn't slowly drift of hand

I believe Lukas already completed a PR that should solve this: #5364

julenka:
Should we also update the docs to mention that constraints are applied to all modes?

Currently it says:

> Constraints on Rotation Specifies on which axis the object will rotate when interacted with.

Which is a big ambiguous

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -660,6 +671,11 @@ private void HandleOneHandMoveStarted()
             PointerData pointerData = GetFirstPointer();
             IMixedRealityPointer pointer = pointerData.pointer;
             moveLogic.Setup(pointerData.GrabPoint, hostTransform.position);
+			
+			// cache objects rotation on start to have a reference for constraint calculations
```
julenka: spacing

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
```diff
@@ -476,6 +476,90 @@ public IEnumerator ManipulationHandlerOneHandMoveFar()
...
+            yield return hand.Show(initialObjectPosition);
+            yield return null;
+
+            // grab the object
+            yield return hand.SetGesture(ArticulatedHandPose.GestureId.Pinch);
+            yield return new WaitForFixedUpdate();
+            yield return null;
+
+            float testRotation = 45;
+            const int numRotSteps = 30;
```
julenka: Might want to do less to get tests to run faster. As Lukas suggested perhaps we should have  a default "num steps" amount that's like, 10 or less.

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
```diff
@@ -476,6 +476,90 @@ public IEnumerator ManipulationHandlerOneHandMoveFar()
 
         }
 
+
+        /// <summary>
+        /// This tests the one hand near rotation and applying different rotation constraints to the object.
+        /// </summary>
+        /// <returns></returns>
+        [UnityTest]
+        public IEnumerator ManipulationHandlerOneHandRotateWithConstraint()
```
julenka: It would also be nice to test two hand rotation as well, unless you have other high priority bugs to get to for GA (it's a nice to have).

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
lukastoenneMS: I have a PR open for unifying those numSteps across tests, but doing that breaks a lot of them right now. So i would postpone that until after GA.
#5358

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -565,7 +566,27 @@ private void HandleTwoHandManipulationUpdated()
...
+                    break;
+                case RotationConstraintType.YAxisOnly:
+                    diffRotation.eulerAngles = Vector3.Scale(diffRotation.eulerAngles, Vector3.up);
+                    break;
+                case RotationConstraintType.ZAxisOnly:
+                    diffRotation.eulerAngles = Vector3.Scale(diffRotation.eulerAngles, Vector3.forward);
+                    break;
+            }
+
+            return diffRotation * hostWorldRotationOnManipulationStart;
```
lukastoenneMS: Regarding my changes in #5364: those affect only the position vector, by ensuring that the grab point is properly fixed in object space.

The rotation is computed independently before that. Bernadette's change improves the stability of the rotation constraint, by constraining the overall rotation since start of the manip instead of each incremental rotation offset.

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
thalbern: will tune it down a bit for now - final steps will be decided by lukas clean up PR :)

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
thalbern: I had a look , but the test for 2 hand rotation will be very different from this one - so i'll make a separate PR for that - i'll add it to the manipulation handler test task. Fixing the last GA bug first

PR #5360, opened 2019-07-18
Add Platform Toolset switch to Build Window

julianloehr-kg:
## Overview

The project solution produced by the Unity build step always targets Platform Toolset v141 (Visual Studio 2017), regardless of Visual Studio Version Build Settings or whether Visual Studio 2017 or 2019 is installed. Build failures arise, when there is no Platform Toolset v141 installed, i.e. only Visual Studio 2019 has been installed. 

This issue has been reported already to Unity : https://issuetracker.unity3d.com/issues/uwp-building-project-for-visual-studio-2019-generates-solution-that-targets-build-tools-for-vs-2017-tool-set-v141

Attempts to figure out whether Platform Toolset v141 is installed seem to be non-trivial. For the time being, a convenience workaround is to override the Platform Toolset when invoking MSBuild. For the user this override is exposed as Appx Build Option in the Build Window. It defaults to Platform Toolset v141, but lets the user choose to override it to v142 in case v141 is not installed.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildDeployPreferences.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildDeployPreferences.cs
@@ -12,6 +12,7 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
     {
         public static Version MIN_SDK_VERSION = new Version("10.0.18362.0");
         private const string EDITOR_PREF_BUILD_CONFIG = "BuildDeployWindow_BuildConfig";
+        private const string EDITOR_PREF_PLATFORM_TOOLSET = "BuildDeployWindow_PlatformToolset";
         private const string EDITOR_PREF_FORCE_REBUILD = "BuildDeployWindow_ForceRebuild";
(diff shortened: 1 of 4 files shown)


keveleigh:
Looks like the build is failing:

>C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.Cpp.Platform.targets(61,5): error MSB8038: Platform Toolset is not defined. Please select one of the available Platform Toolsets in the Project Properties UI. [f:\aipmragent_work\1\a\build\UWP_x86_default\MixedRealityToolkit\MixedRealityToolkit.vcxproj]

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpBuildDeployPreferences.cs
```diff
@@ -28,6 +29,15 @@ public static string BuildConfig
             set => EditorPreferences.Set(EDITOR_PREF_BUILD_CONFIG, value.ToLower());
         }
 
+        /// <summary>
+        /// The current Build Configuration. (Debug, Release, or Master)
```
wiwei: Note to self: Update this when it goes in (since I don't have rights on the fork)

PR #5361, opened 2019-07-18
Fix build failure due to empty scene paths

julianloehr-kg:
## Overview

There are occasions where the EditorBuildSettings enumerates all scenes, regardless whether they are missing or not. In that case the scene is reported as enabled, though has an empty path. The build process later fails, as is cant' find the scene at `""`. Therefore better be safe than sorry and filter out scenes that have an empty path.


wiwei:
Looks great, thanks for contributing!

PR #5362, opened 2019-07-18
[Docs] Updated infographic for the MRTK Architecture

cre8ivepark:
## Overview
Updated architecture infographic image.

## Preview
https://github.com/cre8ivepark/MixedRealityToolkit-Unity/blob/mrtk_development_ArchitectureDiagram/Documentation/Architecture/Overview.md

Updated image file:
https://github.com/cre8ivepark/MixedRealityToolkit-Unity/blob/mrtk_development_ArchitectureDiagram/Documentation/Images/Architecture/MRTK_Architecture.png

## Before & After
2019-07-18 22_47_13-Architecture overview _ Mixed Reality Toolkit Documentation

2019-07-18 22_54_22-2019-07-18 22_47_13-Architecture overview _ Mixed Reality Toolkit Documentation


Troy-Ferrell:
For "Articulated" above Input simulation, can we expand to "Articulated Hands" to be explicit? Also maybe for GGV add "GGV (HL1)" to add some more context for new readers

Also right above input what is the WMR section supposed to be? It seems like a weird breakdown to Unity/WMR/Speech?

wiwei:
> For "Articulated" above Input simulation, can we expand to "Articulated Hands" to be explicit? Also maybe for GGV add "GGV (HL1)" to add some more context for new readers
> 
> Also right above input what is the WMR section supposed to be? It seems like a weird breakdown to Unity/WMR/Speech?

For the things in the WMR section, those are the input providers - the weird thing here is that, the layers aren't all exactly "identical" - i.e. those three boxes should be labelled as those input providers/device managers, but everything else on that row shouldn't have that label. I'm not sure what the right answer is here

Troy-Ferrell:
> everything els

What about just one green box "Input Providers" above "Input"?

wiwei:
> > everything els
> 
> What about just one green box "Input Providers" above "Input"?

Yeah, that works!

cre8ivepark:
Great! I'll apply the changes!

cre8ivepark:
How about this?
MRTK_Architecture

wiwei:
> How about this?
> MRTK_Architecture

The vertical ordering of input simulation and boundary on top of the other things feeeeeeels a bit weird. I think that the old image but with the specific providers removed (and just replaced with the box "input providers" would have been a good middle ground)

Troy-Ferrell:
> > How about this?
> > MRTK_Architecture
> 
> The vertical ordering of input simulation and boundary on top of the other things feeeeeeels a bit weird. I think that the old image but with the specific providers removed (and just replaced with the box "input providers" would have been a good middle ground)

Yea, it gives the sense of dependency/flow for the diagnostics/inputsimulation which isn't true...but that's what we want them to think for the input stack. 

I would maybe just simplify everything. Have the Input Stack, then SpatialAwareness next to it, then maybe TeleportService, and then the dots. We don't need to show all services...we are likely going to add and demote some of these "services" in the future. 

Also is "Focus" supposed to be "FocusProvider"?

julenka:
> I would maybe just simplify everything. Have the Input Stack, then SpatialAwareness next to it, then maybe TeleportService, and then the dots. We don't need to show all services...we are likely going to add and demote some of these "services" in the future.

I'm having a tough time visualizing this, @Troy-Ferrell, any way you could share a quick napkin sketch (just draw on paper and take photo, no need to spend lots of time on it)

julenka:
Note that the Input Simulation service is a input data provider, so this bit could be improved:

image

wiwei:
> Note that the Input Simulation service is a input data provider, so this bit could be improved:
> 
> image
> 
> However, once I start doing that I start reworking the whole diagram and it's a bit of a rabbit hole. I think it's worth re-thinking this diagram, but perhaps later once we've fixed the other GA bugs :-)

Yep, definitely worth revisiting. I'm sure by then we'll have new goodies to put in there :)

cre8ivepark:
Simplified version based on the feedback:
MRTK_Architecture

Issue #5365, opened 2019-07-18
Clean unused scripts


Troy-Ferrell:
## Describe the bug

use the dependency window tool and visual studio to identify scripts no longer in use and that can easily be deprecated

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.11f1]
- MRTK Version [e.g. v2.0]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

PR #5366, opened 2019-07-18
Deprecate unused scripts

Troy-Ferrell:
## Overview

This change is deprecating unused scripts in MRTK, particularly Examples scene helper monobehaviors that are no longer in use (rotate cube, etc).

Two other minor adjustments include in this change:
- Updated meta files for MRTK editor icons
- Moved StandardShader utilities under (Assets/MRTK/Utilities) to their own folder

List of scripts targeted for deprecation:
DemoTouchButton.cs
GrabTouchExample.cs
ToggleBoundingBox.cs
GazeHandHelper.cs
InteractableHighlight.cs
ButtonBackgroundSize.cs
ButtonBackgroundSizeGizmo.cs
ButtonBackgroundSizeOffset.cs
GrabReleaseDetector.cs
OnSelectVisualizerInputController.cs
KeepFacingCamera.cs
KeepThisAlive.cs
StatusText.cs

## Changes
- Fixes: #5365 

## Verification

Assets identified as unused via 3 methods:
1) Dependency window see's no dependency via YAML analysis
2) Find all References in Visual Studio yields no results
3) Search all text for the class name yields no results

TODO: Run EyeTracking example scenes on device

>
>


julenka:
@wiwei Do we have a plan to eventually remove these, or will we just keep them around, deprecated?

wiwei:
@julenka, we shouldn't keep these around forever - I'm drafting up a proposal here for what happens to things that get deprecated/timeline and other stuff

Issue #5252, opened 2019-07-09
GazeProvider GazeTarget returns null when articulated hands are up


julenka:
## Overview
Many customers use GazeProvider to know what user is looking at by examining where the head ray is pointing. For example, Experimental/Features/Utilities/StabilizationPlaneModifier.cs uses the GazeProvider to determine where to place the LSR stabilization plan for Hololens.

The `GazeProvider.GazeTarget` is filled in on startup, but then becomes null whenever articulated hands are present, even though Gazeprovider is enabled.

## Steps to reproduce
You can use the following class to print the gaze target 

```
using Microsoft.MixedReality.Toolkit;
using System.Collections;
using System.Collections.Generic;
using UnityEngine;

public class GazeProviderDebug : MonoBehaviour
{
    // Update is called once per frame
    void Update()
    {
        var gp = MixedRealityToolkit.InputSystem.GazeProvider;
        //DebugInfo.Status($"gazeprovider enabled: {gp.Enabled} target: {gp.GazeTarget}");
        Debug.Log($"gazeprovider enabled: {gp.Enabled} target: {gp.GazeTarget}");
    }
}

```
2. Place a cube at 0, 0, 3, attach the GazeProviderDebug script to the cube.
3. Press play, look at console, notice the following messages:

gazeprovider enabled: True target: Cube (UnityEngine.GameObject)

4. Bring up articulated hand, look at console, notice messages

gazeprovider enabled: True target: 

## Expected Behavior
Even when the articulated hands are up, the console output should read:

gazeprovider enabled: True target: Cube (UnityEngine.GameObject)

julenka:
This is a tricky one -- I briefly looked into it but didn't quickly find solution (spent ~5 minutes on it). Can somebody with knowledge of gaze provider and input provide thoughts on how to fix / hints for where to look? @luis-valverde-ms or @keveleigh  know the most, @googlan  also knows about input and could help.

wiwei:
This could be a side effect of how the whole hands/eyes interactions aren't reallllllly specified and are still being figured out.

I haven't looked at this myself, but this could fall into two buckets:

1) An obvious bug which we just fix.
2) Another instance of the fact that we haven't really figured out the perfect set of hand/eye interaction models, and we'd be doing some hack or workaround that doesn't necessarily create additional clarity.

I hope it's #1

keveleigh:
My guess is it's a side effect of disabling the gaze pointer:

[...]

PR #5369, opened 2019-07-18
Fix GazeProvider.GazeTarget being null when hands are active

julenka:
## Overview
Several components like StabilizationPlaneModifier and apps that want to know where user is looking rely on GazeProvider.GazeTarget (and other raycast info) to be filled in even when gaze is not the primary input. The Input System had special-cased gaze provider for this reason, but wasn't actually filling in the hit data if gaze wasn't active.

This change actually fills in the hit data so that the gaze result is not null when hands / other controllers are up.

## Breaking Change
If you were relying on GazeProvider.GazeTarget == null to determine if gaze input is active, you should now instead use GazeProvider.Pointer.IsInteractionEnabled.

## Changes
- Fixes: #5252 
- Added test to validate fix


Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -517,11 +517,42 @@ public override void Update()
...
             PrimaryPointer = primaryPointerSelector?.Update();
         }
 
+        /// <summary>
+        /// Update the hit information for the gaze provider.
+        /// 
+        /// For the sake of StabilizationPlaneModifier and potentially other 
+        /// components that care where the user's looking, we need to do a gaze 
+        /// raycast even if gaze isn't used for focus.
+        /// Assumes gaze provider is not null;
```
wiwei: >; [](start = 45, length = 1)

Nit: . instead of ;

(I'm using codeflow so suggestions aren't possible AFAIK)

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -517,11 +517,42 @@ public override void Update()
...
             UpdateFocusedObjects();
 
             PrimaryPointer = primaryPointerSelector?.Update();
         }
 
+        /// <summary>
+        /// Update the hit information for the gaze provider.
+        /// 
+        /// For the sake of StabilizationPlaneModifier and potentially other 
+        /// components that care where the user's looking, we need to do a gaze
```
wiwei: Nit: Generally for comments, avoid mentioning what or who is calling you, as it's possible for this comment to get out of date just via that component changing (and it's super easy to not update this comment, as you could make it to StabilizationPlaneModifierdoesn't call this anymore, but unless you grep StabilizationPlaneModifier (i.e. not just the functions you changed in StabilizationPlaneModifier), you won't think to update this)

So long story short, probably just the "Updates the the gaze raycast provider even in scenarios where gaze isn't used for focus"

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
```diff
@@ -521,6 +521,25 @@ private async void RaiseSourceDetected()
...
         }
 
+        internal void UpdateGazeInfoFromHit(MixedRealityRaycastHit raycastHit)
+        {
+            HitInfo = raycastHit;
+            if (raycastHit.transform != null)
+            {
+                GazeTarget = raycastHit.transform.gameObject;
+                var ray = GazePointer.Rays[0];
+                var lhd = (raycastHit.point - ray.Origin).magnitude;
```
wiwei: > var lhd = (raycastHit.point - ray.Origin).magnitude; [](start = 15, length = 53)

Do we also need to update gazeProvider.lastHitDistance?

See lines 266-268.

More generally, it might be worth having a helper function that takes an object and properly sets all three things, so that we don't forget to update them all as a unit (otherwise we risk similar bugs in the future where we only update 2/3 or 1/3 of the things)

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -517,11 +517,42 @@ public override void Update()
...
+        /// 
+        /// For the sake of StabilizationPlaneModifier and potentially other 
+        /// components that care where the user's looking, we need to do a gaze 
+        /// raycast even if gaze isn't used for focus.
+        /// Assumes gaze provider is not null;
+        /// </summary>
+        private void UpdateGazeProvider()
+        {
+            // We only update if interaction is not enabled because if interaction is enabled 
+            // for this pointer, we have already updated hit data
```
wiwei: > we have already updated hit data [](start = 32, length = 33)

Is there any way to assert that we aren't recalc-ing things? This introduces cross-code dependency that is implicit, meaning we could "fix" this issue in another location and not know to update this (i.e. delete it) later.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -517,11 +517,42 @@ public override void Update()
...
+            if (!gazeProviderPointingData.Pointer.IsInteractionEnabled)
+            {
+                var raycastProvider = InputSystem.RaycastProvider;
+                // Should we just use a local variable instead of this member?
+                // Unclear why we use the member everywhere...maybe to reduce allocaionts
```
wiwei: I think that yeah, this is to avoid update-loop allocations.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
```diff
@@ -517,11 +517,42 @@ public override void Update()
...
+                hitResult3d.Clear();
+                LayerMask[] prioritizedLayerMasks = (gazeProviderPointingData.Pointer.PrioritizedLayerMasksOverride ?? FocusLayerMasks);
+
+                QueryScene(gazeProviderPointingData.Pointer, raycastProvider, prioritizedLayerMasks,
+                    hitResult3d, maxQuerySceneResults);
+                ((GazeProvider)InputSystem.GazeProvider).UpdateGazeInfoFromHit(hitResult3d.raycastHit);
```
wiwei: >(GazeProvider) [](start = 17, length = 14)

This has an interesting side effect of preventing other people from implementing their own gaze provider directly using the IMixedRealityGazeProvider interface.

I think that, truth be told, it's really hard to implement the gaze provider correctly without having everything inside GazeProvider.cs, so the value of the interface itself isn't high, but if we are going to do this we should do this consciously and make this tradeoff explicitly.

Assets/MixedRealityToolkit.Tests/PlayModeTests/FocusProviderTests.cs
```diff
@@ -70,6 +70,29 @@ public IEnumerator TestGazeCursorArticulated()
...
+
+            yield return null;
+
+            Assert.NotNull(MixedRealityToolkit.InputSystem.GazeProvider.GazeTarget, "GazeProvider target is null when looking at an object");
+
+            TestHand h = new TestHand(Handedness.Right);
+            yield return h.Show(Vector3.forward * 0.2f);
+            yield return null;
+
+            Assert.NotNull(MixedRealityToolkit.InputSystem.GazeProvider.GazeTarget, "GazeProvider target is null when looking at an object with hand raised");
```
wiwei: Is there also any way to validate that the the object isn't focused with the hand raised (i.e. that interaction enabled is properly being honored?)

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: I could add another method to the GazeProvider interface :-)

But honestly I kind of feel like GazeProvider shouldn't be an interface, for this reason:

> I think that, truth be told, it's really hard to implement the gaze provider correctly without having everything inside GazeProvider.cs

Assets/MixedRealityToolkit.Tests/PlayModeTests/FocusProviderTests.cs
julenka: Yes, this is covered in [PointerEventTests::Test(). ](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/f7f9e051327bc145012fc64804681f03d2cd709a/Assets/MixedRealityToolkit.Tests/PlayModeTests/PointerEventsTests.cs#L174)

```
            // Show hand far enough from the test collider to not give it focus
            var rightHand = new TestHand(Handedness.Right);
            Vector3 rightNoFocusPos = new Vector3(0.05f, 0, 1.0f);
            yield return rightHand.Show(rightNoFocusPos);

            // Should receive focus left events from gaze pointer
            expected.numBeforeFocusChange++;
            expected.numFocusChanged++;
            expected.numFocusExit++;
            handler.state.AssertEqual(expected);
```

Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
julenka: Refactored, but forgot to fix lastHitDistance, so leaving open.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: In the current mrtk, FocusProvider and GazeProvider are quite closely coupled: FocusProvider basically updates all the data for gaze provider. I actually don't think the architects intended this. If you look at the IGazeProvider interface, all its fields are only getters, meaning I think architects intended GazeProvider to update itself. It would actually be pretty easy to see a system where this would work -- in gaze provider's update loop, GazeProvider just does a raycast and updates its fields. But the current code instead has FocusProvider updating GazeProvider. 

I'd rather not add to the interface, since I think its design is not so bad -- to encourage people to have the gaze provider update itself. It's just the specific implementation we have doesn't do that.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
julenka: I looked into whether we could null out and then read some fields, but didn't find any. Instead I refactored the code so that we only ever populate the gazeprovider data in one location -- from the focus provider. Previously we populated the gaze provider data in two locations--once from FocusProvider, once from GazeProvider.InternalGazePointer. Is there still a cross-code dependency?

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
wiwei: Nit: Delete the comment (since it's just for CR, and I think this was answered)

wiwei:
Thanks! Still lgtm.

Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs
keveleigh: I was about to write this same comment! I think I'm mostly agreed with the conclusion here, where the more common data provider to pointer relationship seems more muddled with the gaze provider than any of the hand / controller pointers. On initial thought, I'm not sure about the GazeProvider doing its own raycast (to try to get it to follow the pattern of the other providers / pointers as much as possible), but its possible we'll need to special case it.

PR #5370, opened 2019-07-19
Adds the MixedRealityToolkit.Tools folder and package.

wiwei:
There are a number of features currently in the the MixedRealityToolkit folder that are useful at editor and build time, but are otherwise never going to be shipping with the final application. For example, the screenshot utility is incredibly useful for people who want to take screenshots of their features (useful for MRTK devs especially!) but otherwise is a bit odd to have directly in the MRTK "core" folder, because it's not really a core part of a consumer's development process.

This change spawns a new folder to put these things in (MixedRealityToolkit.Tools) where we can place features (especially ones that are editor-only) that accrue value during the development cycle (or could just generally be useful tools even outside of the MRTK, but due to repo/CI resourcing we're currently putting them in here), but otherwise don't ship with the app itself.

Note that this change doesn't move EVERYTHING over (for example, the BuildWindow tool is still left inside the core part because there is some history there and it would be some degree of a "package" breaking change to move that over.)

This change also adds the new MixedRealityToolkit.Tools.nuget package, along with the corresponding MixedRealityToolkit.Tools.unitypackage

## Notes for reviewers:

I would love to get feedback on the naming and the general principles behind this move. This is somewhat of a reaction to how some helpful but perhaps not "core" utilities have made their way into the core folders of the MRTK (especially ones that seem like they could be generally useful for Unity development).

I would also appreciate feedback on whether or not this SHOULD be a different package, or if people think we should continue to bundle things up in the .Foundation package to:

1) Reduce the mental overhead of having another package.
2) Reduce the cost of also having to update the existing tutorials and such.

I think the answer is "no" because consumers should be able to take just the code needed to build what they want, and then use our tooling and import that if they want (obviously the existing foundation package is already fairly large, so the idea here is we want to not grow it in ways that diverge from our philosophies)


Assets/MixedRealityToolkit.Tools/Microsoft.MixedReality.Toolkit.Tools.asmdef
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "Microsoft.MixedReality.Toolkit.Tools",
+    "references": [
+        "Microsoft.MixedReality.Toolkit",
+        "Microsoft.MixedReality.Toolkit.Editor.Inspectors",
+        "Microsoft.MixedReality.Toolkit.Editor.Utilities"
+    ],
+    "optionalUnityReferences": [],
+    "includePlatforms": [],
```
Troy-Ferrell: I think this needs to be updated to only include "Editor"

julenka:
> I would also appreciate feedback on whether or not this SHOULD be a different package, or if people think we should continue to bundle things up in the .Foundation package...
> I think the answer is "no" because consumers should be able to take just the code needed to build what they want, and then use our tooling and import that if they want (obviously the existing foundation package is already fairly large, so the idea here is we want to not grow it in ways that diverge from our philosophies)

I don't think the overhead is that high to have a separate package, you have to download one more package sure, but presumably you won't be downloading packages that often. I think longer term it will be good to have a separate package, since it will prevent the core packages from getting too large as people add more useful tools.

julenka:
For the build window, do we have a plan / task to migrate it? If not, can we include a readme explaining why the build window is not in the tools package?

Cameron-Micka:
Makes sense to me to move these items to a separate package. Do you want to pull in the texture combiner too? https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Inspectors/TextureCombinerWindow.cs

wiwei:
> Makes sense to me to move these items to a separate package. Do you want to pull in the texture combiner too? https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Inspectors/TextureCombinerWindow.cs

Yasssssssss.

wiwei:
> For the build window, do we have a plan / task to migrate it? If not, can we include a readme explaining why the build window is not in the tools package?

Also, something a bit TBD - I want to gather a little more data before moving this, because this has some history of being in the foundation package and my concern for changing this now is a little higher.

Good point on the readme, lemme do that...

Troy-Ferrell:
Also maybe, update the "Utilities" section in the TOC to now be "Tools"? To keep consistency? I don't think this will affect references as the URLs will remain the same

Documentation/Tools/DependencyWindow.md
```diff
@@ -1,9 +1,9 @@
 
 # Dependency Window
 
-Often in Unity it is difficult to gleam which assets are being used, and what is referencing them. The "Find References in Scene" option works great when you are only concerned with the current scene, but what about your entire Unity project? This is where the [Dependency Window](/Assets/MixedRealityToolkit/Inspectors) can be useful.
+Often in Unity it is difficult to gleam which assets are being used, and what is referencing them. The "Find References in Scene" option works great when you are only concerned with the current scene, but what about your entire Unity project? This is where the [Dependency Window](/Assets/MixedRealityToolkit.Tools/DependencyWindow) can be useful.
```
keveleigh: >/Assets/MixedRealityToolkit.Tools/DependencyWindow

Looks like this was added before your PR, but I don't think a relative link like this resolves to anything in our generated docs.

wiwei:
```suggestion
Often in Unity it is difficult to gleam which assets are being used, and what is referencing them. The "Find References in Scene" option works great when you are only concerned with the current scene, but what about your entire Unity project? This is where the [Dependency Window]((https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Tools/DependencyWindow) can be useful.
```

Documentation/Tools/ScreenshotUtility.md
```diff
@@ -1,9 +1,9 @@
 
 # Screenshot Utility
 
-Often taking screenshots in Unity for documentation and promotional imagery can be burdensome and the output often looks less than desirable. This is where the [ScreenshotUtility](/Assets/MixedRealityToolkit/Utilities/Editor) class comes into play.
+Often taking screenshots in Unity for documentation and promotional imagery can be burdensome and the output often looks less than desirable. This is where the [ScreenshotUtility](/Assets/MixedRealityToolkit.Tools/ScreenshotUtility) class comes into play.
```
keveleigh:
```suggestion
Often taking screenshots in Unity for documentation and promotional imagery can be burdensome and the output often looks less than desirable. This is where the [ScreenshotUtility](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Tools/ScreenshotUtility) class comes into play.
```

Documentation/Tools/ScreenshotUtility.md
```diff
@@ -1,9 +1,9 @@
...
 
-The [ScreenshotUtility](/Assets/MixedRealityToolkit/Utilities/Editor) class aides in taking screenshots via menu items and public APIs within the Unity editor. Screenshots can be captured at various resolutions and with transparent clear colors for use in easy post compositing of images. Taking screenshots from a standalone build is not supported by this tool. 
+The [ScreenshotUtility](/Assets/MixedRealityToolkit.Tools/ScreenshotUtility) class aides in taking screenshots via menu items and public APIs within the Unity editor. Screenshots can be captured at various resolutions and with transparent clear colors for use in easy post compositing of images. Taking screenshots from a standalone build is not supported by this tool.
```
keveleigh:
```suggestion
The ScreenshotUtility class aides in taking screenshots via menu items and public APIs within the Unity editor. Screenshots can be captured at various resolutions and with transparent clear colors for use in easy post compositing of images. Taking screenshots from a standalone build is not supported by this tool. 
```

PR #5373, opened 2019-07-19
break out BaseCursor's target lerp functionality to make it accesible to derived classes

naostranMS:
## Overview

We override UpdateCursorTransform in some of our custom cursor classes, and I wanted to be able to use the lerp functionality of the BaseCursor class without having to duplicate it. It's a pretty small change, and we can workaround it easily enough if it's not generally useful (but I thought it might be at least).

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/InteractiveMeshCursor.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/InteractiveMeshCursor.cs
@@ -49,7 +49,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         private bool hasHand = false;
         private bool isDown = false;
 
-        private new Vector3 targetScale;
+        private float ringDotTargetScale;
         private Vector3 initialScale;
 
         private void Awake()
@@ -75,7 +75,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
             isDown = IsPointerDown;
             hasHover = TargetedObject != null;
 
-            targetScale = Vector3.one * defaultScale;
+            ringDotTargetScale = defaultScale;
             bool showRing = false;
 
             switch (state)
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/InteractiveMeshCursor.cs
```diff
@@ -49,7 +49,7 @@ public class InteractiveMeshCursor : BaseCursor
         private bool hasHand = false;
         private bool isDown = false;
 
-        private Vector3 targetScale;
+        private new Vector3 targetScale;
```
davidkline-ms: You marked targetScale as protected in BaseCursor.cs, is there a reason you decided to hide it (and make it private) in this class?

naostranMS: I did it this way as I'm not really familiar with InteractiveMeshCursor (we don't use it), and so I wanted to just keep the behavior the same (while also wanting to minimize the amount of code changed). If I used the base version instead of hiding, I think it would potentially do a double-lerp of the scale (InteractiveMeshCursor.UpdateCursorTransform calls base.UpdateCursorTransform, and both do lerps, and not quite in the same way).
I'd be happy to rename the variable if you think that would be better, but I'd be hesitant to make any more substantial changes myself due to lack of familiarity.

wiwei: @davidkline-ms, any recommendations for @naostranMS here?

davidkline-ms: I dont have any real recommendation, it just seemed odd to hide a base class' protected member with a derived class private member.

davidkline-ms: > I think it would potentially do a double-lerp of the scale (InteractiveMeshCursor.UpdateCursorTransform calls base.UpdateCursorTransform, and both do lerps, and not quite in the same way).

Question: how does making targetScale private prevent this?

naostranMS: It being private doesn't really matter to this change. I meant that I could have just removed this targetScale and used the now accessible (protected) base class targetScale instead, but then I would be worried about it doing two lerps.
Would you prefer if I changed this to be protected also?

davidkline-ms: Looking at the code, it appears to do two lerps regardless of private or protected :)

I, personally, am not a fan of hiding varaiables.

Thx!

naostranMS: fixed!

davidkline-ms:
safe to ignore the cancelled test pass. it was not intended to be run on this change

PR #5374, opened 2019-07-19
modify BaseCursor to better handle swapping between pointers

naostranMS:
## Overview

We mostly handle cursor management on our own, and use a single cursor instance which gets passed around to whatever pointer is active at a given time. This was causing some problems as some of the cursor code assumes a pretty strong pairing between cursor and pointer, so I've made some changes to hopefully make it a little more flexible without any detrimental effects on the normal usage.

## Changes
1. Various pointer/cursor null-checks so things don't break when they are detached from one another.
2. Make visibleSourcesCount and SourceDownIds refresh at certain points so that they don't get out-of-sync if the cursor missed some source state events.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/AnimatedCursor.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/AnimatedCursor.cs
@@ -59,7 +59,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         {
             base.OnFocusChanged(eventData);
 
-            if (Pointer.CursorModifier != null)
+            if (Pointer?.CursorModifier != null)
             {
                 if ((Pointer.CursorModifier.CursorParameters != null) && (Pointer.CursorModifier.CursorParameters.Length > 0))
                 {
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/AnimatedCursor.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/AnimatedCursor.cs
@@ -59,7 +59,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         {
             base.OnFocusChanged(eventData);
 
-            if (Pointer?.CursorModifier != null)
+            if (((Pointer is UnityEngine.Object) ? ((Pointer as UnityEngine.Object) != null) : (Pointer != null)) && Pointer.CursorModifier != null)
             {
                 if ((Pointer.CursorModifier.CursorParameters != null) && (Pointer.CursorModifier.CursorParameters.Length > 0))
                 {
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
@@ -144,9 +144,18 @@ namespace Microsoft.MixedReality.Toolkit.Input
             get { return pointer; }
             set
             {
+                if (ReferenceEquals(pointer?.BaseCursor, this))
+                {
+                    pointer.BaseCursor = null;
+                }
+
                 pointer = value;
-                pointer.BaseCursor = this;
-                RegisterManagers();
+                if (pointer != null)
+                {
+                    pointer.BaseCursor = this;
+                }
+
+                ResetInputSourceState();
             }
         }
 
@@ -208,7 +217,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// <inheritdoc />
         public virtual void OnSourceDetected(SourceStateEventData eventData)
         {
-            if (eventData.Controller != null)
+            if (Pointer != null && eventData.Controller != null)
             {
                 for (int i = 0; i < eventData.InputSource.Pointers.Length; i++)
                 {
@@ -231,38 +240,26 @@ namespace Microsoft.MixedReality.Toolkit.Input
         /// <inheritdoc />
         public virtual void OnSourceLost(SourceStateEventData eventData)
         {
-            if (eventData.Controller != null)
+            if (Pointer != null && eventData.Controller != null)
             {
                 for (int i = 0; i < eventData.InputSource.Pointers.Length; i++)
                 {
                     // If a source is lost that's using this cursor's pointer, we decrement the count to set the cursor state properly.
                     if (eventData.InputSource.Pointers[i].PointerId == Pointer.PointerId)
                     {
-                        var basePointer = eventData.InputSource.Pointers[i] as BaseControllerPointer;
-
-                        if (basePointer != null &&
-                            basePointer.DestroyOnSourceLost)
-                        {
-                            SourceDownIds.Remove(eventData.SourceId);
-                            Destroy(gameObject);
-                            return;
-                        }
-
                         visibleSourcesCount--;
+                        break;
                     }
                 }
             }
 
-            if (!IsSourceDetected)
-            {
                 SourceDownIds.Remove(eventData.SourceId);
 
-                if (SetVisibilityOnSourceDetected)
+            if (!IsSourceDetected && SetVisibilityOnSourceDetected)
                 {
                     SetVisibility(false);
                 }
             }
-        }
 
         #endregion IMixedRealitySourceStateHandler Implementation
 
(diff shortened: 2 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
```diff
@@ -144,9 +144,18 @@ public virtual IMixedRealityPointer Pointer
             get { return pointer; }
             set
             {
+                if (ReferenceEquals(pointer?.BaseCursor, this))
```
davidkline-ms: Can you add a comment re: why the BaseCursor property is getting set to null when the references are equal?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
```diff
@@ -287,11 +284,15 @@ public virtual void OnBeforeFocusChange(FocusEventData eventData)
         /// <inheritdoc />
         public virtual void OnPointerDown(MixedRealityPointerEventData eventData)
         {
-            foreach (var sourcePointer in eventData.InputSource.Pointers)
+            if (Pointer != null)
             {
-                if (sourcePointer.PointerId == Pointer.PointerId)
+                foreach (var sourcePointer in eventData.InputSource.Pointers)
```
davidkline-ms: nit: consistency: most other loops in this file use for instead of foreach

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
```diff
@@ -305,18 +306,26 @@ public virtual void OnPointerDown(MixedRealityPointerEventData eventData)
         /// <inheritdoc />
         public virtual void OnPointerUp(MixedRealityPointerEventData eventData)
         {
-            foreach (var sourcePointer in eventData.InputSource.Pointers)
+            if (Pointer != null)
             {
-                if (sourcePointer.PointerId == Pointer.PointerId)
+                foreach (var sourcePointer in eventData.InputSource.Pointers)
```
davidkline-ms: nit: consistency: most other loops in this file use for instead of foreach

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
```diff
@@ -455,6 +465,45 @@ public virtual void OnInputDisabled()
...
+        /// updating (for whatever reason).
+        /// </summary>
+        private void ResetInputSourceState()
+        {
+            SourceDownIds.Clear();
+            visibleSourcesCount = 0;
+            if (Pointer != null)
+            {
+                uint cursorPointerId = Pointer.PointerId;
+                foreach (IMixedRealityInputSource inputSource in InputSystem.DetectedInputSources)
```
davidkline-ms: nit: consistency: most other loops in this file use for instead of foreach

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
```diff
@@ -455,6 +465,45 @@ public virtual void OnInputDisabled()
...
+                {
+                    if (inputSource.SourceType != InputSourceType.Head && inputSource.SourceType != InputSourceType.Eyes)
+                    {
+                        foreach (IMixedRealityPointer inputSourcePointer in inputSource.Pointers)
```
davidkline-ms: nit: consistency: most other loops in this file use for instead of foreach

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/AnimatedCursor.cs
```diff
@@ -59,7 +59,7 @@ public override void OnFocusChanged(FocusEventData eventData)
         {
             base.OnFocusChanged(eventData);
 
-            if (Pointer.CursorModifier != null)
+            if (Pointer?.CursorModifier != null)
```
keveleigh: I believe many implementations of `IMixedRealityPointer` are MonoBehaviors, which would cause this check to behave potentially unexpectedly: https://blogs.unity3d.com/2014/05/16/custom-operator-should-we-keep-it/

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
```diff
@@ -231,38 +240,26 @@ public virtual void OnSourceDetected(SourceStateEventData eventData)
...
                 }
             }
 
-            if (!IsSourceDetected)
-            {
                 SourceDownIds.Remove(eventData.SourceId);
 
-                if (SetVisibilityOnSourceDetected)
+            if (!IsSourceDetected && SetVisibilityOnSourceDetected)
                 {
```
keveleigh: Not sure if this is just GitHub formatting, but the indentation here seems inconsistent with the line above.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
keveleigh: I think this might be fine, since this line was present in the original file:
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5374/files#diff-89d7de97a7cae3c9e2c96ea7bc566646L290

I'm good either way though.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
naostranMS: yeah, the indentation is wrong (probably a merge artifact) - I'll fix it

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/AnimatedCursor.cs
naostranMS: I agree that it is problematic - is there an existing MRTK convention for handling this case? We have a have a pretty simple NullIfDestroyed extension in our codebase for this situation (it just tries casting to a MonoBehaviour or Unity.Object and then does the overloaded null check on the casted version), and I could bring that into MRTK if it would help.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
naostranMS: well, it looks good locally now, but still looks bad here :/ hopefully it is just GitHub formatting at this point.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
naostranMS: added

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/BaseCursor.cs
naostranMS: Either way works for me too - just let me know what is preferred.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/AnimatedCursor.cs
davidkline-ms: in general, we have used a two step validation

if (Pointer != null && Pointer.CursorModifier != null)

naostranMS: That works if you're doing the null check through a class type, but not through an interface type. The interface ==/!= is still just doing a reference (in)equality check, without the extra "is destroyed" check that the Unity.Object ==/!= null will do (https://stackoverflow.com/questions/1881459/how-do-i-override-the-equals-operator-for-an-interface-in-c).

naostranMS:
My last commit fixed up the Pointer null checks in BaseCursor. I've left the GazeCursor null check in GazeProvider as there's a bit of a cascading effect there (and lots of other interface null-checks that won't do the UnityEngine.Object null (and destroyed) check. I feel like if the UnityEngine.Object destroyed check for interfaces is really a concern, there should be a consistent way of handling that across all of MRTK.

naostranMS:
> Do you have an example script that we can use to validate the new functionality?

Sorry I missed this initially. I don't really have a good test for this behavior outside of our app.

Issue #4910, opened 2019-06-17
ManipulationHandler uses incorrect origin when object origin is far from the object.


camnewnham:
## Describe the bug

When manipulating an object using a ManipulationHandler and Solver it does not transform around the pointer position or the average of the two grab points. It appears to use the object origin, which in some cases will be a long way from the mesh or collider.

## To reproduce

1. Import MRTK
2. Import a mesh which does not have the origin in the center of the object. This would be the default when exporting as .obj or similar from a CAD platform. For example a cube from (10,10,10) to (11,11,11).
3. Attach a convex MeshCollider
4. Attach a ManipulationHandler
5. Manipulate the object and observe that it behaves unexpectedly (scales around a point a long way away, or moves in fast unexpected movements)

## Expected behavior

(5) The object would be transformed relative to the touch points, or have an option to use mesh origin or bounding box center.

## Your Setup (please complete the following information)

- Unity Version 2018.4.2f1
- MRTK Version v2.0.0 RC2.0

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2

## Additional context

Reproduced for Julia at Hack Week.

MenelvagorMilsom:
@camnewnham @julenka I've been working on fixes for this for a couple of days and I have a few concerns. Myself and @thalbern have been chatting about this.

1. We don't really want to add new features with GA coming up. Giving the option to switch between object origin and mesh origin isn't something we can't add right now. Even if we could, there would still be other issues (see point 3).
1. Given that we can't add this feature right now, we could at least fix it in cases where we do provide options. For example, using the `RotateAboutGrabPoint` option should fix this issue for rotation. However, it doesn't work for far interaction; the object still rotates around the origin. We can fix this (I have it working), but it's not a clean fix. Also, because things are still broken for translate and scale, manipulation overall still feels broken. I feel like this fix needs to be all or nothing for manipulation of these objects to actually feel correct.
1. With the way `ManipulationHandler` is currently set up, the changes required here are quite fundamental to the way we do manipulation. To get some things working I've added a bunch of hacky changes to some fundamental manipulation logic. They don't seem like safe changes so close to GA. The right changes would require an overhaul of a good portion of our manipulation code.

[...]

PR #5379, opened 2019-07-19
Fix manipulation offsets for when origin not mesh centre

MenelvagorMilsom:
## Overview
These changes will make it that most manipulation options will rotate around the grab point for all types of interaction. This means that in most cases, it will not matter that the origin is not at the mesh centre. With smoothing this can still look weird if scaling or rotating quickly, however the object ends up where you expect it to go. 

As I've mentioned [here](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4910#issuecomment-513267204) I wasn't convinced that we should take these changes. There are better ways to do this, but they would require an overhaul of the `ManipulationHandler`, which is not a change we can make right now.

## Changes
- Fixes: #4910 
- Overhaul of `TwoHandMoveLogic` so that it works for rotate around grab point.
- New `TwoHandMoveLogic` now works for one handed near interaction. 
- Changes in `ManipulationHandler` to use this new `TwoHandMoveLogic`.
- Some test changes to respond to the changed logic.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
@@ -310,16 +310,16 @@ namespace Microsoft.MixedReality.Toolkit.UI
             MixedRealityPose pose = new MixedRealityPose();
             pose.Position = (hand1.Position + hand2.Position) / 2;
 
-            Vector3 right = hand1.Position - hand2.Position;
-
             // WIP
-            Vector3 forward = Quaternion.AngleAxis(90, Vector3.up) * right;
+            Vector3 right = hand1.Position - hand2.Position;
+            Vector3 forward = (hand1.Rotation * Vector3.forward + hand2.Rotation * Vector3.forward);
             Vector3 up = Vector3.Cross(forward, right);
-            pose.Rotation = rotateLogic.Update(GetHandPositionMap(), targetRotationTwoHands, constraintOnRotation);
+            //pose.Rotation = rotateLogic.Update(GetHandPositionMap(), targetRotationTwoHands, constraintOnRotation);
+            pose.Rotation = Quaternion.LookRotation(forward);
 
             Debug.DrawRay(pose.Position, right.normalized, Color.red);
             Debug.DrawRay(pose.Position, up.normalized, Color.green);
-            Debug.DrawRay(pose.Position, Vector3.Cross(right, up).normalized, Color.blue);
+            Debug.DrawRay(pose.Position, forward.normalized, Color.blue);
 
             return pose;
         }
@@ -575,7 +575,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
             if ((currentState & State.Moving) > 0)
             {
                 MixedRealityPose pose = GetAveragePointerPose();
-                targetPosition = moveLogic.Update(pose, hostTransform.localScale, true, true);
+                targetPosition = moveLogic.Update(pose, hostTransform.localScale, true);
             }
 
             var handPositionMap = GetHandPositionMap();
@@ -660,7 +660,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
             targetRotation = ApplyConstraints(targetRotation);
             MixedRealityPose pointerPose = new MixedRealityPose(pointer.Position, pointer.Rotation);
-            Vector3 targetPosition = moveLogic.Update(pointerPose, hostTransform.localScale, rotateInOneHandType != RotateInOneHandType.RotateAboutObjectCenter, false);
+            Vector3 targetPosition = moveLogic.Update(pointerPose, hostTransform.localScale, rotateInOneHandType != RotateInOneHandType.RotateAboutObjectCenter);
 
             float lerpAmount = GetLerpAmount();
             Quaternion smoothedRotation = Quaternion.Lerp(hostTransform.rotation, targetRotation, lerpAmount);
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
@@ -237,7 +237,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
         }
         private Dictionary<uint, PointerData> pointerIdToPointerMap = new Dictionary<uint, PointerData>();
         private Quaternion objectToHandRotation;
-        private Vector3 objectToHandTranslation;
         private Vector3 hostLocalGrabPoint;
         private bool isNearManipulation;
         // This can probably be consolidated so that we use same for one hand and two hands
@@ -295,6 +294,34 @@ namespace Microsoft.MixedReality.Toolkit.UI
                 count++;
             }
             return sum / Math.Max(1, count);
+        }
+
+        private MixedRealityPose GetAveragePointerPose()
+        {
+            Debug.Assert(pointerIdToPointerMap.Count > 1);
+
+            var handsEnumerator = pointerIdToPointerMap.Values.GetEnumerator();
+            handsEnumerator.MoveNext();
+            var hand1 = handsEnumerator.Current.pointer;
+            handsEnumerator.MoveNext();
+            var hand2 = handsEnumerator.Current.pointer;
+            handsEnumerator.Dispose();
+            
+            MixedRealityPose pose = new MixedRealityPose();
+            pose.Position = (hand1.Position + hand2.Position) / 2;
+
+            Vector3 right = hand1.Position - hand2.Position;
+
+            // WIP
+            Vector3 forward = Quaternion.AngleAxis(90, Vector3.up) * right;
+            Vector3 up = Vector3.Cross(forward, right);
+            pose.Rotation = rotateLogic.Update(GetHandPositionMap(), targetRotationTwoHands, constraintOnRotation);
+
+            Debug.DrawRay(pose.Position, right.normalized, Color.red);
+            Debug.DrawRay(pose.Position, up.normalized, Color.green);
+            Debug.DrawRay(pose.Position, Vector3.Cross(right, up).normalized, Color.blue);
+
+            return pose;
         }
 
         private Vector3 GetPointersVelocity()
@@ -547,7 +574,8 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
             if ((currentState & State.Moving) > 0)
             {
-                targetPosition = moveLogic.Update(GetPointersCentroid(), IsNearManipulation());
+                MixedRealityPose pose = GetAveragePointerPose();
+                targetPosition = moveLogic.Update(pose, hostTransform.localScale, true, true);
             }
 
             var handPositionMap = GetHandPositionMap();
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -295,6 +294,34 @@ private Vector3 GetPointersCentroid()
...
+            handsEnumerator.MoveNext();
+            var hand2 = handsEnumerator.Current.pointer;
+            handsEnumerator.Dispose();
+            
+            MixedRealityPose pose = new MixedRealityPose();
+            pose.Position = (hand1.Position + hand2.Position) / 2;
+
+            Vector3 right = hand1.Position - hand2.Position;
+
+            // WIP
```
MenelvagorMilsom: I have this marked as WIP. For one, it's doesn't work with 6dof controllers as it doesn't take the controllers rotation into account. Additionally, relying on the `rotateLogic` isn't ideal, as I'm pretty sure there are circumstances where this object will not be Setup with the correct values for the current rotation.

MenelvagorMilsom: This should also include the calculation of `right` above.

MenelvagorMilsom:
Ok, things look to be working now; I've done a bunch of manual testing and automated tests were passing locally. I'm still not sure about how safe these changes are. I've added @wiwei to discuss if this is ready for GA.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
MenelvagorMilsom: Oops, forgot to remove the WIP stuff. I'll do that now.

wiwei:
@MenelvagorMilsom just took a look at this - the changes look to be fairly extensive, but looking at them they seem to be sane to me. Some thoughts:

1) I think that the two hand move logic code is actually an implementation detail of the manipulation handler, but is technically public - those changes are 'technically' breaking but I'm not sure that it's something we ever something we wanted to expose as API surface. It might be worth calling it out in the GA migration notes, and we should probably have some story for marking these things (maybe as "internal" so that they don't get referenced by other people unintentionally in the future)

2) @julenka @thalbern if you guys feel good about this, let's get this in for the GA train, otherwise it'll just ship out with the next iteration (so no biggie either way).

julenka:
> Still I'm opening this PR so we can discuss. There are changes still needed here too, not everything is working fully, so I'm going to use the Do Not Merge label.

You mentioned the PR is ready, could you please update the PR to reflect this? You could just remove this paragraph.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -224,21 +224,12 @@ public Vector3 GrabPoint
...
                 {
-                    if (IsNearPointer())
-                    {
-                        return pointer.Position;
-                    }
-                    else
-                    {
-                        return (pointer.Rotation * initialGrabPointInPointer) + pointer.Position;
-                    }
+                    return (pointer.Rotation * initialGrabPointInPointer) + pointer.Position;
```
julenka: Could you please explain why you removed this? It seems like this is changing behavior of GrabPoint so I want to understand repercussions.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -466,6 +465,37 @@ private void InvokeStateUpdateFunctions(State oldState, State newState)
...
+            pointerIdToPointerMap.Clear();
+
+            // end manipulation
+            State newState = State.Start;
+            InvokeStateUpdateFunctions(currentState, newState);
+            currentState = newState;
+        }
+
+        /// <summary>
+        /// Gets the grab point for the given pointer id
```
julenka: Or zero if none is found?

julenka: Perhaps we should assert here that the pointer exists in the map, returning zero if none is found might lead to some tricky to discover bugs in the future.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -631,16 +662,10 @@ private void HandleOneHandMoveUpdated()
...
-                // Compute the host position such that the grab point falls on the pointer
-                targetPosition = pointerData.GrabPoint - targetRotation * hostLocalGrabPoint;
-            }
-            else
-            {
-                targetPosition = moveLogic.Update(pointerData.GrabPoint, false);
-            }
+            MixedRealityPose pointerPose = new MixedRealityPose(pointer.Position, pointer.Rotation);
+            Vector3 targetPosition = moveLogic.Update(pointerPose, targetRotation, hostTransform.localScale, IsNearManipulation(), rotateInOneHandType != RotateInOneHandType.RotateAboutObjectCenter);
+            //Debug.DrawRay(pointer.Position, Vector3.up, Color.magenta);
```
julenka: nit: remove commented code.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
```diff
@@ -683,11 +709,12 @@ private void HandleOneHandMoveStarted()
...
             objectToHandRotation = worldToPalmRotation * hostTransform.rotation;
-            objectToHandTranslation = worldToPalmRotation * (hostTransform.position - pointer.Position);
+
+            MixedRealityPose pointerPose = new MixedRealityPose(pointer.Position, pointer.Rotation);
+            MixedRealityPose hostPose = new MixedRealityPose(hostTransform.position, hostTransform.rotation);
+            moveLogic.Setup(pointerPose, pointerData.GrabPoint, hostPose, hostTransform.localScale);
 
             Vector3 worldGrabPoint = pointerData.GrabPoint;
-            // Compute the grab point in local space of the host
-            hostLocalGrabPoint = Quaternion.Inverse(hostTransform.rotation) * (worldGrabPoint - hostTransform.position);
```
julenka: can we remove hostLocalGrabPoint variable if we are not initializing it? Or, is it already removed?

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
```diff
@@ -336,16 +336,19 @@ public IEnumerator ManipulationHandlerOneHandMoveNear()
                 yield return hand.MoveTo(initialGrabPosition, numHandSteps);
                 yield return hand.SetGesture(ArticulatedHandPose.GestureId.Pinch);
 
+                yield return new WaitForFixedUpdate();
```
julenka: interesting, any idea why we need to add this?

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
```diff
@@ -336,16 +336,19 @@ public IEnumerator ManipulationHandlerOneHandMoveNear()
                 yield return hand.MoveTo(initialGrabPosition, numHandSteps);
```
julenka: It would be very helpful to add a test case for move/scale/rotate object when center != mesh center. I could see this as a feature that somebody could accidentally break in the future, we should protect against that since it was so hard to get working.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
MenelvagorMilsom: Yeah, my changes require a distinction between the pointer position, the grab position and the object position:

- pointer position: A point on the hand/controller defined by the type of pointer you're using
- grab position: A point on the object where the pointer grabbed the object
- object position: The origin point of the object

The way this was before, for near interaction the pointer position and the grab position would be equal. In fact, for near interaction pointer.Position returns the palm position. Now, the grab position is where you grabbed the object.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
MenelvagorMilsom: It's already removed:
image

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
MenelvagorMilsom: Oh, this hung around from when I was trying something. I've tried removing this and the tests still pass.

Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
MenelvagorMilsom: ### Difference in old(#ff00ff)  vs. new (#00ffff) grab points:
image

keveleigh: >for near interaction pointer.Position returns the palm position

Huh, interesting. I thought we moved the grab pointer to ~a point between your index finger and your thumb~ actually it looks like we just put it right at the position of the index finger? Via:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/3c98d9c582e90214fa4685dc02ae78553c36c492/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs#L87

and

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/3c98d9c582e90214fa4685dc02ae78553c36c492/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs#L120

image

MenelvagorMilsom: Both `ShellHandRayPointer` and `SpherePointer` use the `BaseControllerPointer` implementation of `Position`:
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/6081dcbb5f2eb1e89e26d2064e61920809bdbbb6/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/BaseControllerPointer.cs#L374

Neither pointers define a value for raycastOrigin in their prefabs, so this just returns the pointers transform.

The transform appears to be controlled by ControllerPoseSynchronizer (from which `BaseControllerPointer` inherits). The [`poseAction`](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/6081dcbb5f2eb1e89e26d2064e61920809bdbbb6/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ControllerPoseSynchronizer.cs#L83) inspector option seems to determine where on the controller this ends up being.

I guess if we were to use the pointer's `Rays` property, we could access this value here, but will it work for all pointers? Currently what we're doing here works grand.

julenka: For near interaction, the origin of the grab point should be the average of the index and thumb, I was checking this earlier but didn't notice that the average calculation was commented out, great catch @keveleigh. 

It would be really awesome if we could fix the sphere center computation from palm to average (index, thumb) before our GA release. @MenelvagorMilsom are you able to fix this, since I'm OOF?

julenka: @wiwei @thalbern and @cre8ivepark  FYI - it looks like we may be using the wrong position when doing hit testing for near interaction. It seems like we are using the index tip position when doing our sphere cast, when we should be using the average of index and thumb.

For doing actual movement / manipulation, using the palm (controller position) is a more stable point than the index and thumb, which is why we are having the transform controlled by ControllerPoseSynchronizer. 

However **the hit tests for computing which object is in focus / should receive input** should be done via average of index and thumb, to match the shell.

keveleigh: Ahh interesting. Since we use `pointerPosition` as the origin of the raycast due to its use in updating `Rays[0]`, it feels like we might have a missed setting or something somewhere to update the position to the same value.

>but will it work for all pointers

It should! Hopefully all pointers have their rays updated properly to where they should be raycasting from (which feels like it should be their position?).

>Currently what we're doing here works grand.

Sounds good!

MenelvagorMilsom: Created #5450 and will look into it soon.

PR #5404, opened 2019-07-22
Fix MRTK asset menu layout / remove duplicate menu

davidkline-ms:
This change consolidates duplicate asset menus and places all profile options into the profile menu.

Before:
image

After
image


keveleigh:
Should the MixedRealityInputRecordingProfile go under profiles as well?

davidkline-ms:
> Should the MixedRealityInputRecordingProfile go under profiles as well?

I didnt notice that one... sure, i can add it

PR #5405, opened 2019-07-22
Reduce the length of the names of certain files that exceed the MAX_PATH

matthejo:
The MAX_PATH being exceeded by these files is problematic when the MixedRealityToolkit is included as a submodule in other repositories. I'm certainly open to feedback on revisions to these names that still keep the total path short enough.

## Changes
- Fixes: # .


matthejo:
Going to start against mrtk_development and send a new PR in a bit.

PR #5408, opened 2019-07-23
Update spatial mesh level of detail values, translate in observer

davidkline-ms:
This is a minor breaking change (values in the SpatialMeshLevelOfDetail enum) to eliminate direct triangles per cubic meter encoding in the enum.

This change adds a Medium LOD as well as a lookup method in the windows mesh observer.

Verified by changing the LOD in the observer profile and deploying / running on HoloLens.

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
@@ -236,12 +236,45 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.SpatialAwareness
             {
                 if (value != SpatialAwarenessMeshLevelOfDetail.Custom)
                 {
-                    // For non-custom levels, the enum value is the appropriate triangles per cubic meter.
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
```diff
@@ -236,14 +236,47 @@ public SpatialAwarenessMeshLevelOfDetail LevelOfDetail
...
+                case SpatialAwarenessMeshLevelOfDetail.Coarse:
+                    triangleDensity = 0;
+                    break;
+
+                case SpatialAwarenessMeshLevelOfDetail.Medium:
+                    triangleDensity = 400;
+                    break;
+
+                case SpatialAwarenessMeshLevelOfDetail.Fine:
+                    triangleDensity = 200;
```
keveleigh: Is `Fine` intended to be less dense than `Medium`?

davidkline-ms: good catch. thanks!

Issue #5393, opened 2019-07-20
[Solver] Add degree/distance range to Orbital or Add Y-pos lock to Radial view


cre8ivepark:
## Describe the bug
To achieve shell-parity 'Follow Me' behavior, we need to either:
- Add degree/distance allowance range to Orbital 
- OR Add Y-pos lock to Radial view

Currently, with Orbital Solver, it is possible to place objects with specific Y offset, but it does not have a degree/distance allowance range option which allows lazily movement
Radial Solver has degree/distance range option but it cannot have fixed Y offset. When you move your head up, the object also follows.

Part of the shell-parity items #4200 
Related item: [#4472] Follow Me behavior: no examples available yet in MRTK.

## Degree/Distance allowance range in Radial Solver which is missing in Orbital Solver
This example makes an object stay within 0.3~0.7m and left/right 30degree threshold:
![2019-07-20 10_19_09-Unity 2018 4 2f1 Personal - HandInteractionExamples unity - MRTK-GitHub - Univer](https://user-images.githubusercontent.com/13754172/61572403-dfbe6c00-aad7-11e9-85d2-cfb234fcba16.png)

Alexees:
Wasn't the idea of those solvers to chain them? What about adding Radial first, enabling UpdateLinkedTransform in the inspector, and Orbital second?

wiwei:
@keveleigh, do you have some background on @Alexees  question?

Issue #5380, opened 2019-07-19
Toggle features panel solver moves menu directly in front of face


julenka:
Having menus directly in front of the face is not great for comfort, but the solver in the toggle features panel moves the menu to be right in front of the face. We should update the params of the solver to be closer to the waist.

julenka:
Marking at GA blocker because toggle menu is usually the script people use to implement follow menu behavior (they copy / paste that prefab), so this prefab needs to have the recommended following behavior.

cre8ivepark:
Updated Radial Solver to match the shell's Follow Me behavior. Testing out on the device and tuning the values now.

Issue #4983, opened 2019-06-20
[MRTK Shell Parity] Follow Me behavior does not match the shell


cre8ivepark:
## Describe the bug
Follow Me behavior does not match the shell.
Currently, Slate prefab has 'Follow Me' button which toggles Orbital Solver to mimic the behavior of the shell. However, the behavior is different from the shell.
The biggest difference is: current behavior with orbital solver does not provide 'lazy move' - allow the user to move/rotate a certain amount before it triggers object repositioning.

Part of #4200 UX Controls Parity with HoloLens 2 Shell

## To reproduce
In HandInteractionExamples scene, press 'Follow Me' button on the Slate example.

## Branch
mrtk_development

PR #5412, opened 2019-07-24
[Shell-Parity] Radial Solver update for the 'Follow Me' behavior

cre8ivepark:
## Overview
To achieve shell-parity 'Follow Me' behavior, added vertical position lock option to the Radial view Solver. Radial view already had min/max distance/degree allowance which allowed shell-style 'lazy following' behavior. We just needed to a way to lock vertical position. With original Radial view, object also moved vertically following the headset movement.

- Applied updated Radial view Solver to Slate.prefab and ToggleFeaturesPanel.prefab
- Optimized the min/max distance/degree allowance of the Radial view Solver to match shell behavior

2019-07-23 17_09_12-Shell-Parity

Part of #4200 [MRTK Shell Parity] UX Controls Parity with HoloLens 2 Shell

## Fixes
- #5380 Toggle features panel solver moves menu directly in front of face
- #4983 [MRTK Shell Parity] Follow Me behavior does not match the shell
- #5393 [Solver] Add degree/distance range to Orbital or Add Y-pos lock to Radial view

## MRCs
Lazy following with left/right degree values & min max distance values.
MRTK_FollowMeUpdate1
MRTK_FollowMeUpdate2

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Slate/Slate.prefab
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Slate/Slate.prefab
@@ -1036,14 +1036,14 @@ MonoBehaviour:
   m_PrefabInstance: {fileID: 0}
   m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 7779420039263858270}
-  m_Enabled: 1
+  m_Enabled: 0
   m_EditorHideFlags: 0
   m_Script: {fileID: 11500000, guid: 4684083f6dff4a1d8a790bccc354fcf4, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
   updateLinkedTransform: 0
-  moveLerpTime: 0.3
-  rotateLerpTime: 0.3
+  moveLerpTime: 0.5
+  rotateLerpTime: 0.5
   scaleLerpTime: 0
   maintainScale: 1
   smoothing: 1
@@ -1053,12 +1053,12 @@ MonoBehaviour:
   minDistance: 0.4
   maxDistance: 1
   minViewDegrees: 0
-  maxViewDegrees: 20
+  maxViewDegrees: 25
   aspectV: 1
   ignoreAngleClamp: 0
   ignoreDistanceClamp: 0
   useFixedVerticalPosition: 1
-  fixedVerticalPosition: -0.1
+  fixedVerticalPosition: -0.2
   orientToReferenceDirection: 0
 --- !u!65 &7779420039263858272
 BoxCollider:
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Slate/Slate.prefab
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Slate/Slate.prefab
@@ -216,7 +216,6 @@ MonoBehaviour:
   momentumHorizontal: 0.9
   momentumVertical: 0.9
   panZoomSmoothing: 80
-  panEventReceivers: []
   centerPoint: {fileID: 0}
   leftPoint: {fileID: 0}
   rightPoint: {fileID: 0}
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/FollowMeToggle.cs
```diff
@@ -12,25 +12,25 @@ public class FollowMeToggle : MonoBehaviour
...
         private GameObject visualizationObject = null;
 
-        private Orbital orbital = null;
+        private RadialView radialView = null;
 
         private void Start()
         {
             // Get Orbital Solver component
-            orbital = GetComponent<Orbital>();
+            radialView = GetComponent<RadialView>();
```
Cameron-Micka: Since it appears this class requires a RadialView present to do anything. Maybe we should require one? https://docs.unity3d.com/ScriptReference/RequireComponent.html

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/RadialView.cs
```diff
@@ -196,6 +222,11 @@ public override void SolverUpdate()
                 goalRotation.x = goalRotation.z = 0f;
             }
 
+            if(UseFixedVerticalPosition == true)
```
Cameron-Micka: Nit: spacing after if is missing.

Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Slate/Slate.prefab
```diff
@@ -254,8 +253,13 @@ MonoBehaviour:
           m_StringArgument: 
           m_BoolArgument: 0
         m_CallState: 2
-    m_TypeName: UnityEngine.Events.UnityEvent, UnityEngine.CoreModule, Version=0.0.0.0,
-      Culture=neutral, PublicKeyToken=null
+    m_TypeName: Microsoft.MixedReality.Toolkit.Input.PanUnityEvent, Microsoft.MixedReality.Toolkit.SDK,
```
wiwei: Was this switch to PanUnityEvent something that was swapped automatically during the save? This seems like it would be a change that's unrelated to this radial solver update at first glance.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/FollowMeToggle.cs
wiwei: I think it does looook to be optional (i.e. you could in theory have a slate that doesn't follow you, like in the hand interaaction example scenes case)

The only thing that might be helpful is, if the ToggleFollowMeBehavior is engaged, to have some sort of warning or logging if radialView is null (meaning it's not been set up correctly for follow me behavior.

Alternatively, we could try to loosen the dependency directly on RadialView (i.e. maybe ANY time of "following" solver would work here, and we could enable/disable a general solver object instead of this specific one). I'm trying to imagine a consumer wanting to have their own follow me behavior and choosing a non-radial solver (or doing their own)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/RadialView.cs
wiwei:
```suggestion
            if (UseFixedVerticalPosition)
```

wiwei: It's a boolean so we don't need to do the direct comparison to true (generally if params are named well like they are here, it can be fairly readable to be like "if using a fixed vertical position then..."

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/FollowMeToggle.cs
Cameron-Micka: I'm not so sure it is optional, since they should remove this component if a solver isn't present on this object. But, I agree the logic should switch to the base solver class. In which case we can't use require component. haha

cre8ivepark: @Cameron-Micka Made it Required component. Thanks!

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/RadialView.cs
cre8ivepark: Applied the update!

Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Slate/Slate.prefab
cre8ivepark: It seems like it was automatically updated by Unity.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/RadialView.cs
cre8ivepark: Updated

PR #5415, opened 2019-07-24
Update the build tools to use the new vswhere flag to find msbuild

wiwei:
The build tooling currently has some issues finding instances of msbuild (especially on certain VS2019 installs)

This change adopts the change @julianloehr-kg from:
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5356

With a modification to still fall back to the old one if the new one fails (this can happen if building in VS2017, like what happens on our build machines).

Thanks to @julianloehr-kg for this!

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
@@ -163,7 +163,11 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
 
         private static async Task<string> FindMsBuildPathAsync()
         {
-            var result = await new Process().StartProcessAsync(
+            // Finding msbuild.exe involves different work depending on whether or not users
+            // have VS2017 or VS2019 installed.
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -163,33 +163,42 @@ private static async Task<int> Run(string fileName, string args, bool showDebug,
...
+                var result = await new Process().StartProcessAsync(
                 new ProcessStartInfo
                 {
                     FileName = "cmd.exe",
                     CreateNoWindow = true,
                     UseShellExecute = false,
                     RedirectStandardOutput = true,
                     RedirectStandardError = true,
-                    Arguments = $@"/C vswhere -all -products * -requires Microsoft.Component.MSBuild -property installationPath",
+                    Arguments = findOption.arguments,
```
keveleigh: If we pass this an argument it doesn't support (`-find`), does it error or just not find anything?

wiwei: It reports an error back through stdout from what I remember

Issue #5141, opened 2019-06-29
SolverHandler: Changing transformTarget does not recalculate the Solver


Alexees:
## Describe the bug

The SolverHandler does not update it's inner workings when assigned a new TransformTarget. It's neither recalculating it's transformWithOffset nor is there any API call that would make changing what to operate on easy. There is also no option for a custom reference in TrackedObjectType

## To reproduce

There is not much to reproduce other than checking out the API

## Expected behavior

Setting a new TransformTarget recalculates the Solver to use it

## Additional context

I have a menu that jumps around depending on where it's used.

Issue #5143, opened 2019-06-30
InBetweenInspector/-Editor serializes non existing field


Alexees:
## Describe the bug

The following field does not exist and causes error when choosing HandJointLeft or -Right in "Tracked Object For Second Transform" in the inspector:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/ab6bfc0b97a59d2ba76b2f1695c43115aba3bb63/Assets/MixedRealityToolkit.SDK/Inspectors/Utilities/Solvers/InBetweenInspector.cs#L23

I don't know what this did so I don't know if this could just be removed as the Inspector code is explicitly using this:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/ab6bfc0b97a59d2ba76b2f1695c43115aba3bb63/Assets/MixedRealityToolkit.SDK/Inspectors/Utilities/Solvers/InBetweenInspector.cs#L42-L46

Additional note: InBetweenInspector is the name of the file, InBetweenEditor the name of the class

wiwei:
Seems like dead code to me - a good candidate for someone getting started here making a contribution

Issue #5110, opened 2019-06-28
Allow manual ordering of Solvers for a SolverHandler


romsahel:
## Describe the problem
As noted in the ["Known Issues"](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/README_Solver.html#known-issues), we currently have to set the Solvers' order of execution in case of a problem: it is neither practical nor does it allow it to be done per-instance. 

## Describe the solution you'd like
Currently, the list of solvers is obtained `GetComponents<Solver>()`: exposing the list so that it can be manually ordered within the inspector and from the code would solve this issue.

## Additional context
Solvers

keveleigh:
The list should be returned in the order they're positioned in the Inspector (higher = earlier), so I don't think script execution order needs to be changed. I'll update the known issue to clarify that.

Though! There are some solver improvements planned, so making this list more explicit definitely sounds like something that should be considered.

romsahel:
Oh, alright. Documentation seemed to talk about the _script execution order_! I apologize, I should've checked :), I didn't know the order of elements returned by `GetComponents` depended of the order in which they are added.

> making this list more explicit definitely sounds like something that should be considered.

Indeed! I can see use cases where the list of Solvers should be determined at runtime, after the initialization of the SolverHandler.

Issue #4788, opened 2019-06-09
Solvers (i.e RadialView) cannot be instantiated at runtime


Troy-Ferrell:
## Describe the bug

If loading/building models & gameobjects dynamically, calling AddComponent<RadialView> or similar does not work. 

Solvers require the "SolverHandler" component which in Awake calls

`solvers.AddRange(GetComponents<Solver>());`

After debugging, solvers array is empty. Guessing that AddComponent<RadialView> calls AddComponent<SolverHandler> but the RadialView component has not been added yet, so this array is empty. Further, this array can never be repopulated so on update the list of solvers to call is always empty.

## To reproduce

Create a component which calls AddComponent<RadialView> and notice that the transform never gets updated or debug into the code and breakpoint

## Expected behavior

Like other MRTK components, developers should have the ability to add and remove components at runtime dynamically.

## Your Setup (please complete the following information)

- Unity Version 2018.4
- MRTK Version RC2-Stabilization

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

PR #5416, opened 2019-07-24
Refactor SolverHandler to fix multiple issues

Troy-Ferrell:
## Overview

The solver system has many issues (most of which are posted on GitHub). One of the key limitations for the feature is involved in the SolverHandler class. This change addresses multiple issues:

- Cleaner understanding of what kind of target type is active
=> Simpler types => Head, MotionController, HandJoint, & CustomOverride
=> Inspector now shows valid sub-properties based on the target type

- Developers can define order of solvers in code (via SetSolvers)
- Solvers can now be correctly tracked/instantiated at runtime dynamically
- Solvers can now follow any hand or motion controller (defaults to Left first if set to Both/Any)
- Removed OnValidate from Solver.cs as this is generally poor practice and give inconsistent results
- Removed nonexisting property for InBetween.cs
- Added solvertests to tests package (See verification section below for details)

- Fixed SurfaceMagnetism 
=> Renamed properties to not assume always head tracked transform
=> Created new inspector to make properties easier to understand & group
=> Updated MaxDistance for raycast default from 3m to 50m
=> Updated default Raycast Direction Mode to Tracked Target Forward (ray along head gaze or hand pointer, etc)
- Fixed InBetween
=> Updated names to be clearer
=> Updated refresh of second solver to be accurate
=> Updated tooltips to be accurate
=> Modified inspector to extend from SolverInspector

**NOTE:** This changes property names and will break any existing code that tried to extend Solvers themselves

New Solver Handler Targeting
tracked-target-type

Hand Switching Feature:
hand-switching

New Editor Inspectors:
surface-mag-inspector

InBetween

New Solver Tests:
image

## Changes
- Fixes: #5141 , #5110 , #4642 , #4788 , #5143 

## Verification
Added Solver Tests! 

Furthermore, I will follow up this PR with another one to fix the Solver Example scene which is mostly broken.
- Test dynamic instantiation of solver (Orbital)
- Test switch hands (Orbital)
- Test switch target types at runtime (Orbital)
- Test SurfaceMagnetism (simple raycast against "wall" tracked against head*)
- Test InBetween (between two cubes & modifies PartwayOffset)

Tested basic functionality manually
- VR Motion Controllers
- Editor Hands Simulation
- HL2 Hands on Device
- Custom override with transform in scene
- Dynamically add component to gameobject (both in scene editor & via c# code)
- Modify SolverHandler values in Inspector at runtime

>
>


Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
```diff
@@ -154,16 +154,13 @@ protected set
 
         #region MonoBehaviour Implementation
 
-        protected virtual void OnValidate()
+        protected virtual void Awake()
```
Cameron-Micka: Nit: Now that this happens in Awake (which is way more correct) we could remove the [SerializeField] attribute from SolverHandler since it will be null at serialization/deserialization time.

Troy-Ferrell:
@keveleigh / @wiwei  are there any steps I should look into for a breaking change PR? 

I know I need to update documentation but I will do that in the follow up PR. I need to fix the examples scene as well. This PR is already getting pretty big and I don't want to include those two steps here.

Troy-Ferrell:
> Looks great! I merged this into my HandConstrait experimental to release branch and everything worked/merged nicely.
> 
> I tried the example scene (which was pretty broken before this PR) and everything behaved as I would expect except the surface magnetism solver. Might be worth a sanity check since I'm not 100% sure how it should behave?
> 
> Could you also add a note to:
> https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html for any breaking changes?

yes, the example scene has been broken. I want to do a follow up PR for both the documentation and the example scene updates. This current PR is just getting very big

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
Troy-Ferrell: Done!

Troy-Ferrell:
Updated PR to point to prerelease stabilization branch

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -77,28 +77,28 @@ public LayerMask[] MagneticSurfaces
 
         [SerializeField]
         [Tooltip("Max distance for raycast to check for surfaces")]
-        private float maxDistance = 3.0f;
+        private float maxRaycastDistance = 50.0f;
```
wiwei: >50.0 [](start = 43, length = 4)

I think 50 is further than the default raycast distance of 10 for things like the hand ray right?

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -77,28 +77,28 @@ public LayerMask[] MagneticSurfaces
 
         [SerializeField]
         [Tooltip("Max distance for raycast to check for surfaces")]
```
wiwei: Please using FormerlySerializedAs to avoid breaking previous users of this script

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -77,28 +77,28 @@ public LayerMask[] MagneticSurfaces
...
+        public float MaxRaycastDistance
         {
-            get { return maxDistance; }
-            set { maxDistance = value; }
+            get { return maxRaycastDistance; }
+            set { maxRaycastDistance = value; }
         }
 
         [SerializeField]
         [Tooltip("Closest distance to bring object")]
```
wiwei: Please using FormerlySerializedAs to avoid breaking previous users of this script

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -227,11 +227,11 @@ public bool UseLinkedAltScaleOverride
         }
 
         [SerializeField]
-        [Tooltip("Raycast direction. Can cast from head in facing direction, or cast from head to object position")]
-        private RaycastDirectionMode currentRaycastDirectionMode = RaycastDirectionMode.ToLinkedPosition;
+        [Tooltip("Raycast direction type. Default is forward direction of Tracked Target transform")]
+        private RaycastDirectionMode currentRaycastDirectionMode = RaycastDirectionMode.TrackedTargetForward;
```
wiwei: >TrackedTargetForward [](start = 88, length = 20)

What's the reason for changing the default mode of this solver?

This has a risk of breaking someone in a really hard to debug way... though I guess technically if they've already serialized these scenes it should have saved the old default value.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -77,28 +77,28 @@ public LayerMask[] MagneticSurfaces
...
         [SerializeField]
         [Tooltip("Max distance for raycast to check for surfaces")]
-        private float maxDistance = 3.0f;
+        private float maxRaycastDistance = 50.0f;
 
         /// <summary>
         /// Max distance for raycast to check for surfaces
         /// </summary>
-        public float MaxDistance
+        public float MaxRaycastDistance
```
wiwei: >MaxRaycastDistance [](start = 21, length = 18)

Please add a note on Solver migration in GA here:

[https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -77,28 +77,28 @@ public LayerMask[] MagneticSurfaces
...
         [SerializeField]
         [Tooltip("Closest distance to bring object")]
-        private float closeDistance = 0.5f;
+        private float closestDistance = 0.5f;
 
         /// <summary>
         /// Closest distance to bring object
         /// </summary>
-        public float CloseDistance
+        public float ClosestDistance
```
wiwei: >ClosestDistance [](start = 21, length = 15)

Please add a note on Solver migration in GA here:

[https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/UpdatingToGA.html)

wiwei:
>     public enum TrackedObjectType

The changes to this enum are a bitttttt scarry because they feel so central to existing solver usage.

Do you know if the existing usages of enums are actually serialized as enum names, or enum values? i.e. if someone had configured to have a solver follow the MotionControllerRight before, when they load up their serialized asset, will they suddenly get HandJoint?

If this is the case I would almost rather keep the old values around deprecated with a warning to use the new ones (we'd have to do some "map old to new helper" in order to make use of them).

It's on me to write up the plan for how we get rid of things that are deprecated (i.e. timeline).

---
Refers to: Assets/MixedRealityToolkit/Definitions/Utilities/TrackedObjectType.cs:6 in 3aeb246. [](commit_id = 3aeb2469d2ddb7732966b0fb0942d446796678b6, deletion_comment = False)

wiwei:
> > ```
> > public enum TrackedObjectType
> > ```
> 
> The changes to this enum are a bitttttt scarry because they feel so central to existing solver usage.
> 
> Do you know if the existing usages of enums are actually serialized as enum names, or enum values? i.e. if someone had configured to have a solver follow the MotionControllerRight before, when they load up their serialized asset, will they suddenly get HandJoint?
> 
> If this is the case I would almost rather keep the old values around deprecated with a warning to use the new ones (we'd have to do some "map old to new helper" in order to make use of them).
> 
> It's on me to write up the plan for how we get rid of things that are deprecated (i.e. timeline).
> 
> Refers to: Assets/MixedRealityToolkit/Definitions/Utilities/TrackedObjectType.cs:6 in [3aeb246](https://github.com/microsoft/MixedRealityToolkit-Unity/commit/3aeb2469d2ddb7732966b0fb0942d446796678b6). [](commit_id = [3aeb246](https://github.com/microsoft/MixedRealityToolkit-Unity/commit/3aeb2469d2ddb7732966b0fb0942d446796678b6), deletion_comment = False)

@keveleigh for your thoughts

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/InBetween.cs
```diff
@@ -27,35 +27,49 @@ public float PartwayOffset
         [SerializeField]
         [Tooltip("Tracked object to calculate position and orientation for the second object. If you want to manually override and use a scene object, use the TransformTarget field.")]
         [HideInInspector]
-        private TrackedObjectType trackedObjectForSecondTransform = TrackedObjectType.Head;
+        private TrackedObjectType secondTrackedObjectType = TrackedObjectType.Head;
```
wiwei: >private TrackedObjectType secondTrackedObjectType = TrackedObjectType.Head; [](start = 8, length = 75)

Please also use FormerlySerializedAs here as well

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/InBetween.cs
```diff
@@ -27,35 +27,49 @@ public float PartwayOffset
...
 
         /// <summary>
         /// Tracked object to calculate position and orientation for the second object. If you want to manually override and use a scene object, use the TransformTarget field.
         /// </summary>
-        public TrackedObjectType TrackedObjectForSecondTransform
+        public TrackedObjectType SecondTrackedObjectType
```
wiwei: >SecondTrackedObjectType [](start = 33, length = 23)

Please add a migration note in the GA doc for this change.

I'm kinda torn between keeping things as deprecated and having them reference this - it might be good to be consistent with that (i.e. in all of these public field renames, keep the old around, mark as deprecated, put it as a task on me to do the post-deprecation deletion process). What do you think?

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/InBetween.cs
```diff
@@ -91,23 +106,13 @@ private void UpdateSecondSolverHandler()
...
-                    secondSolverHandler.TrackedObjectToReference = trackedObjectForSecondTransform;
+                    secondSolverHandler.SetTransformOverride(secondTransformOverride);
                 }
             }
         }
-
-        /// <summary>
-        /// This should only be called from the SolverInBetweenEditor to cause the secondary SolverHandler to reattach this solver.
-        /// </summary>
-        public void AttachSecondTransformToNewTrackedObject()
```
wiwei: >AttachSecondTransformToNewTrackedObject [](start = 20, length = 39)

Please add migration notes on how to accomplish this now (which is just updating the property itself!)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -10,45 +10,72 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
     /// <summary>
     /// This class handles the solver components that are attached to this <see href="https://docs.unity3d.com/ScriptReference/GameObject.html">GameObject</see>
     /// </summary>
-    public class SolverHandler : ControllerFinder
```
wiwei: >ControllerFinder [](start = 33, length = 16)

Nit: Also work having a section on the migration docs for this (i.e. no longer extends ControllerFinder, here's what you do instead!)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -10,45 +10,72 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
...
+    public class SolverHandler : MonoBehaviour
     {
         [SerializeField]
         [Tooltip("Tracked object to calculate position and orientation from. If you want to manually override and use a scene object, use the TransformTarget field.")]
-        private TrackedObjectType trackedObjectToReference = TrackedObjectType.Head;
+        private TrackedObjectType trackedTargetType = TrackedObjectType.Head;
```
wiwei: >trackedTargetType [](start = 34, length = 17)

Please add FormerlySerializedAs here too!

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -10,45 +10,72 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
...
 
         /// <summary>
         /// Tracked object to calculate position and orientation from. If you want to manually override and use a scene object, use the TransformTarget field.
         /// </summary>
-        public TrackedObjectType TrackedObjectToReference
+        public TrackedObjectType TrackedTargetType
```
wiwei: >TrackedTargetType [](start = 33, length = 17)

Same comment about the general keep old feild around (but deprecated) vs do the new thing and write migration docs on it.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -10,45 +10,72 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
...
-                    trackedObjectToReference = value;
+                    trackedTargetType = value;
                     RefreshTrackedObject();
                 }
             }
         }
 
         [SerializeField]
-        [Tooltip("When using the hand to calculate position and orientation, use this specific joint.")]
+        [Tooltip("If tracking hands or motion controllers, determines which hand(s) are valid attachments")]
```
wiwei: >If tracking hands or motion controllers, determines which hand(s) are valid attachments [](start = 18, length = 87)

I think the valid options are Left, Right, Both... right?

Because there are other options in the enum, it might be worth mentioning what we actually support (i.e. do we support "other?")

wiwei: This has come up in the past for other uses of this, and when taking an enum we either support EVERYTHING on it, or we should say what we do or do not support

---
In reply to: [308995148](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5416#discussion_r308995148) [](ancestors = 308995148)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -10,45 +10,72 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
...
+                {
+                    trackedHandJoint = value;
+                    RefreshTrackedObject();
+                }
             }
         }
 
+        [SerializeField]
+        [Tooltip("Manual override for when TrackedTargetType is set to CustomOverride")]
+        private Transform transformOverride;
```
wiwei: >private Transform transformOverride; [](start = 8, length = 36)

Out of curiousity, why was the runtime setter for this property done as a function (i.e. SetTransformOverride) instead of as a property like the other ones above?

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -109,6 +129,28 @@ public bool UpdateSolvers
...
+        protected readonly List<Solver> solvers = new List<Solver>();
+
+        /// <summary>
+        /// List of solvers that this handler will manage and update
+        /// </summary>
+        public IReadOnlyCollection<Solver> Solvers
+        {
+            get
+            {
+                return solvers.AsReadOnly();
```
wiwei: > solvers.AsReadOnly(); [](start = 22, length = 22)

Just to check here - can you double check that this doesn't allocate a new list on each call? If not, you can also I thik just cast this to a read only type list, which will accomplish the same thing without the memory allocation.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -134,17 +176,62 @@ public bool UpdateSolvers
...
+        }
 
-        private float lastUpdateTime;
 
-        private GameObject transformWithOffset;
+        // Stores currently attached hand if valid (only possible values Left, Right, or None)
+        protected Handedness currentTrackedHandedness = Handedness.None;
+
+        /// <summary>
+        /// Currently tracked hand or motion controller if applicable
```
wiwei: >Currently tracked hand or motion controller if applicable [](start = 12, length = 57)

Nit: Add a remarks section which contains the comments from above:

These are super valuable to have in autocomplete

// Stores currently attached hand if valid (only possible values Left, Right, or None)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
wiwei: Asking mostly out of consistency, not because one way is better than the other

---
In reply to: [308995399](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5416#discussion_r308995399) [](ancestors = 308995399)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -134,17 +176,62 @@ public bool UpdateSolvers
...
+        {
+            get
+            {
+                if (inputSystem == null)
+                {
+                    MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out inputSystem);
+                }
+                return inputSystem;
+            }
+        }
```
wiwei: I wish your static helper thingy went in....

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -204,89 +283,154 @@ protected void OnDestroy()
...
+                    target = GetMotionController(Handedness.Left);
+                    if (target == null)
+                    {
+                        this.currentTrackedHandedness = Handedness.Right;
+                        target = GetMotionController(Handedness.Right);
+                        if (target == null)
+                        {
+                            this.currentTrackedHandedness = Handedness.None;
+                        }
+                    }
```
wiwei: Does this mean that if I have both controllers in the scene and I call RefreshTrackedObject, it will always prefer the left? Is there any way for the user to specify the handedness preference? I could just be misreading this :)

wiwei: Same comment on the hand joint stuff down below.

---
In reply to: [308996330](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5416#discussion_r308996330) [](ancestors = 308996330)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -204,89 +283,154 @@ protected void OnDestroy()
...
+                        {
+                            currentTrackedHandedness = Handedness.Right;
+                        }
+                        else
+                        {
+                            currentTrackedHandedness = Handedness.None;
+                        }
+                    }
+
+                    target = HandJointService?.RequestJointTransform(this.TrackedHandJoint, currentTrackedHandedness);
```
wiwei: >HandJointService [](start = 29, length = 16)

I think the ? isn't necessary since you already check for null above

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -204,89 +283,154 @@ protected void OnDestroy()
...
+            }
+
+            return null;
         }
 
-        private Transform MakeOffsetTransform(Transform parentTransform)
+        /// <summary>
+        /// Returns true if the solver handler's transform target is not valid
+        /// </summary>
+        /// <returns></returns>
```
wiwei: Nit: Delete empty returns clause

Assets/MixedRealityToolkit.SDK/Inspectors/Utilities/Solvers/InBetweenInspector.cs
```diff
@@ -8,53 +8,50 @@
...
-            DrawPropertiesExcluding(serializedObject, fieldsToExclude);
+            EditorGUILayout.PropertyField(partwayOffsetProperty);
 
             serializedObject.ApplyModifiedProperties();
 
-            if (Application.isPlaying && trackedObjectChanged)
+            if (Application.isPlaying && objectChanged)
             {
-                solverInBetween.AttachSecondTransformToNewTrackedObject();
+                //solverInBetween.AttachSecondTransformToNewTrackedObject();
```
wiwei: Was this meant to be uncommented (or can the if clause be deleted)?

Assets/MixedRealityToolkit.SDK/Inspectors/Utilities/Solvers/SurfaceMagnetismInspector.cs
```diff
@@ -0,0 +1,112 @@
...
+            EditorGUILayout.PropertyField(magneticSurfacesProperty, true);
+            EditorGUILayout.PropertyField(closestDistanceProperty);
+            EditorGUILayout.PropertyField(maxDistanceProperty);
+            EditorGUILayout.PropertyField(currentRaycastDirectionModeProperty);
+            EditorGUILayout.PropertyField(raycastModeProperty);
+
+            if (raycastModeProperty.enumValueIndex != (int)SceneQueryType.SphereOverlap)
+            {
+                if (raycastModeProperty.enumValueIndex != (int)SceneQueryType.SimpleRaycast)
+                {
```
wiwei: Consider deleting this line, since you check for the specific things you care about down below

OR

Alternatively, this is actually a pretty good candidate for a switch statement like:

switch (raycastModeProperty.enumValueIndex):
case (int)SceneQueryType.SphereOverlap:
   InspectorUIUtility.DrawWarning("SurfaceMagnetism does not support SphereOverlap raycast mode");
   break;
case (int)SceneQueryType.BoxRaycast:
   EditorGUILayout.PropertyField(boxRaysPerEdgeProperty);
   EditorGUILayout.PropertyField(orthographicBoxCastProperty);
   EditorGUILayout.PropertyField(maximumNormalVarianceProperty);
   break;
...

It puts the code assocaited with the condtiion really close by (especially for the SphereOverlap case, where the warning message is currently pretty far from the if check)

wiwei: This is a nit btw, it's just usually when I see a  lot of if statements it can be kinda like owaw

---
In reply to: [308997604](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5416#discussion_r308997604) [](ancestors = 308997604)

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
```diff
@@ -0,0 +1,246 @@
...
+            wall.transform.position = Vector3.forward * 10.0f;
+
+            yield return null;
+
+            // Instantiate our test gameobject with solver. 
+            // Set layer to ignore raycast so solver doesn't raycast itself (i.e BoxCollider)
+            var testObjects = InstantiateTestSolver<SurfaceMagnetism>();
+            testObjects.target.layer = LayerMask.NameToLayer("Ignore Raycast");
+
+            yield return new WaitForSeconds(SolverUpdateWaitTime);
```
wiwei: >WaitForSeconds(SolverUpdateWaitTime) [](start = 29, length = 36)

Does this actually add 2 real seconds of wait per test? Or is this "game time" which can be configured to move faster than normal?

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
```diff
@@ -0,0 +1,246 @@
...
+            // Test orbital around right hand
+            yield return TestHandSolver(testObjects.target, inputSimulationService, rightHandPos, Handedness.Right);
+
+            // Test orbital around head
+            testObjects.handler.TrackedTargetType = TrackedObjectType.Head;
+
+            yield return new WaitForSeconds(SolverUpdateWaitTime);
+
+            Assert.LessOrEqual(Vector3.Distance(testObjects.target.transform.position, Vector3.zero), DistanceThreshold);
+
```
wiwei: Can you add a comment about where the camera is located (i.e. is it at 2.5m?) so that this check can have a little more context about what it's about

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
```diff
@@ -0,0 +1,246 @@
...
+            yield return null;
+
+            InputSimulationService inputSimulationService = PlayModeTestUtilities.GetInputSimulationService();
+
+            // Test orbital around right hand
+            yield return TestHandSolver(testObjects.target, inputSimulationService, rightHandPos, Handedness.Right);
+
+            // Test orbital around left hand
+            yield return TestHandSolver(testObjects.target, inputSimulationService, leftHandPos, Handedness.Left);
```
wiwei: Would also be interesting have a test that shows what happens when both hands are in the scene - if people start depending on this behavior, we should have a test for that

wiwei:
Thank you SO much for doing this.

I have a bunch of comments, but I think we can work through them and get them addressed reasonably. 

I wanted to call out that this is a canonical example of a well done PR that includes great tests for features that should have tests a long time ago.

Because of this I feel 100% more confident making changes to solvers going forward

Cameron-Micka:
@wiwei just a heads up my upcoming hand constraint pull request depends on this pull request. So, please let me know if you and Troy need help getting this in.

Troy-Ferrell:
@wiwei thanks for the very helpful comments! I know it's a substantial change. 

@Cameron-Micka I just skimmed through most of the items and I think this is very doable to cover pretty quickly tomorrow morning. It also actually gave me a potential idea to help people with the migration/breaking change. It's particularly around that target type enum

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
Troy-Ferrell: @wiwei  Yes there is actually a related issue #5269 

Although I don't understand why 10m? That's not very useful for VR and you could have large items in a scene

It seems like something people will keep running into this and it's not obvious how to debug. The SurfaceMagnetism in the example scene I think was actually broken because of this property being set to 3m.

Think of it this way:

If I try to use the component out-of-box and raycast to a wall 20m away (AR or VR or hands or head or other raycast type), it doesn't work. It's not clear where to debug and why it doesn't work

If I can use my hands to raycast a small item to a wall 50m away and it's not very accurate or useful for my scenario, it's more obvious that I just need to limit my raycast to certain layermask or to limit distance

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
Troy-Ferrell: gah didn't know FormerlySerializedAs existed! so useful! thanks

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
Troy-Ferrell: @wiwei I actually want to consider looking at removing linkedposition lol. It is the default and it doesn't work. It generates a raystep of 0,0,0 to 0,0,0. 

This shouldn't break customers though because the linkedposition enum is still there if they selected it. Only new components being added will be impacted which is a smaller set of customers...who again I hope were switching it anyways to get it to work. 

I actually know of one customer who may have been dynamically doing it. I'll have to check

Troy-Ferrell:
> > ```
> > public enum TrackedObjectType
> > ```
> 
> The changes to this enum are a bitttttt scarry because they feel so central to existing solver usage.
> 
> Do you know if the existing usages of enums are actually serialized as enum names, or enum values? i.e. if someone had configured to have a solver follow the MotionControllerRight before, when they load up their serialized asset, will they suddenly get HandJoint?
> 
> If this is the case I would almost rather keep the old values around deprecated with a warning to use the new ones (we'd have to do some "map old to new helper" in order to make use of them).
> 
> It's on me to write up the plan for how we get rid of things that are deprecated (i.e. timeline).
> 
> Refers to: Assets/MixedRealityToolkit/Definitions/Utilities/TrackedObjectType.cs:6 in [3aeb246](https://github.com/microsoft/MixedRealityToolkit-Unity/commit/3aeb2469d2ddb7732966b0fb0942d446796678b6). [](commit_id = [3aeb246](https://github.com/microsoft/MixedRealityToolkit-Unity/commit/3aeb2469d2ddb7732966b0fb0942d446796678b6), deletion_comment = False)

I'm 99% sure they are serialized as numerical values. I'll double check though. Yes your points give me an idea. We could keep the old values, mark as obsolete, and then add on the new values I listed. Then in either OnValidate or in the inspector/awake, check if they are deprecated and just auto switch to the correct value and modify TrackedHandedness as well

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
Troy-Ferrell: @wiwei it is real time. We can & should totally use game time though. Although I think the solvers are being updated against realtimeSinceStartup

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
Troy-Ferrell: Just switched comparison to be the actual head transform in case it changes

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
Troy-Ferrell: @wiwei  yea problem is TimeScale ( https://docs.unity3d.com/ScriptReference/Time-timeScale.html ) does not affect realtimesincestartup which is used by the solvers

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/InBetween.cs
Troy-Ferrell: I think that is reasonable. It's more work on our side as developers but it should make it easier for consumers. They can replace all references in their framework to the new name probably more easily since VS can still resolve the reference

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
Troy-Ferrell: (y)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
Troy-Ferrell: Because there is no getter so I made a set function. But it does feel better to make it a public property setter instead

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
Troy-Ferrell: Yes, it's a readonly wrapper around the existing collection
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.asreadonly?view=netframework-4.8

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
Troy-Ferrell: Me too, I am still working on it. Just trying to resolve some gaps in MRTK.cs and the Registry

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
Troy-Ferrell: It will always prefer Left if the TrackedHandedness value is set to Both. If not, then it will try to get the controller or hand of TrackedHandedness value (i.e right or left)

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
Troy-Ferrell: Added to TestHandModality

Issue #4914, opened 2019-06-17
Prefab and meta serialization is not up to date


gilbdev:
## Describe the bug
prefab, meta, and scene assets are not automatically updated in Unity.
the serialization of those files can change when:
- unity version is updated
- some code of the components used in prefab has changed on the serialization part (renaming adding,removing serialized field)

currently, if i make a change in some of the mrtk prefabs, a whole list of diffs will come up.

bonus effect: if you make a tool (see below) in the mrtk, this can be useful when people are upgrading version on their side (to reserialize their profiles and scenes for example)

## Expected behavior

when a new mrtk release is pushed, all files should be up to date regarding serialization, at least the changes regarding mrtk code. (untiy serialization changes is a subject to be discussed, i don't know the exact repercussion for retrocompatibilities).

##
this process can be automated:
(i can provide a better version that displays a progress bar)
```

[MenuItem("Assets/Reserialize Prefabs&Scenes...")]
private static void ReserializeAll()
{
	var array = GetAssets("t:Prefab t:Scene t:ScriptableObject");
	AssetDatabase.ForceReserializeAssets(array);
}
private static string[] GetAssets(string filter)
{
	string[] allPrefabsGUID = AssetDatabase.FindAssets($"{filter}");

	List<string> allPrefabs = new List<string>();
	foreach (string guid in allPrefabsGUID)
	{
		allPrefabs.Add(AssetDatabase.GUIDToAssetPath(guid));
	}
	return allPrefabs.ToArray();
}
```

gilbdev:
reserializing assets also helps pointing out 'missing scripts' like #4738 
(when running my reserialize script, i got an error :
`You are trying to replace or create a Prefab from the instance 'ToggleFeaturesPanel' that references a missing script. This is not allowed.` )

gilbdev:
fixed in v2.0.0

PR #5352, opened 2019-07-17
Reserialize metas

keveleigh:
## Overview
This change reserializes our metas via 2018.4. Updating our serialization will help reduce errant, seemingly unrelated changes in both PRs into this project and other projects that consume the MRTK via source / unitypackage.

For the most part, these changes are just the removal of stale `timeCreated` and `licenseType` info, which are no longer included in meta files.

image

## Changes
- Part of #4914

PR #5429, opened 2019-07-25
Add ReserializeAssetsUtility

keveleigh:
## Overview
This PR adds a version of the tool provided in https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4914 to the repo, under our new Tools directory.

## Changes
- Related to https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4914

### DIFF
--- a/Assets/MixedRealityToolkit.Tools/ReserializeAssetsUtility/ReserializeAssetsUtility.cs
+++ b/Assets/MixedRealityToolkit.Tools/ReserializeAssetsUtility/ReserializeAssetsUtility.cs
@@ -3,10 +3,16 @@
 
 using System.Collections.Generic;
 using UnityEditor;
-using UnityEngine;
 
 namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
 {
+    /// <summary>
+    /// Adds menu items to automate reserializing specific files in Unity.
+    /// </summary>
+    /// <remarks>
+    /// Reserialization can be needed between Unity versions or when the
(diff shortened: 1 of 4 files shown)


keveleigh:
I also just included a single reserialized meta which was added to the repo after #5352

Assets/MixedRealityToolkit.Tools/ReserializeAssetsUtility/ReserializeAssetsUtility.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
+{
+    public class ReserializeUtility
```
wiwei: Can you also add a doccomment about what this class is for (i.e. what it does, benefits, etc).

Usually we'd like to also have some docs on it as well, but in this case because it's pretty small maybe just having the doccomments would be sufficient.

keveleigh: Oh whoops haha definitely

keveleigh: Added

PR #5430, opened 2019-07-25
Reserialize profiles and assets

keveleigh:
## Overview
This PR reserializes our profiles and various asset files. Mostly, these changes add / remove serialized fields that have changed in the underlying definitions.

## Changes
- Related to https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4914


wiwei:
By the way, do you have a recommendation on when we should be running this? Every release? At the beginning? At the end?

keveleigh:
@wiwei To be completely on top of it, any time a script's serialization gets something added or removed. Really though, probably right before we do a release.

Issue #5046, opened 2019-06-25
Add a default MRTK profile for HoloLens 1 Interactions


DrDecipher:
The original question / bug highlighted a gap in usability/configurability of MRTK for HoloLens 1. By default interactions are set up for HoloLens 2 and there's no easy way to get to a HoloLens 1 friendly MRTK profile. This item tracks adding a config profile that's hololens 1 friendly (for HoloLens 1 interactions). 

This profile should use Gestures input simulation mode instead of articulated hands to simulate how HoloLens 1 works.

# Original Question
Gaze cursor is no longer re-enabled after using the 'Hand' in the latest release and can not be altered manually via the inspector.

1. Press Play( Gaze Cursor will be visible/active highlighting correctly )
2. Hold Down the 'Space Bar' to engage 'Hand'.
3. Release 'Space Bar'
4. Gaze Cursor remains inactive and disabled.
5. Can not be activated in the inspector as something is turning it off per frame.

## Expected behavior
After using either 'Hand', the Gaze cursor should reappear and be active after releasing the space bar.

## Your Setup (please complete the following information)

- Unity Version 
Unity : 2019.1.6f1
- MRTK : MRTK RC2.1

Platforms
- HoloLens
- HoloLens 2

wiwei:
This is by design AFAIK to be consistent with how the shell works (i.e. gaze cursor visible, then hand shows up, you have to say select order to get the cursor to show up again).

@julenka FYI, also to correct me if I'm wrong here.

Is there a specific scenario/use case that you're getting blocked on because of this?

DrDecipher:
We are developing for both the HL1 and the HL2.  When testing we are often checking controls with an air-tap and a hand.  This type of behavior should not be hidden. As great as the MRTK2 is becoming there seem to be little consideration for the fact that many of us will be stuck in 'HL1 land' for the foreseeable future. _How can I reenable gaze it after the hand is used in the editor?_ Any assistance would be appreciated :)

wiwei:
@julenka, Q for you, maybe a question for folks who don't want the default shell behavior that the MRTK is mimicing

DrDecipher:
[...]

PR #5431, opened 2019-07-25
Config Profile for HoloLens 1

julenka:
## Overview
By default the MRTK config profile assumes articulated hand input using HoloLens 2 style hand rays + direct interaction, however many people using MRTK are still deploying to HoloLens 1. People often ask how they can [simulate HoloLens 1 style interactions in the Unity Editor](https://stackoverflow.com/questions/56821449/hl1-input-for-mrtk-v2-1).

People building and deploying HoloLens 1 apps can use this configuration in their MRTK.

This change adds an MRTK profile that people can use to simulate HoloLens 1 style GGV interactions.

Fixes: #5046 

## Changes
* New profile to simulate HoloLens 1 interations.
* Profile uses "gesture hand" for input simulation, not the articulated hand
* Set Camera Clipping plane for HoloLens 1 to 0.3 per comfort guidelines at https://docs.microsoft.com/en-us/windows/mixed-reality/comfort

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Profiles/HoloLens1.meta
+++ b/Assets/MixedRealityToolkit.SDK/Profiles/HoloLens1.meta
@@ -1,0 +1,8 @@
+fileFormatVersion: 2
+guid: 2385ba3fd0af66146ba74ee7f05d17c7
+folderAsset: yes
+DefaultImporter:
+  externalObjects: {}
+  userData: 
+  assetBundleName: 
+  assetBundleVariant: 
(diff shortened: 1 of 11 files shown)


wiwei:
Profile lgtm.

Consider adding notes to this doc as well to explain the differences to consumers:

https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Profiles/Profiles.html#hololens-2-profile

i.e. add a "hololens 1 profile" section

I don't believe that people will all find it, but when people ask, it is important for us to be able to point to an answer and help build up a culture of doing a search on the docs.

wiwei:
So, I have no idea why this is failing the build on mrtk_pr, and why it seems to only be this particular change. 

I tried pulling this locally as well and running this in batch mode and it works just fine - I'm trying to retarget to the stabillization branch and trying again.

The worst part is that the tests named "sanity tests" are failing

Issue #5219, opened 2019-07-08
ManipulationHandler Far interaction "Rotate using wrist" no longer works


julenka:
## Overview
If ManipulationHandler one hand rotation mode is configured as "rotate using wrist", the object does not rotate with the wrist when using far interaction

## Steps to reproduce
1. Open HandInteractionExamples scene, and notice that the earth is configured with "One Hand Rotation Mode Far" as "Rotate around Grab Point"
![image](https://user-images.githubusercontent.com/168492/60786968-50d65900-a193-11e9-9192-ed7d7a4ae5be.png)
2. Grab globe with ray and rotate hand

Expected: globe rotates

Actual: nothing rotates

thalbern:
I've spend a lot of time looking into this and make this work, but with how our input system is set up and the values we're getting from the platform API this is just not working. 

Our pointer far interaction is currently querying the Pointer Pose from the Unity XR.WSA Input API. That pose doesn't include any roll values for wrist rotations. 
There's a second pose that could be queried that is the Grip Pose from the same API. However this is meant for near interaction and our input system doesn't provide a way right now to keep two different data sets around. It's either the one or the other. Also whenever querying one of these I'm only operating on one data set provided by the OS, so even if I make some custom logic for this I don't even have access to the two raw data sets in my pose update functions. 

Imho the roll values for making a far interaction wrist rotation work should come from the platform API and not be stitched together on our end. 
If we want to make this work with the data we're receiving right now we would have to make considerable changes in our input system which I won't recommend for this release.

thalbern:
I hacked some code together that combines the values for those two interaction mappings now and stomps over the previously generated pose (sends the input event again as well). 
It's not nice code and I don't think we should put that in, but put it into a draft now in case someone wants to have a look.

thalbern:
this was fixed in the hl2 OS now - won't be adding any fixes on MRTK side - waiting for access to a device to verify before closing this ticket

thalbern:
verified fix - closing ticket

PR #5440, opened 2019-07-26
added fix for roll on far manipulation on hololens 2

thalbern:
current code will just make sure that the object won't drift off in a strange way by zeroing out the roll value we're getting from the api.
the commented code should be the combination of the actual wrist roll from the grip pose with the previously queried pointer pose. 
I couldn't figure out how the hand is oriented , so with the code the manipulated objects will jump at the start.
we can't fix this in manipulation handler. all other platforms are working fine because they give us reasonable rotation values. unfortunately pointer pose on hl2 doesn't :-/

## Changes
- Fixes: #5219


thalbern:
closing this as the issue was properly fixed in the HL2 OS

Issue #5434, opened 2019-07-26
Fails to build when running via Build Window / Appx Build Options when no VS solution already created


Naphier:
## Describe the bug
Trying to use these new build tools and figure out how to get this rolling on CI/CD. 
In Build Window
- If we run Unity Build Options to generate the VS project then run Appx Build Options everything works as expected.
- If we run from Appx build options without a generated VS project the build fails. I believe this is the failure reason:
```
C:\VRR_Repos\WARC-LSOD\Builds\WSAPlayer\WARC-LSOD.sln.metaproj : error MSB4126: The specified solution configuration "master|" is invalid. Please specify a valid solution configuration using the Configuration and Platform properties (e.g. MSBuild.exe Solution.sln /p:Configuration=Debug /p:Platform="Any CPU") or leave those properties blank to use the default solution configuration. [C:\VRR_Repos\WARC-LSOD\Builds\WSAPlayer\WARC-LSOD.sln]
UnityEngine.Debug:Log(Object)
VRR.MixedRealityToolkit.<>c__DisplayClass1_0:<StartProcessAsync>g__OnOutputDataReceived|2(Object, DataReceivedEventArgs) (at Assets/Editor/From MRTK/Extensions/ProcessExtensions.cs:106)
System.Threading._ThreadPoolWaitCallback:PerformWaitCallback()
```
- This same message occurs whether the _Appx Build Options_ are set to Release or Master. Currently set to IL2CPP, Release, x86

## To reproduce
See description. 

## Expected behavior
That the VS project would build and then the appxbundle would build without errors.

## Your Setup (please complete the following information)
- Unity Editor 2018.3.1f1
- MRTK SDK RC 2.0.0
- Windows SDK 10.0.18922.0

## Target Platform (please complete the following information)
- HoloLens

## Additional info
We also get a few compiler errors, but these happen no matter if we're just building the VS project or doing both VS proj and appx bundle. They do not seem to affect the build. Also they don't appear when building via Unity's Build Settings window. Thought they'd go away with Windows SDK 10.0.18922.0 but nope, probably a Unity issue.
```
[...]

PR #5442, opened 2019-07-26
Prevents build failure

Naphier:
## Overview
Prevents build failure when buildInfo.Configuration or buildInfo.BuildPlatform are empty or whitespace. Addresses #5434
Simply adds a couple of null/whitespace checks on the above variables and sets them to safe defaults so that the build can proceed. Warning is displayed in console so dev can debug if necessary.

## Changes
- Fixes: #5434

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
@@ -98,6 +98,18 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
                 IsBuilding = false;
                 return false;
             }
+            
+            if (string.IsNullOrWhiteSpace(buildInfo.Configuration))
+            {
+                Debug.LogWarning("!!! buildInfo.Configuration is null or whitespace. Using 'master'");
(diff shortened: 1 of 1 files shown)


wiwei:
Thanks for sending this PR!

@Naphier, can you take a look at my comments in the issue before this goes in:

https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5434

I think that these changes seem fine, I just want to make sure they're still needed after the couple of bug fixes that have gone into mrtk_development

Naphier:
I'm not completely certain they're needed. Weird thing is that somewhere, while testing, EditorBuildPrefs starting having the right platform. So it's a bit hard to diagnose / reproduce. Hence just a check with warnings. 
Maybe it's better if I just fail the build if either of those are null/whitespace?
Main problem is that MSBuild outputs something like `invalid build option 'master|'` which is far from helpful in figuring out what went wrong.
What do you think?

wiwei:
Ah, it looks like we're chatting in both the issue and the PR - let's keep the discussion in one place (in the issue, because that's where the most information is) - we can resolve stuff there and then circle back here after!

wiwei:
Closing this down while discussion is happening here - I wanted to reiterate that the change itself looks fine, I just want to make sure that it's still needed with the latest in mrtk_development

PR #5444, opened 2019-07-26
Updated the reference to HL2 tutorials

rwinj:
## Changes
- Fixes: # .

### DIFF
--- a/Documentation/GettingStartedWithTheMRTK.md
+++ b/Documentation/GettingStartedWithTheMRTK.md
@@ -20,7 +20,7 @@ To get started with the Mixed Reality Toolkit you will need:
   when building for OpenVR.
 
 
-# HoloLens 2 tutorials
+# Getting started tutorials
 If you are new to MRTK, or MR development, we recommend you check out the [Getting started tutorials](https://docs.microsoft.com/en-us/windows/mixed-reality/mrlearning-base) which uses MRTK v2.
 
 
--- a/Documentation/GettingStartedWithTheMRTK.md
+++ b/Documentation/GettingStartedWithTheMRTK.md
@@ -20,8 +20,8 @@ To get started with the Mixed Reality Toolkit you will need:
   when building for OpenVR.
 
 
-# MR Learning Base Module Tutortial 
-If you are new to MRTK, or MR development, we recommend you check out the [MR Learning Base Module](https://docs.microsoft.com/en-us/windows/mixed-reality/mrlearning-base) which uses MRTK v2.
+# HoloLens 2 tutorials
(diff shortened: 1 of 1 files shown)


Documentation/GettingStartedWithTheMRTK.md
```diff
@@ -20,8 +20,8 @@ To get started with the Mixed Reality Toolkit you will need:
   when building for OpenVR.
 
 
-# MR Learning Base Module Tutortial 
-If you are new to MRTK, or MR development, we recommend you check out the [MR Learning Base Module](https://docs.microsoft.com/en-us/windows/mixed-reality/mrlearning-base) which uses MRTK v2.
+# HoloLens 2 tutorials
+If you are new to MRTK, or MR development, we recommend you check out the [Getting started tutorials](https://docs.microsoft.com/en-us/windows/mixed-reality/mrlearning-base) which uses MRTK v2.
```
wiwei: This change is factually accurate (because this link has since been repurposed for HL2), but I feel like suggests that we're having focus only on HL2 - there are a number of other tutorials on there like:

https://docs.microsoft.com/en-us/windows/mixed-reality/tutorials

We might want to call out that general part first, and then say hey, check out this one which uses the MRTK that covers HL2.

@keveleigh, thoughts?

rwinj: Yes, if you want to be more general about all the available tutorials then you can point to the tutorials overview page you referenced (https://docs.microsoft.com/en-us/windows/mixed-reality/tutorials). I had just removed the "learning module" text from that page past week.

Documentation/GettingStartedWithTheMRTK.md
```diff
@@ -20,8 +20,8 @@ To get started with the Mixed Reality Toolkit you will need:
   when building for OpenVR.
 
 
-# MR Learning Base Module Tutortial 
-If you are new to MRTK, or MR development, we recommend you check out the [MR Learning Base Module](https://docs.microsoft.com/en-us/windows/mixed-reality/mrlearning-base) which uses MRTK v2.
+# HoloLens 2 tutorials
```
wiwei:
```suggestion
# Getting started tutorials
```

Documentation/GettingStartedWithTheMRTK.md
wiwei: Let's go with the suggestion posted above (i.e. getting started tutorials). I think we can expand it to link to the other ones (i.e. hololens 1 tutorial as well). Since this is one of the starting off points, I don't really want to have people think "oh, I don't have a hololens 2, what tutorial should I follow?"

Really these tutorials are generally applicable to other devices as well (i.e. bounding box, manipulation handler)

PR #5452, opened 2019-07-29
Resolve the ArgumentException error in MRTK+Unity startup

wiwei:
When ingesting MRTK for the first time (especially via git) you'll see an ArgumentException warning, which happens as MRTK is figuring out where its files are being placed. This shows as an error but is actualy benign (i.e. it happens in some processing of UPM packages that have no effect on the MRTK).

The cause here was an overly aggressive .Replace call - it replaced ALL instances of "Asset" with the application asset path (even when Asset was showing up much later in the path). It was really only intending to replace the prefix "Asset" string, so this fixes it to only do that.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
@@ -43,13 +43,13 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
 
                 foreach (string asset in importedAssets.Concat(movedAssets))
                 {
-                    string folder = asset.Replace("Assets", Application.dataPath);
+                    string folder = ResolveFullAssetsPath(asset);
                     TryRegisterModuleFolder(folder);
                 }
 
                 foreach (string asset in deletedAssets.Concat(movedFromAssetPaths))
                 {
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -59,6 +59,23 @@ public static void OnPostprocessAllAssets(string[] importedAssets, string[] dele
...
+        /// <summary>
+        /// Resolves the given asset to its full path if and only if the asset belongs to the
+        /// Assets folder (i.e. it is prefixed with "Assets/..."
+        /// </summary>
+        /// <remarks>
+        /// If not associated with the Assets folder, will return the path unchanged.
+        /// </remarks>
+        private static string ResolveFullAssetsPath(string path)
+        {
+            if (path.StartsWith("Assets"))
```
Troy-Ferrell: Add null check for safe measure?

wiwei: I think I'll keep this one as-is - if we did do some null checking, I wouldn't want to proliferate this into all of the private static helpers, and would rather do the checking at the various data junction points (i.e. the actual Unity API). Even in this case, however, I assume that we should be able to depend on strings from Unity APIs being not null, in which case we would only null check if there were a Unity bug)

PR #5456, opened 2019-07-29
Add planning process document

davidkline-ms:
This change adds a document describing the MRTK planning process.

### DIFF
--- a/Documentation/PlanningProcess.md
+++ b/Documentation/PlanningProcess.md
@@ -1,0 +1,78 @@
+# Mixed Reality Toolkit Planning process
+
+Before each milestone, the MRTK team will prioritize features to implement and bugs to fix. A
+milestone will be comprised of one or more iterations, resulting in an official MRTK release.
+
+Items will be distributed between iterations based on resource availablility and customer impact.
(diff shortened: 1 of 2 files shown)


wiwei:
I was thinking about where we might put this, and the more I think about it, I'm curious if having planning process be part of the docs is kinda odd - it kinda feels like the docs should reflect how as a consumer folks should be consuming things, whereas maybe our wiki here is what should be driving contributors.

Otherwise, I'm kinda curious - would you imagine we'd also put iteration plans on the same general location on the docs? Would it make sense to keep the iteration plans and generally planning/how we run the project somewhat close together (so that you can go from overall meta plan to specific plans).

I think that it might be good to get some alignment on this, as it will affect where we choose to put other artifacts in the future.

davidkline-ms:
> I was thinking about where we might put this, and the more I think about it, I'm curious if having planning process be part of the docs is kinda odd 

Good point. I am happy to post this in the repository's wiki. Sound reasonable?

wiwei:
Cool, closing this one out in favor of the wiki

Issue #4340, opened 2019-05-14
Building .NET Master (.NET Native) flavor to HL2 Breaks Hands


andreiborodin:
## Overview
When you build Release version of .NET, everything works on latest MRTK as of today. However, the Master version (.NET Native) compiler breaks the hands.

## Expected Behavior
Hands to work.

## Actual Behavior
Hands don't work.

## Steps to reproduce
1. Take latest MRTK
2. Switch to .NET Scripting Runtime
3. Build UWP
4. Build and deploy to device the Master flavor (.NET Native enabled)

## Unity Editor Version
2018.3.11f1

## Mixed Reality Toolkit Release Version
2.0.0-RC1Refresh

Sample of the log:
ArgumentException: Arg_ArgumentException
   at SharedLibrary!<BaseAddress>+0x3f5e3b​
   at HelloAR U3D!<BaseAddress>+0x141fdbb​
   at HelloAR U3D!<BaseAddress>+0x141fc3b​
   at Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input.WindowsMixedRealityArticulatedHand.UpdateControllerData(XR.WSA.Input.InteractionSourceState interactionSourceState)​
   at Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input.WindowsMixedRealityArticulatedHand.UpdateController(XR.WSA.Input.InteractionSourceState interactionSourceState)​
   at Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input.WindowsMixedRealityDeviceManager.InteractionManager_InteractionSourceDetected(XR.WSA.Input.InteractionSourceDetectedEventArgs args)​
   at HelloAR U3D!<BaseAddress>+0x2154629​
   at UnityEngine.XR.WSA.Input.InteractionManager.SourceEventHandler.Invoke(XR.WSA.Input.InteractionSourceState state)​
   at UnityEngine.XR.WSA.Input.InteractionManager.OnSourceEvent(XR.WSA.Input.InteractionManager.EventType eventType, IntPtr statePtr, XR.WSA.Input.InteractionSourcePressType pressType)​
   at UnityEngine.XR.WSA.Input.InteractionManager.$Invoke19(Int64 instance, Int64* args)​
   at UnityEngine.Internal.$MethodUtility.InvokeMethod(Int64 instance, Int64* args, IntPtr method) ​
(Filename: <Unknown> Line: 0)
[...]

PR #5458, opened 2019-07-29
Add workaround for hands/eyes on .NET Native

keveleigh:
## Overview
There is currently a known issue with Master builds using the .NET backend. In .NET Native, `IInspectable` pointers cannot be marshaled from native to managed code using `Marshal.GetObjectForIUnknown`. The MRTK uses this to obtain the `SpatialCoordinateSystem` in order to receive hand and eye data from the platform.

Native source PR: https://github.com/microsoft/MixedRealityToolkit/pull/228

We are not shipping the binaries precompiled in this case, but users who need them can compile them themselves from the native source repo.

## Changes
- Fixes: #4340 

## Verification
1. Build the binary from the native repo
2. Include it in your plugins folder
3. Build the .NET backend as Master and deploy to your HL2.
4. Verify that hand joints properly render (the HandInteractions example scene is good for this).

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityUtilities.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityUtilities.cs
@@ -23,10 +23,13 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input
 
         /// <summary>
         /// Helps marshal WinRT IInspectable objects that have been passed to managed code as an IntPtr.
+        /// </summary>
+        /// <remarks>
         /// On .NET Native, IInspectable pointers cannot be marshaled from native to managed code using Marshal.GetObjectForIUnknown.
         /// This class calls into a native method that specifically marshals the type as a specific WinRT interface, which
         /// is supported by the marshaller on both .NET Core and .NET Native.
-        /// </summary>    
+        /// Please see https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/InputSystem/HandTracking.html#net-native for more info.
+        /// </remarks>
         private static SpatialCoordinateSystem GetSpatialCoordinateSystem(IntPtr nativePtr)
         {
             try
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityUtilities.cs
```diff
@@ -12,7 +15,34 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input
...
+        private static extern void GetSpatialCoordinateSystem(IntPtr nativePtr, out SpatialCoordinateSystem coordinateSystem);
+
+        public static SpatialCoordinateSystem SpatialCoordinateSystem => spatialCoordinateSystem ?? (spatialCoordinateSystem = GetSpatialCoordinateSystem(WorldManager.GetNativeISpatialCoordinateSystemPtr()));
+
+        /// <summary>
+        /// Helps marshal WinRT IInspectable objects that have been passed to managed code as an IntPtr.
+        /// On .NET Native, IInspectable pointers cannot be marshaled from native to managed code using Marshal.GetObjectForIUnknown.
+        /// This class calls into a native method that specifically marshals the type as a specific WinRT interface, which
+        /// is supported by the marshaller on both .NET Core and .NET Native.
+        /// </summary>
```
wiwei: Nit: In the remarks, it might be worth adding a link to the docs that you're adding as well. I know that this might be a circular reference, but it could help people understand "hey, I just hit this LogError, what do I do?"

keveleigh: Oh, since this is a private function, the summary / remarks don't actually get published anywhere :/

wiwei: That's okay - my comments were more about fulfilling the needs of someone who is reading this code directly (i.e. they ask, why don't hands work?) or they see the log message - they might then come directly to this code, so having the link here would reduce the friction for figuring out how this part works.

keveleigh: Cool yeah, I can put a more full link in. I was thinking more getting an xref working, which isn't super translatable while reading to figure out exactly where you need to go.

keveleigh: Added

PR #5473, opened 2019-07-30
Move extension templates to tool folder

davidkline-ms:
## Overview
The template files used by the ExtensionServiceCreator tool were previously located in the root of the project (peer to the Assets folder). This location is not valid as they cannot be navigated to by the user (from within Unity) and prevents them from being packaged for distribution.

This change moves them into a Templates folder beneath MixedRealityToolkit.Tools/ExtensionServiceCreator and updates the paths within the tool.

## Validation
Confirmed that the tool successfully finds and loads the tempates.


Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/ExtensionServiceCreator.cs
```diff
@@ -68,10 +68,10 @@ private struct PersistentState
         private static readonly string PersistentStateKey = "MRTK_ExtensionServiceWizard_State_Before_Recompilation";
         private static readonly string DefaultExtensionsFolder = "Assets/MixedRealityToolkit.Extensions";
         private static readonly string DefaultExtensionsFolderName = "MixedRealityToolkit.Extensions";
-        private static readonly string ServiceTemplatePath = "ExtensionTemplates/ExtensionScriptTemplate.txt";
-        private static readonly string InspectorTemplatePath = "ExtensionTemplates/ExtensionInspectorTemplate.txt";
-        private static readonly string InterfaceTemplatePath = "ExtensionTemplates/ExtensionInterfaceTemplate.txt";
-        private static readonly string ProfileTemplatePath = "ExtensionTemplates/ExtensionProfileTemplate.txt";
+        private static readonly string ServiceTemplatePath = "Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/Templates/ExtensionScriptTemplate.txt";
```
keveleigh: Does this path resolve properly even when consumed via Nuget?

Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/ExtensionServiceCreator.cs
```diff
@@ -68,10 +68,10 @@ private struct PersistentState
...
+        private static readonly string InspectorTemplatePath = "Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/Templates/ExtensionInspectorTemplate.txt";
```
keveleigh: Also, I think there was some work done previously to support the MRTK in different locations / folder names than the default

keveleigh: This file I think https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs

wiwei: Yep - look up MixedRealityToolkitFiles.cs - has some stuff around identifying where things are located at

Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/ExtensionServiceCreator.cs
davidkline-ms: I will need to work with @andreiborodin on this part. I do not believe they are consumable via NuGet now.

Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/ExtensionServiceCreator.cs
davidkline-ms: Looks like we need to add Extensions and Tools to that list.

As a note, this was a simple string replace. None of this class leverages the toolkit files class.

davidkline-ms: updated the tool source. thanks!

Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/ExtensionServiceCreator.cs
davidkline-ms: talked with @andreiborodin and it looks like these files will be packaged correctly.

Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/ExtensionServiceCreator.cs
```diff
@@ -85,6 +81,38 @@ private struct PersistentState
 
         #endregion
 
+        #region paths
+
+        private string ExtensionsFolder
+        {
+            get
```
keveleigh: nit: inconsistent syntax between this (`get { }`) and the ones following (`get =>`).

Could we turn them all into `private string Foo => MixedRealityToolkitFiles.Bar()`? :)

davidkline-ms: i'll update the syntax :)

davidkline-ms: fixed

PR #5475, opened 2019-07-31
Fixes two omissions in the ExtensionServiceCreator

LocalJoost:
## Overview
1. ExtensionServiceCreator ignores namespace supplied by user, always takes default instead
2. ExtensionServiceCreator creates inspector even when "Generate Inspector" is unselected

## Changes
1. ExtensionServiceCreator now uses NameSpace in stead of DefaultExtensionNamespace while creating the files. This also required adaption of the template files as those now have a missing "using" on top now they are no long in Microsoft.MixedReality.Toolkit.Extensions;
2. Added simple "if(UsesInspector)" around creation of inspector. This was read from the UI and clearly intended to use for this, but was not checked

### DIFF
--- a/Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/ExtensionServiceCreator.cs
+++ b/Assets/MixedRealityToolkit.Tools/ExtensionServiceCreator/ExtensionServiceCreator.cs
@@ -517,10 +517,10 @@ namespace Microsoft.MixedReality.Toolkit.Editor
 
             try
             {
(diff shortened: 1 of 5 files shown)


wiwei:
@Railboy, can you take a look as well?

wiwei:
Thanks for fixing this! Change lgtm, would like to get @Railboy's eyes because he wrote this.

davidkline-ms:
Changed base to stabilization branch so that this makes it into the next release.

LocalJoost:
Glad to be of service. I had run into this limitation for quite some number of times, and I am glad I could lend a helping hand

PR #5480, opened 2019-07-31
Create ci pipeline that packages but does not publish them

davidkline-ms:
This change adds yml files to create a CI pipeline that creates but does not publish packages.

This will enable creation of package artifacts without creating noise for partner teams.


pipelines/ci-packaging-dontpublish.yml
```diff
@@ -0,0 +1,19 @@
...
+  pool:
+    name: On-Prem Unity
+    demands:
+    - Unity2018.3.7f1
+    - COG-UnityCache-WUS2-01
+    - SDK_18362 -equals TRUE
+  steps:
+  - template: templates/ci-dontpublish.yml
+    parameters:
+      packagingEnabled: true
```
keveleigh: Could we add an additional parameter to the existing yml called `publishingEnabled`?

keveleigh: I guess we'd have to create a new pipeline either way (I thought of this first as a way around that), but at least removes some duplicate code? (mostly the other file)

wiwei: I think it would be good to avoid that extra duplication as well

MenelvagorMilsom: Agreed, I think an additional parameter is preferable if possible.

davidkline-ms: i can give that a try :)

Issue #4864, opened 2019-06-13
Consider changing the deployed MRTK directory structure


keith-phillips:
Currently when you deploy the MRTK, it creates a number of top-level directories:
- Assets/MixedRealityToolkit
- Assets/MixedRealityToolkit.Providers
- Assets/MixedRealityToolkit.SDK
- Assets/MixedRealityToolkit.Services

Consider changing the deployed structure of the MRTK to move everything under a single top-level directory, e.g.
- Assets/MRTK/MixedRealityToolkit
- Assets/MRTK/MixedRealityToolkit.Providers
- Assets/MRTK/MixedRealityToolkit.SDK
- Assets/MRTK/MixedRealityToolkit.Services

This would allow us to control the MRTK as a separate git submodule, which provides some nice benefits:
- It's more obvious if we ever make changes to the MRTK code, as opposed to our own scripts
- If we do ever make any changes, we can keep our own branch of the submodule and easily merge with new versions
- It's simpler to extract any patches we make to submit a PR upstream

Currently any user-derived profiles, etc that are created go under a different top-level directory:
- Assets/MixedRealityToolkit.Generated

It would probably make most sense for these to remain at this level, so that user-generated files don't clutter up the directory with the original MRTK code in it.

wiwei:
We have some work which would make our package UPM compliant, which I believe ends up getting placed in the Packages folder instead of the root assets - this might also help with the issue that we're hitting here (it won't clutter up your project folder that way)

gilbdev:
my current working solution : 
- have the whole mrtk repo as a submodule
`git submodule add -b mrtk_release --force your/git/repo/fork/Microsoft_MixedRealityToolkit-Unity Subrepos/MixedRealityToolkit-Unity`
 
- add some symlinks for each fo the 4 subfolders in your Unity project (named Unity folder here)

`mklink /D Unity\Assets\MixedRealityToolkit ..\..\Subrepos\MixedRealityToolkit-Unity\Assets\MixedRealityToolkit`

wiwei:
Okay, finally got around to this one - so I think for our unity package, we'll probably keep the current structure around where it is put initially (for now), but now with https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5483 it's possible to put MRTK in a non-root folder (there were a few issues where some specific things weren't working outside of the non-root folder).

@davidkline-ms @keveleigh FYI.

Issue #4067, opened 2019-04-24
Changing MRTK folder paths results in problems with Controller Mapping


Avalin:
## Overview
Moving the MRTK folders to another folder than Assets (My path was in this case Assets/Libs/MRTK/MixedRealityToolkit... etc) results in issues with Controller Mapping, as the EditorWindow (ControllerWindowPopup.cs) is hardcoded to open up directly from Assets folder. 

## Expected Behavior
My idea would be one or more of potential scenarios:
- Controller Mapping working no matter where MRTK folders are placed
- Easily readable LogErrors when attempting to open the ControllerMapping from wrong path
- Highlighted in documentation that moving the project to a different folder creates issues (Like it is highlighted with changing the camera position) 

## Actual Behavior
NullPointerReference saying it couldn't close the window because it is null from the ControllerWindowPopup.cs

## Steps to reproduce
-Download MRTK RC1
-Generate Custom Profiles (Particularly ControllerMapping profile) 
-Move all folders into a new folder called MRTK 
-Open any of the EditorWindows from the ControllerMapping profile

## Unity Editor Version
2018.3.13f

## MRTK Version
MRTK RC1

wiwei:
I thought that this might have been fixed recently because we did some work to make relative paths work (i.e. to support the MRTK being put in non-root-asset folder locations), but it looks like it didn't quite cover this case. We'll take a look here

wiwei:
A long time later, this got fixed.

PR #5483, opened 2019-07-31
Update the MRTK to support being placed in the non-root Asset folder

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4864
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4067

Up until now, the MRTK has only worked if you had its code in the root Assets folder - there were specific parts of the code that would use Application.dataPath and do things relative to that, instead of using the MixedRealityToolkitFiles helper class to do correct folder resolution.

This fixes it so that the MRTK now SHOULD support other paths (i.e. if consumers want to place things in Assets/lib/mrtk/*), by updating the parts of the MRTK that currently use Asset absolute paths to instead use the MixedRealityToolkitFiles.


Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -33,6 +37,18 @@ public enum MixedRealityToolkitModuleType
     [InitializeOnLoad]
     public static class MixedRealityToolkitFiles
     {
+        /// <summary>
+        /// This controls the behavior of MapRelativePathToAbsolutePath.
+        /// </summary>
+        private enum SearchType
+        {
+            /// <summary>
+            /// This indicates
```
keveleigh: Incomplete?

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/MixedRealityToolkitFilesTests.cs
```diff
@@ -123,6 +154,54 @@ public void TestValidPath(string path, MixedRealityToolkitModuleType expectedMod
...
+            adhocTestDirectoryCreated = true;
+            Directory.CreateDirectory(Path.Combine(Application.dataPath, "lib"));
+            Directory.CreateDirectory(Path.Combine(Application.dataPath, "lib/mrtk"));
+            Directory.CreateDirectory(Path.Combine(Application.dataPath, "lib/mrtk/MixedRealityToolkit.AdhocTesting"));
+            Directory.CreateDirectory(Path.Combine(Application.dataPath, "lib/mrtk/MixedRealityToolkit.AdhocTesting/empty"));
+        }
+
+        private static void RecursiveFolderCleanup(DirectoryInfo folder)
+        {
+            foreach (DirectoryInfo subFolder in folder.GetDirectories())
```
Troy-Ferrell: Why not just call folder.Delete(true) and then you just need to delete the associated meta file?

If you are worried about re-use of folder by other components, add a random guid?

Troy-Ferrell: Actually does the current code delete lib.meta?

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/MixedRealityToolkitFilesTests.cs
```diff
@@ -9,16 +9,18 @@
 using System;
 using System.IO;
 using System.Threading.Tasks;
+using System.Linq;
 
 namespace Microsoft.MixedReality.Toolkit.Tests.Core
 {
     // Tests for the MixedRealityToolkitFiles utility class
-    public class TestFixture_02_MixedRealityToolkitFilesTests
+    public class MixedRealityToolkitFilesTests
```
Troy-Ferrell: Can you also rename the TestFixture_01_...?

SimonDarksideJ: I don't understand why you would want to rename the test functions?  Keep to the standards used for all the other tests and revert these changes.  Only helper methods that are not tests should have such names.

Assets/MixedRealityToolkit/Inspectors/ControllerPopupWindow.cs
```diff
@@ -218,12 +219,12 @@ public static void Show(MixedRealityControllerMapping controllerMapping, Seriali
                     }
                 };
 
-                File.WriteAllText($"{Application.dataPath}{EditorWindowOptionsPath}", JsonUtility.ToJson(empty));
```
SimonDarksideJ: Output functions **should not change**. keep the app data path as these are unity constants. especially as this pr is about fixing reading of files.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
SimonDarksideJ: agreed, either complete or remove trailing ,

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/MixedRealityToolkitFilesTests.cs
wiwei: I can do the other rename separately - the existing naming of TestFixture_#_ adds no value and only clutters things up.

If the intent is specific order, that's a really bad intent - you should NEVER depend on your tests working on a specific order.

Assets/MixedRealityToolkit/Inspectors/ControllerPopupWindow.cs
wiwei: I'm not sure that we're on the same page here about what this is doing.

Today, if you have MRTK in the root Assets folder, this thing works because Application.dataPath is just the assets folder. Note that, this is **editor specific code that runs in editor**

This obviously doesn't work if it's just in a subfolder of Assets. This is what this fixes. It happens to be in this case that the input == a json file about the controller window, and the output == a json file about the controller window, so they are literally the same thing.

wiwei: Another way of saying this is that, I think that the comment here being left is a generally valid statement but it doesn't apply here - if you have an alternative suggestion that leaves this the same (which would, FWIW, be broken), prefer leaving concrete suggestions instead of just generalisms.

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
wiwei: Yep, obviously a mistake, fixed!

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/MixedRealityToolkitFilesTests.cs
wiwei: This code itself doesn't, but the part in TestCleanup does - FWIW, I took this from the MSBuild code that was added here (I guess that @andreiborodin tried a few different things here and this is what ended up working)

I didn't choose to make it common because of the direction where we'll be moving the msbuild code to a different repo in the medium term, and didn't want to add more things to disentangle

Troy-Ferrell: I'm slightly confused. Test cleanup calls RecursiveFolderCleanup("...\Assets\lib") but I don't think this actually deletes lib.meta?

I'm not saying necessarily make RecursiveFolderCleanup() in a common place like Utils. I am asking why can't the function be simplified to like the below? Simpler to read & maintain

CleanupFolder(path)
{
// if dir exists path
var dir = new DirectoryInfo(path);
dir.Delete(true);

// if path + ".meta" file exists
var metaFile = new FileINfo(path);
file.delete();
}

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/MixedRealityToolkitFilesTests.cs
Troy-Ferrell: @SimonDarksideJ I don't see that naming pattern being the standard though. It's only used by a handful of tests. The vast majority are more straightforward. "TestFixture" on the front seems redundant with adding "Tests" to the end of the name and the number doesn't seem to have any purpose or value. 

I agree with @wiwei and say we make the names simple like the other tests so it is more readable for others

Issue #5485, opened 2019-08-01
In-Editor camera/hand simulation can get into a bad state when the mouse goes out of the Unity Editor window. 


hendryems:
## Describe the bug

One of our developers observed that the hand simulation can get into a bad state when the mouse goes out of the Unity Editor window. Specifically, when enabling mouse look (Mouse Right Click), Left Hand Manipulate (Left Shift), Pinch (Mouse Left Click) to drag an object in the HandInteractionExamples scene until the mouse is outside of the Unity Editor window. Releasing all would still result in mouse look state still being active.

## To reproduce

1. Run the HandInteractionExamples scene
2. Grab an object with Mouse Right Click, Left Shift + Mouse Lift Click
3. Start dragging the object until the mouse cursor leaves the Unity Editor
4. Release Mouse Right & Left Click & Left Shift

## Expected behavior

Mouse movement no longer triggers mouse look

## Actual behavior

Mouse movement continues to drive camera

## Your Setup (please complete the following information)

- Unity Version [2018.4.4f1]
- MRTK Version [Microsoft.MixedReality.Toolkit.Foundation 2.0.0-20190730.2]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2

lukastoenneMS:
So this is definitely a Unity bug. Unfortunately they won't be fixing this any time soon, so i'm going to close this issue as well.

[https://issuetracker.unity3d.com/issues/mouse-button-input-remains-active-after-clicking-going-off-screen-and-releasing-it](https://issuetracker.unity3d.com/issues/mouse-button-input-remains-active-after-clicking-going-off-screen-and-releasing-it)

hendryems:
@ericob , do you know if this is something we're tracking on the Unity side already?

lukastoenneMS:
@hendryems I have logged the bug with Unity. The reply was

> I've reached out to our developers about your issue and, unfortunately, it is highly unlikely that this issue would be fixed, given its low severity and the fact that we are hoping to move to the new Input System at some point in the near future.

PR #5487, opened 2019-08-01
Unity button updates don't work reliably outside the viewport.

lukastoenneMS:
## Overview

Looks like this is a Unity bug (or call it undocumented behavior if you like): The [GetMouseButton](https://docs.unity3d.com/ScriptReference/Input.GetMouseButton.html) function does not work reliably outside the viewport window when called in the [Update](https://docs.unity3d.com/ScriptReference/MonoBehaviour.Update.html) function. Doing this in the [FixedUpdate](https://docs.unity3d.com/ScriptReference/MonoBehaviour.FixedUpdate.html) function instead seems to work, although i could only find vague hints in forum posts and all official documentation seems to indicate that Update() should be working.

## Changes
- Fixes: #5485 .

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
+++ b/Assets/MixedRealityToolkit.SDK/Experimental/Features/Support/Scripts/BaseServiceManager.cs
@@ -61,6 +61,25 @@ namespace Microsoft.MixedReality.Toolkit.Experimental
             }
         }
 
+        private void FixedUpdate()
+        {
+            if (Application.isPlaying)
+            {
+                if (Application.isPlaying)
+                {
+                    foreach (IMixedRealityService service in registeredServices.Values)
+                    {
+                        service.FixedUpdate();
+                    }
+
+                    for (int i = 0; i < dataProviders.Count; i++)
+                    {
+                        dataProviders[i].FixedUpdate();
+                    }
+                }
+            }
+        }
+
 
         protected virtual void OnEnable()
         {
--- a/Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
@@ -117,10 +117,6 @@ namespace Microsoft.MixedReality.Toolkit.Input
             if (profile.IsCameraControlEnabled)
             {
                 EnableCameraControl();
-                if (CameraCache.Main)
-                {
-                    cameraControl.UpdateTransform(CameraCache.Main.transform);
-                }
             }
             else
             {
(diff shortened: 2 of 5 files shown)


wiwei:
@keveleigh, @davidkline-ms

lukastoenneMS:
Change currently breaks tests because the `ManualCameraControl` does not always exist in FixedUpdate. Easy enough to fix, but i also think that this fix cannot be correct - all documentation and forum answers state that input should be handled in `Update` rather than `FixedUpdate`. I don't have any idea why that doesn't work here ...

lukastoenneMS:
It seems to be a Unity bug: If you press two buttons simultaneously (right + left or middle) and move the cursor outside of the viewport, only the 1st button released gets the ButtonUp state change signal. The second button release is never registered, you have to go back into the viewport and press that button again.

lukastoenneMS:
Same behavior in Unity 2018 and 2019.

wiwei:
Changing this back to target mrtk_development, as it looks like otherwise this ends up pulling in other changes.

keveleigh:
Could we just lock the mouse to the viewport instead?
https://docs.unity3d.com/ScriptReference/Cursor-lockState.html

That feels like a cleaner change than adding another method call to every service just to fix one, especially if we don't fully understand why it fixes it.

lukastoenneMS:
Yes, this PR should not be merged as-is (have set the DONOTMERGE label). Fixing the tests is simple, but i agree adding another update function for services is not great and shouldn't be necessary in the first place.

I hesitate to add cursor locking at this late stage before GA. It feels like we are inviting other bugs that way, because the state has to be reliably reset to avoid getting stuck in the viewport. Quoting the Unity docs:

> To provide a good user experience the recommended behavior is only to lock or confine the cursor as a result of user action, for example by pressing a button.

The problem with missing mouse button events also happens for other interactions, like grabbing objects or pushing buttons. There is little common ground to determine whether a mouse button push is intended to be used in conjunction with moving, where it could end up outside the viewport window. Locking the cursor on any kind of mouse button press could easily break other interactions, like UnityUI button pushes.

keveleigh:
>Locking the cursor on any kind of **mouse button press** could easily break other interactions, like UnityUI button pushes.

(emphasis mine) Ah right yeah, I said the wrong thing. I was more thinking any time space bar is pressed to bring a hand up, we'd lock the mouse for simulated hand control, then when the hand is removed / inactive, we'd relinquish control.

lukastoenneMS:
> 
> 
> > Locking the cursor on any kind of **_mouse button press_** could easily break other interactions, like UnityUI button pushes.
> 
> (emphasis mine) Ah right yeah, I said the wrong thing. I was more thinking any time space bar is pressed to bring a hand up, we'd lock the mouse for simulated hand control, then when the hand is removed / inactive, we'd relinquish control.

Thing is, this bug is not related to hand simulation at all. Try this:
1. Press & Hold RMB to start rotating camera.
2. Press & Hold LMB as well.
3. Move the mouse outside the viewport.
4. Release LMB first (!)
5. Release RMB second
Now when you move back into the viewport the camera will still be rotating. You have to click RMB a again to stop it. No hands involved.

lukastoenneMS:
The bug is definitely on the Unity side, but unfortunately won't be fixed. Closing this PR.

[https://issuetracker.unity3d.com/issues/mouse-button-input-remains-active-after-clicking-going-off-screen-and-releasing-it](https://issuetracker.unity3d.com/issues/mouse-button-input-remains-active-after-clicking-going-off-screen-and-releasing-it)

Issue #5478, opened 2019-07-31
InteractableHighlight is obsolete


Troy-Ferrell:
## Describe the bug

InteractableHighlight was marked obsolete since it was not very extensible and switched materials breaking any instancing or batching.

## Expected behavior

Create a new InteractableHighlight that provides the convenience of visual changes on focus but in a more extensible and performant way

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.11f1]
- MRTK Version [e.g. v2.0]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

Issue #5050, opened 2019-06-25
Interactable Custom Theme Assembly at Runtime


Railboy:
## Describe the problem

Interactable themes currently depend on editor-only inspector code to be configured correctly. They are also composed of many nested layers of classes. This makes it difficult to assemble and troubleshoot custom themes at runtime.

## Describe the solution you'd like

- A static utility class for assembling custom Interactable themes at runtime.
- Reduced dependence on inspectors for theme configuration.
- A revamped Interactable life cycle that allows for setting a theme after initialization.

## Describe alternatives you've considered

Sticking strictly to scriptable object references. This is not always feasible or desired.

PR #5493, opened 2019-08-02
Add on focus interactable feature and clean up inspectors

Troy-Ferrell:
## Overview
InteractableHighlight.cs was deprecated because it was limited in functionality and broke instancing/batching capabilities. This change adds InteractableOnFocus which uses focus change events to control interactable theme. 

In an effort to make Interactable Themes more extensible, there is also considerable work in the Theme & Interactable inspector code. 90% of the changes keep the same functionality but in a simpler and more componentized manner. There are some slight differences though.

- Fixed issue where remove button in ThemeInspector would remove the wrong theme property
- Populating shader options & names is a bit problematic especially since it is done at the inspector level but is dependent not at the Theme level but the Profile level. A theme asset can be assigned to multiple renderers which is a bit confusing. When creating a new theme asset, it does not auto-populate possibly shader properties. This has been fixed to use MRTK standard shader by default when empty
- Removed weird layout option filtering on ThemeInspector
- Moved Inspector classes to ....UI.Editor namespace
- Fixed InspectorUIUtility to not force callback functions which overcomplicated usage
- Added notes to breaking changes doc

## Changes
- Fixes: #5478 
- Related: #5050 

## Verification

>
>


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/InteractableOnFocus.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// Adds or removes materials to target renderer for highlighting Focused <see href="https://docs.unity3d.com/ScriptReference/GameObject.html">GameObject</see>s.
+    /// </summary>
+    /// <remarks>Useful with focusable <see href="https://docs.unity3d.com/ScriptReference/GameObject.html">GameObject</see>s</remarks>
+    public class InteractableOnFocus : BaseFocusHandler
```
killerantz: This is really cool and having a remote profile on other objects is something I've though about doing for a while.

Some warnings from rendering the inspector for InteractableOnFocus.
`Unexpected top level layout group! Missing GUILayout.EndScrollView/EndVertical/EndHorizontal?
UnityEngine.GUIUtility:ProcessEvent(Int32, IntPtr)`

Two things to consider,
1. It will be interesting to see any other use cases for this pattern, like InteractableOnPress? Do we want to make a base version of this that is more generic? or am I getting ahead of things?
2. When a theme is created the DefaultInteractableStates are added to the theme, which include Default, Focus, Press and Disabled. In this control we only care about default and focus, do we want to only show the default and focus values instead of all 4? There's a way to filter which states show up and which ones will need to be added if they do not exist already. This could be done later, without breaking anything, if we should wait to get some more feedback.

For reference.

Example of created theme with state added at the top

Also, very cool that the MRTK Standard shader is referenced by default for new themes!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/InteractableOnFocus.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            foreach (var profile in this.Profiles)
+            {
+                var themeEngines = profile.CreateThemeEngines();
+
+                themes.AddRange(themeEngines);
+            }
+        }
+
+        private void UpdateThemes(bool isFocused)
```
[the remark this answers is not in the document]
Troy-Ferrell: ahhh good call out!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/InteractableOnFocus.cs
Troy-Ferrell: I'll look into the warnings 

1) This is sort of the goal I think. I think it will only be so fruitful to break out specialties of the main Interactable class. The true win though is making the InteractableThemes extensible. I could as a developer create whatever kind of input handler model I want but then have extensible visual configurations and not have to hardcode anything in my component (like activate this or that). The key for this though is to actually make the InteractableTheme, just a Theme and not dependent on Interactable. It should almost be a separate feature. 

2) Yes I thought about this...but I'm hesitant to do anything atm. I don't want to filter fields because then people might be weirded out why their asset is only half defined. And then for the InteractableOnFocus to force a certain state object kind of like Interactable does is not the direction I want to push forward either...or at least just yet.

Troy-Ferrell:
@killerantz 
All very good feedback!

In the future post-GA, I am interested in working with you & others to refactor the theme system, primarily untangle it from Interactable. There are a few weird aspects that make it hard for things to be populated or configurable or to write code around. 

a) One downside of themes is the inspector code. I could have made InteractableOnFocus just render a property field and you drag/drop the scriptableobject...but it is nicer when you can edit everything in editor without double-clicking down into the asset itself.

b) It's very complicated for a re-usable asset file to have a dependency depending on what it is assigned to per component at runtime which is primarily the assigned Target GameObject. I would have to define multiple themes for every type of shader effectively and is kind of restrictive and not well communicated to the designer. And then there is whether it has text renderers which is per instance not per theme definition etc. 

c) The states object is also defined both in the Theme itself and weirdly overridden by the "profile" containing it. Which is again a bit weird

d) The Theme classes are a bit convoluted to maintain or extend. The theme system is extremely powerful and flexible but the complexity and naming makes it hard to navigate.

killerantz:
I agree and look forward to future conversations.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/VisualProfile.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// Theme scriptableObject for loading theme settings
```
keveleigh: Is this correct? This class isn't a `ScriptableObject`

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/InteractableOnFocus.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// </summary>
+        [SerializeField]
+        [HideInInspector]
+        protected List<VisualProfile> Profiles = new List<VisualProfile>();
+
+        protected InteractableStates.InteractableStateEnum State
+        {
+            get
+            {
+                return HasFocus? InteractableStates.InteractableStateEnum.Focus : InteractableStates.InteractableStateEnum.Default;
```
keveleigh:
```suggestion
                return HasFocus ? InteractableStates.InteractableStateEnum.Focus : InteractableStates.InteractableStateEnum.Default;
```

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Themes/VisualProfile.cs
Troy-Ferrell: No, it was a scriptableobject but I couldn't get the inspector behavior etc I wanted ...fixing

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableOnFocusInspector.cs
```diff
@@ -0,0 +1,94 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities.Editor;
+using System.Collections.Generic;
+using UnityEditor;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.UI.Editor
```
thalbern: add class description please

Issue #5501, opened 2019-08-02
Null Reference Exception in SystemKeyboardExample


Cameron-Micka:
## Describe the bug

When running the HandInteractionExamples the Unity log is spammed with null reference exceptions from the SystemKeyboardExample due to the debug text not being set up in the inspector.

## To reproduce

1. Run the HandInteractionExamples on a HoloLens 2.
2. Download the UnityPlayer.log
3. Notice `NullReferenceException: Object reference not set to an instance of an object.
  at Microsoft.MixedReality.Toolkit.Examples.Demos.SystemKeyboardExample.Update () [0x00000] in <00000000000000000000000000000000>:0 ` throughout the log.

## Your Setup (please complete the following information)

- Unity Version 2018.4.2f1
- mrtk_development

Troy-Ferrell:
Thank you! Just hit this

PR #5502, opened 2019-08-02
Fixing Null Reference Exception in SystemKeyboardExample 

Cameron-Micka:
## Overview

When running the HandInteractionExamples the Unity log is spammed with null reference exceptions from the SystemKeyboardExample due to the debug text not being set up in the inspector. This pull request address this by assigning the reference and performing runtime checks.

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5501

## Verification

1. Run the HandInteractionExamples on a HoloLens 2.
2. Download the UnityPlayer.log
3. Verify no exceptions from Microsoft.MixedReality.Toolkit.Examples.Demos.SystemKeyboardExample.Update

>
>
>

### DIFF
--- a/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Scenes/HandInteractionExamples.unity
+++ b/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Scenes/HandInteractionExamples.unity
@@ -1051,6 +1051,11 @@ PrefabInstance:
       propertyPath: m_LocalScale.z
       value: 1.5000013
       objectReference: {fileID: 0}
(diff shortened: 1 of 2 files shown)


cre8ivepark:
@Cameron-Micka @davidkline-ms I think this should be included in the GA branch. However, this PR was based on dev branch. After David's target branch change, I see many unrelated commits from dev branch. We might need a new PR for GA branch.

davidkline-ms:
yes. we should take a separate PR for stabilization, unless we believe the other fixes contained here (from the dev branch) meet the release bar.

Cameron-Micka:
I can make another PR targeting stabilization if that is desirable?

Issue #5504, opened 2019-08-02
Bounding Box Handle asset cleanup needed for RC1/RC2 to GA upgrade


cre8ivepark:
## Describe the bug
Customers upgrading from RC1, RC2 to GA will see invisible handles since the asset scaling logic has been changed which requires new assets.

PR #5505, opened 2019-08-02
Bounding Box Handle asset cleanup for RC1/RC2 to GA upgrade

cre8ivepark:
## Overview
Based on some internal feedback, unified Bounding Box handle assets. Customers upgrading from RC1, RC2 to GA will see invisible handles(HoloLens 2 style) since the asset scaling logic has been changed which requires new assets. To avoid confusion and to achieve the naming consistency between the inspector and actual handle asset prefabs, updated asset names from 'widget' to 'handle' and deleted old handle assets that won't work with new BB scale values.

## Changes
- Renamed asset names 
- Updated handle assets to new 1m based assets
- Updated handle prefabs
- Updated HandInteractionExample scene, BoundingBoxExamples scene, Slate prefab.

2019-08-02 12_53_28-Unity 2018 4 1f1 Personal - HandInteractionExamples unity - MRTK-Public-Microsof

- Fixes: #5504


cre8ivepark:
Found some additional missing field. Re-assigned them and tested on HoloLens 2.

Issue #5503, opened 2019-08-02
"Make active instance" kills input in editor


Troy-Ferrell:
## Describe the bug

If try to make another gameobject with an MRTK monobehavior the active instance for the entire framework via the editor button "Make this the Active Instance", keyboard input such as camera control or hand simulation breaks. 

Note: The inactive instance does become the active instance or at least appears to.
Note: I have not tried on device to see if the API fails or only in editor

## To reproduce

Open HandINteractionExample scenes
Duplicate MRTK gameobject
Run scene
Navigate around with camera control & hand simulation
On inactive MRTK instance, click "Make this the Active Instance"
Notice now no controls work

Note: Click "Make this the Active Instance" while not in play mode works

## Expected behavior

Switching instances will re-initialize MRTK framework for the new system profiles but this should not break editor interaction

## Your Setup (please complete the following information)

- Unity Version 2018.4.5f1
- MRTK Version 2.0.0_Stabilization (around 8/2)

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

Troy-Ferrell:
@Railboy FYI

Troy-Ferrell:
No longer repos. Closing

PR #5507, opened 2019-08-02
Simplify MRTK service registration & access

Troy-Ferrell:
**IMPORTANT**
This PR is now dependent on #5523 before merging

This change focuses on completing two goals with regards to service registration & access in MRTK. 
1) Simplify system access for consumers without need of redundant local properties calling MixedRealityServiceRegistry.TryGetService()
2) Decrease monolithic infrastructure for service registration in MRTK

How this is accomplished
1) Introduction of CoreServices utility class. System wide reference cache for selective core MRTK systems
2) Initial ingestion of MixedRealityServiceRegistry & CoreServices classes into MixedRealityToolkit class as well as initial deprecation of related interfaces

MRTK Services
-> Added functions to MixedRealityServiceRegistry for GetAllServices() & ClearAllServices()
-> Deprecated Core System References in MixedRealityToolkit.cs
-> Deprecated ActiveSystems & RegisteredMixedRealityServices
-> Part of the internal MRTK object would actually de-register all services so that they could be re-registered in order. The collection was never marked as ordered or commented anywhere. If a function wants to execute on the services list in order, then it must order them itself. This has been updated with an ExecuteAllServicesInOrder() function.
-> MRTK object now uses CoreServices instead of it's internal "cache" references
-> MRTK object now utilize both the Registry & it's internal dictionary for Register, Get, and Unregister interfaces
-> Updated MRTK.IsSystemRegistered<T> to force only IMixedRealityService templates
-> Redundant functionalty of IsServiceRegistered & IsSystemRegistered. Deprecated IsServiceRegistered in favor of IsSystemRegistered
-> Updated tests to account for deprecation
-> OnValidate in MRTK would try to re-initialize services when a new active instance was assigned....but the old instance never destroyed or cleaned up

## Changes
Fixes: #5021 

## Verification
-> All tests passed
-> Test experimental prefab stuff*
-> Built HandInteractionExample in editor & on device
-> Opening & running various scene for editor MRTK use
-> Tested Boundary Visualization example scene with VR device
-> Tested BoundarySystem standalone-prefab from experimental with VR device
-> Testing "Make active instance" dynamically already broken via #5503
-> Updated documentation for deprecated fields

>
>


wiwei:
@Troy-Ferrel, the CoreServices.cs part seems good to me (in terms of reducing the copy pasta of locally caching things in various places) - the other stuff feels a bit risky given the timing now.

I would really have to defer to how @davidkline-ms feels about the other stuff, both from a code and also from a stability/risk perspective.

davidkline-ms:
@wiwei, Troy and i are collaborating on this change. We will ensure that the stability is high and the risk is acceptable.

Assets/MixedRealityToolkit/Utilities/CoreServices.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            if (serviceCache.ContainsKey(serviceType))
+            {
+                return (T)serviceCache[serviceType];
+            }
+            else
+            {
+                T service;
+                MixedRealityServiceRegistry.TryGetService<T>(out service);
+                if (service == null)
```
davidkline-ms: this code feels a bit odd.  I recommend doing something like:

```
T service;

if  (!MixedRealityServiceRegistry.TryGetService<T>(out service))
{
    return default(T);
}

serviceCache.Add(typeof(T), service as IMixedRealityService);
return service;
```

Troy-Ferrell:
I believe the one failure in the CI here is #5552 but of course there is an additional bug with the CI test results write operation failing so can't know for sure. Re-running again

Documentation/UpdatingToGA.md
```diff
@@ -9,10 +9,28 @@ Between the RC2 and GA releases of the Microsoft Mixed Reality Toolkit, changes
...
+- [MixedRealityToolkit](#mixedrealitytoolkit)
 - [Event System](#event-system)
 - [Spatial Awareness](#spatial-awareness)
 - [Clipping Sphere](#clipping-sphere)
 
+### MixedRealityToolkit
+
+The following public properties on the MixedRealityToolkit object have been deprecated.
+
+- `RegisteredMixedRealityServices` - Use [`IMixedRealityServiceRegistrar.GetDataProvider<T>()`](xref:Microsoft.MixedReality.Toolkit.IMixedRealityServiceRegistrar) or [`IMixedRealityDataProviderAccess`](xref:Microsoft.MixedReality.Toolkit.IMixedRealityDataProviderAccess) instead
```
davidkline-ms: Use [`IMixedRealityServiceRegistrar.GetDataProvider<T>()`]

should no longer be valid.

davidkline-ms:
Nice clean changes! I'm going to confirm that the MixedRealityToolkit.*DataProvider* methods are removed and fix the RC->GA doc. Should be good to go after that!

davidkline-ms:
made a minor change to the docs. rerunning the pr checks

Issue #5469, opened 2019-07-30
Can't build HoloLens project after importing MRTK due to missing DLL's


CookieHael:
## Describe the bug

Building in VS to deploy to HoloLens throws a bunch of errors about not finding MRTK dll's where it expects them. See screenshot for specific errors. Can't seem to reproduce fully on a new project, but at least one error that is thrown in my project also gets thrown.

## To reproduce

1. Make new project, set it up for HoloLens
2. Import MRTK
3. Build in Unity and open the solution that is created in the selected folder in VS, build it there.

## Expected behavior

Build properly

## Screenshots

<img width="726" alt="errors" src="https://user-images.githubusercontent.com/36892196/62130695-dd2ef400-b2d9-11e9-985a-e1ff685969b2.png">

## Your Setup (please complete the following information)

- Unity Version 2018.4.4f1 (.NET backend, set up as explained in the microsoft hololens development docs)
- MRTK Version v2.0.0 RC2.1
- Visual Studio 2019 16.2  (also tried building in 2017, same errors. When doing a manual check the actual missing files are just not present where the solution thinks they are.)

## Target Platform (please complete the following information)

- HoloLens

wiwei:
@CookieHael, this is gonna be weird to ask, but can you try putting your project in a folder closer to the root of the drive?

i.e. instead of c:\very\long\path\to\stuff\woaaaaaaaaaaaaaaaw

Do: c:\src

I think that Unity has trouble with very long paths (i.e. longer than MAX_PATH which is 255 characters - I think that the paths in the image above look to be super long)

If this is the case I'll have to update the docs to make sure we warn off of this

CookieHael:
Hi, I'll try that ASAP, gotta get to our IT guy then cause they have my laptop locked down so I can't get to the C drive. If this is the case, it'd probably also be helpful in preventing the error if you guys shortened down your pathnames as well along with telling people to keep their paths short, since "Microsoft.MixedRealityToolkit.something.something.anotherthing.dll" isn't all too short either hehe ;)

CookieHael:
It looks like that fixed it! Tried moving it to C:\users\rhelsen yesterday but even that was too long apparently! It's building fine now! Thanks 👍

wiwei:
Okay I'll use this bug to track updating the docs about this, so hopefully folks can at least search for the problem if they hit it themselves.

[...]

PR #5500, opened 2019-08-02
Moving the HandConstraint Solver from Experimental to Release

Cameron-Micka:
## Overview
This pull request moves the HandConstraint  (and HandConstraintPalmUp) solver from Experimental to Release by adding tests, documentation, and a handful of bug fixes and improvements. Functionality remains pretty much the same from the previous pull request (https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4532) and utilizes solver fixes from @Troy-Ferrell's (https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5416)

The solver has also be genericized to work with motion controllers.

Thank you @julenka for guidance on how to write unit tests that utilize hands!

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5312, https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5304, https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5257, https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5413

## Verification
Please try HandBasedMenuExample scene.
>
> As a reviewer, it is possible to check out this change locally by using the following
> commands (substituting {PR_ID} with the ID of this pull request):
>
> git fetch origin pull/{PR_ID}/head:name_of_local_branch
>
> git checkout name_of_local_branch

PR #5508, opened 2019-08-02
Add a known issues page to the MRTK.

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5469

While we track all of our issues on github, there are some that folks may hit more often - for the ones that are both common and have workarounds (where the real solution may be somewhat in the future) having this known issues page can be helpful.

### DIFF
--- a/Documentation/KnownIssues.md
+++ b/Documentation/KnownIssues.md
@@ -1,0 +1,25 @@
+# Known Issues
+
+This page covers common issues that one might encounter when using the MRTK and their corresponding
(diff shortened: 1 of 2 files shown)


Troy-Ferrell:
@Cameron-Micka FYI for related to your handconstraint PR #5500 

Maybe the .NET workaround info would go in here?

davidkline-ms:
Should this target the stabilization branch?

wiwei:
> What about TOC update though?

I thought I was missing something.....

Issue #5506, opened 2019-08-02
Input simulation 'tap/press' does not work 


cre8ivepark:
## Describe the bug
When I click the mouse, instead of tap/press, the simulated hand only moves vertically.

## To reproduce
Run any scenes in the Unity editor. Click left mouse button to simulate press.

## Screenshots
![MRTK_InputSimulation2](https://user-images.githubusercontent.com/13754172/62397169-ddbdc980-b529-11e9-881b-5946ad54a526.gif)

![MRTK_InputSimulationIssue](https://user-images.githubusercontent.com/13754172/62396533-4015ca80-b528-11e9-8494-d5afb32af309.gif)

## Branch
prerelease/2.0.0_stabilization

Cameron-Micka:
I've re-produced this locally too.

Troy-Ferrell:
Same, it's in stabilization

davidkline-ms:
this was addressed by #5513, if this does not fix the issue, please reopen with any additional information. thanks!

PR #5495, opened 2019-08-02
Fix grab ray pos

MenelvagorMilsom:
## Overview
Moves the grab ray position to the average of the thumb tip and the index tip and fix tests affected by this change.

## Changes
- Fixes: #5450.
- Grab ray position change.
- Change the pinch hand gesture JSON to centre on the new grab point.
- Add a new open hand gesture that is centred on the grab point (mostly for testing)
- Some minor gesture changes to a couple of tests
- Removed NearInteractionTouchableStack test as it was broken (see comments in file)

Issue #5450, opened 2019-07-29
Start of grab ray should be average of index and thumb position


MenelvagorMilsom:
## Describe the bug
`TryGetNearGraspPoint` summary says that the near grasp point should be the average of the index tip position and the thumb tip position. There is commented code here that should do the job:
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/24ad9e66f7aeaa6edd0912f0f17db294ce25e55c/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs#L122

Just need to look into this and make sure that everything still works with this line uncommented.

MenelvagorMilsom:
There's a bit of work to get the tests working with this change. I have a plan of action but I doubt I will get this finished tonight. @wiwei Is this something I can create a PR directly into stabilization for?

PR #5513, opened 2019-08-05
Fix the backwards-incompatible enum change

wiwei:
A previous change to this updated an existing enum by adding a value in the middle of the enum - this is backwards incompatible because the same value is persisted in various profiles, where enums are stored as integer values, rather than string values.

This means that something that used to refer to "pinch" now referred to this new thing, and the same off-by-one problem would affect all of the other options below the new enum

See change:
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5495

Original issue:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5450

Current issue:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5506

### DIFF
--- a/Assets/MixedRealityToolkit/Definitions/Utilities/ArticulatedHandPose.cs
+++ b/Assets/MixedRealityToolkit/Definitions/Utilities/ArticulatedHandPose.cs
@@ -154,10 +154,6 @@ namespace Microsoft.MixedReality.Toolkit.Utilities
             /// </summary>
             Open,
             /// <summary>
-            /// Relaxed hand pose, grab point does not move
-            /// </summary>
-            OpenSteadyGrabPoint,
-            /// <summary>
             /// Index finger and Thumb touching, grab point does not move
             /// </summary>
             Pinch,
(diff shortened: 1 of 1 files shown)


wiwei:
With this change, the finger still gets offset, @MenelvagorMilsom  FYI.

This change is purely about fixing up the pinch/click functionality to work.

Assets/MixedRealityToolkit/Definitions/Utilities/ArticulatedHandPose.cs
```diff
@@ -181,6 +177,10 @@ public enum GestureId
             /// Victory sign
             /// </summary>
             Victory,
+            /// <summary>
+            /// Relaxed hand pose, grab point does not move
```
Troy-Ferrell: nit: since we're here, should we start labeling each value instead of auto-increment i.e
Victor = x,
OpenSteadyGrabPoint = y 

and make this trend for all other enums going forward?

wiwei: I can take a separate pass for existing enums, but yeah, this would be good to do in the future

Issue #4444, opened 2019-05-17
Solver documentation is sparse


ritijain:
## Describe the Issue
Solver model is too convoluted for ease of understanding, and the documentation doesn't help.

## Feature Area
It is hard to leverage the Solver script, and its implementations.
1. Some of the solvers exhibit too many settings. The tooltips on those are not very helpful and can be more descriptive.
2. It would be great if the Solver implementations that come with MRTK have separate documentation pages. The number of settings exposed in them can be a little overwhelming and not all settings inherited from SolverBase.cs are applicable to all types it seems. 

A little more focus on usage of the scripts rather than intent would be helpful. 

## Existing Doc Link
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/README_Solver.html

ritijain:
@gejohnst  - I believe you are addressing some of these issues?

ritijain:
Related - https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4675

wiwei:
Assigning to @gejohnst because he's working on some runtime-programmability of solvers and solver documentation.

Issue #5091, opened 2019-06-27
SolverExamples scene is broken


Troy-Ferrell:
## Describe the bug

The solverexamples scene is broken for testing.

1) Prefabs are spawned and assignment of tracked transform is incorrect for hands. In simulation hand parent gameobjects are always at <0,0,0> while the joint transforms are being updated to world positions. The prefab spawning also seems overkill when you cannot create multiple gameobjects after selecting a solver option

2) Clickable spheres have no interactable profiles to indicate if in focus or onclick or even if that solver is active in scene

3) Not all solvers are shown in example

## To reproduce

Open the solver examples scene and run in editor

## Your Setup (please complete the following information)

- Unity Version 2018.4
- MRTK Version Latest

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

PR #5514, opened 2019-08-05
Solver Examples & Documentation Updates

Troy-Ferrell:
## Overview

This change address additional issues with the Solver example scenes & updates the solver documentation. There are additional minor fixes discovered for Solvers in here as well. See list below for all change details

- Redesigned Solver example scene to be functional & easier to understand
- Provided example usage of InBetween in scene
- Remove SwapVolume script
- Delete outdated Solver example prefabs (most were nonfunctional)
- Updated solver text panel in example scene
- Updated SolverHandler DestroyImmediate to Destroy
- Fix SolverHandler issue in editor when destroying tracked transform unnecessarily
- Fixed inspector isssue with enum popups not working with prefabs well
- Fixed InBetween SecondTrackedTargetType to not show obsolete enum possibilites
- Added warning to SurfaceMagnetism component for the MagneticSurfaces property. One common mistake is not marking a SurfaceMagnetism gameObject as "Ignore Raycast" so it doesn't "Magnetize" against itself
- Added (Obsolete) term to TrackedObjectType summaries
- SurfaceMagnetism did not check SolverHandler.TransformTarget was not null on raycast calculation
- Add new InBetween documentation with images & gif
- Added sections on Orbital & RadialView & SurfaceMagnetism to cover key propertise & common mistakes
- Also added doc link button to SolverHandler

image

SolverExampleScene

## Changes
- Fixes: #4444 , #5091 

## Verification
Ran all playmode & editmode tests
Utilize Solvers in new example scene

>
>


Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            if (handler != null)
+            {
+                this.handler.TrackedTargetType = this.TrackedType;
+                this.handler.TrackedHandness = Handedness.Both;
+                if (this.CustomTrackedObject != null)
+                {
+                    this.handler.TransformOverride = this.CustomTrackedObject.transform;
+                }
+            }
```
wiwei: Nit: Thoughts on dropping this. from this file (i.e. there are a bunch of references to this. where usually we haven't done so in our codebase (and also, in the rest of the file, "this." isn't used)

Assets/MixedRealityToolkit.SDK/Inspectors/Utilities/Solvers/SurfaceMagnetismInspector.cs
```diff
@@ -73,6 +74,19 @@ public override void OnInspectorGUI()
             // Raycast properties
             EditorGUILayout.LabelField("Raycast Properties", EditorStyles.boldLabel);
             EditorGUILayout.PropertyField(magneticSurfacesProperty, true);
+
+            // When raycast from the center of the GameObject, Raycast may hit one of the collider on the GameObject (or children)
+            // This results in the GameObject "magnetizes" against itself. Warn user if this possiblity exists
+            var colliders = surfaceMagnetism.GetComponentsInChildren<Collider>();
+            foreach(var collider in colliders)
```
wiwei:
```suggestion
            foreach (var collider in colliders)
```

Documentation/README_Solver.md
```diff
@@ -1,40 +1,155 @@
...
+
+> [!NOTE]
+> Examples of how to use the Solvers system can be found in the **SolverExamples.scene** file.
+
+## How to change tracking reference
+
+The *Tracked Target Type* property of the [`SolverHandler`](xref:Microsoft.MixedReality.Toolkit.Utilities.Solvers.SolverHandler) component defines the point of reference all solvers will use to calculate their algorithms. For example, a value type of [`Head`](xref:Microsoft.MixedReality.Toolkit.Utilities.TrackedObjectType.Head) with a simple [`SurfaceMagnetism`](xref:Microsoft.MixedReality.Toolkit.Utilities.Solvers.SurfaceMagnetism) component will result in a raycast from the head and in the direction of the user's gaze for solving what surface is hit. Potential values for the `TrackedTargetType` property are:
+
+* *Head* : Point of reference from the camera
+* *MotionController*: Point of reference from a motion controller. Use `TrackedHandedness` property to define details
```
wiwei: Nit: "property to affect the handedness preferences (left, right, either)"

I think details seems interesting, and I think with not too many more words we could say what they are (i.e. left/right/both)

Troy-Ferrell: You're right. I got lazy. Docs can just....keep going on forever :P

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
Troy-Ferrell: Sure, I personally like using this but you're right in that this is not common in the codebase

Documentation/README_Solver.md
```diff
@@ -1,40 +1,155 @@
...
+
+![Solver](../Documentation/Images/Solver/TrackedObjectType-Example.gif)
+<br/>
+*Example of various properties associated with each TrackedTargetType*
+
+## How to chain Solvers
+
+It is possible to add multiple `Solver` components to the same GameObject thus chaining their algorithms. The `SolverHandler` components handles updating all solvers on the same GameObject. By default the `SolverHandler` calls `GetComponents<Solver>()` on Start which will return the Solvers in the order that they appear in the inspector.
+
+Furthermore, setting the *Updated Linked Transform* property to true will instruct that `Solver` to save it's calculated position, orientation, & scale to an intermediary variable accessible by all Solvers (i.e `GoalPosition`). When false, the `Solver` will update the GameObject's transform directly. By saving the transform properties to an intermediary location, other Solvers are able to perform their calculations starting from the intermediary variable. This is because Unity does not allow updates to gameObject.transform to stack within the same frame.
```
wiwei: I wonder if this property should have been named "allow multiple solvers" or something similar to that - I guess this is probably a discussion for a future day, but it's interesting in that, from a user-API perspective, it feels like the property should be named according to intent for what the user is trying to do, but it's more named from an implementation perspective (i.e. it does what it says, but missing is the overall intent)

Troy-Ferrell: I agree. Even "Is Chained Solver" maybe? On the fence about renaming. We could also just update name in inspector but not code

Documentation/README_Solver.md
```diff
@@ -1,40 +1,155 @@
...
 
 The third category is the solver itself. The following solvers provide the building blocks for basic behavior:
 
-* **Orbital**: Locks to a specified position * offset from the referenced object.
-* **ConstantViewSize**: Scales to maintain a constant size relative to the view of the referenced object.
-* **RadialView**: Keeps the object within a view cone cast by the referenced object.
-* **SurfaceMagnetism**, casts rays to surfaces in the world, and align the object to that surface.
-* **Momentum**: Applies acceleration/velocity/friction to simulate momentum and springiness for an object being moved by other solvers/components.
-* **InBetween**: Keeps an object in between two tracked objects.
+* [`Orbital`](xref:Microsoft.MixedReality.Toolkit.Utilities.Solvers.Orbital): Locks to a specified position and offset from the referenced object.
```
wiwei: Consider linking to the text sections below, which provide I think more helpful resources for folks getting started in this area. Down below there are xrefs anyway, so nothing is really lost I think, but you gain from pointing to the most helpful getting started/usage docs

Troy-Ferrell: Good idea

Documentation/README_Solver.md
Troy-Ferrell: There is already a lot of changes here. Let's leave this out for now

cre8ivepark:
Beautiful!

Issue #5384, opened 2019-07-19
Support 2019 for NuGet


andreiborodin:
Currently the MSBuild project generation doesn't work for 2019. To make it work, there needs to be relatively small modification to support BuiltInPackages:
2019.1.6f1\Editor\Data\Resources\PackageManager\BuiltInPackages

This will follow the same approach as standard packages, but putting this work after the current effort is complete.

keveleigh:
Looks like this was fixed via https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5516

PR #5516, opened 2019-08-05
Handle Built-In Packages for MSBuild Project Generation (NuGet Work)

andreiborodin:
## Overview
This change updates the handling of packages for MSBuild project generation to support built-in packages, which is part of the work for #5384. This is done by updating how projects are parsed and dependencies are specified, as well as content items for packages.

## Changes
- Fixes: #5384


Assets/MixedRealityToolkit.Tools/MSBuild/MSBuildTemplates/SDKProjectTemplate.csproj.template
```diff
@@ -17,6 +17,7 @@
     <RelativeBuildFolder>..\Output\</RelativeBuildFolder>
     <BuildFolder>$(MSBuildThisFileDirectory)$(RelativeBuildFolder)</BuildFolder>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+	<Deterministic>false</Deterministic>
```
wiwei:
```suggestion
    <Deterministic>false</Deterministic>
```

wiwei: I think this is a tab that snuck in

wiwei: ><Deterministic>false</Deterministic> [](start = 1, length = 36)

Do you have some more details to share on why we're moving to non-deterministic builds (i.e. when deterministic appears to be the default and the direction that things are moving in?)

Even having some like, maybe some comments inline here as well would be helpful to future readers

Assets/MixedRealityToolkit.Tools/MSBuild/Scripts/AssemblyDefinitionInfo.cs
```diff
@@ -17,15 +20,74 @@ public class AssemblyDefinitionInfo
         public const string EditorPlatform = "Editor";
         public const string TestAssembliesReference = "TestAssemblies";
 
+        /// <summary>
+        /// Creates an instance of <see cref="AssemblyDefinitionInfo"/> for the default projects (such as Assembly-CSharp)
+        /// </summary>
+        /// <param name="assembly">The Unity assembly reference.</param>
+        /// <returns>A new instnace.</returns>
```
wiwei:
```suggestion
        /// <returns>A new instance.</returns>
```

Assets/MixedRealityToolkit.Tools/MSBuild/Scripts/AssemblyDefinitionInfo.cs
```diff
@@ -93,6 +180,60 @@ public void Validate(IEnumerable<CompilationPlatformInfo> availablePlatforms)
...
+        /// </summary>
+        /// <returns>The array of <see cref="SourceFileInfo"/>.</returns>
+        public SourceFileInfo[] GetSources()
+        {
+            if (sourceFiles == null)
+            {
+                List<SourceFileInfo> sourceFiles = new List<SourceFileInfo>();
+                if (assembly != null)
+                {
+                    foreach (string file in assembly.sourceFiles)
```
wiwei: >file [](start = 36, length = 4)

Super nit: sourceFile - "file" as a var already exists in the class context, and when paired with the reference down below (i.e. file.Directory in the else statement) it seeeeems a little confusing

Assets/MixedRealityToolkit.Tools/MSBuild/Scripts/UnityProjectInfo.cs
```diff
@@ -20,107 +19,189 @@ namespace Microsoft.MixedReality.Toolkit.MSBuild
...
+                }
+            }
 
-                    toAdd = new CSProjectInfo(availablePlatforms, Guid.NewGuid(), null, pair.Value, projectOutputPath);
+            CSProjects = new ReadOnlyDictionary<string, CSProjectInfo>(CreateUnityProjects(projectOutputPath));
+        }
+
+        private Dictionary<string, CSProjectInfo> CreateUnityProjects(string projectOutputPath)
+        {
+            // Not all of these will be inflated
```
wiwei: >inflated [](start = 40, length = 8)

This term is new to me, and I can't seem references to this in the rest of this file - is inflation a term for the overall process of expanding out the templates and stuff?

Assets/MixedRealityToolkit.Tools/MSBuild/Scripts/Utilities.cs
```diff
@@ -23,15 +44,25 @@ public static class Utilities
...
+
+        public static string AssetPath { get; }
+
+        public static string BuiltInPackagesPath { get; }
 
         static Utilities()
         {
-            assetsPath = Path.GetFullPath(Application.dataPath);
+            AssetPath = Path.GetFullPath(Application.dataPath);
+            packagesPath = Path.GetFullPath(Application.dataPath.Replace("Assets", "Packages"));
```
wiwei: >Application.dataPath.Replace("Assets", "Packages") [](start = 44, length = 50)

Be wary when the path leading up the application path can also contain the "Assets" word -  

[https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5452/files](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5452/files)

Otherwise a path like: d:/Assets/MyProject/Assets would turn into:

d:/Packages/MyProject/Packages

Where only the last one is the one we really care to replace

Same about any usage of "replace" elsewhere - I *THINK* that Assets is probably the most general so probably the one to really watch out for.

wiwei:
@andreiborodin lgtm, some minor comments

Assets/MixedRealityToolkit.Tools/MSBuild/MSBuildTemplates/SDKProjectTemplate.csproj.template
andreiborodin: One of the packages has an AssemblyInfo that has [assembly: AssemblyVersion("X.Y.*")] where X and Y are numbers, but the * fails the build if it is deterministic

---
In reply to: [310749157](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5516#discussion_r310749157) [](ancestors = 310749157)

Assets/MixedRealityToolkit.Tools/MSBuild/Scripts/UnityProjectInfo.cs
andreiborodin: Ah, inflated as in converted to a C# project, will update wording.

---
In reply to: [310753916](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5516#discussion_r310753916) [](ancestors = 310753916)

Assets/MixedRealityToolkit.Tools/MSBuild/Scripts/Utilities.cs
andreiborodin: Agreed, will update.

---
In reply to: [310757064](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5516#discussion_r310757064) [](ancestors = 310757064)

Issue #5423, opened 2019-07-25
Controllers  shown at original place after teleporting in VR.


rahsingh-ms:
## Describe the bug

We are using the WindowsMixedRealityControllers for our App's VR version and when we attempt to teleport in our app, We can see camera hands at the original position. 

## To reproduce

1. Open VR app on Windows
2. Try teleporting/moving the main camera to some other location
3. We can see controllers left behind at original position.

## Describe the solution:

This issue is happening because of an error in WindowsMixedRealityController.cs class.
In the function 
private async void CreateControllerModelFromPlatformSDK(uint interactionSourceId)

when the GLTF is found from platfor, while creating the gameobject for controller visualization, we are not adding it to under MixedRealityPlayspace but it is getting added directly to root scene which is causing the issue of seeing controllers at original position after teleportation.

![image](https://user-images.githubusercontent.com/17119496/61846260-145d6980-aec4-11e9-981d-f25d546359f5.png)

The fix is in the BaseController.cs file.
In the function: 
protected virtual bool TryRenderControllerModel(Type controllerType, InputSourceType inputSourceType)

we add MixedRealityPlayspace.AddChild(controllerObject.transform);

This line should actually be moved inside TryAddControllerModelToSceneHierarchy function

so that even if we use the platform controller visualization, they will be correctly added to under MixedRealityPlaySpace game object.

## Your Setup (please complete the following information)

issue on mrtk_development branch

## Target Platform (please complete the following information)

- WMR immersive

keveleigh:
Huh, this was something I noticed and thought I had fixed as part of https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4707.
The pointer and grip poses were being properly transformed based on the transform of the playspace, but the source pose wasn't. I'll take a look.

keveleigh:
Actually, should have been fixed via https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5250

rahsingh-ms:
Hi, you are right . But the problem here that for specific case of WMR controllers, in the CreateControllerModelFromPlatformSDK() function of WindowsMixedRealityController.cs file, if we get the fileStream for controller from platform, in that case, erroneously the Visualization gameobject is getting added to the scene and not under the MixedRealityPlaySpace gameobject and that is why the teleportation is not working.

PR #5528, opened 2019-08-06
Fix for issue #5423

anuraag016:
Fix for issue #5423 [Controllers shown at original place after teleporting in VR.](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5423)

## Overview
This fix addresses the issue #5423 [Controllers shown at original place after teleporting in VR.](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5423)

## Changes
The fix is in the BaseController.cs file.
In the function:
protected virtual bool TryRenderControllerModel(Type controllerType, InputSourceType inputSourceType)

we add MixedRealityPlayspace.AddChild(controllerObject.transform);

This line should actually be moved inside TryAddControllerModelToSceneHierarchy function

so that even if we use the platform controller visualization, they will be correctly added to under MixedRealityPlaySpace game object.

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/BaseWindowsMixedRealitySource.cs
@@ -106,14 +106,12 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input
         public void UpdateVelocity(InteractionSourceState interactionSourceState)
         {
             Vector3 newVelocity;
-            bool isVelocityValid = interactionSourceState.sourcePose.TryGetVelocity(out newVelocity);
-            if (isVelocityValid)
+            if (interactionSourceState.sourcePose.TryGetVelocity(out newVelocity))
             {
                 Velocity = newVelocity;
             }
             Vector3 newAngularVelocity;
-            bool isAngularVelocityValid = interactionSourceState.sourcePose.TryGetAngularVelocity(out newAngularVelocity);
-            if(isAngularVelocityValid)
+            if (interactionSourceState.sourcePose.TryGetAngularVelocity(out newAngularVelocity))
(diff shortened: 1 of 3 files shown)


wiwei:
@keveleigh I think we were just chatting about this last week - was this one of the options you were thinking about?

keveleigh:
Unfortunately, I don't believe just this change fixes it (after #5250). I implemented this locally last week and saw incorrect behavior in mrtk_dev.

I'm thinking through a combination of reverting #5250 and adding this change as part of more clearly defining the expected data from source pose events (world space vs local space).

ritijain:
@keveleigh  - is this some

> Unfortunately, I don't believe just this change fixes it (after #5250). I implemented this locally last week ([keveleigh@ba647fc](https://github.com/keveleigh/HoloToolkit-Unity/commit/ba647fc67c976db744f214ea14ecc9c1a7e25cf5)) and saw incorrect behavior in mrtk_dev.
> 
> I'm thinking through a combination of reverting #5250 and adding this change as part of more clearly defining the expected data from source pose events (world space vs local space).

@keveleigh is this something you have the bandwith to try? What was the undesirable behavior that you observed, so that @anuraag016 can try it out as well?

keveleigh:
@ritijain

>is this something you have the bandwith to try?

I've blocked out some time tomorrow to focus on some MRTK issues, including this one!

>What was the undesirable behavior that you observed

After teleporting, the controllers were in an unexpected location. I had originally fixed that with #5250, which is in mrtk_dev but not RC2.1, but I'm no longer convinced I fixed it correctly, after the issue @anuraag016 reported here. Combining the two fixes breaks things again though

keveleigh:
@wiwei @davidkline-ms
I merged my work into this PR, to account for both this more correct fix and the changes I made in #5250. I decided to take the path of defining both events as world space data, since that's likely more useful for an end developer.

I believe this PR is ready for review now!

PR #5530, opened 2019-08-06
Update the build system to have a hook to check for validation errors in the MRTK build

wiwei:
Recently we hit an issue in the MRTK where we had duplicate GUIDs checked in for a few different assets:

#5431

The root issue here was that the meta files were copy pasted, instead of being regenerated at the new location. There was a warning in the console but it was ignored, and since Unity will non-deterministically choose one of them (i.e. either of the duplicate GUIDs could be right) - you end up with cases where you will get one or the other of the assets (and hopefully Unity choose correctly for you!)

I looked into doing a custom logger (i.e. catching LogWarning) but found that it doesn't run soon enough (i.e. warnings will show up, this code will get compiled) so this had to resort to editor-log parsing.

In theory this could be extended in the future to other issues that we hit that show up in the console as well.

### DIFF
--- a/pipelines/templates/tasks/unitybuild.yml
+++ b/pipelines/templates/tasks/unitybuild.yml
@@ -123,6 +123,14 @@ steps:
    }
   displayName: "Build ${{ parameters.Platform }} ${{ parameters.Arch }} ${{ parameters.ScriptingBackend }}"
 
+- task: PowerShell@2
+  displayName: 'Validate build logs'
+  inputs:
+    targetType: filePath
+    filePath: ./scripts/ci/validatebuildlog.ps1
+    arguments: >
+      -LogFile: '$(Build.ArtifactStagingDirectory)\build\${{ parameters.Platform }}_${{ parameters.Arch }}_${{ parameters.ScriptingBackend }}\build\build.log'
+
 - task: PublishBuildArtifacts@1
   enabled: ${{ parameters.PublishArtifacts }}
   displayName: 'Publish ${{ parameters.Platform }} ${{ parameters.Arch }} (${{ parameters.PackagingDir }}) ${{ parameters.ScriptingBackend }}'
--- a/scripts/ci/validatebuildlog.ps1
+++ b/scripts/ci/validatebuildlog.ps1
@@ -1,0 +1,75 @@
+<#
+.SYNOPSIS
+    Validates a build log to check for the presence of non-build related warnings
+    that may negatively affect the quality of the build.
+.DESCRIPTION
+    This script intends to prevent repeats of past issues that have come up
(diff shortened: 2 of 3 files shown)


wiwei:
Note for reviewers, the .meta file is for testing, I will revert that once I have verified that this is working as intended on mrtk_pr

wiwei:
It works!

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=3945

scripts/ci/validatebuildlog.ps1
```diff
@@ -0,0 +1,75 @@
...
+    .SYNOPSIS
+        Checks if the given log file has a conflict guid at the specified line number.
+        Outputs to console if such an error exists.
+        Returns true if there is a duplicate guid.
+    #>
+    if ($LogFileContent[$LineNumber] -match "GUID \[[a-g0-9]{32}?\] for asset '.*' conflicts with") {
+        for ($i = $LineNumber; $i -lt $LogFileContent.Length; $i++) {
+            if ($LogFileContent[$i] -eq "Assigning a new guid.") {
+                Write-Host "Found duplicated GUID, Unity will non-deterministically use one of them, please manually"
+                    "regenerate the intended one by deleting the .meta file and re-opening the Unity editor locally"
```
keveleigh: image
Doesn't look like this line prints

wiwei: Oh my

pipelines/templates/tasks/unitybuild.yml
```diff
@@ -123,6 +123,14 @@ steps:
    }
   displayName: "Build ${{ parameters.Platform }} ${{ parameters.Arch }} ${{ parameters.ScriptingBackend }}"
 
+- task: PowerShell@2
+  displayName: 'Validate build logs'
+  inputs:
+    targetType: filePath
+    filePath: ./scripts/ci/validatebuildlog.ps1
+    arguments: >
+      -LogFile: '$(Build.ArtifactStagingDirectory)\build\${{ parameters.Platform }}_${{ parameters.Arch }}_${{ parameters.ScriptingBackend }}\build\build.log'
```
macborow: It might be nice to add logDirectory and logPath as task parameters (with default values). This would ensure that the log path is always the same in unity build task and any log processing tasks that follow.
There's a way to emit a value from a task and reuse it (using ##vso tags) but I'm not sure how to use it effectively (something I need to experiment a bit more with) and parameters are easy to use.

wiwei: This is a good idea - I'm going to look into this after getting this change in, if only because I will need a little more time to figure out how to refactor this stuff to make the string interpolation work with other parameters (in the param section).

PR #5519, opened 2019-08-05
Merge stabilization -> dev

davidkline-ms:
This change merges the stabilization branch into the mrtk_development branch.

PLEASE NOTE: There was a merge conflict in HandInteractionExamples.unity that was resolved by taking the version from prerelease/2.0.0_stabilization,

PR #5531, opened 2019-08-06
[Cherry-pick from stabilization] add namespace to solverinspector

davidkline-ms:
This change takes the solver inspector namespace fix from the stabilization branch to resolve the asset retargeting error in ci


keveleigh:
I believe this is no longer needed after #5519

Troy-Ferrell:
> I believe this is no longer needed after #5519

@davidkline-ms yes it should already be there:
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.SDK/Inspectors/Utilities/Solvers/SolverInspector.cs

davidkline-ms:
Yep. This was opened to add just that fix to address a CI failure while we worked on the larger merge.

Issue #4800, opened 2019-06-10
Horizontal Surface Magnetism, z-axis Rotation Issue


orenono:
## Describe the bug
I'm using Solver Handler+  Surface Magnetism to have an object float over the room's surfaces.

Everything works as expected on vertical surfaces, however on horizontal surfaces the object will often keep rotating on the Z-axis without stabilizing according to the headset location. 

## To reproduce

I've attached a screenshot of the settings.
## Expected behavior

When placing the object on horizontal surfaces, the object's z-axis should face the HoloLens's direction (as it does on vertical surfaces).

## Screenshots

![Settings](https://user-images.githubusercontent.com/10733786/59207015-2c776480-8b74-11e9-9138-746862fd9904.JPG)

## Your Setup (please complete the following information)

- Unity Version 2018.3.14f
- MRTK Version 2 RC1 Refresh

## Target Platform (please complete the following information)

- HoloLens

Troy-Ferrell:
@orenono there is PR (linked above) to fix this but if I understand your desires correctly, you actually have the wrong parameters set. You should set OrientationMode to "Vertical" (which will translate to "TrackedTarget" - just a rename, same enum value). This will keep the GameObject facing the camera.

PR #5532, opened 2019-08-06
Updates to fix SurfaceMagnetism orientation

Troy-Ferrell:
## Overview
This change fixes a faulty orientation setting in SurfaceMagnetism as well as provides clearer naming & documentation.

Fixes: 

- The Blended mode for OrientationMode in SurfaceMagnetism is supposed to blend rotation input calculated via the Tracked Transform and the normal at the surface hit point. The code was actually only blending between the current model rotation and the tracked transform rotation. And thus, the OrientationBlend property had no impact what so ever.

- Furthermore, the original OrientationMode *Vertical* in documentation is supposed to face the tracked transform (i.e user etc) but actually faced along the surface normal. 

- To simplify implementation, the property *KeepOrientationVertical* was added to give user's more control for their chosen OrientationMode (see related bug #4800 )

- OrientationMode keeps original values but renamed to be more accurate. 
- Updated SurfaceMagnetismInspector to hide OrientationBlend if not in Blended Mode and to hide both that and KeepOrientationVertical if OrientionMode is None
- All related tooltip/summaries are updated to be both clearer & accurate
- Solver documentation page has been updated accordingly
- Added public properties to Solver for LerpTime etc
- Updated SolverTests to set *LerpTime to 0.0 so we can make test run much faster

## Changes
- Fixes: #4800 

## Verification
All tests pass
SolverExample scene works as expected

>
>

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
@@ -47,9 +47,9 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
             None = 0,
 
             /// <summary>
-            /// Face the tracked transform, but always oriented up or down
+            /// Face the tracked transform
             /// </summary>
-            TrackedVertical = 1,
+            TrackedTarget = 1,
 
             /// <summary>
             /// Aligned to surface normal completely
@@ -264,8 +264,8 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
         }
 
         [SerializeField]
-        [Tooltip("How solver will orient model. None = no orieting, TrackedVertical = Face tracked target transform, but always oriented up/down, SurfaceNormal = Aligned to surface normal completely, Blended = blend between tracked transform and surface orientation")]
-        private OrientationMode orientationMode = OrientationMode.TrackedVertical;
+        [Tooltip("How solver will orient model. None = no orieting, TrackedTarget = Face tracked target transform, SurfaceNormal = Aligned to surface normal completely, Blended = blend between tracked transform and surface orientation")]
+        private OrientationMode orientationMode = OrientationMode.TrackedTarget;
 
         /// <summary>
         /// How solver will orient model. See OrientationMode enum for possible modes. When mode=Blended, use OrientationBlend property to define ratio for blending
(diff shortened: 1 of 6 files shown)


Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -264,11 +264,11 @@ public RaycastDirectionMode CurrentRaycastDirectionMode
         }
 
         [SerializeField]
-        [Tooltip("Orientation mode. None = no orienting, Vertical = Face head, but always oriented up/down, Full = Aligned to surface normal completely")]
-        private OrientationMode orientationMode = OrientationMode.Vertical;
+        [Tooltip("How solver will orient model. None = no orieting, TrackedTarget = Face tracked target transform, SurfaceNormal = Aligned to surface normal completely, Blended = blend between tracked transform and surface orientation")]
```
wiwei: >orieting [](start = 58, length = 8)

Nit: orienting

wiwei: err orientation

---
In reply to: [311316917](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5532#discussion_r311316917) [](ancestors = 311316917)

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -47,19 +47,19 @@ public enum OrientationMode
             None = 0,
 
             /// <summary>
-            /// Face the tracked transform, but always oriented up or down
+            /// Face the tracked transform
             /// </summary>
-            Vertical,
+            TrackedTarget = 1,
```
wiwei: >TrackedTarget [](start = 12, length = 13)

Can you also add a note to the RC2->GA migration update about these renames?

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
```diff
@@ -47,19 +47,19 @@ public enum OrientationMode
...
 
             /// <summary>
             /// Aligned to surface normal completely
             /// </summary>
-            Full,
+            SurfaceNormal = 2,
```
wiwei: >SurfaceNormal [](start = 12, length = 13)

Would it be hard to add a test case for one of these (just so that we can get some coverage for these options) - maybe the blended case because that could cover most ground?

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
keveleigh:
```suggestion
        [Tooltip("How solver will orient model. None = no orienting, TrackedTarget = Face tracked target transform, SurfaceNormal = Aligned to surface normal completely, Blended = blend between tracked transform and surface orientation")]
```

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SurfaceMagnetism.cs
Troy-Ferrell: I added TrackedTarget & SurfaceNormal into the original wall test for SurfaceMagnetism

I also changed the lerptime properties so we can test faster for the solvers

Issue #5476, opened 2019-07-31
MRTK locks Unity Editor playmode if no MRTK profile is set in a different scene


fcagnetta:
## The bug
In Editor: if you load a scene without MRTK and try to play it, but you have another scene in your project (not loaded) with MRTK object without an active profile set, you CANNOT enter Playmode.

## To reproduce

1. Create a project with 2 scenes
2. Place MRTK in one scene, with no active profile set
3. Load the other scene
4. Play

## Expected behavior
Editor normally play the scene

## MySetup
- Unity Version [2019.1.12f1]
- MRTK Version [v2.0.0 RC2.1]

## Target Platform
- UWP
- iOS
- Android

PR #5533, opened 2019-08-06
Avoid blocking developer when ActiveProfile null in Editor

Troy-Ferrell:
## Overview
Currently, if the ActiveProfile field is null for the active MRTK instance, then the MRTK system will frustratingly do two things at the editor level:

1) Block the user from entering play mode ever 
2) Spam user to select a profile whenever the MRTK object is selected in the inspector

This change addresses the issue by warning the user but allowing them to continue unimpeded. 

- The entering playmode is only aborted if the developer selects "OK" with the new dialog warning
---> If user selects Ignore, then warning dialog will never be shown again for this Unity session
- The profile object picker is only shown if the user clicks the "Assign MRTK profile" button or they click "Add mrtk to scene & configure". 
- If ActiveProfile is null, then an error message is rendered in the inspector

## Changes
- Fixes: #5476 

## Verification
Test MRTK object without Active Profile & enter play mode
Test adding MRTK gameobject to scene repeatedly
Test "Assign MRTK profile" button

>
>

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs
@@ -14,6 +14,9 @@ namespace Microsoft.MixedReality.Toolkit.Editor
         private int currentPickerWindow = -1;
         private bool checkChange = false;
 
+        // Utility to show object picker for ActiveProfile property since Show command must be called in OnGUI()
+        private static bool forceShowProfilePicker = false;
+
         private void OnEnable()
         {
             activeProfile = serializedObject.FindProperty("activeProfile");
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs
```diff
@@ -51,41 +54,45 @@ public override void OnInspectorGUI()
...
-                    activeProfile.objectReferenceValue = allConfigProfiles[0];
-                    changed = true;
-                    Selection.activeObject = allConfigProfiles[0];
-                    EditorGUIUtility.PingObject(allConfigProfiles[0]);
-                }
-                else
-                {
-                    if (EditorUtility.DisplayDialog("Attention!", "No profiles were found for the Mixed Reality Toolkit.\n\n" +
-                                                                  "Would you like to create one now?", "OK", "Later"))
+                    if (allConfigProfiles.Length > 1)
```
wiwei: >> 1) [](start = 49, length = 4)

Should this be > 0?

Note that this section originally looked for 1 because in that case, we shouldn't prompt (we should just auto-select, since that's the only thing the user can do anyway).

Since we now include the HL2 and HL1 profiles (in addition to the default one) this probably won't get exercised, but changing it to check for empty is probably the more correct thing

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs
```diff
@@ -14,6 +14,9 @@ public class MixedRealityToolkitInspector : UnityEditor.Editor
         private int currentPickerWindow = -1;
         private bool checkChange = false;
```
wiwei: >checkChange [](start = 21, length = 11)

I think that this is no longer used now that it's been deleted in the profile stuff down below - note that it's still SET in OnEnable but not referenced after that

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs
Troy-Ferrell: Yes it should just check for empty

Assets/MixedRealityToolkit/Inspectors/MixedRealityToolkitInspector.cs
Troy-Ferrell: The checkChange or picker window? They both are still used I believe

wiwei: checkChange specifically - I only see it getting updated in OnEnable (but it's never used in an if statement anymore)

Troy-Ferrell: Ahhh checkChange, I was looking at changed. Yes I’ll update

PR #5535, opened 2019-08-07
Mrtk development

rloughrey:
## Changes
- Fixes: # .


wiwei:
@rloughrey Moving things into mrtk_release from mrtk_development is a process where we first branch for stabilization (see the prerelease/2.0.0_stabilization), fix bugs in there and stabilize, and then move things into mrtk_release afterward. This gives us the ability to ensure that we can catch obvious/showstopper bugs in that branch (while allowing for continued development in mrtk_development).

I'm going to close this down because we're currently in the process of doing that stabilization - let us know if there's a specific thing you're trying to get into mrtk_release

Issue #4528, opened 2019-05-22
PressableButton gets stuck in touched/pressed state if disabled while touched/pressed


mrbobbybobberson:
## Overview

In my project, we have buttons that, when pressed, end up causing themselves to be disabled.  For example, we have a TextMeshPro explaining what kind of test the user is about to go through, and there's a "Begin" button under it.  When the user presses "Begin", the parent game object of all that UX gets set inactive to hide it, and the user goes through the testing process.

Later, we reenable the UX to give the user a break and explain what the next test will be.  When the UX is reenabled, the "Begin" button still has stale touchPoints in it from when it got disabled, so it thinks it's still currently touched and pressed.

If the user has kept his/her hand up the whole time and interacts with the button again, the button can recover, since it sees the new touches as updates to the old one, but if the user has lowered and raised his/her hand, MRTK and the button will consider the new touches as though they're coming from a unique controller, and the old touch will be stuck in the button forever, so it will stay in the pressed state forever, and can't trigger its pressed events again.

I worked around the issue by adding the following code to PressableButton, but I didn't test it exhaustively, and I don't know if it's the right fix given the bigger picture of MRTK input:

```
protected void OnDisable()
{
	touchPoints.Clear();
	IsTouching = false;
	isPressing = false;
}
```

## Expected Behavior

I would expect PressableButton to come back without any touch state when it's reenabled.

## Actual Behavior

PressableButton gets stuck in the touched/pressed state and can't be pressed again.

## Steps to reproduce

1. Make a PressableButton in a scene.
2. Hook it up so that when you press the pressable button, it or its parent game object gets set inactive thereby hiding the button.
3. Press the button with your finger (ie: direct manipulation touch)
4. While the button is hidden, lower your hand.
5. Do something to reenable the button (like make it automatically reenable after a period of time or use a Bluetooth mouse or keyboard or voice command to set the game object active again. [BUG] If you have logging or a debugger attached, you'll see that it's in the touched/pressed state when it comes back, even though nothing is touching it.
6. Now raise your hand and try to touch the button again.  [BUG] It won't be pressable anymore, because it's already stuck in the pressed state.

## Unity Editor Version

Same steps as above, but use the editor simulation hand.  Remember "raising" and "lowering" your editor hand is done by pressing or releasing the right keyboard button (e.g. the Shift button).

## Mixed Reality Toolkit Release Version

v2.0.0-RC1-Refresh

sostel:
[...]

PR #5536, opened 2019-08-07
fixed button highlights for buttons that use themes on leaving button sideways and disabling button

thalbern:
-added forcethemeupdate in interactable OnEnable to ensure the right theme is used on reenabling an interactable

-added missing physicaltouch set in handPressCompleted of physicalPressEventRouter
- rearranged if in untouched event in event router so it's behaving like the other events in that class

## Changes
- Fixes: #4528

## Verification
- run tests
- todo: device tests


thalbern:
just fyi, tested on hololens 2 - all good :)

PR #5541, opened 2019-08-07
Remove unused events, interfaces, and patterns from GGVPointer

keveleigh:
## Overview
This is largely a formatting / cleanup pass on GGVPointer, found as a result of my work on #5528.
Mostly, there was an interface that had entirely empty events. It also had a `Start` method that only called the base.

## Verification
1. Make sure GGVPointer still works.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
@@ -1,14 +1,21 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using UnityEngine;
-using System.Collections;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
```diff
@@ -9,24 +9,26 @@
 
 namespace Microsoft.MixedReality.Toolkit.Input
 {
-    public class GGVPointer : InputSystemGlobalHandlerListener, IMixedRealityPointer, IMixedRealityInputHandler, IMixedRealityInputHandler<MixedRealityPose>, IMixedRealitySourcePoseHandler, IMixedRealitySourceStateHandler
+    public class GGVPointer : InputSystemGlobalHandlerListener,
```
Troy-Ferrell: summary?

keveleigh: Good catch!

PR #5548, opened 2019-08-08
SolverHandlerInspector Fix

Alexees:
## Overview
The current implementation does work on a scene object, but it fails when editing a prefab. I am not sure if that's a bug, at least one of my Unity bug reports considered something similar as not being a bug, but this way the value stays.

## Verification
Without the fix, creating a prefab with the SolverHandler script on it will loose the chosen TrackedTargetType every time you reselect it (deselect -> select).
With this fix, the value is remembered.


Troy-Ferrell:
This was fixed in #5514 PR which got merged to the stabilization branch. The proper fix is actually to use EditorGUI.BeginProperty/EndProperty to get all of the prefab tracking. It failed for me too otherwise. 

Once stabilization branch merges back to development again it should be there

wiwei:
@Troy-Ferrell thanks for calling this out! We currently are merging stabilization back into dev, which should have this fix at that point.

PR #5554, opened 2019-08-08
[BugFix] Modified TwoHandMoveLogic to have parity with TwoHandRotateLogic in that the MovementConstraintType can change during runtime

### LINKED ISSUE #5546: ManipulationHandler's TwoHandMoveLogic doesn't properly adjust to the ConstraintOnMovement public field changing (indicating intent to change the type of movement)
## Describe the bug

Currently, if the user is trying to trigger far-field manipulations using the gaze pointer + ManipulationHandler and have defined the ConstraintOnMovement type as not being "FixDistanceFromHead" in the scene, the move logic's update function causes NaN errors to pop up. 

This is due to line 63 in TwoHandMoveLogic; which tries to recalculate the distance ratio based on the current hand distance (which is 0 since this is done by a gaze-driven pointer).

 A simple workaround to this issue would be to allow for dynamic setting of the ConstraintOnMovement variable that the TwoHandMovementLogic would respect when its Update function is called; much like it is for rotation constraint when the TwoHandRotationLogic Update function is called.

## To reproduce

1. Create an object with a ManipulationHandler object, where the constraint on movement option is anything but the "FixDistanceFromHead" type. 
2. Stare at the object, and using a voice command, spoof some pointer events (Down and Dragged) and utilize the gaze pointer as the pointer for those events to respect.
3.  See resulting NaN error

## Expected behavior

The ManipulationHandler

## Screenshots

ManipulationHelper settings the MoveConstraint type on Awake():
image

TwoHandRotateLogic function parameters:
image

TwoHandMoveLogics function parameters:
image

- Unity Version 2018.4.4f1
- MRTK Version 2.0.0-20190730.2

- HoloLens
- HoloLens 2

## Additional context

Spoke to @keveleigh about this, and since it's a small change and I know what needs to be done, I'll be making the changes for it.

Kjakubzak:
## Overview
Modified the TwoHandMoveLogic class to more closely match the TwoHandRotateLogic class in terms of passing in constraints to the Update function (allowing for a more flexible ManipulationHandler). 

This is needed so that the movement constraint type can be dynamically changed on the ManipulationHelper during runtime, so consumers of the SDK don't need to have several ManipulationHandler components on the same object for different movement constraint types. 

## Changes
- Removed the constructor for TwoHandMoveLogic that passed in and defined the MovementConstraintType for the lifespan of the TwoHandMoveLogic component
- Modified the Update function to require a MovementConstraintType object to be passed in whenever called so it allows developers to use the same component for multiple types of movement constraints (though there are only two right now).
- Fixes #5546

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/Input/Handlers/ManipulationHandler.cs
@@ -249,7 +249,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
         private void Awake()
         {
-            moveLogic = new TwoHandMoveLogic(constraintOnMovement);
+            moveLogic = new TwoHandMoveLogic();
             rotateLogic = new TwoHandRotateLogic();
             scaleLogic = new TwoHandScaleLogic();
         }
@@ -592,7 +592,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
             if ((currentState & State.Moving) > 0)
             {
                 MixedRealityPose pose = GetAveragePointerPose();
-                targetPosition = moveLogic.Update(pose, targetRotationTwoHands, targetScale, IsNearManipulation(), true);
+                targetPosition = moveLogic.Update(pose, targetRotationTwoHands, targetScale, IsNearManipulation(), true, constraintOnMovement);
             }
 
             float lerpAmount = GetLerpAmount();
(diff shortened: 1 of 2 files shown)


julenka:
It looks like mrtk_docs job is failing with the following warning:

[19-08-08 09:20:41.492]Warning:[MetadataCommand.ExtractMetadata]Duplicate parameter 'objectScale' found in comments, the latter one is ignored.

Could you please fix? [Detailed output here](https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=4058)

davidkline-ms:
Plan on merging into dev and cherry-picking into stabilization

wiwei:
Looks great!

Issue #5553, opened 2019-08-08
Interactable throws error for onFocus event with selected toggle buttonmode


Troy-Ferrell:
## Describe the bug

Interactable will throw an assertion fail for all OnFocusEnter calls if the OnBeforeFocusChange fails with CanInteract(). 

OnBeforeFocusChange is called which returns early because CanInteract() returns false because of some logic I don't understand fully. Since OnBeforeFocusChange returns early, the focusingPointers is never set properly and thus the following event call to OnFocusEnter right after, throws an error

Inside Interactable.cs
```csharp
       public void OnFocusEnter(FocusEventData eventData)
        {
            Debug.Assert(focusingPointers.Count > 0,
                "OnFocusEnter called but focusingPointers == 0. Most likely caused by the presence of a child object " +
                "that is handling IMixedRealityFocusChangedHandler");
```

Thanks to @KevinKennedy for discovery!

## To reproduce

This seems to primarily occur with Toggle interactable buttons, most likely using the InteractableToggleCollection? 

For a good repo, do the following

1) Open the SolverExample scene (at least in stabilization branch)
2) Press space to get input simulated hand
3) Hover over the "RadialView" interactable button with it already selected
4) See error message in console

## Expected behavior

Users should be able to focus on a selected toggle interactable without errors in console

## Your Setup (please complete the following information)

- Unity Version 2018.4.5
- MRTK Version Stabilization branch 2.0.0

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

julenka:
Thanks for the great repro steps!

PR #5558, opened 2019-08-08
Fix Debug.Assert when hovering over radial set buttons

julenka:
## Overview
You should be able to hover over radial buttons without getting debug asserts. I fixed this bug, added a test to verify the fix, fixed another issue highlighted by the test, and fixed a third logic issue I discovered just by reading the code.

## Changes
- Fixes: #5553
- Add test to validate fix (verified that test does not pass without fix, and passes with fix)
- Test uncovered issue: IsToggled returns incorrect value when radial button instantiated. Fixed.
- Fixed another logic issue: bitwise AND was being used instead of boolean AND.

## Verification
* Tested in editor
* Playmode tests pass.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
@@ -5,7 +5,6 @@ using Microsoft.MixedReality.Toolkit.Input;
 using Microsoft.MixedReality.Toolkit.Utilities;
 using System.Collections;
 using System.Collections.Generic;
-using UnityEditor;
 using UnityEngine;
 using UnityEngine.Events;
 
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
@@ -5,6 +5,7 @@ using Microsoft.MixedReality.Toolkit.Input;
 using Microsoft.MixedReality.Toolkit.Utilities;
 using System.Collections;
 using System.Collections.Generic;
+using UnityEditor;
 using UnityEngine;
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -1198,7 +1197,7 @@ protected virtual bool CanInteract()
                 return false;
             }
 
-            if (Dimensions > 1 && ((dimensionIndex != Dimensions - 1 & !CanSelect) || (dimensionIndex == Dimensions - 1 & !CanDeselect)))
+            if (Dimensions > 1 && ((dimensionIndex != Dimensions - 1 && !CanSelect) || (dimensionIndex == Dimensions - 1 && !CanDeselect)))
```
wiwei: Oh myyyyyyyy

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -940,11 +939,11 @@ public void OnBeforeFocusChange(FocusEventData eventData)
 
         public void OnFocusEnter(FocusEventData eventData)
         {
-            Debug.Assert(focusingPointers.Count > 0,
```
wiwei: When I updated this assert before, it was specifically happening in cases where this script wasn't really being used "correctly" - i.e. this script assumes that it gets the full range of focus events (i.e. IMixedRealityFocusChangedHandler and IMixedRealityFocusHandler.

What this assert means is that, there's something else in the chain that is eating one of the interface events (or it just isn't getting the other set of events but is getting the others). It's possible though that I had misunderstood the calling conventions/requirements here, in which case it would probably be worth removing the assert if that's the case.

julenka:
Argh wrong base branch! This is going into GA, moving there and kicking off a test.

PR #5559, opened 2019-08-09
Update the holographic remoting workaround is only limited to editor scenarios.

wiwei:
Running this code in the context of the deployed device will trigger an error (in the console) about failing to load a perception remoting binary. It turns out that calling https://docs.unity3d.com/ScriptReference/XR.WSA.HolographicRemoting.ConnectionState.html (which we do) will trigger this load attempt. Wrapping this in a UNITY_EDITOR ifdef to prevent us from trying to do this on device (where this workaround isn't necessary)


thalbern:
I like the branch name 😄

gilbdev:
sadly didn't get its way to v2.0.0

Issue #5557, opened 2019-08-08
Hand input broken in stabilization branch


sostel:
## Describe the bug
Hand tracking is not working. Hand rays are not shown. I can't poke or grab holograms. I can't use air-tap.

## To reproduce
1. I've synced to latest in the prerelease/2.0.0_stabilization branch.
2. I've moved the HandInteractionExamples.unity to the first place in the BuildSettings to build and test that scene. 
3. Build and deploy to the device.
4. No hand input is registered. I do see an little opaque gaze cursor that is following my eye gaze. 

I did not change the MRTK config profile. When checking it's set to 'DefaultHoloLens2ConfigurationProfile'. 

## Expected behavior
1. The gaze cursor should not follow my eye gaze. If that's desired, let's replace the visuals as the distinct opaque circle is not recommended.
2. Holograms should be interactable using hand input (near and far). 

## Video
![image](https://user-images.githubusercontent.com/32180789/62742268-c3d12a80-b9f2-11e9-8046-4a033f438a0c.png)

## Your Setup (please complete the following information)
- Unity  2018.4.0f1
- MRTK Version: prerelease/2.0.0_stabilization

## Target Platform (please complete the following information)
- HoloLens 2

wiwei:
Also interesting, holographic remoting also not working in that branch

keveleigh:
@davidkline-ms and I believe we've figured this out. He'll have a PR incoming.

davidkline-ms:
This looks to have been a regression introduced by the recent data provider changes.  The root causes were:

* Enable in the new base class was calling Reset on the data providers instead of Enable
* The Input and Spatial AwarenessSystems were either
    * Not calling base.Enable / base.Disable
    * Calling them at the wrong time.

PR coming soon

davidkline-ms:
fixed in stabilization. merging fix into dev shortly

PR #5561, opened 2019-08-09
Fix 5557 - Hand / controller input broken

davidkline-ms:
Recently, a regression was introduced that caused data providers to not be enabled. Windows Mixed Reality controllers (including hands) rely on Enable to register for platform events.

This PR fixes this and a related error in the new core service base class.

Added calls to base class methods in input and spatial awareness systems.

Fixes: #5557

Verified running on Windows Mixed Reality immersive device.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs
@@ -245,12 +245,16 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 }
             }
 
+            // Ensure data providers are enabled (performed by the base class)
+            base.Enable();
+
             InputEnabled?.Invoke();
         }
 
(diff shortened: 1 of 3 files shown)


julenka:
I think Kevin also wanted this to be verified on a HoloLens 2 device, is it possible to verify that as well?

keveleigh:
@julenka @KevinKennedy

I just validated both spatial mapping and hand input on HoloLens 2, and both worked great!

wiwei:
Is there any way we can have a test to ensure that the right providers are getting enabled along with this PR? I want to make sure we have SOME coverage to ensure that this doesn't recur.

davidkline-ms:
> Is there any way we can have a test to ensure that the right providers are getting enabled along with this PR? I want to make sure we have SOME coverage to ensure that this doesn't recur.

We could potentially write a test-only provider / profile that confirms that the correct method calls (Initialize, Enable, etc) are called. I'll file an issue to track

Issue #5063, opened 2019-06-26
Add an outline in the MRTK shader


benjamincoven:
## Describe the problem

Hello, in many of our projects with have used an outline effect (to higlight objects) like these ones:

https://assetstore.unity.com/packages/vfx/shaders/fullscreen-camera-effects/outline-effect-78608
https://assetstore.unity.com/packages/tools/particles-effects/quick-outline-115488

However passing to MRTK 2 and single pass rendering broke these assets.
We are curently looking for an alternative but because MRTK2 shaders are recommended I am suggesting this feature.

## Describe the solution you'd like

The possibility in one of the MRTK shader to create a colored outline.

## Describe alternatives you've considered

Using an external asset but the shaders are not optimized for HL2 and are not working in Single Pass rendering.

note: the rim light effect does not meet the needs

Cameron-Micka:
Hi @benjamincoven, I think this is a great suggestion for the MRTK shaders! I've built a few one-off prototypes for this type of effect. It has a few asset caveats at the moment. But, I'll see how best I can resolve them and merge them into the MRTK in the future.
 
Here is an example utilizing a modified version of the MRTK/Standard shader:
![Outline](https://user-images.githubusercontent.com/13305729/60206447-6ac58280-9808-11e9-8525-357e236f8139.gif)

benjamincoven:
You are awesome, we would be very happy to test it as early as possible before the next MRTK release.

Also note that we had many problems with the previous outlines that we used in the asset store about the normal of the back faces (same as #4946 ), we are using the new MRTK standard shader of the last week.

Cameron-Micka:
Sounds great, I'll be sure to post back here when I have a pull request for this feature.

Cameron-Micka:
Hi @benjamincoven, a pull request is up for mesh outlines if you want to check it out: https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5562

benjamincoven:
Hi, thank you very much for your work.

Unfortunetly I have some issues with my meshes that are one sided. I tried to set the Cull mode to off on the outline material, it's slightly better but not totally smooth.

Also it looks like that the outline is not renderer when there is another object at the back.

Maybe I missed something when configuring my test?

Here some screenshots:

![image](https://user-images.githubusercontent.com/12698212/62771064-9e98f680-ba9c-11e9-9ea3-0a37bb97e02d.png)

[...]

PR #5562, opened 2019-08-09
Mesh Outlining

Cameron-Micka:
## Overview
This pull request introduces a technique to outline meshes without utilizing post processing effects, which can be costly on mobile mixed reality devices. Many mixed reality applications require a mesh, or portions of a mesh, to be highlighted. The MeshOutline and MeshOutlineHierarchy components can be used to easy achieve this.

Outline

To outline a mesh the object is drawn a second time slightly extruded. This results in extra draw calls but reduced overdraw. Arbitrary meshes are supported. Meshes which do not extrude cleanly can be mitigated by utilizing the MeshSmoother component which asynchronously generates smoothing groups for meshes. Outline visuals are specified via a second material which gives developers freedom to pick a wide array effects.

Outline

Because this effect happens in world space rather than screen space a few limitations apply around sorting and non-uniform scales.

The PR includes:

- [x] Example Scene
- [x] Unit Tests
- [x] Documentation

Lunar model curtesy of [Beau Folsom](https://www.linkedin.com/in/beaufolsom/). 

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5063

## Verification
Please try using the OutlineExamples scene.

>
>


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.MixedReality.Toolkit.Input;
+using UnityEngine;
+using UnityEngine.Events;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// Utility component to hook up Unity events to the OnFocusEnter and OnFocusExit events.
+    /// </summary>
+    public class FocusHandlerTrigger : MonoBehaviour, IMixedRealityFocusHandler
```
Troy-Ferrell: This seems like something that should be joined with Features/Input/Handlers
And of course renamed to FocusHandler

Assets/MixedRealityToolkit.Examples/Demos/StandardShader/Textures/Lander_albedo.png.meta
```diff
@@ -0,0 +1,88 @@
+fileFormatVersion: 2
+guid: 11056bb52f91cdc43ba4fba9accfb226
```
Troy-Ferrell: How strongly do you feel about adding the lander model? The problem with adding these FBX assets into Examples it they bloat the UnityPackage file. The textures and mesh data adds like 7-8MB alone in this change. I know that is typical for assets but that's a 15% increase to the existing package. 

I recommended removing because of the size and primarily because it's the feature is not dependent on it. We can still demonstrate the feature with the moon lander rover or other existing FBX

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        /// <summary>
+        /// Event which is triggered when focus ends.
+        /// </summary>
+        public UnityEvent OnFocusExitEvent;
+
+        /// <inheritdoc />
+        public void OnFocusEnter(FocusEventData eventData)
+        {
+            OnFocusEnterEvent.Invoke();
```
Troy-Ferrell: I think may want to check for OnFocusEnterEvent?.Invoke()?

Assets/MixedRealityToolkit.Tests/PlayModeTests/StandardShaderTests.cs
```diff
@@ -0,0 +1,129 @@
...
+        private readonly Vector3 LightDirection = new Vector3(-0.5f, -0.5f, 0.0f);
+
+        /// <summary>
+        /// Tests if the MeshOutline component can be added an manipulated at runtime.
+        /// </summary>
+        /// <returns></returns>
+        [UnityTest]
+        public IEnumerator TestMeshOutline()
+        {
+            var light = InstantiateDirectionalLight(LightDirection);
```
Troy-Ferrell: I don't understand what this is testing? Just no errors thrown? There are no asserts

Assets/MixedRealityToolkit/Utilities/StandardShader/BaseMeshOutline.cs
```diff
@@ -0,0 +1,72 @@
...
+{
+    /// <summary>
+    /// Abstract component to encapsulate common functionality around outline components. 
+    /// </summary>
+    [HelpURL("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/README_MRTKStandardShader.html#mesh-outlines")]
+    public abstract class BaseMeshOutline : MonoBehaviour
+    {
+        /// <summary>
+        /// The material used to render the outline. Outline materials should normal have "Depth Write" set to Off and "Vertex Extrusion" enabled.
+        /// Most MRTK/Standard features should work as an outline material, but it is recommended to use a simple unlit material.
```
Troy-Ferrell: So I'm still trying to parse this whole PR. But is the outline feature in the scripts or in the MRTK standard shader?

All of the Outline*.mat files in this PR are MRTK standard shader. Why not make them simpler as just unlit as this comment suggests?

Assets/MixedRealityToolkit/Utilities/StandardShader/BaseMeshOutline.cs
```diff
@@ -0,0 +1,72 @@
...
+        protected float outlineWidth = 0.01f;
+
+        #region MonoBehaviour Implementation
+
+        /// <summary>
+        /// Enables users to modify inspector properties while playing in the editor.
+        /// </summary>
+        protected virtual void OnValidate()
+        {
+            ApplyOutlineMaterial();
```
Troy-Ferrell: Why use onvalidate instead of a BaseMesh inspector class? This is for the component right?

Assets/MixedRealityToolkit/Utilities/StandardShader/MeshSmoother.cs
```diff
@@ -0,0 +1,241 @@
...
+    /// Component which can be used to automatically generate smoothed normals on a mesh and pack 
+    /// those normals into a UV set. Smoothed normals can be used for a variety of effects including 
+    /// extruding disjoint meshes along a vertex normal. This behavior is designed to be used in conjunction 
+    /// with the MRTK/Standard shader which assumes smoothed normals are packed into the 3rd UV set.
+    /// </summary>
+    [RequireComponent(typeof(MeshFilter))]
+    [HelpURL("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/README_MRTKStandardShader.html#mesh-outlines")]
+    public class MeshSmoother : MonoBehaviour
+    {
+        private const int smoothNormalUVChannel = 2;
```
Troy-Ferrell: This seems more like an editor level utility? What is the benefit of running this at runtime as a MonoBehavior?

Documentation/README_MRTKStandardShader.md
```diff
@@ -105,18 +107,55 @@ Performant plane, sphere, and box shape clipping with the ability to specify whi
 
 ![primitive clipping gizmos](../Documentation/Images/MRTKStandardShader/MRTK_PrimitiveClippingGizmos.gif)
 
+### Mesh Outlines
+
+Many mesh outline techniques are done using a [post processing](https://docs.unity3d.com/Manual/PostProcessingOverview.html) technique. Post processing provides great quality outlines, but can be prohibitively expensive on mobile Mixed Reality devices. You can find a scene that demonstrates usage of mesh outlines in the  **OutlineExamples** scene under: [MixedRealityToolkit.Examples/Demos/StandardShader/Scenes/](https://github.com/microsoft/MixedRealityToolkit-Unity/tree/mrtk_release/Assets/MixedRealityToolkit.Examples/Demos/StandardShader/Scenes)
```
Troy-Ferrell: nit: expensive on mobile Mixed Reality devices >>>> Expensive on many Mixed Reality platforms. 
Just a better catch all

Documentation/README_MRTKStandardShader.md
```diff
@@ -105,18 +107,55 @@ Performant plane, sphere, and box shape clipping with the ability to specify whi
...
+
+<img src="../Documentation/Images/MRTKStandardShader/MRTK_MeshOutline.jpg" width="900">
+
+[**MeshOutline.cs**](xref:Microsoft.MixedReality.Toolkit.Utilities.MeshOutline) and [**MeshOutlineHierarchy.cs**](xref:Microsoft.MixedReality.Toolkit.Utilities.MeshOutlineHierarchy) can be used render an outline around a mesh renderer. Enabling this component introduces an additional render pass of the object being outlined, but is designed to run performantly on mobile Mixed Reality devices and does not utilize any post processes. Limitations of this effect include it not working well on objects which are non-uniformly scaled and depth sorting issues on overlapping objects.
```
Troy-Ferrell: Can you actually switch these from [**Blah**] to [ \` blah \`  ]  for styling?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
```diff
@@ -0,0 +1,38 @@
...
+using UnityEngine.Events;
+
+namespace Microsoft.MixedReality.Toolkit.UI
+{
+    /// <summary>
+    /// Utility component to hook up Unity events to the OnFocusEnter and OnFocusExit events.
+    /// </summary>
+    public class FocusHandlerTrigger : MonoBehaviour, IMixedRealityFocusHandler
+    {
+        /// <summary>
```
julenka: Do these need to be initialized? Also, I believe MRTK aims to use the pattern internal private serialized fields, public getters and setters.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
julenka: We tend to use the name "FocusHandler" instead of *Trigger, for example see "InputActionHandler"

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Input;
+using UnityEngine;
+using UnityEngine.Events;
+
+namespace Microsoft.MixedReality.Toolkit.UI
```
julenka: might want to move this input SDK/Features/Input/Handlers since the other event handlers are there.

Assets/MixedRealityToolkit/Utilities/StandardShader/BaseMeshOutline.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        [Tooltip("How thick (in meters) should the outline be. Overrides the \"Extrusion Value\" in the MRTK/Standard material.")]
+        [SerializeField]
+        [Range(0.001f, 1.0f)]
+        protected float outlineWidth = 0.01f;
+
+        #region MonoBehaviour Implementation
+
+        /// <summary>
```
julenka: I remember @Railboy recently removed a bunch of code that was using OnValidate incorrectly #4899. Wanted to double check that this was appropriate usage. It seems like we are actually using OnValidate here as a proxy to "when value changes, show changes in editor". Is there another way to do this, maybe with the inspector?

Assets/MixedRealityToolkit/Utilities/StandardShader/MeshSmoother.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        [Tooltip("Should this component automatically smooth normals on awake?")]
+        [SerializeField]
+        private bool smoothNormalsOnAwake = false;
```
julenka: should this have a getter / setter?

Assets/MixedRealityToolkit.Tests/PlayModeTests/StandardShaderTests.cs
julenka: Yeah I assume he is just checking that no null ptrs are thrown. I considered suggesting doing screenshot comparison but that seemed like it could break often and also could be error prone.

Assets/MixedRealityToolkit.Tests/PlayModeTests/StandardShaderTests.cs
```diff
@@ -0,0 +1,129 @@
...
+            var gameObject = GameObject.CreatePrimitive(PrimitiveType.Cube);
+
+            var meshOutline = gameObject.AddComponent<MeshOutline>();
+            meshOutline.OutlineMaterial = InstantiateStandardShaderOutlineMaterial(Color.red);
+            meshOutline.OutlineWidth = 0.02f;
```
julenka: idea: could you:

1. make the outline some color like bright green (0, 255, 0)
2. Capture camera texture
3. ensure that at least 100 pixels are bright green.

screen shot / capture camera texture and at least ensure that some

julenka:
@Cameron-Micka let me know when you'd like me to take another look. This PR looks awesome! Thank you.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
Cameron-Micka: That's a good idea, I'll make that change. I started by using the Interactable behavior (but it makes some assumptions that don't always work and it's pretty heavy for this use) so that's why I put this here.

Assets/MixedRealityToolkit.Examples/Demos/StandardShader/Textures/Lander_albedo.png.meta
Cameron-Micka: I would really like to add this model since it demonstrates proper asset consturction, naming, and PBR workflow. It would be nice to phase out the old lander (and maybe even the rover) since they don't demonstrate the same quality bar I would like to see in the MRTK.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
Cameron-Micka: We shouldn't need to since Unity's serialization system will initialize all serializable properties for us. But, I suppose a user could nullify the event? I can make the change.

Assets/MixedRealityToolkit.Tests/PlayModeTests/StandardShaderTests.cs
Cameron-Micka: Yep @julenka is correct. We've had so many components that just fall over when instantiated at runtime with null references that this is a sanity check for that. Would love to do screenshot comparisons. But, that would require some work on the test and build machine side. :(

Assets/MixedRealityToolkit/Utilities/StandardShader/BaseMeshOutline.cs
Cameron-Micka: Ah, this comment is misleading. Thanks for pointing that out, I'll fix this! It should read that you should use a MRTK/Standard material but it's recommend to make the MRTK/Standard material as bare bones as possible (in other words unlit).

Assets/MixedRealityToolkit/Utilities/StandardShader/BaseMeshOutline.cs
Cameron-Micka: I'm just using OnValidate to apply changes as they occur in the editor. I could make a customer inspector but seems a little overkill?

Assets/MixedRealityToolkit/Utilities/StandardShader/MeshSmoother.cs
Cameron-Micka: We need to do this at runtime for meshes that are loaded dynamically. Such as loading user generated content at runtime.

Documentation/README_MRTKStandardShader.md
Cameron-Micka: Good idea!

Documentation/README_MRTKStandardShader.md
Cameron-Micka: Certainly!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
Cameron-Micka: They shouldn't need to be initialized since Unity's we do that for you for any serializable properties. But, I'm happy to provide a default. Also, I do prefer the private serialized field pattern. I was just following what I found in PressableButton, BoundingBox, etc. But, we should probably fix those too? I'll make this change!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
Cameron-Micka: I like your name better, I'll change that.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/FocusHandlerTrigger.cs
Cameron-Micka: Good idea!

Assets/MixedRealityToolkit/Utilities/StandardShader/BaseMeshOutline.cs
Cameron-Micka: I've heard some of the (warranted) complaints about using OnValidate. But, I do believe this is the intended use of it. The MRTK was using it for initialization in a few places, which is a dangerous pattern. I think it's perfectly acceptable to use OnValidate to respond to edit time property changes.

Assets/MixedRealityToolkit/Utilities/StandardShader/MeshSmoother.cs
Cameron-Micka: That is a good question, since you can't modify the value in code before Awake is called I ommited them. Can you think of when we would need them for a property like this?

Assets/MixedRealityToolkit.Tests/PlayModeTests/StandardShaderTests.cs
Cameron-Micka: I like that idea! Last time I tried to take screenshots on the build machine it kept failing with "missing device context" issues. But, maybe we could think of another way to accomplish this?

Cameron-Micka:
Thank you @Troy-Ferrell & @julenka for the review! I'll make these changes as soon as I can which should (hopefully) be later next week (in the middle of another project ATM).

Issue #5549, opened 2019-08-08
WMR controller Menu Button switched


SchroerDev:
## Describe the bug

The handedness of menu button for windows mixed reality controllers is switched.

I Press the Menu button on the left controller and the InputEventData.Handedness is right.

## To reproduce

Try this class:

```C#
using Microsoft.MixedReality.Toolkit;
using Microsoft.MixedReality.Toolkit.Input;
using UnityEngine;

public class MenuButtonTest : MonoBehaviour, IMixedRealityInputHandler
{

    private IMixedRealityInputSystem inputSystem = null;

    /// <summary>
    /// The active instance of the input system.
    /// </summary>
    protected IMixedRealityInputSystem InputSystem
    {
        get
        {
            if (inputSystem == null)
            {
                MixedRealityServiceRegistry.TryGetService(out inputSystem);
            }
            return inputSystem;
        }
    }

    private void OnEnable()
    {
        InputSystem?.Register(gameObject);
    }

    private void OnDisable()
    {
        InputSystem?.Unregister(gameObject);
    }

    public void OnInputDown(InputEventData eventData)
    {
        Debug.Log("EventDown : " + eventData.MixedRealityInputAction.Description.ToString() 
+ " Handness: " + eventData.Handedness.ToString() + " sourceID: " 
+ eventData.SourceId + "  ----  " + eventData.InputSource.SourceName);
    }

    public void OnInputUp(InputEventData eventData)
    {
        Debug.Log("EventUp : " + eventData.MixedRealityInputAction.Description.ToString() 
+ " Handness: " + eventData.Handedness.ToString() + " sourceID: " 
+ eventData.SourceId + "  ----  " + eventData.InputSource.SourceName);
    }
    
}
```

## MySetup 

- Unity Version  2019.1.11f1
- MRTK Version v2.1

## Target Platform 

- WMR immersive

davidkline-ms:
i am unable to reproduce this (prerelease/2.0.0_stabilization branch) using my WMR immersive headset when running a UWP application.

I will try using the same headset running an OpenVR application

davidkline-ms:
confirmed that the mappings are for the incorrect handedness in the OpenVR controller mapping

davidkline-ms:
Fixed in stabilization. will go into mrtk_development with the next merge.

PR #5578, opened 2019-08-09
Fix button mappings for WMR controllers in OpenVR - 5549

davidkline-ms:
#5549 pointed out that the menu button mappings for the Windows Mixed Reality Motion Controllers were incorrect when running on OpenVR.

This was a result of the button identifiers being entered backwards (left as right, right as left) in the .cs and .asset files.

Verified using customer provided script, running both UWP and OpenVR targets in the editor and examining the debug console output.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityControllerMappingProfile.asset
+++ b/Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityControllerMappingProfile.asset
@@ -2189,7 +2189,7 @@ MonoBehaviour:
         id: 2
         description: Menu
         axisConstraint: 2
-      keyCode: 330
+      keyCode: 332
       axisCodeX: 
       axisCodeY: 
       invertXAxis: 0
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.SDK/Profiles/DefaultMixedRealityControllerMappingProfile.asset
```diff
@@ -2189,7 +2189,7 @@ MonoBehaviour:
         id: 2
         description: Menu
         axisConstraint: 2
-      keyCode: 330
+      keyCode: 332
```
keveleigh: There's another mapping profile as well (not sure what the difference in mappings is, if we could squash them into the default):

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/prerelease/2.0.0_stabilization/Assets/MixedRealityToolkit.Examples/Demos/Input/Scenes/InputActions/InputActions.MixedRealityControllerMappingProfile.asset#L2192

Issue #4491, opened 2019-05-21
[nuget] configuration profile UI missing default types when consuming MRTK as nuget package


naostranMS:
## Describe the bug

When consuming MRTK as a nuget package, the configuration profile inspector doesn't seem to be able to locate the default MRTK system types (e.g. MixedRealityCameraSystem) and so the dropdowns are not populated.

## To reproduce

1. Create a new project, and add MRTK as a nuget package.
2. From the MRTK menu -> Add to Scene and Configure
3. Clone the MRTKConfigurationProfile
4. Select the Camera configuration and clone it
5. Note that Camera System Type is "MixedRealityCameraSystem {Missing}"
6. Click "Try Repair" -> error "No types with the name 'MixedRealityCameraSystem' were found"
7. Try to select a type with the dropdown -> "None" is the only option

This also happens with input providers in the input system configuration (and probably others, but those are all I've tried).

## Expected behavior

Default/built-in MRTK system types are selectable through the configuration profile inspector.

## Screenshots
![image](https://user-images.githubusercontent.com/38891743/58059890-b1e29880-7b23-11e9-9ff2-f921e0877d47.png)
![image](https://user-images.githubusercontent.com/38891743/58059911-c7f05900-7b23-11e9-8614-132aac878093.png)
![image](https://user-images.githubusercontent.com/38891743/58059933-e3f3fa80-7b23-11e9-94fd-500f777f9c85.png)

## Your Setup (please complete the following information)

- Unity 2018.4.0f1
- Microsoft.MixedReality.Toolkit/2.0.0-20190517.1

## Target Platform (please complete the following information)

- Unity editor

Yoyozilla:
blocks nuget consumption

MenelvagorMilsom:
@Yoyozilla I think this is more Andrei's area of expertise. I'm going to reassign to him.

PR #5589, opened 2019-08-09
Fix missing default types in profile configs when consumed via NuGet

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4491

The current type enumeration code uses Unity's assembly CompilationPipeline.GetAssemblies, which only captures things that are going through Unity's compilation pipeline. This doesn't work well when consuming MRTK as compiled binaries - these types are obviously not part of source files at that point.

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
@@ -68,13 +68,15 @@ namespace Microsoft.MixedReality.Toolkit.Editor
         private static List<Type> GetFilteredTypes(SystemTypeAttribute filter)
         {
             var types = new List<Type>();
-            var assemblies = CompilationPipeline.GetAssemblies();
             var excludedTypes = ExcludedTypeCollectionGetter?.Invoke();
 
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Inspectors/PropertyDrawers/TypeReferencePropertyDrawer.cs
```diff
@@ -68,13 +68,15 @@ public class SystemTypeReferencePropertyDrawer : PropertyDrawer
...
-            var assemblies = CompilationPipeline.GetAssemblies();
             var excludedTypes = ExcludedTypeCollectionGetter?.Invoke();
 
+            // We prefer using this over CompilationPipeline.GetAssemblies() because
+            // some types may come from plugins and other sources that have already
+            // been compiled.
+            var assemblies = AppDomain.CurrentDomain.GetAssemblies();
             foreach (var assembly in assemblies)
             {
-                Assembly compiledAssembly = Assembly.Load(assembly.name);
```
keveleigh: No longer need to `Load`?

wiwei: Nope! Load was necessary before because UnityAssembly needed to be turned into System.Reflection.Assembly. Since these assemblies are already System.Reflection.Assembly, no need!

PR #5594, opened 2019-08-10
Introduce Experimental  Feature: Dwell Interaction

ritijain:
## Overview
This PR introduces the Dwell interaction as an Experimental component in MRTK.

## Changes
- This PR adds 2 scripts : DwellProfile and Dwellhandler as core components that a developer can leverage to setup a interactable target as a dwellable object. 
The DwellHandler script exposes UnityEvents which the targets can register to to configure their visual states based on teh dwell progress.

- Developers can also override the CalculateDwellProgress method to get creative with how they would like to leverage the dwell timer in conjunction with visuals.

## Verification
Added a sample scene and scrips to demonstrate various use cases.

>
> git fetch origin pull/{PR_ID}/head:gazedwellv1
>
> git checkout gazedwellv1


Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellProfile.cs
```diff
@@ -0,0 +1,91 @@
...
+namespace Microsoft.MixedReality.Toolkit.Experimental.Dwell
+{
+    [MixedRealityServiceProfile(typeof(DwellProfile))]
+    [CreateAssetMenu(menuName = "Mixed Reality Toolkit/Profiles/Dwell Profile", fileName = "DwellProfile", order = 100)]
+    [Serializable]
+    public class DwellProfile : ScriptableObject
+    {
+        [Tooltip("Pointer type to use for triggering a dwell interaction")]
+        [SerializeField]
+        private InputSourceType dwellTriggerPointerType = InputSourceType.Head;
```
AdamMitchell-ms: >dwellTriggerPointerType [](start = 32, length = 23)

Do you think users of this feature may ever want to allow dwells with more than one input source type?  If so, then consider making this an array.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+
+            CancelDwellAction += OnDwellCancelAction;
+        }
+
+        private void Update()
+        {
+            if (HasFocus && currentDwellState != DwellState.DwellCompleted)
+            {
+                double focusTime = (DateTime.UtcNow - this.focusEnterTime).TotalSeconds;
```
AdamMitchell-ms: >focusTime [](start = 23, length = 9)

nit: rename focusTime to focusDuration (or dwellDuration?) for clarity

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellProfile.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        [Tooltip("Additional time (not including the timeToTriggerDwellInSec) the user needs to keep looking at the UI to trigger select on it.")]
+        [SerializeField]
+        [Range(0, 20)]
+        private float timeToCompleteDwell = 6;
+
+        [Tooltip("Should the system allow for dwell to resume if the pointer exits the target briefly.")]
+        [SerializeField]
+        private bool allowDwellResume = false;
+
```
AdamMitchell-ms: This parameter is not needed.  You can just set timeToAllowDwellResume to 0.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        protected enum DwellState
+        {
+            None = 0,
+            FocusGained,
+            DwellIntended,
+            DwellStarted,
+            DwellCompleted,
+            DwellCanceled,
+            Invalid
```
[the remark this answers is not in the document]
AdamMitchell-ms: Do you mean trailing comma?

---
In reply to: [313146037](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r313146037) [](ancestors = 313146037)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+                double focusTime = (DateTime.UtcNow - this.focusEnterTime).TotalSeconds;
+
+                if (focusTime >= dwellProfile.DwellIntentDelay && currentDwellState == DwellState.FocusGained)
+                {
+                    DwellIntended.Invoke(pointerRef);
+                    currentDwellState = DwellState.DwellIntended;
+                    fillTimer = 0;
+                }
+                else if ((focusTime - dwellProfile.DwellIntentDelay) >= dwellProfile.DwellStartDelay && currentDwellState == DwellState.DwellIntended)
+                {
```
AdamMitchell-ms: super nit: I'd do the state check prior to the math in this if clause.  And consider refactoring the math slightly to add the delays right of the ">=" sign instead of subtracting some of them
(currentState == DwellIntended && (focusTime >= DwellIntendDelay + DwellStartDelay)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// Invoke this event if the dwell is interrupted with alternate modality
+        /// </summary>
+        public UnityAction CancelDwellAction;
+
+        protected bool HasFocus { get; private set; }
+
+        /// <summary>
+        /// Cached pointer reference to track focus events maps to teh same pointer that initiated dwell
+        /// </summary>
+        private IMixedRealityPointer pointerRef;
```
AdamMitchell-ms: >pointerRef [](start = 37, length = 10)

nit: rename to just "pointer"

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+            {
+                case DwellState.None:
+                    dwellProgress = 0;
+                    break;
+                case DwellState.FocusGained:
+                    dwellProgress = 0;
+                    fillTimer = 0;
+                    break;
+                case DwellState.DwellStarted:
+                    fillTimer -= Time.deltaTime;
```
AdamMitchell-ms: >Time.deltaTime [](start = 33, length = 14)

Does this imply that CalculateDwellProgress must be called exactly once per frame?  If so, then maybe do the math in Update() instead?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+                    focusEnterTime = DateTime.UtcNow;
+                    currentDwellState = DwellState.FocusGained;
+                    pointerRef = eventData.Pointer;
+                    fillTimer = 0;
+                }
+                //intent to resume 
+                else if (currentDwellState == DwellState.DwellCanceled && dwellProfile.AllowDwellResume
+                    && pointerRef.InputSourceParent.SourceId == eventData.Pointer.InputSourceParent.SourceId //make sure the returning pointer id is the same
+                    && (DateTime.UtcNow - focusExitTime).TotalSeconds <= dwellProfile.TimeToAllowDwellResume)
+                {
```
AdamMitchell-ms: It seems like you're doing a lot of this logic twice, once in the if clause and again here in reverse.  Consider refactoring to something like this:

if (resume criteria)
{
   // do resume actions here
}
else //maybe restart in all other cases?  Not sure if you need to check anything in this case.
{
   // do restart actions here
}

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellProfile.cs
ritijain: No. This is meant to support only 1 type of input source,

---
In reply to: [313498270](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r313498270) [](ancestors = 313498270)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellProfile.cs
```diff
@@ -0,0 +1,91 @@
...
+        [Tooltip("Time in seconds when gaze can fall off the target and come back.")]
+        [SerializeField]
+        [Range(0, 20)]
+        private float timeToAllowDwellResume = 1;
+
+        public InputSourceType DwellPointerType
```
AdamMitchell-ms: I think there should be a period of time after dwell completion where new dwells are not allowed on the same item.  Is this accounted for anywhere?  Or maybe it's just that as currently coded, you have to focus off then back on again to start a new dwell after completing one?  Maybe that's fine.

ritijain: Yes you have to focus of and then focus back on the target to trigger a fresh dwell.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/ToggleDwellSample.cs
```diff
@@ -0,0 +1,88 @@
...
+        private Color dwellOnColor = Color.white;
+
+        [SerializeField]
+        private Color dwellOffColor = Color.cyan;
+
+        [SerializeField]
+        private Color dwellIntendedColor = Color.cyan;
+
+        [SerializeField]
+        private float dwellVisualCancelDurationInFPS = 60;
```
AdamMitchell-ms: >dwellVisualCancelDurationInFPS [](start = 22, length = 30)

Do you mean in frames?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+
+using Microsoft.MixedReality.Toolkit.Input;
+using System;
+using UnityEngine;
+using UnityEngine.Events;
+
+namespace Microsoft.MixedReality.Toolkit.Experimental.Dwell
+{
+    [Serializable]
+    public class PointerUnityEvent : UnityEvent<IMixedRealityPointer>
```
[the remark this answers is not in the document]
ritijain: Maybe we can do that when this feature is upgraded to be part of regular namespace.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/BaseDwellSample.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
ritijain: > Should samples be in a samples project/assembly so they don't end up in the core MRTK distribution?

Not sure. @keveleigh  thoughts?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
ritijain: > > Time.deltaTime [](start = 33, length = 14)
> 
> Does this imply that CalculateDwellProgress must be called exactly once per frame? If so, then maybe do the math in Update() instead?

Not necessarily. It is up to the caller to determine how often to make this call.

Assets/MixedRealityToolkit.SDK/Inspectors/Dwell/DwellHandlerInspector.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEditor;
+
+namespace Microsoft.MixedReality.Toolkit.Experimental.Dwell.Editor
+{
+    [CustomEditor(typeof(DwellHandler), true)]
+    public class DwellHandlerInspector : UnityEditor.Editor
```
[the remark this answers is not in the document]
ritijain: > I would recommend making these private classes, nested inside the class they are editing. This way, your `FindProperty` calls can use `nameof` which will make this code much more resilient to field renaming bugs. This would look something like:
> 
> ```cs
> partial class DwellHandler
> {
>     [CustomEditor(typeof(DwellHandler), true)]
>     private class DwellHandlerInspector : UnityEditor.Editor
>     {
>         ...
>         this.serializedObject.FindProperty(nameof(DwellHandler.dwellProfile));
>         ...
>     }
> }
> ```

I have not seen that pattern prevelant in MRTK. @wiwei , @keveleigh How do you folks feel about this ?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/BaseDwellSample.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public virtual void DwellStarted(IMixedRealityPointer pointer) { isDwelling = true; }
+
+        public virtual void DwellIntended(IMixedRealityPointer pointer) { }
+
+        public virtual void DwellCanceled(IMixedRealityPointer pointer) { isDwelling = false; }
+        
+        public virtual void DwellCompleted(IMixedRealityPointer pointer) { isDwelling = false; ButtonExecute(); }
+
+        public virtual void ButtonExecute() { }
```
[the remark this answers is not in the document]
ritijain: > I personally find this `ButtonExecute` method a little confusing. For me it would be easier to understand if we just kept it as `DwellCompleted` and the derived classes just override that method. It threw me off because I thought `ButtonExecute` was going to click a button or something, rather than just drive visuals corresponding to dwell completing.

yeah ButtonExecute is implementing the "what actions do i perform when this button is clicked", not doing anything related to dwell visuals :)
ButtonExecute is what will be bound to the unity button's onclick event handler.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+        protected DwellState currentDwellState = DwellState.None;
+        protected float fillTimer = 0;
+
+        private void Awake()
+        {
+            Debug.Assert(dwellProfile != null, "DwellProfile is null, creating default profile.");
+
+            if (dwellProfile == null)
+            {
+                dwellProfile = ScriptableObject.CreateInstance("DwellProfile") as DwellProfile;
```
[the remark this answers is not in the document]
ritijain: > ```cs
> dwellProfile = ScriptableObject.CreateInstance<DwellProfile>();
> ```

That will create compile issue since the  ScriptableObject.CreateInstance returns a scriptable object.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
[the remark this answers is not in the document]
wiwei: Note that there already is a PointerUnityEvent, so when that day happens we'd still have to reconcile that.

If you want to avoid that conversation you can also go DwellUnityEvent (which names it after the situations in which it's used, rather than what it represents). It's still not GREAT exactly but could be a reasonable tradeoff

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+    /// <summary>
+    /// Use this component to add a Dwell modality (https://docs.microsoft.com/en-us/windows/mixed-reality/gaze-and-dwell) to the UI target.
+    /// </summary>
+    public class DwellHandler : MonoBehaviour, IMixedRealityFocusHandler
+    {
+        protected enum DwellStateType
+        {
+            None = 0,
+            FocusGained,
+            DwellIntended,
```
wiwei: >DwellIntended, [](start = 12, length = 14)

It might be worth adding docomments to this enum to better explain the transitions and how they work:

i.e. the state machine is:

Current State       |     Valid Transitions
None                    |  FocusGained
FocusGained       |  None
FocusGained       |  DwellIntended
DwellIntended    |  DwellCanceled
DwellIntended    |  DwellStarted
DwellCanceled   |  None

Also in particular, it would be helpful to explain the term "DwellIntended" - the others feel like they're relatively clear

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+        {
+            get
+            {
+                switch (CurrentDwellState)
+                {
+                    case DwellStateType.None:
+                    case DwellStateType.FocusGained:
+                        return 0;
+                    case DwellStateType.DwellStarted:
+                        return Mathf.Clamp((float)(dwellProfile.TimeToCompleteDwell - FillTimer) / dwellProfile.TimeToCompleteDwell, 0f, 1f);
```
wiwei: >Mathf.Clamp((float)(dwellProfile.TimeToCompleteDwell - FillTimer) / dwellProfile.TimeToCompleteDwell, 0f, 1f); [](start = 31, length = 110)

Nit: Make this a private helper (because it's used twice in the context of this getter)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+                        if ((DateTime.UtcNow - focusExitTime).TotalSeconds > dwellProfile.TimeToAllowDwellResume)
+                        {
+                            CurrentDwellState = DwellStateType.None;
+                            return 0;
+                        }
+                        else if (dwellProfile.TimeToAllowDwellResume > 0)
+                        {
+                            return Mathf.Clamp((float)(dwellProfile.TimeToCompleteDwell - FillTimer) / dwellProfile.TimeToCompleteDwell, 0f, 1f);
+                        }
+                        break;
```
wiwei: >break; [](start = 24, length = 6)

It's weird seeing no return here, given that there are returns in other locations. - Maybe add a return to be consistent, or remove the other ones

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+        [SerializeField]
+        private PointerUnityEvent DwellCompleted = new PointerUnityEvent();
+
+        [SerializeField]
+        private PointerUnityEvent DwellCanceled = new PointerUnityEvent();
+
+        /// <summary>
+        /// Invoke this event if the dwell is interrupted with alternate modality
+        /// </summary>
+        public UnityAction CancelDwellAction;
```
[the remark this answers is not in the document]
wiwei: +1

---
In reply to: [313148556](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r313148556) [](ancestors = 313148556)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellProfile.cs
```diff
@@ -0,0 +1,79 @@
...
+        [Tooltip("Pointer type to use for triggering a dwell interaction")]
+        [SerializeField]
+        private InputSourceType dwellTriggerPointerType = InputSourceType.Head;
+
+        [Tooltip("Delay in seconds until it is determined that the user intends to interact with the target.")]
+        [SerializeField]
+        [Range(0, 2)]
+        private float dwellIntentDelay = 0;
+
+        [Tooltip("Delay in seconds until dwell feedback is started to be shown.")]
```
wiwei: >feedback is started to be shown [](start = 47, length = 31)

Some of these comments are interesting because they suggest feedback/UI, when this portion of the code is designed to be agnostic of that and more about the state transitions - as such, it might be best to keep language here scoped to the data model underlying, instead of the higher level experiences that get built on top of that.

Otherwise, the natural question is "what feedback"? The answer is, there's no feedback in this code, and you have to build it yourself on top of this.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellProfile.cs
```diff
@@ -0,0 +1,79 @@
...
+        [SerializeField]
+        [Range(0, 5)]
+        private float dwellStartDelay = 3;
+
+        [Tooltip("Additional time (not including the timeToTriggerDwellInSec) the user needs to keep looking at the UI to trigger select on it.")]
```
wiwei: >timeToTriggerDwellInSec [](start = 53, length = 23)

I think that this word needs to be updated (it seems like it doesn't exist anymore here)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/BaseDwellSample.cs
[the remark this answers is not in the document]
wiwei: Yep, these should be in MixedRealityToolkit.Examples folder.

In general I would just move all of these things to the MixedRealityToolkit.Examples folder

---
In reply to: [313534236](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r313534236) [](ancestors = 313534236)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+        /// </summary>
+        protected bool HasFocus { get; private set; }
+
+        /// <summary>
+        /// Captures the dwell status 
+        /// </summary>
+        protected DwellStateType CurrentDwellState = DwellStateType.None;
+
+        /// <summary>
+        /// Abstracted value for the how long the dwelled object still needs to be focused to comlete the dwell action
```
wiwei: >long [](start = 41, length = 4)

Nit: What are the units on this, or is it unitless? (If unitless, documenting that it is unitless is helpful)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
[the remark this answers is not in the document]
ryantrem: If there is already an existing PointerUnityEvent, can we just use that? Or is it not a UnityEvent<IMixedRealityPointer>?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
wiwei: Ah, this was a nit

---
In reply to: [314409177](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r314409177) [](ancestors = 314409177)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+                    default:
+                        return 0;
+                }
+
+                return 0;
+            }
+        }
+
+        /// <summary>
+        /// Cached pointer reference to track focus events maps to teh same pointer that initiated dwell
```
wiwei: >Cached pointer reference to track focus events maps to teh same pointer that initiated dwell [](start = 12, length = 92)

Cached pointer reference to track focus events maps to the same pointer that initiated the dwell.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+        /// </summary>
+        private IMixedRealityPointer pointer;
+
+        private DateTime focusEnterTime = DateTime.MaxValue;
+        private DateTime focusExitTime = DateTime.MaxValue;
+
+        /// <summary>
+        /// Exposes whether the target has focus from the pointer defined in dwell profile settings
```
wiwei: >dwell profile settings [](start = 77, length = 22)

Is this the same as the pointer variable above?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+            UpdateFillTimer();
+
+            if (HasFocus && CurrentDwellState != DwellStateType.DwellCompleted)
+            {
+                double focusDuration = (DateTime.UtcNow - this.focusEnterTime).TotalSeconds;
+
+                if (CurrentDwellState == DwellStateType.FocusGained && focusDuration >= dwellProfile.DwellIntentDelay)
+                {
+                    DwellIntended.Invoke(pointer);
+                    CurrentDwellState = DwellStateType.DwellIntended;
```
wiwei: Does the order matter here? If the caller asks for the current state, they will still see that it's in the FocusGained state.

Or is that the intent (in which case having a comment saying that this is the general pattern would also help to provide a better understanding of the mechanics of how these things are used)

wiwei: What I mean by this is, I'm assuming DwellIntended gets invoked sync here, which means if the caller tries to ask this dwell handler what state its in (because maybe it's calling some common code that is not aware it's being called from DwellIntended) this could be weird maybe

---
In reply to: [314431451](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r314431451) [](ancestors = 314431451)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+                }
+                else if (CurrentDwellState == DwellStateType.DwellIntended && (focusDuration - dwellProfile.DwellIntentDelay) >= dwellProfile.DwellStartDelay)
```
wiwei: >(focusDuration - dwellProfile.DwellIntentDelay) [](start = 78, length = 47)

Nit: This section is interesting because it tracks the current time in the current phase - you don't need to do this, but from a readability perspective it would would easier to read to have something like:

GetTimeInCurrentState() - which would then have a cleaner/more readable thing in these if statements, and then also have a single helper where it's clear that each of these operations (i.e. focusDuration - whatever - whatever) is clear that it's how you calculate the time in each state.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+                    FillTimer = FillTimer < 0 ? 0 : FillTimer;
+                    break;
+                case DwellStateType.DwellCompleted:
+                    Debug.Assert(FillTimer < float.Epsilon, $"Fill timer was not 0. It was {FillTimer}.");
+                    break;
+                case DwellStateType.DwellCanceled:
+                    if ((DateTime.UtcNow - focusExitTime).TotalSeconds > dwellProfile.TimeToAllowDwellResume)
+                    {
+                        FillTimer = 0;
+                        CurrentDwellState = DwellStateType.None;
```
wiwei: >CurrentDwellState = DwellStateType.None; [](start = 24, length = 40)

It feels weird that a function called UpdateFillTimer will also update the current dwell state - also interesting because subclasses could forget to update this dwell state change.

If this is intended to address a specific issue, it might be worth having a comment about what this is for

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+                && eventData.Pointer.InputSourceParent.SourceType == dwellProfile.DwellPointerType)
+            {
+                HasFocus = true;
+
+                // check intent to resume
+                if (CurrentDwellState == DwellStateType.DwellCanceled
+                    && pointer.InputSourceParent.SourceId == eventData.Pointer.InputSourceParent.SourceId //make sure the returning pointer id is the same
+                    && (DateTime.UtcNow - focusExitTime).TotalSeconds <= dwellProfile.TimeToAllowDwellResume)
+                {
+                    focusEnterTime = focusEnterTime.AddSeconds(dwellProfile.TimeToAllowDwellResume);
```
wiwei: >focusEnterTime = focusEnterTime.AddSeconds(dwellProfile.TimeToAllowDwellResume); [](start = 20, length = 80)

Can you add a comment about why focusEnterTime isn't just when OnFocusEnter called?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,243 @@
...
+                if (CurrentDwellState == DwellStateType.DwellStarted)
+                {
+                    DwellCanceled.Invoke(eventData.Pointer);
+                    CurrentDwellState = DwellStateType.DwellCanceled;
+                    focusExitTime = DateTime.UtcNow;
+                }
+                else
+                {
+                    CurrentDwellState = DwellStateType.None;
+                    focusExitTime = DateTime.MinValue;
```
wiwei: >MinValue [](start = 45, length = 8)

Out of curiousity, why does this use MinValue (even though the default "no value" is MaxValue when it gets initialized way above?)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/CustomDwellHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+                case DwellStateType.DwellCompleted:
+                    FillTimer = 0;
+                    break;
+                case DwellStateType.DwellCanceled:
+                    var customDwellProfile = dwellProfile as DwellProfileWithDecay;
+                    if (customDwellProfile.AllowDwellDecayOnCancel)
+                    {
+                        FillTimer += Time.deltaTime;
+                    }
+                    break;
```
wiwei: Corresponding question on DwellHandler (around there being a state transition in the base class, that's missing in this one)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/InstantDwellSample.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public override void DwellStarted(IMixedRealityPointer pointer)
+        {
+            base.DwellStarted(pointer);
+        }
+
+        public override void DwellCanceled(IMixedRealityPointer pointer)
+        {
+            isDwelling = false;
```
wiwei: >isDwelling = false; [](start = 12, length = 19)

Nit: I think this can just call base.DwellCancelled right? I think it does the same thing

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/InstantDwellSample.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        public override void DwellCompleted(IMixedRealityPointer pointer)
+        {
+            isDwelling = false;
+            dwellVisualImage.transform.localScale = Vector3.zero;
+            base.DwellCompleted(pointer);
```
wiwei: Same-ish comment here (I think that this function could just be:

dwellVisualImage.transform.localScale = Vector3.zero;
base.DwellCompleted(pointer);

Assets/MixedRealityToolkit.SDK/Inspectors/Dwell/DwellHandlerInspector.cs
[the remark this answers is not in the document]
wiwei: The suggestion makes sense, though because we have prior art here and I think it would be best to keep consistent with the existing code.

As-is is fine because I think we'd value consistency in this case (despite being consistent with something that could obviously be improved) over the incremental value this provides

---
In reply to: [313542173](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r313542173) [](ancestors = 313542173)

Assets/MixedRealityToolkit.SDK/Inspectors/Dwell/DwellProfileWithDecayInspector.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using UnityEditor;
+
+namespace Microsoft.MixedReality.Toolkit.Experimental.Dwell.Editor
```
wiwei: >Microsoft.MixedReality.Toolkit.Experimental.Dwell.Editor [](start = 10, length = 56)

With this inspector, it would probably be also good to have this in the MRTK.Examples folder (because the thing it inspects is going to be moved to the MRTK examples folder as well)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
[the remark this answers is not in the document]
ritijain: The existing UnityPointerEvent is public class PointerUnityEvent : UnityEvent<**MixedRealityPointerEventData**> { } 
so it is a different type. i will rename this one, and put it in a separate to follow MRTK coding guidelines.

wiwei:
@ritijain looks good overall! Let me know when you want me to take another look after this round of comments

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
ritijain: That is a good idea. by documents do you mean in md files or in code comments ?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
ritijain: The CurrentDwellState  is a protected variable and so the caller does not have access to its value. But i think it is good to invoke the events after state transitions for consistent pattern.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
ritijain: my oversight. :(

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
ritijain: updated the comment with better description

Assets/MixedRealityToolkit.SDK/Inspectors/Dwell/DwellProfileWithDecayInspector.cs
ritijain: Sure. When i move it under Examples, I get build error though. 
error CS0115: 'DwellProfileWithDecayInspector.OnInspectorGUI()': no suitable method found to override

Seems like i need to  add it to some exclusion list in build steps. Any pointers on where that needs to be done ?

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        /// <summary>
+        /// Cached pointer reference to track focus events maps to teh same pointer that initiated dwell
+        /// </summary>
+        private IMixedRealityPointer pointerRef;
+
+        private DateTime focusEnterTime = DateTime.MaxValue;
+        private DateTime focusExitTime = DateTime.MaxValue;
+
+        protected DwellState currentDwellState = DwellState.None;
```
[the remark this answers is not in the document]
ritijain: Yes. Depending on the "Dwell experience" that people want, they might want to update this based on desired experiences.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
wiwei: Just code comments - I think that they're generally helpful here and then when we get around to doing MD files we can just XREF to these code comments

wiwei: Yeah sorry, doccomments == the <summary> type code comments that generate stuff through docfx.

comments = regular code comments
documentation markup stuff = MD file stuff

Assets/MixedRealityToolkit.SDK/Inspectors/Dwell/DwellProfileWithDecayInspector.cs
wiwei: Do you have a local branch that I could check out that has your changes in that state?

ritijain: https://github.com/ritijain/MixedRealityToolkit-Unity/tree/gazeDwellv1

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/CustomDwellHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Experimental.Dwell
+{
+    /// <summary>
+    /// Example to demonstrate override
+    /// </summary>
+    public class CustomDwellHandler : DwellHandler
+    {
+        public override float CalculateDwellProgress()
```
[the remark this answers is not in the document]
ritijain: Can't use this syntax since some MRTK projects do not use c#7 yet :(

---
In reply to: [313156541](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r313156541) [](ancestors = 313156541)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
ritijain: Yeah it is ok for the deriving class to "forget". In most dwell variations I have seen the DwellCanceled state can transition to None or Started based on conditions. Those conditions can vary based on the profile supplied to the implementation. It can be left up to the developer to decide what state transition they want.

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
ritijain: Should i rename "DwellIntended" to "DelayedHover" ? (i have received similar feedback from other folks in MRW, that DwellIntended is not clear enough terminology)

---
In reply to: [314125680](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r314125680) [](ancestors = 314125680)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
ritijain: I feel this is clear enough... marking as Won't Fix unless you have strong feelings about this.

ritijain:
> @ritijain looks good overall! Let me know when you want me to take another look after this round of comments

@wiwei  - addressed most comments. Have a few follow-ups which i have left in comments for. Main blocking issue is the custom inspector that needs to go in the Examples folder.  (i could get rid of the custom inspector if it turns into a big issue, but i think it might be worth getting to the bottom of why the build fails and how we can get around it)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/Samples/CustomDwellHandler.cs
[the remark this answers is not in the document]
ryantrem: Is MRTK trying to support <= Unity 2018.3? Everything 2018.3+ supports C# 7.

---
In reply to: [316857577](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5594#discussion_r316857577) [](ancestors = 316857577,313156541)

Assets/MixedRealityToolkit.SDK/Experimental/Features/Dwell/DwellHandler.cs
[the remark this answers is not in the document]
ryantrem: DwellProfile is a ScriptableObject, and CreateInstance<T> specifically returns a T. If there is a problem, I'm not understanding what it is.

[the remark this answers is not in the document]
ryantrem: Oh wait, it looks like despite the above comment you did actually change this. Re-resolving. 🙂

Issue #5568, opened 2019-08-09
[Editor][NuGet] The simulated articulated hand visualization only appears as a single orb in-editor when getting MRTK via NuGet


Kjakubzak:
## Describe the bug

When using MRTK via NuGet packages, the articulated hand simulation visualization is a single orb rather than a full hand.

## To reproduce

1. Pull MRTK using NuGet
2. Set the input simulation type to Articulated
3. Press play in an example scene
4. See a single orb rather than a whole hand

## Expected behavior

A full hand representation is visible, rather than a single orb

## Screenshots

![image](https://user-images.githubusercontent.com/1617757/62796904-37287a00-ba8f-11e9-843c-17c08d6eb7e3.png)

## Your Setup (please complete the following information)

- Unity Version 2018.4.4f1
- MRTK Version 2.0.0-20190730.2

## Target Platform (please complete the following information)
Editor - H2

PR #5603, opened 2019-08-12
Update the MRTK to support being placed in arbitrary folder names up to three directories deep.

wiwei:
In looking into #5568, I found that the existing MRTK files code would look for specific folder names that MRTK gets put into (as in, containing folder names). It had very specific behaviors with respect to NuGetForUnity or consumption as .unitypackage - if you had consumed the code some other way (for example, used a manual nuget restore step separately from NuGetForUnity) this wouldn't work out, because the naming of the folders would be different (i.e. MRTK/*) instead of (Microsoft.MixedReality.Toolkit*/*).

This change works around all that by introducing the concept of "sentinel" files, which are empty files that are named in a specific way and uniquely identify the root of an MRTK module/subfolder. Thus, when the MRTK files code finds a file called MRTK.SDK.sentinel, it knows that that folder is the root of the SDK MRTK subfolder, and it doesn't matter what name the root folder (or folders going up the root) are.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using UnityEditor;
 using UnityEngine;
@@ -34,6 +35,15 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
     /// <summary>
     /// API for working with MixedRealityToolkit folders contained in the project.
     /// </summary>
+    /// <remarks>
+    /// This class works by looking for sentinel files (following the pattern MRTK.*.sentinel,
+    /// for example, MRTK.Core.sentinel) in order to identify where the MRTK is located
+    /// within the project.
+    ///
+    /// If the MRTK is being consumed as code that sits within the Assets folder, the "root"
+    /// MRTK folder must be at most three directories deep - this search code will only reason
+    /// over MRTK folders that sit in a depth range [0, 3].
+    /// </remarks>
     [InitializeOnLoad]
     public static class MixedRealityToolkitFiles
     {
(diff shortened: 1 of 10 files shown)


Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -59,15 +77,25 @@ public static void OnPostprocessAllAssets(string[] importedAssets, string[] dele
...
+
+                foreach (string folder in importedFolders)
+                {
                     TryRegisterModuleFolder(folder);
                 }
 
                 foreach (string asset in deletedAssets.Concat(movedFromAssetPaths))
                 {
-                    string folder = ResolveFullAssetsPath(asset);
+                    string fullAssetPath = ResolveFullAssetsPath(asset);
```
andreiborodin: Could we just pattern match the sentinel files in this callback, so that if a folder is moved we don't do a deep recursive search for each file in that folder?

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
```diff
@@ -59,15 +77,25 @@ public static void OnPostprocessAllAssets(string[] importedAssets, string[] dele
             {
                 searchForFoldersTask.Wait();
 
+                // Identify the unique set of imported folders instead of unique files
+                // in order to avoid duplicating per-folder work.
+                HashSet<string> importedFolders = new HashSet<string>();
                 foreach (string asset in importedAssets.Concat(movedAssets))
                 {
-                    string folder = ResolveFullAssetsPath(asset);
+                    string fullAssetPath = ResolveFullAssetsPath(asset);
```
andreiborodin: Same question as below, could we pattern match for sentinel files here?

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
wiwei: Yep, makes sense!

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
wiwei: Alrighty, posting the resolution in this comment instead of in both. A couple of changes I took here:

1) Make it so that when processing new assets, that we match directly against the regex instead of dropping to folder and then reenumerating. As you mention, this is way better than what I had before.
2) When processing removed assets, fixed that issue where deleting ANY arbitrary file in the root module folder would also cause us to unregister the module folder - now it's only done if the module type matches)

Issue #5545, opened 2019-08-07
GazeProvider is not able to be implemented as editor UI Indicates, and it's impossible to define your own GazePointer


Kjakubzak:
## Describe the bug

Right now, you are not able to properly define and set your own IMixedRealityGazeProvider in the pointer configuration. The FocusProvider implementation takes the passed in IMixedRealityGazeProvider and then casts it to the GazeProvider in UpdateGazeProvider() to call  .UpdateGazeInfoFromHit(). This causes problems for a IMixedRealityGazeProvider that's not the MRTK-defined GazeProvider.

## To reproduce

1. Write a new GazeProvider that implements the IMixedRealityGazeProvider class
2. Assign the new implementation to the GazeProvider option in the Pointer profile 
3. Click play, and see the casting exception occur

![image](https://user-images.githubusercontent.com/1617757/62663010-d41dd280-b92a-11e9-9169-56812d25f3df.png)

## Expected behavior

It doesn't throw an exception and allows you to have your own implementation of the GazeProvider. 

Also, being able to implement your own GazePointer and provide that to the GazeProvider seems like a feature that's missing; given that user-defined pointer definition is available for all other pointer types.
 
## Your Setup (please complete the following information)

- Unity Version 2018.4.4f1
- MRTK Version - mrtk_development - 2.0.0-20190730.2

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2

## Additional context

Spoke to @wiwei and @keveleigh about this already and just wanted to immortalize it in bug form.

keveleigh:
Fixed via #5610. Opened https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5612 to track custom gaze pointers

PR #5610, opened 2019-08-12
Remove concrete GazeProvider reference to allow custom providers

keveleigh:
## Overview
This PR moves UpdateGazeInfoFromHit into the interface definition to remove a cast in FocusProvider to the concrete type.

It also performs some formatting / comment updates to match current MRTK-style more.

## Changes
- Fixes: #5545 (at least, as far as GazeProvider is concerned. There's also the note about potentially defining custom GazePointers, like other pointers are handled)

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
@@ -492,6 +492,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
             GazePointer.BaseCursor?.SetVisibility(true);
         }
 
+        /// <inheritdoc />
         public void UpdateGazeInfoFromHit(MixedRealityRaycastHit raycastHit)
         {
             HitInfo = raycastHit;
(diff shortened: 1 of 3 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/GazeProvider.cs
```diff
@@ -163,8 +163,6 @@ public IMixedRealityInputSource GazeInputSource
 
         private Ray latestEyeGaze = default(Ray);
         private DateTime latestEyeTrackingUpdate = DateTime.MinValue;
-        private bool? isEyeCalibrated = null;
-
         private readonly float maxEyeTrackingTimeoutInSeconds = 2.0f;
 
         #region InternalGazePointer Class
```
Kjakubzak: I think that having the InternalGazePointer class be a self-contained entity within the GazeProvider can lead to confusion for developers; especially when there's a UI element in the editor that allows them to define pointers (other than for gaze). 

It doesn't need to be addressed in this PR, but can a work item be opened up to address this concern in the future? While being able to define your own GazeProvider means you also are able to define your own InternalGazePointer; it isn't very easily discoverable.

Kjakubzak: Also, in the OnPreSceneQuery function for the internal gaze pointer; there's this line:
`gazeProvider.HitPosition = Rays[0].Origin + (gazeProvider.lastHitDistance * Rays[0].Direction);
`
Which then gets blown away by the call to UpdateGazeInfoFromHit() from the FocusProvider.

keveleigh: Agreed! I'll file an issue about InternalGazePointer

keveleigh: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5612

PR #5611, opened 2019-08-12
Added follow behavior used in OOBE (based on follow behavior from shell)

ErianVazquez-Microsoft:
## Overview
Added the follow behavior used in OOBE to the list of sample solvers.

## Changes
Added Follow.cs (a solver used in OOBE) and changed the SolverExample Scene to contain this solver option. This behavior can now be applied to the mug. 

## Verification
Reviewers please checkout the PR locally to validate that I modified SolverExamples.unity to your coding standards. 

git fetch origin pull/{PR_ID}/head:user/ervazque/Add_follow_from_shell

git checkout user/ervazque/Add_follow_from_shell


julenka:
Thanks for adding this!

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Follow.cs
```diff
@@ -0,0 +1,573 @@
...
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Solvers
+{
+    /// <summary>
+    /// Follow solver positions an element relative in front of the forward axis of the reference.
+    /// The element can be loosly constrained (a.k.a. tag-along) so that it doesn't follow until it is too far.
+    /// </summary>
+    public class Follow : Solver
```
julenka: I'm very curious (perhaps @keveleigh  you can explain since you understand the old solvers), isn't the Orbital solver supposed to implement tag-along behavior as well? Note that the name "follow" makes much more sense to me than "orbital" (which to me sounds like a planet orbiting me, so it's a behavior that orbits around as time goes on). But, I'd just like to understand once we get this in, why anybody would want to use Orbital.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Follow.cs
```diff
@@ -0,0 +1,573 @@
...
+    {
+        [SerializeField]
+        [Tooltip("Position lerp multiplier")]
+        private float moveToDefaultDistanceLerpTime = 0.1f;
+
+        /// <summary>
```
julenka: I've always found this comment / description unclear. Instead of describing what the variable is, it would be more helpful to explain how increasing / decreasing the value changes the follow behavior. Also, does the "time" here actually mean "seconds for object to lerp to the desired position"? I don't see how the math leads to that, I always wondered if "time" here was misleading...

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Follow.cs
```diff
@@ -0,0 +1,573 @@
...
+        }
+
+        [SerializeField]
+        [Tooltip("Pitch offset from reference element (relative to Max Distance)")]
+        public float pitchOffset = 0;
+
+        /// <summary>
+        /// Pitch offset from reference element (relative to MaxDistance).
+        /// </summary>
+        /// [SerializeField]
```
julenka: should be removed.

julenka:
@cre8ivepark have a look at this PR. You may actually want to update the toggle panel prefab to use this new follow behavior that more closely matches the shell.

julenka:
Ah, one more suggestion: it would be very nice for this change (which is a new, user facing feature) to attach a .gif of the behavior you are enabling, with screenshots of new scene elements you're adding to the sample. For example, have a look at [this PR](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4532) from @Cameron-Micka  which is an excellent example of a descriptive PR with animation that helps describe the behavior being added.

You can use a variety of tools to do a screen capture and convert to gif, the tool I use is [screentogif](https://www.screentogif.com/)

Troy-Ferrell:
Thanks for the PR. Along with the tests missing there is no documentation which needs to be added. Gifs work great as Julia said. Licecap, terrible name, is another useful tool. 

I am a bit concerned though about adding another solver. Especially it looks like lots of overlap with radial view and orbital. I think radial view is used for the toggle feature on the slates? There are already a lot of solvers and I think developers already have many issues understanding them. The name follow is much better than radial view though. 

Again it just seems like another component to have test overhead, documentation overhead, and for some one to manage and maintain? 

Can we not just update radial view to be better?

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Follow.cs
```diff
@@ -0,0 +1,573 @@
...
+            get { return orientationType; }
+            set { orientationType = value; }
+        }
+
+        [SerializeField]
+        [Tooltip("Min distance from eye to position element around, i.e. the sphere radius")]
+        private float minDistance = 1f;
+
+        /// <summary>
+        /// Min distance from eye to position element around, i.e. the sphere radius.
```
Troy-Ferrell: We need to avoid using eye or head. Solvers should be able to function with different tracked transforms

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Follow.cs
```diff
@@ -0,0 +1,573 @@
...
+        /// </summary>
+        public float DefaultDistance
+        {
+            get { return defaultDistance; }
+            set { defaultDistance = value; }
+        }
+
+        [SerializeField]
+        [Tooltip("The element will stay at least this close to the center of view")]
+        private float maxViewHorizontalDegrees = 30f;
```
Troy-Ferrell: Isn’t this like radial view? Same with the distance properties

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Follow.cs
```diff
@@ -0,0 +1,573 @@
...
+
+            PreviousReferencePosition = refPosition;
+            PreviousReferenceRotation = refRotation;
+            recenterNextUpdate = false;
+
+            UpdateWorkingPositionToGoal();
+            UpdateWorkingRotationToGoal();
+        }
+
+        float AngleBetweenOnXZPlane(Vector3 from, Vector3 to)
```
Troy-Ferrell: Add explicit private to be consistent with codebase

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Orbital.cs
```diff
@@ -121,7 +121,7 @@ private Quaternion SnapToTetherAngleSteps(Quaternion rotationToSnap)
             return Quaternion.Euler(rotationToSnap.eulerAngles.x, newAngle, rotationToSnap.eulerAngles.z);
         }
 
-        private Quaternion CalculateDesiredRotation(Vector3 desiredPos)
+        private Quaternion CalculateDesiredRotation(Vector3 desiredPos, Quaternion previousGoalRotation)
```
Troy-Ferrell: Parameter seems unnecessary. Why not just set desiredRot to GoalRotation?

Assets/MixedRealityToolkit/Definitions/Utilities/SolverOrientationType.cs
```diff
@@ -28,6 +28,10 @@ public enum SolverOrientationType
         /// <summary>
         /// Align parallel to the direction the camera is facing 
         /// </summary>
-        CameraAligned
+        CameraAligned,
+        /// <summary>
+        /// Leave the current goal rotation  
+        /// </summary>
```
Troy-Ferrell: Also not clear how MaintainGoal and Unmodified are fully different. Most of the time GoalRotation will just be transform.rotation. In linked transforms, I guess there would be a difference. But this is extremely confusing to a developer who won’t understand how solvers work.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Follow.cs
Troy-Ferrell: I really like the idea of tool tips explaining the mechanism of what happens. MRTK has a lot of tool tips that are just redundant of the parameter name instead of really illuminating and helpful to the developer

davidkline-ms:
Closing as it has been folded into #5713

PR #5615, opened 2019-08-12
Fix up performance degradation associated with postponed actions.

wiwei:
The work to add postponed actions/registrations in https://github.com/microsoft/MixedRealityToolkit-Unity/commit/c712eb128deea8cdc4633d11b76a6207f440fe7c was missing a clearing of the postponedActions and postponedObjectActions.

This leads to the array growing really really large over time, where on each input event, it would proceed to try to register/unregister the same thing (that had already been registered/unregistered). As time goes on, it grows so large that each update tick's eventing work would get overwhelmed by the busy work trapped in postponedActions and postponedObjectActions.

### DIFF
--- a/Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
+++ b/Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
@@ -126,6 +126,9 @@ namespace Microsoft.MixedReality.Toolkit
                         Unregister(obj.Item2);
                     }
                 }
+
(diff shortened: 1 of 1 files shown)


julenka:
Things look good on HL2, verified this doesn't break stuff

wiwei:
Looks like @Troy-Ferrel already has this covered in another PR: https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5618

Issue #5391, opened 2019-07-19
would like to be able to configure FocusProvider to target colliders instead of rigid bodies when both are present


naostranMS:
## Describe the problem

In FocusProvider.PointerHitResult.Set(MixedRealityRaycastHit hit, ...) and MixedRealityRaycaster.TryGetPrioritizedPhysicsHit, it uses the RaycastHit.transform.gameObject as the target gameObject. In the case of compouned colliders (example here: [https://docs.unity3d.com/Manual/class-Rigidbody.html](url)), the transform will be that of the parent rigidbody rather than the particular collider that was hit.

E.g. when raycasting at the cylinder in this setup (where the Parent object has a RigidBody, and the children each have colliders):
![image](https://user-images.githubusercontent.com/38891743/61568111-d204da00-aa36-11e9-97f6-c1f8ace9d02a.png)
you get the following raycast result:
![image](https://user-images.githubusercontent.com/38891743/61568062-910cc580-aa36-11e9-8ebb-55ecfe290248.png)

In our forked version of HTK, we changed it to use the collider as the focus object to get more granularity, and without that in MRTK, we end up with a lot of broken functionality due to the focus target being different than expected (and also due to prioritization changes because the collider game objects are often on a different layer than the parent rigidbody).

## Describe the solution you'd like

I would like to have a simple way of configuring or overriding the default FocusProvider behavior to use hit.collider.gameObject instead of hit.transform.gameObject.

## Describe alternatives you've considered

1. It might be possible to subclass the FocusProvider and override the needed functionality, but it seems like that would take a lot of extra/duplicated code in the override for what is a very small difference.
2. I imagine the data could be rearchitected to work with the existing behavior, but I think that would be a pretty substantial undertaking.

Alexees:
Hi @naostranMS I felt this task was fun and I did a little something, maybe you're interested in checking it out:
https://github.com/HoloSpaces/MixedRealityToolkit-Unity/tree/mrtk_removeMixedRealityRaycastResult

All you need to do is opening the MixedRealityInputSystem profile you're using and enable "FocusIndividualCompoundCollider"

Would love to hear if that works for you.

naostranMS:
@Alexees thanks for looking into this!
[...]

PR #5617, opened 2019-08-12
Expose configuration in FocusProvider to target colliders instead of rigid bodies

ritijain:
## Overview

This PR addresses issue #5391. 

## Changes
This change introduces a configuration property in the Input System Settings in MRTK profiles - "FocusIndividualCompoundCollider". This configurable property is passed as a argument to the Raycast provider implementation and ultimately propagates to the TryGetPrioritizedPhysicsHit logic to determines, if the raycast hit result should pick individual colliders or the root rigidbody transform.

When the hit is determined, the profile property is again used to determine which gameobject to set as the rayast hit object .

image

## Verification
Validation requested from @anuraag016  via product integration.

Created a local scene with a rigidbody and compound colliders and verified that the correct item is focused.

FocusIndividualCompoundCollider = false;
colliderfalse

FocusIndividualCompoundCollider = true
collider
;

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Physics/MixedRealityRaycaster.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Physics/MixedRealityRaycaster.cs
@@ -16,9 +16,9 @@ namespace Microsoft.MixedReality.Toolkit.Physics
         /// <param name="prioritizedLayerMasks"></param>
         /// <param name="physicsHit"></param>
         /// <returns>Whether or not the raycast hit something.</returns>
-        public static bool RaycastSimplePhysicsStep(RayStep step, LayerMask[] prioritizedLayerMasks, out RaycastHit physicsHit)
+        public static bool RaycastSimplePhysicsStep(RayStep step, LayerMask[] prioritizedLayerMasks, bool focusIndividualCompoundCollider, out RaycastHit physicsHit)
         {
-            return RaycastSimplePhysicsStep(step, step.Length, prioritizedLayerMasks, out physicsHit);
+            return RaycastSimplePhysicsStep(step, step.Length, prioritizedLayerMasks, focusIndividualCompoundCollider, out physicsHit);
         }
 
         /// <summary>
@@ -29,7 +29,7 @@ namespace Microsoft.MixedReality.Toolkit.Physics
         /// <param name="prioritizedLayerMasks"></param>
         /// <param name="physicsHit"></param>
         /// <returns>Whether or not the raycast hit something.</returns>
-        public static bool RaycastSimplePhysicsStep(RayStep step, float maxDistance, LayerMask[] prioritizedLayerMasks, out RaycastHit physicsHit)
+        public static bool RaycastSimplePhysicsStep(RayStep step, float maxDistance, LayerMask[] prioritizedLayerMasks, bool focusIndividualCompoundCollider, out RaycastHit physicsHit)
         {
             Debug.Assert(maxDistance > 0, "Length must be longer than zero!");
             Debug.Assert(step.Direction != Vector3.zero, "Invalid step direction!");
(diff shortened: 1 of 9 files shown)


ritijain:
@anuraag016, @naostranMS, @davidkline-ms - FYI

davidkline-ms:
@wiwei, @keveleigh,

this has some API and profile changes. if we are going to take if for dev, we should strongly consider it for the stabilization branch. thoughts?

Assets/MixedRealityToolkit/Utilities/Physics/MixedRealityRaycaster.cs
```diff
@@ -38,14 +38,14 @@ public static bool RaycastSimplePhysicsStep(RayStep step, float maxDistance, Lay
...
+                : TryGetPrioritizedPhysicsHit(UnityEngine.Physics.RaycastAll(step.Origin, step.Direction, maxDistance, UnityEngine.Physics.AllLayers), prioritizedLayerMasks, focusIndividualCompoundCollider, out physicsHit);
         }
 
         /// <summary>
         /// Box raycasts each physics <see cref="Microsoft.MixedReality.Toolkit.Physics.RayStep"/>.
         /// </summary>
         /// <returns>Whether or not the raycast hit something.</returns>
-        public static bool RaycastBoxPhysicsStep(RayStep step, Vector3 extents, Vector3 targetPosition, Matrix4x4 matrix, float maxDistance, LayerMask[] prioritizedLayerMasks, int raysPerEdge, bool isOrthographic, out Vector3[] points, out Vector3[] normals, out bool[] hits)
+        public static bool RaycastBoxPhysicsStep(RayStep step, Vector3 extents, Vector3 targetPosition, Matrix4x4 matrix, float maxDistance, LayerMask[] prioritizedLayerMasks, int raysPerEdge, bool isOrthographic, out Vector3[] points, out Vector3[] normals, bool focusIndividualCompoundCollider,out bool[] hits)
```
AdamMitchell-ms: >bool focusIndividualCompoundCollider [](start = 259, length = 36)

You should probably put this param before all of the out params.  Also need a space after the comma.

Troy-Ferrell:
I really like the idea of this feature. I think developers would appreciate the greater power and there has been some feedback to for them to better control what gets focus, etc. Few problems I see though

1) FocusIndividualCompoundCollider is a mouthful and for most people doesn't really inform them of what it is
2) There is no documentation with this change to teach people or tests to verify correct functionality
3) The global value isn't really catch all. Basically could we think of a better approach that would allow hit.collider.gameobject and hit.transform.gameobject results? I could see developers wanting one in some scenarios of their app and the other for separate parts
4) I don't see any way to alter this behavior at runtime. It's set in the profile but I can't modify the FocusProvider functionality directly after instantiation
5) Also I feel like MixedRealityRaycaster.Raycast* functions may need a refactor in general. Especially MixedRealityRaycaster.RaycastBoxPhysicsStep is just insanely long in the parameter list

Assets/MixedRealityToolkit/Utilities/Physics/MixedRealityRaycaster.cs
```diff
@@ -163,7 +163,12 @@ public static bool TryGetPrioritizedPhysicsHit(RaycastHit[] hits, LayerMask[] pr
                 for (int hitIdx = 0; hitIdx < hits.Length; hitIdx++)
                 {
                     RaycastHit hit = hits[hitIdx];
-                    if (hit.transform.gameObject.layer.IsInLayerMask(priorityLayers[layerMaskIdx]) &&
+                    GameObject targetGameObject = hit.transform.gameObject;
```
AdamMitchell-ms: >hit.transform.gameObject [](start = 50, length = 24)

I'd use ternary syntax here.

GameObject targetGameObject = (focusIndividualCompoundCollider) ? hit.collider.gameObject : hit.transform.gameObject;

davidkline-ms:
Adding this to the must fix for GA as it is both an interface API and profile breaking change.

@ritijain, can you please retarget this change to prerelease.2.0.0_stabilization in the next day so that we can test and validate?

thanks!

ritijain:
> Adding this to the must fix for GA as it is both an interface API and profile breaking change.
> 
> @ritijain, can you please retarget this change to prerelease.2.0.0_stabilization in the next day so that we can test and validate?
> 
> thanks!

@davidkline-ms  - retargeted to pre-release. Validation pending.

naostranMS:
this looks like it should work for #5391

Assets/MixedRealityToolkit/Utilities/Physics/MixedRealityRaycaster.cs
wiwei:
```suggestion
        public static bool RaycastBoxPhysicsStep(RayStep step, Vector3 extents, Vector3 targetPosition, Matrix4x4 matrix, float maxDistance, LayerMask[] prioritizedLayerMasks, int raysPerEdge, bool isOrthographic, bool focusIndividualCompoundCollider, out Vector3[] points, out Vector3[] normals, out bool[] hits)
```

wiwei: Per Adam's comment

wiwei:
Change looks good - I can approve once Adam's nits get addressed

Yes, targetting at stabilization is good because this isn't something we want folks to absorb in a future release.

ritijain:
> Change looks good - I can approve once Adam's nits get addressed
> 
> Yes, targetting at stabilization is good because this isn't something we want folks to absorb in a future release.

@wiwei - PR updated.

Issue #5572, opened 2019-08-09
Performance drops to 45fps on one hand detected, after a lot of manipuation. Drops to 22 fps on two hands detected.


cre8ivepark:
## Describe the bug
After a lot of manipulation interaction (one or two-handed), the Profiler shows the framerate drop to 45fps on two hands detected even there are no objects in the viewport. (no hand mesh/joint)
Presenting a single hand seems fine (60fps). Only drops when both hands are detected.

After pausing and restarting the app and manipulating the HL2 style bounding box, then bringing up a single hand can cause the framerate to drop to 45 fps, and bringing up 2 hands causes frame rate drop to 22 fps.

## To reproduce
In the HandInteractionExample scene, keep interacting with the objects with ManipulationHandler and BoundingBox. Use two-handed manipulation for scaling/rotating.

## Expected behavior
No framerate drop

## Your Setup (please complete the following information)
- Unity Version 2018.4.

## Target Platform (please complete the following information)
- HoloLens 2

## Branch
- prelease

Troy-Ferrell:
This behavior is also pretty noticeable in Unity editor on desktop. Looking quickly into the Unity profiler, these are the bigger hitters that start taking up computation on hands shown (Input simulation Hands)

BoundingBox (Multiple instances) -> Update()
MRTK.Update() -> Huge increase
MRTK.LateUpdate() -> Huge increaase
Physics->Processing

davidkline-ms:
Is this a recent regression?

Troy-Ferrell:
@davidkline-ms hard to fully say. The BoundingBox isn't recent and I believe was part of #4895 . The MRTK update and MRTK late update may have been there a while too. Haven't investigated yet

Troy-Ferrell:
The MRTK update with hands appears to be mostly in UpdatePointers(), and in particular the Graphics raycast for every pointer Ray step is very expensive

DefaultControllerPointer has 2 rays instead of just one but setting to 1 breaks Unity UI slider?

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5615

PR #5618, opened 2019-08-12
Input, Hand joints, and BoundingBox perf improvements

Troy-Ferrell:
## Overview
There are three areas of bugs found for perf improvements into the stabilization branch.

1)
The most dramatic perf impact was identified by @wiwei . Only on device, the showing and hiding of hands would cause input to handle events and a list of "postponed" action events were never cleared over time. This caused the list to balloon through app usage and testing with Julia at one point reached 3000+ action list! Being looped through every frame. This caused framerate to collapse into below 30 territory. 

2)
BoundingBox called update scale handle for every BoundingBox in the scene for every rotation/scale handle on a given BB, regardless of visibility or state. This caused quite a significant overhead on the CPU.
-> If handles are not visible (mostly wireframeOnly = true), then proximity scaling does not execute at all
-> BoundingBox now more efficiently acquires all possible pointers
-> BoundingBox only test pointer types that are visible (i.e ShowScaleHandles = true, etc.)
-> BoundingBox only tests pointer results that are within a sphere of influence of the BB. Thus, if the user is far away etc. then no need to calculate proximity for every handle
-> Fixed some summary tags and empty tooltips in BoundingBox properties

3)
HandJointUtils did not efficiently acquire a given hand joint. Every call will cause the class to try to find the latest input system and try to find the given hand by looping through detected controllers. Furthermore, in most cases (particularly SpherePointer), the calling class already had the controller reference and thus only needed to cast to IMixedRealityHand and call TryGetJoint() itself which is way more efficient.

## Changes
- Fixes: #5572 

## Verification
@julenka  tested HandInteractionExample scene on HL2 device
All tests passed

>
>


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -649,109 +749,94 @@ public bool ShowRotationHandleForZ
...
-        public bool DrawTetherWhenManipulating
-        {
-            get { return drawTetherWhenManipulating; }
-            set { drawTetherWhenManipulating = value; }
-        }
+
+        /// <summary>
+        /// Check to draw a tether point from the handles to the hand when manipulating.
+        /// </summary>
+        public bool DrawTetherWhenManipulating => drawTetherWhenManipulating;
```
keveleigh: This (and quite a few other properties below) had their setters removed in this PR. Was this intentional?

julenka:
I have verified these changes look good on HL2

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -5,6 +5,7 @@
 using Microsoft.MixedReality.Toolkit.Utilities;
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
```
julenka: What requires this?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -130,8 +134,19 @@ public enum RotationHandlePrefabCollider
             Sphere,
             Box
         }
-        #endregion Enums
 
+        /// <summary>
+        /// Container for handle references and states (including scale and rotation type handles)
+        /// </summary>
+        private class Handle
```
julenka: Why not a struct?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
julenka: This also means that DrawTetherWhenManipulating will not be set-table from code, violates principle "make stuff configurable from code"

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1412,57 +1489,19 @@ private void AddLinks()
...
 
-                ballsProximate.Add(HandleProximityState.FullsizeNoProximity);
-                ballRenderers.Add(midpointVisual.GetComponent<Renderer>());
                 if (handleMaterial != null)
                 {
-                    Renderer renderer = midpointVisual.GetComponent<Renderer>();
                     ApplyMaterialToAllRenderers(midpointVisual, handleMaterial);
                 }
             }
-            //if (rotationHandlePrefab == null)
```
julenka: oof. thanks!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Troy-Ferrell: AT line 2206, it's basically to be like c++ inline keyword
[MethodImpl(MethodImplOptions.AggressiveInlining)]

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Troy-Ferrell: Because our handle list is
1) small
2) contains reference types

Making this a struct would require a copy back of the struct every for loop. With List<> you can't do handles[i].state = blah. 

Making the handles member a Handle[] array is annoying because it's not filled in one place. So we need an indexCounter to be tracked/incremented everywhere

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Troy-Ferrell: no it wasn't....I didn't realize it only created getters and not setters

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/FingerCursor.cs
```diff
@@ -171,14 +171,18 @@ protected virtual bool IsNearGrabbableObject(out float dist)
         /// <returns></returns>
         protected bool TryGetJoint(TrackedHandJoint joint, out Vector3 position, out Quaternion rotation)
         {
-            if (Pointer != null && Pointer.Controller != null)
+            if (Pointer != null)
             {
-                if (HandJointUtils.TryGetJointPose(joint, Pointer.Controller.ControllerHandedness, out MixedRealityPose handJoint))
+                var hand = Pointer.Controller as IMixedRealityHand;
```
julenka: i could see this being common code (wanting to get the position of a hand from the pointer or controller). Should we just expose a method HandJointUtils.TryGetJoint(Controller) and HandJointUtils.TryGetJoint(Pointer) ?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/LinePointer.cs
```diff
@@ -140,10 +140,11 @@ public override void OnPostSceneQuery()
         {
             base.OnPostSceneQuery();
 
-            LineBase.enabled = IsInteractionEnabled;
-            BaseCursor?.SetVisibility(IsInteractionEnabled);
+            bool isEnabled = IsInteractionEnabled;
```
julenka: I'm not sure what these changes accomplish. What am I missing?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
```diff
@@ -105,22 +105,24 @@ public override void OnPreSceneQuery()
...
-            // For now, assume that anything that is a sphere pointer is a hand
-            // TODO: have a way to determine if this is a fully articulated hand and return 
-            // ray origin if it's a sixdof
+            // If controller is of kind IMixedRealityHand, return average of index and thumb
             if (Controller != null && Controller is IMixedRealityHand)
             {
-                HandJointUtils.TryGetJointPose(TrackedHandJoint.IndexTip, Controller.ControllerHandedness, out MixedRealityPose index);
-                HandJointUtils.TryGetJointPose(TrackedHandJoint.ThumbTip, Controller.ControllerHandedness, out MixedRealityPose thumb);
-                if (index != null && thumb != null)
+                var hand = Controller as IMixedRealityHand;
```
julenka: About my comment above about HandJointUtils.TryGetJoint(Controller). Seems like a great example of where that method would be useful.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Slate/HandInteractionPanZoom.cs
```diff
@@ -681,13 +681,17 @@ private void SetHandDataFromController(IMixedRealityController controller, IMixe
...
         }
+
         private bool TryGetHandPositionFromController(IMixedRealityController controller, TrackedHandJoint joint, out Vector3 position)
         {
-            if (controller != null &&
-                HandJointUtils.TryGetJointPose(joint, controller.ControllerHandedness, out MixedRealityPose pose))
-            {
-                position = pose.Position;
-                return true;
+            var hand = controller as IMixedRealityHand;
```
julenka: hah, again :-)

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -936,12 +1065,14 @@ public GameObject Target
...
-        }
+        /// <summary>
+        /// The collider reference tracking the bounds utilized by this component during runtime
+        /// </summary>
+        public BoxCollider TargetBounds { get; private set; }
+
+        private List<Handle> handles;
 
-        private List<Vector3> inputPoints = new List<Vector3>();
+        private List<Transform> corners;
```
julenka: Should this be initialized?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -2025,172 +2120,203 @@ private void UpdateRigHandles()
...
 
-        private void HandleProximityScaling(bool forceFar = false)
-        {
-            if (corners?.Count > 0 || balls?.Count > 0)
-            {
-                inputPoints.Clear();
-                //only use proximity effect if nothing is being dragged or grabbed
-                if (currentPointer == null && forceFar == false)
+                // Find all valid pointers
+                foreach (var inputSource in CoreServices.InputSystem.DetectedInputSources)
```
julenka: Not needed for this PR, but I feel like we should create an IEnumerator in InputSystem that enumerates over all pointers, since I imagine many people want to iterate over all pointers in the scene.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -2025,172 +2120,203 @@ private void UpdateRigHandles()
...
+                        {
+                            proximityPointers.Add(pointer);
+                        }
                     }
-                    if (TryGetPointerPoint(Handedness.Right, out point))
+                }
+
+                // Get the max radius possible of our current bounds plus the proximity
+                float maxRadius = Mathf.Max(Mathf.Max(currentBoundsExtents.x, currentBoundsExtents.y), currentBoundsExtents.z);
+                maxRadius *= maxRadius;
```
julenka: Hm....why are we squaring maxRadius here?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -2025,172 +2120,203 @@ private void UpdateRigHandles()
...
+                foreach (var pointer in proximityPointers)
+                {
+                    if (IsPointWithinBounds(pointer.Position, maxRadius))
                     {
-                        inputPoints.Add(point);
+                        proximityPoints.Add(pointer.Position);
                     }
-                    if (TryGetEyeGazePoint(out point))
+
+                    if (IsPointWithinBounds(pointer.Result.Details.Point, maxRadius))
```
julenka: Could pointer.Result be null here?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/FingerCursor.cs
Troy-Ferrell: Yea I wrestled with that. It just seems so easy to accidentally select one of the HandJointUtils options that are way less efficient. I mean that was the original bug here.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/LinePointer.cs
Troy-Ferrell: IsInteractionEnabled isn’t a variable, it’s basically a function. Each call is really testing many variables and levels of inheritance. It’s minor but seemed wasted overhead to do like I believe 5-8 Boolean conditions repeatedly instead of just cashing the results.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
Troy-Ferrell: @julenka, these were instances of people calling HandJointUtils when they didn’t have to after doing a find all references

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Troy-Ferrell: If so same for balls and links. There’s a deep rabbit hole in this file to fix/change a lot

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Troy-Ferrell: Love it!

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Troy-Ferrell: Because the comparison later is against sqrMagnitude. Sqrt is very expensive in distance

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Troy-Ferrell: Ooo yes good catch!

Assets/MixedRealityToolkit/Services/BaseEventSystem.cs
```diff
@@ -126,6 +126,9 @@ public override string ToString()
                         Unregister(obj.Item2);
                     }
                 }
+
+                postponedActions.Clear();
```
thalbern: ❤️

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -2025,172 +2120,203 @@ private void UpdateRigHandles()
...
-                        {
-                            cornersProximate[i] = ScaleHandleByProximity(corners[i], cornerRenderers.Count > 0 ? cornerRenderers[i] : null, cornerVisuals[i], cornersProximate[i], inputPoints, scaleHandleSize);
-                        }
+                        // If handle can't be visisble, skip calculations
+                        if (!IsHandleTypeVisible(handles[i].Type))
+                            continue;
 
-                        for (int i = 0; i < balls.Count; ++i)
+                        // Perform comparison on sqr distance since sqrt() operation is expensive in Vector3.Distance()
+                        float sqrDistance = (handles[i].HandleVisual.position - point).sqrMagnitude;
```
thalbern: ❤️

thalbern:
thanks Troy!

Issue #5600, opened 2019-08-12
Button labels are misaligned in HandInteractionExample


thalbern:
## Describe the bug

The text in the buttons in HandInteractionExample look misplaced. See Screenshot:
![image](https://user-images.githubusercontent.com/36998103/62871945-5552dc00-bd14-11e9-89e1-d5d889016f9e.png)

## To reproduce

Open HandInteractionScene 
Observe misplaced text in Editor
or
Press Play and observe misplaced text in WMR headset

## Expected behavior

Text should be centered in the button

## Your Setup (please complete the following information)

- Stabilization v2.0.0
used latest package from 08-10

## Target Platform (please complete the following information)

- Unity Editor
- WMR immersive

thalbern:
![image](https://user-images.githubusercontent.com/36998103/62876079-145ec580-bd1c-11e9-8840-8dacc8fd05cc.png)
another screenshot from the bounding box example

cre8ivepark:
I also have experienced this when I pull/switch branch. Deleting the Library folder and restarting Unity resolved the issue. The issue is TextMesh Pro loses its text alignment property (center) and reverts back to default 'left' alignment. Fresh import of the MRTK release package also does not show this symptom.

julenka:
Closing as not repro based on Yoon's comments. If people follow upgrade instructions (delete Library folder first), they will not hit this.

cre8ivepark:
Found proper solution. Upgrading TextMesh Pro to 1.4.1 resolves the issue. 
![2019-08-12 15_24_38-](https://user-images.githubusercontent.com/13754172/62902791-8ed81700-bd15-11e9-84ee-fce17725e1ea.png)

cre8ivepark:
@julenka @thalbern Found the clear solution. Updated manifest file to use the latest version of TextMesh Pro. #5619

cre8ivepark:
Fixed with #5619

thalbern:
thanks for looking into it Yoon! 
(just because it was mentioned before, I didn't upgrade a package. i was just starting a project from scratch and importing our latest package)

PR #5619, opened 2019-08-12
Update TextMesh Pro version to 1.4.1 in the Package manifest file

cre8ivepark:
## Overview
Our Packages/manifest.json file has old TextMesh Pro version (1.3.0) which causes alignment issues on certain cases. Manually updating through Unity Package Manager resolves the issue.
Since there is no specific reason for keeping 1.3.0, updating it to the latest version 1.4.1.

## Changes
- Fixes: #5600


davidkline-ms:
azp run

Issue #5604, opened 2019-08-12
Nullptr in Overlap.SolverUpdate after Upgrade from RC2 to G1


julenka:
## Overview
After I updated my personal project from RC2 to GA, I am getting NullPtr every frame when I hit "Play" in my sample scene

## Repro steps
1. Unzip the project Assets folder located [here](https://microsoft-my.sharepoint.com/:u:/p/jschwarz/EXlokMtzbbxArfKNoRsVOO0BgFfKuW6hGDh3K0H68sUOUQ?e=kjyzpO), move it to a new test project
2. Open the project, then open the BouncyBall/SampleScene.unity scene, and then hit play

You will get a large number of errors like:

    NullReferenceException: Object reference not set to an instance of an object
    Microsoft.MixedReality.Toolkit.Utilities.Solvers.Overlap.SolverUpdate () (at Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Overlap.cs:13)
    Microsoft.MixedReality.Toolkit.Utilities.Solvers.Solver.SolverUpdateEntry () (at Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs:245)
    Microsoft.MixedReality.Toolkit.Utilities.Solvers.SolverHandler.LateUpdate () (at Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs:326)

This project contains my old project (all my files are in the BouncyBall folder), plus the new MRTK GA release (in the other MRTK folders). All MRTK files have been updated to their newer version, except PinchSlider.cs

## Expected Behavior
The expectation is to not get any errors.

## Unity Editor Version
2018.4.4f1

## Mixed Reality Toolkit Release Version
GA

julenka:
Assigning to Troy since he has been modifying solvers recently.

davidkline-ms:
I'll pitch in and investigate while @Troy-Ferrell works on #5572

PR #5622, opened 2019-08-12
Overlap does not check for SolverHandler.TransformTarget null

Troy-Ferrell:
## Overview
Overlap solver was not updated to check if TransformTarget could be null (i.e no hand is shown). This fix adds that check and adds SolverTest for this type of Solver

## Changes
- Fixes: #5604 

## Verification
All tests pass

>
>

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
@@ -289,6 +289,44 @@ namespace Microsoft.MixedReality.Toolkit.Tests
             Assert.LessOrEqual(Vector3.Distance(testObjects.target.transform.position, handPosition), HandDistanceThreshold, "HandConstraint solver is not within {0} units of the hand", HandDistanceThreshold);
         }
 
+        /// <summary>
+        /// Test the Overlap solver and make sure it tracks the left simulated hand exactly
+        /// </summary>
+        /// <returns></returns>
+        [UnityTest]
+        public IEnumerator TestOverlap()
+        {
+            // Instantiate our test gameobject with solver.
+            var testObjects = InstantiateTestSolver<Overlap>();
+            testObjects.handler.TrackedTargetType = TrackedObjectType.HandJoint;
+            var targetTransform = testObjects.target.transform;
+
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
```diff
@@ -289,6 +289,44 @@ public IEnumerator TestHandConstraint()
...
+            TestUtilities.AssertAboutEqual(targetTransform.rotation, Quaternion.identity, "Overlap not at original rotation");
+
+            // Test that the solver flies to the position of the left hand
+            var handPosition = Vector3.forward - Vector3.right;
+            var handRotation = Quaternion.LookRotation(handPosition);
+            var leftHand = new TestHand(Handedness.Left);
+            yield return leftHand.Show(handPosition);
+            yield return leftHand.SetRotation(handRotation);
+            
+            yield return WaitForFrames(2);
```
julenka: why 2?

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
```diff
@@ -289,6 +289,44 @@ public IEnumerator TestHandConstraint()
...
+            var hand = PlayModeTestUtilities.GetInputSimulationService().GetHandDevice(Handedness.Left);
+            Assert.IsNotNull(hand);
+            Assert.IsTrue(hand.TryGetJoint(TrackedHandJoint.Palm, out MixedRealityPose pose));
```
julenka: Why not just use handjointutils?

Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs
```diff
@@ -289,6 +289,44 @@ public IEnumerator TestHandConstraint()
...
+
+            TestUtilities.AssertAboutEqual(targetTransform.position, pose.Position, "Overlap solver is not at the same position as the left hand.");
+            Assert.IsTrue(Quaternion.Angle(targetTransform.rotation, pose.Rotation) < 2.0f);
+
+            // Make sure the solver did not move when hand was hidden
+            yield return leftHand.Hide();
+            yield return WaitForFrames(2);
```
julenka: why do we need to wait here?

PR #5624, opened 2019-08-13
TryUnregisterModuleFolder now returns on path deletion

Alexees:
## Overview
The MixedRealityToolkitFiles.cs TryUnregisterModuleFolder iterates a foreach loop and deletes entries within it, but instead of exiting it sets a bool to true which it returns at the end continuing to iterate the loop resulting in an Exception.

## Changes
Now returns on entry find.

## Verification
I recognized that when I moved the MRTK generated folders over to another one and then removed the old folder.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs
@@ -184,7 +184,7 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
         private static bool TryUnregisterModuleFolder(string folder)
         {
             string normalizedFolder = NormalizeSeparators(folder);
-            bool found = false;
+
             foreach (var modFolders in mrtkFolders)
             {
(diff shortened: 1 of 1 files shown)


wiwei:
FYI, I think with the change here:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5603/files

Going through all the mrtkFolders may actually be required - in practice, I don't think you'll actually see this really, but it might be good to keep this code as-is until the other PR goes in.

wiwei:
Okay, I think that with the new changes highlighted in that PR this is going to be necessary to handle NuGet consumption cases, even if it doesn't show up in non-nuget cases

wiwei:
If I hadn't made that other PR though, this would have been in a heartbeat

Alexees:
@wiwei do you mean #5603, cause if so, this did not fix the problem. The loop still alters itself:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/2367777b57dbdf6ee4b306aaf1c5860c378118a6/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityToolkitFiles.cs#L220-L230

Issue #5580, opened 2019-08-09
Gaze cursor shows up when hands are in near touch mode


sostel:
## Describe the bug
The gaze cursor shows up when the hands go into near mode. 

## To reproduce
1. Sync to the latest '2.0.0_stabilization' branch 
2. Start Unity and move the HandsInteractionAll scene to the first place in the Build Settings
3. Open the VS project and enable the "Gaze Input" capability
4. Compile and deploy to HoloLens 2

*Please note:*
This issue doesn't occur if the "Gaze Input" capability is disabled. However, in this case, once the hands are not detected anymore, the head gaze cursor will not show up automatically. Instead the user has to say "Select" first.

## Expected behavior
The gaze cursor should only show up if either "no hands" are detectable and/or "no hands" + an activation command (e.g., say select) is issued.

## Screenshots
![image](https://user-images.githubusercontent.com/32180789/62802461-cdaf6800-ba9c-11e9-9a93-9189198aca6d.png)

## Your Setup (please complete the following information)
- Unity 2018.4.0f1
- MRTK 2.0.0_stabilization

## Target Platform (please complete the following information)
HoloLens 2

julenka:
This might be my bug caused by showing / hiding cursors in the cursor state machine. @wiwei and @keveleigh are you guys planning to look at this Friday or Monday? I can help. I was under impression Kurtis and will were busy with other things.

wiwei:
@julenka I think your cursor state machine changes were pretty long ago though - I just tried pulling latest stabilization and wasn't able to hit this (first tried in editor just because it's faster, then tried deploying to device and also wasn't able to hit this).

I'm pinging some others now to see if they can see this on their devices.

sostel:
I just tested after syncing to the latest in the stabilization branch. I don't see the issue occur in the Editor, but I do see it on the device 🤔

sostel:
To add:
I've tested with the "Gaze Input" capability disabled and the cursor will _not_ show up automatically in that case. Instead the user has to say "select" to make it show up again. I've updated the description above.

cre8ivepark:
@sostel Confirmed. The gaze cursor shows up when 'Gaze Input' is checked in the Visual Studio.

cre8ivepark:
@sostel Isn't this expected behavior for ET? Checking 'Gaze Input' shows the gaze cursor and user have to change the cursor prefab to EyeGazeCursor to make it subtle?
[...]

PR #5627, opened 2019-08-13
Hide eye cursor in near touch mode

sostel:
## Overview
This addresses issue #5580: "Gaze cursor shows up when hands are in near touch mode". An additional check was added to only show the eye gaze cursor when no other near or far pointers are active.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
@@ -63,8 +63,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
 
         private void UpdateStateEyeGaze(int numNearPointersActive, int numFarPointersActive)
         {
-            // If there are any far pointers active while eye gaze is valid, then
-            // eye gaze should be disabled.
+            // Only enable eye gaze as a pointer if there are no other near or far pointers active.
             bool isEyeGazePointerActive = numFarPointersActive == 0 && numNearPointersActive == 0;
 
             gazePointerState = isEyeGazePointerActive ?
--- a/Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
@@ -65,7 +65,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
         {
             // If there are any far pointers active while eye gaze is valid, then
             // eye gaze should be disabled.
-            bool isEyeGazePointerActive = numFarPointersActive == 0;
+            bool isEyeGazePointerActive = numFarPointersActive == 0 && numNearPointersActive == 0;
 
             gazePointerState = isEyeGazePointerActive ?
                 GazePointerState.GazePointerActive :
--- a/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/GazePointerStateMachineTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/GazePointerStateMachineTests.cs
@@ -133,27 +133,26 @@ namespace Microsoft.MixedReality.Toolkit.Tests.InputSystem
             var gsm = new GazePointerVisibilityStateMachine();
             Assert.IsTrue(gsm.IsGazePointerActive, "Eye gaze pointer should be visible on start");
 
-            // With the hand raised, eye gaze pointer should still exist because only far interaction causes the 
-            // eye gaze pointer to go away.
+            // With one hand is raised and being in near touch mode, the eye gaze pointer should be disabled.
             gsm.UpdateState(1 /*numNearPointersActive*/, 0 /*numFarPointersActive*/, 0 /*numFarPointersWithoutCursorActive*/, true);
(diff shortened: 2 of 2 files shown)


Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
```diff
@@ -65,7 +65,7 @@ private void UpdateStateEyeGaze(int numNearPointersActive, int numFarPointersAct
         {
             // If there are any far pointers active while eye gaze is valid, then
```
wiwei: Nit: Also update the comment here stating that it checks for both near and far

sostel: Good catch! Will do ASAP.

wiwei: Whereeverrrrrrrrrrrrrrr you areeeeeeeeeeeeeeeeeeeeeeeeeeeeee

sostel: Ehm?

sostel:
> One more thing - please add another test line to:
> 
> TestHeadGazeHandAndSpeechBehaviour
> 
> https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Tests/EditModeTests/InputSystem/GazePointerStateMachineTests.cs
> 
> I think we should have some SOME test coverage of this, and since we already have test code, may as well update that

Oh, good pointer! Didn't even know that this existed. I saw that there is already a test for "TestEyeGazeHandAndSpeechBehaviour". I've just updated it.

sostel:
Alrighty, addressed all your comments, @wiwei! :) Anything else?

cre8ivepark:
Tested on HL2 with & without the 'Gaze Input' in the VS. It looks good.
I'll do test on HL1 and WMR immersive headset.

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
```diff
@@ -63,9 +63,8 @@ public void UpdateState(int numNearPointersActive, int numFarPointersActive, int
 
         private void UpdateStateEyeGaze(int numNearPointersActive, int numFarPointersActive)
         {
-            // If there are any far pointers active while eye gaze is valid, then
-            // eye gaze should be disabled.
-            bool isEyeGazePointerActive = numFarPointersActive == 0;
+            // Only enable eye gaze as a pointer if there are no other near or far pointers active.
+            bool isEyeGazePointerActive = numFarPointersActive == 0 && numNearPointersActive == 0;
```
keveleigh: It looks like one side effect of this would be that we can no longer use the eye gaze pointer in combination with near interaction, which I think I remember was a use case some of our partners wanted / were using.

sostel: In that case, they can use EyeTrackingTarget to support simultaneous use of eye gaze and hand input. https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/EyeTracking/EyeTracking_TargetSelection.html#2-independent-eye-gaze-specific-eyetrackingtarget

sostel: Alternatively we could simply hide the cursor, but then buttons would still be highlighted based on eye gaze, which may be confusing with hands highlighting other buttons.

cre8ivepark: Right... it would be better to make this optional. In the Periodic Table app too. The user can grab and manipulate the element while reading through the text panel with ET auto-scroll.

2019-08-13 11_42_55-Photos

sostel: ET auto scroll should already handle this, as this was supposed to work even with hand rays active. I might be wrong. Worth testing it again.

cre8ivepark: I see. I'll also test with the Periodic Table app

cre8ivepark: @sostel @keveleigh Verified. Eye-tracking scroll works with while manipulating objects with hands.

julenka:
Thank you sophie!

Assets/MixedRealityToolkit.Services/InputSystem/GazePointerVisibilityStateMachine.cs
sostel: Yay! :)

wiwei:
@sostel, I think there are some test failures that need to be patched up here.

sostel:
> @sostel, I think there are some test failures that need to be patched up here.

Sigh. Yes... I only updated "TestEyeGazeHandAndSpeechBehaviour". Turns out there are other tests for "TestEyeGazeToHeadGazeTransition" and "TestHeadGazeToEyeGazeTransition" 🙄 Let me check them.

sostel:
Yay! Done! Thanks everybody for you feedback! Thanks for the merge, @davidkline-ms!

Issue #5616, opened 2019-08-12
Buttons play sounds even when approached from behind


julenka:
## Overview
Buttons should only play touch down, up sounds, when they are actually being pressed! Since we cannot press buttons from the back, we should not hear sounds when approaching a button from the back

## Repro steps
Approach a button from behind using your finger
Move finger through collider
Notice that a pressed and released sound plays

## Expected Behavior
The pressed and released sounds should only play when the button is actually compressed. If the button state and visual doesn't respond to the finger movement, then a sound should not be playing.

Otherwise, the user will be confused because they are hearing feedback that the button is responding to them, when it is actually not responding. 

## Mixed Reality Toolkit Release Version
GA

cre8ivepark:
I think this is because I have moved the audio clips from PressableButton to Interactable to support audio feedback on both near and far interaction. #5577 
PressableButton has the mechanism to detect press direction. However, I think Interactable doesn't.

PR #5628, opened 2019-08-13
added filter for interactable press receiver and updated hololens 2 b…

thalbern:
## Overview
Interactable press receiver can now filter for far, near, or both event types. This allows us to properly configure buttons so they play sounds only from the front.

Also includes a fix for our InspectorField serialization whenever we're serializing newly added fields that aren't part of existing data/assets.

## Changes
- Fixes: #5616, #5647

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnPressReceiver.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnPressReceiver.cs
@@ -16,15 +16,15 @@ namespace Microsoft.MixedReality.Toolkit.UI
         [InspectorField(Type = InspectorField.FieldTypes.Event, Label = "On Release", Tooltip = "The button is released")]
         public UnityEvent OnRelease = new UnityEvent();
 
-        public enum PressType
+        public enum InteractionType
         {
             NearAndFar = 0,
             NearOnly = 1,
             FarOnly = 2
         }
 
-        [InspectorField(Label = "Press Interaction Type", Tooltip = "Specify whether press event is for near or far interaction", Type = InspectorField.FieldTypes.DropdownInt, Options = new string[] { "Near and Far", "Near Only", "Far Only" })]
-        public int PressTypeFilter = (int)PressType.NearAndFar;
+        [InspectorField(Label = "Interaction Filter", Tooltip = "Specify whether press event is for near or far interaction", Type = InspectorField.FieldTypes.DropdownInt, Options = new string[] { "Near and Far", "Near Only", "Far Only" })]
+        public int InteractionFilter = (int)InteractionType.NearAndFar;
 
         private bool hasDown;
         private State lastState;
@@ -42,8 +42,8 @@ namespace Microsoft.MixedReality.Toolkit.UI
         /// <returns>true if interactable state matches filter</returns>
         private bool IsFilterValid()
         {
-            if (PressTypeFilter == (int)PressType.FarOnly && isNear
-                || PressTypeFilter == (int)PressType.NearOnly && !isNear)
+            if (InteractionFilter == (int)InteractionType.FarOnly && isNear
+                || InteractionFilter == (int)InteractionType.NearOnly && !isNear)
             {
                 return false;
             }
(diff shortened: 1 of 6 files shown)


cre8ivepark:
Tested on HL2 device with 'Far only' filter option in the Interactable. It works well!
Also updated the blue Round button prefab as well.

2019-08-13 16_05_21-Unity 2018 4 5f1 Personal - PressableButtonExample unity - MRTK-Public-Microsoft

cre8ivepark:
Also tested on HL1 and WMR headset.

julenka:
I'm okay with this, but it is not ideal. Would love to hear from @Troy-Ferrell @killerantz  or @Railboy  if we could have a better approach for serializing these fields.

julenka:
I reset my feedback since I haven't tested thoroughly enough. @cre8ivepark  or @thalbern  are you able to test and verify on HL2? We should do that before checking in.

Troy-Ferrell:
> I'm okay with this, but it is not ideal. Would love to hear from @Troy-Ferrell @killerantz  or @Railboy  if we could have a better approach for serializing these fields.

What do you mean serializing these fields?

cre8ivepark:
@julenka I'll do testing for HL2 & HL1

cre8ivepark:
Tested on HL2 and HL1. Looks good!

thalbern:
did another test on HL2 and looks good to me

thalbern:
upgrade tests failed for me. using existing assets and updating to new code will break the existing pressreceivers - they need to be  manually removed and readded

thalbern:
did several other upgrade scenarios which all failed for me. don't merge this change!

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnPressReceiver.cs
```diff
@@ -11,33 +12,71 @@ namespace Microsoft.MixedReality.Toolkit.UI
...
         public UnityEvent OnRelease = new UnityEvent();
 
+        public enum PressType
+        {
+            NearAndFar = 0,
+            NearOnly = 1,
+            FarOnly = 2
+        }
+
+        [InspectorField(Label = "Press Interaction Type", Tooltip = "Specify whether press event is for near or far interaction", Type = InspectorField.FieldTypes.DropdownInt, Options = new string[] { "Near and Far", "Near Only", "Far Only" })]
```
julenka: I am not a fan of the name "Press Interaction Type" Suggestions:

* Near / Far Interaction Filter
* Near or Far Interaction?
* Interaction Filter
* Interaction Distance Filter
* Event Filter Type
* Event Filter

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnPressReceiver.cs
```diff
@@ -11,33 +12,71 @@ namespace Microsoft.MixedReality.Toolkit.UI
...
+        public int PressTypeFilter = (int)PressType.NearAndFar;
```
julenka: @Troy-Ferrell and @Railboy  berni + i couldn't find a way to actually use an enum for the public field, it looks like the custom inspector only supports int lists. Do you have ideas / suggestions for how to use an enum here?

Assets/MixedRealityToolkit/Utilities/InspectorFields/InspectorGenericFields.cs
```diff
@@ -31,7 +31,10 @@ public static void LoadSettings(T target, List<InspectorPropertySetting> setting
                 foreach (var attr in attrs)
                 {
                     object value = InspectorField.GetSettingValue(settings, propInfo.Name);
-                    propInfo.SetValue(target, value);
+                    if (value != null)
```
julenka: what happens in the null case? Does a default value fill in?

thalbern: afais it's just taking whatever default is defined in the class - but i'll check again if that's really the case

julenka:
Verified that I can have two receivers: a near and a far, and do different things based on whether I pressed near and far:

image

Assets/MixedRealityToolkit/Utilities/InspectorFields/InspectorGenericFields.cs
thalbern: yeah confirmed. it's getting the defaults from the class and this is just overwriting whatever default value is already in there

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnPressReceiver.cs
thalbern: I vote for "Interaction Filter"

thalbern:
did some more testing on HL2 and upgrade tests with old buttons - looks good to me now.

cre8ivepark:
Doing upgrade testing with Periodic Table app now.

thalbern:
> Finished upgrade test with Periodic Table app. Looks good!

thanks Yoon! :)

PR #5630, opened 2019-08-13
Fix build warnings

keveleigh:
## Overview
image

Some of our build steps were generating warnings. This fixes them! And some formatting.

For some reason, these didn't show up in our "normal" build steps, just in some of the later steps.


wiwei:
>                 keyboardText = touchscreenKeyboard.text;

I think these also need to be updated to:

KeyboardText

(Unless there is some base class that I'm missing)

---
Refers to: Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Script/SystemKeyboardExample.cs:80 in dcb59a5. [](commit_id = dcb59a501e644ac9bb60387220aa1e68f7cd791c, deletion_comment = False)

keveleigh:
@wiwei ooh yeah you're right.

PR #5631, opened 2019-08-13
Fix #5626: Discover and set virtual reality SDK(s) for the target platform

### LINKED ISSUE #5626: Apply Default Settings does not add the SDK in Unity 2019
What appears to be a behavior change in Unity 2019 is causing the XR Settings SDK to not be correctly selected when using the "Apply Mixed Reality Default Settings" message box.

Unity 2018.4 does not experience this issue.

davidkline-ms:
This change asks Unity for the set of supported virtual reality SDKs for the currently selected target platform and then explicitly sets those that are supported by MRTK v2.

Unity 2018 set default SDKs, 2019 appers to no longer do so.

Fixes: #5626 

Verification:
* Clean Unity 2018 and 2019 projects
* Import private test foundation package
* Apply settings via dialog
* Confirm correct settings
* Switch platform (to uwp or to sdk based on original setting)
* Reload project
* Apply settings via dialog
* Confirm correct settings

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
@@ -158,6 +158,9 @@ namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
             }
         }
 
+        /// <summary>
+        /// Discover and set the appropriate XR Settings for the current build target.
+        /// </summary>
         private static void ApplyXRSettings()
         {
             BuildTargetGroup targetGroup = EditorUserBuildSettings.selectedBuildTargetGroup;
--- a/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
+++ b/Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using Boo.Lang;
 using Microsoft.MixedReality.Toolkit.Editor;
 using System;
 using System.IO;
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using Boo.Lang;
```
wiwei: I'm going to add a script to make sure we don't have any instances of Boo.Lang anymore.

As much as I love Boo:

image

wiwei: This should probably be System.Collections

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
```diff
@@ -124,7 +125,7 @@ private static void ShowSettingsDialog()
                         case 0:
                             EditorSettings.serializationMode = SerializationMode.ForceText;
                             EditorSettings.externalVersionControl = "Visible Meta Files";
-                            PlayerSettings.virtualRealitySupported = true;
```
keveleigh: Is this now superseded by `PlayerSettings.SetVirtualRealitySupported`?

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
davidkline-ms: agreed. i'll fix

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
davidkline-ms: yes. i am not calling SetVirtualRealitySupported. i am hoping that a SetStereoRenderMode gets added so i can set all supported platforms at once (dont make the customer reopen to see the dialog)

Assets/MixedRealityToolkit/Utilities/Editor/Setup/MixedRealityEditorSettings.cs
wiwei: https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5632

PR #5632, opened 2019-08-13
Add a code validation step to the PR pipeline.

wiwei:
We've seen a couple of instances of Boo.Lang getting checked in, where we really should be using System.Collections. More generally, there can be other patterns that we may check in inadvertently that we should avoid going forward, and having this be checked by a linter/during the checkin pipeline is better than assuming that all PR will catch it.

### DIFF
--- a/scripts/ci/validatecode.ps1
+++ b/scripts/ci/validatecode.ps1
@@ -16,6 +16,7 @@
     .\validatecode.ps1 -Directory c:\path\to\MRTK\Assets
 #>
 param(
+    [Parameter(Mandatory=$true)]
     [string]$Directory
 )
 
@@ -35,10 +36,6 @@ function CheckBooLang(
         return $true;
     }
     return $false
-}
-
-if (-not $Directory) {
-    throw "-Directory is a required flag"
 }
 
 function CheckFile(
--- a/scripts/ci/validatecode.ps1
+++ b/scripts/ci/validatecode.ps1
@@ -1,0 +1,78 @@
+<#
+.SYNOPSIS
+    Validates the code to check for common patterns and usage that shouldn't be
+    checked in.
+.DESCRIPTION
+    This currently checks:
+
+    - That Boo.Lang isn't used anywhere in the code. This is an autocomplete option
+      that occurs when using Lists and other collections - the right thing to do
+      is to use System.Collections
+
+    Returns 0 if there are no issues, non-zero if there are.
+.PARAMETER Directory
+    The directory containing the code to validate. Only CSharp files
+.EXAMPLE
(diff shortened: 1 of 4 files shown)


wiwei:
Kicking off a build which should fail because there is a usage of Boo.Lang in the code today.

wiwei:
It worked! https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=4250

scripts/ci/validatecode.ps1
```diff
@@ -0,0 +1,78 @@
...
+    #>
+    if ($FileContent[$LineNumber] -match "^using Boo\.Lang;") {
+        Write-Host "An instance of Boo.Lang was found in $FileName at line $LineNumber "
+        Write-Host "Use System.Collections instead."
+        return $true;
+    }
+    return $false
+}
+
+if (-not $Directory) {
```
MenelvagorMilsom: Can you not make this a mandatory parameter?
``` PowerShell
[Parameter(Mandatory=$true)]
[string]$Directory
```

wiwei: Ah, didn't know, thanks for the pointer

scripts/ci/validatecode.ps1
```diff
@@ -0,0 +1,75 @@
...
+        if (CheckBooLang $FileName $fileContent $i) {
+            $containsIssue = $true
+        }
+    }
+    return $containsIssue
+}
+
+Write-Output "Checking $Directory for common code issues"
+
+$codeFiles = Get-ChildItem $Directory *.cs -Recurse | Select-Object FullName
```
julenka: Did you consider using [Select-String](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/select-string?view=powershell-6)?

something like:

```
$matches = gci -r *.cs | sls -Pattern "^using Boo\.Lang;"

if ( -not $matches -eq $null )
{
    Write-Output "Issues found, please see above for details"
    exit 1;
}
else {
    Write-Output "No issues found"
    exit 0;
}
```

You could even be smarter and exit early.

wiwei: I thought about doing something like this, but specifically:

1) I think I'll want to check for multiple things, so having a system that gives different handlers the ability to check the same lines (without re-searching through each file) is valuable. Obviously I'm not using that capability right now so it's a bit weird for me to say this, but I plan on augmenting this in the future and would like to do a single pass per-file

2) Exitting early isn't great because we should collect ALL issues at once, so that people can run the thing once and then get all of the problems, instead of finding one problem, fixing it, running it again, and so on.

Issue #4953, opened 2019-06-18
Need docs for how to configure pointer profiles


julenka:
## Overview
Provide documentation on [profiles README](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Profiles/Profiles.html) for how to configure pointer profiles for the following common use cases:

1. Use GGV on HoloLens 2.
2. Use only the poke pointer, no hand rays (turn hand rays off)

Troy-Ferrell:
For issue 2 of this bug, that should be tracked by #4954

PR #5633, opened 2019-08-13
Clean up and simplify pointer documentation

Troy-Ferrell:
## Overview
This change cleans and updates much of the Pointer documentation. Originally there were three different pages for pointers and each had overlapping but slightly different content. This organizes the content into Pointer Architecture and Pointers in Practice. 

- Added links to pointer classes on docs via xref
- Add code snippets for how to do key things
- Fixed some image resizing
- Fixed some formating, line spacing, etc.
- Added additional information on how to configure pointers via profiles
- Added summary tag and tooltips to some code
- Deprecates README_Pointer.md, removal from TOC and updates on getting started/readme pages

## Changes
- Fixes: #4953 

## Verification

>
>


cre8ivepark:
I have updated some old images and added an overview image for the pointers. Please let me know if you see anything need to be updated.
MRTK_Pointers

Troy-Ferrell:
TouchPoint

> I have updated some old images and added an overview image for the pointers. Please let me know if you see anything need to be updated.

Awesome! Few thoughts.

1) Pointer mediator is actually a bit tricky. I was reading into the code more and the PointerMediator is created per input source. So it's weird to show the pointer mediator connecting all the pointers when you wouldn't have pokepointer with parabolic pointer on the same input source.

2) TouchPointer is for touchscreens yes? The hand image suggests articulated hand. I'd rather show like a phone icon or surface with a hand etc

cre8ivepark:
Updated version:
MRTK_Pointers

Documentation/Input/Pointers.md
```diff
@@ -1,39 +1,239 @@
...
-Event | Description | Handler
---- | --- | ---
-Before Focus Changed / Focus Changed | Raised on both the game object losing focus and the one gaining it every time a pointer changes focus. | [`IMixedRealityFocusChangedHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityFocusChangedHandler)
+- *Pointing Extent* - Defines the max distance for which a Pointer can interact with a GameObject.
+
+- *Pointing Raycast Layer Masks* - This is a prioritized array of LayerMasks to determine what possible GameObjects any given Pointer can interact and the order of interaction to attempt. This is may be useful to ensure Pointers interact with UI elements first before other scene objects.
+
+![Pointer Profile Example](../Images/Input/Pointers/PointerProfile.png)
+
+The default MRTK Pointer Profile configuration includes the following pointer classes and associated prefabs out-of-box. The list of pointers available to the system at runtime is defined under *Pointer Options* in the Pointer profile. Developers can utilize this list to reconfigure existing Pointers, add new Pointers, or delete one.
```
wiwei: >following [](start = 60, length = 9)

This instance of the word following feels like it's a bit out of place, because the text that follows this doesn't contain a list of pointer classes of associated prefabs - maybe this should be "preceeding" because it's referring to the content in the picture?

Documentation/Input/Pointers.md
```diff
@@ -1,39 +1,239 @@
...
+The following classes are the out-of-box MRTK pointers available and defined in the default *MRTK Pointer Profile* outlined above. Each pointer prefab provided under *Assets/MixedRealityToolkit.SDK/Features/UX/Prefabs/Pointers* contains one of these pointer components attached.
+
+![MRTK Default Pointers](../Images/Input/Pointers/MRTK_Pointers.png)
+
+
+#### Far pointers
+
+##### [`LinePointer`](xref:Microsoft.MixedReality.Toolkit.Input.MousePointer)
+
+ A base pointer class, draws lines from the source of the input (i.e. the controller) in the pointer direction. This is also used by other things like the Shell Hand Ray Pointer and the teleport pointers (which also draw lines to indicate where teleportation will end up at).
```
wiwei: It might be worth adding a note that this pointer itself isn't directly instantiated, and is only used through subclassing of the shell hand ray pointer and teleport pointers.

The reason I mention this is because the next line contains additional content (i.e. rotation stuff) which makes me think that actually this is more useful itself than it is (i.e. DefaultControllerPointer isn't mentioned here, which is actually the main one that is actually referenced in the profile.)

Documentation/Input/Pointers.md
```diff
@@ -1,39 +1,239 @@
...
+
+        private void IMixedRealityPointerHandler.OnPointerClicked(MixedRealityPointerEventData eventData) 
+        {
+            material.color = color_OnSelect;
+        }
+```
+
+#### Register globally
+
+blah need to add code example for CoreServices.InputSystem.Register etc
```
wiwei: It feels like this section should probably just point to a general "how to register for input events" instead of having to necessarily copy paste and have another define of that... so maybe consider having a reference to somewhere else (which I don't think exists FWIW).

Documentation/Input/Pointers.md
```diff
@@ -1,39 +1,239 @@
...
-## Pointer Result
+    private void OnPrimaryPointerChanged(IMixedRealityPointer oldPointer, IMixedRealityPointer newPointer)
+    {
+        ...
+    }
+
+    private void OnDisable()
+    {
+        CoreServices.InputSystem?.FocusProvider?.UnsubscribeFromPrimaryPointerChanged(OnPrimaryPointerChanged);
+        OnPrimaryPointerChanged(null, null);
```
wiwei: >OnPrimaryPointerChanged(null, null); [](start = 8, length = 36)

Nit: Add a comment stating that this is designed to flush out the current primary pointer

Documentation/Input/Pointers.md
Troy-Ferrell: Gah, and it's totally wrong reference

Troy-Ferrell: Also DefaultControllerPointer is only the prefab which is mentioned earlier on and is mentioned with ShellHandRayPointer

Documentation/Input/Pointers.md
Troy-Ferrell: Move image back up to make sense

Documentation/Input/Pointers.md
Troy-Ferrell: Ah I see what you are talking about now

Troy-Ferrell:
> I found that some of the new images has .PNG instead of .png which will be broken on HTML version. I tried to change the file extensions to lower case but git did not detect them as changed file. Could you please check? Or we might have to update the .md file to use proper upper case.
> 
> Broken image example:
> https://github.com/microsoft/MixedRealityToolkit-Unity/blob/cd86fbf5d08a179982383b128e67e3a394f79c26/Documentation/Input/Pointers.md

Got the fix in!

Documentation/Input/Pointers.md
```diff
@@ -1,39 +1,235 @@
...
+##### [`TouchPointer`](xref:Microsoft.MixedReality.Toolkit.Input.TouchPointer)
+
+The TouchPointer is responsible for working with Unity Touch input (i.e. touchscreen). These are 'far interactions' because the act of touching the screen will cast a ray from the camera to a potentially far location in the scene.
+
+##### [`MousePointer`](xref:Microsoft.MixedReality.Toolkit.Input.MousePointer)
+
+The MousePointer powers a screen to world raycast for far interactions, but for mouse instead of touch.
+
+<img src="../../Documentation/Images/Pointers/MRTK_MousePointer.png" width="400">
+
```
thalbern: ` [!NOTE] Mouse support is not available by default in MRTK and can be enabled by adding a new *Input Data Provider* of type @Microsoft.MixedReality.Toolkit.Input.UnityInput.MouseDeviceManager to the MRTK input profile.`

Documentation/README_Pointers.md
```diff
@@ -1,76 +1,5 @@
-# Pointers #
+# Pointer #
```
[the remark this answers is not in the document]
Troy-Ferrell: Because there may still be links out there referencing it. It's a pretty old file. Thus it's basically in a deprecated state and taken off the TOC until the next iteration. Just like all of the [Obsolete] tags

thalbern:
doc build failed : (Documentation/Input/Pointers.md)Text in the first line of Note/Section/Video is not valid. Will be rendererd to <blockquote>

Issue #5634, opened 2019-08-13
Interactable - "Remove Event" button does not work, NullPtr


julenka:
## Overview
You should be able to remove events using the "remove event" button in interactable. Unfortuantely, the button does not work for removing all but the first event, and instead throws large number of exceptions.

![image](https://user-images.githubusercontent.com/168492/62977286-70385580-bdd3-11e9-93d0-3de785fc750b.png)

## Steps to reproduce
1. Open any interactable, and under "event", add a second event
2. Try to remove that second event

## Expected Behavior
I can remove the event!

## Actual Behavior
Oh no! I cannot actually remove the second event.

Troy-Ferrell:
I fixed a similar issue with removing themes in the inspector code so curious if it is the same problem just different place....looking now

PR #5636, opened 2019-08-13
Fix inspector event removal bug

Troy-Ferrell:
## Overview
When removing an interactable event receiver item, the for loop rendering will still proceed thinking the number of items in the list has not changed. Thus resulting in out of bounds index access. 

Fix now breaks out of rendering list till next redraw which happens immediately with changed object.

## Changes
- Fixes: #5634 

## Verification
One can now add and remove interactable event receivers 

>
>

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
+++ b/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/InteractableInspector.cs
@@ -322,7 +322,11 @@ namespace Microsoft.MixedReality.Toolkit.UI.Editor
                 for (int i = 0; i < events.arraySize; i++)
                 {
                     SerializedProperty eventItem = events.GetArrayElementAtIndex(i);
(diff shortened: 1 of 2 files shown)


julenka:
There's something else I noticed: the InteractableReceiverListInspector.RemoveEvent is never called, instead InteractableInspector.RemoveEvent is called. Additionally, the methods are just duplicates of one another. Can we clean that up here? It was super confusing debugging that code...

Troy-Ferrell:
> There's something else I noticed: the InteractableReceiverListInspector.RemoveEvent is never called, instead InteractableInspector.RemoveEvent is called. Additionally, the methods are just duplicates of one another. Can we clean that up here? It was super confusing debugging that code...

Actually they are different. The serializedObject is different in each class and the RenderEventSettings is a static function. Hence why the parameter is passed to determine which private function to call. 

This setup can definitely be improved and better compenitized possibly. But I was looking to do the minimum change since this is stabilization

Issue #5625, opened 2019-08-13
Spatial Observer Volume ignores MRTK Playspace transform


fast-slow-still:
## Describe the bug

This one is my bad, I missed it when fixing up other SpatialMesh issues with Playspace transform.

When the WindowsMixedRealitySpatialMeshObserver sets the ObserverVolume, it incorrectly accounts for the MRTK Playspace transform (and any other transform on the camera hierarchy).

The tangible outcome is that when there is a non-identity transform on the Playspace, the Spatial Meshes will no longer be centered around the camera

## To reproduce

1. Set a large translation on the playspace (e.g. teleport a few meters).
2. Notice that the spatial meshes are centered around where you were, rather than where you are.

## Expected behavior

The spatial meshes to recenter around the viewer as the viewer teleports around.

## Additional context

There are two conflicting approaches to this problem. It is important to note that, whatever else, the volume passed into observer.SetVolumeXXX(volume) is expected to be in the same space as the head pose. That is, in the space of the camera's parent.

Approach 1) Treat WindowsMixedRealitySpatialMeshObserver.ObserverOrigin as being in the same space as the head pose (i.e. already in the camera's parent's space). In that case, the error is in UpdateObserver(), where ObserverOrigin should be CameraCache.Main.transform.**localPosition**, rather than **position**.

Approach 2) Treat ObserverOrigin as being in world space (Unity's root coordinate system). In that case, UpdateObserver() is correct, but ConfigureObserverVolume() needs to transform the world space ObserverOrigin into the camera's parent's space before passing it to Observer.SetVolumeXXXX(). This would presumably use CameraCache.Main.transform.parent.InverseTransformPoint(ObserverOrigin), with appropriate null check(s).

Since ObserverOrigin is publicly settable, it is probably expected that it is in world space, but the expected space isn't actually explicitly stated, so I might be wrong there. But if it is generally in world space, then Approach 2) would be preferred. Otherwise, Approach 1) is a simpler fix.

Note that everything here also applies to ObserverRotation.

wiwei:
I'm guessing this is referring to:

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/4474/files

fast-slow-still:
Yes, I fixed the mesh placement there, but spaced and forgot to correct the observer volume.

davidkline-ms:
> Yes, I fixed the mesh placement there, but spaced and forgot to correct the observer volume.

[...]

PR #5638, opened 2019-08-13
Fix 5625 - fix observer volume

davidkline-ms:
- Document that the observer origin is in world space.
- Update the observer volume update to properly transform the coordinate space.

@fast-slow-still, can you please validate this change (review the code and test your scenario)?

Fixes: #5625

### DIFF
--- a/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
+++ b/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealitySpatialMeshObserver.cs
@@ -679,20 +679,32 @@ namespace Microsoft.MixedReality.Toolkit.WindowsMixedReality.SpatialAwareness
         /// </summary>
         private void ConfigureObserverVolume()
         {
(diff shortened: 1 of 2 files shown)


davidkline-ms:
@fast-slow-still,

There is a test failure that appears to be related to this change. It feels like the transform is causing the meshes to not be placed correctly... I will add you as a collaborator to my fork. please feel free to push any changes.

fast-slow-still:
It looks like the test failure was a red herring. After running the failing test both in isolation and in the full suite several times locally without error, I re-ran the tests on CI and all three passed. That was before any code changes.

I did push a few minor tweaks. Mostly I clarified the distinction between camera space and the camera's parent's space, and took advantage of the camera's parent being known to be the MixedRealityPlayspace.

I also added the rotation of the ObserverVolume's orientation from world to Playspace.

These are all pretty minor, so if you want to rollback my push, I'm fine with that. My situation was already fixed by your changes.

I do want to stress that this only affects the region of the world in which spatial meshes are generated, but does not affect the positioning of meshes.

davidkline-ms:
Thanks @fast-slow-still! I love your changes.

Issue #5640, opened 2019-08-13
[HL1] Manipulation Handler - Object moves with head movement


cre8ivepark:
## Describe the bug
In HoloLens 1, when you grab and move an object with ManipulationHandler.
Hololens 2 and WMR immersive headset do not show this symptom.

- The object moves with hand (correct behavior)
- The object also moves with the head (should not move)
![MRTK_ManipulationHL1](https://user-images.githubusercontent.com/13754172/62984557-869cdc00-bde8-11e9-906e-8a2c4e609317.gif)
![MRTK_ManipulationHL1a](https://user-images.githubusercontent.com/13754172/62986445-7f79cc00-bdf0-11e9-8817-c88049b4a82a.gif)

## To reproduce
1. Go to HandInteractionExamples scene.
2. Gaze at any movable object that has ManipulationHandler.
3. Grab and move your head

## Expected behavior
The object should not move with your head. Verified with WMR immersive headset: it only moves with the motion controller.

## Branch
prerelease/stabilization_2.0.0

PR #5645, opened 2019-08-14
Change GGVPointer rotation to closer match shell

MenelvagorMilsom:
## Overview
- With articulated hands and 6dof controllers the pointer rotation is the hand/controller's rotation. This is used in `TwoHandMoveLogic` (which is also used for one hand) so that the object rotates around the pointer as it rotates:
image
- For GGV hands, the pointer rotation was the `InternalGazePointer` rotation, which meant that as you rotated your head, using the logic above, the object would rotate around your hand:
image
- These changes instead create a rotation out of the vector from the head to the hand, which rotates as you rotate your hand around your body but stable is if you only turn your head.
image

## Changes
- Fixes: #5640.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
@@ -194,6 +194,14 @@ namespace Microsoft.MixedReality.Toolkit.Input
         {
             get
             {
+                // Previously we were simply returning the InternalGazeProvider rotation here.
+                // This caused issues when the head rotated, but the hand stayed where it was.
+                // Now we're returning a rotation based on the vector from the camera position
+                // to the hand. This rotation is not affected by rotating your head.
+                //
+                // The y value is set to 0 here as we want the rotation to be about the y axis.
+                // Without this, one-hand manipulating an object would give it unwanted x/z 
+                // rotations as you move your hand up and down.
                 Vector3 look = Position - CameraCache.Main.transform.position;
                 look.y = 0;
                 return Quaternion.LookRotation(look);
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
@@ -190,7 +190,15 @@ namespace Microsoft.MixedReality.Toolkit.Input
         public virtual Vector3 Position => sourcePosition;
 
         /// <inheritdoc />
-        public virtual Quaternion Rotation => Quaternion.LookRotation(gazeProvider.GazePointer.Rays[0].Direction);
+        public virtual Quaternion Rotation
+        {
+            get
+            {
+                Vector3 look = Position - CameraCache.Main.transform.position;
+                look.y = 0;
+                return Quaternion.LookRotation(look);
+            }
+        }
 
         #endregion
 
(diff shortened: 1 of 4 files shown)


Assets/MixedRealityToolkit.Tests/PlayModeTests/BoundingBoxTests.cs
```diff
@@ -228,7 +228,7 @@ public IEnumerator ScaleViaHoloLens1Interaction()
 
             var endBounds = bbox.GetComponent<BoxCollider>().bounds;
             TestUtilities.AssertAboutEqual(endBounds.center, new Vector3(0.033f, 0.033f, 1.467f), "endBounds incorrect center");
-            TestUtilities.AssertAboutEqual(endBounds.size, Vector3.one * .567f, "endBounds incorrect size");
+            TestUtilities.AssertAboutEqual(endBounds.size, Vector3.one * .561f, "endBounds incorrect size");
```
julenka: Heh. Maybe we should increase the tolerance here.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/GGVPointer.cs
```diff
@@ -190,7 +190,15 @@ public void OnPreSceneQuery()
         public virtual Vector3 Position => sourcePosition;
 
         /// <inheritdoc />
-        public virtual Quaternion Rotation => Quaternion.LookRotation(gazeProvider.GazePointer.Rays[0].Direction);
+        public virtual Quaternion Rotation
+        {
+            get
+            {
+                Vector3 look = Position - CameraCache.Main.transform.position;
+                look.y = 0;
```
wiwei: Nit: Can we also add a comment about the intent here (i.e. otherwise it's future someone asking why Y specifically is being zeroed out)

wiwei:
Looks good! I think it would be good to have a test case for this as well, so that we don't regress this going forward.

julenka:
> code looks good but I did not test on device. I'm currently helping with another GA blocker. Could somebody, perhaps @cre8ivepark help out and do a test on HL1?

I freed up, so validating on HL1.

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
```diff
@@ -800,6 +800,60 @@ public IEnumerator ManipulationHandlerMinMaxScale()
...
+            // Rotate Head and readjust hand
+            int numRotations = 10;
+            for (int i = 0; i < numRotations; i++)
+            {
+                MixedRealityPlayspace.Transform.Rotate(Vector3.up, 180 / numRotations);
+                yield return hand.MoveTo(originalHandPosition, numHandSteps);
+                yield return null;
+
+                // Test Object hasn't moved
+                TestUtilities.AssertAboutEqual(initialObjectPosition, testObject.transform.position, "Object moved while rotating head");
```
wiwei: Is it also possible to assert that the rotation hasn't changed as well?

MenelvagorMilsom: Can do

MenelvagorMilsom:
> Tested on HoloLens 1. Head movement issue is fixed.
> This might be a separate issue: when you grab and move, the amount of vertical movement is smaller than horizontal movement. (I haven't tested this on HL2) It is much easier to move horizontally.

@cre8ivepark I noticed this too, but I checked and it appeared to be the case prior to these changes too. Let me know if you find otherwise.

Assets/MixedRealityToolkit.Tests/PlayModeTests/ManipulationHandlerTests.cs
MenelvagorMilsom: I had to add a 0.25° tolerance for the test to pass. I figured that's an acceptable level of error.

PR #5579, opened 2019-08-09
Add tests for data provider registeration

Troy-Ferrell:
## Overview

Data provider registration has changed dramatically in the stabilization branch. As per desires for original design, ownership of dataprovider registeration has moved entirely to the particular MixedRealityService utilizing that DataProvider. These tests are added to get coverage on this paradigm.

## Changes
- Fixes: #5563 

## Verification
All tests pass

> As a reviewer, it is possible to check out this change locally by using the following
> commands (substituting {PR_ID} with the ID of this pull request):
>
> git fetch origin pull/{PR_ID}/head:name_of_local_branch
>
> git checkout name_of_local_branch

PR #5648, opened 2019-08-14
Fix 5641 - SceneSystem defined as a data provider

davidkline-ms:
IMixedRealitySceneSystem was defines as both a CoreSytem (inherits BaseCoreSystem) and an IMixedRealityDataProvider.

This is incorrect and was preventing the scene service from being loadable.

### DIFF
--- a/Assets/MixedRealityToolkit.Services/SceneSystem/MixedRealitySceneSystem.cs
+++ b/Assets/MixedRealityToolkit.Services/SceneSystem/MixedRealitySceneSystem.cs
@@ -19,7 +19,7 @@ namespace Microsoft.MixedReality.Toolkit.SceneSystem
     /// This part handles the runtime parts of the service.
     /// </summary>
     [HelpURL("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/SceneSystem/SceneSystemGettingStarted.html")]
-    public partial class MixedRealitySceneSystem : BaseCoreSystem, IMixedRealitySceneSystem
+    public partial class MixedRealitySceneSystem : BaseCoreSystem
     {
         /// <summary>
         /// Async load operation progress amount indicating that we're ready to activate a scene.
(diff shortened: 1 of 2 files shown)


Assets/MixedRealityToolkit.Services/SceneSystem/MixedRealitySceneSystem.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.MixedReality.Toolkit.SceneSystem
     /// This part handles the runtime parts of the service.
     /// </summary>
     [HelpURL("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/SceneSystem/SceneSystemGettingStarted.html")]
-    public partial class MixedRealitySceneSystem : BaseCoreSystem, IMixedRealitySceneSystem
+    public partial class MixedRealitySceneSystem : BaseCoreSystem
```
keveleigh: Shouldn't this stay the same?

```suggestion
    public partial class MixedRealitySceneSystem : BaseCoreSystem, IMixedRealitySceneSystem
```

davidkline-ms: Unneeded. BaseCoreSystem already specifies IMixedRealityService

davidkline-ms: Good catch! My bad reading.. :D

wiwei:
We should absolutely have a test for this so that things don't error like this going forward.

davidkline-ms:
> We should absolutely have a test for this so that things don't error like this going forward.

Agreed. This will need to be done for _all_ core services and MRTK team provided extension services.

Troy-Ferrell:
> We should absolutely have a test for this so that things don't error like this going forward.
> 
> Agreed. This will need to be done for all core services and MRTK team provided extension services.
> Not sure how to make a catch-all for this though

I can look at possibly including with #5579

PR #5649, opened 2019-08-14
Merge stabilization to dev

keveleigh:
## Overview
Periodic merge of stabilization to dev


wiwei:
Just to check - is this coming from keveleigh:merge-stabilization because there were some merge conflicts that had to be resolved?

It's a bit unorthodox to see a merge from not-stabilization

keveleigh:
@wiwei Correct.

https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5502 and #5630 both edited SystemKeyboardExample.

The null checks in #5502 probably should have been added to stabilization.

Issue #5363, opened 2019-07-18
On HL1 and Oasis, keyboard cannot be re-opened after it has been closed using HandInteractionExample "Show Keyboard" button


Alexees:
## Describe the bug

Since the TouchScreenKeyboard does have a Unity Bug, which means it's neither null, nor reporting an inactive state even though it's closed, @julenka created a potential experimental fix #5018.
The problem is, the fix also does not work. For the fix to work the callbacks from the InputPane, Hiding and Showing need to fire, but they don't, at least not on WMR Immersive.

The code must be correct, so my question is, is it another bug, some configuration I have to do in Unity or something else?

## To reproduce

1. Create a new scene
2. add MixedRealityKeyboard.cs to a gameobject
3. add `ShowKeyboard();` to the start method
4. build and run on UWP native platform with script debugging enabled and all the network capabilities on
5. set a breakpoint in OnKeyboardShowing

it will never fire

## Expected behavior

The keyboard should just work rather, but for now I would be happy if those callbacks were fired.

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.4.2f1]
- MRTK Version current dev branch

## Target Platform (please complete the following information)

- WMR immersive

wiwei:
@julenka, do you remember if keyboard issue was also written for WMR Immersive? I remember the original keyboard issues we hit were with with HL devices.

Alexees:
> do you know which OS build you're running? I believe there was a bug with those events that was fixed in 18362.

@keveleigh I guess I better answer you here, because this is a new issue while the other was a closed PR. I checked the System and can confirm that it's 18362.239

Now I did a little investigation.
On that System (Windows 10 Home) the following happened:
- Opening the Keyboard does not even show it
- From the beginning of the opening, the Keyboard.active state stays true
- The mentioned callbacks of the InputPane are indeed not being called

the first issue is actually a weird one because at work, on another machine, it works. It has either to do with:
-WMR Headset VS Emulator
- Scene with or without MRTK
- Windows 10 Home Vs Enterprise
- Version difference
will test in a couple of hours as I'm not at work right now

Alexees:
Ok, this was a major f***up.
I came to the following conclusion:

- The Keyboard was not showing because I was calling its Open() method too (I think that's the reason). The InputField I was operating on seemed to do this already.
[...]

PR #5651, opened 2019-08-14
Fix keyboard show/hide for HL1 and WMR 

julenka:
# Overview
We want our windows system keyboard to work not just for HoloLens 2, but also for HoloLens (HL1) and Windows Mixed Reality (WHR). In HL1 and WMR there are OS bugs where the keyboard showing and hiding events do not fire. Our keyboard example was assuming those events fired correctly, and therefore not showing a keyboard in OnShow if the keyboard was visible. Remove this check.

## Changes
- Fixes: #5363

## Verification
Tested on HL1, HL2 and WMR

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Experimental/Features/UX/MixedRealityKeyboard.cs
+++ b/Assets/MixedRealityToolkit.SDK/Experimental/Features/UX/MixedRealityKeyboard.cs
@@ -123,7 +123,7 @@ namespace Microsoft.MixedReality.Toolkit.Experimental.UI
 
         public void ShowKeyboard()
         {
-            // 8/14/2019: We show the keyboard even when the keyboard is already visible because on HoloLens 1
+            // 2019/08/14: We show the keyboard even when the keyboard is already visible because on HoloLens 1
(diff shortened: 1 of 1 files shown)


Assets/MixedRealityToolkit.SDK/Experimental/Features/UX/MixedRealityKeyboard.cs
```diff
@@ -123,11 +123,13 @@ public void HideKeyboard()
 
         public void ShowKeyboard()
         {
-            if (state == KeyboardState.Showing)
-            {
-                Debug.Log($"MixedRealityKeyboard.ShowKeyboard called but keyboard already visible");
-                return;
-            }
+            // 8/14/2019: We show the keyboard even when the keyboard is already visible because on HoloLens 1
```
wiwei:
```suggestion
            // 2019/08/14: We show the keyboard even when the keyboard is already visible because on HoloLens 1
```

wiwei: Super nit - to be locale aware doing YYYY MM DD is usually pretty unambiguous.

Issue #5654, opened 2019-08-14
Optional extension service inspector and scriptableObject profile are not reported as optional


Troy-Ferrell:
## Describe the bug

When using the Extension Service wizard, if you uncheck the optional inspector and profile classes to be created. You get an error on the next page

Furthermore, it breaks compilation for your build. As the created Extension service makes a property variable for your scriptableobject profile so the variable must be deleted

Whenever this issue is fixed, the KnownIssues page needs to be updated (per #5660 )

```csharp
[MixedRealityExtensionService(SupportedPlatforms.WindowsStandalone|SupportedPlatforms.MacStandalone|SupportedPlatforms.LinuxStandalone|SupportedPlatforms.WindowsUniversal|SupportedPlatforms.WindowsEditor)]
	public class TroyTestService : BaseExtensionService, ITroyTestService, IMixedRealityExtensionService
	{
// TroyTestServiceProfile does not exist in the project
		private TroyTestServiceProfile troyTestServiceProfile;

		public TroyTestService(IMixedRealityServiceRegistrar registrar,  string name,  uint priority,  BaseMixedRealityProfile profile) : base(registrar, name, priority, profile) 
		{
			troyTestServiceProfile = (TroyTestServiceProfile)profile;
		}

		public override void Initialize()
		{
			// Do service initialization here.
		}

		public override void Update()
		{
			// Do service updates here.
		}
	}
```

![image](https://user-images.githubusercontent.com/25975362/63047491-6a9d4700-be89-11e9-9dfd-00368ede5679.png)

![image](https://user-images.githubusercontent.com/25975362/63047397-39bd1200-be89-11e9-826f-249c413ec0b5.png)

## To reproduce

Go To extension service wizard
Type in a service name
Uncheck the profile & inspector click next
View error

## Expected behavior

Should say successfully created everything

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.11f1]
- MRTK Version - stabilization Aug 14th

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

Troy-Ferrell:
Still an issue

PR #5660, opened 2019-08-14
Add service extension wizard known issue

Troy-Ferrell:
## Overview

Extension Service Wizard is broken for optional inspector and profile. This change adds information to the known issues list for GA

## Changes
Related #5654


SimonDarksideJ:
This seems a very odd "workaround" recommending users edit official MRTK files to make changes.
Is it not better to simply update the create service feature to make the optional elements mandatory and allow the user to remove files they don't need.

This seems an overly complicated documentation workaround that is only going to have users creating issues as they wont' see the docs until AFTER they encounter the issue.  Seems a very bad user experience @davidkline-ms

PR #5662, opened 2019-08-15
Fix additional Solver added during runtime not being updated

julianloehr-kg:
## Overview

Though #5416 fixed dynamically adding Solver during runtime, it just fixed the case where the GameObject has no Solver attached at all. When having a GameObject with an already attached Solver, adding another Solver (replacing the other one, or chaining them up) didn't work as intended. This is because the SolverHandler queries the list of its Solver just once.

This PR fixes those cases, by having the Solvers check and make sure they're in the list of solvers of their respective SolverHanlder.

Moreover this removes the necessary to use DestroyImmediate when replacing a Solver with another. This PR changes the Solver Example to reflect this as well.

While trying the Solver Example Scene, I noticed the Custom Target tracking wasn't working, because the manipulation handlers were not referencing the correct root GameObject. I fixed the references and included those changes in the Solver Example commit. Though I can commit that fix in a separate PR, if desired.

## Verification

Run the Solver Example Scene and all other scenes where Solvers are part of.


Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
```diff
@@ -98,19 +103,10 @@ private void RefreshSolverHandler()
...
-                DestroyImmediate(currentSolver);
-                currentSolver = null;
-            }
-
-            if (handler != null)
-            {
-                DestroyImmediate(handler);
-                handler = null;
-            }
+            foreach (Solver solver in GetComponents<Solver>())
```
SimonDarksideJ: Consider using a for loop instead of foreach, for performance.

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
```diff
@@ -69,24 +69,29 @@ public void SetOrbital()
...
-            currentSolver = gameObject.AddComponent<T>();
-            handler = GetComponent<SolverHandler>();
+            currentSolver = gameObject.EnsureComponent<T>();
+            currentSolver.enabled = true;
             RefreshSolverHandler();
         }
 
         private void RefreshSolverHandler()
         {
+            if(handler == null)
```
SimonDarksideJ: What if the handler has changed programatically?  This would cause the hander to NOT be refreshed

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
```diff
@@ -210,12 +210,18 @@ protected virtual void Awake()
...
+        /// When a solver becomes enabled, it must make sure it is registered at its <see cref="SolverHandler"/>.
+        /// Also it should update its internal state to the system, in case it was disabled far away.
         /// </summary>
         protected virtual void OnEnable()
         {
             if (SolverHandler != null)
             {
+                if(!SolverHandler.Solvers.Contains(this))
+                { 
+                    SolverHandler.Solvers = SolverHandler.GetComponents<Solver>();
```
SimonDarksideJ: SO I can see this is the crux of the change, however is multiple solvers get added / removed, this is going to be refreshed / updated multiple times.  Might be better to have a call back to the SolverHandler to refresh it's solvers and check the "Solvers" list against the component list, instead of repeatedly calling "GetComponents" which is costly.

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
```diff
@@ -98,19 +103,10 @@ private void RefreshSolverHandler()
             }
         }
 
-        private void DestroySolver()
+        private void DisableSolver()
```
SimonDarksideJ: I can see the reasons for the rename, but these have been bulked together in a related change.  If these are to be renamed, there should be a corresponding issue request to discuss the change and a separate PR to address it.

Each PR should change specifically only what it is intended to address or we lose tracability.

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
wiwei: This is something that we talked a bunch as a team about before, and in later version of unity/c# this isn't the same issue that it's been.

I'm happy to have this conversation separate from this PR to get numbers here, but I'm also not concerned because this isn't not a critical path (not update loop).

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
wiwei:
```suggestion
            if (handler == null)
```

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
```diff
@@ -210,12 +210,18 @@ protected virtual void Awake()
...
         /// <summary>
-        /// Typically when a solver becomes enabled, it should update its internal state to the system, in case it was disabled far away
+        /// When a solver becomes enabled, it must make sure it is registered at its <see cref="SolverHandler"/>.
+        /// Also it should update its internal state to the system, in case it was disabled far away.
         /// </summary>
         protected virtual void OnEnable()
         {
             if (SolverHandler != null)
             {
+                if(!SolverHandler.Solvers.Contains(this))
```
wiwei:
```suggestion
                if (!SolverHandler.Solvers.Contains(this))
```

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
```diff
@@ -210,12 +210,18 @@ protected virtual void Awake()
...
+                { 
+                    SolverHandler.Solvers = SolverHandler.GetComponents<Solver>();
+                }
```
wiwei: A couple of thoughts here - instead of having consumers of an object determine what solvers it manages, I think it would be better to instead of a system where instead the Solver will tell the SolverHandler that "hey, you should update your list" and the solver handler, in its LateUpdate() call, can then do the work to refresh its list once.

This has some benefits of only having to do the GetComponents<> call once if enabling a bunch of solvers in the same update loop, and also makes it so that you avoid propagating bugs in the Solvers themselves up the chain. What I mean by this is, if there were a bug in some solver code that updated SolverHandler.Solvers to be empty, it would now break all other solvers attached to that handler.

If instead, the was just an api that was like:

SolverHandler.QueueSolverListUpdate();

Then the surface area for error would also be less.

I *think* that this would also eliminate the need to do the:

public IReadOnlyCollection<Solver> Solvers

Change in SolverHandler.cs

So suggestion here is to:

protected virtual void OnEnable()
        {
            if (SolverHandler != null)
            {
                if(!SolverHandler.Solvers.Contains(this))
                { 
                    SolverHandler.QueueSolverListUpdate();
                }

And then in solver handler, in LateUpdate, just have the code that's currently in this section there instead.

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
wiwei: The naming change is fine by me - the intent here is to provide a change along with the qualifying stuff that demonstrates that it works. I don't think that the magnitude of the changes here warrant having a split even though I generally agree with the principle of splitting things up into smaller changes.

---
In reply to: [314231637](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5662#discussion_r314231637) [](ancestors = 314231637)

wiwei:
@Troy-Ferrell, because you just made some changes here as well, could you take a peek at SolverExampleManager.cs?

@julianloehr-kg Thanks for making this change! Can you also add a test case to: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Tests/PlayModeTests/SolverTests.cs to make sure that there's coverage of this example?

This will help make sure that we don't regress this in the future.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
wiwei: Had same comment, though in more words, maybe let's have this chat in my slightly longer comment.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
Troy-Ferrell: I would actually go a different route.

The GetComponents<Solver>() is very dangerous. and as @wiwei said bad performance. Furthermore, the SolverManager is supposed to manage the solvers...but a Solver telling a manager exactly what to manage seems very weird and problematic. It's probably better for a Solver to just say hey! manage me in a sort of auto-registeration

1) So SolverHandler would be updated to not set a default Solver list 
2) In Solver.Start(), have the solver register itself with the SolverHandler
3) In OnDestroy, solver unregisters itself

We should keep the Solver property as readonly. There is the SetSolvers() function if a higher order manager system wants to override the list

```csharp
public class solver
{
	Start()
	{
		if (SolverHandler != null)
		{
			SolverHandler.RegisterSolver(this)
		}
	}

	OnDestroy()
	{
		if (SolverHandler != null)
		{
			SolverHandler.UnregisterSolver(this)
		}
	}
}

public class SovlerHandler
{
	RegisterSolver(Solver s)
	{
		if s != null & !solvers.Contains(s)
		{
			solvers.Add(s);
		}
	}

	UnregisterSolver(Solver s)
	{
		if (s != null && solvers.Contains(s))
		{
			solvers.Remove(s);
		}
	}
}

```

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -173,11 +173,11 @@ public bool UpdateSolvers
         /// <summary>
         /// List of solvers that this handler will manage and update
         /// </summary>
-        public IReadOnlyCollection<Solver> Solvers
+        public ICollection<Solver> Solvers
```
Troy-Ferrell: Revert this and instead use the RegisterSolver and SetSolvers

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
Troy-Ferrell: I'm fine with this change being included but I did DestroyImmediate on purpose for a different reason.

In a real application, re-using resources and not wasting cycles on GarbageCollection makes sense. But this is designed to be an example scene to learn from. 

The solvers already have a crap load of properties to understand. I preferred only showing one solver at a time on the gameobject so it is very clear to the new developer. 

The idea being is you could select a solver and then in editor select that gameobject and modify the properties to test things out. Even disabled, having possibly 3-5 solvers or more as we grow, clutters up that inspector

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
julianloehr-kg: I can't follow, the following code also just takes the assumption that `handler` is still referencing the correct handler, and does not gather the correct handler reference every time.
This is for (a) it is the first call and the SolverHandler component has just been added to the GameObject, (b) someone/something has removed the SolverHandler component, causing a new component to be added.

If the handler has been changed programatically, the `handler` reference wouldn't be `null` therefore this is skipped.

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
julianloehr-kg: > But this is designed to be an example scene to learn from.

Though Unity Docs clearly state that they recommend to use `Destroy` instead of `DestroyImmediate`, as `DestroyImmediate` is intended for Editor Scripts only. So this would show learners a wrong use of Unity API.

The learning argument could also be used, to show them how they can simply switch between different Solvers, without doing complicated component destroy and rebuild procedures. But instead going the natural Unity was of simply disabling components that are temporarily not used.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
julianloehr-kg: That was my thought as well, but the fact that `SolverHandler.Solvers` is not readonly, but others can programatically change the list, and the Docs state that the update order shall be the order of components in the Inspector confused me. I will change this as it is indeed less error prone.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
julianloehr-kg: `IReadOnlyCollection` unfortunately does not have the `Contains` method, that was the reason for the change. Further the property still exposes a `set` method, that allows changing the list, so why offering the list as ReadOnly in the first place?

julianloehr-kg:
I will change this to the proposed Register-Unregister pattern.

@wiwei Regarding regression test, I will see what i can come up with. Have not that much experience with Unity Tests yet.

Assets/MixedRealityToolkit.Examples/Demos/Solvers/Scripts/SolverExampleManager.cs
Troy-Ferrell: I’m advocating destroying the components, not the manner in which they are destroyed. 

Switching between solvers dynamically is an edge case of common developer usage. The larger reason for the example scene is to try the solvers out in scene view and play with each solver’a properties. Honestly, the solver scene description and UX could heavily be improved all up

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/Solver.cs
Troy-Ferrell: Yes the docs will need to be updated as well. 

The GetComponents call by Unity will return components in the order that they appear in the inspector. SolverHandler by default calls GetComponents < Solver >

Troy-Ferrell:
For testing, search for “SolverTests.cs” in the MRTK.Tests section. You can add and leverage the existing items there

julianloehr-kg:
The proposed change now utilizes the Register-Unregister pattern for the Solver-SolverHandler relation. SolverHandler internally updates its solver list once during LateUpdate, when a new Solver has been added, or the solver list needs to be updated otherwise, e.g. triggered programmatically. The solver list update tries to guarantee Inspector ordering, by utilizing LINQ operations.

Further added a test, for adding another solver during runtime and switching between the two.

Lastly, I kept the current Solver Example mostly as it is, except for the `DestroyImmediate`. The fix does allow for just destroying the solver component and adding a new one, instead of destroying the Solver and SolverHandler Components entirely.

Assets/MixedRealityToolkit.SDK/Features/Utilities/Solvers/SolverHandler.cs
```diff
@@ -310,6 +310,14 @@ private void Update()
 
         private void LateUpdate()
         {
+            if (updateSolversList)
+            {
+                IEnumerable<Solver> inspectorOrderedSolvers = GetComponents<Solver>().Intersect(solvers);
```
Troy-Ferrell: How does Intersect() guarantee order? 

Also I don't think GetComponents work. this assumes all solvers will want to associate with this particular SolverHandler which is a bit dangerous. Like InBetween is an interesting scenario here. 

It also makes it difficult for inserting new solvers dynamically. If I can add a new solver dynamically, then it appears this will be put at the end of the list? But how would I insert that solver higher via GetComponents<>? Furthermore, if I was to register a solver with a solverhandler that is not on the sameobject...you could never insert it before other solvers.

julianloehr-kg: `solvers` will always be a list of only the solver that have registered, or have been set programmatically by something that wants to override the managed solvers.

`GetComponents` returns the list of all solvers in Insepctor order attached to this Gameobject. Interset() takes two lists and returns a list of elements that are in both lists, but keeping the order of the first list. See [Intersect<TSource>(IEnumerable<TSource>, IEnumerable<TSource>)](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.intersect?view=netframework-4.8#System_Linq_Enumerable_Intersect__1_System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEnumerable___0__), specifically the example and remarks section:

> When the object returned by this method is enumerated, `Intersect` yields distinct elements occurring in both sequences in the order in which they appear in first.

So `inspectorOrderedSolvers` is the list of solvers attached to this gameobject in Inspector order, but only those in the `solvers` list.  Finally using `Union` we add the remaining solvers, that are not attached to this Gameobject, but are in the `solvers` list.

So the order will always be: Insepctor order of Solvers attached to this Gameobject that have been registered, then remaining Solvers not attached to this Gameobject, in the order of their appearance in `solvers` (By default order of `Register` calls).

If you dynamically add a Solver, `AddComponent` will add them at the end of the list of components. Though you can mix up the order and manually place it at another position. Currently this change will only take effect, when you do it right away when adding the solver, before LateUpdate runs during that frame. So maybe there should be a public API to trigger the Solver List Update. Also the `Solvers` setter maybe should trigger Solver List Update, so the Insepctor Order is guaranteed.

I think the underlying problem is, what shall the ordering of solver updates be? Inspector Order, Order of register call or programatically set order?  And further how to resolve conflicts due to the different ways to add solvers and manipulate the solver list. Like, default is Inspector Order, but what happens when I override the solver list? Does the list order take precedence, is that reflected back to the Inspector, as in the component order is changed? Or does the inspector takes precedence and the list order is changed internally to reflect the current Inspector, like it is now? 

Regarding Solvers that are not attached to the SolverHandler Gameobject, that is an edge case. That edge case didn't work with the previous implementation. Now it does, but the Inspector Ordering is hard to apply, like which Inspector takes precedence. I'm open for discussion as to how the ordering shall apply there. To create such edge case, registering the solver has to be done manually in code by the developer, though.

Troy-Ferrell: @julianloehr-kg thanks for the great & detailed response. 

I did look at the Intersect docs but did not note the remarks. Good call out. Thank you.

I am going to file a bug that defining precedence/order for SolverHandler is a bit nuanced and problematic and will include all the items we discussed above. 

I'd rather get all your good changes here and fix the actual problem this PR was created for initially. We can do a separate PR to fix how priority can be easily defined for Solvers by consumers.

PR #5677, opened 2019-08-15
PressableButton MovingButtonVisuals inspector fix

AdamMitchell-ms:
## Overview
Fixes an issue where you could not modify the Moving Button Visuals on a PressableButton.

## Changes
The call to serializedObject.Update() seems to revert the value, so to fix the issue I've called serializedObject.ApplyModifiedProperties() just after the EditorGUILayout.PropertyField() call which draws and modifies the Moving Button Visuals.  It's a bit unconventional to call ApplyModifiedProperties twice in a single call to OnInspectorGUI, but I don't think it will cause any issues.

@Troy-Ferrell, you introduced the call to serializedObject.Update() about a week ago.  I'm open to alternative suggestions for how to fix it if you have any concerns.

Also switch to PropertyScope instead of Begin/EndProperty

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
@@ -50,6 +50,8 @@ namespace Microsoft.MixedReality.Toolkit.UI
                 // Convert world to local distances and vice versa whenever we switch the mode
                 if (value != distanceSpaceMode)
                 {
+                    // NOTE: This code is copied/pasted in PressableButtonInspector because
+                    // the inspector can only modify the serialized field and not the property
                     distanceSpaceMode = value;
                     float scale = (distanceSpaceMode == SpaceMode.Local) ? WorldToLocalScale : LocalToWorldScale;
                     startPushDistance *= scale;
@@ -209,7 +211,11 @@ namespace Microsoft.MixedReality.Toolkit.UI
             get { return movingButtonVisuals != null ? movingButtonVisuals.transform : transform; }
         }
 
-        private float WorldToLocalScale
+        /// <summary>
+        /// Transform for world to local space in the world direction of press
+        /// Multiply world scale positions by this value to convert to local space
+        /// </summary>
+        public float WorldToLocalScale
         {
             get
             {
@@ -217,7 +223,11 @@ namespace Microsoft.MixedReality.Toolkit.UI
             }
         }
 
-        private float LocalToWorldScale
+        /// <summary>
+        /// Transform for local to world space in the world direction of a press
+        /// Multiply local scale positions by this value to convert to world space
+        /// </summary>
+        public float LocalToWorldScale
         {
             get
             {
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
@@ -50,8 +50,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
                 // Convert world to local distances and vice versa whenever we switch the mode
                 if (value != distanceSpaceMode)
                 {
-                    // NOTE: This code is copied/pasted in PressableButtonInspector because
-                    // the inspector can only modify the serialized field and not the property
                     distanceSpaceMode = value;
                     float scale = (distanceSpaceMode == SpaceMode.Local) ? WorldToLocalScale : LocalToWorldScale;
                     startPushDistance *= scale;
@@ -215,7 +213,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
         /// Transform for world to local space in the world direction of press
         /// Multiply world scale positions by this value to convert to local space
         /// </summary>
-        public float WorldToLocalScale
+        private float WorldToLocalScale
         {
             get
             {
@@ -227,7 +225,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
         /// Transform for local to world space in the world direction of a press
         /// Multiply local scale positions by this value to convert to world space
         /// </summary>
-        public float LocalToWorldScale
+        private float LocalToWorldScale
         {
             get
             {
--- a/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
+++ b/Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
@@ -254,9 +254,11 @@ namespace Microsoft.MixedReality.Toolkit.Editor
             EditorGUILayout.PropertyField(movingButtonVisuals);
             EditorGUILayout.LabelField("Press Settings", EditorStyles.boldLabel);
 
+            serializedObject.ApplyModifiedProperties();
+
             var pos = EditorGUILayout.GetControlRect();
-            // Utilize EditorGUI.BeginProperty to keep prefab bolding and other editor field tracking
-            EditorGUI.BeginProperty(pos, DistanceSpaceModeLabel, distanceSpaceMode);
+            // Utilize EditorGUI.PropertyScope to keep prefab bolding and other editor field tracking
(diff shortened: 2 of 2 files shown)


thalbern:
Thanks for finding this @AdamMitchell-ms . We should move this to the stabilization branch. Just saw that this is broken there as well.

thalbern:
tested fix locally in stabilization branch

Troy-Ferrell:
@AdamMitchell-ms thanks for the find!

Unfortunately though, the coordinate space mode isn't being saved. I'll write up more in a second

Troy-Ferrell:
Ok it should be fully fixed now for all issues. I'd say it's both cleaner but also worse code : P 

I couldn't for the life of me get Update() and ApplyModifiedProperties to actually do as expected in multiple attempts. So instead I just resorted to using PropertyField() and then on change do the plane distance scaling myself...and that is the ugly part. We now have redundant code for the scaling procedure. The problem is I can't call the setter method as the serialized field is updated and the values won't be different. I would rather have this redundant code in this one spot that is clearly related to the main class then create a public function that is only useful for as editor code

Troy-Ferrell:
Ok it should be fully fixed now for all issues. I'd say it's both cleaner but also worse code : P

I couldn't for the life of me get Update() and ApplyModifiedProperties to actually do as expected in multiple attempts. So instead I just resorted to using PropertyField() and then on change do the plane distance scaling myself...and that is the ugly part. We now have redundant code for the scaling procedure. The problem is I can't call the setter method as the serialized field is updated and the values won't be different. I would rather have this redundant code in this one spot that is clearly related to the main class then create a public function that is only useful for as editor code

Troy-Ferrell:
@keveleigh is fixing the tree for removing mrtk_development commits from the stabilization path

davidkline-ms:
Test failure: TestGazeProviderTargetNotNull

GazeProvider target is null when looking at an object
Expected: not null
But was: null

at Microsoft.MixedReality.Toolkit.Tests.FocusProviderTests+<TestGazeProviderTargetNotNull>d__3.MoveNext () [0x00076] in F:\aipmragent_work\1\s\Assets\MixedRealityToolkit.Tests\PlayModeTests\FocusProviderTests.cs:86
at UnityEngine.TestTools.TestEnumerator+<Execute>c__Iterator0.MoveNext () [0x00031] in C:\buildslave\unity\build\Extensions\TestRunner\UnityEngine.TestRunner\NUnitExtensions\Attributes\TestEnumerator.cs:29

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -50,6 +50,8 @@ public SpaceMode DistanceSpaceMode
                 // Convert world to local distances and vice versa whenever we switch the mode
                 if (value != distanceSpaceMode)
                 {
+                    // NOTE: This code is copied/pasted in PressableButtonInspector because
+                    // the inspector can only modify the serialized field and not the property
```
thalbern: can't we move this code into a public function called ConvertPlaneDistances(SpaceMode toMode) and then call this function here and in the inspector?

thalbern: if you do that you don't have to expose WorldToLocalScale and LocalToWorldScale public

thalbern:
> Test failure: TestGazeProviderTargetNotNull
> 
> GazeProvider target is null when looking at an object
> Expected: not null
> But was: null
> 
> at Microsoft.MixedReality.Toolkit.Tests.FocusProviderTests+d__3.MoveNext () [0x00076] in F:\aipmragent_work\1\s\Assets\MixedRealityToolkit.Tests\PlayModeTests\FocusProviderTests.cs:86
> at UnityEngine.TestTools.TestEnumerator+c__Iterator0.MoveNext () [0x00031] in C:\buildslave\unity\build\Extensions\TestRunner\UnityEngine.TestRunner\NUnitExtensions\Attributes\TestEnumerator.cs:29

that looks like some random failure - the code shouldn't interfere with that test - retriggered CI

thalbern:
> > Test failure: TestGazeProviderTargetNotNull
> > GazeProvider target is null when looking at an object
> > Expected: not null
> > But was: null
> > at Microsoft.MixedReality.Toolkit.Tests.FocusProviderTests+d__3.MoveNext () [0x00076] in F:\aipmragent_work\1\s\Assets\MixedRealityToolkit.Tests\PlayModeTests\FocusProviderTests.cs:86
> > at UnityEngine.TestTools.TestEnumerator+c__Iterator0.MoveNext () [0x00031] in C:\buildslave\unity\build\Extensions\TestRunner\UnityEngine.TestRunner\NUnitExtensions\Attributes\TestEnumerator.cs:29
> 
> that looks like some random failure - the code shouldn't interfere with that test - retriggered CI

seems like it's passed now but checks section still shows the old failed build next to the new one that passed :/

davidkline-ms:
> seems like it's passed now but checks section still shows the old failed build next to the new one that passed :/

That is very strange. We can ignore the previous failure.

@thalbern, do you think we should address this first?

> can't we move this code into a public function called ConvertPlaneDistances(SpaceMode toMode) and then call this function here and in the inspector?

thalbern:
> > seems like it's passed now but checks section still shows the old failed build next to the new one that passed :/
> 
> That is very strange. We can ignore the previous failure.
> 
> @thalbern Bernadette Thalhammer FTE, do you think we should address this first?
> 
> > can't we move this code into a public function called ConvertPlaneDistances(SpaceMode toMode) and then call this function here and in the inspector?

yeah. actually i can quickly write this up and test on my machine if @Troy-Ferrell  is busy

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
Troy-Ferrell: Yes, I chose not to surface the public method. We could make one. I guess it's really pick your poison. They're both ugly cases to handle this one scenario just for the editor.

thalbern: i'm observing a very strange behavior - I debugged a bit in the inspector because I thought I had used that setter before and seems like you can call the setter from the inspector code - however, for some reason the hololens2 version of the button doesn' t apply the changes properly. So PressableButtons work with calling the setter but Hololens 2 won't - They both however call that setter, so there must be something else interfering afterwards in the hololens 2 version of that button

AdamMitchell-ms: What about doing something like this?

                var pressableButton = (PressableButton)serializedObject.targetObject;
                pressableButton.DistanceSpaceMode = (PressableButton.SpaceMode)distanceSpaceMode.enumValueIndex;

Then the setter logic can stay in one location.  The thing I'm unsure of is how to make sure that the values don't get overwritten by a later call to serializedObject.ApplyModifiedProperties.  Is that what serializedObject.Update() does?  Copy the values from the real object onto the serialized representation?

Troy-Ferrell: @AdamMitchell-ms the setter was being used in the original version. The Update() is supposed to copy the object values into the serializedobject and is why I had that originally. 

But exactly as the purpose for this PR being created, I had a lot of weird behavior using ApplyModifiedProperties & Update(). I tried multiple different iterations and values would just not get saved properly. 

The EditorGUILayout.PropertyField() is the only thing thus far that seems to cover all the bases correctly for whatever reason.

thalbern: it's only in the derived class where this is causing issues for me ¯\_(ツ)_/¯

thalbern:
i tested this a lot but it's already quite late, so if someone else could confirm I didn't break it 👍 :-D

Troy-Ferrell:
Creating from scratch and editing prefabs for PressableButton & PBHL2 both seem to work as expected. Moving visuals also seems correct as well!

Issue #5683, opened 2019-08-16
Add a warning for Internet connection requirement in Dictation Handler doc


Kent1LG:
## Describe the Issue
I'm currently working on an application which will be used without an internet connection. I wanted to use Dictation and it took me a long time to understand that an internet connection is required for Dictation to work. 

## Feature Area
Please can you add a warning on the documentation about the internet connection requirement?
 
## Existing Doc Link
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Input/Dictation.html?q=dictation

wiwei:
@Kent1LG definitely - I was checking out the Unity docs for this and they also don't mention this which is weird, but looking at https://docs.microsoft.com/en-us/windows/mixed-reality/voice-input-in-unity, this is mentioned there.

I'll update this to make sure that others don't have to go through your pain

PR #5686, opened 2019-08-16
Update the Dictation documentation to specify the internet connection and microphone requirements

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5683

The dictation system requires an internet connection because the underlying windows APIs (and Unity APIs which consume the windows APIs) requires those two capabilities to be enabled.


keveleigh:
Are you gonna put this change into stabilization as well, so it makes it into our GA docs?

wiwei:
> Are you gonna put this change into stabilization as well, so it makes it into our GA docs?

No, I'm planning on just keeping these updates to mrtk_development.

PR #5687, opened 2019-08-16
Clean up 'coming soon' documentation.

wiwei:
Some of these docs haven't been updated in a while, despite saying coming soon - as a principle, we should not check in things that are "coming soon" unless the work to do that has already been done and on your box locally and you're about to hit the button to create a PR. Otherwise, we risk creating a "graveyard" of documents that were dead before they even had the chance to live.

Some of these things being removed will be valuable to have docs on, and they can be re-added once they exist, but until then having empty pages has less value than nothing at all.


sostel:
Also, instead of removing images and pages, it would have been nice to have a call out for action first. I can take a stab at it this afternoon.

wiwei:
> Isn't the heat map image used in different doc pages? Even if the description of the tutorial is missing, we could at least describe the demos. Hence, I disagree with "until then having empty pages has less value than nothing at all" because it at least communicates that there is indeed some documentation coming instead of developers asking themselves, whether this will be ever addressed.

This change cleans up the usage in other doc pages as well. There's another composite image that has the same picture as part of some larger PNG.

I don't agree on having empty pages. It's extremely dissappointing as a reader to stumble across a months old documentation that still says coming soon. It gets someone's hopes up and then lets them down.

Documentation/EyeTracking/EyeTracking_ExamplesOverview.md
```diff
@@ -47,18 +47,6 @@ For more precisely positioning your hologram, you can use additional input from
...
 <br>
 
-
-[**Visualization of Visual Attention**](EyeTracking_Visualization.md)
-
-Information about where users look at is an immensely powerful tool to assess usability of a design and to identify problems in efficient work streams. 
-This tutorial discusses different eye tracking visualizations and how they fit different needs. 
-We provide basic examples for logging and loading eye tracking data and examples for how to visualize them. 
-
-**Summary**: Two-dimensional attention map (heatmaps) on slates. Recording & replaying eye tracking data.
```
keveleigh: This sample does exist: https://github.com/microsoft/MixedRealityToolkit-Unity/blob/prerelease/2.0.0_stabilization/Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/Scenes/EyeTrackingDemo-05-Visualizer.unity

wiwei:
In the same way, having TODOs in code is a great way to leave teasers of what someone thought about doing but never got around to, and never wrote down in detail for the next person to continue the thread. It leads to the build up of stuff that undermines the confidence in the code (and docs)

Documentation/EyeTracking/EyeTracking_ExamplesOverview.md
wiwei: I can keep this blurb around, but just delete the link

sostel:
> Isn't the heat map image used in different doc pages? Even if the description of the tutorial is missing, we could at least describe the demos. Hence, I disagree with "until then having empty pages has less value than nothing at all" because it at least communicates that there is indeed some documentation coming instead of developers asking themselves, whether this will be ever addressed.

I was thinking of the main MRTK landing page, where it is used: 
image

sostel:
> In the same way, having TODOs in code is a great way to leave teasers of what someone thought about doing but never got around to, and never wrote down in detail for the next person to continue the thread. It leads to the build up of stuff that undermines the confidence in the code (and docs)

Yeah, totally agree.

wiwei:
Closing this out to create one that targets stabilization

PR #5690, opened 2019-08-16
Clean up "coming soon" docs

wiwei:
Some of these docs haven't been updated in a while, despite saying coming soon - as a principle, we should not check in things that are "coming soon" unless the work to do that has already been done and on your box locally and you're about to hit the button to create a PR. Otherwise, we risk creating a "graveyard" of documents that were dead before they even had the chance to live.

Some of these things being removed will be valuable to have docs on, and they can be re-added once they exist, but until then having empty pages has less value than nothing at all.


Troy-Ferrell:
FYI There is also Building Window documentation page. I created it as a placeholder hoping someone who knows about the build window could fill it in. Unfortunately, I know very little about the build window : (

wiwei:
> FYI There is also Building Window documentation page. I created it as a placeholder hoping someone who knows about the build window could fill it in. Unfortunately, I know very little about the build window : (

Alright, removed that as well. I kept the mention of it on the getting started page, but removed the empty page that it linked to

PR #5692, opened 2019-08-16
Move the platform capability documentation to the features section.

wiwei:
This page wasn't really about the architecture of the MRTK at all, and interrupted the flow of reading through the architecture section (which is written in a way where each section builds off of the learnings from the previous page)

This also moves it into the root, because there isn't really a great place to stick this (i.e. it would be good to get a "features" folder to match the TOC, but I wanted to keep this particular change scoped) - cross platform was interesting but it doesn't really fulfill the cross-platform notion to consumers (i.e. I would think there would be more about cross-platform information here, but it's just a system capability checker, which exists even if things weren't cross platform)

### DIFF
--- a/Documentation/DetectingPlatformCapabilities.md
+++ b/Documentation/DetectingPlatformCapabilities.md
@@ -1,4 +1,4 @@
-# Determining platform capabilities
+# Detecting platform capabilities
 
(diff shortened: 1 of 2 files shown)


davidkline-ms:
doc generation error:

(Documentation/toc.yml#L0)Invalid file link:(~/Documentation/Architecture/CrossPlatform/DeterminingPlatformCapabilities.md).

Issue #5282, opened 2019-07-11
IMixedRealityTeleportHandler not working


pocce90:
## Describe the bug
Trying implementing the interface "IMixedRealityTeleportHandler" in a script attached on a gameObject. When I teleport myself, the system never enter inside the implemented interface methods

## To reproduce
- implement IMixedRealityTeleportHandler interface in a script
- attach the script on something in the scene
- try to teleport

## Expected behavior

The system executes the code inside the implemented methods

## Your Setup (please complete the following information)

- Unity Version 2018.3.12
- MRTK Version 2.1

## Target Platform (please complete the following information)

- WMR immersive

keveleigh:
You'll need to register as a listener of the teleport system. An example of how to do this (with the spatial awareness system, though the pattern should be the same) can be found at and around https://github.com/microsoft/MixedRealityToolkit-Unity/blob/5c172ecc4eb92fbd41924547ad0e654798e21e4e/Assets/MixedRealityToolkit.Examples/Demos/SpatialAwareness/Scripts/DemoSpatialMeshHandler.cs#L39

I'm also looking into an abstract class to abstract away much of this common code and make registering more straightforward.

wiwei:
Changing this one to a doc bug

PR #5701, opened 2019-08-16
Add documentation for the teleport system.

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5282

The teleport system is primarily interesting for VR cases, and I added some docs here mostly so that I could have a section to put the blurb on "how to register for teleport events"

We'd gotten some questions on how to use it (see the issue lined above) and it's clearly not-obvious how you get events, so this is necessary.


Documentation/TeleportSystem/Overview.md
```diff
@@ -0,0 +1,88 @@
...
+The teleport system is a sub-system of the MRTK that handles teleporting the user when the
+application is using an opaque display. For AR experiences (like HoloLens), the teleportation
+system is not active. For immersive HMD experiences (OpenVR, WMR) the teleport system can
+be enabled.
+
+## Enabling and disabling
+
+The teleport system can be enabled or disabled by toggling the checkbox in its profile:
+
+![Teleport system profile](../../Documentation/Images/TeleportSystem/profile.png)
```
thalbern: do we actually need that image or are we ok with just describing the path of that checkbox / property?

we have a docs guideline for images to only include them if absolutely necessary. Someone would have to update them whenever that profile changes
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/DocumentationGuide.html#images--screenshots
(I know we're already showing profile screenshots in a lot of places)

wiwei: That's fair - I think in this case the profile is simple enough that text is sufficient. I've gotten rid of the image.

Documentation/TeleportSystem/Overview.md
```diff
@@ -0,0 +1,88 @@
...
+
+void EnableTeleportSystem()
+{
+    CoreServices.TeleportSystem.Enable();
+}
+```
+
+## Events
+
+The teleport system exposes events through the [`IMixedRealityTeleportHandler`](xref:Microsoft.MixedReality.Toolkit.Teleport.IMixedRealityTeleportHandler)
```
Troy-Ferrell: Should we link to teleport pointers in the pointers.md?

Documentation/TeleportSystem/Overview.md
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    void OnDisable()
+    {
+        // Unregistering when disabled is important, otherwise this class will continue
+        // to recieve teleportation events.
+        CoreServices.TeleportSystem.UnregisterHandler<IMixedRealityTeleportHandler>(this);
+    }
+
+    void Update()
```
Troy-Ferrell: update seem unnecessary

Documentation/TeleportSystem/Overview.md
wiwei: I think there's some value to this, I haven't quite figured out the content yet, so I'll probably proceed with this change first and we can add the other parts as it comes to me.

Documentation/TeleportSystem/Overview.md
wiwei: Yep, removed

Issue #4960, opened 2019-06-19
Eye tracking exemple's page warning about Gaze input capabilite


Kent1LG:
Eye tracking exemple's page dosn't warn you about Visual Studio's capability "Gaze Input".
It should probably have a link for the "Getting started with Eye Tracking" page (https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/EyeTracking/EyeTracking_BasicSetup.html).

## Existing Doc Link
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/EyeTracking/EyeTracking_ExamplesOverview.html

wiwei:
Annnnd it took me way to long to get to this one, but I think I finally am adding a note to that doc

PR #5703, opened 2019-08-16
Update the eye tracking sample docs to reference the Gaze Input capability

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4960

Just like building an eye tracking app from scratch, using the samples also requires the Gaze Input capability, but it isn't mentioned in the docs. As a result, people using the samples on device will wonder why it doesn't work (because we didn't tell them to use that special capability)

### DIFF
--- a/Documentation/EyeTracking/EyeTracking_ExamplesOverview.md
+++ b/Documentation/EyeTracking/EyeTracking_ExamplesOverview.md
@@ -64,8 +64,8 @@ We provide basic examples for logging and loading eye tracking data and examples
 
 ### Prerequisites
 
-Note that using the eye tracking samples on device requires a device that supports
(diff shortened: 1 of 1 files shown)


sostel:
What about the section here where there is even a screenshot with the Gaze Input checkbox checked?
https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Documentation/EyeTracking/EyeTracking_BasicSetup.md#testing-your-unity-app-on-a-hololens-2

wiwei:
kk, linked to that section instead, ptal

wiwei:
@sostel, done, ptal!

sostel:
> @sostel, done, ptal!

Thank you for taking care of this, Will!

PR #5722, opened 2019-08-20
added release 2.0.0 / ga to version dropdown in docs

thalbern:
-adjusted contribute link to always point to mrtk_development
-added version 2.0.0 ga drop down

### DIFF
--- a/docfx.json
+++ b/docfx.json
@@ -51,11 +51,13 @@
       "_disableNavbar": false,
       "_appLogoPath": "Documentation/Images/mrt_logo_icon.png",
       "_appFaviconPath": "Documentation/Images/favicon.ico",
-      // remove this section if you want to restore default behavior which publishes a version of the docs that keeps the "improve this doc" link pointing to the publishing branch
+      // section contribute start
+	  // remove this section if you want to restore default behavior which publishes a version of the docs that keeps the "improve this doc" link pointing to the publishing branch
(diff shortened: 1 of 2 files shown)


docfx.json
```diff
@@ -50,12 +50,12 @@
...
       "_disableNavbar": false,
       "_appLogoPath": "Documentation/Images/mrt_logo_icon.png",
-      "_appFaviconPath": "Documentation/Images/favicon.ico"
-      // re-add this section if you want to publish a version of the docs that keeps the "improve this doc" link pointing to mrtk_development branch
-      //"_gitContribute": {
-      //  "repo": "https://github.com/Microsoft/MixedRealityToolkit-Unity",
-      //  "branch": "mrtk_development"
-      //}
+      "_appFaviconPath": "Documentation/Images/favicon.ico",
+      // remove this section if you want to restore default behavior which publishes a version of the docs that keeps the "improve this doc" link pointing to the publishing branch
```
wiwei:
```suggestion
      // remove this "_gitContribute" section if you want to restore default behavior which publishes a version of the docs that keeps the "improve this doc" link pointing to the publishing branch
```

wiwei: If only because I guess things could get added after this next scope, and it could be ambiguous where the section begins and ends.

thalbern: added section markers

PR #5726, opened 2019-08-20
fixed bounds calculation (#5711)

keveleigh:
Cherry-pick #5711 into stabilization


davidkline-ms:
This change was intended to also go into the stabilization branch and 2.0.0. Approving as the "final" code change in 2.0.0

Issue #5445, opened 2019-07-26
Add a snippet/usage guide for how to register for specific scenarios


wiwei:
The input docs are helpful in how to listen to specific events, but don't translate those into human terms:

https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Input/InputEvents.html

We should probably have some usage guide saying:

"If you want to listen for hand events, you can use this following code snippet..."

wiwei:
Also important to call out that Input system starts up in a non-defined order with respect to arbitrary other monobehaviors, so some degree of waiting of checking is required.

PR #5731, opened 2019-08-20
Update InputEvent documentation

Troy-Ferrell:
## Overview

Usage of input events was not well explained and examples had to be searched for in the examples package. This change gives greater detail to how input events are handle by the input system, how to register for global event handling, and how to mark an event as used. Further, there are some minor tweaks to other input docs.

## Changes
- Fixes: #5445


Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
 # Input Events
 
-At the script level you consume input events by implementing one of the event handler interfaces shown in the table below.
+The list below outlines all available input event interfaces to be implemented by a custom MonoBehavior component. These interfaces will be called by the MRTK Input System to handle custom app logic based on user input interactions. [Pointer input events](pointers.md#pointer-event-interfaces) are handled slightly differently than the standard input event types below.
```
wiwei: I feel like the old stuff had an interesting tidbit about only recieving events when in focus - it feels like that's still an important part to capture.

We may not need to add the part about why the old global thing sucks but the new thing is awesome, but the part that talked about how to expect to get them seems to be useful to keep around.

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+
+## Input events in action
+
+At the script level, input events can be consumed by implementing one of the event handler interfaces shown in the table above. When an input event fires via an user interaction, the following takes place:
+
+1. The MRTK input system recognizes that an input event has occurred
+1. The MRTK input system fires the relevant interface function of the input event to all registered global input handlers
+1. For every active pointer registered with the input system
+    1. The input system determines which GameObject is in focus for the current pointer
+    1. The input system utilizes [Unity's Event System](https://docs.unity3d.com/Manual/EventSystem.html) to fire the relevant interface function for all matching components on the focused GameObject
```
wiwei: Do you know.. if it's all matching components, or the first one that it finds? I think that we stop when finding the first one that handles it.

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+| [`IMixedRealityHandJointHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityHandJointHandler) | Hand Joints Updated | Raised by articulated hand controllers when hand joints are updated. |
+| [`IMixedRealityHandMeshHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityHandMeshHandler) | Hand Mesh Updated | Raised by articulated hand controllers when a hand mesh is updated. |
+| [`IMixedRealityInputActionHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputActionHandler) | Action Started / Ended | Raise to indicate action start and end for inputs mapped to actions. |
+
+## Input events in action
+
+At the script level, input events can be consumed by implementing one of the event handler interfaces shown in the table above. When an input event fires via an user interaction, the following takes place:
+
+1. The MRTK input system recognizes that an input event has occurred
+1. The MRTK input system fires the relevant interface function of the input event to all registered global input handlers
```
wiwei: "registered global input handlers" consider adding a link to the section below on global events)

wiwei: It would be interesting to note the short-circuit functionality - i.e. if someone is handling it at any of these places, it won't call the next step.

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+        - Note: The Unity Event System will bubble up to search the parent GameObject if no components matching the desired interface are found on the current GameObject.
+1. If no global input handlers are registered and no GameObject is found with a matching component/interface, then the input system will call each fallback registered input handler
+
+> [!NOTE]
+> [Pointer input events](pointers.md#pointer-event-interfaces) are handled in a slightly different manner than the input event interfaces listed above. In particular, pointer input events are handled only by the GameObject in focus by the pointer that fired the input event - as well as any global input handlers. Regular input events are handled by GameObjects in focus for all active pointers.
+
+### Input event interface example
+
+The code below demonstrates use of the [`IMixedRealitySpeechHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealitySpeechHandler) interface. When the user says the words "smaller" or "bigger" while focusing on a GameObject with this `ShowHideSpeechHandler` class, the GameObject will scale itself by half or twice as much.
```
wiwei: Should we also show the registration for global speech event? Otherwise we should add the note that this will only work if the person is looking or pointing at the object

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+    }
+}
+```
+
+> [!NOTE]
+> [`IMixedRealitySpeechHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealitySpeechHandler) input events require that the desired keywords are pre-registered in the [MRTK Speech Commands Profile](Speech.md).
+
+## Register for global input events
+
+To create a component that listens for global input events, disregarding what GameObject may be in focus, a component must register itself with the Input System. Once registered, any instances of this MonoBehavior will receive input events along with any GameObject(s) currently in focus and other global registered listeners.
```
wiwei: It might be worth caveating that by handling an event, it will no longer be delivered to other things (i.e. if the global thing handles it, it won't get delivered to objects in focus)

Documentation/Input/InputProviders.md
```diff
@@ -6,17 +6,17 @@ Input providers are registered in the **Registered Service Providers Profile**,
...
+| [`Unity Touch Device Manager`](xref:Microsoft.MixedReality.Toolkit.Input.UnityInput.UnityTouchDeviceManager) | Unity Touch Controller  |
+| [`Windows Dictation Input Provider`](xref:Microsoft.MixedReality.Toolkit.Windows.Input.WindowsDictationInputProvider) | *None*  |
+| [`Windows Mixed Reality Device Manager`](xref:Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input.WindowsMixedRealityDeviceManager) | WMR Articulated Hand, WMR Controller, WMR GGV Hand |
+| [`Windows Speech Input Provider`](xref:Microsoft.MixedReality.Toolkit.Windows.Input.WindowsSpeechInputProvider) | *None* |
 
 Dictation and Speech providers don't create any controllers, they raise their own specialized input events directly.
 
-Custom input providers can be created implementing the [`IMixedRealityInputDeviceManager`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputDeviceManager) interface.
\ No newline at end of file
+Custom input providers can be created by implementing the [`IMixedRealityInputDeviceManager`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputDeviceManager) interface.
```
wiwei: The existing line is the ultimate teaser... nothing for you to change here, I just wonder if it would have been better if we deleted this. It doesn't give enough useful stuff for people to operate on.

Documentation/Input/Pointers.md
```diff
@@ -127,6 +127,19 @@ Touch Started / Updated / Completed | Raised by touch-aware pointers like [`Poke
 > [!NOTE]
 > [`IMixedRealityFocusChangedHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityFocusChangedHandler) and [`IMixedRealityFocusHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityFocusHandler) should be handled in the objects they are raised on. It is possible to receive focus events globally but, unlike other input events, global event handler won't block receiving events based on focus (the event will be received by both global handler and a corresponding object in focus).
 
+#### Pointer input events in action
+
+1. The MRTK input system recognizes an input event has occurred
```
wiwei: It feels like this is a slight duplicate of the stuff already in the input doc above - thoughts on linking/consolidating or otherwise reducing copy pasta?

Documentation/Input/InputEvents.md
Troy-Ferrell: I don't understand the comment. There is an important tag saying you only receive events when in focus

Documentation/Input/InputEvents.md
Troy-Ferrell: @julenka to confirm? It's all components on the GameObject?

Documentation/Input/InputEvents.md
Troy-Ferrell: do you mean the .Use()? I have the section later on. I could re-add a note that any point that the process will stop if the event has been marked used

Documentation/Input/InputEvents.md
Troy-Ferrell: There is a section on global registeration.
This example does say "When the user says the words "smaller" or "bigger" while focusing on a GameObject with this `ShowHideSpeechHandler` class," and only on the object in focus

Documentation/Input/InputEvents.md
Troy-Ferrell: I don't believe this is true. @julenka to confirm?

Documentation/Input/InputProviders.md
Troy-Ferrell: Indeed, I'd say leave it as it at least gives people a direction. There should be work to come back and elaborate here

Documentation/Input/Pointers.md
Troy-Ferrell: It is duplicated. One points at the other and the other points back. 

I don't know where would be the consolidated position? It's not worth it's own page and there is no all up input events for both pages to be a child of.

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
 # Input Events
```
thalbern:
```suggestion
# Input events
```

Documentation/Input/InputEvents.md
thalbern:
```suggestion
The list below outlines all available input event interfaces to be implemented by a custom MonoBehavior component. These interfaces will be called by the MRTK input system to handle custom app logic based on user input interactions. [Pointer input events](pointers.md#pointer-event-interfaces) are handled slightly differently than the standard input event types below.
```

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+| [`IMixedRealitySourcePoseHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealitySourcePoseHandler) | Source Pose Changed | Raised on source pose changes. The source pose represents the general pose of the input source. Specific poses, like the grip or pointer pose in a six DOF controller, can be obtained via `IMixedRealityInputHandler<MixedRealityPose>`. |
+| [`IMixedRealityInputHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputHandler) | Input Down / Up | Raised on changes to binary inputs like buttons. |
+| [`IMixedRealityInputHandler<T>`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputHandler`1) | Input Changed | Raised on changes to inputs of the given type. **T** can take the following values: <br/> - *float* (e.g returns analog trigger)<br/> - *Vector2* (e.g returns gamepad thumbstick direction) <br/> - *Vector3* (e.g return position of tracked device) <br/> - *Quaternion* (e.g returns orientation of tracked device)<br/> - [MixedRealityPose](xref:Microsoft.MixedReality.Toolkit.Utilities.MixedRealityPose) (e.g. returns fully tracked device) |
```
thalbern:
```suggestion
| [`IMixedRealityInputHandler<T>`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputHandler`1) | Input Changed | Raised on changes to inputs of the given type. **T** can take the following values: <br/> - *float* (e.g returns analog trigger)<br/> - *Vector2* (e.g returns gamepad thumbstick direction) <br/> - *Vector3* (e.g returns position of tracked device) <br/> - *Quaternion* (e.g returns orientation of tracked device)<br/> - [MixedRealityPose](xref:Microsoft.MixedReality.Toolkit.Utilities.MixedRealityPose) (e.g. returns fully tracked device) |
```

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+| [`IMixedRealitySpeechHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealitySpeechHandler) | Speech Keyword Recognized | Raised on recognition of one of the keywords configured in the *Speech Commands Profile*. |
+| [`IMixedRealityDictationHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityDictationHandler) | Dictation<br/> Hypothesis <br/> Result <br/> Complete <br/> Error | Raised by dictation systems to report the results of a dictation session. |
+| [`IMixedRealityGestureHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityGestureHandler) | Gesture events on: <br/> Started <br/> Updated <br/> Completed <br/> Canceled | Raised on gesture detection. |
+| [`IMixedRealityGestureHandler<T>`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityGestureHandler`1) | Gesture Updated / Completed | Raised on detection of gestures containing additional data of the given type. See [**Gesture Events**](Gestures.md#gesture-events) for details on possible values for **T**. |
```
thalbern:
```suggestion
| [`IMixedRealityGestureHandler<T>`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityGestureHandler`1) | Gesture Updated / Completed | Raised on detection of gestures containing additional data of the given type. See [**gesture events**](Gestures.md#gesture-events) for details on possible values for **T**. |
```

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+| [`IMixedRealityHandJointHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityHandJointHandler) | Hand Joints Updated | Raised by articulated hand controllers when hand joints are updated. |
+| [`IMixedRealityHandMeshHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityHandMeshHandler) | Hand Mesh Updated | Raised by articulated hand controllers when a hand mesh is updated. |
+| [`IMixedRealityInputActionHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityInputActionHandler) | Action Started / Ended | Raise to indicate action start and end for inputs mapped to actions. |
+
+## Input events in action
+
+At the script level, input events can be consumed by implementing one of the event handler interfaces shown in the table above. When an input event fires via an user interaction, the following takes place:
+
+1. The MRTK input system recognizes that an input event has occurred
```
thalbern:
```suggestion
1. The MRTK input system recognizes that an input event has occurred.
```

Documentation/Input/InputEvents.md
thalbern:
```suggestion
1. The MRTK input system fires the relevant interface function of the input event to all registered global input handlers.
```

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+1. The MRTK input system fires the relevant interface function of the input event to all registered global input handlers
+1. For every active pointer registered with the input system
```
thalbern:
```suggestion
1. For every active pointer registered with the input system:
```

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+    1. The input system determines which GameObject is in focus for the current pointer
```
thalbern:
```suggestion
    1. The input system determines which GameObject is in focus for the current pointer.
```

Documentation/Input/InputEvents.md
thalbern:
```suggestion
    1. The input system utilizes [Unity's event system](https://docs.unity3d.com/Manual/EventSystem.html) to fire the relevant interface function for all matching components on the focused GameObject.
```

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+    1. The input system utilizes [Unity's Event System](https://docs.unity3d.com/Manual/EventSystem.html) to fire the relevant interface function for all matching components on the focused GameObject
+        - Example: Components implementing the interface [`IMixedRealitySpeechHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealitySpeechHandler) will be searched for when a speech command is recognized
```
thalbern:
```suggestion
        - Example: Components implementing the interface [`IMixedRealitySpeechHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealitySpeechHandler) will be searched for when a speech command is recognized.
```

Documentation/Input/InputEvents.md
```diff
@@ -1,23 +1,169 @@
...
+        - Note: The Unity Event System will bubble up to search the parent GameObject if no components matching the desired interface are found on the current GameObject.
```
thalbern:
```suggestion
        - Note: The Unity event system will bubble up to search the parent GameObject if no components matching the desired interface are found on the current GameObject.
```

Documentation/Input/Pointers.md
thalbern: Can't you instead of copy pasting say something like:

```suggestion
Pointer input events are recognized and handled by the MRTK input system in a similar way than [regular input events](InputEvents.md#input_events_in_action). The difference being that pointer input events are handled only by the GameObject in focus by the pointer that fired the input event - as well as any global input handlers. Regular input events are handled by GameObjects in focus for all active pointers.
```

Documentation/Input/InputEvents.md
wiwei: Yep - that's what I'm referring to - this text seems to suggest it will keep going, but it can (and sometimes does) stop early

Documentation/Input/InputEvents.md
wiwei: By handling, I guess I meant marking the event as used

Documentation/Input/InputEvents.md
wiwei: The important tag shows up later, after the reader has already gone through the interface information (and potentially already started trying to implement those interfaces). Moreover, the important comment shows up in the section called "registering for global events"

Basically, I think that still having the important tag stuff up here about requiring focus is still good to tell people before telling them about the interfaces (i.e. 

step 1) tell people how to start listening for events in general
step 2) tell people exactly which interfaces for which events

wiwei:
lgtm pending resolving the current open comments

Documentation/Input/InputEvents.md
Troy-Ferrell: ok I see. @wiwei do you think at the very top or with the "Input events in action" since that actually explains how to utilize them?

Documentation/Input/InputEvents.md
Troy-Ferrell: Ah yes ok

Documentation/Input/InputEvents.md
Troy-Ferrell: What's the change here? I don't see a delta?

Documentation/Input/InputEvents.md
Troy-Ferrell: So testing global handlers, Use() does not cause it to not call other GameObjects. Use() will not objects with focus though

Documentation/Input/InputEvents.md
Troy-Ferrell: Testing it out, all components on a focused gameobject will receive the callback even if one of them marks the eventData.use()

Documentation/Input/Pointers.md
```diff
@@ -115,96 +115,111 @@ Useful Sphere Pointer properties:
...
 
 > [!NOTE]
 > [`IMixedRealityFocusChangedHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityFocusChangedHandler) and [`IMixedRealityFocusHandler`](xref:Microsoft.MixedReality.Toolkit.Input.IMixedRealityFocusHandler) should be handled in the objects they are raised on. It is possible to receive focus events globally but, unlike other input events, global event handler won't block receiving events based on focus (the event will be received by both global handler and a corresponding object in focus).
 
+#### Pointer input events in action
+
+Pointer input events are recognized and handled by the MRTK input system in a similar way as [regular input events](InputEvents.md#input-events-in-action). The difference being that pointer input events are handled only by the GameObject in focus by the pointer that fired the input event - as well as any global input handlers. Regular input events are handled by GameObjects in focus for all active pointers.
+
+1. The MRTK input system recognizes an input event has occurred
```
thalbern: I meant we only add the sentence from the top and remove this list as we're linking to the regular input events (in action) anyways

Troy-Ferrell: Ah I see. I would rather keep the list. The flow is slightly different in the list so it’s not an exact duplicate and see the entire flow with the key change is more informative to new developers than linking to another page, giving a comment of what’s different roughly, and then have them piece it together between the two pages.

PR #5736, opened 2019-08-21
Add daily build with ARM64 build flavour

MenelvagorMilsom:
## Overview
Adds a daily build intended to be scheduled once a day. This build allows us to verify additional build flavours without clogging the main CI and PR pipelines. To begin with, I've added an ARM64 build that only runs on the daily build.
Alongside these changes I needed to modify the build scripts to allow us to build ARM64 on CI. This included adding `ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch` to the VS project. I believe this issue has been seen elsewhere for ARM64 builds.

### DIFF
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
@@ -237,7 +237,16 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
                 rootNode.Add(propertyGroupNode);
             }
 
-            propertyGroupNode.Add(new XElement(defaultNamespace + "ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch", "None"));
+            var newNode = propertyGroupNode.Element(defaultNamespace + "ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch");
+            if (newNode != null)
+            {
+                // If this setting already exists in the project, ensure it's value is "None"
+                newNode.Value = "None";
+            }
+            else
+            {
+                propertyGroupNode.Add(new XElement(defaultNamespace + "ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch", "None"));
+            }
 
             rootNode.Save(projectFilePath);
 
--- a/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
+++ b/Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
@@ -97,6 +97,15 @@ namespace Microsoft.MixedReality.Toolkit.Build.Editor
                 return false;
             }
 
+            // Need to add ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch to MixedRealityToolkit.vcxproj
+            if (buildInfo.BuildPlatform == "arm64")
+            {
+                if (!UpdateVSProj(buildInfo))
+                {
+                    return IsBuilding = false;
+                }
+            }
+
             // Now that NuGet packages have been restored, we can run the actual build process.
             exitCode = await Run(msBuildPath, 
                 $"\"{solutionProjectPath}\" {(buildInfo.Multicore ? "/m /nr:false" : "")} /t:{(buildInfo.RebuildAppx ? "Rebuild" : "Build")} /p:Configuration={buildInfo.Configuration} /p:Platform={buildInfo.BuildPlatform} {(string.IsNullOrEmpty(buildInfo.PlatformToolset) ? string.Empty : $"/p:PlatformToolset={buildInfo.PlatformToolset}")} {GetMSBuildLoggingCommand(buildInfo.LogDirectory, "buildAppx.log")}",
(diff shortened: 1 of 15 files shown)


Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -206,6 +215,37 @@ private static async Task<string> FindMsBuildPathAsync()
             return string.Empty;
         }
 
+        private static bool UpdateVSProj(IBuildInfo buildInfo)
```
macborow: It's a pity that it is necessary to patch existing project files instead of handling that during project generation.
Do we also need to modify any templates for the MSBuild process?

pipelines/templates/tasks/unitybuild.yml
```diff
@@ -7,11 +7,19 @@ parameters:
   ScriptingBackend: 'default'  # [optional] default|.NET
   PublishArtifacts: false
   PackagingDir: 'Win32'
+  UnityVer: 'Unity2018'
```
macborow: Why not pass the actual path in here? I.e. make it default to {$Env:${Unity2019Version)} and pass in the path to the editor if different.

pipelines/ci-daily.yml
```diff
@@ -0,0 +1,23 @@
...
+  MRTKVersion: 2.0.0
+  packagingEnabled: false
+
+jobs:
+- job: CIDeveloperValidation
+  timeoutInMinutes: 120
+  pool:
+    name: On-Prem Unity
+    demands:
+    - Unity2018.3.7f1
```
macborow: Would it make sure to separate Unity2018 and Unity2019 tasks into separate jobs? This way we won't have to be extra careful about having agents with *both* versions present.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
MenelvagorMilsom: >It's a pity that it is necessary to patch existing project files instead of handling that during project generation.

Yeah, as far as I can tell this is a known issue. I'm not sure there's anything we can do for the time being.

> Do we also need to modify any templates for the MSBuild process?

I'm not sure, I didn't try it. I'm not sure if the MSBuild stuff works at all for Unity2019 or arm64

pipelines/templates/tasks/unitybuild.yml
MenelvagorMilsom: Sounds good to me.

pipelines/ci-daily.yml
MenelvagorMilsom: Yeah, sounds reasonable.

MenelvagorMilsom: I guess we can run tests on 2019 too, wouldn't do any harm.

MenelvagorMilsom:
Looks like most tests are failing in Unity2019. I confirmed this locally, but didn't look into it further.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -206,6 +215,37 @@ private static async Task<string> FindMsBuildPathAsync()
...
+            var defaultNamespace = rootNode.GetDefaultNamespace();
+            var propertyGroupNode = rootNode.Element(defaultNamespace + "PropertyGroup");
+            
+            if (propertyGroupNode == null)
+            {
+                propertyGroupNode = new XElement(defaultNamespace + "PropertyGroup", new XAttribute("Label", "Globals"));
+                rootNode.Add(propertyGroupNode);
+            }
+
+            propertyGroupNode.Add(new XElement(defaultNamespace + "ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch", "None"));
```
wiwei: Do you know if running this twice will cause it to add two elements (which are both this thing?)

I think when I did the capability thing (i.e. gaze input) I had to check to see if it existed before adding it, otherwise it would actually add it twice.

wiwei:
> Looks like most play mode tests are failing in Unity2019. I confirmed this locally, but didn't look into it further.

I think that @googlan mentioned the same thing a while back when we he was looking at 2019

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -99,6 +99,15 @@ public static async Task<bool> BuildAppxAsync(UwpBuildInfo buildInfo, Cancellati
                 return false;
             }
 
+            // Need to add ResolveAssemblyWarnOrErrorOnTargetArchitectureMismatch to MixedRealityToolkit.vcxproj
```
keveleigh: I *think* I saw that this was fixed in the latest VS 2019

https://github.com/microsoft/msbuild/issues/3746

davidkline-ms:
@MenelvagorMilsom, once the merge conflicts are resolved, is this good to go in?

MenelvagorMilsom:
@davidkline-ms I was holding out on this one until someone's had a look at tests failing for Unity2019. As far as I'm aware no one has had a look yet.

MenelvagorMilsom:
@davidkline-ms the usual PR build won't be able to verify if tests are passing in 2019 now. If we want to check that I can rebase this branch and test on the daily build.

Issue #5737, opened 2019-08-21
Canvas only reflects the case where it is being used for MRTK


Alexees:
## Describe the bug

A lot of UI elements can be used for visualization only (Loading-disc anyone?). This means, one might not necessarily want to interact with them in any way. The Canvas Inspector does not reflect that choice. Worse, it simply shows the default fields from the Canvas class, of which some are deprecated or not shown at all, because they are only populated by the Unity CanvasEditor class (this is another story).
Additionally, the Canvas can only work in WorldSpace mode and therefore should only show properties relevant to that mode.

## Expected behavior

The proper way for indicating one does not want to interact with a canvas is to remove the GraphicsRaycaster, which is what the custom inspector should reflect.

## Screenshots

Unity Version 2018.4.6f1
MRTK Version mrtk_development

## Target Platform (please complete the following information)

Editor

## Additional context

I think it's annoying to be warned about a missing CanvasUtility if you want to use the canvas for  visuals only

Alexees:
Closing via #5738

PR #5738, opened 2019-08-21
Full canvas inspector support

Alexees:
## Overview
The current Canvas inspector merely draws the base inspector, which doesn't reflect the Unity default inspector in the least. Additionally, it does not respect Canvases that might only be used for visualization.

## Changes
- Fixes: #5737

The inspector now fully reflects the current state of the configuration:

- A Canvas without a GraphicRaycaster only reminds of the absence, which might be what you want
- Switching to other camera modes switches to the Unity default inspector completely, which means you're not annoyed by the warning about the CanvasUtility 
- The WorldSpace mode now reflects the use of the required CanvasUtility in that the Camera field doesn't show when properly configured. And it also shows up like the Unity inspector would


thalbern:
@Alexees can you retarget your branch to mrtk_development please?

thalbern:
Hi @Alexees. Had a look at your changes in Unity editor and it seems like whenever I'm configuring canvas for world space I'm loosing the option to switch back to another render mode (along with some other properties that are disappearing)

new version:
image
old version: 
image

Alexees:
> I like the convenience "add component" buttons you've added 🚀 .
> 
> Change looks good but the current state will break inspector behavior. After switching to world space mode you can never switch back again.

That's strange, did not happen to me when I made this. I'll check this out on wednesday.

Alexees:
@thalbern I was wrong, it was intended this way. I wanted the Inspector to look as little cluttered as possible. To get the space dropdown back you have to remove the CanvasUtility, which is kind of the key component to dinstinguish between an MRTK Canvas and a default one.
At the moment you also have to switch to world space on the canvas component first to even get the button for adding the Utility.
I'm open for suggestions. Do you think it makes sense to have the "Add CanvasUtility" button be renamed to just "Make MRKT Canvas" or something, so the user doesn't have to care what components go into this? Likewise a "Remove MRTK capability" which does the same in reverse?

thalbern:
> @thalbern Bernadette Thalhammer FTE I was wrong, it was intended this way. I wanted the Inspector to look as little cluttered as possible. To get the space dropdown back you have to remove the CanvasUtility, which is kind of the key component to dinstinguish between an MRTK Canvas and a default one.
> At the moment you also have to switch to world space on the canvas component first to even get the button for adding the Utility.
> I'm open for suggestions. Do you think it makes sense to have the "Add CanvasUtility" button be renamed to just "Make MRKT Canvas" or something, so the user doesn't have to care what components go into this? Likewise a "Remove MRTK capability" which does the same in reverse?

I like the decluttering you did, though I still wouldn't take that dropdown away from users - I think it's not clear that you have to remove the utility component to get that option back. The buttons will for sure make it better but I would still show that dropdown - we can grey it out and show the tooltip that it needs to be converted back to "normal" canvas / non mrtk canvas to be editable again?
that way it's clearer to the user what's going on?

Alexees:
@thalbern I'd add that "making it a MRTK canvas" should be shown no matter the canvas space and automatically configures as world space. additionally, since the camera is required to be null and a Canvas Utility as well, both configurations can be made automatically.
Then, a button shows up along with the greyed out dropdown that would remove all of that

thalbern:
> @thalbern Bernadette Thalhammer FTE I'd add that "making it a MRTK canvas" should be shown no matter the canvas space and automatically configures as world space. additionally, since the camera is required to be null and a Canvas Utility as well, both configurations can be made automatically.
> Then, a button shows up along with the greyed out dropdown that would remove all of that

Sounds good!

Alexees:
@thalbern you might want to have another go?

thalbern:
> @thalbern Bernadette Thalhammer FTE you might want to have another go?

looking!

Assets/MixedRealityToolkit.Services/InputSystem/Editor/MixedRealityCanvasInspector.cs
```diff
@@ -0,0 +1,342 @@
...
+                    EditorGUILayout.HelpBox($"Canvas does not contain any {typeof(NearInteractionTouchableUnityUI).Name} components for supporting near interaction.", MessageType.Warning);
+                    if (GUILayout.Button($"Add {typeof(NearInteractionTouchableUnityUI).Name}"))
+                    {
+                        Undo.AddComponent<NearInteractionTouchableUnityUI>(canvas.gameObject);
+                    }
+                }
+            }
+            else
+            {
+                if (GUILayout.Button("Make MRTK canvas"))
```
thalbern: Can we add some tooltip or info msg here as well that explains what's going to happen or link to documentation, just so it's more clear what's going on.

Assets/MixedRealityToolkit.Services/InputSystem/Editor/MixedRealityCanvasInspector.cs
```diff
@@ -0,0 +1,342 @@
...
+            {
+                return false;
+            }
+
+            bool isMRTKCanvas = canvas.GetComponent<CanvasUtility>() != null;
+
+            if (isMRTKCanvas)
+            {
+                if (GUILayout.Button("Revert MRTK canvas"))
+                {
```
thalbern: I would add some tooltip or info msg that says we're going to remove CanvasUtility and NearInteractionTouchableUnityUI

Alexees:
@thalbern I changed both button names and added button hover tooltips

wiwei:
It looks like this got affected by the license/cla bug (i.e. it is in a forever pending state). I'm going to close and reopen this PR to force it to refresh.

wiwei:
Yay, it fixed it, merging this in now.

PR #5747, opened 2019-08-21
Document extension services; how to create, register and use

davidkline-ms:
This document briefly describes creating extension services, getting them registered with the MRTK and how application code can access and use the services.


davidkline-ms:
closing to try to get cla to validate correctly

Issue #4607, opened 2019-05-29
Error when the number of hand mesh vertices decreases


tarukosu:
## Describe the bug
I'm using MixedRealityToolkit.InputSystem?.RaiseHandMeshUpdated method.
HandMeshInfo data is created based on Leap Motion hand mesh data.

When the number of vertices decreases, the following error occured.
```
Mesh.vertices is too small. The supplied vertex array has less vertices than are referenced by the triangles array.
UnityEngine.Mesh:set_vertices(Vector3[])
Microsoft.MixedReality.Toolkit.Input.BaseHandVisualizer:OnHandMeshUpdated(InputEventData`1) (at Assets/MixedRealityToolkit/Providers/Hands/BaseHandVisualizer.cs:151)
Microsoft.MixedReality.Toolkit.Input.<>c:<.cctor>b__174_40(IMixedRealityHandMeshHandler, BaseEventData) (at Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs:1412)
UnityEngine.EventSystems.ExecuteEvents:Execute(GameObject, BaseEventData, EventFunction`1)
Microsoft.MixedReality.Toolkit.BaseEventSystem:HandleEvent(BaseEventData, EventFunction`1) (at Assets/MixedRealityToolkit/Services/BaseEventSystem.cs:32)
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:DispatchEventToGlobalListeners(BaseInputEventData, EventFunction`1) (at Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs:357)
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:HandleEvent(BaseEventData, EventFunction`1) (at Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs:287)
Microsoft.MixedReality.Toolkit.Input.MixedRealityInputSystem:RaiseHandMeshUpdated(IMixedRealityInputSource, Handedness, HandMeshInfo) (at Assets/MixedRealityToolkit.Services/InputSystem/MixedRealityInputSystem.cs:1421)
Microsoft.MixedReality.Toolkit.WindowsMixedReality.Input.LeapMotionHand:UpdateHandMesh(HandMeshInfo) (at Assets/LeapMotionInputSimulation/Providers/LeapMotionHand.cs:171)
[...]

PR #5750, opened 2019-08-21
Fix error when the number of hand mesh vertices decreases

wiwei:
On certain platforms it's possible for hand mesh vertex array sizes to decrease between events, and the current infrastructure doesn't support that.

This addresses it (using the fix provided in https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4607) and adds a test that shows it working.

### DIFF
--- a/Assets/MixedRealityToolkit/Providers/Hands/BaseHandVisualizer.cs
+++ b/Assets/MixedRealityToolkit/Providers/Hands/BaseHandVisualizer.cs
@@ -176,6 +176,16 @@ namespace Microsoft.MixedReality.Toolkit.Input
             {
                 Mesh mesh = handMeshFilter.mesh;
 
+                // On some platforms, mesh length counts may change as the hand mesh is updated.
+                // In order to update the vertices when the array sizes change, the mesh
+                // must be cleared per instructions here:
+                // https://docs.unity3d.com/ScriptReference/Mesh.html
+                if ((mesh.vertices?.Length ?? 0) != 0 &&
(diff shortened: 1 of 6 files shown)


Assets/MixedRealityToolkit/Providers/Hands/BaseHandVisualizer.cs
```diff
@@ -176,6 +176,16 @@ public void OnHandMeshUpdated(InputEventData<HandMeshInfo> eventData)
             {
                 Mesh mesh = handMeshFilter.mesh;
 
+                // On some platforms, mesh length counts may change as the hand mesh is updated.
+                // In order to update the vertices when the array sizes change, the mesh
+                // must be cleared per instructions here:
+                // https://docs.unity3d.com/ScriptReference/Mesh.html
+                if ((mesh.vertices?.Length ?? 0) != 0 &&
```
Cameron-Micka: [mesh.vertices](https://docs.unity3d.com/ScriptReference/Mesh-vertices.html) returns a **copy** of a mesh's vertices (crazy, right?). So, you may want to cache them once before the conditional, rather than potentially creating two copies. Also, I'm pretty Unity guarantees mesh verticies, indicies, etc. are never null (they can be empty) so you could remove those checks too.

wiwei: Thanks for calling that out - I ended up caching the last vertex array (to avoid having an additional copy each time).

wiwei:
@Cameron-Micka, can you take another look?

Issue #5205, opened 2019-07-04
No Way to Have Extension Service Marked as UWP Only? (i.e. not for editor?)


mtaulty:
## Describe the bug

I have a Mixed Reality Extension Service. I have marked it so that the only "supported platform" is "WindowsUniversal". I do not want that service to be instantiated inside of the Unity Editor.

However, when I run my code in the editor I find that the service is instantiated.

## To reproduce

Make a MixedRealityExtensionService by deriving from BaseExtensionService. Mark it with a MixedRealityExtensionService attribute and specify only "WindowsUniversal" for the supported platforms.

Add it to your extension services configuration in the toolkit configuration and then debug the project in Visual Studio and see whether your service is created or not.

## Expected behavior

I would not expect the service to be instantiated in the editor, only on a UWP device.

## Screenshots

N/A

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.4.3f1]
- MRTK Version [e.g. v2.0 RC2.1]

## Target Platform (please complete the following information)

- HoloLens/2

## Additional context

The code in PlatformUtility.cs seems to do the checking here. 

I am looking at a call to GetSupportedPlatformMask and when running in the editor it is returning "WindowsUniversal | WindowsEditor". My service says it supports "WindowsUniversal" and so the code decides to instantiate my service inside of the editor but by marking my service as "WindowsUniversal" only, I expected to keep it away from running inside of the editor.

Is the logic there correct?

wiwei:
@davidkline-ms

Troy-Ferrell:
I believe this is fixed in stabilization. @mtaulty try the prerelease/2.0.0_stabilization or even mrtk_development?

From MRTK.cs, line 170
```csharp
#if !UNITY_EDITOR
            if (!Application.platform.IsPlatformSupported(supportedPlatforms))
#else
            if (!EditorUserBuildSettings.activeBuildTarget.IsPlatformSupported(supportedPlatforms))
#endif
            {
                return false;
            }
```

Troy-Ferrell:
Looks like primary logic is here:

```csharp
#if !UNITY_EDITOR
            if (!Application.platform.IsPlatformSupported(supportedPlatforms))
#else
            if (!EditorUserBuildSettings.activeBuildTarget.IsPlatformSupported(supportedPlatforms))
#endif
            {
                return false;
            }
```

[...]

PR #5771, opened 2019-08-23
Fix SupportedPlatforms checking

Troy-Ferrell:
## Overview
The SupportedPlatforms enum has Editor listings but it's not exactly accurrate and it's a bit weird. I could be running in WindowsEditor but targeting different platforms such as WindowsStandalone, UWP, etc.

If running in Editor, platform comparison checks are handled by looking at the current BuildTarget while if running in Player, platform comparison checks are handled by looking at the actual app endpoint.

The problem is there is currently no way to indicate that a service/dataprovider should target a certain platform BUT also only on device or only in editor. 

@wiwei, added the SupportedApplicationModes enum flags which this PR leverages to allow services and data providers to target running in Editor, Player, or Both. 

PlatformUtility now supports functions to query against SupportedApplicationModes.
ExtensionServiceWizard now allows for targeting against SupportedApplicationModes

**Breaking changes:**
IMixedRealityServiceConfiguration => Added SupportedApplicationModes RuntimeModes and does not support None results
IMixedRealityServiceRegistrar => RegisterService interface has SupportedApplicationModes default param now

This should be mostly internal folks or rare occurrences.

## Changes
- Fixes: #5205 

## Verification
All tests pass
TODO:

>
>


Assets/MixedRealityToolkit.Extensions/SceneTransitionService/SceneTransitionService.cs
```diff
@@ -14,7 +14,10 @@ namespace Microsoft.MixedReality.Toolkit.Extensions.SceneTransitions
     [HelpURL("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Extensions/SceneTransitionService/SceneTransitionServiceOverview.html")]
     public class SceneTransitionService : BaseExtensionService, ISceneTransitionService, IMixedRealityExtensionService
     {
-        public SceneTransitionService(IMixedRealityServiceRegistrar registrar, string name, uint priority, BaseMixedRealityProfile profile) : base(registrar, name, priority, profile)
+        public SceneTransitionService(IMixedRealityServiceRegistrar registrar,
```
Troy-Ferrell: @Railboy should the SceneTransitionService be able to run in play mode editor?

Troy-Ferrell:
@keveleigh / @davidkline-ms , what is the easiest way to test OpenVRDeviceManager?

keveleigh:
>Although the SupportedPlatforms enum has WindowsEditor and related entries, these are actually ignored. 

That doesn't sound right. It should be accounted for with

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/fbaa5c2dc7e9e9ab924fcf716d00dc4fb80dca8a/Assets/MixedRealityToolkit/Utilities/PlatformUtility.cs#L79

when running in the editor.

Assets/MixedRealityToolkit/Utilities/PlatformUtility.cs
```diff
@@ -20,22 +20,28 @@ private static SupportedPlatforms GetSupportedPlatformMask(RuntimePlatform runti
             switch (runtimePlatform)
             {
                 case RuntimePlatform.WindowsPlayer:
-                case RuntimePlatform.WindowsEditor:
                     supportedPlatforms |= SupportedPlatforms.WindowsStandalone;
                     break;
+                case RuntimePlatform.WindowsEditor:
```
keveleigh: I think this change will break most existing services, which don't explicitly mark the editor as a supported platform (like WMRDeviceManager) when running in the editor:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Assets/MixedRealityToolkit.Providers/WindowsMixedReality/WindowsMixedRealityDeviceManager.cs#L26

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
```diff
@@ -167,11 +167,7 @@ public void ResetConfiguration(MixedRealityToolkitConfigurationProfile profile)
                 return false;
             }
 
-#if !UNITY_EDITOR
             if (!Application.platform.IsPlatformSupported(supportedPlatforms))
-#else
-            if (!EditorUserBuildSettings.activeBuildTarget.IsPlatformSupported(supportedPlatforms))
```
keveleigh: The idea here was that you care more about the active build target in the editor than the fact that you're in the editor.

I added the editor as an explicit target to mark editor-only services, which did muddy the definition a little bit (for example, you can't currently mark something as in-app only, ignoring running in the editor).

Assets/MixedRealityToolkit/Utilities/PlatformUtility.cs
keveleigh: There may be some more changes needed here

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
keveleigh: Removing this would make things a little less configurable, i think, since all services that *can* run in the editor *will* run in the editor, even if one only runs when the build target is Standalone and one only runs when the build target is UWP.

Troy-Ferrell:
> > Although the SupportedPlatforms enum has WindowsEditor and related entries, these are actually ignored.
> 
> That doesn't sound right. It should be accounted for with
> 
> https://github.com/microsoft/MixedRealityToolkit-Unity/blob/fbaa5c2dc7e9e9ab924fcf716d00dc4fb80dca8a/Assets/MixedRealityToolkit/Utilities/PlatformUtility.cs#L79
> 
> when running in the editor.

Sorry, I updated my PR comments to be more accurate. The problem isn't turning on for WindowsEditor. It's turning off. If I purposefully leave out WindowsEditor in my supportedplatforms mask, it will still run because my build target is on UWP. See bug #5205

Assets/MixedRealityToolkit/Services/MixedRealityToolkit.cs
Troy-Ferrell: I think I see what you mean but it's more confusing. If I mark something as supported in WindowsEditor, that should be the end all truth. 

I see what you mean in that if I mark a service as WindowsStandalone and WindowsEditor, it will still run in-editor with build target UWP or really any windows build target. But saying we only use BuildTarget in editor is not obvious compared to marking a service that yes I can run in editor

keveleigh:
>Hmmm, sounds like we need another parameter like bool canRuninEditor or bool onlyOnDevice etc?

Yeah, separating out the concept of editor / in-app from platform seems like a good idea!

Assets/MixedRealityToolkit.Extensions/SceneTransitionService/SceneTransitionService.cs
julenka: Ping @Railboy can you help answer here?

Alexees:
I Integrated this Into my initial idea of a type based platform approach for more fine grained selectability.
You might want to check it out. It's not finished in that it's not clear how the conditions stack.

https://github.com/HoloSpaces/MixedRealityToolkit-Unity/tree/mrtk_typeBasedPlatformSupportWithEditorRuntime

Troy-Ferrell:
> I Integrated this Into my initial idea of a type based platform approach for more fine grained selectability.
> You might want to check it out. It's not finished in that it's not clear how the conditions stack.
> https://github.com/HoloSpaces/MixedRealityToolkit-Unity/tree/mrtk_typeBasedPlatformSupportWithEditorRuntime

Very cool stuff. I took a look at your code. I really like the idea of an interface that people can more easily customize/control via IPlatformSupport

I feel like though IPlatformSupport should fully replace the SupportedPlatforms enums. It's seems weird/much more additional complexity to try to have both. 

Also the enum added by @wiwei , `SupportedApplicationModes`, is a separate enum dimension that would still be needed with IPlatformSupport

SupportedApplicationModes is sort of like test vs production toggle. Because I may be targeting UWP apps with a certain service but should it run while testing in editor and while in "production" on device? Or just one of the either.

Alexees:
> I feel like though IPlatformSupport should fully replace the SupportedPlatforms enums

Oh man... Since you directed me to your PR with yet another enum I thought I had to adjust to that... my initial design effectively incorporated SupportedApplicationModes already and now I removed it 💩

Alexees:
This PR addresses the distinction between Editor and Runtime. Mine addresses the distinction between subcategories on platforms, like OculusGo and Oculus Quest, which are both Android.

Troy-Ferrell:
@alexees is exactly right. I believe the main thing up for question is how the two should play together. 

Alex created a IPlatformSupport interface which could go in conjunction with SupportedPlatforms/SupportedApplicationModes....but at the same time, we could potentially just use IPlatformSupports exclusively and internally in those concrete classes they may use the various enums etc Or some other alternative.

julenka:
@Troy-Ferrell what's the status on this PR? Is anything blocking you here? Should we not merge it for some reason? Do you need a review?

Troy-Ferrell:
> @Troy-Ferrell what's the status on this PR? Is anything blocking you here? Should we not merge it for some reason? Do you need a review?

Mainly I keep de-prioritizing it. 

The other component is I am just not that happy with the current solution. It can unblock the referenced issue but the whole where am I supported question for components seems overcomplicated.

Also the dataprovider/system class attributes seems like they are getting overloaded with parameters.

I'll try to re-review this soon

Alexees:
Wanted to add that the ARFoundation integration now introduced a situation where distinguishing between devices within a platform would be nice. Right now one has to manually remove UnityARCameraSettings from Android support if you want to build for Oculus Quest for example, you'll otherwise end up with a black screen.

wiwei:
Closing out old/stale PRs

PR #5781, opened 2019-08-26
Get PressableButton working well with UnityUI

AdamMitchell-ms:
## Overview
PressableButton and its inspector previously only worked with NearInteractionTouchable.  With this change, it now also works well with NearInteractionTouchable**UnityUI**.  This makes it possible for users to create UnityUI Buttons which resemble MRTK buttons in appearance and functionality.

## Breaking Changes
- NearInteractionTouchableCollider is removed.
- NearInteractionTouchable's LocalBounds and LocalCenter properties are now getter-only.  To set them, you now must call SetLocalBounds() and SetLocalCenter().

## Changes
- NearInteractionTouchable and NearInteractionTouchableUnityUI now derive from NearInterationTouchableSurface.
- PressableButton and its inspector now use NearInterationTouchableSurface rather than either concrete class.
- Created Prefabs for buttons based on UnityUI but which are styled to look like MRTK buttons.
   - Had to create new or extend existing utility classes to support this.
- Added a section to the HandInteractionExamples scene to showcase the new style of buttons.

Here's how they look and behave in the sample scene.

Pressable Button UnityUI 1

Pressable Button UnityUI 2

Pressable Button UnityUI 3


thalbern:
> > ```
> >     protected float startPushDistance = 0.0f;
> > ```
> 
> For UnityUI buttons, this only seems to work if the start push distance is 0. You just need to add a UnityUI Graphic (ex. Image) as far forward as you want the touch interaction to start.
> Maybe I should add a check for that somewhere? Or just have the Inspector not render this field if there is a NearInteractionTouchableUnityUI?
> 
> Refers to: Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs:64 in [4928ceb](https://github.com/microsoft/MixedRealityToolkit-Unity/commit/4928cebd0109b5b2d1ebe279c29d855afb402f42). [](commit_id = [4928ceb](https://github.com/microsoft/MixedRealityToolkit-Unity/commit/4928cebd0109b5b2d1ebe279c29d855afb402f42), deletion_comment = False)

Is the start plane in case of the unity ui outside the collider? we only receive updates after we've entered the collider of the button.

Assets/MixedRealityToolkit.Services/InputSystem/Editor/NearInteractionTouchableInspector.cs
```diff
@@ -118,4 +90,36 @@ public override void OnInspectorGUI()
             }
         }
     }
-}
\ No newline at end of file
+
+    public class NearInteractionTouchableInspectorBase : UnityEditor.Editor
```
thalbern: can you add a  [UnityEditor.CustomEditor(typeof(BaseNearInteractionTouchable), true)] here so this is the fallback for all the NearInteractionTouchables? Then you also won't have to define the NearInteractionTouchableUnityUI editor further up.

Assets/MixedRealityToolkit.Services/InputSystem/INearInteractionTouchable.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    public interface INearInteractionTouchable
```
thalbern: add documentation to public interface

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -158,16 +151,16 @@ private void DrawButtonInfo(ButtonInfo info, bool editingEnabled)
...
             if (editingEnabled && newPressDistance != info.PressDistance)
             {
                 EnforceDistanceOrdering(ref info);
                 info.PressDistance = Mathf.Clamp(newPressDistance, info.ReleaseDistance, info.MaxPushDistance);
             }
 
             // MAX PUSH
-            Handles.color = Color.cyan;
-            float newMaxPushDistance = DrawPlaneAndHandle(endPlaneVertices, info.TouchCageLocalBounds.size * 0.5f, info.MaxPushDistance, info, "Max Push Distance", editingEnabled);
+            Handles.color = Color.red;
```
thalbern: we already defined red for the release plane

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/RectTransformCubeScaler.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+[ExecuteInEditMode]
+[RequireComponent(typeof(RectTransform))]
+public class RectTransformCubeScaler : MonoBehaviour
+{
```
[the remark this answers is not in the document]
thalbern: where is this class used? I'm not sure I understand the use case. Can you add some doc and maybe use it somewhere in an example?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PhysicalPressEventRouter.cs
```diff
@@ -3,18 +3,22 @@
 
 using Microsoft.MixedReality.Toolkit.UI;
 using UnityEngine;
+using UnityEngine.UI;
 
 namespace Microsoft.MixedReality.Toolkit.Examples.Demos
 {
     ///<summary>
-    /// This class exists to route <see cref="Microsoft.MixedReality.Toolkit.UI.PressableButton"/> events through to Interactable.
+    /// This class exists to route <see cref="Microsoft.MixedReality.Toolkit.UI.PressableButton"/> events through to <see cref="Interactable"/> and/or <see cref="UnityEngine.UI.Button"/>.
```
thalbern: i'm not super happy to extend this class because we have to throw it out (or redo) soon.

The unity ui button case does not share anything with the interactable, so it looks a bit weird - the class is handling two completely different things - should we create a base class and specialized versions for those two cases?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PhysicalPressEventRouter.cs
```diff
@@ -44,6 +53,11 @@ public void OnHandPressTouched()
                     routingTarget.SetPress(false);
                 }
             }
+
+            if (uiButton != null && InteractableOnClick == PhysicalPressEventBehavior.EventOnTouch)
+            {
+                uiButton.onClick.Invoke();
```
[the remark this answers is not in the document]
thalbern: I like the executeEvents approach!
Though I wouldn't put this into pressablebutton right away. We would have to move that EventBehavior enum over as well and we can't get rid of the router class without taking care of interactable, which I wouldn't want to put on Adam right now. 
So just to avoid having a version where users have to set this up in different locations depending on if it's unity ui or interactable I would say we leave this in the router until I find time to fix this for Interactable as well. We're currently looking into migration scripts, so as soon as that's in place it should hopefully be an easy one to just move those fields from the router into pressablebutton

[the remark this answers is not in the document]
thalbern: if we add this we should adjust the doc comments and InteractableOnClick member as well so it's not only referring to interactable

Assets/MixedRealityToolkit.Services/InputSystem/INearInteractionTouchable.cs
thalbern: can we rename this to INearInteractionTouchableDirected? It's not going to be used by the volume near interaction touchable and doesn't make sense to move it into base class. I think renaming would make it easier to understand what it's for.

Cameron-Micka:
Just a quick update. @AdamMitchell-ms and I spoke in person yesterday and I'm going to create a seperate PR that addresses the scale issues observed when using the MRTK/Standard shader. It turns out on UGUI components the unity_ObjectToWorld matrix is not the transformation matrix of the gameobject the Image component lives on, but that of it's parent Canvas component. So, I'm going to push in extra scaling information using a BaseMeshEffect to preserve UI batching.

TL;DR Please feel free to ignore button border visuals for this PR. The hope is we can use the exact same materials for UGUI renderers and traditional MeshRenderers.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PhysicalPressEventRouter.cs
```diff
@@ -3,18 +3,22 @@
...
+    /// This class exists to route <see cref="Microsoft.MixedReality.Toolkit.UI.PressableButton"/> events through to <see cref="Interactable"/> and/or <see cref="UnityEngine.UI.Button"/>.
     /// The result is being able to have physical touch call Interactable.OnPointerClicked.
     ///</summary>
     public class PhysicalPressEventRouter : MonoBehaviour
     {
         [Tooltip("Interactable to which the press events are being routed. Defaults to the object of the component.")]
         public Interactable routingTarget;
 
+        [Tooltip("Unity UI Button to invoke on press of PressableButton")]
+        public Button uiButton;
```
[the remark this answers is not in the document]
AdamMitchell-ms: Pulled out into a separate class now.  It's much simpler now; just a single function with a single line.

Assets/MixedRealityToolkit.Services/InputSystem/BaseNearInteractionTouchable.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     public abstract class BaseNearInteractionTouchable : MonoBehaviour
     {
         [SerializeField]
-        protected TouchableEventType eventsToReceive = TouchableEventType.Touch;
```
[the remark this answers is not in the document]
AdamMitchell-ms: Reverted these breaking changes.

Assets/MixedRealityToolkit.Tests/PlayModeTests/PressableButtonTests.cs
```diff
@@ -28,11 +31,43 @@ public void Setup()
...
+
+        private GameObject InstantiateDefaultPressableButton(string prefabFilename)
+        {
+            var path = Path.Combine(PrefabDirectoryPath, prefabFilename);
+            Object pressableButtonPrefab = AssetDatabase.LoadAssetAtPath(path, typeof(Object));
             GameObject testButton = Object.Instantiate(pressableButtonPrefab) as GameObject;
 
+            if (PressableButtonTestPrefabs[prefabFilename])
+            {
+                //TODO: does this get cleaned up by TearDown?
```
thalbern: should be safe to not destroy this as the play test base class is deleting all the root level gameobjects in the tests scenes on teardown

AdamMitchell-ms: Thanks.  Yes I ran all of the tests with my latest changes.  They all pass and none leave any gameobjects behind.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PressableButton.cs
```diff
@@ -188,16 +187,16 @@ private set
         public bool IsPressing { get; private set; }
 
         /// <summary>
-        /// The press direction of the button as defined by a NearInteractionTouchable.
+        /// The press direction of the button as defined by an INearInteractionTouchable.
```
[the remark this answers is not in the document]
AdamMitchell-ms: Updated with the latest naming.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -254,6 +252,15 @@ public override void OnInspectorGUI()
             EditorGUILayout.PropertyField(movingButtonVisuals);
             EditorGUILayout.LabelField("Press Settings", EditorStyles.boldLabel);
 
+            if (touchable is NearInteractionTouchableUnityUI && startPushDistance.floatValue < 0.0f)
+            {
+                EditorGUILayout.HelpBox("UnityUI based PressableButtons don't support negative Start Push Distances.", UnityEditor.MessageType.Warning);
+                if (GUILayout.Button("Set to 0"))
```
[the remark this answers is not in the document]
AdamMitchell-ms: Sounds good.  Done now.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
```diff
@@ -171,7 +171,7 @@ private bool FindClosestTouchableForLayerMask(LayerMask layerMask, out BaseNearI
 
             for (int i = 0; i < numColliders; ++i)
             {
-                var touchable = queryBuffer[i].GetComponent<ColliderNearInteractionTouchable>();
+                var touchable = queryBuffer[i].GetComponent<BaseNearInteractionTouchable>();
```
julenka: @LukasTeonneMS, I think this should be fine because the UnitUI touchables will not be hit by the spherecast anyway

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PhysicalPressEventRouter.cs
```diff
@@ -3,11 +3,13 @@
 
 using Microsoft.MixedReality.Toolkit.UI;
 using UnityEngine;
+using UnityEngine.EventSystems;
+using UnityEngine.UI;
```
julenka: do we need these two usings?

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/RectTransformCubeScaler.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities
+{
+    /// <summary>
+    /// RectTransforms do not scale 3d objects (such as unit cubes) to fit within their bounds.
+    /// This helper class will apply a scale to fit a unit cube into the bounds specified by the RectTransform.
+    /// The Z component is scaled to the min of the X and Y components.
+    /// </summary>
```
julenka: I'm still not sure where I would use this, could you provide an example of when I'd want to use it, and where it's used in MRTK? I don't see where it's used in this PR but perhaps I am missing something.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -116,10 +108,16 @@ private ButtonInfo GatherCurrentInfo()
...
-            info.TouchCageLocalBounds = new Bounds(touchCage.center, touchCage.size);
+            info.LocalCenter = touchable.LocalCenter;
+            info.PlaneExtents = touchable.Bounds;
 
-            Vector3 pressDirLocal = (touchable != null) ? -1.0f * touchable.LocalForward : Vector3.forward;
-            Vector3 upDirLocal = (touchable != null) ? touchable.LocalUp : Vector3.up;
+            Vector3 pressDirLocal = (touchable != null) ? touchable.LocalPressDirection : Vector3.forward;
+            Vector3 upDirLocal = Vector3.up;
+
+            if (touchable is NearInteractionTouchable touchableConcrete)
```
julenka: Woah cool, I didn't know you could do that!

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -116,10 +108,16 @@ private ButtonInfo GatherCurrentInfo()
...
+            {
+                upDirLocal = touchableConcrete.LocalUp;
```
julenka: should this be NearInteractionTouchableSurface?

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
```diff
@@ -231,7 +229,7 @@ private float DrawPlaneAndHandle(Vector3[] vertices, Vector3 halfExtents, float
             {
                 float handleSize = HandleUtility.GetHandleSize(vertices[1]) * 0.15f;
 
-                Vector3 dir = (touchable != null) ? -1.0f * touchable.LocalForward : Vector3.forward;
+                Vector3 dir = (touchable != null) ? touchable.LocalPressDirection : Vector3.forward;
```
julenka: Since  we are  now negating LocalForward to be LocalPressDirection, will this mess up people who have created custom pressable buttons, or other custom elements that have near interaction touchables? Say in our next release, if those assets still have LocalForward to be pointing at (0, 0, 1), will they no longer be able to press buttons until they re-configure the buttons? If yes, let's try to figure out a plan for migration of our customers. A surprising number of customers create copies of our prefabs and then edit / customize them, and the breaks on upgrade might make them reluctant to upgrade and get all the goodness we are adding

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
```diff
@@ -13,7 +12,7 @@ namespace Microsoft.MixedReality.Toolkit.Input
     /// Add a NearInteractionTouchable to your scene and configure a touchable surface
     /// in order to get PointerDown and PointerUp events whenever a PokePointer touches this surface.
     /// </summary>
-    public class NearInteractionTouchable : ColliderNearInteractionTouchable
+    public class NearInteractionTouchable : NearInteractionTouchableSurface
```
julenka: @wiwei Adam and I were lamenting that NearInteractionTouchable needs to inherit NearInteractionTouchableSurface.

My ideal would be:

NearInteractionTouchableSurface : BaseNearInteractionTouchable
NearInteractionTouchableVolume : BaseNearInteractionTouchable
NearInteractionTouchableUnityUI : BaseNearInteractionTouchable

I'd like to create a plan to get us there. Not needed for this PR since one of MRTK members can do this. 

Something like this:
- Mark NearInteractionTouchable as Deprecated 
- Provide migration guidance "if this is something that should be touched only from the front, use surface, otherwise use volume"?
- In one release after, delete NearInteractionTouchable

I can create an issue to follow up on this

Assets/MixedRealityToolkit.Tests/PlayModeTests/PressableButtonTests.cs
```diff
@@ -53,9 +86,9 @@ public void TearDown()
         #region Tests
 
         [UnityTest]
-        public IEnumerator ButtonInstantiate()
+        public IEnumerator ButtonInstantiate([ValueSource(nameof(PressableButtonsTestPrefabFilenames))] string prefabFilename)
```
julenka: Woah cool

Documentation/README_Button.md
```diff
@@ -23,7 +31,7 @@ The [`Button.prefab`](https://github.com/Microsoft/MixedRealityToolkit-Unity/blo
...
+Simply drag [`PressableButtonHoloLens2.prefab`](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButtonHoloLens2.prefab), [`PressableButtonHoloLens2UnityUI.prefab`](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButtonHoloLens2UnityUI.prefab), or [`PressableButtonHoloLens2Unplated.prefab`](https://github.com/Microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButtonHoloLens2Unplated.prefab) into the scene. These button prefabs are already configured to have audio-visual feedback for the various types of inputs, including articulated hand input and gaze.
```
julenka: If I dragged a UnityUI button not on a UnityUI canvas, would they work? I thought UnityUI buttons needed to be on UnityUI canvases. It might be good to have a quick Hhow to add pressable buttons to UnityUI canvas" section.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs
lukastoenneMS: But why is it necessary?

AdamMitchell-ms: Per request from @julenka, I refactored the class hierarchy.  ColliderNearInteractionTouchable has been deleted.  See this whiteboard drawing of the old vs new hierarchy.  (NIT = NearInteractionTouchable)

mrtk_nearinteraction_class_hierarchy (2)

The red INIT was an interface I had created in an earlier iteration of this PR to let PressableButton talk with NearInteractionTouchable AND NearInteractionTouchableUnityUI.  Now it talks with them both via NearInteractionTouchableSurface.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/PressableButtons/PhysicalPressEventRouter.cs
AdamMitchell-ms: Good callout.  I've now reverted all changes to this file.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Utilities/RectTransformCubeScaler.cs
AdamMitchell-ms: I use it in a prefab.  Specifically, I needed it for [PressableButtonHololens2UnityUI.prefab](https://github.com/AdamMitchell-ms/MixedRealityToolkit-Unity/blob/unityui-nearfield-improvements2/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Prefabs/PressableButtonHololens2UnityUI.prefab).

image

PressableButtonHololens2 needs a cube for the compressable visual.  But in Unity, a cube ignores a RectTransform's Width/Height.  So to get it to fit into the bounds specified by the size, I created this little script.

To see this for yourself, create a canvas in Unity.  Then place an image in it and change the Width/Height of the rect transform.  The image will be resized to fit inside that size.  Now create a cube and give it a rect transform.  Changing the width/height has no effect on the cube.  The cube's size is determined only by the Scale of the transform hierarchy.  So this script takes the rect transform's width/height and applies a scale which will cause a 1x1x1 sized object to fit within the bounds.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
AdamMitchell-ms: Yes, it's available as of C# 7.  This actually reminds me...
@julenka, does MRTK target .NET 4.7 and above?

If you target .NET 4.6 or below, then I should remove the use of this syntax.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
AdamMitchell-ms: NearInteractionTouchableSurface has no LocalUp property.  It only has LocalCenter, LocalPressDirection, and Bounds.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
AdamMitchell-ms: For existing NearInteractionTouchables, this should still work.  No migration plan should be required.

Whatever they have set as their local forward direction on their own prefabs (or prefab variants of MRTK buttons) will be negated into LocalPressDirection.  So the value produced by this line will be identical now as it was prior to my change.  The existing button prefabs in MRTK already have non-default localForward, and they function correctly.  Therefore the MRTK buttons are equivalent to the upgrade scenario, so I think this should satisfy testing.

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
AdamMitchell-ms: @julenka , I thought your ideal was what I've now done (other than the fact that "NearInteractionTouchable" should ideally be called something like "NearInteractionTouchableCollider" or something like that.  For reference, here's the whiteboard drawing we came up with of what I had done in an earlier iteration of this PR vs what is now done.

mrtk_nearinteraction_class_hierarchy (2)

Assets/MixedRealityToolkit.Tests/PlayModeTests/PressableButtonTests.cs
AdamMitchell-ms: I know.  It's kind of like magic.  I learned about this relatively recently when working on [UnityGLTF](https://github.com/KhronosGroup/UnityGLTF).  Unity knows how to parse this in its test window and it shows up like this.

image

Documentation/README_Button.md
AdamMitchell-ms: Ok, I've clarified that a bit and put it in a separate section.

Assets/MixedRealityToolkit.SDK/Inspectors/UX/Interactable/PressableButtonInspector.cs
AdamMitchell-ms: Regarding PressableButtonsTestCanvas.prefab, it is an example prefab.  I named it that because it lives right next to the existing "TestCanvas.prefab" which is all of the pre-existing UnityUI stuff in the HandInteractionExamples scene.  I was trying to be consistent with the existing naming scheme.

[Hand tracking prefabs](https://github.com/AdamMitchell-ms/MixedRealityToolkit-Unity/tree/unityui-nearfield-improvements2/Assets/MixedRealityToolkit.Examples/Demos/HandTracking/Prefabs)

Assets/MixedRealityToolkit.Services/InputSystem/NearInteractionTouchable.cs
julenka: Oh, sorry. Let me update my ideal. 

NearInteractionTouchableSurface : BaseNearInteractionTouchable
NearInteractionTouchableVolume : BaseNearInteractionTouchable
NearInteractionTouchableUnityUI : NearInteractionTouchableSurface 

And then NearInteractionTouchable no longer exists, instead NearInteractionTouchable would be renamed to NearInteractionTouchableSurface.

I was trying to say that I think the name "NearInteractionTouchable" is misleading, since it's really a touchable surface. It threw me a bit that NearInteractionTouchable : NearInteractionTouchableSurface but then I remembered that was the plan.

AdamMitchell-ms:
The auto merge caused a build break.  I've manually merged and resolved the break.

@julenka , @wiwei , or @keveleigh ,
Can one of you "/azp run" for me?

Issue #5004, opened 2019-06-21
Can't delete select and delete pointer options in Input Profile


thalbern:
## Describe the bug

It's currently not possible to select and delete existing pointer options in the input profile. The button for deleting is there but we can't select any elements in the list so the button is not working.

## To reproduce
-Select the MRTK gameobject 
- navigate to Input
- inspect the pointer options
- try selecting one of them and pressing the - button for deleting them

## Expected behavior
- pointers entries should be selectable and deletable

## Your Setup (please complete the following information)

- Unity Version 2018.4
- MRTK Version 2 RC2.1

Troy-Ferrell:
@thalbern FYI as workaround, if you set handness to nothing then the pointer should be active for any input

Troy-Ferrell:
@thalbern Also if you double click directly into the pointer profile, then you can select & delete pointers

PR #5783, opened 2019-08-26
Fix pointer profile inspector

Troy-Ferrell:
## Overview
This PR has two changes:

1) The Pointers profile used `ReordableList` to easily draw the inspector which is awesome...but when drawn as a subprofile, the entire inspector is reallocated every draw. Thus there is no way to save state between renders resulting in the bug #5004 . This PR draws the list manually with our own add/delete buttons similar to speech profile and other similar list inspectors in MRTK

2) A lot of old code still uses BeginHorizontal/Vertical which is both problematic and harder to read. Switching to EditorGUILayout.VerticalScope() {} clearly defines what is being draw in this horizontal/vertical and allows valid clean up if the logic returns early

The PR looks more deadly than it is. It's largely because of all the tab formating with the new using() { }

## Changes
- Fixes: #5004 

## Verification
Navigated all pages and browed basic usage
All tests pass

>
>

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Inspectors/Input/Handlers/SpeechInputHandlerInspector.cs
+++ b/Assets/MixedRealityToolkit.SDK/Inspectors/Input/Handlers/SpeechInputHandlerInspector.cs
@@ -102,93 +102,93 @@ namespace Microsoft.MixedReality.Toolkit.Input.Editor
 
         private void ShowList(SerializedProperty list)
         {
-            EditorGUI.indentLevel++;
-
-            // remove the keywords already assigned from the registered list
-            var handler = (SpeechInputHandler)target;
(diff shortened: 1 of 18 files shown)


wiwei:
@Troy-Ferrel, do you think we should ban usage of BeginHorizontal/indentLevel to reduce mismatched inspector bugs going forward?

Troy-Ferrell:
> @Troy-Ferrell, do you think we should ban usage of BeginHorizontal/indentLevel to reduce mismatched inspector bugs going forward?

Yes, I would say in general. There is one potential edge case. One semi negative thing with the scopes is that you know force variable scope. Generally you can just move things out to be defined top of function or whatever. But of course for some code areas....it can be a jungle to move around with your UI scope definition...but this should be rare

PR #5791, opened 2019-08-27
Improve profile inspector performance by removing a type search bottleneck

lukastoenneMS:
**Note: Solution here doesn't work because Interfaces in the registry may be assignable from the given service, but map to an incompatible profile type.
E.g. InputSimulationService implements both IInputSimulationService and IMixedRealityEyeGazeDataProvider, but IMixedRealityEyeGazeDataProvider maps to MixedRealityEyeTrackingProfile instead of MixedRealityInputSimulationProfile.
Don't have a good way to disambiguate here if the service does not match one of the registered interface types exactly**

## Overview

Profile inspectors can be really slow, due to a lookup function `GetProfileTypesForService` that loops over assets (!) to find profile types matching a given service type. This is becoming a severe bottleneck (as indicated by the internal comment), especially for profiles with a lot of nested "child" profiles for data providers, themes, etc., for example the Interactable profiles.

## Changes

The profile types can be examined based on reflection instead, which is also more reliable in case there is no default profile instance for the service type. This can be done statically once, so the inspector doesn't have to loop over assets or types in every single draw call.

## Verification

Create a new service or data provider or Interactable theme. If the service has a dedicated profile type, as indicated by the `MixedRealityServiceProfile` attribute, then the profile selector should only show this type of profile.

For example:
- Select the MixedRealityToolkit object
- Select the Input subprofile in the inspector
- Add a new Data Provider
- Set the data provider type to Microsoft.MixedReality.Toolkit.Input.InputSimulationService
- Open the profile selector, it should only show input simulation profiles

### DIFF
--- a/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
+++ b/Assets/MixedRealityToolkit/Inspectors/Profiles/BaseMixedRealityProfileInspector.cs
@@ -22,6 +22,31 @@ namespace Microsoft.MixedReality.Toolkit.Editor
 
         private static StringBuilder dropdownKeyBuilder = new StringBuilder();
 
+        private static readonly Dictionary<Type, Type> ProfileTypeForService;
+
+        // Static constructor initializes some static members
+        static BaseMixedRealityProfileInspector()
+        {
+            ProfileTypeForService = new Dictionary<Type, Type>();
+
+            foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
+            {
+                foreach (var type in assembly.GetTypes())
+                {
+                    if (type.IsSubclassOf(typeof(BaseMixedRealityProfile)))
+                    {
+                        foreach (MixedRealityServiceProfileAttribute serviceProfileAttribute in type.GetCustomAttributes(typeof(MixedRealityServiceProfileAttribute), true))
+                        {
+                            if (!ProfileTypeForService.ContainsKey(serviceProfileAttribute.ServiceType))
+                            {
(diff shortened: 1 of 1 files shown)


wiwei:
> **Note: Solution here doesn't work because Interfaces in the registry may be assignable from the given service, but map to an incompatible profile type. E.g. InputSimulationService implements both IInputSimulationService and IMixedRealityEyeGazeDataProvider, but IMixedRealityEyeGazeDataProvider maps to MixedRealityEyeTrackingProfile instead of MixedRealityInputSimulationProfile. Don't have a good way to disambiguate here if the service does not match one of the registered interface types exactly**

Is this what IsProfileForService tries to address, or is that not really applicable anymore given the changes to use reflection?

lukastoenneMS:
> 
> 
> > **Note: Solution here doesn't work because Interfaces in the registry may be assignable from the given service, but map to an incompatible profile type. E.g. InputSimulationService implements both IInputSimulationService and IMixedRealityEyeGazeDataProvider, but IMixedRealityEyeGazeDataProvider maps to MixedRealityEyeTrackingProfile instead of MixedRealityInputSimulationProfile. Don't have a good way to disambiguate here if the service does not match one of the registered interface types exactly**
> 
> Is this what IsProfileForService tries to address, or is that not really applicable anymore given the changes to use reflection?

Yes, the reflection-based method is an attempt at solving the performance problem, but doesn't work because of the service->profile type mapping ambiguity. I'm not sure why this works when mapping via asset files, it might just be an ordering thing, where the input sim profile (alphabetically) comes first and so you don't get back the eye tracking profile when asking for the input sim service.

I'm sure this is solvable some way, but i haven't looked into it any further yet.

wiwei:
Closing out old PRs

Issue #5772, opened 2019-08-23
Migration and breaking changes mitigations


wiwei:
Filing an issue to that we need to have a proposal written for how we're going to manage breaking changes and other migitations that we need to provide to customers as they upgrade from build to build.

The nature of the MRTK is one where MUCH of the code is public API surface, so by definition many things are potentially breaking. Some changes are tougher to stomach than others, but now that we have a definition for what breaking changes are:

https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/BreakingChanges.html

We need to have some proposal and thoughts for how we can lessen the impact of those sorts of changes on consumers.

MenelvagorMilsom:
So I began trying this in Python (PyYAML) yesterday evening and this morning. Here are some notes so far:
- Reading Unity's YAML files is a little more awkward than expected.
  - From what I can see `%TAG` shorthands in YAML are valid per `---` document and not per file, but Unity files have only one `%TAG` at the top of the file. This causes errors with PyYAML. This should be ok however, we don't need any information from the tags so this line and all !u![0-9]* strings from the start of `---` documents.
  - PyYAML wasn't capturing the `fileID` specified at the beginning of every `---` document. This is needed in order to reference existing objects and fields. This can be amended by converting each `---` document in the file into a YAML object where the name is `_<fileID>`. This way, we only have one document in the YAML string. At the moment I've stripped away some information that I'll need in order to reconstruct the file, but it should be simple to add it in a way that doesn't interfere.
  - Both of these fixes can be done in a single pass over the document. Once these changes have been made to the YAML string, we should be in a position to make the changes we need.
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td nowrap>
      <pre><code>%YAML 1.1
%TAG !u! tag:unity3d.com,2011:
--- !u!29 &1
OcclusionCullingSettings:
  m_ObjectHideFlags: 0
  serializedVersion: 2
  m_OcclusionBakeSettings:
    smallestOccluder: 5
    smallestHole: 0.25
    backfaceThreshold: 100
  m_SceneGUID: 00000000000000000000000000000000
  m_OcclusionCullingData: {fileID: 0}</code></pre>
    </td>
    <td nowrap>
      <pre><code>%YAML 1.1
---
_1:
  OcclusionCullingSettings:
    m_ObjectHideFlags: 0
    serializedVersion: 2
[...]

Issue #4213, opened 2019-05-07
Make hand mesh / joint visualization a platform setting


thalbern:
## Describe the problem

On Hololens 2 we usually don't want to visualize hand mesh / joints for performance reason, however for eg the hand input simulation in editor we still want to show the joints. Currently you have to modify the profile (or setting in the example scene) to switch the mesh / joints on / off whenever you switch your debugging / testing scenario from editor to hololens 2. It would be great to have this as a per platform setting to avoid having to change the profile everytime you're deploying on device.

## Describe the solution you'd like

A per platform setting for hand mesh / joing visualization.

Yoyozilla:
This would be nice to have. We should also add a doc in the perf section to note this.

Yoyozilla:
Docs needs to be soon.

wiwei:
@Yoyozilla, is the Urgency-Soon issue for the documentation guidelines, or for everything?

Yoyozilla:
urgency soon on Doc.

Yoyozilla:
removing urgency soon as we have the HL2 default profile now.

wiwei:
This is still a painpoint for folks using this, and I'm plus1ing this because folks are still keen on getting this fixed.

The hand joint problem is still there even with the HL2 profiles

PR #5797, opened 2019-08-27
Make the mesh/joint visualization have independent editor/playersettings

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4213

This makes it so that the hand mesh/joint visualization is has a separate editor and player setting - this allows people have to hand joint and mesh in-editor (for in-editor simulation) without also having the same setting in the actual player. Having these settings on in-player has a perf cost, so we typically recommend that people have these features turned off when actually deployed to device.

This change updates the profile by having a flag based config for this (i.e. none, editor, player, or both). Note that this is obviously not backwards compatible, and thus this is a breaking change, to which I'm going to add a migration script (working with @MenelvagorMilsom and @thalbern on this one) once some of the migration infrastructure is a little clearer. The goal here is to make it so that people can just run a script and get the right changes to their existing profiles.

See https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5772 for tracking of breaking changes and how we deal with migrating them

This also changes the mechanics of EnableHandMeshVisualization and EnableHandJointVisualization to only affect the currently active application mode (so that doing things like using the in-editor simuilation and the hand visualization toggling script won't affect player settings).

image


Documentation/InputSystem/HandTracking.md
```diff
@@ -25,6 +25,17 @@ The hand mesh is used if fully defined mesh data is provided by the hand trackin
 
 Hand mesh display can have a noticeable performance impact, for this reason it can be disabled entirely by unchecking **Enable Hand Mesh Visualization** option.
 
+## Hand Visualization Settings
```
thalbern:
```suggestion
## Hand visualization settings
```

Issue #5395, opened 2019-07-20
Gaze Cursor's SetVisibility() not working


Pycorax:
## Bug Description
`MixedRealityToolkit.InputSystem.EyeGazeProvider.GazeCursor.SetVisibility()` does not do anything.

## To Reproduce

1. Call `MixedRealityToolkit.InputSystem.EyeGazeProvider.GazeCursor.SetVisibility(false);`,
2. Observe the cursor not turning invisible

## Expected behavior
Upon calling `MixedRealityToolkit.InputSystem.EyeGazeProvider.GazeCursor.SetVisibility(false);`, the cursor should turn invisible.

## Setup

- Unity Version 2018.40f1
- MRTK Version 2.0.0 RC2.1

## Target Platform
- HoloLens

## Additional context
Calling `MixedRealityToolkit.InputSystem.EyeGazeProvider.GazeCursor.GameObjectReference.SetActive(false);` seems to work as a temporary workaround to achieve the desired effect of hiding the cursor.

michel-de-verdelhan:
I have the same problem. I found that the Gaze cursor visibility is actually driven by 
FocusProvider.ReconcilePointers() which re enable it based on it's internal state machine. I think adding an option to disable this behaviour would be helpful

timGerken:
@wiwei, is there a way to work around this in the 2.0.0 release without waiting for 2.0.1?

In 2.0.0-RC2.1 I was able to call `inputSystem.GazeProvider.Enabled = state;` to effectivly show/hide the gaze cursor, but when moving to 2.0.0 GA, apparently the FocusProvider has a dependency on the GazeProvider being enabled, and leads to the following exception:
```
NullReferenceException: Object reference not set to an instance of an object
Microsoft.MixedReality.Toolkit.Input.FocusProvider.UpdatePointer (Microsoft.MixedReality.Toolkit.Input.FocusProvider+PointerData pointerData) (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:903)
Microsoft.MixedReality.Toolkit.Input.FocusProvider.UpdatePointers () (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:840)
Microsoft.MixedReality.Toolkit.Input.FocusProvider.Update () (at Assets/MixedRealityToolkit.Services/InputSystem/FocusProvider.cs:517)
[...]

Issue #4954, opened 2019-06-18
Provide example of how to enable / disable hand rays at runtime


julenka:
## Overview
We have had two requests for how to disable the far interaction (line pointer, hand rays, etc) at runtime. For example, this [stackoverflow question](https://stackoverflow.com/questions/56248329/how-to-turn-off-motion-controller-line-pointer/56435895#56435895).

Since it is fairly involved, it would be good to provide an example for how to enable / disable the following pointers at runtime:

- hand rays / motion controller line pointer
- poke pointer
- grab pointer / sphere pointer.

Troy-Ferrell:
@julenka extending off of this, there has been multiple people asking how to disable hand-related pointers, basically just keep GGV. As well as how to disable eyes. All of this at runtime dynamically of course.

Some of this stuff is easily do-able (GazeProvider.UseEyeGaze) but it is not obvious or documented. It ma be good to have both 
1) A clear mechanism/code path/design arch in MRTK to turn on pointers/services/etc
2) Dedicated documentation page for how to disable/enable various inputs

julenka:
I actually have an example that allows you to turn on / off ggv pointer, hand ray pointer, sphere pointer from code, and to have that update without needing to require hiding hands. I haven't gotten it up in MRTK but in case it helps, I made a custom pointer mediator class:

```
using Microsoft.MixedReality.Toolkit.Input;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;

namespace Assets.Julia
{
    public class MyPointerMediator : DefaultPointerMediator
    {
        public bool IsGrabEnabled { get; set; } = true;
        public bool IsPokeEnabled { get; set; } = true;
        public bool IsHandRayEnabled { get; set; } = true;
        public bool IsHeadGazeEnabled { get; set; } = false;


        public override void UpdatePointers()
        {
            base.UpdatePointers();
            if (!IsPokeEnabled || !IsGrabEnabled)
            {
                foreach (var item in nearInteractPointers)
                {
                    if (!IsPokeEnabled && item is PokePointer)
                    {
                        item.IsActive = false;
                    }
                    if (item is SpherePointer && !IsGrabEnabled)
                    {
                        item.IsActive = false;
                    }
                }
            }

            if(!IsHandRayEnabled || !IsHeadGazeEnabled)
            {
                foreach (var item in farInteractPointers)
                {
[...]

PR #5799, opened 2019-08-27
Provide a way to override the gaze cursor visibility

wiwei:
https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5395

This updates IMixedRealityGazeProvider to provide an override for the gaze cursor visibility.

Note that this doesn't fix the issue by making GazeCursor.SetVisibility "just work", because the thing that controls the gaze cursor visibility is a state machine controlled by higher level concepts than the cursor itself. I thought about adding some sort of mediation/more complicated set of flags (i.e. IMixedRealityCursor.SetVisibilityOverride) but found that doing such a thing would end up making things more complicated on that level (i.e. it introduces complexity in an area that SHOULD be simple, rather than building complexity at the more appropriate layers)

This change adds tests to show that it works, and then also adds docs to explain how to use this.

### DIFF
--- a/Assets/MixedRealityToolkit.Tests/PlayModeTests/InputSystem/GazeProviderTests.cs
+++ b/Assets/MixedRealityToolkit.Tests/PlayModeTests/InputSystem/GazeProviderTests.cs
@@ -1,0 +1,108 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#if !WINDOWS_UWP
+// When the .NET scripting backend is enabled and C# projects are built
+// The assembly that this file is part of is still built for the player,
+// even though the assembly itself is marked as a test assembly (this is not
+// expected because test assemblies should not be included in player builds).
+// Because the .NET backend is deprecated in 2018 and removed in 2019 and this
+// issue will likely persist for 2018, this issue is worked around by wrapping all
+// play mode tests in this check.
+
+using NUnit.Framework;
+using System.Collections;
+using UnityEngine.TestTools;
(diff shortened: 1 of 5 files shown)


Assets/MixedRealityToolkit.Tests/PlayModeTests/InputSystem/GazeProviderTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            // Showing the hand should cause the gaze cursor to be not visible.
+            var inputSimulationService = PlayModeTestUtilities.GetInputSimulationService();
+            yield return PlayModeTestUtilities.ShowHand(Handedness.Right, inputSimulationService);
+
+            // Verify that since the hand is shown, the gaze cursor is not visible.
+            Assert.IsFalse(CoreServices.InputSystem.GazeProvider.GazeCursor.IsVisible);
+
+            // Set the override to true, and then wait a frame in order for the
+            // internal state machine to reconcile this property change.
+            CoreServices.InputSystem.GazeProvider.GazeCursorVisibilityOverride = false;
```
thalbern: it says in the comment that the override is set to true, but you're setting it to false?

thalbern: also shouldn't the asset fail after setting it to false because the gaze cursor should be hidden?

wiwei:
Chatted with @julenka offline, this is going to be rolled up into https://github.com/microsoft/MixedRealityToolkit-Unity/issues/4954 instead of having its own one-off solution.

I think that in making this change, I was too focused on this specific scenario, and wasn't thinking about the other scenarios. I'm going to close this out and dupe it out to the other one.

PR #5808, opened 2019-08-28
Cleaned up NuGet Package Duplicates

andreiborodin:
Removed copies of License/Version from folders packages together, and also updates the nuspec to exclude the yoc and the StandardAssets.meta.


davidkline-ms:
:shipit:

Issue #5770, opened 2019-08-23
custom profiles created prior to v2.0.0 are not recognized


gilbdev:
## Describe the bug
During RC2.1 to v2.0.0 migration, profile that were previously created (in RC1 or RC2) could not be selected as configuration profile (not listed when clicking on the litttle circle to select asset corresponding to type).

Also in the project window, the old profile does not display the blue mrtk icon, but the default unity asset  grey one.

## To reproduce

1. create any custom profile in RC2
2. follow migration step to v2.0.0
3. when adding mrtk to scene and configure, ythe 'old' profile won't show

## Expected behavior

old profile should be listed in the list of configuration profile just like new ones.

## Workaround
old profile can still be drag and dropped.
by selecting, modifying and saving a profile, the icon turns 'blue' and issue seems to fixed. (this may be a hint to make an automatic migration procedure that select all profile assets and force save them ?)

## Your Setup (please complete the following information)

- Unity Version 2018.4.2f1
- MRTK Version RC2.1 to v2.0.0

## Target Platform (please complete the following information)

- HoloLens

wiwei:
I was able to repro this as well - what's happening here appears to be that the asset database isn't being fully refreshed after importing the new package.

In particular, if you follow the instructions here:
https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/ReleaseNotes.html#upgrading-projects-to-200

After the step 5 and before step 6: For each scene in the project...

There should be another step, which involves closing Unity again and deleting the Library/ folder. This forces an asset database refresh, which will make all of the profiles "whole" again (in terms of Unity knowing that they are really, actually, for real, those types of profiles).

I think there is an alternative here where we could either do:
1) Add some post-package import callback to force an asset database refresh (though I'm not sure that this will work because I don't *think* that a package can get a callback for its own install completion).
2) Add some sort of MRTK menu button to force a "post install" button - the only bad thing about this is that it would reserve some relatively valuable real estate (i.e. the menus) for a thing you only need to do once on upgrade (and never again during regular flows).

I'm going to update the documentation to call out this extra step (and also say why it's needed)

@gilbdev FYI

wiwei:
Also note that this is why saving or just moving/renaming and moving things back would fix things up (it triggers an asset database refresh for that file)

PR #5813, opened 2019-08-28
Update the migration/upgrade notes with correct instructions on how to upgrade between MRTK versions.

wiwei:
#5770

The existing instructions weren't quite complete, because when you approached step #6 (re-adding MRTK objects to the scene) there is a strong likelihood that the profiles (and other assets outside of the MRTK) haven't been refreshed in the asset database. As a result, doing things like adding the MRTK to the scene and picking a profile wouldn't pick up existing profiles (because the asset database wouldn't have the correct information.)

This requires forcing Unity to regenerate its asset database - the easiest way to do this is just to close it and delete the Library/ folder (which causes an asset database refresh)

See https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5770#issuecomment-525947248 for alternatives considered


wiwei:
Note that I need to cherry pick this into a couple of branches

PR #5819, opened 2019-08-29
Fix Platform Toolset selection not applied for msbuild

julianloehr-kg:
## Overview

This is a follow up fix for #5360. The `PlatformToolset` property of the `UwpBuildInfo` wasn't set when starting an appx build. Leaving it to the default empty string value, `/p:PlatformToolset` was ommited during msbuild invokation.

In addition, this PR also contains an fix for the copy-pasted `PlatformToolset` preferences API in-code documentation.

## Changes
- Follow up fix for #5360


wiwei:
lgtm, thanks @julianloehr-kg

PR #5822, opened 2019-08-29
Remove the SimulatedHandData timestamps since they prevent changes in…

lukastoenneMS:
## Overview

Remove the timestamps from SimulatedHandData because they are breaking tests.

These timestamps are supposed to prevent frequent changes, but they do this in tests as well, where multiple changes per frame are possible. Due to the timestamps being based on UtcNow (which is also quite inaccurate) they can prevent updates randomly.

## Changes
- Fixes: Test failures observed in #5264

### DIFF
--- a/Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
+++ b/Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
@@ -157,14 +157,8 @@ namespace Microsoft.MixedReality.Toolkit.Input
                 // TODO implement custom hand device update frequency here, use 1000/fps instead of 0
                 if (msSinceLastHandUpdate > 0)
                 {
-                    if (HandDataLeft.Timestamp > lastHandUpdateTimestamp)
-                    {
-                        UpdateHandDevice(profile.HandSimulationMode, Handedness.Left, HandDataLeft);
-                    }
-                    if (HandDataRight.Timestamp > lastHandUpdateTimestamp)
-                    {
-                        UpdateHandDevice(profile.HandSimulationMode, Handedness.Right, HandDataRight);
-                    }
+                    UpdateHandDevice(profile.HandSimulationMode, Handedness.Left, HandDataLeft);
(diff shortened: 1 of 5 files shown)


Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationService.cs
```diff
@@ -157,14 +157,8 @@ public override void LateUpdate()
                 // TODO implement custom hand device update frequency here, use 1000/fps instead of 0
                 if (msSinceLastHandUpdate > 0)
                 {
-                    if (HandDataLeft.Timestamp > lastHandUpdateTimestamp)
```
lukastoenneMS: **Note**: ManipulationHandlerThrow test is currently broken by this, because the way hand velocity is calculated relies on UpdateVelocity being called no more than once per frame.

lukastoenneMS: Fixed the throwing test. It was doing an additional WaitForFixedUpdate after releasing, which recomputes the hand velocity filtering and makes it deviate from the expected rigid body velocity. Previously this wasn't happening because the hand data update was blocked by the timestamps.

Assets/MixedRealityToolkit.Tests/PlayModeTests/TestHand.cs
```diff
@@ -61,7 +61,10 @@ public IEnumerator MoveTo(Vector3 newPosition, int numSteps = 30, bool waitForFi
         {
             Vector3 oldPosition = position;
             position = newPosition;
-            yield return PlayModeTestUtilities.MoveHandFromTo(oldPosition, newPosition, numSteps, gestureId, handedness, simulationService);
+            for (var iter = PlayModeTestUtilities.MoveHandFromTo(oldPosition, newPosition, numSteps, gestureId, handedness, simulationService); iter.MoveNext(); )
```
lukastoenneMS: I have changes these to yield each individual step, which makes it easier to investigate test failures.

Issue #5748, opened 2019-08-21
Focus position locks on select gesture to spatial mesh 


CDiaz-MS:
## Describe the bug

The focus position gets locked to the spatial mesh on select gesture.

## To reproduce

1. Create new scene with MRTK config
2. Enable Spatial Awareness System, make sure input simulation is enabled 
3. Using the in-editor hand input simulation, move the hand focus on the spatial mesh
4. Click the left mouse to switch to the select gesture
5. Move the mouse and the focus position gets locked on the spatial mesh

## Expected behavior

If an object is not interactable, then focus should not lock position on the mesh.  Currently the focus gets locked on select with the spatial mesh.

## Additional context

To fix this behavior you comment out pointer.IsFocusLocked in MixedRealityInputSystem.cs:

```
        public void RaisePointerDown(IMixedRealityPointer pointer, MixedRealityInputAction inputAction, Handedness handedness = Handedness.None, IMixedRealityInputSource inputSource = null)
        {
            //pointer.IsFocusLocked = (pointer.Result?.Details.Object != null);

            pointerEventData.Initialize(pointer, inputAction, handedness, inputSource);

            HandlePointerEvent(pointerEventData, OnPointerDownEventHandler);
        }
```

Issue #5621, opened 2019-08-12
SpherePointer to cache TryGetNearGraspPoint


Troy-Ferrell:
## Describe the bug

SpherePointer leverages TryGetNearGraspPoint() for many commonly used properties (i.e IsNearObject and IsInteractionEnabled). These properties are called multiple times which causes the TryGetNearGraspPoint and resulting Physics.CheckSphere to be redundantly called everytime.

## To reproduce

Read SpherePointer.cs => IsNearObject & IsInteractionEnabled

## Expected behavior

For a given frame update, the results of TryGetNearGraspPoint and Physics.CheckSphere could potentially be cached. The checksphere may be problematic if objects are instantiated during an update loop. But the TryGetNearGrapsPoint should not be altering.

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.3.11f1]
- MRTK Version [e.g. v2.0]

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

julenka:
fixed.

Issue #4203, opened 2019-05-07
Hand ray disappears when hand gets near SR


chbecker-ms:
Hand rays should only disappear when hands are near grababble or touchable things, not when the hand is near a non-interactive element like SR (the surface mesh).

## Additional Context
This is happening because SpherePointer.IsNearGrabbable will return true if it is near ANY collidable, but it should only return true if it is near something that is grabbable (has NearInteractionGrabbable component on it).

And instead of doing an extra sphere cast we should just take advantage of focus system and just check if the sphere pointer is focusing anything.

## Workaround
Here's how you can work around this issue, for the SR mesh:

```
LayerMask toIgnore = LayerMask.GetMask("Spatial Awareness") | UnityEngine.Physics.IgnoreRaycastLayer;return UnityEngine.Physics.CheckSphere(position, NearObjectRadius, ~toIgnore);
```
Add that in SpherePointer.cs wherever Physics.CheckSphere runs.

wiwei:
This bug was filed a while back I think in response to some code review. At the time this was brought up, but the author actually fixed it in the same review and so this bug immediately went stale.

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Cursors/FingerCursor.cs#L199

The original bug references this section, which used to do a sphere raycast (instead of just an interactiongrabbable search)

tl;dr this was fixed in the same PR that this issue was filed against so it's not relevant anymore

julenka:
suggestion for fix: create a 'grabbable layer mask' similar to what the [poke pointer does with poke layer masks](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/PokePointer.cs), and in SpherePointer.IsNearObject, query using the grab layer mask filter.

Addison had previously created this solution, but then I reverted it to reduce risk from his change, because the layer mask was unrelated to his PR. We should bring that change back in.

julenka:
Marking urgency soon since it blocked several people who were using sr meshes in MRTK.

julenka:
fixed.

PR #5823, opened 2019-08-29
Turn off hand rays only when hand is near a grabbable object

julenka:
## Overview
If the hand is near the spatial mesh (SR), hand rays should not turn off. In fact, rays should only turn off if the hand is near a grabbable or touchable thing. This PR updates SpherePointer to return `IsNearObject` and `IsInteractionEnabled` only when the hand is near a collider that has a NearInteractionGrabbable on it. It also adds Grab Layer Masks, which allows developers to further prioritize and filter which objects a pointer can interact with. 

After this change, an object must both be on a Grabbable Layer as well as have a NearInteractionGrabbable component in order for hand rays to turn off. The Grabbable Layer is by default set to everything **except** Ignore Raycast and Spatial Awareness.

Fixes: #5748, #4203, #5621 

## Changes
- Add a Grabbable Layer Mask to Sphere Pointer to allow devs to easily disable all grabbable objects, and to improve performance
- The default Grabbale Layer Mask for the grab pointer is everything but Ignore Raycast and Spatial Awareness so that sphere casts don't check the spatial mesh (for perf).
- I also am caching the sphere query results to ensure we only do 3 sphere casts per frame (down from 11 without caching).
-  I snuck in a fix to pointer locking -- previously a pointer would lock when you clicked on any collidable, even if the collidable did not actually have any pointer handlers on it, or above it. now, the pointer will only lock if the pointer is actually grabbing an object that would handle pointer events. I may actually end up pulling that change into a separate PR.
- Add tests to verify my fix, and documentation reflecting the new grabbable behavior
- I also fixed the sphere pointer inspector -- it was actually not displaying several sphere pointer fields that it should have
- I removed the DebugSphere because it was not very useful, added a bit of extra code, and had bugs (did not turn off when debugsphere == false).

## Verification
- All playmode tests pass, new tests pass
- Verified that my new tests would have failed in old code
- Verified that my change actually fixes the bug (ray no longer hides when near SR)
- Runs on HoloLens 2 with no adverse perf impact.


Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
```diff
@@ -88,20 +116,21 @@ public override void OnPreSceneQuery()
...
+                var toQuery = new SpherePointerQueryInfo[] { queryBufferNearObjectRadius, queryBufferInteractionRadius };
+                for (int j = 0; j < toQuery.Length; j++)
                 {
-                    if (debugSphere == null)
+                    for (int i = 0; i < layerMasks.Length; i++)
                     {
-                        debugSphere = GameObject.CreatePrimitive(PrimitiveType.Sphere).transform;
-                        debugSphere.localScale = Vector3.one * SphereCastRadius * 2;
-                        Destroy(debugSphere.gameObject.GetComponent<Collider>());
+                        if (toQuery[j].TryUpdateQueryBufferForLayerMask(layerMasks[i], pointerPosition, triggerInteraction))
```
julenka: I think a for loop over toQuery here is excessive, remove

davidkline-ms:
> After this change, an object must both be on a Grabbable Layer as well as have a NearInteractionGrabbable component

why the layer requirement? can the same be accomplished by just checking for NearInteractionGrabbable/Touchable?

julenka:
> > After this change, an object must both be on a Grabbable Layer as well as have a NearInteractionGrabbable component
> 
> why the layer requirement? can the same be accomplished by just checking for NearInteractionGrabbable/Touchable?

The default Grabbale Layer Mask for the grab pointer is everything but Ignore Raycast and Spatial Awareness so that sphere casts don't check the spatial mesh (for perf). The poke pointer already uses this approach and I wanted to mimic it. Additionally, this will make it easy for devs to disable large collections of UI by just changing layers without needing to go down and remove grabbables from every component.

Troy-Ferrell:
FYI @julenka I believe this also fixes #5621

Documentation/Input/Pointers.md
```diff
@@ -90,6 +89,7 @@ Useful Poke Pointer properties:
 - *TouchableDistance*: Maximum distance a which a touchable surface can be interacted with
 - *Visuals*: Game object used to render finger tip visual (the ring on finger, by default).
 - *Line*: Optional line to draw from fingertip to the active input surface.
+- *Poke Layer Masks* - A prioritized array of LayerMasks to determine which possible GameObjects a the pointer can interact with and the order of interaction to attempt. Note that a GameObject must also have a `NearInteractionTouchable` component in order to interact with a poke pointer.
```
thalbern:
```suggestion
- *Poke Layer Masks* - A prioritized array of LayerMasks to determine which possible GameObjects the pointer can interact with and the order of interaction to attempt. Note that a GameObject must also have a `NearInteractionTouchable` component in order to interact with a poke pointer.
```

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/Pointers/SpherePointer.cs
```diff
@@ -16,7 +17,7 @@ public class SpherePointer : BaseControllerPointer, IMixedRealityNearPointer
 
         [SerializeField]
         [Min(0.0f)]
-        [Tooltip("Additional distance between SphereCastRadius and NearObjectRadius")]
+        [Tooltip("Additional distance on top of sphere cast radius when pointer is considered 'near' an object and far interaction will turn off")]
         private float nearObjectMargin = 0.2f;
         /// <summary>
         /// Additional distance between <see cref="BaseControllerPointer.SphereCastRadius"/> and <see cref="NearObjectRadius"/>.
```
julenka:
```suggestion
        /// Additional distance on top of<see cref="BaseControllerPointer.SphereCastRadius"/> when pointer is considered 'near' an object and far interaction will turn off.
```

julenka:
Completing PR. All checks passed with my code change. Re-ran only docs test because I only made changes to comments.

davidkline-ms:
@julenka, is this a 2.2.0 fix? If so, I can cherry-pick it into the stabilization branch.

keveleigh:
@davidkline-ms This was in the 2.1 release: 
image

Issue #5828, opened 2019-08-30
Support for Both Hand controllers


Alexees:
## Describe the problem

While implementing the Oculus Go and the Oculus Quest, two things arose. First of all, the Oculus Go uses one controller for both hands, just one at a time.
Secondly, the "Oculus Integrations" packages provides one Prefab with all controller models for all Oculus platforms, which means the Quest uses the same prefab for both hands, just configures them differently (I did this using the ControllerVisualizer class).

I went for left first in case there's a check for left or right controller in the case of both.

## Describe the solution you'd like

I like to be able to choose "Both Hands" for a controller, so it does not matter which hand it's spawned for.

## Describe alternatives you've considered

Adding the same controller, once for left, once for right. But that's rather unnecessarily tedious

## Additional context

I already did that, just wanted to get my upcoming pull request its context first.

davidkline-ms:
Question: Does Handedness.Any work for this scenario?

Alexees:
> Question: Does Handedness.Any work for this scenario?

I guess so, did not consider this. The comment on both says:

> The controller is identified as being either left and/or right handed.

and the comment for other is
> Reserved, for systems that provide alternate hand state.

Since we are talking about left and right hand only I thought it should be as precise as that. I'm not even sure what an alternate hand state could be, but in this case it's definitely not.
If it does not matter though, I'll use any.

Zee2:
Closing as stale. I believe this is essentially irrelevant for XRI (and the Go is now discontinued anyway... :( )

Please reopen if this is still a critical issue.

PR #5830, opened 2019-08-30
Both hand controller model support

Alexees:
- Fixes: #5828


Assets/MixedRealityToolkit/Inspectors/Utilities/HandednessInspectorGUI.cs
```diff
@@ -0,0 +1,29 @@
+﻿
```
julenka: Even though it's technically not needed any more, for consistency please add a copyright block as described in [coding guidelines here](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/CodingGuidelines.html#script-license-information-headers)

Assets/MixedRealityToolkit/Inspectors/Utilities/HandednessInspectorGUI.cs
```diff
@@ -0,0 +1,29 @@
...
+using UnityEditor;
+using UnityEngine;
+
+public static class HandednessInspectorGUI
```
julenka: Please add to the appropriate namespace -- MRTK classes should be in a mrtk namespace. Please see [this guideline](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/CodingGuidelines.html#mrtk-namespace-rules) for more.

Assets/MixedRealityToolkit/Inspectors/Utilities/HandednessInspectorGUI.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    private static readonly GUIContent[] HandednessSelections =
+    {
+        new GUIContent("Left Hand"),
+        new GUIContent("Right Hand"),
+        new GUIContent("Both Hands"),
+    };
+
+    public static void DrawControllerHandednessDropdown(SerializedProperty controllerHandedness)
```
julenka: Please comment public methods, [refer to this guideline](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/DocumentationGuide.html#property-method-event-summary-blocks)

julenka: It would be good to mention that this actually modifies the input controllerHandedness as well.

Assets/MixedRealityToolkit/Inspectors/Utilities/HandednessInspectorGUI.cs
julenka: I stand corrected!

https://docs.opensource.microsoft.com/content/releasing/copyright-headers.html#copyright-headers-in-microsoft-oss-project

"Microsoft should not add a Microsoft copyright header to third party contributions to Microsoft OSS projects. We should never remove copyright notices from files."

No need to add this.

Assets/MixedRealityToolkit/Providers/BaseController.cs
```diff
@@ -136,7 +136,7 @@ public bool SetupConfiguration(Type controllerType, InputSourceType inputSourceT
                             profileFound = true;
 
                             // If it is an exact match, assign interaction mappings.
-                            if (controllerMappings[i].Handedness == ControllerHandedness &&
+                            if ((controllerMappings[i].Handedness & ControllerHandedness) != 0 &&
```
SimonDarksideJ: Be aware, that Controllers are not detected as "both", so where you might have a mapping defined as both, the Platform will still report the "hand" as either left or right.  Better to simply test if you have a "Both" definition to apply to either left or right, e.g.
> controllerMappings[i].Handedness == ControllerHandedness || controllerMappings[i].Handedness == Handedness.Both

Assets/MixedRealityToolkit/Inspectors/Utilities/HandednessInspectorGUI.cs
Alexees: > I don't feel a separate file is really needed in this case as it's only used by a single area, but abstraction can help the flow of code.
> 
> My recommendation would be to put the function inside the inspector class, but keep it as a separate function for clarity and not create a new class file.

Do you mean DrawControllerHandednessDropdown? It's used in two classes, ControllerPoseSynchronizerInspector and MixedRealityControllerVisualizationProfileInspector. If you want I create a seperate pull request that deals with this extraction

Alexees:
> I don't feel a separate file is really needed in this case as it's only used by a single area, but abstraction can help the flow of code.
> 
> My recommendation would be to put the function inside the inspector class, but keep it as a separate function for clarity and not create a new class file.

Do you mean DrawControllerHandednessDropdown? It's used in two classes, ControllerPoseSynchronizerInspector and MixedRealityControllerVisualizationProfileInspector. If you want I create a seperate pull request that deals with this extraction

Assets/MixedRealityToolkit/Providers/BaseController.cs
Alexees: That's what the binary comparison is for, is it not? If controllerMappings[i].Handedness is "Both" it does not matter what hand ControllerHandedness is, it works for both.

SimonDarksideJ: I would ask why you need to set it in the PoseSynchroniser?  Mainly because no SDK has a "Both" setting from the SDK, Both is an approximation for the toolkit to enable configuration that will work in either hand.

Once a controller is being visualised, it's either in the Left or Right hand, as determined from the underlying SDK

Assets/MixedRealityToolkit/Inspectors/Utilities/HandednessInspectorGUI.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.MixedReality.Toolkit.Utilities.Editor
+{
+    public static class HandednessInspectorGUI
+    {
+        private static readonly GUIContent[] HandednessSelections =
+        {
+        new GUIContent("Left Hand"),
+        new GUIContent("Right Hand"),
+        new GUIContent("Both Hands"),
```
davidkline-ms: Please consider using "Any Hand" as "Both Hands" could be read as the controller in intended to be held in two hands (ex: common gamepad controllers)

Assets/MixedRealityToolkit/Providers/BaseController.cs
davidkline-ms: > controllerMappings[i].Handedness is "Both" it does not matter what hand ControllerHandedness is, it works for both

This is where Any might be a better description.

Assets/MixedRealityToolkit/Inspectors/Utilities/HandednessInspectorGUI.cs
Alexees: Then the comment on the Both enum entry should be rephrased because it reads:
> The controller is identified as being either left and/or right handed.

Assets/MixedRealityToolkit/Providers/BaseController.cs
Alexees: > I would ask why you need to set it in the PoseSynchroniser?

because with the Oculus Go, the controller is symmetrical (ControllerVisualization any hand), and can be held in either hand (PoseSynchronizer any hand).
It also just for convenience, because you can have a single synchronizer class and a single prefab, within which you configure what hand it is individually. Oculus's visuals, for all their platforms, comes within one prefab, so you just toggle their visibility as needed.

Alexees: > This is where Any might be a better description.

@davidkline-ms Any already exists in the enum but also includes Other as an option. This leads me to the question, why not opening this up for all handedness choices you can make?

Issue #5836, opened 2019-08-30
TouchableVolume cannot be touched from all sides


julenka:
## Steps to reproduce
Add a cube to the scene and add a touchable volume to it, configure it to raise pointer events.

Add the following component to see when pointer events get raised:

```csharp
public class PrintPointerEvents : MonoBehaviour, IMixedRealityPointerHandler
{
    public void OnPointerDown(MixedRealityPointerEventData eventData)
    {
        if (eventData.Pointer is SpherePointer)
        {
            Debug.Log($"Grab start from {eventData.Pointer.PointerName}");
        }
        if (eventData.Pointer is PokePointer)
        {
            Debug.Log($"Touch start from {eventData.Pointer.PointerName}");
        }
    }

    public void OnPointerClicked(MixedRealityPointerEventData eventData) {}
    public void OnPointerDragged(MixedRealityPointerEventData eventData) {}
    public void OnPointerUp(MixedRealityPointerEventData eventData) {}
}
```

Move a simulated hand through all 6 faces of the cube.

## Expected Behavior
Touch start happens on all 6 faces.

## Actual Behavior
Touch start only occurs on some of the faces.

## Notes
This bug occurs because the ray that PokePointer builds changes as the finger goes through the surface of the cube. This is because the normal returned by NearInteractionTouchableVolume.DistanceToTouchable is not correct.

Below, the red cross is the ray start, and the cyan cross the result hit point. You can see that as the hand goes through the cube volume, the ray start moves all the way to the other side of the cube, causing the hit point to be large, 

![touchvolume](https://user-images.githubusercontent.com/168492/64040366-b8fd4780-cb11-11e9-8c6d-246af69ec849.gif)

I think if the finger is inside the volume, it would be great if we could not do a raycast and just return - the distance from the finger hit to the closest point on the volume bounds.

lukastoenneMS:
@julenka Thanks for investigating this!

I couldn't really reproduce what you describe, entering any of the 6 cube faces gives a "TouchStarted" event as expected.

However, i noticed that the touchable still recomputes the closest point while you are moving around, and Unity seems to have some kind of quantization going on that makes it snap to particular points on a plane. Every time that happens i get a new touch event too.

[...]

PR #5849, opened 2019-09-02
Use correct center point for computing touchable normal inside a coll…

lukastoenneMS:
## Overview

Fix incorrect normal calculation for NearInteractionTouchableVolume in the case of casting from within a collider.

## Changes
- Fixes: #5836


wiwei:
@lukastoenneMS @julenka thoughts on adding a small test case for this?

wiwei:
Change lgtm - it seems like an easy one to regress going forward.

julenka:
> @lukastoenneMS @julenka thoughts on adding a small test case for this?

I think a test case that does the following would work:

1. Places cubes in 10 random positions around origin (say, within a 2 meter radius from origin)
2. For each cube, touch the cube from all 6 sides, and ensure touch up and down events are received.

To ensure we don't regress behavior in future, I think we should add this test before we close the bug related to this.

Issue #5489, opened 2019-08-01
In-Editor Input simulation does not allow for head tilt/camera roll


hendryems:
## Describe the problem

From the [documentation](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/InputSimulation/InputSimulationService.html), it doesn't look like head tilting (camera roll) is something that's supported:

![image](https://user-images.githubusercontent.com/50596066/62320785-01174480-b456-11e9-8ef2-b2177c92b6d9.png)

Here's what's desired:
![image](https://user-images.githubusercontent.com/50596066/62320882-3d4aa500-b456-11e9-957f-3ea08b39fe2a.png)

This is also something that's supported in the [HL simulator](https://docs.microsoft.com/en-us/windows/mixed-reality/advanced-hololens-emulator-and-mixed-reality-simulator-input):

![image](https://user-images.githubusercontent.com/50596066/62320966-6c611680-b456-11e9-84d7-dbbdcc251698.png)

This is critical to be able to test out code that more accurately mimics user's head movement without having to deploy onto a physical HMD.

## Describe the solution you'd like

Extend `ManualCameraControl.GetCameraControlRotation` ([source](https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Services/InputSimulation/ManualCameraControl.cs)) to include additional keys that supports this:

```
            if (this.isGamepadLookEnabled)
            {
                ...
            }

            KeyboardLookTick(ref rot);

            if (this.ShouldMouseLook)
            {
                ...
            }

```

KeyboardLookTick implementation would be something like this:

```
    private void KeyboardLookTick(ref Vector3 rot)
    {
        // TODO: Make keys customizable from profile
        if (Input.GetKey(KeyCode.Comma))
        {
            rot.z += InputCurve(Time.deltaTime * this.keyboardRotationSensitivity);
        }
        else if (Input.GetKey(KeyCode.Period))
        {
            rot.z -= InputCurve(Time.deltaTime * this.keyboardRotationSensitivity);
        }
    }
```

## Describe alternatives you've considered

N/A

## Additional context

N/A

lukastoenneMS:
[...]

PR #5853, opened 2019-09-03
Unified input methods and new features for input simulation

lukastoenneMS:
## Overview

Input simulation camera control and hand control originated from two separate features and have inconsistent support for rotation as well as slightly different approaches to key binding.

This PR aims to unify the support for keyboard/mouse/game controller input and complement existing features.

## Changes
- Fixes: #5489 
- New  editor tools window to expose input simulation state of head and hands at runtime. This also provides useful features such as "Reset" buttons to bring head/hands back to default state. This window also incorporates the previous "Input Recording" window.
- Optional viewport icons for controlling hands, similar to icons in HTK.
- Introduces a generic `KeyBinding` struct that encodes both KeyCode and mouse button. This comes with editor support for quickly assigning buttons with a key press. (_Note: Unity editor does not support shift key presses on their own, so these have to be assigned using the legacy enum popup).
- Unifies handling of mouse rotations for camera and hands, including cursor locking and visibility. The previous 6 keys for rotating hands have been removed in favor of intuitive mouse rotation and/or changing angles in the editor window.
- Support for roll angle of head and hands, using the scroll wheel during rotation (pitch and yaw are controlled by mouse x/y movement as before).
- New input axis "UpDown" to control camera movement along the Up vector. Defaults to Q/E keys and the 3rd game controller axis (triggers)

**Breaking Changes:** InputSimulationProfile has changed, with some existing KeyCode fields replaced by the new KeyBinding struct (this might reset user profiles to default settings). Some fields have been renamed, but should be handled by FormerlySerializedAs attributes.

## Verification
- Read through input simulation and input animation recording documentation and check for correctness and clarity.
- Run the hand tracking test scene in the editor and try out input simulation features:
  * Camera movement using WASD and game controller
  * Camera rotation by holding RMB
  * Hand movement using LShift/Space + mouse movement
  * Hand rotation using LShift/Space + RMB + mouse movement (it becomes less complicated when toggling hands on using T/Y first)
  * Try camera/hand roll by using the mouse wheel while rotating
  * Move camera up/down using Q/E keys or game controller triggers
- Open the Input Simulation window and try buttons
- Also try assigning new key bindings using the little buttons on the right.
- Try toggling hands via the icons in the viewport in play mode.


julenka:
I've gotten myself into this state, and can't seem my hands. What does this state mean, actually?

image

julenka:
I can't seem to figure out how to roll the hand. When I press shift or space and use mouse  wheel, my hand disappears.

julenka:
Now that mouse wheel rolls the hand, how do you move the hand in Z?

Assets/MixedRealityToolkit.SDK/Features/Input/InputSimulationIndicators.cs
```diff
@@ -0,0 +1,112 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.MixedReality.Toolkit.Utilities;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
```
julenka: Hm...shouldn't this code live near where the input simulation files are?

Assets/MixedRealityToolkit/Definitions/Devices/ControllerMappingLibrary.cs
```diff
@@ -201,6 +206,8 @@ public static class ControllerMappingLibrary
         /// <returns></returns>
         public static InputManagerAxis[] UnityInputManagerAxes => new[]
         {
+            new InputManagerAxis { Name = UP_DOWN, Gravity = 3, Dead = 0.001f, Sensitivity = 3, Snap = true, Invert = false, Type = InputManagerAxisType.KeyOrMouseButton, PositiveButton = "e", NegativeButton = "q" },
+            new InputManagerAxis { Name = UP_DOWN,  Dead = defaultDeadZone, Sensitivity = 1, Invert = false, Type = InputManagerAxisType.JoystickAxis, Axis = 3 },
```
wiwei: What are your thoughts on adding this to the end of the array, instead of the beginning? I'm curious if someone out there is currently using this array to index into position 0 (they expect AXIS_1) but since we're hijacking that existing position, they get broken.

It might be possible to do static analysis on others' code to see if they're using it that way but, if at all possible, at the end is safest (same reason why we add enum values at the end rather than the beginning even if they logically make sense at the beginning)

Assets/MixedRealityToolkit.SDK/Features/Input/InputSimulationIndicators.cs
```diff
@@ -0,0 +1,112 @@
...
+                else
+                {
+                    iconHandRight = iconHandUntrackedRight;
+                }
+
+                imageHandRight.sprite = iconHandRight;
+            }
+        }
+
+        public void ToggleLeftHand()
```
wiwei: >ToggleLeftHand [](start = 20, length = 14)

Nit: Can you also add doccomments to these four functions (i.e. toggle/reset/*)?

Mainly because ToggleLeftHand toggles the visibility (which doesn't come up as a word in the function name) (otherwise, is it toggling its tracking state/not tracking state? - it's a bit ambiguous)

Same comment around reset (i.e. what is it resetting? visibility state? tracked state? simulation state? all up state?)

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
```diff
@@ -0,0 +1,255 @@
...
+        // Value can be used as index in AllCodeNames array.
+        internal static readonly Dictionary<Tuple<KeyType, int>, int> KeyBindingToEnumMap;
+        // Maps enum index to a KeyBinding, for assignment after selecting an enum value.
+        internal static readonly Dictionary<int, KeyBinding> EnumToKeyBindingMap;
+
+        // Static constructor to initialize static fields
+        static KeyBinding()
+        {
+            KeyCode[] KeyCodeValues = (KeyCode[])Enum.GetValues(typeof(KeyCode));
+            string[] KeyCodeNames = Enum.GetNames(typeof(KeyCode));
```
wiwei: >KeyCodeNames [](start = 21, length = 12)

This isn't used in this constructor - was this supposed to be used? (Or can we delete this?)

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
```diff
@@ -0,0 +1,255 @@
...
+                    AddEnumValue(KeyType.Key, (int)kc);
+                }
+
+                AllCodeNames = names.ToArray();
+            }
+        }
+
+        [SerializeField]
+        private KeyType bindingType;
+        public KeyType BindingType => bindingType;
```
wiwei: > public KeyType BindingType => bindingType; [](start = 7, length = 43)

Nit: Consider adding a space after BindingType

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
```diff
@@ -0,0 +1,255 @@
...
+        [SerializeField]
+        private int code;
+        internal int Code => code;
```
wiwei: >internal [](start = 8, length = 8)

Was this supposed to be public? I haven't seen this private/internal combination in the codebase yet.

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
```diff
@@ -0,0 +1,255 @@
...
+        public override string ToString()
+        {
+            string s = "";
+
+            s += bindingType.ToString();
+
+            switch (bindingType)
+            {
+                case KeyType.Key:
+                    s += ": " + ((KeyCode)code).ToString();
```
wiwei: >((KeyCode)code).ToString(); [](start = 32, length = 27)

Another thing to consider in the future:

[https://docs.microsoft.com/en-us/dotnet/api/system.enum.getname?view=netframework-4.8](https://docs.microsoft.com/en-us/dotnet/api/system.enum.getname?view=netframework-4.8)

For this, probably not worth changing to that link.

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
```diff
@@ -0,0 +1,255 @@
...
+            Mouse = 1,
+            Key = 2,
+        }
+
+        /// <summary>
+        /// Enum for interpreting the mouse button integer index.
+        /// </summary>
+        public enum MouseButton : int
+        {
+            Left = 0,
```
wiwei: > Left = 0, [](start = 11, length = 10)

Nit: I think that our guidelines say to include a None value like:

[https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/enum)

I think we can go either way, it would be good to be consistent within the same file (i.e. KeyType.None and MouseButton.None). If only because this would also let us return MouseButton.None from the "TryGet..." function down below in failure case (instead of returning a "valid" mouse button)

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityInputSimulationProfile.cs
```diff
@@ -3,6 +3,7 @@
 
 using Microsoft.MixedReality.Toolkit.Utilities;
 using UnityEngine;
+using UnityEngine.Serialization;
 
 namespace Microsoft.MixedReality.Toolkit.Input
```
wiwei: >MixedReality [](start = 20, length = 12)

Can you also add a link/mention on this issue:

[https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5772](https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5772)

And also a new section here:

[https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Updating.html](https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Updating.html)

To describe the changes here (i.e. deletion hand yaw/rotate controls and how to use that functionality in the new world?)

wiwei: It's cheaper for us to do this as we go along the month than trying to write this all at the end (and potentially forgetting some stuff as we go)

---
In reply to: [320507000](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5853#discussion_r320507000) [](ancestors = 320507000)

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
wiwei: Ah... this was done to keep some consistency with the existing InputSimulationMouseButton, I guess nevermind then.

---
In reply to: [320505346](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5853#discussion_r320505346) [](ancestors = 320505346)

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityInputSimulationProfile.cs
```diff
@@ -87,30 +112,34 @@ public class MixedRealityInputSimulationProfile : BaseMixedRealityProfile
         [Header("Hand Simulation")]
         [SerializeField]
         [Tooltip("Enable hand simulation")]
-        private HandSimulationMode handSimulationMode = HandSimulationMode.Articulated;
-        public HandSimulationMode HandSimulationMode { get => handSimulationMode; set => handSimulationMode = value; }
+        private HandSimulationMode defaultHandSimulationMode = HandSimulationMode.Articulated;
```
wiwei: >defaultHandSimulationMode [](start = 35, length = 25)

Is it possible to FormerlySerializedAs here? It looks like this still reasonably equivalent.

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityInputSimulationProfile.cs
```diff
@@ -87,30 +112,34 @@ public class MixedRealityInputSimulationProfile : BaseMixedRealityProfile
...
+        private KeyBinding leftHandManipulationKey = KeyBinding.FromKey(KeyCode.LeftShift);
+        public KeyBinding LeftHandManipulationKey => leftHandManipulationKey;
         [SerializeField]
         [Tooltip("Key to manipulate the right hand")]
-        private KeyCode rightHandManipulationKey = KeyCode.Space;
-        public KeyCode RightHandManipulationKey => rightHandManipulationKey;
+        private KeyBinding rightHandManipulationKey = KeyBinding.FromKey(KeyCode.Space);
+        public KeyBinding RightHandManipulationKey => rightHandManipulationKey;
+        [SerializeField]
+        [Tooltip("Controls how mouse look control is activated")]
```
wiwei: >Controls how mouse look control is activated [](start = 18, length = 44)

I think this tooltip was copy pasted from above (but should reference hand rotaation instead).

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationWindow.cs
```diff
@@ -136,13 +159,113 @@ private void OnGUI()
...
+                bool resetHand = GUILayout.Button("Reset");
+
+                if (newPosition != headTransform.position)
+                {
+                    headTransform.position = newPosition;
+                }
+                if (newRotation != headTransform.rotation.eulerAngles)
+                {
+                    headTransform.rotation = Quaternion.Euler(newRotation);
+                }
```
wiwei: Super nit: Move these into an else {} statement of resetHand - no need to set them if we're just gonna reset them to identity.

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/KeyBindingInspector.cs
```diff
@@ -0,0 +1,125 @@
...
+                {
+                    if (KeyBinding.EnumToKeyBindingMap.TryGetValue(newIndex, out KeyBinding kb))
+                    {
+                        bindingType.intValue = (int)kb.BindingType;
+                        code.intValue = kb.Code;
+                    }
+                }
+            }
+
+            // if (GUI.Button(position, KeyBinding.CodeToString(code.intValue)))
```
wiwei: >// if (GUI.Button(position, KeyBinding.CodeToString(code.intValue))) [](start = 12, length = 68)

Nit: I think dead code that was commented out right?

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/InputSimulationService.cs
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    [MixedRealityDataProvider(
+        typeof(IMixedRealityInputSystem),
+        SupportedPlatforms.WindowsEditor | SupportedPlatforms.MacEditor | SupportedPlatforms.LinuxEditor,
+        "Input Simulation Service",
+        "Profiles/DefaultMixedRealityInputSimulationProfile.asset",
+        "MixedRealityToolkit.SDK")]
+    [HelpURL("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/InputSimulation/InputSimulationService.html")]
+    public class InputSimulationService :
```
wiwei: If there's any way to preserve the file history here that would be great -this is showing up as a pure add/delete (meaning you lose history at this change). Was trying to review this against the old (and it's hard because it shows up as a pure add)

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityInputSimulationProfile.cs
```diff
@@ -87,30 +112,34 @@ public class MixedRealityInputSimulationProfile : BaseMixedRealityProfile
...
+        private KeyBinding handRotateButton = KeyBinding.FromMouseButton(KeyBinding.MouseButton.Right);
```
wiwei: >private KeyBinding handRotateButton = KeyBinding.FromMouseButton(KeyBinding.MouseButton.Right); [](start = 8, length = 95)

What are your thoughts on adding a comment regarding the priority order of hand rotate vs camera rotate? Otherwise the question here is around "hey, both the hand and camera have the same mouse binding" - AFAICT the camera wins out if both try to be active.

wiwei: Ah wait, it's the other way around.

---
In reply to: [320511622](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5853#discussion_r320511622) [](ancestors = 320511622)

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/InputSimulationService.cs
```diff
@@ -0,0 +1,430 @@
...
+            // Reset in cases where mouse position is jumping
+            if (gainedFocus || cursorLockChanged)
+            {
+                ResetMouseDelta();
+                return new MouseDelta();
+            }
+            else
+            {
+                // Use frame-to-frame mouse delta in pixels to determine mouse rotation.
+                // The traditional GetAxis("Mouse X") method doesn't work under Remote Desktop.
```
wiwei: It feels like this comment should apply to the else statement (could we move it closer there?).

Otherwise it looks like the comment says "we shouldn't do this" and then the code immediately does it (in what seems like the typical/first thing that is read)

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/InputSimulationService.cs
```diff
@@ -0,0 +1,430 @@
...
+                return new MouseDelta()
+                {
+                    screenDelta = screenDelta,
+                    worldDelta = worldDelta,
+                    viewportDelta = CameraCache.Main.ScreenToViewportPoint(screenDelta),
+                };
+            }
+        }
+
+        private const float MouseWorldDepth = 0.5f;
```
wiwei: >private const float MouseWorldDepth = 0.5f; [](start = 8, length = 43)

What's the significance of the .5 number? (and is it designed to also match the other .5f values down below?)

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/Microsoft.MixedReality.Toolkit.Services.InputSimulation.Editor.asmdef
```diff
@@ -1,15 +1,20 @@
...
-        "Microsoft.MixedReality.Toolkit.Services.InputSimulation"
-    ],
-    "optionalUnityReferences": [],
-    "includePlatforms": [
-        "Editor"
-    ],
-    "excludePlatforms": [],
-    "allowUnsafeCode": false
+{
+    "name": "Microsoft.MixedReality.Toolkit.Services.InputSimulation.Implementation",
```
wiwei: >InputSimulation [](start = 53, length = 15)

I think this file is still called "InputSimulation.Editor" but it doesn't match the name of the asmdef - I think this file should be renamed to InputSimulation.Implementation

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Input
+{
+    /// <summary>
+    /// Utility class to manage toggling of mouse rotation and associated features,
+    /// such as cursor visibility/locking
+    /// </summary>
+    /// <param name="wantsJumping">Show the cursor</param>
```
wiwei: >/// <param name="wantsJumping">Show the cursor</param> [](start = 4, length = 54)

Delete this line (I think it was copy pasted from the other wantsJumping location down below)

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+                if (toggle)
+                {
+                    if (isRotating)
+                    {
+                        // pressing escape will stop capture
+                        isRotating = !KeyInputSystem.GetKeyDown(cancelRotationKey);
+                    }
+                    else
+                    {
+                        // any kind of click will capture focus
```
wiwei: >// any kind of click will capture focus [](start = 24, length = 39)

I think the comment here and the code here don't seem to match up - it looks like it's getting keydown for the rotation key specifically, not any type of click.

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+                else if (rotationKey.BindingType == KeyBinding.KeyType.Key)
+                {
+                    // if mousebutton is either control, shift or focused
+                    UnityEngine.Cursor.lockState = CursorLockMode.Locked;
+                    // save current cursor visibility before hiding it
+                    wasCursorVisible = UnityEngine.Cursor.visible;
+                    UnityEngine.Cursor.visible = false;
+                }
+
+                // do nothing if (this.MouseLookButton == MouseButton.None)
```
wiwei: >MouseLookButton [](start = 39, length = 15)

I don't think that MouseLookButton is defined - was this supposed to be something else?

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        private static void OnEndRotating(KeyBinding rotationKey)
+        {
+            --numRotating;
+
+            if (numRotating == 0)
+            {
+                if (rotationKey.BindingType == KeyBinding.KeyType.Mouse)
+                {
+                    // if mousebutton is either left, right or middle
```
wiwei: >left, right or middle [](start = 48, length = 21)

Was this a comment from an older iteration? The helper function doesn't seem to have anything specific to left/right/middle mouse button

wiwei: Same comment on the thing down below:

// if mousebutton is either control, shift or focused

---
In reply to: [320515680](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5853#discussion_r320515680) [](ancestors = 320515680)

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+            else if (wasRotating && !isRotating)
+            {
+                OnEndRotating(rotationKey);
+            }
+        }
+
+        private static void OnStartRotating(KeyBinding rotationKey)
+        {
+            if (numRotating == 0)
+            {
```
wiwei: Why is this (and the function down below) using a == 0 check (if it's only called internally in this class and is gated on transitions - i.e. false -> true, true->false?). It seems like this should only transition from 0 -> 1 and then 1 -> 0

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+                {
+                    // recenter the cursor (setting lockCursor has side-effects under the hood)
+                    UnityEngine.Cursor.lockState = CursorLockMode.Locked;
+                    UnityEngine.Cursor.lockState = CursorLockMode.None;
+
+                    // restore the cursor
+                    UnityEngine.Cursor.visible = wasCursorVisible;
+                }
+
+    #if UNITY_EDITOR
```
wiwei: >#if UNITY_EDITOR [](start = 4, length = 16)

This asmdef looks to be editor-only - I'm curious why UNITY_EDITOR is specicially needed here (would be worth adding a comment here for that reason)

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/InputSimulationService.cs
```diff
@@ -0,0 +1,430 @@
...
+        BaseInputSimulationService,
+        IInputSimulationService,
+        IMixedRealityEyeGazeDataProvider,
+        IMixedRealityCapabilityCheck
+    {
+        private ManualCameraControl cameraControl = null;
+        private SimulatedHandDataProvider handDataProvider = null;
+
+        private HandSimulationMode handSimulationMode;
+        /// <inheritdoc />
```
thalbern: pls document all public properties / functions

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/ManualCameraControl.cs
```diff
@@ -0,0 +1,136 @@
...
+            profile = _profile;
+        }
+
+        private static float InputCurve(float x)
+        {
+            // smoothing input curve, converts from [-1,1] to [-2,2]
+            return Mathf.Sign(x) * (1.0f - Mathf.Cos(0.5f * Mathf.PI * Mathf.Clamp(x, -1.0f, 1.0f)));
+        }
+
+        public void UpdateTransform(Transform transform, MouseDelta mouseDelta)
```
thalbern: add doc comment to public function

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityInputSimulationProfile.cs
```diff
@@ -11,22 +12,38 @@ namespace Microsoft.MixedReality.Toolkit.Input
     [HelpURL("https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/InputSimulation/InputSimulationService.html")]
     public class MixedRealityInputSimulationProfile : BaseMixedRealityProfile
     {
+        [SerializeField]
+        [Tooltip("Indicator buttons to show input simulation state in the viewport")]
```
thalbern: can you add doc comments as well pls , tooltips won't shop up in the doc portal

thalbern: also some properties don't have tooltip (eg mouseRotationSensitivity)

lukastoenneMS:
> 
> 
> Thank you for continuing to work on input simulation, Lukas! It's a very popular feature.
> 
> Couple things to start with just from testing it, then will do a bigger code review once functionality is resolved:
> 
>     1. I found a bug -- my hand always disappears when I try to use the scroll wheel.
> 
>     2. I could not find in the documentation what the little arrows do in the overlay -- could you please explain, and add to the docs?
> 
> 
> image
> 
>     1. I often press and hold space and then press RMB to move the camera, but now that pressing space + RMB rotates the hand, which is something I do much less often. I think others also use RMB + space to move around a lot, and have that muscle memory. Can we brainstorm a different way to rotate the hand? For example, to rotate hand when spacebar is pressed, perhaps hold down ctrl key?

1. Scrolling fixed, this was because my mouseDelta conversions were producing world-space delta instead of screen scale (0..1).
1. I've added explanation to the docs now: 
image
1. Yes, that's why i made two separate key bindings for head and hand rotation, but wasn't sure if using a different key should be the default setting. Ctrl could work for hand rotation. Generally i've tried to keep the key bindings as close to the original as possible to not grind against muscle memory too much.

Assets/MixedRealityToolkit.SDK/Features/Input/InputSimulationIndicators.cs
lukastoenneMS: The indicators are an optional asset that is defined by the profile (default profile also lives in the SDK).

I'm not sure what the policy is regarding optional assets tied to services. It's just looks to me like we put all default assets in SDK, only have scripts in Services.

Assets/MixedRealityToolkit/Definitions/Devices/ControllerMappingLibrary.cs
lukastoenneMS: No objection. This input axis feature is not very robust in general, but Unity is introducing a new input system anyway, so i would put more effort into it at this point.
[https://forum.unity.com/threads/input-system-update.508660/](https://forum.unity.com/threads/input-system-update.508660/)

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
lukastoenneMS: Left over from previous iteration, will remove.

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
lukastoenneMS: Not sure what you mean here.

Assets/MixedRealityToolkit.Services/InputSimulation/KeyBinding.cs
lukastoenneMS: These are marked internal so that the inspector can access them. I didn't want to make them public because the value is not very meaningful on its own, the TryGetKeyCode and TryGetMouseButton functions should be used instead if one needs to know which internal KeyCode or mouse button is used.

The inspector needs access to some internals of the KeyBinding to convert to and from a plain enum index, which allows selecting a KeyBinding directly from a combined list. I've changed the inspector slightly so it now accesses the enum dictionaries and doesn't need the Code property.

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/InputSimulationService.cs
lukastoenneMS: I can move it back into the Editor folder for this PR, and then make a pure file move commit later.

The purpose of moving these files is to keep the service implementation (which happens to be editor-only atm) separate from actual inspectors etc.. That will also make it easier to change assembly settings if we ever figure out how to solve the input sim profile warnings in standalone builds #3971

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityInputSimulationProfile.cs
lukastoenneMS: I will change the default key binding to Ctrl for the hand rotation, so (by default) there won't be a conflict. And then if you set them to the same binding again it will just do both camera and hand rotation at the same time.

My thinking originally was that it would be better to just have one key for enabling rotation for anything, but it seems that the combinations are a bit fiddly, and by having dedicated keys for both rotations it just extends the current system and doesn't mess with muscle memory so much. There's just no ideal solution here, just more or less annoying tradeoffs ...

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
lukastoenneMS: This is a refcount to make sure that cursor visibility and locking is handled correctly when both head and hands are rotating at the same time.
* Cursor should not be unlocked when one rotation stops while the other is still going.
* Cursor must be unlocked when all rotations have stopped.

It's still not 100% reliable because it only handles the input sim subsystem - if any other system uses these static cursor features it can mess up the cursor state too. We don't have a global API for managing this. But at least this common case of head + hands rotations works.

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/InputSimulationService.cs
lukastoenneMS: Added comments
```cpp
        // Default world-space distance for converting screen/viewport scroll offsets into world space depth offset.
        // The pixel-to-world-unit ratio changes with depth, so have to chose a fixed distance for conversion.
        private const float mouseWorldDepth = 0.5f;
        // Center of the viewport is at (0.5, 0.5)
        private readonly Vector2 viewportCenter = new Vector2(0.5f, 0.5f);
```

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/Microsoft.MixedReality.Toolkit.Services.InputSimulation.Editor.asmdef
lukastoenneMS: I've moved files back into Editor for the time being, can do a clean file move later for git history.

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
lukastoenneMS: Legacy comment, fixed.

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
lukastoenneMS: Legacy comment, fixed.

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
lukastoenneMS: Legacy comment, fixed.

Assets/MixedRealityToolkit.SDK/Features/Input/InputSimulationIndicators.cs
```diff
@@ -0,0 +1,112 @@
...
+                    if (MixedRealityServiceRegistry.TryGetService<IMixedRealityInputSystem>(out IMixedRealityInputSystem inputSystem))
+                    {
+                        inputSimService = (inputSystem as IMixedRealityDataProviderAccess).GetDataProvider<IInputSimulationService>();
+                    }
+                }
+                return inputSimService;
+            }
+        }
+
+        /// <inheritdoc />
```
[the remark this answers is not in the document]
lukastoenneMS: done.

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/MouseRotationProvider.cs
lukastoenneMS: 1. This code originated in ManualCameraControl, which was previously in the all-platforms Microsoft.MixedReality.Toolkit.Services.InputSimulation asmdef. Since it's only used by the input simulation implementation class i have moved it into the Editor subfolder.
1. The Editor subfolder contains both genuine editor code (inspectors, windows) as well as implementation classes that just happen to be editor-only at this point. In the future it's possible that these implementation classes will be enabled for all platforms, so i would prefer to keep this #ifdef.

Assets/MixedRealityToolkit.Services/InputSimulation/Implementation/InputSimulationService.cs
lukastoenneMS: Yeah, it's a confusing legacy comment. You're correct, it probably applies to the else case, moved it there.

Assets/MixedRealityToolkit.Services/InputSimulation/Editor/InputSimulationWindow.cs
lukastoenneMS: It's just a UI script, the "else" case runs in, like, 99.999% of draw calls :)
No need to make it complicated IMO.

Assets/MixedRealityToolkit.Services/InputSimulation/MixedRealityInputSimulationProfile.cs
lukastoenneMS: done!

julenka:
After updating to the latest, I don't see any viewport buttons.  Regression?

julenka:
Lukas, suggestion perhaps for a different bug: Can we have a new simulation hand config: "hand open facing user"? This would be to make it easy to open a hand menu (common task now in MRTK).

I can make a new bug for that if needed, would be awesome to get that in this new input sim update.

lukastoenneMS:
> 
> 
> Lukas, suggestion perhaps for a different bug: Can we have a new simulation hand config: "hand open facing user"? This would be to make it easy to open a hand menu (common task now in MRTK).
> 
> I can make a new bug for that if needed, would be awesome to get that in this new input sim update.

I would rather do that in a new PR.

I suspect that adding this particular hand pose will cause some visual weirdness with our current joint position interpolation, because flipping over the hand will make all joints pass through the center. The hand will shrink and expand to come out the other side in the open-facing-user pose. To avoid that we could interpolate the relative joint rotations, keeping bone lengths constant, closer to what an actual animation system would do.

lukastoenneMS:
> 
> 
> Couple feedbacks after testing:
> 
>     1. I can't see the hand status icons any more.
> 
>     2. The hand rotation with ctrl is working much better for my muscle memory, but can we make it more sensitive? I want to easily be able to flip my hand to bring up a hand menu.

1. Can you check if the "Indicators Prefab" in the input sim profile is set correctly? I can't repro ...
2. I've added a separate rotation factor for the hand rotation now, since it should be faster than the head rotation. Also regrouped settings in the inspector slightly to have common shared settings at the top.

julenka:
> Can you check if the "Indicators Prefab" in the input sim profile is set correctly? I can't repro ...
It's set to None, but my prefab is the default. Are you sure you checked in the changs to the input sim profile?

image

I see a change to  Examples/Demos/EyeTracking/General/Profiles/EyeTrackingDemoInputSimulationProfile.asset
but not  the default profile.

julenka:
@davidkline-ms  FYI, this PR contains a breaking change that will requrire some profile upgrades for people that have custom input simulation profiles. Lukas posted update steps here: https://github.com/lukastoenneMS/MixedRealityToolkit-Unity/blob/camera_roll_sim/Documentation/Updating.md#input-simulation-profile-changes

Issue #5013, opened 2019-06-21
Add Interactable event via script


fcagnetta:
## Describe the problem

It would be useful to add events to interactables via script. Imagine cases in which we add the script at runtime.

- [x] Add OnClick from dynamically instantiated Interactable

## Describe the solution you'd like

It would be great a simple way to do that (still I can't find out how). Ok, maybe not same as OnClick event that is completely exposed, but something easy to implement.

## Describe alternatives you've considered

Use a global listener instead of interactable script

## Additional context

Same for interactable receiver script, and so on.

Troy-Ferrell:
@fcagnetta , hello! Can you elaborate on what you mean by "add events"? Do you mean just subscribe to the OnClick Unity event inside Interactable.cs and thus be able to assign actions at runtime when a user selects the Interactable? If so, this is possible.

`
var interactable = myGameObject.AddComponent<Interactable>();
interactable .OnClick.AddListener(YourFunction);
`

fcagnetta:
Hi Troy, sorry I try to explain better:

If you programmatically add Interactable component to an object, and then you want to add a listener for OnClick event and for OnFocus (on and off) event, and X other event you may need, how can you do that? OnClick event is automatically provided so you can just add a listener to it, but what about the others?

In the inspector you can just click add event and choose among a list of events, then you drag the script on which the method you want to call resides onto.

How to script this?
Thanks

Troy-Ferrell:
Ah I see now. I know in Unity inspector it says "Add event", but to be all on the same page, you mean add **InteractableEvent** objects dynamically. The "Add event" in the inspector I am pretty sure just represent **InteractableEvent** objects. The property we want is Interactable.Events and I believe if added correctly this could work. But the **InteractableEvent** creation isn't super obvious. For example, protected SetupEvent() in Interactable gets called on Awake so you would have to link in those pieces yourself? 

@killerantz can you help with better guidance?

fcagnetta:
Meanwhile I would like to add a bit of code of a use case for the OnClick() event registration:

```
Interactable interactable = gameObject.AddComponent<Interactable>();
interactable.IsGlobal = true;
interactable.OnClick.AddListener(() => SomeGuy.Instance.HelloWorld());

```
I get a null exception when Interactable.cs runs SetupStates() (line 272) 
`StateManager = States.SetupLogic();` - States is null. Maybe a call to GetStates() is needed instead of using States as is.

[...]

Issue #5652, opened 2019-08-14
InteractableInspector::RenderEventSettings is only taking already serialized properties into account


thalbern:
## Describe the bug
If new data gets added to any of the Interactable events, those fields won't show up in the inspector for already existing assets. 

## To reproduce

Add a new field in code to one of the Interactable Receivers eg PressReceiver and observe on opening any of the buttons that that field won't show up until you readd the receiver or switch the drop down to a different event .

## Expected behavior

New fields should be shown for new and already existing assets

## Your Setup (please complete the following information)

- Stabilization GA , MRTK v2.0

## Target Platform (please complete the following information)

- Unity Editor Bug

related to #5647

PR #5854, opened 2019-09-03
Add Interactable event listeners via script

julenka:
## Overview
Most UI frameworks allow for event listeners to be added not only from a GUI editor, but also from code. MRTK currently makes it easy for event listeners on Interactable (except OnClick) to be configured in the editor, but configuring at run-time from script is not possible.

This PR does the following:
1. Fixes bugs preventing adding events at run-time
1. Adds utility scripts to Interactable  to simplify adding Interactable event listeners 
1. Adds tests to verify all event listeners available in interactable by default
1. Adds examples to documentation

### Demo: Old vs. new way to add event listeners
Here's the old way to listen to focus enter/exit events:

```csharp
public static void AddFocusEvents(Interactable interactable)
{
    var ie = new InteractableEvent();
    var fr = new InteractableOnFocusReceiver(ie.Event);
    ie.Receiver = fr;
    interactable.Events.Add(ie);
    ie.Event.AddListener(() => Debug.Log("Focus on"));
    ie.OnFocusOff.AddListener(() => Debug.Log("Focus off"));
}
```

And the new way to add focus enter/exit events:

```csharp
public static void AddFocusEvents(Interactable interactable)
{
    var onFocusReceiver = interactable.AddReceiver<InteractableOnFocusReceiver>();
    onFocusReceiver.OnFocusOn.AddListener(() => Debug.Log("Focus on"));
    onFocusReceiver.OnFocusOff.AddListener(() => Debug.Log("Focus off"));
}
```

- Fixes: #5013

## Verification
Verify that new tests pass, and that they did not pass before.

### DIFF
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
@@ -24,10 +24,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
             get { return onFocusOff; }
         }
 
-        public InteractableOnFocusReceiver() : base()
-        {
-
-        }
+        public InteractableOnFocusReceiver() : this(new UnityEvent()) { }
 
         /// <summary>
         /// Raised when focus has entered the object
@@ -39,10 +36,7 @@ namespace Microsoft.MixedReality.Toolkit.UI
 
         private bool hadFocus;
 
-        public InteractableOnFocusReceiver(UnityEvent ev) : base(ev)
-        {
-            Name = "OnFocus";
-        }
+        public InteractableOnFocusReceiver(UnityEvent ev) : base(ev, "OnFocus") { }
 
         /// <inheritdoc />
         public override void OnUpdate(InteractableStates state, Interactable source)
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
@@ -15,7 +15,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
         public UnityEvent OnFocusOff = new UnityEvent();
 
         private bool hadFocus;
-        private State lastState;
 
         public InteractableOnFocusReceiver(UnityEvent ev) : base(ev)
         {
@@ -25,11 +24,9 @@ namespace Microsoft.MixedReality.Toolkit.UI
         /// <inheritdoc />
         public override void OnUpdate(InteractableStates state, Interactable source)
         {
-            bool changed = state.CurrentState() != lastState;
-
             bool hasFocus = state.GetState(InteractableStates.InteractableStateEnum.Focus).Value > 0;
 
-            if (hadFocus != hasFocus && changed)
+            if (hadFocus != hasFocus)
             {
                 if (hasFocus)
                 {
@@ -42,7 +39,6 @@ namespace Microsoft.MixedReality.Toolkit.UI
             }
 
             hadFocus = hasFocus;
-            lastState = state.CurrentState();
         }
     }
 }
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.MixedReality.Toolkit.Utilities.Editor;
+using UnityEngine;
 using UnityEngine.Events;
 
 namespace Microsoft.MixedReality.Toolkit.UI
@@ -11,8 +12,30 @@ namespace Microsoft.MixedReality.Toolkit.UI
     /// </summary>
     public class InteractableOnFocusReceiver : ReceiverBase
     {
+        [SerializeField]
         [InspectorField(Type = InspectorField.FieldTypes.Event, Label = "On Focus Off", Tooltip = "Focus has left the object")]
-        public UnityEvent OnFocusOff = new UnityEvent();
+        private UnityEvent onFocusOff = new UnityEvent();
+
+        /// <summary>
+        /// Raised when focus has left the object
+        /// </summary>
+        public UnityEvent OnFocusOff
+        {
+            get { return onFocusOff; }
+        }
+
+        public InteractableOnFocusReceiver() : base()
+        {
+
+        }
+
+        /// <summary>
+        /// Raised when focus has entered the object
+        /// </summary>
+        public UnityEvent OnFocusOn
+        {
+            get { return uEvent; }
+        }
 
         private bool hadFocus;
 
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
@@ -21,20 +21,14 @@ namespace Microsoft.MixedReality.Toolkit.UI
         /// <summary>
         /// Raised when focus has left the object
         /// </summary>
-        public UnityEvent OnFocusOff
-        {
-            get { return onFocusOff; }
-        }
+        public UnityEvent OnFocusOff => onFocusOff;
 
         public InteractableOnFocusReceiver() : this(new UnityEvent()) { }
 
         /// <summary>
         /// Raised when focus has entered the object
         /// </summary>
-        public UnityEvent OnFocusOn
-        {
-            get { return uEvent; }
-        }
+        public UnityEvent OnFocusOn => uEvent;
 
         private bool hadFocus;
 
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
@@ -13,8 +13,14 @@ namespace Microsoft.MixedReality.Toolkit.UI
     /// </summary>
     public class InteractableOnFocusReceiver : ReceiverBase
     {
+        /// <summary>
+        /// Creates receiver that raises focus enter and exit unity events
+        /// </summary>
         public InteractableOnFocusReceiver() : this(new UnityEvent()) { }
 
+        /// <summary>
+        /// Creates receiver that raises focus enter and exit unity events
+        /// </summary>
         public InteractableOnFocusReceiver(UnityEvent ev) : base(ev, "OnFocusOn") { }
 
         /// <summary>
--- a/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
+++ b/Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
@@ -13,17 +13,15 @@ namespace Microsoft.MixedReality.Toolkit.UI
     /// </summary>
     public class InteractableOnFocusReceiver : ReceiverBase
     {
-        [SerializeField]
-        [FormerlySerializedAs("OnFocusOff")]
-        [InspectorField(Type = InspectorField.FieldTypes.Event, Label = "On Focus Off", Tooltip = "Focus has left the object")]
-        private UnityEvent onFocusOff = new UnityEvent();
+        public InteractableOnFocusReceiver() : this(new UnityEvent()) { }
+
+        public InteractableOnFocusReceiver(UnityEvent ev) : base(ev, "OnFocusOn") { }
 
         /// <summary>
         /// Raised when focus has left the object
         /// </summary>
-        public UnityEvent OnFocusOff => onFocusOff;
-
-        public InteractableOnFocusReceiver() : this(new UnityEvent()) { }
+        [InspectorField(Type = InspectorField.FieldTypes.Event, Label = "On Focus Off", Tooltip = "Focus has left the object")]
+        public UnityEvent OnFocusOff = new UnityEvent();
 
         /// <summary>
         /// Raised when focus has entered the object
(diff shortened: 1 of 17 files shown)


Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
```diff
@@ -11,11 +12,32 @@ namespace Microsoft.MixedReality.Toolkit.UI
...
+        /// Raised when focus has left the object
+        /// </summary>
+        public UnityEvent OnFocusOff
+        {
+            get { return onFocusOff; }
+        }
+
+        public InteractableOnFocusReceiver() : base()
+        {
+
```
julenka: Initialize name here

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
```diff
@@ -25,11 +47,9 @@ public InteractableOnFocusReceiver(UnityEvent ev) : base(ev)
         /// <inheritdoc />
         public override void OnUpdate(InteractableStates state, Interactable source)
         {
-            bool changed = state.CurrentState() != lastState;
-
             bool hasFocus = state.GetState(InteractableStates.InteractableStateEnum.Focus).Value > 0;
 
-            if (hadFocus != hasFocus && changed)
+            if (hadFocus != hasFocus)
```
julenka: FYI @Troy-Ferrell interactable doesn't actually update its state when instantiated from code -- it is always in "disabled" state. Troy are you already fixing this or should I file a new bug?

julenka: Note: I still think this change is valid, since i'm not sure why we need to predicate this particular event on a state change. We are already tracking state via hadFocus.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
```diff
@@ -11,11 +12,32 @@ namespace Microsoft.MixedReality.Toolkit.UI
     /// </summary>
     public class InteractableOnFocusReceiver : ReceiverBase
     {
+        [SerializeField]
         [InspectorField(Type = InspectorField.FieldTypes.Event, Label = "On Focus Off", Tooltip = "Focus has left the object")]
-        public UnityEvent OnFocusOff = new UnityEvent();
+        private UnityEvent onFocusOff = new UnityEvent();
```
julenka: Add formerlyserializedas

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -127,6 +127,38 @@ protected static IMixedRealityInputSystem InputSystem
         /// </summary>
         public bool FocusEnabled { get { return !IsGlobal; } set { IsGlobal = !value; } }
 
+        /// <summary>
+        /// Event receivers can be used to listen for different
+        /// events at runtime. This method allows receivers to be dynamically added at runtime.
+        /// </summary>
```
julenka: Add an example of how to add event listener in doccomment

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -127,6 +127,38 @@ protected static IMixedRealityInputSystem InputSystem
...
+            result.Event = interactableEvent.Event;
+            interactableEvent.Receiver = result;
+            Events.Add(interactableEvent);
+            return result;
+        }
+
+        /// <summary>
+        /// Returns the first receiver of type T on the interactable,
+        /// or null if nothing is found.
+        /// </summary>
```
julenka: add example

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -61,6 +61,146 @@ public IEnumerator TestAddInteractableAtRuntime()
             yield return null;
         }
 
+        /// <summary>
+        /// Tests that an interactable component can be added to a GameObject
```
julenka: Interactable

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -61,6 +61,146 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+            onFocusReceiver.OnFocusOn.AddListener(() => didHover = true);
+            onFocusReceiver.OnFocusOff.AddListener(() => didUnHover = true);
+            CameraCache.Main.transform.LookAt(interactable.transform);
+            yield return null;
+            CameraCache.Main.transform.LookAt(Vector3.forward);
+            yield return null;
+            Assert.True(didHover, "Interactable did not receive hover event");
+            Assert.True(didUnHover, "Interactable did not receive un-hover event");
+
+            // OnGrab
```
julenka: Move these all out into methods, I think.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
Troy-Ferrell: @julenka, no my changes atm are currently all around theme consumption/management in Interactable

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
```diff
@@ -11,11 +12,32 @@ namespace Microsoft.MixedReality.Toolkit.UI
...
+
+        /// <summary>
+        /// Raised when focus has left the object
+        /// </summary>
+        public UnityEvent OnFocusOff
```
Troy-Ferrell: nit: maybe Public UnityEvent OnFocusOff => onFocusOff for simpler code?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
```diff
@@ -11,11 +12,32 @@ namespace Microsoft.MixedReality.Toolkit.UI
...
+        {
+            get { return onFocusOff; }
+        }
+
+        public InteractableOnFocusReceiver() : base()
```
Troy-Ferrell: comments? summary

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
Troy-Ferrell: Same with others if so

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/ReceiverBase.cs
```diff
@@ -16,8 +17,23 @@ public abstract class ReceiverBase
 
         public bool HideUnityEvents;
         protected UnityEvent uEvent;
+
+        /// <summary>
+        /// Each Receiver has a base Event it raises, (in addition to others).
+        /// </summary>
+        public UnityEvent Event
```
Troy-Ferrell: Why is this Event public set/get but the other specific inherited event types are only public getters?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/ReceiverBase.cs
```diff
@@ -16,8 +17,23 @@ public abstract class ReceiverBase
...
+        {
+            get { return uEvent; }
+            set { uEvent = value; }
+        }
+
         public MonoBehaviour Host;
 
+        public ReceiverBase()
+        {
+            uEvent = new UnityEvent();
```
Troy-Ferrell: why not protected UnityEvent uEvent = new UnityEvent()? Reason for constructor?

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
```diff
@@ -127,6 +127,38 @@ protected static IMixedRealityInputSystem InputSystem
...
+        public T GetReceiver<T>() where T : ReceiverBase
```
wiwei: >GetReceiver [](start = 17, length = 11)

It's also odd to me that it returns the first found of a certain type (meaning if I add two of the same type, it will never return the other one) - this particular type of API surface (first one wins) is of an odd design (usually you want to provide deterministic way of getting a specific instance).

i.e. if I call AddReciever twice and then lose the second one, there's no way for me to get the second one using these typical APIs (I guess I could iterate over the events).

Maybe there should be an overload that returns all that match (keeping the singular case for convenience/typical usage)? Or should there never be multiples there?

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -61,6 +61,146 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+            PlayModeTestUtilities.SetHandSimulationMode(HandSimulationMode.Gestures);
+
+            // This should not throw an exception
+            var interactable = cube.AddComponent<Interactable>();
+
+            // Subscribe to interactable's on click so we know the click went through
+            bool wasClicked = false;
+            interactable.OnClick.AddListener(() => { wasClicked = true; });
+
+            TestHand th = new TestHand(Handedness.Right);
```
wiwei: >th [](start = 21, length = 2)

testHand

It doesn't take that many more characters and is more readable.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
```diff
@@ -61,6 +61,146 @@ public IEnumerator TestAddInteractableAtRuntime()
...
+            bool didHover = false;
+            bool didUnHover = false;
+            onFocusReceiver.OnFocusOn.AddListener(() => didHover = true);
+            onFocusReceiver.OnFocusOff.AddListener(() => didUnHover = true);
+            CameraCache.Main.transform.LookAt(interactable.transform);
+            yield return null;
+            CameraCache.Main.transform.LookAt(Vector3.forward);
+            yield return null;
+            Assert.True(didHover, "Interactable did not receive hover event");
+            Assert.True(didUnHover, "Interactable did not receive un-hover event");
```
wiwei: If we already have coverage of this in another area feel free to ignore this comment - note that it might be better to have another line like:

Assert.True(didHover, "Interactable did not receive hover event");
Assert.False(didUnHover, "Interactable shouldn't have received unhover event");

After looking at the interactable, to show that the right operations are occuring in the right way.

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
wiwei: The test method is a bit long, but it's within the scope of the naming of this function, and the code is readable within this test as well, so it's fine IMO.

The other thing to consider is having each thing in its own test (i.e. TestOnGrab, TestOnFocus)... etc, that way if only one thing fails, the other tests at least get to run, and you get more granular information about what failed.

---
In reply to: [320324131](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5854#discussion_r320324131) [](ancestors = 320324131)

wiwei: Actually yeah, the second suggestion is probably better, but it's up to you.

---
In reply to: [320468834](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5854#discussion_r320468834) [](ancestors = 320468834,320324131)

wiwei:
Generally lgtm, pending question about GetReciever

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
julenka: Creatd #5866 5866

julenka:
Just discovered issue: the secondary / extra events like OnRelease for the press receiver don't show up in editor. Need to fix that before checking in.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/ReceiverBase.cs
julenka: Good catch. Removed the setter.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/ReceiverBase.cs
julenka: Removed this constructor.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/ReceiverBase.cs
julenka: It's because in AddReceiver I am specifying generic type, and generic types cannot AFAIK be instantiated with params. So, I need to do:

```
            var interactableEvent = new InteractableEvent();
            var result = new T();
            result.Event = interactableEvent.Event;
```

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Interactable.cs
julenka: I added overload for returning all matches.

julenka: But actually it seems odd that you'd have multiple of a type. But, it is possible so I added an overload.

Troy-Ferrell:
> Need

FYI, sounds possibly related or dupe of #5652

Assets/MixedRealityToolkit.Tests/PlayModeTests/InteractableTests.cs
julenka: done. good idea.

Assets/MixedRealityToolkit.SDK/Features/UX/Interactable/Scripts/Events/InteractableOnFocusReceiver.cs
julenka: done. for all of them.

julenka:
> > Just discovered issue: the secondary / extra events like OnRelease for the press receiver don't show up in editor. Need to fix that before checking in.
> 
> @julenka FYI, sounds possibly related or dupe of #5652

Ended up being because I was using the [SerializeField] then event patters. It looks like the values don't get saved properly if you don't have a public setter. I fixed by reverting that change, not I just have public UnityEvents like before.

PR #5855, opened 2019-09-03
CI Build Fix

andreiborodin:
This change is to fix the CI Build which is failing because the System.Core.dll is being overwritten by MSBuild. The reason for why this started happening recently, is unclear, however, the fix is to set AddAdditionalExplicitAssemblyReferences to false.

Will test a CI build on this change before commiting.


Assets/MixedRealityToolkit.Tools/MSBuild/MSBuildTemplates/SDKProjectTemplate.csproj.template
```diff
@@ -19,6 +19,7 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <!--NOTE: Currently one of the Unity packages has a * in AssemblyVersion identifier in the AssemblyInfo file, so have to do this.-->
     <Deterministic>false</Deterministic>
+	<AddAdditionalExplicitAssemblyReferences>false</AddAdditionalExplicitAssemblyReferences>
```
macborow: It might be best to consistently use spaces for indentation and not mix them with tabs?

andreiborodin: Oops, good catch.

wiwei:
@andreiborodin I kicked off a build here: 

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=4886

PR #5864, opened 2019-09-03
Add a code checker to validate that we don't have empty doccomments.

wiwei:
These doccomments get added automatically by IDEs and end up being checked in empty - they tend to increase line count/verbiage without adding value, because they get checked in empty. The generated docs aren't any different because they're there, so it's better to either add some comment there or delete the lines entirely.


scripts/ci/validatecode.ps1
```diff
@@ -38,6 +38,36 @@ function CheckBooLang(
...
+function CheckEmptyDoccomment(
+    [string]$FileName,
+    [string[]]$FileContent,
+    [int]$LineNumber
+) {
+    <#
+    .SYNOPSIS
+        Checks if the given file (at the given line number) contains an empty <returns></returns>
+        or an <param ...></param> doccomment. These are typically added automatically by IDEs,
+        but empty doccomments don't at value to code.
```
keveleigh:
```suggestion
        but empty doccomments don't add value to code.
```

wiwei:
If people are okay with this change, I will fix up all of the existing usage.

There's a possible downside here which is that, because the IDE does this automatically, it'll become a burden, so I'm open to doing this or not doing this.

wiwei:
Here's how you fix the issues:

find ./ -name "*.cs" -exec sed -i -b '/\/\/\/[ ]*<param[ a-zA-Z0-9"=]*>[ ]*<\/param>/d' {} \;
find ./ -name "*.cs" -exec sed -i -b '/\/\/\/[ ]*<returns[ a-zA-Z0-9"=]*>[ ]*<\/returns>/d' {} \;

Assets/MixedRealityToolkit.Examples/Demos/EyeTracking/DemoScrollPanZoom/Scripts/BaseClasses/PanZoomBase.cs
```diff
@@ -207,7 +207,6 @@ public void AutoPan()
         /// <summary>
         /// Scroll sideways.
         /// </summary>
-        /// <param name="speed"></param>
```
keveleigh: Do we want to just remove them or (more work, but) should we actually define them all?

keveleigh: I guess, as long as there's some tool that lets us also see where param / return documentation is missing

wiwei: Defining them all would be the best.

I think in order of things of desirability:
1. Having documentation there.
1. Having no empty documentation text.
1. Having no documentation at all.

The reason being, having just empty doc text is the ultimate teaser, and also doesn't actually affect the generated docfx docs (I compared a before/after generated thing and it's the same API docs)

The hope here is that people will just update their things to either delete them (i.e. <returns></returns> where it's obvious what it returns) or update them (param names).

The main goal of this is to move from 3 to 2

wiwei:
> 200 files? That's a lot. How did you fix after you found them?

Ah, I mentioned this above:

> Here's how you fix the issues:
> 
> find ./ -name "_.cs" -exec sed -i -b '////[ ]_<param[ a-zA-Z0-9"=]_>[ ]_</param>/d' {} ;
> find ./ -name "_.cs" -exec sed -i -b '////[ ]_<returns[ a-zA-Z0-9"=]_>[ ]_</returns>/d' {} ;

Note that I don't want to paste this into the script itself/message so as to encourage people to actually write docs instead of deleting them.

Issue #5869, opened 2019-09-04
Touching the Oculus Rift motion controller performs a press (select)


Alexees:
## Describe the bug

When using Rift or Quest controllers, just touching the triggers performs select actions, hence button presses and such.
That's because PrimaryIndexTrigger Touch and Press use the same keycode they listen to, which in case of Press is wrong.

The mappings, Unity's mapping description and Oculus's official mapping support seem to be out of order anyways.

While OculusTouchController.cs has 4 PrimaryIndexTrigger definitions, Unity's documentation only contain 3, of which OculusTouchController.cs misses out on "Squeeze", which is probably "Axis1D.SecondaryIndexTrigger" according to its axis.

PrimaryIndexTrigger Press is therefore redundant and basically needs to be a digital implementation of PrimaryIndexTrigger Squeeze.

the problem is that the default "select" InputAction is used for digital types, but the Trigger press is analogue and there is no logic for combining the two yet.

## To reproduce

1. Build to a Rift and touch the trigger. You'll already see the beam becoming a straight line

## Expected behavior

Only pulling the trigger performs the default select action

## Your Setup (please complete the following information)

- Unity Version 2018.4.6f1
- MRTK Version mrtk_development

## Target Platform (please complete the following information)

- OpenVR
- Oculus Quest

Alexees:
The only KeyCode fired for the trigger press is Mouse0 (on the Quest that is), but this goes for both controllers.
What would be the best approach for reading an axis digitally? As far as I can tell, InputActions match if the ID and the AxisType is identical. This means even though its a SingleAxis reading, it must be digital and the implementation for that reads keycodes...

SimonDarksideJ:
Just reviewing through this and looks like some changes happened somewhere that changed the default behaviour, so I'll sort out a PR to resolve them, both here and the XRTK.

a few points:
* 1, "Axis1D.PrimaryIndexTrigger" - this is the axis value for the trigger mapped to axis 9 which is the Squeeze value in Unity OpenVR
* 4, "Axis1D.PrimaryIndexTrigger Press", should be the digital equivalent of this but as you point out its been mapped to the Touch keycode
* Touching the trigger should start the pointer / select, if you feel it shouldn't then you need only update the controller mapping to your preference.

I'll get the Axis / button mappings cleaned up

Alexees:
[...]

PR #5872, opened 2019-09-04
Digital axis for joysticks

Alexees:
## Overview
The current implementation for the default select InputAction forces Unity InputSystem bases controllers to use digital input like buttons.
Now with the Oculus Quest, there is no such input for the front trigger other than the KeyCode mouse0, which fires for both controllers.
To distinguish between them we need to read an axis as a button.

This PR makes it possible to provice an axis when using Keycode.None for a digital input to read an axis.

## Changes
- Fixes: #5869


Troy-Ferrell:
FYI, another PR is out for same bug: #5874 

@SimonDarksideJ FYI as well

SimonDarksideJ:
Would recommend closing this in favour of #5874 which better addresses the Axis as buttons effect.  Now handled properly in the Button check.
This change would handle 0/1 buttons but would not handle button release.

keveleigh:
Closing after merge of #5874, which fixes the same issue.

Issue #5871, opened 2019-09-04
Oculus Rift controller can not evaluate Near Touch at all


Alexees:
## Describe the bug

Neither Near touch nor Trigger Touch can be evaluated to true by the controller. That's because the MixedRealityInteractionMapping for that defines the AxisType as digital and provides an Axis code instead of a keycode for it.

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/6dd8bd5d17d6ba49a09b4532048a80b520622242/Assets/MixedRealityToolkit.Providers/OpenVR/OculusTouchController.cs#L35

The evaluating code however only checks keycodes in case of the digital axis type, which in this case is Keycode.None.

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/6dd8bd5d17d6ba49a09b4532048a80b520622242/Assets/MixedRealityToolkit/Providers/UnityInput/GenericJoystickController.cs#L86-L91

Maybe the AxisType just needs to be changed to SingleAxis and that's it.

## Expected behavior

NearTouch is interpreted some boolean way

## Your Setup (please complete the following information)

- Unity Version 2018.4.6f1
- MRTK Version mrtk_development

## Target Platform (please complete the following information)

- OpenVR
- Oculus Quest

SimonDarksideJ:
This is certainly a bug and not sure where the change to remove the axis -> digital check disappeared to.  I'll get that reinstated and with #5869 I'll clean up the mappings.

keveleigh:
This should be fixed

PR #5874, opened 2019-09-04
Fix to update Oculus mapping and resolve Axis button resolution

SimonDarksideJ:
## Overview
Fix to update Oculus mapping and resolve Axis button resolution

* Mappings updated, including default asset (example controller assets unaffected)
* GenericJoystickController updated to move Axis button validation to the Digital Button check, removed from Axis check

## Changes
- Fixes: #5869 & #5871


Assets/MixedRealityToolkit/Providers/UnityInput/GenericJoystickController.cs
```diff
@@ -145,25 +141,6 @@ protected void UpdateSingleAxisData(MixedRealityInteractionMapping interactionMa
...
                         InputSystem?.RaiseFloatInputChanged(InputSource, ControllerHandedness, interactionMapping.MixedRealityInputAction, interactionMapping.FloatData);
                     }
-                    return;
-                default:
-                    Debug.LogWarning("Unhandled Interaction");
-                    return;
-            }
-
-            // If our value changed raise it.
-            if (interactionMapping.Changed)
```
julenka: Why are we removing this section? It seems to be raising OnInputDown and Up events. Is it handled somewhere else?

Alexees:
As far as I know, GetAxis returns precise float values while GetAxisRaw returns just the mins and maxs. Considering we are talking about digital input here, doesn't that mean I have to pull a trigger fully to get a signal instead of passing the deadzone?

Assets/MixedRealityToolkit/Providers/UnityInput/GenericJoystickController.cs
SimonDarksideJ: Not removed, just moved to the Button handler.
Did various tests in both cases and button is the better fit and handles more scenarios.  Will be validating and making the same change in the XRTK following this testing.

SimonDarksideJ:
That is an interesting proposition @Alexees, that would work for NearTouch scenarios where 0 or 1 are the only values, But "Pressed" uses max for action but < 1 (Or maybe 1 is hold, less < 1 is not held, using 1/0 for pressed maybe?) for release, so wouldn't fit.
Better not to over complicate it with having multiple ways to validate those events.

SimonDarksideJ:
> Out of curiosity, what's the best way to verify this fix, Simon? Would it be to try to use the HandInteractionExamples scene with an Oculus Touch controller?

@julenka One demo scene we are missing from the HoloToolkit examples is the Controller scene, where it demonstrated inputs / events for all combinations.  I thought the InputActions scene might have given this but sadly it's quite lacking.  Might be worth a PR to upgrade that old scene to MRTK and add it?

I tested using a custom script I've used many times in controller development / testing.

Alexees:
Sorry for my english, doesn't seem to be good enough to do complex sentences and still transport the right meaning :) I'll try again:
I checked the amount of trigger pull you need to perform to get a click going in the Quest shell and it's not even a quarter. Talking deadzone here, I would say it's about .2f or something. When the GenericJoystickController uses GetAxis, you need to pull the trigger all the way through (tiny clearance at the end), because have all values in between. With GetAxisRaw however, we only need to pass the deadzone, which is at .19f, which feels better, and relates closer to the shell.
I don't understand your explanation about the difference between "Near Touch" and "Pressed". For near touch you say 0 or 1 are the only values, but with "Pressed" that's the same case, it's just either smaller and bigger than the deadzone (AxisRaw) or smaller 1 and 1 (GetAxis), which is just a shift of where to distinguish and I think it should be around the deadzone.

SimonDarksideJ:
(From the Unity Specs)
So nearTouch is an Axis but only produces a 0 or a 1
Touch is a button, which produces On or Off (bool)

As to Squeeze, this is an Axis that provides a range between 0 and 1.

So for Pressed, you are basically stating, rather than 0 or 1, there is a deadzone you want to use, so pressed is activated 1/3rd in it's range?  Correct?

And you are also saying, GetAxisRaw is more refined which will allow us to get this deadzone value?

At present, none of the MRTK controller logic has any deadzone logic (neither does XRTK) but it did come up in discussion when we were updating native Oculus, which does implement it.

---

So to sum up, if I understand correctly, you want Pressed to Fire if the trigger is pressed more than 1/3rd of the range, instead of the current minimum of fully pressed value of 1?

If that correct @Alexees ?

Alexees:
> At present, none of the MRTK controller logic has any deadzone logic 

Unity's InputSystem uses deadzones for axis by definition:

image

> So for Pressed, you are basically stating, rather than 0 or 1, there is a deadzone you want to use

I want it to be 0-or-1, but instead of everything below 1 is 0 and just 1 is 1, it's nicer to have everything below the deadzone as 0 and everything above as 1. Simply using AxisRaw gives us that.

SimonDarksideJ:
After a bunch of research, the consensus is that using the "Dead" parameter in the input definitions is a bad idea, as it decreases your overall volume of response.  E.G. only getting a 75% range of motion.
https://forum.unity.com/threads/axis-input-dead-problematic-for-diagonal-input.399269/

The better answer (especially for situations like this) is to use an internal / manual deadzone for controls that need it.

So coming back to the requirement.  When do you want pressed to fire (without affecting the squeeze range), at 30%, more or less @Alexees ?

Alexees:
I think 20% is more likely accurate.

> After a bunch of research, the consensus is that using the "Dead" parameter in the input definitions is a bad idea, as it decreases your overall volume of response

I guess this is then getting moved over to the mappings, right?

Assets/MixedRealityToolkit/Providers/UnityInput/GenericJoystickController.cs
```diff
@@ -118,24 +132,6 @@ protected void UpdateSingleAxisData(MixedRealityInteractionMapping interactionMa
 
             var singleAxisValue = UInput.GetAxis(interactionMapping.AxisCodeX);
 
-            switch (interactionMapping.InputType)
-            {
-                case DeviceInputType.TriggerPress:
```
keveleigh: This change will also require some changes to the OpenVR WMR Controller, which also uses `SingleAxis` + `*Press` to simulate a digital input from an axis:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/89fb1f941856963b0012c6755ddaf61a94d4b999/Assets/MixedRealityToolkit.Providers/OpenVR/WindowsMixedRealityOpenVRMotionController.cs#L37

Either we update the grip axes there to be `Digital`, or we can reinstate this code and change the above `MixedRealityInteractionMapping`s to be `SingleAxis` as well.

SimonDarksideJ: Interesting, that wasn't my experience but I'll test some more, thanks for catching this @keveleigh

SimonDarksideJ: Addressed.   reinstated use case for testing TrigerPress used by WMR for Grip.

Ideally, this should be expressed as the two input types for Grip, position and press.  However, in analysis, this would have been a major breaking change that should be scoped later if required.

For now, the update reenables digital style input for the WMR trigger.

SimonDarksideJ:
Pushed update to enable raw data axis, this simply gives greater precision.  Dead zones however, will need to be handled by the developer for now as it would have caused too large a breaking change.

PR #5881, opened 2019-09-04
BuildNumber to be SemVer 1.0.0 Compatible

andreiborodin:


wiwei:
Adding a note that this is only necessary because this issue is open:

https://github.com/GlitchEnzo/NuGetForUnity/issues/273

@andreiborodin, can you also open an issue tracking removal of this (i.e. I assume that this is just a temporary workaround) that links to that above issue as well?

andreiborodin:
This is also only in the 2.0.1 release, we won't be bringing it to mrtk_development.

Issue #5878, opened 2019-09-04
When using the Mixed Reality Toolkit/Standard shader with UGUI, scale is not utilized.


Cameron-Micka:
## Describe the bug

On Unity UI components the unity_ObjectToWorld matrix is not the transformation matrix of the local transform the Graphic component lives on, but that of it's parent Canvas. Many MRTK/Standard shader effects require object scale to be known to renderer correctly.

## To reproduce

1. Select an Image component under a UGUI canvas.
2. Change the material to "HolographicBackPlate"
3. Notice that the material does not render as expected.

## Expected behavior

Materials look the same on normal renderers and UGUI Graphic renderers.

## Screenshots

Current result:
![image](https://user-images.githubusercontent.com/13305729/64280105-44f1e380-cf05-11e9-9e9c-cb0fdbe86c31.png)

Desired result:
![image](https://user-images.githubusercontent.com/13305729/64280129-576c1d00-cf05-11e9-9c8b-174da98f086b.png)

## Your Setup (please complete the following information)

- Unity Version [e.g. 2018.4.6f1]
- mrtk_development branch

## Target Platform (please complete the following information)

- HoloLens
- HoloLens 2
- WMR immersive
- OpenVR

PR #5882, opened 2019-09-04
UGUI Scale Support for the MRTK/Standard Shader

Cameron-Micka:
## Overview
On Unity UI components the unity_ObjectToWorld matrix is not the transformation matrix of the local transform the Graphic component lives on, but that of it's parent Canvas. Many MRTK/Standard shader effects require object scale to be known. To solve this issue the ScaleMeshEffect has been created to pass scale information to the shader while preserving batching.

This PR also introduces a StandardShaderUtility class which will be fleshed out over time to include more utilities for common operations.

Example of the UI that instructs a user on how to fix the issue (note the same material is being used on the default Unity Quad and UGUI Image):

Fix

Before:
image

After:
image

## Changes
- Fixes: https://github.com/microsoft/MixedRealityToolkit-Unity/issues/5878


Assets/MixedRealityToolkit/StandardAssets/Shaders/MixedRealityStandard.shader
```diff
@@ -679,6 +678,13 @@ Shader "Mixed Reality Toolkit/Standard"
 #else
                 o.scale.z = length(mul(unity_ObjectToWorld, float4(0.0, 0.0, 1.0, 0.0)));
 #endif
+                // uv3.y will contain a negative value when rendered by a UGUI and ScaleMeshEffect.
+                if (v.uv3.y < 0.0)
+                {
+                    o.scale.x *= v.uv2.x;
+                    o.scale.y *= v.uv2.y;
+                    o.scale.z *= v.uv3.x;
```
AdamMitchell-ms: >3.x [](start = 37, length = 3)

Did you mean for this to be uv3?  And X?  Not sure if this is a copy/paste error or intentional because Z is a special case for UGUI.

If it's intentional, add a comment explaining why.

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <param name="vh">The vertex helper to populate with new vertex data.</param>
+        public override void ModifyMesh(VertexHelper vh)
+        {
+            var rectTransform = transform as RectTransform;
+            var scale = new Vector2(rectTransform.rect.width * rectTransform.localScale.x,
+                                    rectTransform.rect.height * rectTransform.localScale.y);
+
+            var canvas = GetComponentInParent<Canvas>();
+            var depth = new Vector2((canvas ? (1.0f / canvas.transform.lossyScale.z) : 1.0f) * rectTransform.localScale.z, 
+                                    -1.0f);
```
AdamMitchell-ms: I see.  Here's where you're setting it to the -1 you mentioned in the shader.  Maybe add a comment here saying that that's why you're setting it to -1 here.

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        /// <summary>
+        /// Stores scaling information into UV channel attributes during UI mesh construction.
+        /// </summary>
+        /// <param name="vh">The vertex helper to populate with new vertex data.</param>
+        public override void ModifyMesh(VertexHelper vh)
+        {
+            var rectTransform = transform as RectTransform;
+            var scale = new Vector2(rectTransform.rect.width * rectTransform.localScale.x,
```
AdamMitchell-ms: >localScale [](start = 77, length = 10)

Will this work for nested rectTransforms with non-1 localScale?

Ex:
Canvas
  > Parent (scale not 1.0)
    > ImageWithScaleMeshEffect

You might need to iterate through the hierarchy accumulating localScales.  Or alternately, use rectTransform.scale / canvas.scale maybe?

Assets/MixedRealityToolkit.Services/InputSystem/Editor/MixedRealityCanvasInspector.cs
```diff
@@ -341,5 +357,26 @@ public override void OnInspectorGUI()
 
             serializedObject.ApplyModifiedProperties();
         }
+
+        private static List<Graphic> GetGraphicsWhichRequireScaleMeshEffect(Object[] targets)
+        {
+            var output = new List<Graphic>();
```
AdamMitchell-ms: just for fun, you could consider implementing this using LINQ.

using System.Linq;

targets.SelectMany(target => (target as Canvas).GetComponentsInChildren<Graphic>())
.Where(graphic =>StandardShaderUtility.IsUsingMrtkStandardShader(graphic.material) && graphic.GetComponent<ScaleMeshEffect>() == null);

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1827,9 +1826,7 @@ private void SetMaterials()
             {
                 float[] color = { 1.0f, 1.0f, 1.0f, 0.75f };
 
-                Shader shader = Shader.Find("Mixed Reality Toolkit/Standard");
-
-                handleMaterial = new Material(shader);
+                handleMaterial = new Material(StandardShaderUtility.MrtkStandardShader);
```
julenka: @thalbern FYI, for your refactor. Don't lose this change

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace Microsoft.MixedReality.Toolkit.Input.Utilities
```
julenka: Should this be in Services, or SDK? @davidkline-ms I often wonder about this, whether to put things in services or SDK. What are the rules to follow for deciding if something should go into services or SDK? Maybe a one sentence rule of thumb to follow? Here's a strawman: "If it related to getting data, it goes into services. Otherwise (UI, interactions, etc) put it in SDK"

Assets/MixedRealityToolkit/Utilities/StandardShader/StandardShaderUtility.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using UnityEngine;
+
+namespace Microsoft.MixedReality.Toolkit.Utilities
```
julenka: Also I wonder if this should go into SDK, not core MRTK

Assets/MixedRealityToolkit.Services/InputSystem/Editor/MixedRealityCanvasInspector.cs
AdamMitchell-ms: PS.  If you do that, then this function could return IEnumerable<Graphic> instead of List<Graphic>

---
In reply to: [321028562](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5882#discussion_r321028562) [](ancestors = 321028562)

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
Cameron-Micka: Good idea! I’ll add a comment expanding what is going on here and in the shader. The way the data is packed is a little obuscated.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -1815,9 +1816,7 @@ private void SetMaterials()
             {
                 float[] color = { 1.0f, 1.0f, 1.0f, 0.75f };
 
-                Shader shader = Shader.Find("Mixed Reality Toolkit/Standard");
-
-                wireframeMaterial = new Material(shader);
+                wireframeMaterial = new Material(StandardShaderUtility.MrtkStandardShader);
```
Troy-Ferrell: I was just thinking about this. So many inspector classes default to MRTK shader. Great to have it in one place!

Assets/MixedRealityToolkit/StandardAssets/Shaders/MixedRealityStandard.shader
```diff
@@ -327,11 +327,10 @@ Shader "Mixed Reality Toolkit/Standard"
                 float4 vertex : POSITION;
                 float2 uv : TEXCOORD0;
 #if defined(LIGHTMAP_ON)
-                float2 lightMapUV : TEXCOORD1;
-#endif
-#if defined (_VERTEX_EXTRUSION_SMOOTH_NORMALS)
-                fixed3 smoothNormal : TEXCOORD2;
+                float2 uv1 : TEXCOORD1;
```
Troy-Ferrell: Add comments here for what uv1, uv2, uv3 mean

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.MixedReality.Toolkit.Input;
+using Microsoft.MixedReality.Toolkit.Utilities;
```
Troy-Ferrell: Documentation page?

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    /// <summary>
+    /// On Unity UI components the unity_ObjectToWorld matrix is not the transformation matrix of the local 
+    /// transform the Graphic component lives on, but that of it's parent Canvas. Many MRTK/Standard shader 
+    /// effects require object scale to be known. To solve this issue the ScaleMeshEffect will store scaling 
+    /// information into UV channel attributes during UI mesh construction. Ideally we would store the scale 
+    /// in one attribute but UGUI only supports two scalers per attribute (even in the tangent attribute).
+    /// </summary>
+    [RequireComponent(typeof(RectTransform))]
```
Troy-Ferrell: Set helpurl? or better yet on BaseMeshEffect if not already there?

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
Cameron-Micka: That's a good point, it might not? Let me verify and get back to you.

Assets/MixedRealityToolkit.Services/InputSystem/Editor/MixedRealityCanvasInspector.cs
Cameron-Micka: Haha, I haven't heard many people use the work fun and LINQ in the same sentence. ;) I do like the idea of returning an IEnumerable. I'll fix that!

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
Cameron-Micka: To be honest I don't understand the divide between SDK, Services, and Providers. What does the dependency graph look like? And, can they live on their own? I'm happy to put this wherever makes sense. :)

Assets/MixedRealityToolkit/Utilities/StandardShader/StandardShaderUtility.cs
Cameron-Micka: Since the shader and it's utilities live in core, I believe this also needs to stay in core since it's a dependency.

Assets/MixedRealityToolkit/StandardAssets/Shaders/MixedRealityStandard.shader
Cameron-Micka: I made the names more generic since they now can be mutliple things. But, I agree. I'll comment what they can be in different circumstances.

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Cameron-Micka: I think @thalbern is in the process of refactoring BoundingBox. So might make sense to add then?

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
Cameron-Micka: Will do!

Assets/MixedRealityToolkit/StandardAssets/Shaders/MixedRealityStandard.shader
Troy-Ferrell: yea or at least acknowledge that these are uv slots for special components, depending on the feature

Assets/MixedRealityToolkit.SDK/Features/UX/Scripts/BoundingBox/BoundingBox.cs
Troy-Ferrell: I mean for the ScaleMeshEffect

Cameron-Micka: Ah gotcha, yeah I can add a blub to the MRTK/Shader page.

Cameron-Micka:
Now that https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5781 is in. I'm going to pull in those changes and apply these fixes to the example content from that PR.

AdamMitchell-ms:
Thanks for updating the Unity UI Pressable button prefabs!

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
Cameron-Micka: I just tested this on PressableButtonsTestCanvas by scaling the PressableButtons RectTransform. Luckily everything still works I believe due to the way vertices get rebuilt.

Assets/MixedRealityToolkit.Services/InputSystem/Utilities/ScaleMeshEffect.cs
julenka: You can figure this out by looking at the asmdef files. It looks like SDK depends on a bunch of stuff, but services usually depends only on MixedRealityToolkit.

I think if it isn't a core MRTK service like the input system, spatial mesh system, or boundary system, it should go into SDK.

Cameron-Micka: Sounds good! I'll move this into SDK.

Cameron-Micka: We spoke offline, going to keep this script in Services because MixedRealityCanvasInspector depends on it.

PR #5857, opened 2019-09-03
Decouple InteractableThemes Part 1/N

Troy-Ferrell:
## Overview
This change is the first major step in refactoring Interactable Themes to fix a multitude of issues including the following:
- Decouple Themes from Interactable so they are separate features (i.e Themes should be self-managed)
- Decouple inspector hard-coded management from theme/interactable execution and management
- Simplify code complexity to improve robustness and maintainability
- Other Bug Fix
=> Shader types and properties can now be deterministically defined
=> Changing states on a theme will populate ThemeProperties with correct number of default values
=> Theme scriptableobject assets weren't actually saved correctly because the code would fill in the holes but not update the scriptableobject

There are still some loose ends in this feature as well as bugs filed on github to be addressed, hence part 1/N. Furthermore, tests and documentation need to be updated at a future part. Majority of changes here are automatic updates in the theme asset files

There are also some material asset changes due to some MRTK property changes in a different PR

Breaking Changes:

Great care was taken to ensure MRTK consumers would not break with absorption of this new theme model. The vast, vast majority of changes were applied to internal classes and constructs used by the system that should largely be unknown to the community(i.e see #1). Individuals who have created their own theme engines (i.e extending InteractableThemeBase) will have to register one new function.
Finally, there is an important property for shaders that must be migrated. I added code as a temp workaround to handle this at the inspector and run-time levels. However, users should use the editor migrate button to auto-handle all Assets (MRTK > Migration > Update Themes)

1) Rename of classes below
=> Anyone using these classes directly will be broken. These classes though are 90% internal framework constructs and thus impact on community will be minimal. Furthermore, functionality is not dramatically changed and few consumers will largely just need to fix by copy/pasting new names

2) Inherited classes of InteractableThemeBase now requires GetDefaultThemeDefinition() method which must be properly filled out by custom implementation.
-> Furthermore, new properties AreShadersSupported (false default) and IsEasingSupported (true default) need to be overridden by custom implementations if they want different default behavior

3) Simplified shader properties in Theme Property (i.e ThemeStateProperty)
=> Old variables are no serialized protected to disuade consumers from leveraging them but not break existing customers during migration.
=> Migration is auto-handled in the inspector for users. At runtime, values will temporarily be copied over correctly
=> Users should utilize the menu item MRTK > Migration > Update Themes to correctly update all scriptableobjects in their Assets

The Details:

Theme renamed to ThemeContainer
-> Removed Name property, redundant with scriptableobject.name
-> Removed CustomSettings property
[...]

PR #5889, opened 2019-09-05
Interactable Refactor Part 2/N

Troy-Ferrell:
## Overview
Continuation from #5857 

**Note:** This is merging into a feature branch, not MRTK_Dev

- Moved Theme-related files from Interactable to new VisualThemes feature folder
- Elimnated dependency on InteractableTypeFinder, in favor of TypeExtensions
- Update properties and helper methods of InteractableEvent
- Extrapolated inspector code for InteractableEvent to it's own utility class, simplified change/remove structure too
- Fixed InteractableThemeInspector to update if fields change in the targeted class
- Delete InteractableTypeFinder because utilization was phased out in favor of simpler approaches (created new TypeCacheUtility instead)
- Cleaned up States scriptableobject class and StateInspector to simplified and more robust model

## Changes
- Fixes: #5652 

## Verification

>
>


Assets/MixedRealityToolkit/Inspectors/Utilities/InspectorFieldsUtility.cs
```diff
@@ -144,6 +186,32 @@ public static void UpdatePropertySettings(SerializedProperty prop, int type, obj
             }
         }
 
+        public static List<InspectorFieldData> GetInspectorFields(System.Object target)
+        {
+            List<InspectorFieldData> fields = new List<InspectorFieldData>();
+            Type myType = target.GetType();
+
+            foreach (PropertyInfo prop in myType.GetProperties())
```
julenka: btw, I'm not convinced this actually works

julenka: as in, the code didn't seem to be working for me before the change. I had properties that I wanted to render, and they were not actually getting updated.

Troy-Ferrell: Yes that bug is exactly #5652 

I added functionality to make it work. The one edge case that is not accounted for is you changing any of the property names will be unrecoverable. There is no FormallySerializedAs in this situation. 

Regarding moving the visual themes to a different folder, no it won't affect customers because the guids are still the same, the classes haven't changed (at least the theme engine ones), the namespaces haven't changed, etc. I don't see how they could be impacted.

test

Issue #5879, opened 2019-09-04
Enable research mode


Weasy666:
I'm currently working on a project in which i need access to the raw sensor stream. To gain access there are some steps that need to be done in Package.appxmanifest and Assembly-CSharp.csproj, but these files are replaced with every 'Build Unity Project'. So...i added two checkboxes to the build window, which let the developer choose if the two file will be modified to allow sensor access and to allow unsafe code in the build Assembly-CSharp project.
![image](https://user-images.githubusercontent.com/11967542/64291786-435a0700-cf69-11e9-9fa5-724461ae81e6.png)

Now my question is if there is interest in something like this. I could start a pull request and add this. I'm not sure if a research mode exists for HoloLens 2, so i would look to make the checkboxes or at least the file modifying checkbox visible dependent on the target device.

wiwei:
@MenelvagorMilsom FYI

@Weasy666  This seems reasonable to me - note that there is already precedent in the build window for adding a capability in after the project is generated (see Appx Build Options -> Gaze Input Capability). There's some code there that already adds capabilities in (and makes sure we don't duply-add them).

If you have already done some stuff here, it would probably be cool to follow that existing pattern, but yeah, even if it's not something everyone is necessarily using, having it be an easy option for folks who need it would be great to have in the toolkit.

Weasy666:
I have seen the Gaze Input Capability, but in my workflow i am not building the Appx from inside the Unity Editor. I need to hide most of the code behind preprocessor directives, because the Windows APIs i need (SoftwareBitmap and so on) are not accessible in the editor project.
So i build the unity project, then open the built solution and do most of the C# work in the build/generated project.  And i think other people need to do it that way too….so the options should be available in the 'Unity Build Options' tab.

I think i can whip up a pull request with my changes tomorrow in the course of the day. There we can discuss the details of the change. Do you know if research mode exists for HoloLens 2?

Weasy666:
@wiwei pull request is up.
There are two things that need to be clarified. I asked them in the PRs comment.

wiwei:
Regarding research mode, I'm following up on that internally.

I'll take a look at the PR, thanks for sending that out!

wiwei:
@Weasy666, got confirmation from folks working on research mode that it's being worked on for HL2.

Weasy666:
Nice to hear and thanks for following up on that!
[...]

PR #5891, opened 2019-09-05
Add Research Mode and Unsafe Code to Unity Build options

Weasy666:
## Overview
Continuation from #5879

## Changes
As stated in the issue, this PR adds two checkboxes to the build windows which make it possible to enable Research Mode and/or Unsafe Code for builds that target UWP/WSA and HoloLens 1 as device. If other platforms or devices are targeted the checkboxes won't be visible.

## Questions
1. The current changes only enable this options for builds from Unity Editor. Should i also add these two as command line arguments?
2. Does this need some kind of UnitTests?


Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -408,6 +438,125 @@ public static void AddGazeInputCapability(IBuildInfo buildInfo)
...
+                ignNsAttribute = new XAttribute("IgnorableNamespaces", "rescap");
+                rootNode.Add(ignNsAttribute);
+            }
+
+            if (!ignNsAttribute.Value.Contains("rescap"))
+            {
+                ignNsAttribute.Value += " rescap";
+            }
+
+            // If the capabilities container tag is missing, make sure it gets added.
```
wiwei: Nit: Consider moving this (and the lines in AddGazeInputCapability) to a help function, since they do the same thing (i.e. ensuring and then returning the capabilities element).

Alternatively, it could also be good to have a function that's like AddCapability(string capabilityType, string capabilityName), where you could call it like:

```csharp
AddCapability(rootNode.GetDefaultNamespace()+"DeviceCapability", "gazeInput");
AddCapability(rescapNs + "Capability", "perceptionSensorsExperimental");
```

This should reduce some of the copyiage that occurs here and would simplify stuff a bit

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -408,6 +438,125 @@ public static void AddGazeInputCapability(IBuildInfo buildInfo)
...
+            }
+        }
+
+        /// <summary>
+        /// An overload of AddResearchModeCapability that will read the AppX manifest
+        /// from the build output and update the manifest file with the rescap aka
+        /// research mode capability.
+        /// </summary>
+        /// <param name="buildInfo">An IBuildInfo containing a valid OutputDirectory</param>
+        public static void AddResearchModeCapability(IBuildInfo buildInfo)
```
wiwei: Also a nit: With the change as-is, I believe that the manifest is cracked open twice. I might do a little refactoring here after - it would be worth probably just opening the manifest once (i.e. read once, write once), and making it so that we use the function above that are actually... yeah, only defined as being public for testing purposes (I think it would probably be okay do delete that portion of the comment at that time).

To be a bit clearer, it might be better to have something in UwpPlayerBuildTools.cs that instead of:

```csharp
  var uwpBuildInfo = innerBuildInfo as UwpBuildInfo;
                    Debug.Assert(uwpBuildInfo != null);
                    if (uwpBuildInfo.GazeInputCapabilityEnabled)
                    {
                        UwpAppxBuildTools.AddGazeInputCapability(uwpBuildInfo);
                    }

                    if (EditorUserBuildSettings.wsaSubtarget == WSASubtarget.HoloLens)
                    {
                        if (uwpBuildInfo.ResearchModeCapabilityEnabled)
                        {
                            UwpAppxBuildTools.AddResearchModeCapability(uwpBuildInfo);
                        }

                        if (EditorUserBuildSettings.wsaGenerateReferenceProjects && uwpBuildInfo.AllowUnsafeCode)
                        {
                            UwpAppxBuildTools.AllowUnsafeCode(uwpBuildInfo);
                        }
                    }
```

Is more like:

```csharp
  var uwpBuildInfo = innerBuildInfo as UwpBuildInfo;
                    Debug.Assert(uwpBuildInfo != null);

void UpdateManifest(UwpBuildInfo uwpBuildInfo)
{
string manifestFilePath = GetManifestFilePath(buildInfo);
            if (manifestFilePath == null)
            {
                throw new FileNotFoundException("Unable to find manifest file");
            }

            var rootElement = XElement.Load(manifestFilePath);

if (uwpBuildInfo.GazeInputCapabilityEnabled)
{
            UwpAppxBuildTools.AddGazeInputCapability(rootElement);
}

                    if (EditorUserBuildSettings.wsaSubtarget == WSASubtarget.HoloLens && uwpBuildInfo.ResearchModeCapabilityEnabled)
                    {
                            UwpAppxBuildTools.AddResearchModeCapability(uwpBuildInfo);
                    }
            rootElement.Save(manifestFilePath);
}
void UpdateAssemblyCSharp(UwpBuildInfo  uwpBuildInfo)
{
if (EditorUserBuildSettings.wsaGenerateReferenceProjects && uwpBuildInfo.AllowUnsafeCode)
                        {
                            UwpAppxBuildTools.AllowUnsafeCode(uwpBuildInfo);
                        }
}

// And then in the actual function call:
UpdateManifest(uwpBuildInfo);
UpdateAssemblyCSharp(uwpBuildInfo);

```

Obviously with better formatting than the mess of indentation that I have above

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -408,6 +438,125 @@ public static void AddGazeInputCapability(IBuildInfo buildInfo)
...
+        /// Note that this function is only public to poke a hole for testing - do not
+        /// take a dependency on this function.
+        /// </remarks>
+        public static void AllowUnsafeCode(XElement rootNode)
+        {
+            foreach (XElement propertyGroupNode in rootNode.Descendants(rootNode.GetDefaultNamespace() + "PropertyGroup"))
+            {
+                if (propertyGroupNode.Attribute("Condition") != null)
+                {
+                    var allowUnsafeBlock = propertyGroupNode.Element("AllowUnsafeBlock");
```
wiwei: Should this:

```csharp
var allowUnsafeBlock = propertyGroupNode.Element("AllowUnsafeBlock");
```

Actually be

var allowUnsafeBlock = propertyGroupNode.Element("AllowUnsafeBlock**s**");

Notice the "s" in "Blocks".

When I search for this I see "did you mean AllowUnsafeBlocks" instead:

https://www.google.com/search?q=AllowUnsafeBlock

wiwei:
> 1. The current changes only enable this options for builds from Unity Editor. Should i also add these two as command line arguments?

I think that we can be consistent here with what happened for the gaze capability - we can add it when we need it or when other's need it. If you find that you really want to use this in batch mode then doing it now its probably good, if not, there's no rush.

> 2. Does this need some kind of UnitTests?

Yes thank you SO much for asking :)

It would be great to get some a couple of tests here:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/Assets/MixedRealityToolkit.Tests/EditModeTests/Core/Utilities/BuildAndDeploy/UwpAppxBuildToolsTest.cs#L71

i.e. one that tests the new capability that gets added, and then another that tests the csproj modification.

Basically you can just copy paste a dummy XML string blob for the csproj (probably some minimal amount of XML) and then just verify that after calling your new function, that the right tags are getting added (and not getting duply added if you call it twice!).

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
Weasy666: Yes, absolutely! Was a typo. A few lines down (L543) it is correct. That's what you get for once not copy&pasting 😅

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
Weasy666: Good idea. I will move this to a helper function.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
Weasy666: Yeah, currently it is opened twice. I will think about it. Hm...there is already a method for updating the manifest `UpdateAppxManifest` which does a little bit more than just adding capabilities. Is it ok using that method instead of writing another `UpdateManifest` which just updates the capabilities? Or should i extract that capabilites part from `UpdateAppxManifest` and use the following method which i have already added while addressing your previous comment.
```csharp
public static void AddCapabilities(IBuildInfo buildInfo)
{
    string manifestFilePath = GetManifestFilePath(buildInfo);
    if (manifestFilePath == null)
    {
        throw new FileNotFoundException("Unable to find manifest file");
    }

    var rootElement = XElement.Load(manifestFilePath);
    var uwpBuildInfo = buildInfo as UwpBuildInfo;

    Debug.Assert(uwpBuildInfo != null);
    if (uwpBuildInfo.GazeInputCapabilityEnabled)
    {
        AddGazeInputCapability(rootElement);
    }

    if (EditorUserBuildSettings.wsaSubtarget == WSASubtarget.HoloLens)
    {
        if (uwpBuildInfo.ResearchModeCapabilityEnabled)
        {
            AddResearchModeCapability(rootElement);
        }

        if (EditorUserBuildSettings.wsaGenerateReferenceProjects && uwpBuildInfo.AllowUnsafeCode)
        {
            AllowUnsafeCode(rootElement);
        }
    }

    rootElement.Save(manifestFilePath);
}
```

Weasy666:
Ok UnitTests are done. Hope i didn't add to many of them. 😅

Weasy666:
There is one question left. It is about the `EditorUserBuildSettings.wsaSubtarget == WSASubtarget.HoloLens` condition for displaying the checkboxes and adding the capability to the manifest. I assume that this condition is true for both HoloLenses and as HoloLens 2 currently seems to not have a research mode, i think this needs to somehow be addressed.
Do you know of a way to differentiate both devices at that point in the build pipeline?

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/Utilities/BuildAndDeploy/UwpAppxBuildToolsTest.cs
```diff
@@ -119,5 +120,174 @@ private static void AssertSingleGazeInputCapability(XElement rootElement)
...
+        {
+            XElement rootElement = XElement.Parse(TestManifest);
+            UwpAppxBuildTools.AddResearchModeCapability(rootElement);
+            AssertSingleResearchModeCapability(rootElement);
+        }
+
+
+
+        /// <summary>
+        /// Validates that the AddGazeInputCapability will also add the
```
wiwei:
```suggestion
        /// Validates that AddResearchModeCapability will also add the
```

wiwei: I added an extra the in my original comment too

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/Utilities/BuildAndDeploy/UwpAppxBuildToolsTest.cs
```diff
@@ -119,5 +120,174 @@ private static void AssertSingleGazeInputCapability(XElement rootElement)
...
...E_CLOUD_HUB;ENABLE_CLOUD_PROJECT_ID;ENABLE_CLOUD_SERVICES_USE_WEBREQUEST;ENABLE_CLOUD_SERVICES_UNET;ENABLE_CLOUD_SERVICES_BUILD;ENABLE_CLOUD_LICENSE;ENABLE_EDITOR_HUB;ENABLE_EDITOR_HUB_LICENSE;ENABLE_WEBSOCKET_CLIENT;ENABLE_DIRECTOR_AUDIO;ENABLE_DIRECTOR_TEXTURE;ENABLE_TIMELINE;ENABLE_EDITOR_METRICS;ENABLE_EDITOR_METRICS_CACHING;ENABLE_MANAGED_TRANSFORM_JOBS;ENABLE_MANAGED_ANIMATION_JOBS;INCLUDE_DYNAMIC_GI;INCLUDE_GI;ENABLE_MONO_BDWGC;RENDER_SOFTWARE_CURSOR;INCLUDE_PUBNUB;ENABLE_VIDEO;ENABLE_CUSTOM_RENDER_TEXTURE;ENABLE_LOCALIZATION;PLATFORM_METRO;UNITY_METRO;UNITY_METRO_API;UNITY_WSA;UNITY_WINRT_10_0;PLATFORM_WINRT;UNITY_WINRT;ENABLE_WINRT_PINVOKE;ENABLE_LAZY_METHOD_CACHING;ENABLE_MOVIES;ENABLE_CRUNCH_TEXTURE_COMPRESSION;ENABLE_NETWORK;ENABLE_EVENT_QUEUE;UNITY_PLATFORM_THREAD_TO_CORE_MAPPING;ENABLE_SUBSTANCE;PLATFORM_SUPPORTS_ADS_ID;ENABLE_UNITYWEBREQUEST;ENABLE_CLOUD_SERVICES;ENABLE_CLOUD_SERVICES_ADS;ENABLE_CLOUD_SERVICES_ANALYTICS;ENABLE_CLOUD_SERVICES_PURCHASING;ENABLE_CLOUD_SERVICES_CRASH_REPORTING;CURL_STATICLIB;ENABLE_VR;ENABLE_AR;PLATFORM_OVERRIDES_APPLICATION_PACKAGE_NAME;ENABLE_DOTNET;NETFX_CORE;ENABLE_WINMD_SUPPORT;WINDOWS_UWP;UNITY_UWP;UNITY_WSA_10_0;</DefineConstants>
```
wiwei:
```suggestion
                    <DefineConstants>DEBUG;</DefineConstants>
```

wiwei: Nit: I think that the contents within this don't actually matter for the test, but they do make the line SUPPPPPPPPPPPPPPPPPPPPER long and also make future readers wonder if this is actually necessary (i.e. how do I generate this super long string of things and why is mine different and does it matter, etc)

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -359,15 +389,53 @@ private static string GetMSBuildLoggingCommand(string logDirectory, string logFi
...
+            if (EditorUserBuildSettings.wsaSubtarget == WSASubtarget.HoloLens)
+            {
+                if (uwpBuildInfo.ResearchModeCapabilityEnabled)
+                {
+                    AddResearchModeCapability(rootElement);
+                }
+
+                if (EditorUserBuildSettings.wsaGenerateReferenceProjects && uwpBuildInfo.AllowUnsafeCode)
+                {
+                    AllowUnsafeCode(rootElement);
```
wiwei: I think that rootElement in this location is the AppX manifest, not the Assembly-CSharp manifest (which is the type of manifest that unsafe code has to be on) - I think that this may have gotten a lost in the refactor, and I can definitely take a little bit of blame here for not being a little clearer in my request!

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
```diff
@@ -104,6 +104,10 @@ private enum PlatformToolset
 
         private readonly GUIContent uninstallLabel = new GUIContent("Uninstall First", "Uninstall application before installing");
 
+        private readonly GUIContent researchModeCapabilityLabel = new GUIContent("Enable Research Mode", "Enables research mode of HoloLens 1. This allows access to raw sensor data.");
+
+        private readonly GUIContent allowUnsafeCode = new GUIContent("Allow Unsafe Code", "Modify 'Assembly-CSharp.csproj' to allow use of unsafe code.");
```
wiwei:
```suggestion
        private readonly GUIContent allowUnsafeCode = new GUIContent("Allow Unsafe Code", "Modify 'Assembly-CSharp.csproj' to allow use of unsafe code. Do not use this in production.");
```

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
wiwei: Yep, this is good! I think that there's one more thing that needs to be addressed here (which I opened a new comment for, so I'm resolving this one now)

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
wiwei: i.e. to be clear, I think that AllowUnsafeCode should probably be like what it was before, and be called as a same-level sort of sibling do AddCapabilities (where AddCapabilites is called).

```diff
-                    if (uwpBuildInfo.GazeInputCapabilityEnabled)
-                    {
-                        UwpAppxBuildTools.AddGazeInputCapability(uwpBuildInfo);
-                    }
+                    UwpAppxBuildTools.AddCapabilities(uwpBuildInfo);
+                    UwpAppxBuildTools.AllowUnsafeCode(uwpBuildInfo);

```

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
Weasy666: Hm...aren't there some cases were unsafe code is needed in production? So would it better to say "..unsafe code. Use this only in production if you know exactly what you are doing."  
And should i delete the 1 in line 107 from "...mode of HoloLens 1. This..."?

Assets/MixedRealityToolkit.Tests/EditModeTests/Core/Utilities/BuildAndDeploy/UwpAppxBuildToolsTest.cs
Weasy666: Yeah, this does make absolute sense. I'll condense it down to only DEBUG;

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
Weasy666: yeah...that is some `refactor copy&paste past midnight`-error 😅 sorry about that.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
```diff
@@ -359,15 +389,53 @@ private static string GetMSBuildLoggingCommand(string logDirectory, string logFi
...
+            }
+
+            rootElement.Save(manifestFilePath);
+        }
+
+        /// <summary>
+        /// Adds a capability to the given rootNode, which must be the read AppX manifest from
+        /// the build output.
+        /// </summary>
+        /// <param name="rootNode">An <see cref="XElement"/> containing the AppX manifest from
```
wiwei: So unfortunately our docs validation system doesn't see these crefs as being valid (even though these are fairly well known things).

In order to avoid that, it might be good to just remove the <see cref="XElement" /> part and replace it with it with just XElement.

(Same for XName and XAttribute down below:

[19-09-16 04:37:36.346]Warning:[MetadataCommand.ExtractMetadata]Invalid cref value "!:XElement" found in triple-slash-comments for AddCapability, ignored.
[19-09-16 04:37:36.346]Warning:[MetadataCommand.ExtractMetadata]Invalid cref value "!:XName" found in triple-slash-comments for AddCapability, ignored.
[19-09-16 04:37:36.346]Warning:[MetadataCommand.ExtractMetadata]Invalid cref value "!:XAttribute" found in triple-slash-comments for AddCapability, ignored.)

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
wiwei: Yeah your point is a good one - I think I just want to call out that people need to be careful with using this (i.e. Be careful using this in production).

wiwei:
> Ok UnitTests are done. Hope i didn't add to many of them. 😅

They're great! I wish more people would do what you did!

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
Weasy666: Ok, will do that. Is there a GitHub repo of your docs validation system? So that i can open an issue...i think official Microsoft classes should be recognized by it.

wiwei: For this we just use docfx:

https://microsoft.github.io/MixedRealityToolkit-Unity/Documentation/Contributing/DevDocGuide.html

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_release/docfx.json

I would say it's a bug that we don't currently handle those other types, but in the interest of not blocking behind addressing that (i.e. it could be really easy or a realllllll huge can of worms) it's probably easier in this case just to use the non-cref based text.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
Weasy666: I added a method like `UpdateAppxManifest` that can be used as a "central" point for changes that need to be made in the project file.

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/UwpAppxBuildTools.cs
Weasy666: I got rid of them and will file an issue at dotnet/docfx that points to this pull request. Let's see what happens ^^

Assets/MixedRealityToolkit/Utilities/BuildAndDeploy/BuildDeployWindow.cs
Weasy666: Went with your last suggestion and put "Be careful using this in production." at the end of the tooltip.

PR #5892, opened 2019-09-05
Updating the MRTK/Standard shader render modes to match Unity's new terminology.

Cameron-Micka:
## Overview

One of our partners encountered confusion over the the rendering modes found on the MRTK/Standard shader. Since adding the rendering mode's to match Unity's terminology Unity has changed their terminology. So, this change simply brings us back to naming convention parity (this is not a breaking change, just a user facing naming swap).

I've also improved our documentation around rendering modes and tweaked the specular reflection formula to match the Unity/Standard shader.

MRTK_RenderingModes

## Verification
Try creating new materials and seeing if the selected rendering mode matches your expectations.

>
>
>
>


Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
```diff
@@ -16,25 +16,25 @@ public abstract class MixedRealityShaderGUI : ShaderGUI
     {
         protected enum RenderingMode
         {
-            Opaque,
-            TransparentCutout,
-            Transparent,
-            PremultipliedTransparent,
-            Additive,
-            Custom
+            Opaque = 0,
```
keveleigh: Does Unity provide any enums for theses values we could use, to always stay in sync and know when we're not?

Cameron-Micka: They do have these enums, but they are in an internal class. ☹️

keveleigh: 😢

Assets/MixedRealityToolkit/Inspectors/MixedRealityStandardShaderGUI.cs
```diff
@@ -593,9 +593,9 @@ protected void FluentOptions(MaterialEditor materialEditor, Material material)
                     materialEditor.ShaderProperty(borderLightUsesHoverColor, Styles.borderLightUsesHoverColor, 2);
                 }
 
-                if (mode == RenderingMode.TransparentCutout || mode == RenderingMode.Transparent ||
-                    (mode == RenderingMode.Custom && customMode == CustomRenderingMode.TransparentCutout) ||
-                    (mode == RenderingMode.Custom && customMode == CustomRenderingMode.Transparent))
+                if (mode == RenderingMode.Cutout || mode == RenderingMode.Fade || mode == RenderingMode.Transparent ||
```
keveleigh: It seems like this is adding a new case to this `if` (there were two on this line and now there are three):

```suggestion
                if (mode == RenderingMode.Cutout || mode == RenderingMode.Fade ||
```

Cameron-Micka: Yeah, I found that little bug when looking over the rename. It should be all 3 conditionals (not just the 2). But, thanks for double checking!

Assets/MixedRealityToolkit/Inspectors/MixedRealityShaderGUI.cs
```diff
@@ -16,25 +16,25 @@ public abstract class MixedRealityShaderGUI : ShaderGUI
...
+            Cutout = 1,
+            Fade = 2,
+            Transparent = 3,
```
davidkline-ms: Changing Transparent from 2 to 3 may break some materials as the dev would have chosen transparent and get fade after updating (i believe the int value of the enum is what is serialized)

keveleigh: I believe that's by-design. The names have changed, but the meanings/behaviors are the same both before and after `int`-wise.

`TransparentCutout` -> `Cutout`
`Transparent` -> `Fade`
`PremultipliedTransparent` -> `Transparent`

Cameron-Micka: Yes @keveleigh is correct! It’s confusing but the names have been remapped by Unity. If Unity really wanted to get pedantic the could have renamed Transparent to Translucent, but that may have been too confusing to users.

Issue #5775, opened 2019-08-25
ManipulationHandler is sluggish when moving in y-direction (HL v1)


ei2kpi-ptc:
## Overview
When I use ManipulationHandler (MRTK 2.0.0) with a HoloLens v1 on a GameObject, I can target the object with my gaze, press down with my hand and can move it around.

When I move my hand right / left or forward / backward (relative to my head) it seems to work as expected but when I move my hand up / down, the object moves a lot more sluggishly

## Expected Behavior
Movement in all axes should work the same. I remember there being a HandGrabbable.cs script in HoloToolkit that worked great!

## Steps to reproduce
1. Set up MRTK with default HoloLens 1 profile
2. Create a GameObject
3. Drop ManipulationHandler on it with the following settings:
![image](https://user-images.githubusercontent.com/44373229/63644038-a1efcd00-c6ad-11e9-92fd-4dd4f95e7693.png)
4. Hit Play, and use the input simulation service to target the object by gaze and then try to manipulate it's position by using Space + the mouse to tap on the object and then move the mouse to move the "hand"
5. Moving the mouse left and right works great, moving it up and down is really sluggish moving the mouse forward / backward (rolling mouse wheel) works great too

## Unity Editor Version
Unity 2019.2.0f1

## Mixed Reality Toolkit Release Version
v2.0.0

clementsauvard:
I can also confirm that this is a bit annoying compared to RC2.1 To me it seems like a sensitivity issue, 
moving the object to the left/right or forward/backward look the same compared to RC2.1, but when moving it up or down it doesn't move as far as i would expect it to go.

See below:
![SmartSelect_20190829-132816_Drive](https://user-images.githubusercontent.com/15730879/63939703-2f456f80-ca68-11e9-8130-1894a46a7dc5.gif)


This makes me think about X-axis and Y-axis mouse sensitivity settings found in certain FPS games. 
It could be interesting to add such parameters to the ManipulationHandler script, although i don't know if it would create issues with other parameters like Smoothing.

wiwei:
@julenka, @thalbern

It looks like the difference between GGV (HL1 style) and hand ray (HL2 style) is that the rotation fields are actually critical to the up/down functionality:

```csharp
public Vector3 Update(MixedRealityPose pointerCentroidPose, Quaternion objectRotation, Vector3 objectScale, bool isNearMode, bool usePointerRotation, MovementConstraintType movementConstraint)
        {
[...]

PR #5896, opened 2019-09-06
Ux/sluggish ggv yaxis

MenelvagorMilsom:
## Overview
As part of [this PR](https://github.com/microsoft/MixedRealityToolkit-Unity/pull/5645), I changed how GGV rotation worked. As part of my change, I zeroed the y value of the look rotation vector. Before I did this, moving your hand up and down would impart pitch rotation on manipulated objects in much the same way that moving you hand left and right would impart yaw rotation. I wasn't sure at the time if this was something we wanted. 

These changes remove this line where the y value is zeroed. As a result, vertical manipulation no longer feels sluggish in GGV, but you will also impart some pitch rotations on the manipulated object. It is possible to counteract this by setting a y-axis only rotation constraint on manipulation handler.

### Before
zeroY

### After
noZeroY

## Changes
- Fixes: #5775.


wiwei:
Nice! Any chance we can add a test case for this (to show that given a pose change in GGV hand or Articulated hand, the stuff should still be roughly the same in the Y axis).

PR #5897, opened 2019-09-06
Fix Appx building from Unity player completion dialog not using correct settings.

julianloehr-kg:
## Overview

The `Build Appx` button in the Unity build completion dialog did not trigger the same build as the `Build Appx` button in the Appx build window tab. Specifically most build preferences were not taken into account. 

I compared both `UwpBuildInfo` from `BuildPlayer` and `BuildAppx` and added the missing settings to the former one. Though I am not sure if there are still hidden preferences that must be matched as well. 

## Verification

Start a Unity player build, then in the completion dialog use the `Build Appx` button to build the Appx right away. The Appx should be build using the correct settings from the Appx build tab.


wiwei:
This change lgtm, I think going forward I'll have some refactoring to do here to prevent out-of-syncness from happening again (i.e. if we add another flag or Unity adds another flag)

